PR #1084, opened 2022-08-11
break: secure file upload and download (public and private)

gunar:
Card: https://trello.com/c/cS7bfBXL/2019-restrict-access-to-user-uploaded-files-to-bops-uniform-only

- prevents user uploads from being publicly accesible
- makes all S3 routes go through the API
- adds a new API Token concept to only allow BoPS to download user files
- side-effect: prevents users from downloading their own files


gunar:
Rebased and ready for review.

gunar:
Hey @jessicamcinchak, quick question. This PR requires adding a new env var. I've added it to `s3://pizza-secrets/api.env.test` which makes the CI tests pass. Now as for the Pizza, what would be the right place to put this `API_FILE_KEY` environment variable? Here maybe? https://github.com/theopensystemslab/planx-new/blob/3d0b63fef2ba4a5c352fabf6f8ce855880415962/.github/workflows/pull-request-main.yml#L347

jessicamcinchak:
hey @gunar - without fully understanding this PR yet, I'd suggest adding the new variable to the root .env file as well as the api env.test and then it'll be available in the pizza's docker container and during local development.

any variables defined like L347 `X=${{ env.X }}` will be printed by Actions fyi; Github only masks values like `X=${{ secrets.X }}`. I'm generally aiming to decrease how many variables we need to keep in Github Secrets and use remote `root.env` file as source of truth for all environments using docker :slightly_smiling_face: 

lemme know if you have any issues with this approach though!

gunar:
That makes sense, and it worked! Thanks!

api.planx.uk/server.js
```diff
@@ -490,18 +500,63 @@ app.post("/download-application", async (req, res, next) => {
...
+
+  if (!filePath) {
+    return next({ status: 404, message: "file not found" });
+  }
+
+  try {
+    const { body, headers } = await getFileFromS3(filePath)
+
+    res.set(headers);
+    res.send(body);
```
jessicamcinchak, line 559: nit: I would personally love to see only endpoints in `server.js` and then add these methods/logic to existing `s3.js` file or a new file. same goes for test cases you've added to `server.test.js`. API does all the things and I've been findng sub-files helpful for mental compartmentalization & quicker investigation/troubleshooting

editor.planx.uk/src/@planx/components/Review/Public.test.tsx
```diff
@@ -103,3 +110,210 @@ it("should not have any accessibility violations", async () => {
...
+const mockLink = "my-file.png";
+const uploadedPlanUrl = "http://someurl.com/whjhnh65/plan.png";
+
+beforeEach(() => {
+  global.URL = {
+    createObjectURL: jest.fn(() => mockLink),
+  } as any;
+});
+
+it("should render file upload link", async () => {
```
jessicamcinchak, line 123: nit: "link" feels a bit confusing here because it's an important change in this PR that these are actually _NOT_ hyperlinks anymore in the sense of HTML tags. can "link" be updated to "filename" or something similar please?

editor.planx.uk/src/components/ImagePreview.tsx
```diff
@@ -0,0 +1,8 @@
+import { useFileUrl, UseFileUrlProps } from "@planx/components/shared/hooks";
+import React from "react";
+
+export default function ImagePreview(props: UseFileUrlProps) {
+  const { fileUrl } = useFileUrl(props);
+
+  return <img src={fileUrl} alt="" />;
```
jessicamcinchak, line 7: nit: I think empty `alt` here is probably an accessibility violation. i wonder if there's a simple standard template we coud use like `Preview of uploaded file: ${fileUrl}` ?

infrastructure/application/index.ts
```diff
@@ -279,6 +280,9 @@ export = async () => {
     sslPolicy: "ELBSecurityPolicy-TLS-1-2-Ext-2018-06",
     certificateArn: certificates.requireOutput("certificateArn"),
   });
+  const fileApiKey = new random.RandomPassword("file-api-key", {
```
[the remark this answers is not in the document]
jessicamcinchak: Can these specific rotation docs also perhaps be added to the Pulumi/infrastructure/application README? Definitely a useful one I don't want to lose track of in this PR once merged.

jessicamcinchak (commented):
thanks for all your work & patience on this one, apologies it's been a particularly slow review process as this one just touches so much & requires more cross-project communication than usual! 

i'm happy with the code & went through a full test LDC service: file upload components, review page, and CSV all look good - but both submissions to Uniform & BOPS failed (you can [see logs here](https://hasura.1084.planx.pizza/console/events/one-off-scheduled-events/processed)). Once file conflicts are resolved here, I'm happy to try again - I think once we have a successful staging submission to both BOPS & Uniorm and can confirm that they are able to pick up the files as expected then this could merge :+1:

gunar:
Next steps:

1. Rebase
2. Submit application from Pizza to BOPS Staging
3. Check that it was submitted correctly to Uniform
3. Give BOPS token, ask them to add it to the their request's headers, confirm it worked
4. Merge :tada:

gunar:
Pizza isn't building. I'll investigate.

gunar:
It's Friday I'm in… a ready-for-review state :joy: :musical_note:

api.planx.uk/s3/getFile.ts
```diff
@@ -0,0 +1,27 @@
+import S3 from "aws-sdk/clients/s3";
```
gunar, line 1: Worth noting that I had to change this otherwise `tsc` would try to compile the whole sdk and the GH Action worker would run out of memory.

```
- import { S3 } from "aws-sdk";
+ import S3 from "aws-sdk/clients/s3";
```

In a separate PR I'll suggest increasing the maximum amount of memory available for the worker (just in case).

gunar:
Actually, we already had a plan for this PR didn't we? Sorry I forgot.

@jessicamcinchak on Tuesday when you're back do you mind submitting an application from this Pizza into BOPS Staging? Happy to learn how to do it myself too but at this point I just want this PR to be over with :sweat_smile:

api.planx.uk/s3/getFile.ts
gunar: Worth noting about the "worth noting" above that the Docker build would fail silently (ie the error message was there but it would return a Linux zero code) which made it hard for me to find the root cause at first.

gunar:
[Restrict access to user-uploaded files to BOPS/Uniform only](https://trello.com/c/cS7bfBXL/2019-restrict-access-to-user-uploaded-files-to-bops-uniform-only)

infrastructure/application/index.ts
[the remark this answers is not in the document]
jessicamcinchak: additional question here in the context of BOPS (I was trying to share the secret with the devs during the call today): doesn't seem like `pulumi config get file-api-key --stack staging` here works as normal - is there an alternative way to get this?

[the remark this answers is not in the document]
gunar: I'm exporting this variable so that would be `pulumi stack output --show-secrets fileApiKeys`.

However, this isn't staging. This is a pizza, so the secret is hard-coded.

PR #1143, opened 2022-09-02
feat: sync production data to staging weekly

gunar:
Continuation of #913 and #1123:

- Pulls published_flows from production (the last 2 published flows per flow)
- The fetching of flows is now spread across multiple requests as to avoid Hasura running out of memory
- Requests run in parallel but are throttled

### To test it:

```
$ docker compose down --volumes --remove-orphans
$ docker compose up --build
$ # Wait for the seeding script to finish
$ # check that your local database now containes published flows from production
```

### DIFF
--- a/scripts/seed-database/upsert-production-flows.js
+++ b/scripts/seed-database/upsert-production-flows.js
@@ -1,4 +1,9 @@
-const { GraphQLClient } = require("graphql-request");
+import groupBy from "lodash/groupBy";
+import cloneDeep from "lodash/cloneDeep";
+import { GraphQLClient } from "graphql-request";
+import pThrottle from 'p-throttle';
+const args = process.argv.slice(2);
+
 const PRODUCTION_GRAPHQL_URL = 'https://hasura.editor.planx.uk/v1/graphql';
 const LOCAL_GRAPHQL_URL = process.env.HASURA_GRAPHQL_URL;
 const LOCAL_GRAPHQL_ADMIN_SECRET = process.env.HASURA_GRAPHQL_ADMIN_SECRET;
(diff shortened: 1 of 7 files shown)


scripts/seed-database/upsert-production-flows.js
```diff
@@ -50,65 +60,197 @@ const LOCAL_GRAPHQL_ADMIN_SECRET = process.env.HASURA_GRAPHQL_ADMIN_SECRET;
...
+    const {
+      false: teamsToUpsertById,
+      true: teamsToUpsertBySlug
+    } = groupBy(productionTeams, (team) => localTeamsSlugs.includes(team.slug));
+
+    const {
+      false: flowsToUpsertById,
+      true: flowsToUpsertBySlug
+    } = groupBy(
+      productionFlows.map(flow => ({ ...flow, version: 1, })),
```
jessicamcinchak, line 94: :+1: this is clever and cleaner than querying the most recent operation per flow! is it worth having a brief code comment here too as a reminder that that flow version is overwritten to match the operation version?

scripts/seed-database/upsert-production-flows.js
```diff
@@ -50,65 +60,197 @@ const LOCAL_GRAPHQL_ADMIN_SECRET = process.env.HASURA_GRAPHQL_ADMIN_SECRET;
...
+    seq: 1,
+    src: "1",
+    create: {
+      data: {
+        edges: [],
+        nodes: {}
+      },
+      type: "http://sharejs.org/types/JSONv0"
+    }
+  },
```
jessicamcinchak, line 254: for simplicity, I _think_ this will also work with just `data: {}`

jessicamcinchak (approved):
This is working for me locally now! Can you please just confirm that all CI steps are passing before merging - I'd guess that may require destroying & recreating vultr instance?

PR #1164, opened 2022-09-26
fix: Clones should be asked in correct order

DafyddLlyr:
image

Test flow on staging where this bug can be recreated - https://editor.planx.dev/testing/clone-order-example ❌ 
Pizza where this is fixed - https://1164.planx.pizza/testing/clone-order-example ✅ 

## Problem
Cloned nodes are not asked in the correct order within the flow

## Cause
The depth first sort algorithm used (`dfs()`) current creates a unique set of node ids. This means that clones will appear only once in this set. When `sortIdsDepthFirst()` is called, this results in the returned array for `upcomingCardIds()` "hoisting" cloned nodes to their first graph wide index position.

## Solution
 - Use an array in the `dfs()` algroithim, which means that clones will be listed in all graph positions
 - When calling `sortIdsDepthFirst()`, only sort nodes within the context of their locality within the graph to ensure clones are correctly positioned

## Issue / questions
This relies on `nodeIds[0]` being the first sequential node. This works in simple test flows, complex prod flows, and units tests. Can we actually make this assumption?

### DIFF
--- a/editor.planx.uk/package.json
+++ b/editor.planx.uk/package.json
@@ -90,7 +90,7 @@
     "@testing-library/user-event": "^14.4.3",
     "@turf/helpers": "^6.5.0",
     "@types/draft-js": "^0.11.9",
-    "@types/jest": "^29.0.0",
+    "@types/jest": "^27.4.3",
     "@types/jest-axe": "^3.5.4",
     "@types/lodash": "^4.14.182",
     "@types/marked": "^4.0.3",
--- a/editor.planx.uk/pnpm-lock.yaml
+++ b/editor.planx.uk/pnpm-lock.yaml
@@ -77,7 +77,7 @@ specifiers:
   '@testing-library/user-event': ^14.4.3
   '@turf/helpers': ^6.5.0
   '@types/draft-js': ^0.11.9
-  '@types/jest': ^29.0.0
+  '@types/jest': ^27.4.3
   '@types/jest-axe': ^3.5.4
   '@types/lodash': ^4.14.182
   '@types/marked': ^4.0.3
@@ -250,7 +250,7 @@ devDependencies:
   '@testing-library/user-event': 14.4.3_zqdhmevb64vvgf27bikfwyob6q
   '@turf/helpers': 6.5.0
   '@types/draft-js': 0.11.9
-  '@types/jest': 29.0.0
+  '@types/jest': 27.5.2
   '@types/jest-axe': 3.5.4
   '@types/lodash': 4.14.182
   '@types/marked': 4.0.3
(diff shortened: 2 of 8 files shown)


DafyddLlyr:
So this works for me @jessicamcinchak if you're ok to take another look and try it out?

Ideally I'd like to create a test case to describe the problem we were experiencing - no luck so far, will keep thinking on it over the weekend.

Essentially the issue seemed to be that the sort was running against the entire flow's node list instead of just the `nodeIds` we care about, leading to incorrect ordering. I'm not sure if this was because of a manual duplication of nodes sometime in the history of the flow - I wasn't able to pin it down. I'll take another look next week with a clear mind 🧠

jessicamcinchak (commented):
Hey! Finally revisting this one with fresh eyes and still getting some strange ordering behavior when I go through a full flow (Bucks LDC). Specifically, I think I'm seeing clones in the right order now when moving forward, but I think the "back" button is skipping clones?

Here's the simplest way I can currently reproduce: 
- go to https://1164.planx.pizza/opensystemslab/immunity-check (using editor side panel & graph highlight is useful here)
- pick "repair a window" from list of works, this takes you down alteration path
- see cloned question were the "works carried out > 4 years ago" as expected, select yes
- see question "have the works been completed", click "back" & skip back to "List the changes..."

I _think_ there still might be a bit of strange portal behavior too: 
- when working through a full Bucks LDC, I reached "upload floor plan" prompt _before_ any about the applicant questions - lemme know if you can't reproduce this one and I can try to isolate further!

DafyddLlyr:
So a way forward here has been agreed on the dev call 26/10/22.

 - Merge this PR with failing test skipped
 - Be aware that a new issue can now arise when navigating back from a clone
   - The hope here is that this is much less of an issue for user journeys, and is less likely to be encountered by an end user 
 - (Short / medium term) Revisit "Back" functionality when tackling upcoming "Navigation" tickets
   - How should back work? Is just going to the previous breadcrumb fine? 
 - (Medium term) Change how clones work! Do not use duplicated IDs, just a reference. 
   - Implementation could be similar to portals
   - Consider migration of existing flows
 
 I'll merge this once I'm back on 07/11/22 so that I'm about to deal with any issues which might pop up.

editor.planx.uk/package.json
```diff
@@ -90,7 +90,7 @@
     "@testing-library/user-event": "^14.4.3",
     "@turf/helpers": "^6.5.0",
     "@types/draft-js": "^0.11.9",
-    "@types/jest": "^29.0.0",
+    "@types/jest": "^27.4.3",
```
DafyddLlyr, line 93: I wanted to use the more semantically correct `test.failing()` which my linter was allowing / suggesting (https://jestjs.io/docs/api#testfailingname-fn-timeout). 

This didn't work as our version of jest is controlled by `create-react-app` and not our own `package.json` 👍

PR #1177, opened 2022-09-30
feat: Setup Save & Return email notifications on user prompt, not automatically

DafyddLlyr:
## What does this PR do?

Sets up email "Reminder" and "Expiry" event when a user explicitly hits the "Save and return to this application later" button. Previously this was done automatically once a session was started.

Event triggers used to be hit with an insert to `lowcal_sessions`, it will now be an update on `lowcal_sessions.has_user_saved`.

### To test
 - Start a new session on flow, e.g. XXXX
 - Once you have entered email x2, a `lowcal_session` record will be inserted
 - Event will **not** be set up in Hasura (Events > One-off Scheduled Events > Pending Events)
 - Back on the flow, hit "Save and return to this application later" 
 - A reminder and expiry event has now been set up in Hasura, with the timestamp matching when you hit "Save and return..." ✅

### DIFF
--- a/api.planx.uk/saveAndReturn/sendEmail.js
+++ b/api.planx.uk/saveAndReturn/sendEmail.js
@@ -10,7 +10,7 @@ const sendSaveAndReturnEmail = async (req, res, next) => {
         message: "Required value missing"
       });
     const response = await sendSingleApplicationEmail(template, email, sessionId);
-    response.hasUserSaved = setupEmailEventTriggers(sessionId);
+    response.hasUserSaved = await setupEmailEventTriggers(sessionId);
     return res.json(response);
   } catch (error) {
     return next({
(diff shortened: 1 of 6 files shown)


DafyddLlyr:
[Only trigger reminder emails if user has clicked 'Save' at least once](https://trello.com/c/7Ve0ACB4/2107-only-trigger-reminder-emails-if-user-has-clicked-save-at-least-once)

jessicamcinchak:
Just started reviewing - but here's one quirk to share upfront: I can't seem to successfully "resume" my existing session on this pizza. Can you see if you can recreate?

DafyddLlyr:
@jessicamcinchak This is the classic Pizza "Unable to find a published version of this flow" error - will be glad to see the back of this when https://github.com/theopensystemslab/planx-new/pull/1143 is merged 💪 

I've published the Bucks LDC flow now - if you start a new app and try resuming it should work.

jessicamcinchak (approved):
Looks good to me! And pizza works like a charm now, thanks for reminder about that publishing quirk :slightly_smiling_face:

PR #1178, opened 2022-09-30
fix: add status to pay notifications

jessicamcinchak:
see thread https://opensystemslab.slack.com/archives/C0405SQDP8B/p1663263727756039

we are logging pay notifications regardless of payment status, and we pushed a manual submission for a failed payment.

https://docs.payments.service.gov.uk/api_reference/#payment-status-meanings


DafyddLlyr (approved):
Brill 👍

PR #1179, opened 2022-09-30
fix(ci): Use unique keys for React and Storybook build cache

DafyddLlyr:
image

Fixes issue where only Storybook build was getting passed to Vultr, as a result of [the recent parallelization of CI jobs ](https://github.com/theopensystemslab/planx-new/pull/1154)

I've tried a few (now squashed) commits on this branch to make sure the cache is behaving as expected 👍

### DIFF
--- a/.github/workflows/pull-request-main.yml
+++ b/.github/workflows/pull-request-main.yml
@@ -170,7 +170,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ./${{ env.EDITOR_DIRECTORY }}/build
-          key: ${{ runner.os }}-${{ hashFiles('editor.planx.uk/**', '!editor.planx/build/**') }}
+          key: ${{ runner.os }}-${{ hashFiles('editor.planx.uk/**', '!editor.planx/build/**') }}-react
       - uses: pnpm/action-setup@v2.2.2
         if: steps.cache-react-build-assets.outputs.cache-hit != 'true'
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request-main.yml
```diff
@@ -392,13 +392,13 @@ jobs:
...
           path: ./${{ env.EDITOR_DIRECTORY }}/build
-          key: ${{ runner.os }}-${{ hashFiles('editor.planx.uk/**', '!editor.planx/build/**') }}
+          key: ${{ runner.os }}-${{ hashFiles('editor.planx.uk/**', '!editor.planx/build/**') }}-react
       - name: Download Storybook build assets
         id: cache-storybook-assets
         uses: actions/cache@v3
         with:
           path: ./${{ env.EDITOR_DIRECTORY }}/build
-          key: ${{ runner.os }}-${{ hashFiles('editor.planx.uk/**', '!editor.planx/build/**') }}
+          key: ${{ runner.os }}-${{ hashFiles('editor.planx.uk/**', '!editor.planx/build/**') }}-storybook
```
jessicamcinchak, line 401: should this match the change on L211? so `!editor.planx/build/storybook/**`??

jessicamcinchak (approved):
Thanks for this fix! One question, but CI links all working as expected :+1:

.github/workflows/pull-request-main.yml
DafyddLlyr: Good catch - L211 is me not cleaning up after myself properly from a previous commit sorry! 

Storybook is already build in the `/build/storybook` directory (in the Editor's `package.json` script `build-storybook`). We actually want Storybook to rebuild if there's any change in the Editor directory - the cache key should change if there's a been an update to any file in the `/editor.planx` directory.


Change pushed up!

gunar:
Nice catch! Sorry! Thanks!

PR #1180, opened 2022-10-02
fix: prevent duplicate submissions per sessionId and destination

jessicamcinchak:
Since fixing the "don't show first node in the graph" issue when resuming between Pay & Send last week, it seems that the Send component has become more prone to automatically rerendering/reloading/etc. We hit this issue twice with manaul submissions and yesterday with a live submission. There have also been successful single-submisssions in between though.

While we're fully investigating and diagnosing that, this should help add a layer of protection to avoid duplicate successful submissions (and by extension notifications in the Slack channel) per sessionId & destination. Because the Send component creates Hasura Events, and then those events run with some level of parallelization, I'm not certain this will catch every case - but it seems to work as intended in testing.

This PR won't make any difference to the duplicate Pay notifications we're also occassionally seeing, but I'd argue that those are actually perhaps useful right now as we continue to debug that redirect as long as they continue to reflect a **single** payment.

### DIFF
--- a/api.planx.uk/send/bops.js
+++ b/api.planx.uk/send/bops.js
@@ -121,6 +121,11 @@ const sendToBOPS = async (req, res, next) => {
   }
 };
 
+/**
+ * Query the BOPS audit table to see if we already have an application for this session
+ * @param {string} sessionId 
+ * @returns {object|undefined} bops_applications.response
+ */
 async function checkBOPSAuditTable(sessionId) {
   const application = await client.request(
     `
--- a/api.planx.uk/send/bops.js
+++ b/api.planx.uk/send/bops.js
@@ -13,20 +13,20 @@ const sendToBOPS = async (req, res, next) => {
     });
   }
 
+  // confirm that this session has not already been successfully submitted before proceeding
+  const submittedApp = await checkBOPSAuditTable(payload?.planx_debug_data?.session_id);
+  if (submittedApp?.message === "Application created") {
+    res.status(200).send({
+      sessionId: payload?.planx_debug_data?.session_id,
+      bopsId: submittedApp?.id,
+      message: `Skipping send, already successfully submitted`,
+    });
+  }
+
   // confirm this local authority (aka team) is supported by BOPS before creating the proxy
   //   XXX: we check this outside of the proxy because domain-specific errors (eg 404 "No Local Authority Found") won't bubble up, rather the proxy will throw its' own "Network Error"
   const isSupported = ["BUCKINGHAMSHIRE", "LAMBETH", "SOUTHWARK"].includes(req.params.localAuthority.toUpperCase());
   if (isSupported) {
-    // confirm that this session has not already been successfully submitted
-    const submittedApp = await checkBOPSAuditTable(payload?.planx_debug_data?.session_id);
-    if (submittedApp?.message === "Application created") {
-      res.status(200).send({
-        sessionId: payload?.planx_debug_data?.session_id,
-        bopsId: submittedApp?.id,
-        message: `Skipping send, already successfully submitted`,
-      });
-    }
-
     // a local or staging API instance should send to the BOPS staging endpoint
     // production should send to the BOPS production endpoint
     const domain = `https://${req.params.localAuthority}.${process.env.BOPS_API_ROOT_DOMAIN}`;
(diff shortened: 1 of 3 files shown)


api.planx.uk/send/bops.js
```diff
@@ -1,27 +1,37 @@
...
+  if (!payload) {
+    next({
+      status: 400,
+      message: `Missing application payload data to send to BOPS`,
+    });
+  }
+
+  // confirm that this session has not already been successfully submitted before proceeding
+  const submittedApp = await checkBOPSAuditTable(payload?.planx_debug_data?.session_id);
+  if (submittedApp?.message === "Application created") {
```
jessicamcinchak, line 18: this check ensures that we will still proceed & "retry" for failed submissions. 

as an example, see BOPS prod case from last week that we have multiple audit entries for, but all prior ones have response `{ "message": "Validation failed: File The selected file must be a PDF, JPG or PNG" }`

api.planx.uk/send/bops.js
```diff
@@ -1,27 +1,37 @@
...
+    res.status(200).send({
+      sessionId: payload?.planx_debug_data?.session_id,
+      bopsId: submittedApp?.id,
+      message: `Skipping send, already successfully submitted`,
+    });
```
jessicamcinchak, line 23: we still return success 200 if we skip so the Hasura event log doesn't show an error or failed event.

api.planx.uk/send/uniform.js
```diff
@@ -35,6 +35,16 @@ const sendToUniform = async (req, res, next) => {
     });
   }
 
+  // confirm that this session has not already been successfully submitted before proceeding
+  const submittedApp = await checkUniformAuditTable(payload?.sessionId);
+  if (submittedApp?.submissionStatus === "PENDING" && submittedApp?.canDownload) {
```
jessicamcinchak, line 40: it's trickier to check the uniform response for "success", but I think this combination is sufficient? open to other ideas here!

jessicamcinchak:
[Improve pay and make sure every paid app is sent](https://trello.com/c/nTxKcBLL/2092-improve-pay-and-make-sure-every-paid-app-is-sent)

api.planx.uk/send/bops.js
```diff
@@ -1,27 +1,37 @@
...
+  const { payload } = req.body;
+  if (!payload) {
+    next({
+      status: 400,
+      message: `Missing application payload data to send to BOPS`,
+    });
+  }
+
+  // confirm that this session has not already been successfully submitted before proceeding
+  const submittedApp = await checkBOPSAuditTable(payload?.planx_debug_data?.session_id);
```
DafyddLlyr, line 17: nit: For a future PR! I'm also wrote a function which relied on `payload?.planx_debug_data?.session_id` for BOPS - I think these two functions basically confirm that it's a core part of the process and shouldn't be buried in the "temporary" (?) debug data.

api.planx.uk/send/bops.js
DafyddLlyr: This will certainly work, but does leave us vulnerable to a BOPS change here. This seems unlikely, but if there's a way we could also check a HTTP response code or similar we could have a bit more confidence here.

DafyddLlyr: I think we could capture this from BOPS and keep it in the table, as these changes are really aimed at future duplicate applications which happen in close succession as opposed to checking an historic entry vs a recent one.

api.planx.uk/send/uniform.js
DafyddLlyr: Again, it could be worth improving what we capture back from Uniform to have a bit more certainty here - but I think there's zero problems with this and it does describe the check correctly 👍

DafyddLlyr (approved):
Great stuff - a few very minor suggestions/comments but no required changes for sure 👍 

Appreciate the quick turnaround here 😄

api.planx.uk/send/bops.js
jessicamcinchak: Right, we don't store a status code currently, just this message response and some other not-so-useful response headers.

In the future, I was thinking the interesting improvement here could be moving away from querying the audit tables altogether and actually just checking if Hasura already has an event queued up for this session id & destination. To do that, I think we'd want to create a tracked view that pulls event info from the pg metadata tables that we could then query via GraphQL. I think it'd have the extra benefit of possibly catching more cases, as the event is created far before the audit entry is inserted.

PR #1181, opened 2022-10-03
feat: Copy over applicant address if same as property address

DafyddLlyr:
## Problem
- Applicant address not pulling through on Uniform if the same as the property address

## Solution
 - Check `passport.data.resident`
 - If true, use applicant address. Otherwise, still populate fields with "_address" values.
 - **Update** - It's not quite this simple...
   - The `_address` and `applicant.address` values in the passport are not of the same type 
   - One is constructed from OS Data (e.g. uprn, lat long, etc), and the other from the AddressInput component
   - I've updated type names to make this clearer
   - Luckily the XML schema allows an applicant's address to be either a bs7666:BS7666Address or a common:InternationalAddress node. See snippet of the XSD schema below -
 
```xml
<xsd:element name="ExternalAddress">
  <xsd:complexType>
    <xsd:choice>
      <xsd:element ref="bs7666:BS7666Address"/>
      <xsd:element name="InternationalAddress" type="apd:InternationalAddressStructure"/>
    </xsd:choice>
  </xsd:complexType>
</xsd:element>
```
https://ecab.planningportal.co.uk/uploads/schema/OneAppCommon/OneAppCommon-2-0-1.xsd

## Testing
 - I've sent a few test applications to Kev, waiting on feedback currently


DafyddLlyr:
[Copy over applicant address if same as property address](https://trello.com/c/YPV33h3f/2121-copy-over-applicant-address-if-same-as-property-address)

jessicamcinchak:
[Improve test coverage for Uniform integration](https://trello.com/c/QwUs9nXs/2008-improve-test-coverage-for-uniform-integration)

DafyddLlyr:
Testing has indicated that this is actually a little bit more complex - working on a solution now.

DafyddLlyr:
Re-tested - now getting the expected results 👌 

image

DafyddLlyr:
### Update

Uniform does allow a BS7666Address node to populate the applicant address field, but does not display all values, even when the exact same node is used. Possibly Uniform is taking the UPRN and generating its own address? (Pure speculation...!)

image

I've gone for an alternative approach here of just mapping a `SiteAddress` -> `PersonAddress`

Test applications sent, awaiting feedback.

editor.planx.uk/src/@planx/components/AddressInput/model.ts
```diff
@@ -3,7 +3,7 @@ import { object, string } from "yup";
 
 import { MoreInformation, parseMoreInformation } from "../shared";
 
-export type UserData = {
+export type PersonAddress = {
```
jessicamcinchak, line 6: nit: this component is also used to collect organisation addresses - maybe `ApplicantAddress` or `UserAddress` is a bit more generalized and fits better with existing language?

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -28,7 +28,7 @@ import Input from "ui/Input";
 import InputLabel from "ui/InputLabel";
 import { fetchCurrentTeam } from "utils";
 
-import type { Address, FindProperty } from "../model";
+import type { FindProperty, SiteAddress } from "../model";
```
jessicamcinchak, line 31: :heavy_plus_sign: thanks for this naming distinction!

editor.planx.uk/src/@planx/components/Send/uniform/xml.ts
```diff
@@ -145,8 +152,45 @@ export function makeXmlString(
...
+          <apd:IntAddressLine>${escape(address?.town)}</apd:IntAddressLine>
+          <apd:IntAddressLine>${escape(address?.county)}</apd:IntAddressLine>
+          <apd:Country>${escape(address?.country)}</apd:Country>
+          <apd:InternationalPostCode>${escape(
+            address?.postcode
+          )}</apd:InternationalPostCode>
+        </common:InternationalAddress>
+      </common:ExternalAddress>
+    `;
+  };
```
jessicamcinchak, line 193: :+1:

jessicamcinchak (approved):
Thanks for continued adjustments here, hope it's confirmed on Uniform's side soon!

editor.planx.uk/src/@planx/components/AddressInput/model.ts
DafyddLlyr: Yeah this is a good point 👍 On the XML side of things I'm trying to make it distinct from a site address and generating an address for either the agent or applicant so `ApplicantAddress` or `UserAddress` doesn't quite work either.

Could it be as simple as just `Address` and `SiteAddress`?

jessicamcinchak: Yep happy with `Address` & `SiteAddress` !

DafyddLlyr:
Finally working...! 😅 

image

Address is correct, but in slightly different format. I'm happy to go ahead with this and wait for further feedback.

PR #1184, opened 2022-10-04
fix: Remove duplicate "Proposal" value in Uniform

DafyddLlyr:
After testing with Kev, I've identified `common:OperationsDescription` as the problem node which was duplicating descriptions for Proposed applications. 

The descriptions where never duplicated for Existing LDC apps (see second screenshot).

image (5)
image (4)


DafyddLlyr:
[Investigate duplicate descriptions on Uniform](https://trello.com/c/0zWHP7qi/2120-investigate-duplicate-descriptions-on-uniform)

gunar (approved):

PR #1185, opened 2022-10-04
build: upgrade passport >= 0.6.0

jessicamcinchak:
https://github.com/theopensystemslab/planx-new/security/dependabot/1

Passport is used by Google Auth in API, to test: confirm you can login & logout of this pizza

### DIFF
--- a/api.planx.uk/package.json
+++ b/api.planx.uk/package.json
@@ -28,7 +28,7 @@
     "mime": "^3.0.0",
     "nanoid": "^3.3.4",
     "notifications-node-client": "^5.1.1",
-    "passport": "^0.5.3",
+    "passport": "^0.6.0",
     "passport-google-oauth20": "^2.0.0",
     "pino-noir": "^2.2.1",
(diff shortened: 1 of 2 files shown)


DafyddLlyr:
We previously had issues upgrading this dependency. Once it hit staging and we hit `/logout` the following error was logged - https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1658848785967489

I then downgraded here - https://github.com/theopensystemslab/planx-new/pull/1037

There is an override in `package.json` which should keep up clear of the affected sub-dependency, but upgrading passport would be good to do. Looking at the issues there and at session-cookie I'm not sure the issue is currently fixed.

I can logout of the pizza just fine, but a console error is thrown. Happy to try again for sure on this one 👍

DafyddLlyr:
Thanks @builditben for the thorough testing here - if you could please merge this just now it will trigger a deploy to staging where we can test.

If all's good there let's push try and push this single change onwards to production also today and draw a line under it 💪

PR #1186, opened 2022-10-04
ci(pizza): teardown pizzas manually

gunar:
- Allows us to teardown pizzas by manually triggering the GitHub Actions teardown workflow
- I can only test it by merging it (and seeing if the button shows up on the Actions tab of this repo) so I might need a couple of PRs to get this right

Here's a picture of a workflow_dispatch action and its input fields:

image

(of course, our fields will be different. the image above is for illustration purposes only)


DafyddLlyr (approved):
Great idea! 👌

PR #1187, opened 2022-10-05
docs: update Feedback Fish project ID

jessicamcinchak:
we have two project ids floating around, but there should only be one. This one is already correctly set in the Staging & Production deploy Github workflows, it was only wrong for local & pizza Editor envs which was causing some confusing emails.

i've additonally overwritten the remote editor.env file in S3. 

https://opensystemslab.slack.com/archives/C5Q59R3HB/p1664960742399609


DafyddLlyr (approved):
👍

PR #1188, opened 2022-10-05
fix: Only update has_user_saved column once upon initial Save

DafyddLlyr:
**Problem**

**Solution**

### DIFF
--- a/api.planx.uk/saveAndReturn/utils.js
+++ b/api.planx.uk/saveAndReturn/utils.js
@@ -108,6 +108,7 @@ const sendSingleApplicationEmail = async (template, email, sessionId) => {
       reference: null,
       emailReplyToId: team.notifyPersonalisation.emailReplyToId,
     };
+    if (!session.has_user_saved) await setupEmailEventTriggers(sessionId);
     return await sendEmail(template, email, config);
   } catch (error) {
     throw Error(error.message)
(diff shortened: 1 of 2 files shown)


DafyddLlyr:
PR workflow current stuck in "queue", due to issues with GitHub actions - will take a look in the morning to try and move this forward.

image

api.planx.uk/saveAndReturn/utils.js
```diff
@@ -108,6 +108,7 @@ const sendSingleApplicationEmail = async (template, email, sessionId) => {
       reference: null,
       emailReplyToId: team.notifyPersonalisation.emailReplyToId,
     };
+    if (!session.has_user_saved) await setupEmailEventTriggers(sessionId);
```
jessicamcinchak, line 111: a little confused reading this line - should the logic here be to only setup email triggers if `session.has_user_saved` is TRUE ?

api.planx.uk/saveAndReturn/utils.js
```diff
@@ -309,6 +310,7 @@ const stringifyWithRootKeysSortedAlphabetically = (ob = {}) =>
 
 // Update lowcal_sessions.has_user_saved column to kick-off the setup_lowcal_expiry_events & 
 // setup_lowcal_reminder_events event triggers in Hasura
+// Should only run once on initial save of a session
```
jessicamcinchak, line 313: :+1:

jessicamcinchak (approved):
thanks for getting to the bottom of this one! 

one logic question, but trust you'll straighten it out before merge :slightly_smiling_face:

api.planx.uk/saveAndReturn/utils.js
DafyddLlyr: Yeah I'll add a comment here - it is correct but reads weirdly. Might be better to rename the function.

If the user has not yet saved, setup the emailEventTriggers. The `has_user_saved` value gets toggled within this function, which is what Hasura is listening to to set them up.

PR #1190, opened 2022-10-06
refact(api): add types

gunar:
Adds types to the API service.


DafyddLlyr (approved):
Thanks for picking this up! ✅

PR #1191, opened 2022-10-07
bug: Send payload should build despite additional breadcrumbs

DafyddLlyr:
Addresses the following error - https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1665145845295839

The affected user here had an additional breadcrumb which was subsequently removed from the flow, leading to the following error.

We could / should catch this at the reconciliation stage but a frontend guard will always be required as sometimes reconciliation is skipped (e.g. returning from pay).


jessicamcinchak (approved):
good safeguarding, thank you!!

PR #1192, opened 2022-10-08
New rich text editor

peterszerzo:
New rich text editor with the following changes:
- placeholder support with interpolation.
- better UX for links.
- link text validation for accessibility.

https://user-images.githubusercontent.com/6738398/196024579-378919fe-cec1-497f-95b6-202a409b7d1f.mp4


peterszerzo:
@builditben I found accurate alternatives to most of the `any`'s in the PR and documented where I wasn't able to find / added todo comments documenting the next steps in clearing them out.

As far as `() => void` is concerned, I think it's as accurate as can be (takes no arguments, returns nothing meaningful, so essentially a very basic side effect), maybe you can elaborate on what specifically can go wrong with this typing in place?

peterszerzo:
@builditben I found solutions for the 'type text after link' and disabled the open on click behavior using these boolean flags: https://github.com/theopensystemslab/planx-new/blob/new-rte/editor.planx.uk/src/ui/RichTextInput.tsx#L72-L73.

jessicamcinchak (approved):

jessicamcinchak (approved):
thanks for these upgrades, rich text editing feels much smoother and more predictable! 

only noticed one visual regression in the editor that i think is introduced here but not positive? doesn't seem to effect any public-facing buttons, so I don't think we need to care about it, but flagging just in case it's obvious to you how/where it popped up (removing @draft-js-plugins/buttons i'd guess? had previously assumed these buttons were also material ui) 

when editing a graph node, the bottom buttons are a bit squished now: 
Screenshot from 2022-11-10 11-58-17

before/production:
Screenshot from 2022-11-10 11-58-00

happy for this to merge to staging whenever you feel ready!

peterszerzo:
@jessicamcinchak that's a good catch, I looked around a bit for possible causes and unfortunately the only thing I can think of is to remove the [old rich text editor-specific theming](https://github.com/theopensystemslab/planx-new/blob/80a425fefbcc729a6879006137a2eb74d79e7e44/editor.planx.uk/src/theme.ts#L211-L241). I'll have a look at how the .pizza link behaves after that.

PR #1198, opened 2022-10-10
feat(ci): allow running staging db sync manually

gunar:
Allow running staging db sync manually, so we can run tests and eventually actually run in manually in case some content is changed and we need it to be on staging immediately.


jessicamcinchak (approved):
this is a smart feature, thanks!

PR #1204, opened 2022-10-14
user-research: switch cursor when drawing on map

jessicamcinchak:
request for a pizza that can be used in user-research testing, no need to code review yet!

https://1204.planx.pizza/testing/maps-test/preview


jessicamcinchak:
[Turn red dot into a crosshair – for testing on a pizza](https://trello.com/c/UqC0QJAx/2135-turn-red-dot-into-a-crosshair-for-testing-on-a-pizza)

PR #1207, opened 2022-10-14
feat: create a ContactInput component type

jessicamcinchak:
**Changes:**
Introduces a `ContactInput` component type which will replace individual text inputs used to collect appplicant names, phone numbers, emails, etc.

The passport representation for applicants (or agents, proxies, etc) changes as follows when switching from many `TextInput` components to a single `ContactInput` component and has been updated in the BOPS & Uniform schema mappings (the schemas will support _both_ so there's no strict adoption rule/timeline of the new component):
Before:
```json
"passport": {
  "data": {
    "applicant.agent.name.first",
    "applicant.agent.name.last",
    "..."
   }
 }
 ```
 
 After: 
 ```json
"passport": {
   "data": {
     "applicant.agent": {
       "name.first",
       "name.last",
       "..."
    }
   }
 }
 ```
**How to test:**
- I've updated this pizza's `about-the-applicant` flow to use the new component: https://1207.planx.pizza/opensystemslab/about-the-applicant
Screenshot from 2022-11-04 12-28-32
- I've published a parent LDC flow to capture that change: https://1207.planx.pizza/buckinghamshire/apply-for-a-lawful-development-certificate/preview?analytics=false
- You should go through the LDC flow, confirm that you see the new ContactInput components and that your passport is consistent with how it used to be, and that all necessary applicant fields are still sent to BOPS & Uniform as expected

---
**Prior context:**
One of a couple experiments this week related to navigation (see also #1209). This one aims to address a very common use case which is entering your contact details, currently spread out over ~6 individual text inputs: https://editor.planx.uk/opensystemslab/about-the-applicant. Largely follows the same structure as our existing `AddressInput` component.

Pros:
- Maintains accessibilty
- Doesn't touch graph logic
- Quick to rollout

Cons: 
- Lack of flexibility / editor control
- Loss of individual "description" to supplement input label
- Is a bandaid to the dream feature

Todos and cleanup if moving forward (estimate 1):
- [x] Add valid email regex (see `TextInput` model)
- [x] Clean up `SummaryList`/`Review` page presentation
- [x] Ensure passport variables will match existing flow, check if any data models need to be adjusted for CSV/BOPS/Uniform
- [x] Add tests

### CODE REVIEW

**Comment by jessicamcinchak** [comment]
[[Explore] Form component](https://trello.com/c/33XA0rdR/2146-explore-form-component)

File: editor.planx.uk/src/@planx/components/ContactInput/Public.tsx [hunk 1]
```diff
@@ -0,0 +1,156 @@
...
+      organisation: "",
+      phone: "",
+      email: "",
+    },
+    onSubmit: (values) => {
+      // map internal props to expected granular passport keys before submitting
+      values["name.first"] = values.firstName;
+      values["name.last"] = values.lastName;
+      values["company.name"] = values.organisation;
+      values["phone.primary"] = values.phone;
```
jessicamcinchak, line 45: this mapping feels particularly clumsy, but `formik.handleChange` breaks on an input if its' direct name includes a `.`

see this attempt: 
Screenshot from 2022-11-04 08-55-42

then formik failed to reconcile the passport and we had duplicate: `name.first: <>` & `name: { first: <> }` like this which prevented being able to chagne the first name input:
Screenshot from 2022-11-04 08-54-10

i fiddled with this for awhile and eventually just settled on this mapping, but maybe i'm forgetting some syntax to accomplish this in a more eloquent way? any better ideas?

editor.planx.uk/src/@planx/components/ContactInput/Public.tsx
```diff
@@ -0,0 +1,156 @@
...
+            onChange={formik.handleChange}
+            id={`${props.id}-title`}
+            inputProps={{
+              "aria-describedby": formik.errors.title
+                ? `${ERROR_MESSAGE}-${props.id}-title`
+                : "",
+            }}
+          />
+        </InputRowItem>
+      </InputLabel>
```
jessicamcinchak, line 80: perhaps interesting to note that GovUK suggests avoiding asking for a person's title, or at minimum making it optional: https://design-system.service.gov.uk/patterns/names/

i remember previous conversations in planning about this where partners were very resistant to the idea of not collecting this field because officers are used to referencing his information when making followup phone calls, but might be worth raising again when this goes to UAT

jessicamcinchak:
code reviews very welcome here! but please note that testing/publishing content on the pizza is unreliable until [this data sync bug](https://trello.com/c/fMnVFVy5/2187-bug-publishing-flows-on-staging) is sorted!

editor.planx.uk/src/@planx/components/ContactInput/Public.tsx
DafyddLlyr: So it certainly looks like this should be do-able - there's an option in the docs to avoid nesting https://formik.org/docs/guides/arrays#avoid-nesting and a `getIn()` method which allows us to access values/errors with the following syntax - `value={getIn(formik.values, "first.name")}`

Spent a bit of time on it and I can get it all working apart from the Yup Schema which doesn't seem to handle this. I think mapping is the sensible way forward here!

Question - if we're mapping values anyway, is there a way we could retain the original mappings with something like this and remove both complexity elsewhere and (small) issues for reconciling users?
```ts
values[`${parentPassportValue}.name.first`] = values.firstName;
```

editor.planx.uk/src/@planx/components/ContactInput/Public.tsx
```diff
@@ -0,0 +1,156 @@
...
+  title: string;
+  firstName: string;
+  lastName: string;
+  organisation: string;
+  phone: string;
+  email: string;
+  "name.first"?: string;
+  "name.last"?: string;
+  "company.name"?: string;
+  "phone.primary"?: string;
```
DafyddLlyr, line 27: Are these required here or leftover from trying to get them to work?

DafyddLlyr (commented):
Great stuff! 👍 Really appreciate the level of testing here

One thing not relating to a specific line of code - I expect that this could break integrations (e.g. Kev's PDF maker). I don't think that's necessarily our responsibility but we should pre-warn Southwark and Bucks that once they update their components the passport will have an alternate structure.

editor.planx.uk/src/@planx/components/ContactInput/model.ts
```diff
@@ -0,0 +1,48 @@
...
+  phone: string().required("Phone number is required"),
+  email: string()
+    .required("Email is required")
+    .test({
+      name: "valid",
+      message: (() => {
+        return "Enter a valid email";
+      })(),
+      test: (value: string | undefined) => {
+        return Boolean(value && emailRegex.test(value));
```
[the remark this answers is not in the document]
jessicamcinchak: Good call! Now updated to be a nice clean one-liner: ` email: string().email("Enter a valid email").required("Email is required")`

editor.planx.uk/src/@planx/components/ContactInput/Public.tsx
jessicamcinchak: These are indeed actually required, now fixed, thanks for catching :+1:

editor.planx.uk/src/@planx/components/ContactInput/Public.tsx
jessicamcinchak: This is such a finicky topic that I've been feeling stuck & unsatisfied around too - I've hit the same issues with formik & Yup. 

BUT, I think I landed on a better mapping that fits your second question well:

I'm using this representation for the passport - which keeps the existing/expected passport keys in the top level, and adds a `_contact.[props.fn]` nested object to the passport that we can reference internally for back button/change behavior. This feels like a similar patern to FindProperty's `_address` passport field and accounts for the likelihood that we will have multiple ContactInputs per flow.

The tricky bit was writing custom `makeData` & `getPreviouslySubmittedData` methods rather than referencing the shared utils, but I think it's worth it for this passport representation which then negates all those silly `||` conditionals I originally introduced in the Send schemas. Thoughts?

Screenshot from 2022-11-15 15-18-51

editor.planx.uk/src/@planx/components/shared/utils.ts
```diff
@@ -60,24 +59,31 @@ export const makeData = <T>(
...
-      passport.data?.[bopsDictionary.applicant_last_name],
-    ]
-      .map((x) => String(x).toLowerCase().trim())
-      .join("|") === "test|test"
-  ) {
-    const url = new URL(urlThatMightBeReplaced);
-    url.hostname = url.hostname.replace("planx.uk", "planx.dev");
-    return url.href;
-  }
+export const useStagingUrlIfTestApplication =
```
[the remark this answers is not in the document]
jessicamcinchak: Yep good thinking here - but was able to remove most all of these `||` cases with this update: https://github.com/theopensystemslab/planx-new/pull/1207#discussion_r1023151808

Definitely good thinking for future though, I don't expect this will be the last time we think about this issue!

editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
```diff
@@ -337,7 +337,11 @@ function ContactInput(props: ComponentProps) {
     <>
       <dt>{props.node.data.title ?? "Contact"}</dt>
       <dd>
-        {title + " " + firstName + " " + lastName}
+        {title ? (
+          <>{title + " " + firstName + " " + lastName}</>
+        ) : (
+          <>{firstName + " " + lastName}</>
+        )}
```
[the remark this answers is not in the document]
jessicamcinchak: yes updated, thanks!

editor.planx.uk/src/@planx/components/ContactInput/Public.tsx
```diff
@@ -38,14 +29,26 @@ export default function ContactInputComponent(props: Props): FCReturn {
...
-      values["company.name"] = values.organisation;
-      values["phone.primary"] = values.phone;
+      // map values to the existing/expected passport structure to minimize conditional handling later in Send schemas, etc
+      const newPassportData: any = {};
+      newPassportData[`${props.fn}.title`] = values.title;
+      newPassportData[`${props.fn}.name.first`] = values.firstName;
+      newPassportData[`${props.fn}.name.last`] = values.lastName;
+      newPassportData[`${props.fn}.company.name`] = values.organisation;
+      newPassportData[`${props.fn}.phone.primary`] = values.phone;
+      newPassportData[`${props.fn}.email`] = values.email;
```
DafyddLlyr, line 39: This is a solid solution I think and you can see from this commit and it's doing a great job of stripping complexity out elsewhere 🎉

editor.planx.uk/src/@planx/components/ContactInput/Public.tsx
```diff
@@ -38,14 +29,26 @@ export default function ContactInputComponent(props: Props): FCReturn {
...
+
+      const passportData = {
+        [`_contact.${props.fn}`]: { [`${props.fn}`]: values },
```
DafyddLlyr, line 42: I think I'm missing something here sorry - what's the value / purpose of retaining both structures of contact info?

DafyddLlyr (commented):
Really solid improvement - thanks for keeping at it.

One question left on the code, think I'm maybe missing something sorry. Happy to chat through if simpler.

editor.planx.uk/src/@planx/components/ContactInput/Public.tsx
jessicamcinchak: retaining the non-`.` passport variables (eg `firstName`) allows us to pre-fill the formik input fields when going "back" & "changing" !

jessicamcinchak: this is part of the stickyness of the original goal to be able to use a name with a `.` in it for the name/value of the input itself - but in lieu of Formik/Yup not particularly liking that - we'll maintain `_contact...` as an internal lookup that is read by `initialValues`, `getPreviouslySubmittedData` etc

DafyddLlyr: Perfect - knew there was a reason and there was something I wasn't seeing. Thanks for the explanation ✅

DafyddLlyr (approved):
✅

PR #1208, opened 2022-10-14
feat: Capture payload on Uniform application audit table

DafyddLlyr:
## What does this PR do?
- Adds a `payload` column on the `uniform_application` audit table
- This will be helpful for debugging as we're now doing a better job of cleaning up the scheduled event tables (#1206)

### To test
 - An application submitted to Uniform will have the `payload` column populated


jessicamcinchak (approved):
Thanks for picking this up together with the event cleanup! Will be appreciated during future investigations I'm sure.

PR #1220, opened 2022-10-18
feat: Add UniformPayload types, remove XML string concatenation

DafyddLlyr:
## What does this PR do?
 - Add types for `UniformPayload` generated from the current XML payload
 - Use these types to instantiate a `UniformPayload`, and then generate and XML document which gets sent to Uniform
 - Replaces the current `makeXMLString()` functionality with the above methods

### Comments
We've discussed moving the entire "Send" functionality to the the server side a few times. I've kept this in mind and hopefully the code here is self contained enough that it could in future be moved server side without much effort. The `UniformPayload` class could move now and be called as part of the send functionality server side - however it feels more consistent to have this co-located currently, leaving open the door to move it later. Happy for thoughts on this!

**Namespaces**
Ideally, the XML namespaces would have been captured as part of the type definition. Something like this - 

```ts
interface XmlNode {
  $text?: string;
  $prefix: **Prefix Enum**;
}

interface Proposal extends XmlNode { 
  _someAttribute: string;
  $text: "Text content of the node"
  $prefix: Prefix.Portaloneapp
}

...
```

There's nothing in the docs of fast-xml-parser that natively allows us to define a namespace/prefix as part of the object. I also had a look at a few other libraries which didn't seem to support this. I could iterate deeply over the object whilst modifying keys but in order to move forward and timebox this I've parked this for now.

### Next steps...
I'll mark this as open for review once I get feedback from Kev that the test submissions I've sent have worked as expected.


DafyddLlyr:
[Implement UniformPayload interface to improve Uniform XML construction](https://trello.com/c/6PD2SuLw/2144-implement-uniformpayload-interface-to-improve-uniform-xml-construction)

DafyddLlyr:
Test applications going through to Wycombe staging without issue now...! 🎉 

I suspect the issues we've been having testing this were down the previous issues with the staging environments and not related to this code change which is great.

The following two submissions can be found on the Pizza DB / Hasura - 

image

image

editor.planx.uk/src/@planx/components/Send/uniform/UniformPayload/model.ts
```diff
@@ -0,0 +1,302 @@
...
+
+export class UniformPayload implements IUniformPayload {
+  sessionId: string;
+  passport: Store.passport;
+  files: string[];
+
+  proposalCompletionDate: string;
+  siteAddress: SiteAddress;
+
+  "portaloneapp:Proposal": Proposal;
```
jessicamcinchak, line 32: this is a smart simple way to cheat namespaces! i had definitely gotten very hung up trying to get them typed properly without any success :rabbit: :hole:

editor.planx.uk/src/@planx/components/Send/uniform/UniformPayload/model.ts
```diff
@@ -0,0 +1,302 @@
...
+    };
+    const payload = {
+      "?xml": xmlDeclaration,
+      "portaloneapp:Proposal": this["portaloneapp:Proposal"],
+    };
+    const buildOptions: Partial<XmlBuilderOptions> = {
+      ignoreAttributes: false,
+      attributeNamePrefix: "_",
+      format: true,
+      suppressEmptyNode: true,
```
jessicamcinchak, line 296: is it right that `suppressEmptyNode: true` is what now handles still generating empty field placeholders for any "undefined" passport variables?

editor.planx.uk/src/@planx/components/Send/uniform/UniformPayload/model.ts
```diff
@@ -0,0 +1,302 @@
...
+    "_xmlns:org": "http://www.govtalk.gov.uk/financial/OrganisationIdentifiers",
+    "_xmlns:pdt": "http://www.govtalk.gov.uk/people/PersonDescriptives",
+    "_xmlns:apd": "http://www.govtalk.gov.uk/people/AddressAndPersonalDetails",
+    "_xmlns:core": "http://www.govtalk.gov.uk/core",
+    "_xmlns:common": "http://www.govtalk.gov.uk/planning/OneAppCommon-2006",
+  });
+
+  private getRequiredFiles = (): FileAttachment[] => [
+    {
+      "common:Identifier": "N10049",
```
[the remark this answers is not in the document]
DafyddLlyr: This is a great point and there are other TODOs in the code that I think we should test once we've made this initial change.

Looking at the schema, this looks like it should be optional - 

```xml
<xsd:element name="Identifier" type="core:PopulatedStringType" minOccurs="0"/>
```

https://ecab.planningportal.co.uk/uploads/schema/OneAppCommon/OneAppCommon-2-0-1.xsd

I've added an explicit TODO here to better test or document this.

editor.planx.uk/src/@planx/components/Send/uniform/UniformPayload/model.ts
DafyddLlyr: Yeah I spent a lot of time trying to get to the bottom of this without success and settled on this as a pragmatic step in the right direction. Not one I fancy revisiting in a hurry tbh..!

editor.planx.uk/src/@planx/components/Send/uniform/UniformPayload/model.ts
DafyddLlyr: Good question!

`suppressEmptyNode` renders nodes as `<something />` instead of `<something></something>` in the case of missing values.

This seemed like a sensible approach to help debugging and to more clearly indicate we don't have a value for something.

DafyddLlyr:
Changes since last review - 
 - Rebase to `main` and pulled in tests for `ContactInput`, `boundary.geojson` and `review.html`
   - All tests passed once function names were updated 🎉 
 -  Update `UniformPayload` constructor to take in an options object, instead of a list of arguments
 - Add types to test file

DafyddLlyr:
This may not have been tested as thoroughly as I would ideally like, I do have a good degree of confidence that's we're good to merge this. The reasons being - 

 - Our existing tests all pass on this branch
 - Kev & Maj have confirmed that they can receive applications and all documents to an Idox staging environment (including `review.html` and `boundary.geojson`)
 - We've had screenshots of LDCP and LDCE applications where all properties appear to be correctly mapped

If there are any issues here, I'd suspect they would be in the mapping of values which should be easily fixed and identified.

PR #1221, opened 2022-10-19
Add an initial set of testing principles

builditben:
After our discussion about testing today, I thought it was worth putting together a little document with a minimal set of principles around out test approach. I've also documented some terminology since testing language can sometimes be a little ambiguous.

I have only included principles that I think are already part of the team's culture but do feel free to comment/question if any of these don't fit or if I've got these wrong. Also, feel free to add to or amend these - this is really just a starter document to have a place to capture any principles that we land on now or later on.

The key reason I think this is important is to establish an explicit standard for "code complete" - a job well done (at least from an engineering perspective) is a feature that is functional and has the appropriate level of testing. This is my attempt to establish exactly what is appropriate for our context.


jessicamcinchak (commented):
Really succint and helpful in my opinion! 

I could imagine one additional section in this doc that is something like "Examples" or "Where we think we're doing this well currently" for each category/principal if you think that could fit? As a dev starting new work in one of these categories, I always appreciate being able to read over a similar example first. Also probably a helpful inventory exercise of our current tests/coverage (no pressure to do that exercise solo, some good examples are probably still TODO :sweat_smile:)

doc/principles/0001-testing-principles.md
```diff
@@ -0,0 +1,34 @@
...
+  * **UI Component tests** are unit tests for UI components. These largely follow the same model as unit tests but, rather than testing purely functional interfaces, they attempt to simulate user interactions. Like unit tests, any state provided to these tests is always statically mocked.
+
+  * **Service tests** simulate the use of a service with mocked dependencies. They interact with a service (usually an API) as an an external application/user would but use statically mocked data to ensure tests are deterministic and side-effect free.
+
+  * **End-to-end tests** cover a user journey through a system. These tests interact with the application from the outside, with dummy data but a full running test stack (Web UI, APIs, Database, etc). In this form of testing, only external services are mocked - the full application is under test.
+
+
+## Principles
+
+1. Linting should run in the pipeline on every commit.
```
gunar, line 26:
```diff
- pipeline
+ CI pipeline
```

doc/principles/0001-testing-principles.md
```diff
@@ -0,0 +1,34 @@
...
+1. Unit tests should cover all logically complex code.
```
gunar, line 27: This seems vague (and hence unnecessary)? The terminology itself might be enough.

doc/principles/0001-testing-principles.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Context
+
+The intent of this document is to provide an overview of our testing terminology and approach to ensure we share the same understanding and are working from the same set of assumptions and principles about how we test as a team.
+
+## Terminology
+
+  * **Linting** is a type of static analysis for simple code patterns. These tests are entirely automated and are used to catch classes of errors that can be detected in how code has been written (i.e. missing HTML attributes for accessibility).
+
+  * **Unit tests** directly interact with individual units of code. They are small and focused on testing a single functional interface at a time. Unit tests always mocks external dependencies and are entirely deterministic.
```
gunar, line 15:
```diff
- Unit tests always mocks external dependencies and are entirely deterministic.
+ Unit tests should be deterministic and generally should not have external dependencies but if required these dependencies should be mocked.
```

doc/principles/0001-testing-principles.md
```diff
@@ -0,0 +1,34 @@
+# 1. Testing Principles
```
gunar, line 1: Even though these are principles, they still do define the architecture of our code base. In that sense, maybe they should live alongside ADRs? (`doc/architecture/decisions`).

gunar (commented):
Thanks for putting this in writing. I've added some comments :+1:

DafyddLlyr:
This is great thanks 🙌 

I'd certainly second @jessicamcinchak's suggestions of examples - this can serve as an excellent jumping off point. Happy to help point those out.

It may technically fall under the grouping of "Service test" but it might be nice to mention our introspection tests also (see `hasura.planx.uk/tests`). I'd say that we do have a principle of ensuring we cover our permissions model with introspection tests which may complement a more specific service test.

DafyddLlyr (approved):
Looking great 👍

jessicamcinchak (approved):
thanks again for this! very happy for examples to be a separate PR :heavy_plus_sign:

gunar (approved):
:+1: great stuff

PR #1224, opened 2022-10-21
fix(sync): insert all teams before inserting flows

gunar:


jessicamcinchak (approved):
thanks for continuing to troubleshoot this one! :crossed_fingers:

PR #1231, opened 2022-10-24
refact(ci): set-output is being deprecated

gunar:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


jessicamcinchak (approved):
thanks for catching & updating this one!

PR #1232, opened 2022-10-25
WIP: API module splitting

ghost:
I thought it might be helpful to break the API out into slightly more modular files (an easy non-functional change I can get stuck-in with).

What this PR is aiming to do is to split API routes out into self-contained modules which require little context to understand and change.

The first pass at this is splitting up the responsibility of `server.ts` to explicitly set-up routes in context-specific routing files. I'd like to go further and split out the handlers for each of these routes eventually but this is a course-grained pass at doing just that.

Any feedback or input is very welcome (I'm still learning our team's style of communicating and working patterns so I won't be offended if you tell me a change like this is more work to review than it's worth!).


api.planx.uk/src/app/index.ts
```diff
@@ -0,0 +1,13 @@
+// The middleware of an Express app is excecuted in sequence.
+// This means that the order of `app.use(...)` statements is significant.
+//
+// This file exposes the express app after middlewar has been applied in a meaningful order:
+//
+//   1. init       - this sets up app configuration for all routes
+//   2. cookieless - this adds routes that are available without cookies
+//   3. secure     - this adds authenticated routes
+//   4. error      - this adds catchall middleware for error handling
```
DafyddLlyr, line 9: Appreciate the documentation here 👍 

I'm not _exactly_ sure of the implementation, but it would be great to show not tell here if possible - I found the sequential importing and exporting of the `app` a little unclear initially.

Would it be possible for application layers to export an array of routes and/or functions, which can then be used here by `app`? It looks as if Express does support this (docs: https://expressjs.com/en/4x/api.html#app.use).

For example - 
```ts
import express from 'express';
import init from './init';
import cookieless from './cookieless';
import secure from './secure';
import error from './error';

const app = express();
app.use(init)
app.use(cookieless)
app.use(secure)
app.use(error)

export default app;
```

 What do you think?

DafyddLlyr (commented):
Thanks for jumping into this 🙌 

You're correct - generally smaller PRs are better but in some instances like this it's totally unavoidable so no worries on that one.

A few initial thoughts - 
 - Totally agree with the course-grained approach. It's a real bugbear of mine that some routes have complex handlers which could be separated out. I definitely support tackling this as a separate task to a wider re-structuring.
 - `init` and `error` make perfect sense to me as separate and distinct layers of the app
 - The grouping of `cookieless` and `secure` feels slightly less concrete to me and I'd certainly be up for understanding the motivation here a little clearer. Grouping by function feels a more natural choice to be (though this isn't necessarily mutually exclusive). For example, group "Pay" or "GIS" routes together.

api.planx.uk/src/app/secure.ts
```diff
@@ -0,0 +1,265 @@
...
+app.post("/validate-session", validateSession);
+
+assert(process.env.HASURA_PLANX_API_KEY);
+app.use("/webhooks/hasura", useHasuraAuth);
+app.post("/webhooks/hasura/delete-expired-sessions", hardDeleteSessions);
+app.post("/webhooks/hasura/create-reminder-event", createReminderEvent);
+app.post("/webhooks/hasura/create-expiry-event", createExpiryEvent);
+app.post("/webhooks/hasura/send-slack-notification", sendSlackNotification);
+
+export default app;
```
jessicamcinchak, line 265: thanks for this proposal! i think the only sticky point for me on first read through is also this "cookieless" & "secure" split, which doesn't quite feel like the best fitting nomenclature just yet.

in my mind, the language to describe & group these endpoints is perhaps closer to "authenticated" & "public" ? 

where "authenticated" would group endpoints that use our own auth methods (eg `useJWT`, `useHasuraAuth`) and/or a service-specific token/credential in its request (eg Pay, BOPS & Uniform, etc).

and "public" would group our endpoints that are open, no auth or jwt/cookie check necessary (eg `/gis/`, `/analytics/`, etc). 

does that also make sense to others? 

"cookieless" especially feels like it might get a bit conflated with the client-side concept of cookies. and per our planx service agreements, it's emphasized that NO parts of the public-facing planx service use or leave cookies on applicants machines (eg why we are able to not have a cookies banner)!

DafyddLlyr (approved):
Appreciate you've marked it wip and there are a few conflicts to resolve, but I'm happy with what's here and it feels like a solid step forward 👍

PR #1233, opened 2022-10-25
test pizza

gunar:
- prevents user uploads from being publicly accesible
- makes all S3 routes go through the API
- adds a new API Token concept to only allow BoPS to download user files
- side-effect: prevents users from downloading their own files


gunar:
Spawning a new pizza to identify why #1084 isn't building.

PR #1236, opened 2022-10-26
fix: Correctly handle lowcal_session.has_user_saved

DafyddLlyr:
## What does this PR do?
- Please see a proper explanation co-located on the offending code!
  - There was no user impact here - duplicate events were not getting created 👌 
- Adds test cases to capture intended behaviour
- Tidies up old JSDocs

### DIFF
--- a/api.planx.uk/saveAndReturn/sendEmail.test.ts
+++ b/api.planx.uk/saveAndReturn/sendEmail.test.ts
@@ -1,59 +1,29 @@
 import supertest from "supertest";
 import app from "../server";
 import { queryMock } from "../tests/graphqlQueryMock";
-import { mockFlow, mockLowcalSession } from "../tests/mocks/saveAndReturnMocks";
+import { 
+  mockFlow, 
+  mockGetHumanReadableProjectType, 
+  mockLowcalSession, 
+  mockSetupEmailNotifications, 
+  mockSoftDeleteLowcalSession, 
+  mockValidateSingleSessionRequest 
+} from "../tests/mocks/saveAndReturnMocks";
 
 // https://docs.notifications.service.gov.uk/node.html#email-addresses
 const TEST_EMAIL = "simulate-delivered@notifications.service.gov.uk"
+const SAVE_ENDPOINT = "/send-email/save"
 
 describe("Send Email endpoint", () => {
   beforeEach(() => {
     queryMock.reset();
-    queryMock.mockQuery({
-      name: "GetHumanReadableProjectType",
-      data: {
-        project_types: [
-          { description: "New office premises" }
-        ],
-      },
-      variables: {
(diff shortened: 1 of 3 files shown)


api.planx.uk/saveAndReturn/utils.ts
```diff
@@ -120,7 +110,7 @@ const sendSingleApplicationEmail = async (
       reference: null,
       emailReplyToId: team.notifyPersonalisation.emailReplyToId,
     };
-    const firstSave = !session.has_user_saved;
+    const firstSave = !session.hasUserSaved;
```
DafyddLlyr, line 113: Previously, this would always evaluate to `true` as `session.has_user_saved` was `undefined`.

However, it did not set up duplicate events in Hasura on each call. This is because the event trigger watching the `lowcal_sessions.has_user_saved` column wasn't being fired as the value never changed - and overwrite from `true` to `true` does not count as a change which is good to be aware of. 

I'm not exactly clear on where the responsibility for this is - potentially the GraphQL client wasn't performing a redundant mutation, or Hasura was ignoring one.

api.planx.uk/saveAndReturn/sendEmail.test.ts
```diff
@@ -211,4 +181,42 @@ describe("Send Email endpoint", () => {
...
+        lowcal_sessions: [{ ...mockLowcalSession, has_user_saved: true }]
+      }
+    });
+    
+    await supertest(app)
+      .post(SAVE_ENDPOINT)
+      .send(data)
+      .expect(200)
+    expect(callsToSetupEventsMutation()).toBe(0);
+  });
```
gunar, line 221: Nice, succinct tests 👍

api.planx.uk/saveAndReturn/utils.ts
gunar: So it would still send an email on the first save, the code being accidentally correct?

gunar (approved):
Thanks for fixing this. This bug tells me we need to migrate to [TypeScript types that are automatically generated from the GraphQL types](https://hasura.io/blog/your-guide-to-graphql-with-typescript/#step-three-generate-the-typings) 👍

api.planx.uk/saveAndReturn/utils.ts
DafyddLlyr: Exactly 😬

DafyddLlyr:
> Thanks for fixing this. This bug tells me we need to migrate to [TypeScript types that are automatically generated from the GraphQL types](https://hasura.io/blog/your-guide-to-graphql-with-typescript/#step-three-generate-the-typings) 👍

Couldn't agree more - I was going to bring this exact thing up at the dev call today 👍

PR #1237, opened 2022-10-26
fix: file type is possibly undefined on draw boundary upload

jessicamcinchak:
https://john-opensystemslab-io.airbrake.io/projects/329753/groups/3393061708783262295


DafyddLlyr (approved):
Wish we knew what they uploaded to trigger this - we've seen this error a few times before.

Great fix though - thank you!

PR #1238, opened 2022-10-26
fix(sync): delete before insert

gunar:
The existing sync code had a bug in the scenario where staging and
production environments have the same team ids and team slugs
(which are unique contraints) but where each environment would have these
ids and slugs be in different pairings.

Here instead of upsert we're doing a delete-then-insert mechanism.


jessicamcinchak (approved):
steps feel clear, thanks for moving this forward!

gunar:
:wink:

PR #1240, opened 2022-10-27
Refactor workflows to use path filter action

ghost:
A small refactor of our file change stage to delegate filtering complexity to a Github Action.

This new action should allow us to add more complex filtering logic as this evolves but this change preserves current behavior.


gunar (approved):
Looks good to me. Thanks for this. Although I was the one to set this up, you're welcome to always request review from @theopensystemslab/planx so that the whole team gets tagged.

:shipit:

PR #1243, opened 2022-10-28
test: Add test coverage report to CI

ghost:
This PR adds a test coverage reporting for the API.

A summary of coverage is reported on for each test run
A full coverage report is available with `pnpm test:coverage`
Test coverage is checked in the CI pipeline (see note below about minimum coverage threshold)

image

I'm suggesting here that we initially fix test coverage of functions to not drop below the current coverage of  ~45%. I'm keen to gradually improve this over time up to 100% so this can ratchet this up as more tests are added. Happy to discuss and/or remove this if anyone objects.

### DIFF
--- a/api.planx.uk/package.json
+++ b/api.planx.uk/package.json
@@ -38,6 +38,7 @@
     "dev": "ts-node-dev --files index.ts",
     "start": "node ./dist/index.js",
     "test": "TZ=Europe/London jest",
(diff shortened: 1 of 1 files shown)


jessicamcinchak (approved):
I'm honestly never 100% sure about exactly how to interpret or what to make of coverage charts, but very happy to give it a go and try to pay more attention to metrics like this!

Summary in CI step is nice :+1: 
Screenshot from 2022-10-28 16-30-21

PR #1248, opened 2022-11-02
Inject passport variables into rich text editor output

peterszerzo:
Initially part of #1192, plucked off of it to reduce scope.


peterszerzo:
Abandoning in favor of #1263 to make rebasing easier.

PR #1249, opened 2022-11-04
Feat: Send GeoJSON to Uniform

ghost:
https://trello.com/c/U8m3NL5I/2122-send-copy-of-site-map-image-to-uniform-as-a-geojson

This PR adds some dynamic file generation to the send component so that Uniform validators have access to an individual GeoJSON file for manual review. 

This could be extended to also include an HTML page (and/or derived pdf/image) as an additional file (I'll address this in a follow-up PR).

A few notes about this implementation:

- Mocking zip creation was a bit of a pain because of the way jest mocks are hoisted. I'd like to clean this up eventually but I haven't worked out how to tidy this up without breaking the mocking behavior.
- I've created a helper file to share some breadcrumb filtering code between bops and uniform. This is a placeholder to either find a better home for this kind of shared function or to create one. I think a generic breadcrumb/passport library would be helpful.
- I'd like to add much more testing around uniform sending but, as discussed, this can emerge gradually and over several PRs.

### DIFF
--- a/editor.planx.uk/src/@planx/components/Send/uniform/index.ts
+++ b/editor.planx.uk/src/@planx/components/Send/uniform/index.ts
@@ -51,7 +51,7 @@ export function getUniformParams(
 
   // this is the body we'll POST to the /uniform endpoint - the endpoint will handle file & .zip generation
   return {
-    xml: makeXmlString({ passport, sessionId, uniqueFiles, hasBoundary }),
+    xml: makeXmlString({ passport, sessionId, files: uniqueFiles, hasBoundary }),
     csv: makeCsvData(breadcrumbs, flow, passport, sessionId),
     geojson: geoJSONBoundary,
     files: uniqueFiles,


editor.planx.uk/src/api/upload.ts
```diff
@@ -6,16 +6,18 @@ async function uploadFile(
...
-    },
-  });
-  const token = await res.json();
+    {
+      headers: {
+        "Content-Type": "application/json",
+      },
+    }
+  );
+  const token = JSON.parse(res.data);
```
jessicamcinchak, line 20: I know this is still a draft, so apologies if overstepping, but I was peaking through initials changes and realized I might have failed to explain something here! 

The way that we send files to Uniform is not via S3, which is what this file/endpoint handles (for user-uploaded files specifically through the FileUpload component), but rather by generating a temporary zip folder locally and attaching that to Idox's Submission API (which unfortunately does not have normal internet documentation I can link, but there is a Postman Collection you can find in 1Password that's the closest thing we have to docs from the Idox team!). 

On the frontend, it might be helpful to look at `getUniformParams` which [returns](https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/src/%40planx/components/Send/uniform/index.ts#L42-L48) a general outline of files that Uniform will require. Then, follow that through to [this API file](https://github.com/theopensystemslab/planx-new/blob/main/api.planx.uk/send/uniform.js), which outlines how we generate files from those params and actually submit them to Uniform! 

Of course always many ways to do the thing and improvements very welcome - but just want to clear up any S3 file upload versus local file stuff early! Please just shout if anything still isn't clear!

api.planx.uk/send/uniform.js
```diff
@@ -203,6 +210,11 @@ async function createZip(stringXml, csv, files, sessionId) {
     zip.addLocalFile(csvPath);
     deleteFile(csvPath);
 
+  // build an optional GeoJSON file for validators
+  if (geojson) {
+    const geoBuff = Buffer.from(JSON.stringify(geojson, null, 2));
+    zip.addFile("boundaryGeoJSON.json", geoBuff);
```
jessicamcinchak, line 216: we'll need to make sure this hardcoded file name is also present in the XML ([see here](https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/src/%40planx/components/Send/uniform/xml.ts#L34-L45)) - I think we forgot to mention that these XML entries are what tell the Uniform DMS to find/download each file from our zip ! without an entry, it's likely we'd successfully send the file, but it wouldn't actually be pulled down alongside other application files. reference "other" should be fine for this first go

api.planx.uk/send/uniform.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      csv: [["1", "2", "3"]],
+      geojson: undefined,
+      files: [],
+      sessionId: "123",
+    };
+    await createZip(payload);
+    expect(mockAddFile).not.toHaveBeenCalled();
+    expect(mockWriteZip).toHaveBeenCalledTimes(1);
+  });
+});
```
jessicamcinchak, line 63: :raised_hands: initial tests & framework for future mocking very much appreciated here!

editor.planx.uk/src/@planx/components/Send/helpers.ts
```diff
@@ -0,0 +1,22 @@
...
+    if (breadcrumbData) {
+      // scan the breadcrumb's data object (what got saved to passport)
+      // and extract the first instance of any geojson that's found
+      const geojson = Object.values(breadcrumbData).find(
+        (v) => v.type === "Feature"
+      );
+      return geojson;
+    }
+  }
+}
```
jessicamcinchak, line 22: :heavy_plus_sign: thanks for introducing shared helpers here, definitely makes sense!

api.planx.uk/send/uniform.js
jessicamcinchak: one more thought that might be interesting to keep thinking about during UAT: I wonder if .geojson file extension versus .json would be more likely to "automatically open" with whatever spatial software councils likely have installed on their computers?

editor.planx.uk/src/@planx/components/Send/helpers.ts
```diff
@@ -0,0 +1,22 @@
+import { TYPES } from "@planx/components/types";
+import { Store } from "pages/FlowEditor/lib/store";
+
+export function findGeoJSON(
+  flow: Store.flow,
+  breadcrumbs: Store.breadcrumbs
+): { type: "Feature" } | undefined {
+  const boundaryQuestionID = Object.keys(breadcrumbs).find(
+    (questionId) => flow[questionId]?.type === TYPES.DrawBoundary
```
jessicamcinchak, line 9: nit: rather than `questionId` we often refer to keys of the breadcrumbs object as `nodeId` (to avoid confusion with "question" also being a specific component type too)

jessicamcinchak (commented):
Took a first pass through this & looks really good - thanks especially for adjusting approach to generating file via API and introducing some initial tests!! Feels like you found that nice balance of MVP _and_ solid foundation-setting/good practices here :slightly_smiling_face: 

Just shout if any comments / small edits aren't clear!

editor.planx.uk/src/@planx/components/Send/uniform/xml.ts
```diff
@@ -42,6 +42,10 @@ export function makeXmlString(
       <common:FileName>application.csv</common:FileName>
       <common:Reference>Other</common:Reference>
     </common:FileAttachment>
+    <common:FileAttachment>
+      <common:FileName>boundary.geojson</common:FileName>
+      <common:Reference>Other</common:Reference>
+    </common:FileAttachment>
```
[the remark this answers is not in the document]
DafyddLlyr: It would also then be useful to have a test case for this in `xml.test.ts`, so that when we merge https://github.com/theopensystemslab/planx-new/pull/1220 we can be sure that the logic has been correctly implemented there.

DafyddLlyr (changes requested):
Looking great - nice and clear and really appreciate the tests. Only one small bit of feedback that I picked up on.

Here's what I think the next stage of testing would look like. It would be good to pair on this I think as there's a few submission tricks I'm not sure we've covered (e.g. how to not pay for an application...!).

 - Send an application to the Wycombe Staging Uniform instance
   - I use Wycombe council offices as as address (HP11 1BB)
   - This is because only Wycombe has a Uniform staging DMS (Document Management System) which allows partners to actually receive and interrogate submitted documents
- Ask Kev to check that application and geojson have been received
  - We'll need to provide the submission id for reference
  - Ideally this wouldn't be Kev - I think we should loop Emily in here as well as she mentioned she had other contacts at Wycombe.

api.planx.uk/send/uniform.js
```diff
@@ -225,16 +231,14 @@ async function createZip({ xml: stringXml, csv, geojson, files, sessionId }) {
...
   zip.addLocalFile(csvPath);
   deleteFile(csvPath);
 
-  // build a temporary GeoJSON file
-  const geojsonPath = path.join(tmpDir, "boundary.json");
-  await writeFile(geojsonPath, JSON.stringify(geojson, null, 2));
-  zip.addLocalFile(geojsonPath);
-  deleteFile(geojsonPath);
-
+  // build an optional GeoJSON file for validators
```
DafyddLlyr, line 237: 👍

DafyddLlyr (approved):
Perfect - thanks very much for those changes.

PR #1250, opened 2022-11-04
fix(sync): id sequence

gunar:


gunar:
[Bug publishing Flows on Staging](https://trello.com/c/fMnVFVy5/2187-bug-publishing-flows-on-staging)

jessicamcinchak (approved):
Thanks for this ! Just spent some time on the pizza and can publish flows as expected, which I'm not currently able to do on staging or prior pizzas. 

I am a little surprised that this change doesn't touch `operations` since all of the airbrake errors yesterday emerged as operations foreign key violations, and operations only links to flows, not publshed_flows. But seems to fix the publishing bug for now, and we can keep an eye out if operations & actor_id/user_id bugs emerge again later on.

PR #1251, opened 2022-11-04
Feat: Add explicit linting rules

ghost:
Adding some baseline linting rules from https://typescript-eslint.io/

I marked this as a feature because it does have a few logical changes (e.g. the route `/gis/:localAuthority` was calling rather than passing a function - also see the `Promise.race` as a replacement for `async` calls within a promise).

### DIFF
--- a/api.planx.uk/gis/index.test.ts
+++ b/api.planx.uk/gis/index.test.ts
@@ -17,7 +17,7 @@ describe("locationSearchWithTimeout", () => {
       { x: 50, y: 50, siteBoundary: "[]" },
       timeout
     );
-    expect(promise).resolves.toStrictEqual(expect.any(Object));
+    return expect(promise).resolves.toStrictEqual(expect.any(Object));
   });
 
   test("an immediate timeout", async () => {


api.planx.uk/.eslintrc
```diff
@@ -0,0 +1,37 @@
...
+  "parser": "@typescript-eslint/parser",
+  "plugins": ["@typescript-eslint"],
+  "extends": [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "prettier"
+  ],
+  "rules": {
+    "@typescript-eslint/no-non-null-assertion": "warn",
+  },
```
DafyddLlyr, line 12: nit / opinion:  Could we add the following rule here?

```json
"@typescript-eslint/no-unused-vars": ["warn", { "varsIgnorePattern": "^_", "argsIgnorePattern": "^_"  }]
```

This follows the pattern we're currently using and I personally find it sometimes to be a bit more explicit than just commas. Not a strongly held opinion!

DafyddLlyr (approved):
This is very helpful thanks and will help reduce a lot of noise in PRs hopefully, especially once we've addressed all issues and do some automatic linting 👍 

One small question / comment but happy for this to move forward as is.

PR #1258, opened 2022-11-14
feat: Add published_flows.version computed field

DafyddLlyr:
## What does this PR do?
 - Adds the computed field `version` to the `published_flows` table
 - Updates queries to for published_flows to sort by version, not by id
 - A few `gql` tag tidy ups here and there for better DX

## Why?
- Issue raised here on Slack - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1668361260199859
- Data sync script does not update the `published_flows_id_seq` sequence, so newly published flows on staging appear not be published. However, it's just a sorting issue - they are there, just not returned when sorting by id.

## Questions / discussion
This raised a few issues for me which I'm not totally clear on...

- Why do we always insert to published flows? Is there any value to keeping old published flows?
- Was sorting by id doing something I was missing? `created_at` would have also worked for this.
- This could have been solved up just updating `published_flows_id_seq` after the weekly data sync, but this felt like a better solution to me. However, all sequences on staging are now not in the correct place which could lead to other similar issues if relied on elsewhere. Is there an overall more robusst approach here (e.g. pgdump)?

### DIFF
--- a/hasura.planx.uk/migrations/1668434843791_run_sql_migration/up.sql
+++ b/hasura.planx.uk/migrations/1668434843791_run_sql_migration/up.sql
@@ -1,6 +1,0 @@
-CREATE OR REPLACE FUNCTION get_published_flow_version(pf published_flows)
-RETURNS INTEGER AS $$
-    SELECT max(o.version) FROM operations o
-    WHERE o.flow_id = pf.flow_id
-    AND o.updated_at < pf.created_at
-$$ LANGUAGE sql STABLE;
--- a/api.planx.uk/helpers.ts
+++ b/api.planx.uk/helpers.ts
@@ -26,7 +26,7 @@ const getMostRecentPublishedFlow = async (id: string): Promise<Flow["data"]> => 
     gql`
       query GetMostRecentPublishedFlow($id: uuid!) {
         flows_by_pk(id: $id) {
-          published_flows(limit: 1, order_by: { version: desc }) {
+          published_flows(limit: 1, order_by: { created_at: desc }) {
             data
           }
         }
@@ -47,7 +47,7 @@ const getPublishedFlowByDate = async (id: string, created_at: string) => {
         flows_by_pk(id: $id) {
           published_flows(
             limit: 1,
-            order_by: { version: desc },
+            order_by: { created_at: desc },
             where: { created_at: {_lte: $created_at} }
           ) {
             data
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
@@ -265,7 +265,7 @@ export const editorStore = (
       query: gql`
         query GetFlow($id: uuid) {
           flows(limit: 1, where: { id: { _eq: $id } }) {
-            published_flows(order_by: { version: desc }, limit: 1) {
+            published_flows(order_by: { created_at: desc }, limit: 1) {
               created_at
             }
           }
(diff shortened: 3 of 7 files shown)


gunar:
Does this mean we can remove `fix_id_sequence.sql` and its invocation from `scripts/seed-database/seed-db.sh`?

hasura.planx.uk/migrations/1668434843791_run_sql_migration/up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE OR REPLACE FUNCTION get_published_flow_version(pf published_flows)
```
gunar, line 1: Instead of adding a new sql function and correlated computed field, wouldn't it be simpler to sort by `created_at`?

gunar (commented):
> Why do we always insert to published flows? Is there any value to keeping old published flows?

There's so much good work that goes into creating PlanX flows that I think it'd be irresponsible not to keep older versions. I'd be happy to entertain the idea of keeping just the "diffs" but it can get tricky pretty quickly.

>  Was sorting by id doing something I was missing? created_at would have also worked for this.

Yeah this was weird. I couldn't find anything on it so I must assume it wasn't doing anything funny and that we can switch over to `created_at`.

>  This could have been solved up just updating published_flows_id_seq after the weekly data sync, but this felt like a better solution to me.

I agree the id sequence thing must go (one way or the other).

> However, all sequences on staging are now not in the correct place which could lead to other similar issues if relied on elsewhere.

I'd add this to your `up.sql` file:

```sql
SELECT setval('published_flows_id_seq', max(id)) FROM published_flows;
```

> Is there an overall more robusst approach here (e.g. pgdump)?

Always open to suggestions.

DafyddLlyr:
Hey guys, I'd love an opinion on if you think `published_flow.version` helpful / worthwhile?

There's no immediate benefit here - functionally sorting by `created_at` would deliver us the same result.

I see the benefit as `version` being more meaningful and explicit if comparing a flow and it's related published flows, and possibly in future more semantic versioning for the end users.

jessicamcinchak:
Hey sorry I've been slow to this one! 

Here's a couple thoughts / answers to questions you posed: 
- Why do we always insert to published flows? Is there any value to keeping old published flows?
  - We insert because of the flattened nature (eg snapshot of this flow and all it's nested flows/external portals at point in time), but it's just a lazy solution to not having to work out how to recreate a "published" state from operations etc
- Was sorting by id doing something I was missing? created_at would have also worked for this.
  - My guess here would simply be that id is indexed, and therefore a bit more performant to sort on than a date?
- This could have been solved up just updating published_flows_id_seq after the weekly data sync, but this felt like a better solution to me. However, all sequences on staging are now not in the correct place which could lead to other similar issues if relied on elsewhere. Is there an overall more robusst approach here (e.g. pgdump)?
  - I obviously am also pro-pgdump for data syncing so we don't have think so hard about these various dependencies between flows<>published_flows<>operation. And as a once-weekly overnight operation currently, I don't think we have to be too concerned about the time difference between this query approach or a pgdump (which will likely be slower?)
  
Regarding the computed field, I am curious from a performance perspective if it's quite expensive to sort on it? It likely cannot be indexed, correct? 

While I appreciate this work I will be honest I'm not sure I totally understand the immediate need/benefit overall besides fixing the data sync! I think there's a lot of room for improvement around published_flows table and how we use it, and I guess I'd prefer to prioritize making a more holistic plan and not introduce too many computed fields/psql functions as bandaids in the meantime?

DafyddLlyr:
Thanks both for the feedback here - appreciate it!

I've now...
 - Dropped `version` in favour of a more holistic future improvement
 - Sorted by `created_at`
 - Recreated Vultr environment 

I don't think that the suggestion of adding `SELECT setval('published_flows_id_seq', max(id)) FROM published_flows;` to a migration script will work, as these run before any data is populated (that's my understanding...!).

jessicamcinchak (approved):

PR #1260, opened 2022-11-15
fix: Remove Material-UI v4, fix Editor button styling

DafyddLlyr:
## Context
- Buttons appeared visually incorrect when the node edit modal was open in the Editor
- This was flagged both [here on a review of the rich text editor](https://github.com/theopensystemslab/planx-new/pull/1192#pullrequestreview-1175513882) and [here on Slack](https://opensystemslab.slack.com/archives/C5Q59R3HB/p1668444153093379)

**"Create" button incorrectly styled**
image

**"Publish" button style changing when modal open**
image

## What caused this?
- When the node modal was open, an additional instance of the `MuiButtonBase-root` class was being applied to buttons which was overwriting our styling
- This was caused by a double import of Material UI styles, caused by an import from MUI v4 as part of the Rich Text Editor PR (comment on code)

## Solution
- Fix import
- Update other instances of MUI v4 being imported
- Update Eslint rules
- Remove Material-UI v4! Apologies, this should have been done sooner - docs here: https://mui.com/material-ui/migration/migration-v4/#remove-old-packages

## To test
 - Buttons are styled correctly on the Pizza


editor.planx.uk/src/ui/RichTextInput.tsx
```diff
@@ -1,6 +1,5 @@
 import "./RichTextInput.css";
 
-import IconButton from "@material-ui/core/IconButton";
```
DafyddLlyr, line 3: This was the problem line that broke how buttons appeared, as it was importing Material UI v4 styles which were being re-applied to buttons.

gunar (approved):
:+1:

image

PR #1263, opened 2022-11-15
Inject passport variables into rich text editor output

peterszerzo:
Initially part of #1192, plucked off of it to reduce scope.

### DIFF
--- a/editor.planx.uk/src/ui/RichTextInput.tsx
+++ b/editor.planx.uk/src/ui/RichTextInput.tsx
@@ -36,6 +36,7 @@ import {
   ReactRenderer,
   useEditor,
 } from "@tiptap/react";
+import { map } from "ramda";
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/ui/RichTextInput.tsx
```diff
@@ -126,6 +126,52 @@ export const fromHtml = (htmlString: string) => {
...
+        : null;
+    })(doc)
+  );
+};
+
+/**
+ * Traverse a nested object/array and apply a modification at each level. If the modifier returns `null`, it leaves the result unchanged.
+ * Used to inject placeholder values into a document structure.
+ */
+const modifyDeep =
```
gunar, line 152: Can we get unit tests for this? :upside_down_face:

PR #1268, opened 2022-11-16
Reproduce performance issue

peterszerzo:
Setting up a new story to look into performance issues/regressions reported in nodes like https://editor.planx.dev/opensystemslab/permitteddevelopment/nodes/_root/nodes/eiN5C9wgDe/edit.

You can access the story under http://localhost:6006/?path=/story/design-system-atoms-form-elements-richtextinput--performance.

Open questions:
- how much of the performance hit is caused by `RichTextEditor`, `ListManager`, or any Material UI rendering overhead in general? (you can change the editor between `RichTextInput` and `Input` to see the difference).
- where can improvements be made?

Screen Shot 2022-11-16 at 9 49 30 AM

### DIFF
--- a/editor.planx.uk/src/@planx/components/Checklist/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/Checklist/Editor.tsx
@@ -372,7 +372,7 @@ export const ChecklistComponent: React.FC<ChecklistProps> = (props) => {
             </InputRow>
 
             <InputRow>
-              <RichTextInput
+              <Input
                 name="description"
                 value={formik.values.description}
                 placeholder="Description"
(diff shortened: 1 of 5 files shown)


gunar:
[Text input takes 1s to respond to keyboard input](https://trello.com/c/zoYs9u8U/2204-text-input-takes-1s-to-respond-to-keyboard-input)

peterszerzo:
(this is going to be a living set of notes)

Here are some preliminary findings:
- `<RichTextInput />` renders twice on each keystroke, for reasons that so far seem completely internal.
- on the 200-list example, one keypress event takes 500ms on my computer with `<RichTextInput />` and 250ms when it is swapped with a regular `<Input />`, so the problem is somewhat rich text-related but not entirely rich text-related. Curious to see how old rich text compares to new rich text.
- `<ListManager />` stuff e.g. drag-and-drop, update logic seems to have no noticeable performance impact (although I am switching to `ramda` methods for updating the list, just in case it matters).

Potential solutions:
- could we make sections of that node collapsible?

jessicamcinchak:
Thank you for all these notes, we'll definitely reference them again - but we think this fix will ultimately be a logic task, not related to RichText, so closing this PR out!

PR #1269, opened 2022-11-16
fix: filenames should be stored encoded in planx passport

jessicamcinchak:
By referencing prior application logs (`bops_applications` etc) and commit history (eg prior use of `s3.getSignedUrlPromise()` [here](https://github.com/theopensystemslab/planx-new/pull/1084/files#diff-b586e54c057fe718a94735b71eba906153dc3b470962fca637586f853bcb8f73L39) which returned an encoded URL) we confirmed that we used to store encoded filenames in the planx passport, and therefore send encoded filenames to BOPS. Filenames on S3 are currently not, and have never been, encoded.

This attempts to quickly patch that regression, recognizing that we may want to reconsider a more robust approach in the future  and will surely want to improve test coverage in the short term.

Here's a minimal flow for testing that will submit to BOPS staging: https://1269.planx.pizza/buckinghamshire/files-test/preview (but BOPS will need to switch back to our Docker file-api-key to access, not use a Pulumi managed one yet)

To test:
- Upload a filename with spaces, confirm in the Hasura Event log that it succeeded with 200 (broken on staging now)
- Upload a filename without spaces, confirm in the Hasura Event log that it succeeded with 200 (same as staging now)
- Confirm that the file URL is stored encoded in the PlanX passport (if in a /preview window, open your console and type `window.api.getState().computePassport()` to inspect from any point in a flow!)


jessicamcinchak:
@DafyddLlyr added helper methods sourced from here to handle further encoding! https://github.com/GeorgePhillips/node-s3-url-encode/blob/master/index.js

My latest test has succeeded :crossed_fingers:

DafyddLlyr (approved):
🏴󠁧󠁢󠁷󠁬󠁳󠁿

PR #1271, opened 2022-11-16
chore(api): use pnpm dev on local environment

gunar:
Allows for live-reload on development environments.
Doesn't affect staging/production as these are deployed using pulumi (not docker compose).
Might make pizzas slightly less performant but overall shouldn't affect it that much.


gunar:
Hey @theopensystemslab/planx, can we review/approve/merge this?

jessicamcinchak (approved):

PR #1275, opened 2022-11-17
feat: Generate full API url for uploaded file on the server

DafyddLlyr:
## What does this PR do?
 - Offers a (hopefully!) more robust approach to S3 file name handling than the fix offered by https://github.com/theopensystemslab/planx-new/pull/1269
 - Reverts back to our previous method of generating the fileUrl server side instead of client side
 - Tidies up some redundant code
 - (In theory!) Should have zero affect on submissions or ongoing sessions - this just controls how the URL is generated
 - Enhances test suite
   - I've split out the s3 tests from `server.test.js`
   - I've annotated the updated test cases to help highlight them 
   - Once merged I'm happy to pick up increasing code coverage here as another PR

**Testing**
- This should be tested on Uniform (+DMS) and BOPS staging before we progress it, just so that we have a good degree of confidence that it doesn't break anything.


api.planx.uk/s3/s3.test.ts
```diff
@@ -0,0 +1,196 @@
...
+
+const mockGetObject = jest.fn(() => ({
+  promise: () => Promise.resolve(getObjectResponse)
+}))
+
+const mockGetSignedUrl = jest.fn(() => {
+  const randomFolderName = "nanoid";
+  const modifiedKey = "modified%20key";
+  return `
+    https://test-bucket.s3.eu-west-2.amazonaws.com/${randomFolderName}/${modifiedKey}?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-SignedHeaders=host
```
DafyddLlyr, line 19: Hopefully this is explicit enough to test that "we take the file name that S3 provides us with".

api.planx.uk/s3/s3.test.ts
```diff
@@ -0,0 +1,196 @@
...
+      .post("/private-file-upload")
+      .field("filename", "some filename")
+      .expect(422)
+      .then(res => {
+        expect(mockPutObject).not.toHaveBeenCalled();
+        expect(res.body.error).toBe("missing file")
+      })
+  });
+
+  it("private-file-upload - should upload file", async () => {
```
DafyddLlyr, line 70: Updated test

api.planx.uk/s3/s3.test.ts
```diff
@@ -0,0 +1,196 @@
...
+      .post("/public-file-upload")
+      .field("filename", "some filename")
+      .expect(422)
+      .then(res => {
+        expect(mockPutObject).not.toHaveBeenCalled();
+        expect(res.body.error).toBe("missing file")
+      })
+  });
+
+  it("public-file-upload - should upload file", async () => {
```
DafyddLlyr, line 96: Updated test

api.planx.uk/s3/uploadFile.ts
```diff
@@ -33,13 +34,22 @@ export const uploadPrivateFile = async (
...
-    key,
+    fileUrl,
   };
 };
 
+// Construct an API URL for the uploaded file
+const buildFileUrl = async (key: string, path: "public" | "private") => {
+  const s3 = s3Factory();
+  const s3Url = await new URL(s3.getSignedUrl('getObject', { Key: key }));
+  const fileUrl = `${process.env.API_URL_EXT}/file/${path}${s3Url.pathname}`
```
DafyddLlyr, line 49: `pathname` has a leading `/` so this is correct despite looking weird!

api.planx.uk/s3/s3.test.ts
```diff
@@ -0,0 +1,196 @@
...
+}))
+
+let getObjectResponse = {};
+
+const mockGetObject = jest.fn(() => ({
+  promise: () => Promise.resolve(getObjectResponse)
+}))
+
+const mockGetSignedUrl = jest.fn(() => {
+  const randomFolderName = "nanoid";
```
gunar, line 16: should this be ` = nanoid()`?

api.planx.uk/s3/s3.test.ts
```diff
@@ -0,0 +1,196 @@
...
+    getObject: mockGetObject,
+    getSignedUrl: mockGetSignedUrl,
+  };
+};
+
+jest.mock("aws-sdk/clients/s3", () => {
+  return jest.fn().mockImplementation(() => {
+    return s3Mock();
+  })
+});
```
gunar, line 35: FWIW I would've been fine with a refactor into dependency injection as well.

api.planx.uk/s3/s3.test.ts
```diff
@@ -0,0 +1,196 @@
...
+
+jest.mock("nanoid", () => ({
+  ...(jest.requireActual("nanoid")),
+  customAlphabet: jest.fn(() => () => "nanoid")
```
gunar, line 39: Sorry, why do we need this?

api.planx.uk/s3/uploadFile.ts
```diff
@@ -33,13 +34,22 @@ export const uploadPrivateFile = async (
...
     file_type: fileType,
-    key,
+    fileUrl,
   };
 };
 
+// Construct an API URL for the uploaded file
+const buildFileUrl = async (key: string, path: "public" | "private") => {
+  const s3 = s3Factory();
+  const s3Url = await new URL(s3.getSignedUrl('getObject', { Key: key }));
```
gunar, line 48: Shouldn't the `await` be placed immediately before the `s3` method?

```diff
- const s3Url = await new URL(s3.getSignedUrl('getObject', { Key: key }));
+ const s3Url = new URL(await s3.getSignedUrl('getObject', { Key: key }));
```

api.planx.uk/s3/uploadFile.ts
gunar: This commend might be worth of becoming a code-comment.

api.planx.uk/s3/uploadFile.ts
gunar: Also, won't the URL returned by `s3.getSignedUrl` expire at some point?

gunar (commented):
Did an initial code-only review. Had a couple of questions but I like the new organization of the code.

api.planx.uk/s3/s3.test.ts
DafyddLlyr: This was intentional, so that in the test we can do things like...

```ts
expect(res.body).toEqual({
  file_type: "text/plain",
  fileUrl: expect.stringContaining("/file/private/nanoid/modified%20key")
});
```

api.planx.uk/s3/s3.test.ts
DafyddLlyr: Good catch - this is leftover from a refactor 👌 Tidied up!

api.planx.uk/s3/uploadFile.ts
DafyddLlyr: I believe that this was what I had initially, and I just tried it again.

`const s3Url = new URL(await s3.getSignedUrl('getObject', { Key: key }));` returns a tslint warning `'await' has no effect on the type of this expression.ts(80007)`

Turns out this is because `getSignedUrl()` is actually synchronous - I've not just dropped the `await` altogether.

Docs: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#getSignedUrl

DafyddLlyr: > Also, won't the URL returned by s3.getSignedUrl expire at some point?

Yeah this is an interesting one.

We're only calling `getSignedUrl()` here to get the `URL.pathname` to build a link to our own API, and we use this to get the file key which is later used in `getFileFromS3()`. The fact that the URL will expire doesn't matter - the `pathname` value is consistent.

This feels like a bit of an odd way to just get a key for a file, but it's a reversion to our previous method of doing this and allows us to get the key correctly encoded without worrying about encoding / decoding using the same method that S3 uses.

Also: I'm pretty sure the files are still public on S3 unless I'm mistaken? We're just not exposing that bucket URL any longer, and routing consumers through our API where we check an API key. It's been a while since we originally discussed this but I'd imagine we want to address this right?

gunar: Yes, we'll want to lock down the S3 bucket so that only our API service has access to it, and all public access will need to go through our API service.


If `pathname` doesn't change then we're good, yes :+1:

gunar:
Updating the pizza failed :(

DafyddLlyr:
Once destroyed and re-created, and all actions re-run, we're back in business! 🙌

api.planx.uk/server.test.js
```diff
@@ -218,180 +218,3 @@ describe.skip("fetching GIS data from Digital Land for supported local authoriti
...
-
-    await supertest(app)
-      .get(`/file/private/${filePath}`)
-      .set({ 'api-key': 'test' })
-      .expect(200)
-      .then(() => {
-        expect(mockGetObject).toHaveBeenCalledTimes(1);
-      });
-  });
-});
```
jessicamcinchak, line 397: thanks for re-locating these tests to s3/ :+1:

editor.planx.uk/src/api/encode.ts
```diff
@@ -1,33 +0,0 @@
...
-};
-
-function encodeS3URI(filename: string): string {
-  return encodeURI(filename) // Do the standard url encoding
-    .replace(/(\+|!|"|#|\$|&|'|\(|\)|\*|\+|,|:|;|=|\?|@)/gim, function (match) {
-      return encodings[match];
-    });
-}
-
-export { encodeS3URI };
```
jessicamcinchak, line 33: :+1: great to see this removed

editor.planx.uk/src/api/upload.ts
```diff
@@ -14,21 +12,16 @@ async function uploadPublicFile(
...
   file: any,
   { onProgress }: { onProgress?: (p: any) => void } = {}
 ) {
   const { data } = await handleUpload(file, { onProgress, path: "private" });
-  const passport = useStore.getState().computePassport();
 
-  return useStagingUrlIfTestApplication(passport)(
-    `${process.env.REACT_APP_API_URL}/file/private/${encodeS3URI(data.key)}`
-  );
+  return data.fileUrl;
```
jessicamcinchak, line 24: I'm not sure I understand why this isn't now ```return useStagingUrlIfTestApplication(passport)(data.fileUrl)``` ?

jessicamcinchak (approved):
appreciate this cleanup! pizza seems to be working as expected. i did leave one comment/question about `useStagingUrlIfTestApplication` which isn't exactly test-able on the pizza though. 

otherwise seems like this has had multiple eyes & a thorough review and happy for it to merge when you are!

editor.planx.uk/src/api/upload.ts
DafyddLlyr: This is a good question, and important to be clear on 👍 

`handleUpload()` still conditionally routes private uploads to the staging API if the user is "test test".

So the return value here will be coming from the staging API if the user is "test test" and will already be constructed correctly.

jessicamcinchak: that makes sense, thank you!!

PR #1277, opened 2022-11-17
fix: Omit authorization from BOPS audit table

DafyddLlyr:
## Problem
- Since https://github.com/theopensystemslab/planx-new/pull/1227 was merged (which enforces auth to send to BOPS), these auth headers have been stored in plain text in the `bops_applications` audit table

## Solution
- Omit the offending header when saving to the audit table
- I'll clean up any BOPS submissions since the above PR was merged just for due diligence here


jessicamcinchak (approved):
thanks for catching this!

DafyddLlyr:
Staging and Prod DBs have been cleaned up with the following SQL - 

```sql
UPDATE bops_applications 
SET req_headers = req_headers - 'authorization'
```

PR #1278, opened 2022-11-17
fix: fetch files urls rather than using S3 methods internally

jessicamcinchak:
When we encoded our stored URLs to fix BOPS yesterday, we inadvertantly broke Uniform because the internal S3 methods were attempting to match our encoded URLs to the un-encoded URLs stored on S3.

This PR reverts back to our original way of downloading Uniform files, which simply fetches our own file links.

Once merged and deployed, we have at least 1 Uniform application that will need to be manually submitted.

Testing flow: https://1278.planx.pizza/buckinghamshire/temporary-uniform-files-test/preview?analytics=false


api.planx.uk/.env.test.example
```diff
@@ -41,4 +41,3 @@ UNIFORM_TOKEN_URL=👻
 UNIFORM_SUBMISSION_URL=👻
 
 SLACK_WEBHOOK_URL=👻
-FILE_API_KEY=👻
```
jessicamcinchak, line 44: this is just cleaning up a duplicate entry

api.planx.uk/hasura/metadata/index.ts
```diff
@@ -57,7 +57,7 @@ const createScheduledEvent = async (args: RequiredScheduledEventArgs) => {
           value_from_env: "HASURA_PLANX_API_KEY"
         }],
         retry_conf: {
-          num_retries: 3,
+          num_retries: 1,
```
jessicamcinchak, line 60: this is actually the change that I intended to make earlier in #1273 - which actually only adjusted the Slack _notification_ retry, not the actual send event - it made my testing feedback loop faster here

api.planx.uk/send/uniform.js
```diff
@@ -419,14 +416,19 @@ async function retrieveSubmission(token, submissionId) {
...
  *
- * @param {string} filePath - s3 `path/key` to file
+ * @param {string} url - our file URL, eg api.planx.uk/file/private/path/key
  * @param {string} path - file name for download
  * @param {string} folder - AdmZip archive
  */
-const downloadFile = async (filePath, path, folder) => {
-  const { body } = await getFileFromS3(filePath);
+const downloadFile = async (url, path, folder) => {
+  const res = await fetch(url, { headers: { 'api-key': process.env.FILE_API_KEY } });
```
jessicamcinchak, line 424: interesting to note that without the header, the files were actually still showing up in the zip, but could not be opened. that may make for unpleasant troubleshooting in the future?

gunar: good shout. we probably want to check `res.ok`

https://developer.mozilla.org/pt-BR/docs/Web/API/Fetch_API/Using_Fetch#Checking_that_the_fetch_was_successful

jessicamcinchak: addressed this in latest push !

submissions on the pizza are failing with the below error (but succeeding locally) - so we're going to ensure those are sorted before any merge but thanks for initial phone review :upside_down_face: 
```
"message": "FetchError: request to https://api.1278.planx.pizza/file/private/yjki3e20/Screenshot%20from%202022-11-15%2015-18-51%281%29.png failed, reason: connect ECONNREFUSED 127.0.1.1:443\"}
```

DafyddLlyr:
I think the issue we’re hitting atm is with the Express API (within docker-compose) calling itself via https. I think the code as is would work on staging and prod, but at the expense of not allowing us to test locally and on Pizzas as well as we would like.

There’s probably a work around with env vars where we can modify the URL depending the environment.

Just been thinking about it and the only real reason we’re doing an extra external call to the API is for Express do decode the URL for us right? Under the hood it’s using the exact same function `getFileFromS3()`. 

The following is pretty straightforward and works on all funky filenames I've found so far - 

```ts
const downloadFile = async (url, path, folder) => {
  const s3Key = url.split("/").slice(-2).join("/");
  const decodedS3Key = decodeURIComponent(s3Key);
  const { body } = await getFileFromS3(decodedS3Key);

  fs.writeFileSync(path, body);

  folder.addLocalFile(path);
  deleteFile(path);
};
```

DafyddLlyr (approved):

PR #1252, opened 2022-11-08
chore: Convert api/editor to TS

DafyddLlyr:
## What does this PR do?

- Convert `api.planx.uk/editor` from JavaScript to TypeScript
- A few additional tidy ups to helper functions

PR #1282, opened 2022-11-19
test: Increase test coverage for API /webhooks & /hasura-metadata dirs

DafyddLlyr:
## What does this PR do?
 - Brings test coverage for the API `/webhook` and `hasura-metadata` modules up to 100%
 - Bumps API global functions coverage threshold to 60%
 - Delivers on this promise from last week 😅 https://github.com/theopensystemslab/planx-new/pull/1252#discussion_r1016424337

### DIFF
--- a/api.planx.uk/webhooks/hardDeleteSessions.test.ts
+++ b/api.planx.uk/webhooks/hardDeleteSessions.test.ts
@@ -3,12 +3,13 @@ import app from "../server";
 import { queryMock } from "../tests/graphqlQueryMock";
 
 const { post } = supertest(app);
+const ENDPOINT = "/webhooks/hasura/delete-expired-sessions";
 
 describe("Delete expired sessions webhook", () => {
   afterEach(() => queryMock.reset());
 
   it("fails without correct authentication", () => {
-    post("/webhooks/hasura/delete-expired-sessions")
+    post(ENDPOINT)
       .expect(401)
       .then((response) => {
         expect(response.body).toEqual({
@@ -33,7 +34,7 @@ describe("Delete expired sessions webhook", () => {
       ]
     });
 
-    await post("/webhooks/hasura/delete-expired-sessions")
+    await post(ENDPOINT)
       .set({ Authorization: process.env.HASURA_PLANX_API_KEY })
       .expect(200)
       .then((response) => {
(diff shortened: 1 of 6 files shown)


api.planx.uk/webhooks/hardDeleteSessions.test.ts
```diff
@@ -0,0 +1,67 @@
...
+        delete_lowcal_sessions: {
+          returning: [
+            { id: "abc123" },
+            { id: "xyz789" },
+          ]
+        }
+      },
+      ignoreThesePropertiesInVariables: [
+        "oneWeekAgo"
+      ]
```
DafyddLlyr, line 34: I spent waaaay to long messing with Jest's fake timers trying to get a usable date for this mock query.

Things I learned - 
 - `QueryMock` has this awesome property `ignoreThesePropertiesInVariables` that allowed me to totally bypass the problem ([docs](https://github.com/zth/graphql-query-test-mock/blob/master/README.md#how-to-i-handle-unstable-variables-like-date-or-random-things-that-change-for-every-query))
 - Jest's fake timers do not work with Express. I never got to the _exact_ reason for this but it's to do with how Express uses the native timer (so Jest's implementation gets ignored?)
 - I was able to get around this using [`@sinon-js/fake-timers`](https://github.com/sinonjs/fake-timers). Something like this - 
 
 
 ```ts
 let clock: FakeTimers.InstalledClock;
  
beforeAll(() => clock = FakeTimers.install());
afterEach(() => clock.uninstall());

test("something", () => {
    clock.setSystemTime(new Date(2022, 10, 8))
    // Test stuff....
});
```

In this instance `ignoreThesePropertiesInVariables` worked great as the timing logic doesn't need to be tested, I just needed values for my mock query.

File: api.planx.uk/webhooks/sendNotifications.ts [hunk 2]
```diff
@@ -24,11 +24,8 @@ const sendSlackNotification = (req: Request, res: Response, next: NextFunction):
       }
 
       const bopsMessage = `:incoming_envelope: New BOPS submission *${data?.bops_id}* [${data?.destination_url}]`;
-      slack.send(bopsMessage)
-        .then(() => {
-          res.status(200).send({ message: "Posted to Slack", data: bopsMessage });
-        })
-        .catch(error => next(error));
```
DafyddLlyr, line 31: This error wasn't getting passed to the `catch` block of the `try/catch` but getting passed directly to `errorHandler()` via `next()`. This meant that the error was losing some of it's context.

The small refactor here reads more clearly (to me), and now throws it's error to the `try/catch` handler wrapping this function.

gunar:
Thank you for the extra test coverage :+1:

PR #1284, opened 2022-11-22
feat: Enhance FeedbackFish metadata

DafyddLlyr:
## What does this PR do?
 - Fixes how existing metadata is sent to the FeedbackFish API when user submits information via the "Report an inaccuracy" buttons
 - Send additional ("standard") metadata to allow George and Alastair more context when understanding feedback (e.g. UPRN, proposal.propertyType). This gets sent via the widget and our custom forms.
 - Adds a basic `/admin/feedback` endpoint that allows logged in Editor to download a CSV of feedback

## Context
This ticket has morphed from looking at alternatives to FeedbackFish, to improving the current integration with a view to replacing it with a custom version in the medium term. Tickets for the follow on task of designing and building this have been added to Trello.

I certainly see the current implementation of the "download feedback CSV" endpoint as a temporary measure as it relies on reverse engineering the undocumented FeedbackFish API which could theoretically change at any moment. I'd like to think in future that we could keep the endpoint but switch out the implementation.

This endpoint relies on a cookie I've added this as a query param as opposed to adding it as a secret. The idea here is that this is easier / quicker to update when required as there's no (documented) auth endpoint I can call to generate it.

**Metadata submitted to FeedbackFish**
image

### DIFF
--- a/api.planx.uk/admin/feedback/downloadFeedbackCSV.test.ts
+++ b/api.planx.uk/admin/feedback/downloadFeedbackCSV.test.ts
@@ -1,0 +1,131 @@
+import { Feedback, parseFeedback } from './downloadFeedbackCSV';
+import supertest from "supertest";
+import app from "../../server";
+import { authHeader } from "../../tests/mockJWT";
+import Axios from "axios";
+
+jest.mock("axios");
+const mockAxios = Axios as jest.Mocked<typeof Axios>;
+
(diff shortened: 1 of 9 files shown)


DafyddLlyr:
[Investigate alternatives to Feedback Fish](https://trello.com/c/DyBHULbZ/1953-investigate-alternatives-to-feedback-fish)

PR #1285, opened 2022-11-22
fix: Filenames in Uniform zip must match filenames in XML

DafyddLlyr:


api.planx.uk/send/uniform.js
```diff
@@ -211,7 +211,7 @@ export async function createZip({
       // Ensure unique filename by combining original filename and S3 folder name, which is a nanoid
       // Uniform requires all uploaded files to be present in the zip, even if they are duplicates
       // Must match unique filename in editor.planx.uk/src/@planx/components/Send/uniform/xml.ts
-      const uniqueFilename = file.split("/").slice(-2).join("-");
+      const uniqueFilename = decodeURIComponent(file.split("/").slice(-2).join("-"));
```
DafyddLlyr, line 214: Previously `decodeURIComponent()` was missing here which meant that the XML had nicely formatted file names but the zip did not.

If they don't match, the Uniform connector fails to pick up these files.

jessicamcinchak (approved):
Thanks for sorting this one out! Please shout if you need a hand (or just second set of eyes) on any resubmissions

PR #1287, opened 2022-11-22
fix: border-style kebab case warning

DafyddLlyr:
Quick fix - this is currently logging out an error when displayed, and when Editor tests are running.

image

Visually this still looks as expected - 

image
image


jessicamcinchak (approved):
thank you for cleaning this one up !

PR #1288, opened 2022-11-22
fix: Files uploaded by test users should be handled by staging

DafyddLlyr:
## What's the problem?
 - Currently, test users use the name "test test" on Production to send applications to the BOPS Staging environment
 - Applications sent via this method are failing, as they cannot access images uploaded via the PlanX Production API

## What's the solution?
 - If a user is using "test test", upload via the Staging API
 - Likewise, correctly format the URL added to their passport
   - This would be made redundant by https://github.com/theopensystemslab/planx-new/pull/1275

## What are the limitations?
 - For this to work, a user has to enter their name before uploading files. This is fine for LDC flows, but may not always hold true.
 - Needs to be tested on prod to know it works 😬 

### Notes
A HTTP interceptor would be a nice refactor here to clean things up a little if our general pattern is "a test application should use the staging API". Example Axios docs - https://axios-http.com/docs/interceptors

### DIFF
--- a/editor.planx.uk/src/api/upload.ts
+++ b/editor.planx.uk/src/api/upload.ts
@@ -1,4 +1,6 @@
+import { useStagingUrlIfTestApplication } from "@planx/components/shared/utils";
 import axios from "axios";
+import { useStore } from "pages/FlowEditor/lib/store";
 
 import { encodeS3URI } from "./encode";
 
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/api/upload.ts
```diff
@@ -40,14 +43,19 @@ function handleUpload(
...
 
+  const passport = useStore.getState().computePassport();
+
+  // Private uploads for test applicaitons should be handled by the staging environment
   const paths = {
-    public: "public-file-upload",
-    private: "private-file-upload",
+    public: `${process.env.REACT_APP_API_URL}/public-file-upload`,
+    private: useStagingUrlIfTestApplication(passport)(
+      `${process.env.REACT_APP_API_URL}/private-file-upload`
```
DafyddLlyr, line 52: This might be over cautious, but it's possible for an Editor to have a passport with "test test" whilst building a flow and uploading public files. In these cases, we don't want those files to be inaccessible from the prod public interface.

jessicamcinchak (approved):

PR #1289, opened 2022-11-22
fix: Allow public uploads to be accessed via API

DafyddLlyr:
## What's the problem?
- Publicly uploaded files are not accessible currently, and give a CORS error
- This is because files are now being served via the API

image

### Note
 - This may not work still on custom domains - we can change to "cross-origin" if required but this feels worth a shot...!
 - I'm pretty certain that these headers are set via Helmet in Express, but I'm a little unclear on Pulumi / AWS setup here. I see this as a bit of a temporary measure which is worth trying until we nail down the exact fix / implementation. Ideally we'd still limit access to certain domains if at all possible.


jessicamcinchak (approved):

gunar:
By publicly uploaded files so you mean like editor images and such?

PR #1290, opened 2022-11-22
fix: Set public image CORP to "cross-site"

DafyddLlyr:
Please see https://github.com/theopensystemslab/planx-new/pull/1289 for context

As suspected, in order to work on custom domains this value needs to be set to `cross-site`.

image

https://planningservices.buckinghamshire.gov.uk/temporary-pay-test


gunar (approved):
Always good to start out strict and lax as required.

PR #1292, opened 2022-11-24
wip: sync production users to staging & docker environments

jessicamcinchak:
**DO NOT MERGE: This code doesn't work, for demonstration & discussion only!**

**Problem:** 
The data sync applies `users` via Hasura Seeds (`users.id` have been manually adjusted to match staging for google oauth). When a new user is added via the Hasura console to a pizza environment, their pizza user id is not going to match their staging user id, therefore triggering a foreign key violation when attempting to edit a flow. 

This pevents planning officers like Emily from UAT'ing a feature that has an Editor representation, for example "Please confirm you can add a ContactInput component to a flow." 

More context here: 
- https://opensystemslab.slack.com/archives/C01E3AC0C03/p1669222564451839
- https://john-opensystemslab-io.airbrake.io/projects/329753/groups/3413445689621976952

**Proposed solutions:**
First attempt / this code:
- Extend the existing data sync script to query `users` from production and insert them into the local instance. Remove the manual Hasura seeds. This means ids would be synced across all environments and we wouldn't have to manually add users to pizzas (obviously it makes a strong case for addressing user permisisons sooner than later, but I don't think the risk of a user accessing a pizza is any greater than their current production/staging access). 
- This doesn't work unless the production graphql client in the data sync script is authenticated. `users` can only (rightly) be queried by the Hasura admin role, whereas other tables we're syncing can be queried by the public role. Authenticating the production graphql client in the data sync script makes this mix of queries & mutations begin to feel very precarious to me, and feels very risky to edit & test.

What I think we should actually do:
- Create a new pg database user with read access to necessary production tables and write access to non-production databases - this user will _only_ be responsible for data syncs
- That user does a pgdump of select tables from production (flows, teams, users, published_flows) and pg_restores into a  local/pizza/staging database where only Hasura migrations have been applied (no seed data)

Thoughts? Other proposed solutions?

**Short term pizza fix:**
Rather than adding new users via the Hasura Console add them via sql tab and match their id to their staging id. 

For example:
```INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (20, 'Jessica', 'McInchak', 'jessica@opensystemslab.io', true);```

### DIFF
--- a/hasura.planx.uk/seeds/1595528762802_teams_and_users.sql
+++ b/hasura.planx.uk/seeds/1595528762802_teams_and_users.sql
@@ -1,9 +1,9 @@
-INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (1, 'John', 'Rees', 'john@opensystemslab.io', true) ON CONFLICT (id) DO NOTHING;
-INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (2, 'Alastair', 'Parvin', 'alastair@opensystemslab.io', true) ON CONFLICT (id) DO NOTHING;
-INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (6, 'Gunar', 'Gessner', 'gunargessner@gmail.com', true) ON CONFLICT (id) DO NOTHING;
-INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (13, 'Sarah', 'Scott', 'sarah@opensystemslab.io', true) ON CONFLICT (id) DO NOTHING;
-INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (20, 'Jessica', 'McInchak', 'jessica@opensystemslab.io', true) ON CONFLICT (id) DO NOTHING;
-INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (33, 'Dafydd', 'Pearson', 'dafydd@opensystemslab.io', true) ON CONFLICT (id) DO NOTHING;
-INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (36, 'Benjamin', 'Williams', 'benjamin@opensystemslab.io', true) ON CONFLICT (id) DO NOTHING;
-SELECT setval('users_id_seq', max(id)) FROM users;
-SELECT setval('teams_id_seq', max(id)) FROM teams;
+-- INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (1, 'John', 'Rees', 'john@opensystemslab.io', true) ON CONFLICT (id) DO NOTHING;
+-- INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (2, 'Alastair', 'Parvin', 'alastair@opensystemslab.io', true) ON CONFLICT (id) DO NOTHING;
+-- INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (6, 'Gunar', 'Gessner', 'gunargessner@gmail.com', true) ON CONFLICT (id) DO NOTHING;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: hasura.planx.uk/seeds/1595528762802_teams_and_users.sql [hunk 1]
```diff
@@ -1,9 +1,9 @@
...
+-- INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (2, 'Alastair', 'Parvin', 'alastair@opensystemslab.io', true) ON CONFLICT (id) DO NOTHING;
+-- INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (6, 'Gunar', 'Gessner', 'gunargessner@gmail.com', true) ON CONFLICT (id) DO NOTHING;
+-- INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (13, 'Sarah', 'Scott', 'sarah@opensystemslab.io', true) ON CONFLICT (id) DO NOTHING;
+-- INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (20, 'Jessica', 'McInchak', 'jessica@opensystemslab.io', true) ON CONFLICT (id) DO NOTHING;
+-- INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (33, 'Dafydd', 'Pearson', 'dafydd@opensystemslab.io', true) ON CONFLICT (id) DO NOTHING;
+-- INSERT INTO public.users (id, first_name, last_name, email, is_admin) VALUES (36, 'Benjamin', 'Williams', 'benjamin@opensystemslab.io', true) ON CONFLICT (id) DO NOTHING;
+-- SELECT setval('users_id_seq', max(id)) FROM users;
+-- SELECT setval('teams_id_seq', max(id)) FROM teams;
```
jessicamcinchak, line 9: If we synced production `users`, we would no longer maintain or run Hasura seeds

scripts/seed-database/upsert-production-flows.js
```diff
@@ -57,18 +58,20 @@ const LOCAL_GRAPHQL_ADMIN_SECRET = process.env.HASURA_GRAPHQL_ADMIN_SECRET;
           theme
           notify_personalisation
         }
+        users {
+          id
+          first_name
+          last_name
+          email
+          is_admin
+        }
```
jessicamcinchak, line 67: `users` can only be queried by Hasura's admin role, whereas other tables that we've synced so far can be queried by the public role.

Adding authentication to the `productionGraphqlClient` defined above feels less than ideal, because this file has a number of mutations in it that would accidently get executed by the wrong client in testing, etc.

gunar:
I'm wondering what it would take for pizzas and dev envs to delegate the **whole** login process to planx.dev. I reckon we can't have wildcards in callback URLs but I guess it should be possible to write some glue code that runs on production and forwards the secrets to the required environment...

image


That's probably a lot of work though.

I'm trying to think what the consequences of a sync would be. Besides the work involved in juggling database credentials and making sure things are locked down, if we sync production users into staging then we'd lose the ability of having staging-only users. Having staging-only users is something that was a requirement in the past — is that not the case anymore?

We could also solve this with... process :) Writing down a process where we make sure every time we need to add a new user, we add them to staging and to the migration file at the same time.

Just some thoughts.

jessicamcinchak:
@gunar I don't know of any current requirements for staging-only users. I understand this could be solved with process (eg the SQL snippet to support user testing on pizzas _now_), but I don't think it'd be appropriate to start storing partner users details in our Hasura Seed script for sync purposes (feels not cool from a personal privacy / open repo standpoint, right?)

gunar:
Cool, no need for staging-only users :white_check_mark:

One possible solution would be write a `users`-sync'ing script that is separate from the current `upsert-production-flows.js` and that runs **exclusively** on `staging` and `pizzas`. This allows a few things:
1. Staging will have production users so UAT can happen
2. Pizzas will have production users so UAT can happen
3. Development environments can continue to work "offline" and without any admin permissions on the production database
4. Production is generally unaffected

jessicamcinchak:
Closing this PR, adding to dev call agenda to discuss next steps (but latest comment/proposal sounds work-able!)

PR #1295, opened 2022-11-28
test(e2e): add pay component test using playwright

gunar:
- adds End-to-End tests for the Pay component using Playwright
- the test running on the `preview` URL be redirected to the online yet sandbox `GOV.UK` API

### DIFF
--- a/e2e/playwright/assets/payFlowPublishedFlow.json
+++ b/e2e/playwright/assets/payFlowPublishedFlow.json
@@ -1,0 +1,47 @@
+{
+  "data": {
+    "_root": {
+      "edges": [
+        "a1ED9farLS",
+        "KzuRf6vqc1",
+        "NNdOGmxgfG",
+        "Dszu2W0S8t",
+        "D5fp40TTUD"
+      ]
+    },
+    "D5fp40TTUD": {
+      "data": {
+        "heading": "Application sent",
+        "moreInfo": "<h2>You will be contacted</h2>\n        <ul>\n        <li>if there is anything missing from the information you have provided so far</li>\n        <li>if any additional information is required</li>\n        <li>to arrange a site visit, if required</li>\n        <li>to inform you whether a certificate has been granted or not</li>\n        </ul>",
+        "contactInfo": "You can contact us at **planning@lambeth.gov.uk**<br/>What did you think of this service? Please give us your feedback using the link in the footer below.",
+        "description": "A payment receipt has been emailed to you. You will also receive an email to confirm when your application has been received."
+      },
+      "type": 725
+    },
+    "KzuRf6vqc1": {
+      "data": {
+        "output": "pay.test",
+        "formula": "10"
+      },
+      "type": 700
+    },
+    "NNdOGmxgfG": {
+      "data": {
+        "fn": "pay.test",
+        "color": "#EFEFEF",
+        "title": "Pay for your application",
+        "description": "<p>The planning fee covers the cost of processing your application.         Find out more about how planning fees are calculated          <a href=\"https://www.gov.uk/guidance/fees-for-planning-applications\" target=\"_self\">here</a>.</p>"
+      },
+      "type": 400
+    },
+    "a1ED9farLS": {
+      "data": {
+        "title": "Pay test"
+      },
+      "type": 110
+    }
+  },
+  "created_at": "2022-11-30T10:42:02.471335+00:00",
+  "publisher_id": 1,
+  "summary": "first publish"
+}
--- a/e2e/playwright/pay.spec.ts
+++ b/e2e/playwright/pay.spec.ts
@@ -1,0 +1,264 @@
+import { test, expect, type Page } from "@playwright/test";
+import { gqlAdmin, insertTestUser } from "./utils";
+
+import publishedFlow from "./assets/payFlowPublishedFlow.json";
+
+const URL = "http://localhost:3000";
+const TEAM_SLUG = "buckinghamshire";
+const FLOW_SLUG = "pay-test";
+
+// Test card numbers to be used in gov.uk sandbox environment
+// reference: https://docs.payments.service.gov.uk/testing_govuk_pay/#if-you-39-re-using-a-test-39-sandbox-39-account
+const SUCCESS_CARD_NUMBER = "4444333322221111";
+const INVALID_CARD_NUMBER = "4000000000000002";
+
+test.describe("Payment flow", async () => {
+  const USER_EMAIL = "test@test.com";
+
+  let userId = 0;
+  let teamId = 0;
+  let flowId = "";
+  let publishedFlowId = "";
+
+  test.beforeEach(async () => {
+    userId = await insertTestUser(USER_EMAIL);
+    teamId = await createTeam(TEAM_SLUG);
+    flowId = await createFlow({ teamId, flowSlug: FLOW_SLUG });
+    publishedFlowId = await insertPublishedFlow({ flowId });
+  });
+
+  test.afterEach(async () => {
+    const analyticsIds = await getAnalyticsByFlowId(flowId);
+    await deleteTestData({
+      teamId,
+      flowId,
+      publishedFlowId,
+      userId,
+      analyticsIds,
+    });
+  });
+
+  test("Should pay within GOV.UK Pay and reach the Confirmation page", async ({
+    page,
+  }) => {
+    await navigateToPayComponent({ page, mode: "preview" });
+
+    await page.getByText("Pay using GOV.UK Pay").click();
+
+    await fillGovUkCardDetails(page)(SUCCESS_CARD_NUMBER);
+    await page.locator("#confirm").click();
+    const paymentRef = await awaitForPaymentResponse(page);
+
+    expect(paymentRef).toBeTruthy();
+    expect(await page.getByText(paymentRef!).textContent()).toBeTruthy();
+  });
+
+  test("Should retry and succeed a failed GOV.UK payment", async ({ page }) => {
+    await navigateToPayComponent({ page, mode: "preview" });
+
+    await page.getByText("Pay using GOV.UK Pay").click();
+
+    await fillGovUkCardDetails(page)(INVALID_CARD_NUMBER);
+    await page.locator("#return-url").click();
+    await page.getByText("Retry payment").click();
+    await fillGovUkCardDetails(page)(SUCCESS_CARD_NUMBER);
+    await page.locator("#confirm").click();
+    const paymentRef = await awaitForPaymentResponse(page);
+
+    expect(paymentRef).toBeTruthy();
+    expect(await page.getByText(paymentRef!).textContent()).toBeTruthy();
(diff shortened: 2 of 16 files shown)


e2e/playwright/pay.spec.ts
```diff
@@ -0,0 +1,359 @@
...
+          affected_rows
+        }
+        delete_teams(where: {}) {
+          affected_rows
+        }
+      }
+    `);
+  });
+
+  test.skip("Should pay and send the application", async ({ page }) => {
```
gunar, line 117: The `unpublished` flow was skipped from the tests. In order to test it correctly we'll have to add more data to the tests:

* insert every flow that LDC uses as a portal
* insert the operation, team, and published_flow for every inserted flow

Let me know what you guys think.

e2e/playwright/pay.spec.ts
```diff
@@ -0,0 +1,359 @@
...
+    page,
+  }) => {
+    await navigateToPayComponent({ page, mode: "preview" });
+
+    await page.getByText("Pay using GOV.UK Pay").click();
+
+    // GOV.UK Pay
+    await page
+      .getByRole("textbox", { name: "Card number" })
+      .fill("4444333322221111");
```
jessicamcinchak, line 139: Could we store the mock card numbers as variables and link to [the docs](https://docs.payments.service.gov.uk/testing_govuk_pay/#mock-card-numbers-and-email-addresses) please? 

Can we also have one number for each "Testing outcome" as defined by Gov UK (eg "success", "failure because card declined" etc)?

e2e/playwright/pay.spec.ts
```diff
@@ -0,0 +1,359 @@
...
+
+    await page.locator("button#submit-card-details").click();
+    await page.getByRole("button", { name: "Confirm payment" }).click();
+
+    await page.getByLabel("Email address").fill(TEST_EMAIL);
+    await page.getByTestId("continue-button").click();
+
+    await page.waitForResponse(
+      (resp) => resp.url().includes("create-send-events") && resp.ok()
+    );
```
jessicamcinchak, line 167: The first request that planx makes when returning from Gov Pay is still _within_ the Pay component to double check the status of the payment, I'm a bit confused why this test is jumping past that to check if the next component has been reached? 

Can we check for this request instead please? https://github.com/theopensystemslab/planx-new/blob/main/api.planx.uk/server.ts#L340-L375

e2e/playwright/pay.spec.ts
```diff
@@ -0,0 +1,359 @@
+import { test, expect, type Page } from "@playwright/test";
+import { readFileSync } from "fs";
+import path from "path";
+import { gqlAdmin, insertTeam, insertTestUser } from "./utils";
+
+import team from "./assets/buckinhamshire.json";
+import ldcflow from "./assets/ldcFlow.json";
+import publishedFlows from "./assets/publishedFlowLDC.json";
+import operation from "./assets/operationLDC.json";
```
jessicamcinchak, line 9: If this test is intended to test the Gov Pay integration, I'm not sure I understand why it's important for flow content to reflect a full LDC flow? Why not just a simple published flow that has a SetValue (in order to set amount = 1), Pay (reads from "amount" passport field), then Confirmation (because confirmation page should display pay reference number on success) and **focus on testing different Pay _states_** (eg success, failure & retry, etc)? 

This seems to introduce a fair bit of noise to replicate production content (if we do want to use a LDC flow, can we point to production or staging rather than storing static JSON here?) and as a side effect has to deal with complex non-Pay functionalities like `uploadFileToInput`, `navigateToPayComponent`, etc which I imagine make it more difficult to determine when this test fails for a Pay-specific reason?

e2e/playwright/pay.spec.ts
```diff
@@ -0,0 +1,359 @@
...
+      .fill("123");
+
+    await page.getByLabel("Town or city").fill("Test");
+    await page.getByLabel("Postcode").fill("HP111BB");
+    await page
+      .getByLabel("Email")
+      .fill("simulate-delivered@notifications.service.gov.uk");
+
+    // GOV.UK Pay submit button is actionable but if there is no delay here the click is ignored
+    await page.waitForTimeout(1000);
```
jessicamcinchak, line 157: :+1:

e2e/playwright/pay.spec.ts
```diff
@@ -0,0 +1,359 @@
...
+
+    await page.locator("button#submit-card-details").click();
+    await page.getByRole("button", { name: "Confirm payment" }).click();
```
jessicamcinchak, line 160: It would also be helpful to have a test that captures the journey of clicking "Cancel payment" within Gov UK Pay - which should trigger a state = "failed" & reset on our end (reset meaning you land back on Pay component and Gov Pay inputs are empty when you proceed to pay again).

e2e/playwright/pay.spec.ts
jessicamcinchak: And, again, since this test is about testing Pay, I would think it shouldn't have a Send component at all!

jessicamcinchak (commented):
thanks for getting these started, really interesting to see how e2e works with Playwright too! 

left some comments (mostly about how this surprisingly feels like it's testing a lot _besides_ Pay to me) and would be happy to chat about more during dev call, or just shout before then if anything isn't clear

e2e/playwright/pay.spec.ts
jessicamcinchak: I think only testing `/preview` makes sense for now, as our goal is to test applicant Pay journeys & applicants should never be on `/unpublished` routes !

gunar:
Superb feedback @jessicamcinchak. 100% agree. Will update.

e2e/playwright/pay.spec.ts
DafyddLlyr: Also more generally, I'd be interested to see what an alternative that uses assets downloaded from the DB looks like. This would allow us to test journeys still worked if the structure or data changed.

1. Flows / teams are generated on prod
2. Data copied onto test environment's db
3. Download them as assets (or directly query...?)
4. Run tests against "real" data

I don't think this needs to be achieved in this PR but I would be curious if this is a model we could use. For example, if we changed for the `teams` table was structured we wouldn't need to manually update our assets as part of that change.

e2e/playwright/pay.spec.ts
```diff
@@ -0,0 +1,359 @@
...
+  await page.getByText("None of the above apply to me").click();
+  await page.getByTestId("continue-button").click();
+
+  await page
+    .getByText("The information contained in this application is truthful")
+    .click();
+  await page.getByTestId("continue-button").click();
+
+  await page.getByText("South Buckinghamshire").click();
+  await page.getByTestId("continue-button").click();
```
DafyddLlyr, line 358: I agree with @jessicamcinchak here that in this instance we'd get more value from focusing on Pay and the various states which I think was the original intention.

What you have here though is great to see, and would likely be used elsewhere if we adopt Playwright, so it's still useful work which I think can / should be repurposed 👌

DafyddLlyr (commented):
I think the main points have been covered - really helpful feedback thanks @jessicamcinchak !

I'm looking forward to catching up later at the dev call and seeing what you think about Playwright 👍

e2e/playwright/pay.spec.ts
gunar: I'm jotting down the code (commit) for posterity so we can utilise it as a starting point at some point in the future.

gunar:
I believe I have now addressed all feedback: :+1:

gunar:
[End to end tests for 'Pay' component](https://trello.com/c/e4TOm6GJ/1443-end-to-end-tests-for-pay-component)

jessicamcinchak (commented):
This reads much more clearly now and feels more focused on the ultimate goal of testing Pay, thanks for addressing so much of my initial feedback! Still a few outstanding questions, but seems close

hasura.planx.uk/migrations/1669979245259_alter_table_public_analytics_logs_alter_column_analytics_id/up.sql
```diff
@@ -0,0 +1,4 @@
+-- There was an inconsistency with `analytics` and `analytics_logs`.
+-- The `analytics_id ` column in `analytics_logs` table had an `int` value, but the `analytics`'s id is a `bigint`.
+-- This causes a few issues when we are querying on GraphQL using variables.
+ALTER TABLE "public"."analytics_logs" ALTER COLUMN "analytics_id" TYPE int8;
```
jessicamcinchak, line 4: should this be `TYPE bigint` rather than `int8` ? 

via https://www.postgresql.org/docs/9.1/datatype-numeric.html
> SQL only specifies the integer types integer (or int), smallint, and bigint. The type names int2, int4, and int8 are extensions, which are also used by some other SQL database systems.

also minor nit: i personally appreciate bug fixes as separate PRs if they are completely un-related to existing open PRs !

e2e/playwright/pay.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+      (resp) => resp.url().includes(`pay/${TEAM_SLUG}`) && resp.ok()
+    );
+    await fillGovUkCardDetails(page)(SUCCESS_CARD_NUMBER);
+    await page.locator("#confirm").click();
+    await resumeApplication(page);
+    await page.waitForResponse(
+      (resp) => resp.url().includes("create-send-events") && resp.ok()
+    );
+
+    expect(await page.getByText("Application sent").textContent()).toBeTruthy();
```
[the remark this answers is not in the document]
gunar: done

e2e/playwright/assets/payFlowPublishedFlow.json
```diff
@@ -0,0 +1,57 @@
...
+    "Dszu2W0S8t": {
+      "data": {
+        "title": "Send",
+        "destinations": [
+          "bops",
+          "uniform"
+        ]
+      },
+      "type": 650
+    },
```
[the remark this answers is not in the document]
gunar: done

jessicamcinchak (approved):
Thanks for most recent changes :+1:  I think these test flows read a lot more focused now, and if they fail we'll know it's because of Pay! 

My final nit is that I'd still love to see one third test case which is: `"Should retry and succeed a cancelled GOV.UK payment"`. We know cancelled payments are showing up regularly in production, and it'd be a very similar flow to your invalid card test - but rather than entering card details and clicking "confirm" the first time you connect to Gov Pay, you can just immediately click "cancel" then expect tto have a payment ref with failed status and to see the "Retry" button within Planx. 

But nonetheless these tests feel like a very solid intro to playwright :tada: I'm happy for them to merge whenever (I imagine you might want to coordinate this a bit with Ben though since his playwright tests are branching from here?)

gunar:
Thanks for the review! I've coordinated with @builditben, and I'll be merging this now so he can rebase his work. Also, I'll create a new PR to address the two remaining comments.

PR #1225, opened 2022-10-21
feat: Increase map size in DrawBoundary (alt)

DafyddLlyr:
Branched from https://github.com/theopensystemslab/planx-new/pull/1223 to try another style suggested by @Alastairparvin at Show & Tell

image

PR #1299, opened 2022-11-30
feat: Increase DrawBoundary map size

DafyddLlyr:
Follow up to implement the changes made in #1225 

Feedback from UR - 

> Tested with 6 people in R.45. I don't think there are lots of feedback on the enlargement of the map. The users didn't really know what it used to be like before the changes and they are still struggling a bit when the map doesn't show the whole of their property and they have to zoom out to see the whole of their site and then they are slightly nervous when the snap points disappear. But map enlarged map appears to be clear to users and doesn't seem to have an issue with the enlarged view.


DafyddLlyr:
[Make draw boundary map bigger / full screen option](https://trello.com/c/c06s6rPl/1918-make-draw-boundary-map-bigger-full-screen-option)

jessicamcinchak (approved):

PR #1300, opened 2022-12-01
feat: switch to crosshair as drawing cursor

jessicamcinchak:
In conclusion of Lam's user-testing feedback: 
> Turn red dot into a crosshair: Tested with 6 people in R.45. All users understood the cross-hair icon as a drawing tool. Don’t seem to mind the over icon also overlapping on it. (Lam: But we should get rid of it when we can)


DafyddLlyr (approved):
✅

PR #1223, opened 2022-10-20
feat: Increase map size in DrawBoundary

DafyddLlyr:
## What does this PR do?
 - Above the "md" breakpoint, increase the map size in the DrawBoundary component
 - Below this, keep things as is
 - The new width and height of 80vw x 80vh is chose quite arbitrarily - it seems to allow the map + instructions to appear on the screen at the same time. I expect this could be changed through feedback and UAT.

### Also...
- Tidy up styling, continue to migrate touched components from JSS to Emotion

image

https://user-images.githubusercontent.com/20502206/197004072-10393eb3-ef06-4339-9e5d-071243f92c98.mov

PR #1304, opened 2022-12-02
fix: Adjust space on 'Upload location plan'

DafyddLlyr:
This is a fix for styling issue I missed when resolving merge conflicts between increasing map size and converting buttons to links.

|Before|After|
|---|---|
|image|image|


DafyddLlyr:
@builditben thanks for the review!

The map size change is intentional and was introduced here - https://github.com/theopensystemslab/planx-new/pull/1225

I will say that the design does look a little unusual to me - my original inclination was to expand from centre like this - https://github.com/theopensystemslab/planx-new/pull/1223

The size increase and expanding to the right only has been signed off and user tested though 👍

PR #1305, opened 2022-12-02
Upgrade tiptap packages. Move hard break to common extensions

peterszerzo:
Potential fixes for the errors described here: https://opensystemslab.slack.com/archives/C01E3AC0C03/p1669975760312039?thread_ts=1669972221.878829&cid=C01E3AC0C03.

I have not been able to reproduce any shift+enter issues on the storybook, we might have to keep digging here.


DafyddLlyr (approved):
Thanks very much @peterszerzo, great stuff!

I can't replicate the issue on the Pizza, but can consistently on Staging and Prod 👍

PR #1307, opened 2022-12-05
refactor: Simplify MUI theme generation

DafyddLlyr:
## What is the problem?
Currently, due to `createPalette` being called too early in the theme generation process, some styles are not being correctly derived from a team's primary colour. Please see example below where the dark version of PlanX's default colour is being applied on hover, instead of a dark version of the team's primary colour.

|Staging|Pizza|
|---|---|
|chrome-capture-2022-11-5|chrome-capture-2022-11-5 (1)|

## What does this PR do?
 - Fixes bug by not calling `createPalette()` too early which generates derived colours (e.g. `theme.palette.primary.dark` which is used on hover here). Instead this will only happen when `generateTeamTheme()` is called, by which time we always have a primary (or default) colour set.
 - Simplify generation of `defaultTheme` and `teamTheme`

### DIFF
--- a/editor.planx.uk/src/theme.ts
+++ b/editor.planx.uk/src/theme.ts
@@ -1,14 +1,41 @@
 import {
   createTheme,
   responsiveFontSizes,
+  Theme,
   ThemeOptions,
 } from "@mui/material/styles";
 // eslint-disable-next-line no-restricted-imports
-import createPalette from "@mui/material/styles/createPalette";
-
-import { TeamTheme } from "./types";
+import createPalette, {
+  PaletteOptions,
+} from "@mui/material/styles/createPalette";
 
 const GOVUK_YELLOW = "#FFDD00";
+
+const DEFAULT_PRIMARY_COLOR = "#000661";
+
+const DEFAULT_PALETTE: Partial<PaletteOptions> = {
+  primary: {
+    main: DEFAULT_PRIMARY_COLOR,
+    contrastText: "#fff",
+  },
+  background: {
+    default: "#fff",
+    paper: "#f2f2f2",
+  },
+  secondary: {
+    main: "#EFEFEF",
+  },
+  text: {
+    secondary: "rgba(0,0,0,0.6)",
+  },
+  action: {
+    selected: "#F8F8F8",
+    focus: GOVUK_YELLOW,
+  },
+  error: {
+    main: "#E91B0C",
+  },
+};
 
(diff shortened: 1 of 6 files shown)


editor.planx.uk/src/theme.ts
```diff
@@ -45,33 +72,11 @@ export const linkStyle = (primaryColor?: string) => ({
...
 });
 
-/**
- * Get Global theme options
- * The global theme is used in editor, and as the base theme in Preview/Unpublished which can be
- * merged with Team specific options
- */
-export const getGlobalThemeOptions = (): ThemeOptions => {
+const getThemeOptions = (primaryColor: string): ThemeOptions => {
   const palette = createPalette({
```
DafyddLlyr, line 76: Here's the bug - calling this without the team's primary colour generates secondary colours (e.g. `theme.palette.primary.dark`) which are then never overwritten.

editor.planx.uk/src/theme.ts
```diff
@@ -45,33 +72,11 @@ export const linkStyle = (primaryColor?: string) => ({
...
+    ...DEFAULT_PALETTE,
     primary: {
```
gunar, line 78: This is a shallow merge and unless I'm mistaken it's overwriting `primary.contrastText` into `undefined`.

editor.planx.uk/src/theme.ts
```diff
@@ -219,33 +225,16 @@ export const getGlobalThemeOptions = (): ThemeOptions => {
...
+const generateTeamTheme = (
+  primaryColor: string = DEFAULT_PRIMARY_COLOR
+): Theme => {
+  const themeOptions = getThemeOptions(primaryColor);
+  const theme = responsiveFontSizes(createTheme(themeOptions));
+  return theme;
 };
 
-const globalTheme = createTheme(getGlobalThemeOptions());
+// A static MUI theme based on PlanX's default palette
```
gunar, line 237: :+1:

gunar (approved):

editor.planx.uk/src/theme.ts
DafyddLlyr: Good catch thank you - you're correct. Fixed with a `deepmerge` 👌

gunar: Good :+1:

jessicamcinchak (approved):

PR #1311, opened 2022-12-06
doc: Integration design proposal

ghost:
This is a write-up of some of the ideas we have been discussing about integrations.

There are still a few gaps and this is definitely open for comments and revision. 

Do feel free to add changes directly to this branch (fix-forward only - no rebasing should be needed).

### DIFF
--- a/doc/architecture/decisions/0004-integration-patterns.md
+++ b/doc/architecture/decisions/0004-integration-patterns.md
@@ -36,7 +36,7 @@ Where downstream consumers cannot support webhooks directly, PlanX can host 3rd 
 
 ### Overview
 
-The flow for each of these patterns can be described in the following sequence diagram where steps 1 to 3 show the **webhook** pattern and steps 4 to 9 show the **webhook** pattern:
+The flow for each of these patterns can be described in the following sequence diagram where steps 1 to 3 show the **webhook** pattern and steps 4 to 9 show the **connector** pattern:
 
 ```mermaid
 sequenceDiagram


doc/architecture/decisions/0004-integration-patterns.md
```diff
@@ -0,0 +1,237 @@
...
+
+- Status: Draft
+- Date: December 2022
+
+## Context
+
+PlanX allows service designers to create, edit and publish digital services to the public. Where these services have a back-office processing stage, session data must be made available to the appropriate **downstream consumers** (3rd party platforms) after the completion of each **session** (a user's journey through a service). An **integration** is any pattern that allows PlanX to securely pass session data to these downstream consumers when configured to do so by the "Send" component within a service.
+
+The PlanX platform is currently scaling the number of teams using the platform, some of whom will require integrations to downstream consumers which are not currently supported. Open Digital Planning has also announced a fund for integrations with PlanX/BOPs/Digital Land. Both of these factors mean that a clear and robust set of integration patterns is required in the near future.
+
```
jessicamcinchak, line 11: The definition of **integration** here is an interesting one for me: in this document we're (rightly) focusing on essentially one-directional _submission_ integrations, but I think existing partners (eg DLUHC) also want to say that "Planx integrates with Digital Land" (which is currently an integration in the other direction, but perhaps in the future is bi-directional - eg we feed back info to digital land if an applicant contests a planning constraint). 

I'm wondering if we can be a more explicit about some of this context up front here, and then shift into the subset of integrations that support the submission of Planx applications. I'll think more on ways to write that.

doc/architecture/decisions/0004-integration-patterns.md
```diff
@@ -0,0 +1,237 @@
...
+PlanX will notify downstream consumers of the availability of session data via **Webhooks**.
+
+Where downstream consumers cannot support webhooks directly, PlanX can host 3rd party **Connectors** that transform and send session data.
+
+### Overview
+
+The flow for each of these patterns can be described in the following sequence diagram where steps 1 to 3 show the **webhook** pattern and steps 4 to 9 show the **connector** pattern:
+
+```mermaid
```
jessicamcinchak, line 41: :mermaid: i haven't seen this used in markdown before, very handy !

DafyddLlyr (commented):
Thanks very much for this - great to have this written down as a starting point.

A few comments, looking forward to seeing the rest of the feedback and progressing this.

PR #1314, opened 2022-12-06
feat: add audit table to track reconciled sessions

jessicamcinchak:
this audit table will let us do two key things:
- track how many times a save & return session is successfully resumed (group by session_id & count)
- better troubleshoot applicant issues for cases like [this one recently reported in #help-issues](https://opendigitalplanning.slack.com/archives/C0241GWFG4B/p1669724872840919) 

future scope: 
- delete `response` values when an associated lowcal_session is marked as deleted_at. this will let us keep session ids & resumed timestamps for analytics, but not retain personal session data. associated ticket here https://trello.com/c/FlGxMbps/2078-data-retention-create-single-scheduled-task-to-purge-application-data-at-set-time-interval
- create a proper foreign key relationship between `lowcal_sessions.id` & `reconciliation_requests.session_id` ? (reconciliation_requests follows pattern of other existing audit tables like planning_constraints_requests, bops_ & uniform_applications which means no upfront foreign keys and session ids are stored as text for simplicity)

to test: 
- go through a save & return flow
- save & resume your session
- confirm that an entry has been created in table `reconciliation_requests`


jessicamcinchak:
[Add audit table for reconciliation requests](https://trello.com/c/ELBIsLmL/2222-add-audit-table-for-reconciliation-requests)

jessicamcinchak:
> I think it might be implicitly obvious but, to make it explicit, could we add something like a "reconcilliationSkipped" field to the table? That might make it a lot easier to understand the meaning of the log entry without having to parse the response object.

thanks @builditben ! this is a smart idea for having a bit more easily-filterable context, especially as we don't plan to keep the actual response object around longterm. currently, i'm only setting `reconciliation_skipped = true` when returning from Pay, as this will also make it easy to filter out those cases for analytics like i wondered above. is that also your definition of "reconcilation skipped" ? 

pizza rebuilding with updated database migration now :+1:

api.planx.uk/saveAndReturn/validateSession.ts
```diff
@@ -29,12 +30,14 @@ const validateSession = async (
...
-        return res.status(200).json({
+        const responseData = {
           message: "Payment process initiated, skipping reconciliation",
           alteredNodes: null,
           removedBreadcrumbs: null,
           reconciledSessionData: sessionData.data,
-        });
+        };
+        await createAuditEntry(sessionId, responseData);
+        return res.status(200).json(responseData);
```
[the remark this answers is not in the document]
jessicamcinchak: agree it feels better to over-audit here than not collect it! especially now with an easy field to filter on for a more accurate count of what the applicant would actually experience as "resuming" for analytics purposes

jessicamcinchak:
> Is reconciliation also skipped on [validateSession.ts#L124](https://github.com/theopensystemslab/planx-new/blob/39b94707f66862d8b9af64c4f131ad07ee480205/api.planx.uk/saveAndReturn/validateSession.ts#L124)?

In my opinion, not technically in that case! Rather, the reconciliation methods ran & found that there had been no content changes, therefore no user data adjustments - but these definitions are tricky. 

I understand it like this for now, but up for debate!:
- are they returning from Pay? (if yes, `reconciliation_skipped = true`)
- has published content changed? (`reconciliation_skipped = false` regardless of answer)
- are changed content nodes found in user's breadcrumbs? (`reconcilation_skipped = false` regardless of answer)

api.planx.uk/saveAndReturn/validateSession.ts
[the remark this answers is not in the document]
DafyddLlyr: Yep, this makes sense to me also - we might find ourselves removing it when we come to "Request to pay" but it's sensible to have it here now.

api.planx.uk/saveAndReturn/validateSession.ts
```diff
@@ -29,12 +30,14 @@ const validateSession = async (
...
       if (sessionData.data.govUkPayment?.state?.status === "created") {
-        return res.status(200).json({
+        const responseData = {
           message: "Payment process initiated, skipping reconciliation",
           alteredNodes: null,
           removedBreadcrumbs: null,
           reconciledSessionData: sessionData.data,
-        });
+        };
+        await createAuditEntry(sessionId, responseData, true);
```
DafyddLlyr, line 39: nit/thought: Certainly not required for this PR, but it might be nice / helpful to have an `AuditService` or similar creating BOPS / Uniform / Reconciliation records going forward.

DafyddLlyr: nit: Could we use an options object avoid a boolean param for `reconciliationSkipped` here? Looking at  `createAuditEntry(sessionId, responseData, true)` it's not inherently obvious what the final parameter means.

Alternatively, we could capture this variable and use it for control flow in this function.

```js
const isReconciliationSkipped = someLogic()
if (isReconciliationSkipped) {
  doStuff()
else {
  doOtherStuff()
}
await createAuditEntry(sessionId, responseData, isReconciliationSkipped);

**Review by DafyddLlyr — APPROVED** [review]
This is great! A few very minor nits/ideas - no show stoppers.

In terms of stats (e.g. avg. number of requests per session, time between lowcal_session creation and reconciliation_request creation) I think we are covering a whole lot here 👍

**Comment by DafyddLlyr** [comment]
I'm actually hitting issues also in testing sorry @jessicamcinchak - too keen with my green tick sorry...! ✅ 

Looking into it now but I'm getting a 500 error when calling `/validate-session` on a published flow.

**Comment by jessicamcinchak** [comment]
@DafyddLlyr & @builditben - sorry about that, this was a `Boolean` not `boolean` graphql type error on my side, rebuilding now for fresh testing!

File: api.planx.uk/saveAndReturn/validateSession.ts [hunk 3]
```diff
@@ -10,6 +10,7 @@ import {
 import { Breadcrumb, LowCalSession, Node } from "../types";
 
 const client = publicGraphQLClient;
+const adminClient = adminGraphQLClient;
```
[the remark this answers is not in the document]
jessicamcinchak: updated!

api.planx.uk/saveAndReturn/validateSession.ts
jessicamcinchak: rethinking this a bit: since this boolean isn't a super straightforward interpretation here anyways, I've edited the table schema again and made the third argument `message` - which will have the same benefit of letting us filter for analytics but capture more scenarios more clearly, and more easily scale to future reconciliation scenarios. also makes for for much less ambiguous final parameter!

DafyddLlyr: Great solution - much more scalable that adding multiple bool columns in future 👍

PR #1315, opened 2022-12-06
fix: Different approach on map size increase

DafyddLlyr:
Here's a fix for some of the issues encountered with using `vw` units for increasing the map size. 

Using `vw` caused issues when browsers zoomed in and out, and did not produce a consistent right hand margin.

https://user-images.githubusercontent.com/20502206/206158604-2ccaf0b7-b9fd-4df7-b325-1571958d8c9e.mov

### DIFF
--- a/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
@@ -32,7 +32,7 @@ interface MapContainerProps {
 const dynamicMapSizeStyle = (theme: Theme): Record<string, any> => {
   const mainContainerWidth = `${theme.breakpoints.values.md}px`;
   const mainContainerMargin = `((100vw - ${mainContainerWidth}) / 2)`;
-  const mapMarginRight = "100px";
+  const mapMarginRight = "150px";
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
```diff
@@ -25,19 +25,32 @@ interface MapContainerProps {
...
+
+  const style = {
+    [theme.breakpoints.up("md")]: {
+      height: "70vh",
+      width: `calc(${mainContainerMargin} + ${mainContainerWidth} - ${mapMarginRight})`,
+    },
+  };
+
+  return style;
+};
```
DafyddLlyr, line 45: This style could also be more terse, but I personally think the more explicit is more understandable.

```js
width: `calc(${theme.breakpoints.values.md}px + ((100vw - ${theme.breakpoints.values.md}px) / 2) - 100px`,
```

jessicamcinchak (approved):
Very clear, thanks for cleaning up!

DafyddLlyr:
Tested and approved by @Alastairparvin who was having issues with the previous implementation ✅

PR #1316, opened 2022-12-06
fix: Confirm component styling

DafyddLlyr:
## What does this PR do?
 - Consistently styles the `Confirm` component, specifically `SuggestionDrawer`
   - Use PlanX checkboxes, not plain MUI checkboxes
   - Use `MoreInfo` side panel, not plain MUI drawer
   - Update component to Emotion CSS engine
- Fix bug introduced in #1220 which throws an error in the console when opening the `MoreInfo` component

## Also...
This form never submits (and never has...?) 👀 Once #1284 is tested and merged I'll come back and wire this up

|Before|After|
|---|---|
|image|image|

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/Preview/MoreInfo.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Preview/MoreInfo.tsx
@@ -14,33 +14,33 @@ interface StyledDrawerProps extends DrawerProps {
   drawerWidth: number;
 }
 
-const Root = styled(Drawer)<StyledDrawerProps>(
-  ({ theme, drawerWidth, open }) => ({
+const Root = styled(Drawer, {
+  shouldForwardProp: (prop) => prop !== "drawerWidth",
+})<StyledDrawerProps>(({ theme, drawerWidth, open }) => ({
+  width: drawerWidth,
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/@planx/components/shared/Preview/MoreInfo.tsx
```diff
@@ -14,33 +14,33 @@ interface StyledDrawerProps extends DrawerProps {
   drawerWidth: number;
 }
 
-const Root = styled(Drawer)<StyledDrawerProps>(
-  ({ theme, drawerWidth, open }) => ({
+const Root = styled(Drawer, {
+  shouldForwardProp: (prop) => prop !== "drawerWidth",
```
DafyddLlyr, line 18: Passing the prop `drawerWidth` into the styled `Root` component (line 66) automatically passes it down to the MUI Drawer component, which does not recognise the prop and throws an error in the console.

`shouldForwardProp` allows us to specify that we should not forward `drawerWidth` along to the underlying component.

Docs: https://mui.com/system/styled/#styled-component-options-styles-component

jessicamcinchak (approved):
thanks for spotting this one !

PR #1320, opened 2022-12-08
fix(public): add missing doncaster domain config

gunar:
Fixes bug reported here:

https://opensystemslab.slack.com/archives/C5Q59R3HB/p1670506544605919?thread_ts=1670500507.687649&cid=C5Q59R3HB

### DIFF
--- a/editor.planx.uk/src/routes/utils.ts
+++ b/editor.planx.uk/src/routes/utils.ts
@@ -55,6 +55,7 @@ const PREVIEW_ONLY_DOMAINS = [
   "planningservices.buckinghamshire.gov.uk",
   "planningservices.lambeth.gov.uk",
   "planningservices.southwark.gov.uk",
+  "planningservices.doncaster.gov.uk",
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/routes/utils.ts
```diff
@@ -55,6 +55,7 @@ const PREVIEW_ONLY_DOMAINS = [
   "planningservices.buckinghamshire.gov.uk",
   "planningservices.lambeth.gov.uk",
   "planningservices.southwark.gov.uk",
+  "planningservices.doncaster.gov.uk",
```
DafyddLlyr, line 58: nit: Could we also drop "1036.planx.pizza" from this list now?

DafyddLlyr (approved):
Thanks for picking up and identifying so quickly 👍 

I'll add a step to the `README.md` on https://github.com/theopensystemslab/planx-team-onboarding to capture this step.

PR #1323, opened 2022-12-09
test: Save and return e2e tests

ghost:
This patch adds a simple save and return suite and some utilities for writing further e2e tests.

### DIFF
--- a/.github/workflows/end-to-end-tests.yml
+++ b/.github/workflows/end-to-end-tests.yml
@@ -47,9 +47,9 @@ jobs:
         working-directory: e2e
       - name: Install Playwright Browsers
         run: pnpm playwright install --with-deps
+        working-directory: e2e
       - name: Start test containers
         run: ./scripts/start-containers-for-tests.sh
-        working-directory: e2e
       - name: Playwright Tests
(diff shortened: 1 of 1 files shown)


DafyddLlyr (approved):
This is great! 

Not taken a super close look at the two previous Playwright PRs but it was good to do a deep dive here and get my head around it - this approach looks solid to me. Tests are clear and concise, and the modular approach (`returnToSession`, `saveSession`) works well - having these building blocks is going to be helpful going forward.

Added a small commit to fix the CI issues which seems to have done the job 👌 

One hanging question about `scripts/start-containers-for-tests.sh` - that can get addressed here if you'd like or in another PR (iterating on GitHub actions is generally a lot easier once it's onto `main` as you can trigger it as and when you want).

e2e/playwright/context.ts
```diff
@@ -188,25 +188,16 @@ export interface InitialContext {
   };
 }
 
-export interface Context {
+type Context = InitialContext & {
```
DafyddLlyr, line 191: 👌

DafyddLlyr (approved):

PR #1325, opened 2022-12-12
feat: Data sanitation cron job

DafyddLlyr:
## What does this PR do?

* Sets up a cron job on Hasura which runs nightly and hits the `/webhooks/hasura/sanitise-application-data` endpoint
* This endpoint iterates over a series of operations identified [on the Trello ticket](https://trello.com/c/FlGxMbps/2078-data-retention-create-single-scheduled-task-to-purge-application-data-at-set-time-interval)
* These operations sanitise (redact) sensitive data from the identified tables
  * This preserves stats (e.g. number of `uniform_applications` since XXX) and some debugging ability in place which dropping records would not allow
  * See note of foreign keys - this is worth addressing I think
  * Hasura events logs are of little value, and are just dropped

## To test
* Create some records on the Pizza, and modify their `created_at` values to be longer than 6 months ago
* Manually call the `/webhooks/hasura/sanitise-application-data` endpoint, using the correct auth headers


hasura.planx.uk/migrations/1670838102095_add_sanitised_at_columns/up.sql
```diff
@@ -0,0 +1,14 @@
...
+COMMENT ON COLUMN "public"."session_backups"."sanitised_at" IS E'Records if a record has been sanitised of personal data ';
+
+alter table "public"."uniform_applications" add column "sanitised_at" timestamptz null;
+COMMENT ON COLUMN "public"."uniform_applications"."sanitised_at" IS E'Records if a record has been sanitised of personal data ';
+
+alter table "public"."bops_applications" add column "sanitised_at" timestamptz null;
+COMMENT ON COLUMN "public"."bops_applications"."sanitised_at" IS E'Records if a record has been sanitised of personal data ';
+
+alter table "public"."reconciliation_requests" add column "sanitised_at" timestamptz null;
+COMMENT ON COLUMN "public"."reconciliation_requests"."sanitised_at" IS E'Records if a record has been sanitised of personal data ';
```
DafyddLlyr, line 14: I'm not massively keen on this approach - I think we had a foreign key set up on these audit tables from `lowcal_sessions.id` we could avoid having `sanitised_at` columns across audit tables. 

This would have the added benefit of deleting these audit records 6 months from when the session has expired / been submitted, as opposed to 6 months from when the audit record was created which is not the same thing.

This PR is already bigger than I'd like though so I'd be happy to take this on as a separate task if we think it's worthwhile to tackle.

DafyddLlyr:
[Data retention - create single scheduled task to purge application data at set time interval](https://trello.com/c/FlGxMbps/2078-data-retention-create-single-scheduled-task-to-purge-application-data-at-set-time-interval)

hasura.planx.uk/migrations/1670838102095_add_sanitised_at_columns/up.sql
```diff
@@ -0,0 +1,14 @@
...
+COMMENT ON COLUMN "public"."session_backups"."sanitised_at" IS E'Tracks if personal data has been sanitised from record';
+
+alter table "public"."uniform_applications" add column "sanitised_at" timestamptz null;
+COMMENT ON COLUMN "public"."uniform_applications"."sanitised_at" IS E'Tracks if personal data has been sanitised from record';
+
+alter table "public"."bops_applications" add column "sanitised_at" timestamptz null;
+COMMENT ON COLUMN "public"."bops_applications"."sanitised_at" IS E'Tracks if personal data has been sanitised from record';
+
+alter table "public"."reconciliation_requests" add column "sanitised_at" timestamptz null;
+COMMENT ON COLUMN "public"."reconciliation_requests"."sanitised_at" IS E'Tracks if personal data has been sanitised from record';
```
DafyddLlyr, line 14: I'm not massively keen on this approach - I think we had a foreign key set up on these audit tables from lowcal_sessions.id we could avoid having sanitised_at columns across audit tables.

This would have the added benefit of deleting these audit records 6 months from when the session has expired / been submitted, as opposed to 6 months from when the audit record was created which is not the same thing.

This PR is already bigger than I'd like though so I'd be happy to take this on as a separate task if we think it's worthwhile to tackle.

jessicamcinchak: I don't think there's any case against foreign keys! Agree it would help simplify this and mean we don't require adding a `sanitised_at` to each individual audit table.

api.planx.uk/webhooks/sanitiseApplicationData/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  try {
+    results = await Promise.all(operations.map(operationHandler));
+  } catch (error) {
+    // Unhandled error, flag with Airbrake
+    return next({
+      error,
+      message: `Failed to sanitise application data. ${(error as Error).message}`,
+    });
+  }
+  await postToSlack(results);
```
jessicamcinchak, line 27: Is there value in posting the successes to Slack in this case, or do we really just care about visiblity of failed CRON jobs? I totally understand the little confidence boost that comes from an affirmative notification and am fine with both at first, but imagine it may start to feel noisy quick.

api.planx.uk/webhooks/sanitiseApplicationData/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  
+  const operationFailed = results.find(result => result.status === "failure");
+  if (operationFailed) res.status(500)
+  return res.json(results);
+};
+
+const postToSlack = async (results: OperationResult[]) => {
+  const slack = SlackNotify(process.env.SLACK_WEBHOOK_URL!);
```
jessicamcinchak, line 35: :+1: nice we can re-use same webhook here as #planx-notifications and not have to manage extra tokens

api.planx.uk/webhooks/sanitiseApplicationData/operations.ts
```diff
@@ -0,0 +1,214 @@
...
+      errorMessage: (error as Error).message,
+    }
+  };
+
+  return operationResult;
+}
+
+export const sanitiseLowcalSessions: Operation = async () => {
+  const mutation = gql`
+    mutation SanitiseLowcalSessions($retentionPeriod: timestamptz) {
```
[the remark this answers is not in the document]
jessicamcinchak: I'm also interested in this question! Related to prior comment about sanitising after a week and hard-deleting after 6 months. 

I think we could communicate this policy as:
- We have anonymous analytics for a service over _all time_
- We have fine-grained save&return/reconciliation/submission analytics _per session_ for the _last six months_
  - Eg anonymous analytics could tell you an app was submitted/reached Confirmation page, but fine-grained audit tables could tell you _where_ it was submitted to. I think the latter is perhaps still more a function of troubleshooting anyways rather than full service stats since Send destinations don't change frequently. 

Probably one for Alastair & partners to help confirm!

api.planx.uk/webhooks/sanitiseApplicationData/operations.ts
```diff
@@ -0,0 +1,214 @@
...
+  );
+  return result;
+};
+
+export const sanitiseSessionBackups: Operation = async () => {
+  const mutation = gql`
+    mutation SanitiseSessionBackups($retentionPeriod: timestamptz) {
+      update_session_backups(
+        _set: {
+          flow_data: null
```
jessicamcinchak, line 90: nit: I wonder if rather than `null` or `{}` we could do populate flow_data with something like `_SANITISED_` here? Similar to how you can find `_REMOVED_` in `users` table currently? But I recognize that may make less sense/not be possible with a JSONB data type the same as a text field.

jessicamcinchak (commented):
Thanks for this! Code and functionality looks good to me, but definitely got me thinking about a few new questions around what our retention is/should actually be in practice

api.planx.uk/webhooks/sanitiseApplicationData/operations.ts
jessicamcinchak: noticed a new detail here today related to this: replacing users.email with `_REMOVED_` isn't actually sustainable because there's a unique constraint :see_no_evil: so scratch my `_SANITISED_` suggestion here and let's just aim for consistent soft delete processes / date fields :white_check_mark:

hasura.planx.uk/migrations/1670838102095_add_sanitised_at_columns/up.sql
DafyddLlyr: Perfect - I'll look at picking this up as a separate task.

api.planx.uk/saveAndReturn/utils.ts
```diff
@@ -210,7 +210,7 @@ const getPersonalisation = (
 
 /**
  * Mark a lowcal_session record as deleted
- * Sessions older than a week cleaned up nightly by cron job delete_expired_sessions on Hasura
+ * Sessions older than 6 months cleaned up nightly by cron job sanitise_application_data on Hasura
```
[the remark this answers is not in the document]
DafyddLlyr: Yeah this is true - it is more lenient than what we currently have. Due to this and related questions I opened https://github.com/theopensystemslab/planx-new/pull/1354 so we can separate the actual policy from the implementation.

Once we get sign off from Alastair (and partners?) there, I'll make any changes needed here.

api.planx.uk/webhooks/sanitiseApplicationData/index.ts
DafyddLlyr: You're bang on the money here - we really don't want Slack post from every PR, plus staging and prod each night...! I'll change this to only post failures, and just manually check it for the first couple of days after it's merged 👍

api.planx.uk/webhooks/sanitiseApplicationData/operations.ts
DafyddLlyr: Yeah this was my first idea and I hit this exact issue with emails also! 

I also considered something `{ "data": "**REDACTED**" }` for all JSONB column - I figured it might be visually helpful but in reality being able to easily query join to `flows` and then checking `sanitised_at IS NOT NULL`  / `sanitised_at IS NULL` is what I think would prove most helpful.

PR #1327, opened 2022-12-12
feat(dev): generate TS type definitions from the GraphQL Objects

gunar:
Replaces hand-crafted (and manually updated) type definitions for some of our types with a script that pulls the type definition from our GraphQL API.

### Open questions

- When do we run this script?
  - Do we leave it for the dev to make sure to run it?
  - Should we hook it up to CI?
  - Should we add a Husky script?(I've tried this and it's uneconomic but I could be convinced otherwise)


gunar:
I concur with your sentiment that the domain model and the data objects will never be perfectly aligned - and it is futile to even attempt to couple them. That being said, this PR is not attempting to do that. Here, we are simply making the existing data objects from our GraphQL API available in TypeScript. I agree that renaming some of our GraphQL objects could be helpful - and even creating some new abstractions - but I believe such tasks are separate from what this PR is trying to achieve. I believe having type-checking for the responses from our GraphQL queries is ultimately advantageous.

api.planx.uk/.env.test.example
```diff
@@ -41,3 +41,4 @@ UNIFORM_TOKEN_URL=👻
 UNIFORM_SUBMISSION_URL=👻
 
 SLACK_WEBHOOK_URL=👻
+FILE_API_KEY=👻
```
jessicamcinchak, line 44: This is already defined above on L17, grouped in the `# AWS infra` section rather than `# Integrations` fyi!

jessicamcinchak (commented):
Hey, apologies I've been a bit slow to properly review this one! 

I think this is a cool proof-of-concept of something that we _can_ technically do and a handy feature of working with GraphQL & TS, but I ultimately don't understand why we'd want to do it in our own API at this stage.

Here's a bit of reasoning for me:
- I think the value of our API is to wrap custom, quite domain-specific functions that are _not_ simple database operations that we could otherwise be doing with GraphQL alone - therefore more explicit manual types make a lot of sense to me, even if it means us humans are maintaining those types over time rather than a re-generation script
- We consume our GraphQL APIs in both our API & Editor, which for me means that it actually makes most sense for GraphQL types to just live in the GraphQL/database layer as they currently do! 
- This would have perhaps made more sense to me at the point of initially migrating our API to TS as sort of "starter" types. Reading through this diff, it feels like there are multiple points where our functionality/intentions are becoming _less_ clear because of the leaky abstractions as Ben better articulates. These are key examples for me:
Screenshot from 2022-12-15 08-35-37
Screenshot from 2022-12-15 08-34-21

So, while I appreciate this PR as a good discussion starter and definitely wouldn't hesitate to dust it off in the future or re-evaluate this approach at another time, I'd personally prefer if it closes un-merged right now (there wasn't a wider partner/infrastructure prioritization saying otherwise, right?)

gunar:
We want to move in the direction of improving the PlanX SDK (ie Client). Keeping this branch as the client might build upon and extend these generated types in the future.

PR #1328, opened 2022-12-13
fix(wip): Fix upload in Docker environments

DafyddLlyr:
Reverts change made in https://github.com/theopensystemslab/planx-new/pull/1271

This change meant that Pizzas were calling minio to handle file uploads, which isn't running as a service on Pizzas.

This is a quick and pragmatic revert / fix forward, I think more generally it might be more meaningful to have more fine grained values for `NODE_ENV` such as `"production" | "staging" | "pizza" | "test" | "local"`

We could also just override this in `docker-compose.pizza.yaml` if having it locally makes a significant difference 👌

### DIFF
--- a/api.planx.uk/s3/utils.ts
+++ b/api.planx.uk/s3/utils.ts
@@ -17,7 +17,7 @@ function useMinio() {
   } else {
     // Points to Minio
     return {
-      endpoint: "http://minio:9000",
+      endpoint: "http://127.0.0.1:9000",
       s3ForcePathStyle: true,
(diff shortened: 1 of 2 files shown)


jessicamcinchak (approved):
File uploads working for me, thanks for troubleshooting this one together !

gunar:
Thanks for fixing this. In case it wasn't clear, having it locally does make a difference because it gets us auto-reload of the api code. Yes, `docker-compose.pizza.yaml` should work I think.

PR #1330, opened 2022-12-13
chore: delegated error notifications

ghost:
I came across an issue with airbrake which led to me doing a quick refactor to log errors in an environment specific way.

This generic logger is delegated the responsibility of knowing how to appropriately log an error for the specific environment (dev/prod).


DafyddLlyr (approved):
This is nice - makes a lot of sense as a small refactor thanks 👍 

No action needed, but just to flag - some of this logic is also repeated in `api.planx.uk/airbrake.ts`

PR #1331, opened 2022-12-13
feat: add "copy" to editor-facing flow actions menu

jessicamcinchak:
lets editors make a unique copy of a flow within the same team

borrows logic from https://github.com/theopensystemslab/planx-team-onboarding/blob/main/client.js

future scope: 
- "move" flow to a new team
  - for now, a dev can manually hit the endpoint without `insert` param and manually copy/paste response `data` for one-off copy requests _betweeen_ teams or use onboarding script to copy canonical 3 services to a new team :slightly_smiling_face: 
- "copy to flow" action from "internal portal" node menu
- connect planx-team-onboarding scripts to use this endpoint

### DIFF
--- a/api.planx.uk/editor/copyFlow.test.ts
+++ b/api.planx.uk/editor/copyFlow.test.ts
@@ -1,0 +1,196 @@
+import supertest from "supertest";
+
+import { queryMock } from "../tests/graphqlQueryMock";
+import { authHeader } from "../tests/mockJWT";
+import app from "../server";
+import { Flow } from "../types";
+
+beforeEach(() => {
+  queryMock.mockQuery({
+    name: "GetFlowData",
+    matchOnVariables: false,
+    data: {
+      flows_by_pk: {
+        data: mockFlowData,
+      },
+    },
+  });
+
+  queryMock.mockQuery({
+    name: "InsertFlow",
+    matchOnVariables: false,
+    data: {
+      insert_flows_one: {
+        id: 2,
+      },
+    },
+  });
+
+  queryMock.mockQuery({
+    name: "InsertOperation",
+    matchOnVariables: false,
+    data: {
+      insert_operations_one: {
+        id: 1,
+      },
+    },
+  });
+});
+
+it("returns an error if authorization headers are not set", async () => {
+  const validBody = {
+    insert: false,
+    replaceValue: "T3ST",
+  };
+
+  await supertest(app)
+    .post("/flows/1/copy")
+    .send(validBody)
+    .expect(401)
+    .then((res) => {
+      expect(res.body).toEqual({
+        error: "No authorization token was found"
+      });
+    });
+});
+
+it("returns an error if required replacement characters are not provided in the request body", async () => {
+  const invalidBody = {
+    insert: false,
+  };
+
+  await supertest(app)
+    .post("/flows/1/copy")
+    .send(invalidBody)
+    .set(authHeader())
+    .expect(400)
+    .then((res) => {
+      expect(res.body).toEqual({
+        error: "Missing required values to proceed"
+      });
+    });
+});
+
+it("returns copied unique flow data without inserting a new record", async () => {
+  const body = {
+    insert: false,
+    replaceValue: "T3ST",
+  };
+
+  await supertest(app)
+    .post("/flows/1/copy")
+    .send(body)
+    .set(authHeader())
+    .expect(200)
+    .then((res) => {
+      expect(res.body).toEqual(mockCopyFlowResponse);
+    });
+});
+
+it("inserts copied unique flow data", async () => {
+  const body = {
+    insert: true,
+    replaceValue: "T3ST",
+  };
+
+  await supertest(app)
+    .post("/flows/1/copy")
+    .send(body)
+    .set(authHeader())
+    .expect(200)
+    .then((res) => {
+      expect(res.body).toEqual(mockCopyFlowResponseInserted);
+    });
+});
+
+// the original flow
+const mockFlowData: Flow["data"] = {
+  "_root": {
+    "edges": [
+      "rUilJQTag1",
+      "kNX8Rej9rk"
+    ]
+  },
+  "rUilJQTag1": {
+    "type": 100,
+    "data": {
+      "text": "Copy or paste?"
+    },
+    "edges": [
+      "Yh7t91FisE",
+      "h8DSw40zNr"
+    ]
+  },
+  "Yh7t91FisE": {
+    "type": 200,
+    "data": {
+      "text": "Copy"
+    }
+  },
+  "h8DSw40zNr": {
+    "type": 200,
+    "data": {
+      "text": "Paste"
+    }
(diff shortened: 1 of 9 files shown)


jessicamcinchak:
you'll have to wait and see in the morning ! 
Screenshot from 2022-12-13 20-44-01

api.planx.uk/editor/copyFlow.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (!req.user?.sub) {
+      return next({ status: 401, message: "User ID missing from JWT" });
+    }
+
+    // Fetch the original flow
+    const flow: Flow = await getFlowData(req.params.flowId);
+
+    // Ensure the newly copied flow has unique nodeIds
+    const randomReplacementCharacters = Math.random().toString(36).slice(2,6);
+    const uniqueFlowData = makeUniqueFlow(flow.data, randomReplacementCharacters);
```
DafyddLlyr, line 20: This is really worth thinking / talking about.

Nodes ids must be unique for flow navigation to work, we saw this issue with clones where non-unique node ids cause a lot of trouble.

From memory, the reason John was just replacing three characters was so that so could track if required where a flow came from (so there would be a relationship that devs could understand between `gtUKoDON` and `gtUKoLAM`). I don't think this has been something we've needed or used much, and especially if Editors can generate multiple copies this is going to have less and less value.

So my suggestion (please shoot holes in this!) would be -
 - Make the entire node Id unique to avoid issues with uniqueness in the long term
 - If we really care about understanding "where" a flow came from we can add a `flow_copy_events` audit table, or a `copied_from` column on flows. I'm not sure what the value of this would be apart from peace of mind?

api.planx.uk/server.ts
```diff
@@ -477,6 +478,9 @@ app.get(
   graphQLVoyagerHandler({ graphQLURL: "/introspect-all", validateUser: true })
 );
 
+// use with query param `insert=true` (optional)
```
DafyddLlyr, line 481: nit: This is in the body, not a query param

DafyddLlyr: A type for the body might be helpful also as opposed to just a comment here. The main reason for this is that if somebody were to post the body of `{ insert: "false" }` this would currently incorrectly evaluate `shouldInsert` to  be `true`.

```ts
const shouldInsert = Boolean(req.body?.insert);
```

I think that we probably need to have a better approach to handling validation at the API level across the board - something to discuss!

editor.planx.uk/src/pages/Team.tsx
```diff
@@ -322,10 +331,7 @@ const Team: React.FC<{ id: number; slug: string }> = ({ id, slug }) => {
                 flow={flow}
                 key={flow.slug}
                 teamId={id}
-                onDeleteSuccess={() => {
-                  fetchFlows();
-                }}
-                onRenameSuccess={() => {
+                onSuccess={() => {
```
DafyddLlyr, line 334: 👍

DafyddLlyr (changes requested):
One very small thing to take a look at with the `insert` variable but this is great.

Maybe we can talk unique node Ids later?

api.planx.uk/helpers.ts
```diff
@@ -20,6 +22,53 @@ const getFlowData = async (id: string): Promise<Flow> => {
...
+      version: 1,
+      creator_id: creatorId,
+    }
+  );
+
+  if (data) await createAssociatedOperation(data?.insert_flows_one?.id);
+  return data?.insert_flows_one;
+};
+
+// Add a row to `operations` for an inserted flow, otherwise ShareDB throws a silent error when opening the flow in the UI
```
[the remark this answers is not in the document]
DafyddLlyr: Yeah we have talked to handling this implementation detail at a db level (a trigger following an INSERT into `flows`). It's repeated a few times currently throughout the code base.

I think the decision was to keep the explicit insert and repeat where needed, until we re-evaluate our relationship with ShareDB I guess.

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -140,6 +141,22 @@ export const editorStore = (
     );
   },
 
+  copyFlow: async (flowId, teamId) => {
+    const token = getCookie("jwt");
+
+    return axios.post(
+      `${process.env.REACT_APP_API_URL}/flows/${flowId}/copy`,
```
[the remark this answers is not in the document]
jessicamcinchak: These are important questions ! 

My rational for API here was essentially: 
- `makeUniqueFlow` was already defined in the API helper methods because another endpoint depends on it, I don't want to repeat it in the Editor too (`dataMerged` is an example of another helper that is defined in both places currently, and definitely supports a case I think for an isomorphic node-module for logic methods like these)
- recognizing that we don't have the "move" action just yet, I imagine there may be occassional value in a dev being able to manually hit this function as an endpoint and just generate the unique copied flow data that could then be manually pasted into a new Hasura record under _any_ team (in addition to content editors being able to access from Editor)

editor.planx.uk/src/pages/Team.tsx
```diff
@@ -195,16 +195,10 @@ const FooterLinks = () => (
...
+  onSuccess: () => void;
 }
 
-const FlowItem: React.FC<FlowItemProps> = ({
-  flow,
-  teamId,
-  onDeleteSuccess,
-  onRenameSuccess,
-}) => {
+const FlowItem: React.FC<FlowItemProps> = ({ flow, teamId, onSuccess }) => {
```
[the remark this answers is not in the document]
jessicamcinchak: yep I think that's right - I'll update this to `refreshFlows`

api.planx.uk/editor/copyFlow.ts
DafyddLlyr: Yeah this totally works for me, it's a sensible approach here 👍

DafyddLlyr: This makes perfect sense to me, thanks for the summary 👌

api.planx.uk/editor/copyFlow.test.ts
```diff
@@ -0,0 +1,196 @@
...
+    "type": 110,
+    "data": {
+      "title": "Why do you want to copy this flow?",
+      "type": "short"
+    }
+  }
+};
+
+const mockCopyFlowResponse = {
+  message: `Successfully copied undefined`, // 'undefined' just reflects that we haven't mocked a flow.name here!
```
DafyddLlyr, line 185: 👍

jessicamcinchak:
[Allow content (cards, cards plus children, flows(?)) to be copied / pasted across flows and services.](https://trello.com/c/wMHLWdn1/678-allow-content-cards-cards-plus-children-flows-to-be-copied-pasted-across-flows-and-services)

DafyddLlyr (approved):
Thanks for those changes and comments 👍 Happy to merge once tests pass!

PR #1334, opened 2022-12-15
feat: Ordnance Survey proxy

DafyddLlyr:
## What does this PR do?
 - Creates a new API endpoint `/proxy/ordnance-survey`
   - This endpoint forwards valid requests for OS API services to the OS API, with an API key appended
 - Adds `ORDNANCE_SURVEY_API_KEY` secret to the API
 - Refactors some API code into a new `proxy` module

## Why do we need this?
Currently, our OS API key is exposed as part of our direct requests to the OS API. Using this proxy will allow us to hide this key.

## What's next?
- We'll need to merge https://github.com/theopensystemslab/map/pull/241 in the Map repo, and publish a new version (still in progress!)
- Update the version used in PlanX, add in the correct `osProxyEndpoint` value
- If testing is successful, remove `REACT_APP_ORDNANCE_SURVEY_KEY` from secrets


DafyddLlyr:
[(Map) Hide OS key from requests](https://trello.com/c/em3Asij7/1916-map-hide-os-key-from-requests)

api.planx.uk/proxy/ordnanceSurvey.ts
```diff
@@ -0,0 +1,52 @@
...
+    onProxyRes: (proxyRes, req) => setCORPHeaders(proxyRes, req),
+    pathRewrite: (fullPath, req) => appendAPIKey(fullPath, req)
+  })(req, res, next)
+};
+
+const isValid = (req: Request): boolean => isAllowListed(req) || isPlanX(req);
+
+const isAllowListed = (req: Request): boolean => MAP_ALLOWLIST.some(re => re.test(req.headers?.referer as string));
+
+const isPlanX = (req: Request): boolean => Boolean(req.headers.referer?.match(/^https:\/\/.*planx\.(pizza|dev|uk)\//i)?.length);
```
jessicamcinchak, line 33: I think we'll also need to account for custom domain names here (or in ALLOWLIST because I guess we'll actually want cross-origin CORS for that subset)?

api.planx.uk/server.ts
```diff
@@ -657,36 +653,6 @@ const server = new Server(app);
...
-      Authorization: `Bearer ${
-        process.env[
-          `GOV_UK_PAY_TOKEN_${req.params.localAuthority}`.toUpperCase()
-        ]
-      }`,
-    },
-    ...options,
-  });
-}
-
```
jessicamcinchak, line 689: appreciate this cleanup / re-location :+1:

jessicamcinchak (approved):
Looks good! Left one question about how custom domains will fit in here - but happy for this to merge for further testing before or after that's addressed since these endpoints won't actually be called yet

api.planx.uk/proxy/ordnanceSurvey.ts
DafyddLlyr: This is a great shout and something I just plain didn't consider 🙌  

I've gone for domains ending in `.gov.uk/` because...
 - No maintenance when we add new domains
 - Very few illicit calls to the API are likely to come from non-supported `.gov.uk` domains

I've also just simplified the `isValid()` functionality - one allow list works just fine, and adding "same-site" CORP headers really doesn't achieve anything if we're already properly validating.

PR #1335, opened 2022-12-15
fix: logger function with wrong number of args

ghost:
I'm not sure how this got through (`tsc` picks this up as an error at compile) but I've fixed it now!


DafyddLlyr (approved):
Thank you for fixing this!

PR #1336, opened 2022-12-16
feat: add "move" to editor-facing flow actions menu

jessicamcinchak:
lets editors move an existing flow to a different team. builds on #1331 

TODO
- double check what happens if you move a flow that is referenced as an external portal (do references break or are they OK? flowId isn't changing so I'm hoping they're fine from a schema perspective, but the visual node text for that portal in the graph might fall out of sync??)

future scope:
- decided how role and/or team-based permissions fit into this - right now any editor can move any flow to any team


editor.planx.uk/src/pages/Team.tsx
```diff
@@ -287,6 +301,21 @@ const FlowItem: React.FC<FlowItemProps> = ({ flow, teamId, refreshFlows }) => {
                 handleCopy();
               },
             },
+            {
+              label: "Move",
+              onClick: () => {
+                const newTeam = prompt("New team");
```
jessicamcinchak, line 307: ideally, this would probably be a clever dropdown of all existing team slugs besides the current one to select from - but a `prompt` is a simpler MVP implementation here and I think we can safely assume that only experienced/familiar content editors are going to be "moving" flows right now anyways

api.planx.uk/editor/moveFlow.ts
```diff
@@ -0,0 +1,82 @@
...
+        status: 400,
+        message: `Unable to find a team matching slug ${req.params.teamSlug}, exiting move`,
+      });
+    }
+  } catch (error) {
+    return next(error);
+  }
+};
+
+const getTeamIdBySlug = async (slug: Team["slug"]): Promise<Team["id"]> => {
```
DafyddLlyr, line 40: I appreciate this way of doing it - `Team["id"]` and `Team["slug"]` is more useful and meaningful I think.

editor.planx.uk/src/pages/Team.tsx
DafyddLlyr: Yeah I think this is totally fine for now whilst we have a small number of teams, and a small number of editors using this 👍

api.planx.uk/editor/moveFlow.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      id: 1,
+      team_id: 2,
+    },
+    data: {
+      update_flows_by_pk: {
+        id: 2,
+      },
+    },
+  });
+});
```
DafyddLlyr, line 32: This gets the test passing ✅ 
 - Return array of teams
 - Use strings for ids

```suggestion
  queryMock.mockQuery({
    name: "GetTeam",
    variables: {
      slug: "new-team",
    },
    data: {
      teams: [{
        id: "1",
      }],
    },
  });

  queryMock.mockQuery({
    name: "UpdateFlow",
    variables: {
      id: "1",
      team_id: "1",
    },
    data: {
      update_flows_by_pk: {
        id: "1",
      },
    },
  });
});
```

DafyddLlyr (approved):
Great stuff - again nice and simple and look forward to being able to call this from our onboarding scripts.

Small comment on tests that should fix this issue for you 👍 

The question about permissions is a good one - we'll need to keep this in mind when we cross that bridge.

Portals is another good one - I tested internal and external portals (in a pretty basic way) and it behaved as expected. As it's just the `team_id` column changing and no flow data or structure I would hope this is just fine?

api.planx.uk/editor/moveFlow.ts
```diff
@@ -0,0 +1,82 @@
...
+  next: NextFunction
+): Promise<Response | NextFunction | void> => {
+  try {
+    if (!req.user?.sub) {
+      return next({ status: 401, message: "User ID missing from JWT" });
+    }
+
+    if (!req.params?.flowId || !req.params?.teamSlug) {
+      return next({ status: 400, message: "Missing required values to proceed" });
+    }
```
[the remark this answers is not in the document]
jessicamcinchak: Very curious, have also been noticing this repeated pattern :+1:

api.planx.uk/editor/moveFlow.test.ts
jessicamcinchak: thank you for spotting those finicky details in the mocks ! passing now :white_check_mark:

PR #1337, opened 2022-12-16
chore: fix over-enthusiastic error logger

builditben:
Error logs should be quieter (and not error on setup!)

Note: `DEBUG=true` is now needed to see full logs


gunar (approved):

DafyddLlyr:
Merged as Ben won't be here for a few weeks 👍

PR #1343, opened 2022-12-20
docs: add deployment steps to README

jessicamcinchak:
Branch protection rules were recently adjusted, but I realized that I had incorrectly assumed yesterday what that meant for how we should deploy now - so writing down the process as I understand it after getting everything back in alignment today. 

Edits welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,11 +86,12 @@ git push --force
 
 ### Troubleshooting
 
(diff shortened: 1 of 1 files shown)


DafyddLlyr (approved):
This is perfect - thank you 👍

PR #1345, opened 2022-12-20
feat(ci): cron to deploy to production automatically

gunar:
## What is in this PR?

- deploy automatically `main` into `production` on a schedule
- allows manual triggering of this action for manual deployments

## How to test it?

After this PR is approved, I'll merge this into `main`, I'll ask on Slack if the team is comfortable with a production deploy, then try and execute the action manually (using the "dispatch" option). If the manual execution works well, the automatic execution will most likely work too.


gunar:
[Automatic deployment to production every Tue-Fri at 9am](https://trello.com/c/97w66NpJ/2217-automatic-deployment-to-production-every-tue-fri-at-9am)

.github/workflows/cron-deploy-to-production.yml
```diff
@@ -0,0 +1,23 @@
...
+    - name: sync
+      env:
+        # Uses `github.token`: the default, ephemeral token automatically generated by GitHub
+        # https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token
+        PERSONAL_ACCESS_TOKEN: ${{ github.token }}
+      run: |
+        set -xe
+        git clone "https://theopensystemslab:${PERSONAL_ACCESS_TOKEN}@github.com/planx-new.git" planx-new
+        cd planx-new || exit 1
+        git push --force origin origin/main:production
```
jessicamcinchak, line 23: with our recently updated `production` branch protection rules, i'd expect this `push` command to fail because it will be lacking an "approval" ? i think that perhaps a few of our next steps here have turned to be a bit contradictory to each other :upside_down_face:

jessicamcinchak (commented):
thanks for drafting this up! 

left one question. overall i'd prefer if this wasn't merged until after break when we're all back to regular schedules, so we're around to actively troubleshoot any errors - is that timing ok? 

can you also please update the "Deployments" section in the root README if this is implemented so we all have a clear, shared reference point for how deployments work?

.github/workflows/cron-deploy-to-production.yml
gunar: Yeah let's discuss this


image

gunar:
Agreed on enable the CRON only after the holidays season. Nonetheless, perhaps we can merge this PR now just with the manual workflow dispatch? It'd allow me (us) to test this script in the meantime. Then I'll create a reminder to enable the CRON next year (or anyone else will be able to as well).

jessicamcinchak (approved):
Talked about this during the dev call and decided: 
- We'll disable scheduled CRON until after the holidays
- This can merge to main now so the manually triggered action can be tested this week
- We'll disable branch protection rules on `production` only to enable the above (rules stay in place on `main` and we continue requiring at least 1 reviewer there)

Would still appreciate if you can update README deployment section with details of new action please. 

Thanks!!

PR #1347, opened 2022-12-21
fix: render HTML tags in the Contact Us section

gunar:
- Fixes https://trello.com/c/WZOdzJ4c/2227-confirmation-component-contact-us-section-not-rendering-html-content-correctly
- The rich text editor converts chars into their escaped version, `<` turns into `&lt;` for example. To render the tags correctly, a replacement regex was added to "unescape" chars before rendering.
- Alternatively, and to avoid using a regex, we could add a plugin to the tiptap editor to convert the text to Markdown first then convert it back to HTML, which feels less safe than just replacing the escaped chars. Here is an example: https://codesandbox.io/s/tiptap-0sqm3i?file=/src/components/Tiptap.tsx:4861-4920

### DIFF
--- a/editor.planx.uk/src/ui/ReactMarkdownOrHtml.test.tsx
+++ b/editor.planx.uk/src/ui/ReactMarkdownOrHtml.test.tsx
@@ -1,4 +1,7 @@
-import { incrementHeaderElements } from "./ReactMarkdownOrHtml";
+import {
+  incrementHeaderElements,
+  replaceEscapedTags,
+} from "./ReactMarkdownOrHtml";
 
 describe("Header element incrementor", () => {
   it("should return HTML unaltered if it does not contain H1 or H2 tags", () => {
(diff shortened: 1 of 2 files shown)


DafyddLlyr (changes requested):
So I think there's a simpler approach here.

The core issue here comes from the fact that there's a mixture of HTML and markdown in the original source (`ConfirmationEditor` > `initialValues` > `contactInfo`). This means we end up with a mixture being rendered by `ReactMarkdownOrHtml` (which can handle either input).

If we just update the original input to be just HTML in `ConfirmationEditor` this should fix this.

```js
contactInfo:
  props.node?.data?.contactInfo ||
    `You can contact us at <em>planning@lambeth.gov.uk</em>
     <br/><br/>
     What did you think of this service? Please give us your feedback using the link in the footer below.`,
```

We will need to manually update Confirmation nodes which have the incorrect markdown, but there's just a handful of them.

In future we should be able to catch this mixture of markdown and HTML by just checking that the RichTextEditor displays correctly in the Editor modal. This is how it currently looks which we should have picked up on - 

image

gunar:
If the objective is to render HTML inputted in the Rich Text Editor, the suggested solution won't work. The editor will still escape HTML:
image

DafyddLlyr:
I'm getting a different result when I try this - 

image

image

There's a Pizza here with the suggested change implemented - https://github.com/theopensystemslab/planx-new/pull/1366

gunar:
It will render correctly with the initial string, but when the user edits and adds a new HTML tag, the tag will be escaped

Bug:

https://user-images.githubusercontent.com/7684574/211355376-26c773c2-bd7c-4f06-88de-8688de828f38.mov

Fix:

https://user-images.githubusercontent.com/7684574/211355389-73c61fb9-15a8-4c49-ac96-925cec8aa655.mov

jessicamcinchak:
Closing in favor of #1366 - because an intended consequence of this change would allow you to add any HTML tags such as inputs, etc

PR #1348, opened 2022-12-21
fix: Send current metadata via FeedbackFish widget, not initial metadata

DafyddLlyr:
## What's the problem?
 Currently the `metadata` passed into the FeedbackFish widget is only calculated on initial render of the `Footer` and `PhaseBanner` components. We want it to be "live" with the applicant's current state, not just the initial state on load.

## How is this fixed?
Call `getFeedbackMetadata()` when the widget is opened (using the `window`'s `message` event).

## Also...
There's a bit of a rearranging of the `Footer` component here to split things up a little bit.

## To test
- Navigate a few questions into a flow and submit feedback via the `Footer` or `PhaseBanner`
- The network tab should show the full "live" metadata being sent (e.g. breadcrumbs)

### DIFF
--- a/editor.planx.uk/src/components/Footer.tsx
+++ b/editor.planx.uk/src/components/Footer.tsx
@@ -42,40 +42,53 @@ export interface Props {
   children?: React.ReactNode;
 }
 
+// XXX: This component is located in the DOM within the Footer component, but as it's appearance is controlled by a global window "message" event. This means that its appearance can be controlled by other instances of the FeedbackFish widget (such as in the PhaseBanner)
+const FeedbackPrivacyNote = ({
+  onClose,
+}: {
+  onClose: (e: React.MouseEvent) => void;
+}) => (
+  <Dialog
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/components/Footer.tsx
```diff
@@ -42,49 +42,85 @@ export interface Props {
...
-          //      was received. If not then never show feedback warning msg
-          if (width === 0 && height === 0) {
-            feedbackFishPostMessageWorkingCorrectly = true;
-          }
-          setFeedbackPrivacyNoteVisible(
-            width > 0 && height > 0 && feedbackFishPostMessageWorkingCorrectly
-          );
-        }
-      } catch (err) {}
+      setMetadata(getFeedbackMetadata());
```
DafyddLlyr, line 81: This is the only new logic in this component, everything else is just splitting up / tidying up / adding comments.

jessicamcinchak (approved):
Works as expected in testing, thanks for catching & updating!

PR #1350, opened 2022-12-22
chore: Commit Editor environmental variables

DafyddLlyr:
## What does this PR do?
 - Simplify secret management by committing our Editor `.env` file
 - These are all publicly exposed, and are not secrets
 - This work follows on from implementing a proxy for OS API calls #1334

### DIFF
--- a/scripts/push-secrets.sh
+++ b/scripts/push-secrets.sh
@@ -3,7 +3,6 @@
 cd "$(dirname "${BASH_SOURCE[0]}")"
 
 aws s3 cp ./../.env s3://pizza-secrets/root.env
-aws s3 cp ./../api.planx.uk/.env s3://pizza-secrets/api.env.test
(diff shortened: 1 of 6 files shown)


scripts/push-secrets.sh
```diff
@@ -3,7 +3,6 @@
 cd "$(dirname "${BASH_SOURCE[0]}")"
 
 aws s3 cp ./../.env s3://pizza-secrets/root.env
-aws s3 cp ./../api.planx.uk/.env s3://pizza-secrets/api.env.test
-aws s3 cp ./../editor.planx.uk/.env s3://pizza-secrets/editor.env
+aws s3 cp ./../api.planx.uk/.env.test s3://pizza-secrets/api.env.test
```
DafyddLlyr, line 6: There was also a mistake here - the local API env file is named `.env.test`

jessicamcinchak (approved):
thank you!

PR #1351, opened 2022-12-22
test(skip): Branching filters logic bug

DafyddLlyr:
## What does this PR do?
 - Rearranges existing filter tests into groups
 - Adds a new (failing / skipped) test case which exposes the branching filters logic bug described here - https://trello.com/c/JYnCHyLF/2010-fix-improve-how-filters-work

## What's happening?
Here's the flow used for the test case on staging - https://editor.planx.uk/testing/branch-filter-test

The image below demonstrates the incorrect path taken (in green) - 
<p align="center">
  
</p>

Expected Behaviour - The second filter should take me down the `immunityPath2` route

Observed Behaviour - The second filter takes me down the `noResultPath2` route

I haven't had a huge amount of success in narrowing down the cause of this issue, but I believe it's likely the logic in `upcomingCardIds()`.

This test case exposes the fact that the graph appears to navigate from the `fork` node, to the `immunityFlag2` node when we would expect it to go to `immunityPath2`. On the first filter we can see it takes us from `setImmunity` directly to `immunityPath1` as expected

## Proposal
 - We can merge this PR, or close and it and keep it as reference - either is fine for now
 - Write an abundance of tests in order to better understand and document `upcomingCardIds()`
 - Once in place, consider refactoring `upcomingCardIds()` if we have not exposed the issue in the previous step


editor.planx.uk/src/pages/FlowEditor/lib/__tests__/filters.test.ts
```diff
@@ -1,163 +1,80 @@
...
 import { vanillaStore } from "../store";
+import flowWithBranchingFilters from "./mocks/flowWithBranchingFilters.json";
+import flowWithSingleFilter from "./mocks/flowWithSingleFilter.json";
 
 const { getState, setState } = vanillaStore;
+const { upcomingCardIds, resetPreview, record, currentCard, collectedFlags } =
+  getState();
 
 // https://i.imgur.com/k0kkKox.png
+describe("Single filter", () => {
```
jessicamcinchak, line 10: naming nit: rather than "Single filter", can we please refer to this as "Root filter"? as i understand this bug, it's not actually about the number of the filter type nodes, but rather their placement in the graph (eg directly under `_root` or on a branch). Even if I had multiple filters that were all direct edges of `_root`, they should still behave as expected!

jessicamcinchak (approved):
Thanks for writing a new test and tidying up the others. I'm happy for it merge even if skipped as a jumping off point for any of us to pick this up next. Can you please just make sure this PR gets linked to the existing Trello card so we don't lose track of this piece of work? 

And one small consideration: your minimal flow on staging to recreate is very helpful, but I think it will get overwritten next Sunday by the production data sync - so perhaps you can also copy it to editor.planx.uk/testing ?

editor.planx.uk/src/pages/FlowEditor/lib/__tests__/filters.test.ts
DafyddLlyr: This is correct - I'll make the change.

I originally had it as "single filters" and "multiple filters" - I redefined the latter but not the original 👍

DafyddLlyr:
[Fix / Improve how filters work](https://trello.com/c/JYnCHyLF/2010-fix-improve-how-filters-work)

DafyddLlyr:
Solid feedback as always thank you @jessicamcinchak 

- [ ] Trello card linked
- [ ] Flow copied to prod (and link in PR description replaced)
- [ ] Renamed from single filter to root filter

PR #1352, opened 2022-12-22
fix: always show DrawBoundary area

jessicamcinchak:
was testing FOIYNPP services and noticed that the area calculation was hidden when the file upload option was toggled off, but I think this is probably unintentional and we want to always show the area calculation - it was still being written to the passport in the background! 

now we'll render this when `props.hideFileUpload` is true:
Screenshot from 2022-12-22 10-52-24


DafyddLlyr (approved):
Good catch! This certainly looks like a regression I introduced when I added the `MapFooter` component 🤦‍♂️

PR #1354, opened 2022-12-22
doc: Data retention policy proposal

DafyddLlyr:
Follow up from #1325 - will share with @Alastairparvin also 👍

### DIFF
--- a/doc/architecture/decisions/0005-data-retention-policy.md
+++ b/doc/architecture/decisions/0005-data-retention-policy.md
@@ -10,6 +10,8 @@ Accepted
 
 We have agreed that we need to implement a data retention policy, and the basics of this has been created in [#1325](https://github.com/theopensystemslab/planx-new/pull/1325).
 
+This document and policy specifically covers user data related to applications submitted via PlanX. Further documents will lay out our policies on Flow Data and Analytics.
+
 As part of this, it surfaced that we do not have an agreed and formalised data retention policy which lives outside of the code to serve as a reference.
 
 ## Decision
--- a/doc/architecture/decisions/0005-data-retention-policy.md
+++ b/doc/architecture/decisions/0005-data-retention-policy.md
@@ -83,6 +83,12 @@ Here are the broad catergory of changes -
 | **Event Metadata** | Indefinite | 6 months from creation date, then deleted | 
 | **Feedback** | Indefinite | Indefinite
 
+
+### Timetable
+The new policy will be implemented once this proposal has been accepted. This means that records over 6 months old at the point the feature is merged to production will be sanitised or deleted.
+
+An automated scheduled task will run each night to sanitise and delete records which meet the agreed criteria.
+
 ### What does "Sanitation" mean?
 Currently, we would simply remove / delete and user data (e.g. email, breadcrumbs, passport) whilst retaining the remainder of the metadata associated with a record (such as `id`, `created_at`). This would allow to still have enough information to resconstruct user journeys if required to do so for auditing purposes.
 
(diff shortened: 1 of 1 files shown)


doc/architecture/decisions/0005-data-retention-policy.md
```diff
@@ -0,0 +1,100 @@
...
+
+### What does "Sanitation" mean?
+Currently, we would simply remove / delete and user data (e.g. email, breadcrumbs, passport) whilst retaining the remainder of the metadata associated with a record (such as `id`, `created_at`). This would allow to still have enough information to resconstruct user journeys if required to do so for auditing purposes.
+
+In future, a more fine grained approach could be implemented to only sanitise personally identifiable data using tags.
+
+## Questions
+ - Is 6 months too long?
+ - Do we actually need Session Backups now that we have Lowcal Sessions?
+ - Is there value in retaining reconciliation events indefinately?
```
jessicamcinchak, line 91: One more question I'd add to this list: "When does this retention policy take effect?" Eg once merged are we immediately sanitising records `created_at` more than 6 months ago? Or does the first sanitisation happen 6 months from merge? 

The first options seems ideal and sensible to me, but feels worth asking & being explicit about since this has been a fuzzy topic for so long!

doc/architecture/decisions/0005-data-retention-policy.md
```diff
@@ -0,0 +1,110 @@
...
+    FeedbackFish (last 250 records, rolling): 2022-01-14, 2022-12-31
+    Feedback Learning Log (anonymised): 2022-01-14, 2022-12-31
+```
+
+Here are the broad catergory of changes - 
+
+|| Current Retention Period | Proposed Retention Period |
+|---|---|---|
+| **User Data**| Lowcal Sessions - 7 days, then deleted. Session Backups - Indefinite | 6 months from submission date, then sanitised and held indefinitely |
+| **Uploaded Files**| Indefinite | 6 months from upload date |
```
DafyddLlyr, line 81: There's a ticket to implement this here - https://trello.com/c/XRcoHTyW/2237-implement-agreed-data-retention-policy-for-user-uploaded-files

jessicamcinchak (approved):
thanks again for this!

final nitty thought: is it worth adding onre more sentence in "Context" to just explicitly say that this policy targets/covers "user data" related to submitted applications. 

just thinking it may make sense to have a complimentary doc stating policies for "content / flow data / anonymized user data aka analytics" in near future and be clear that's not part of this (yet) !

doc/architecture/decisions/0005-data-retention-policy.md
```diff
@@ -0,0 +1,111 @@
...
+| **Feedback** | Indefinite | Indefinite
+
+
+### Timetable
+The new policy will be implemented once this proposal has been accepted. This means that records over 6 months old at the point the feature is merged to production will be sanitised or deleted.
+
+An automated scheduled task will run each night to sanitise and delete records which meet the agreed criteria.
+
+### What does "Sanitation" mean?
+Currently, we would simply remove / delete and user data (e.g. email, breadcrumbs, passport) whilst retaining the remainder of the metadata associated with a record (such as `id`, `created_at`). This would allow to still have enough information to resconstruct user journeys if required to do so for auditing purposes.
```
gunar, line 94: Is this true though? If we delete breadcrumbs, can we still reconstruct the user journey?

Also, small typo on "resconstruct"

gunar (approved):

doc/architecture/decisions/0005-data-retention-policy.md
DafyddLlyr: Thanks for this - I've not clarified my meaning here. What I'm referring to is "high level" user journeys (e.g. when was an application started, payed for, and submitted) as opposed to the granular question by question journey.

jessicamcinchak (approved):

PR #1357, opened 2022-12-22
fix(ci): add missing namespace

gunar:
Should fix the error below

https://github.com/theopensystemslab/planx-new/actions/runs/3759553898/jobs/6389256274


DafyddLlyr:
Merged so we can try the script on a deploy this morning @gunar 👌

PR #1368, opened 2023-01-10
feat: Setup alt theme behind feature flag

DafyddLlyr:
This PR sets up an alternate theme which can be toggled using the `ALT_THEME` feature flag. This will allow @ianjon3s to work on theming in relative isolation.

### DIFF
--- a/editor.planx.uk/src/theme.ts
+++ b/editor.planx.uk/src/theme.ts
@@ -9,6 +9,7 @@ import createPalette, {
   PaletteOptions,
 } from "@mui/material/styles/createPalette";
 import { deepmerge } from "@mui/utils";
+import { hasFeatureFlag } from "lib/featureFlags";
 
 const GOVUK_YELLOW = "#FFDD00";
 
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/theme.ts
```diff
@@ -226,11 +227,23 @@ const getThemeOptions = (primaryColor: string): ThemeOptions => {
   return themeOptions;
 };
 
+const getAltThemeOptions = (primaryColor: string): ThemeOptions => {
+  const themeOptions = getThemeOptions(primaryColor);
+  const altThemeOptions: ThemeOptions = {
+    typography: {
+      fontFamily: "Arial",
```
jessicamcinchak, line 234: in the meantime before this becomes a super polished theme, i like that it can be a prank alastair theme :smiling_imp:

DafyddLlyr: I was going to go Comic Sans but couldn't do it to myself 😂

jessicamcinchak (approved):

PR #1369, opened 2023-01-10
fix(ci): Install Cypress binary

DafyddLlyr:
Looks like a breaking change I introduced here - https://github.com/theopensystemslab/planx-new/pull/1323/commits/452ea348461715894d28e01c1ab86207b1a949f0

It worked on the above PR but it looks like current CI builds are now failing.


DafyddLlyr:
Closing - we will just replace the Cypress tests with Playwright tests, so fixing this right now isn't necessary.

PR #1370, opened 2023-01-10
feat(pizza): sync down users from production

gunar:
- have created a new user in production called `github_actions` and have given this user restricted access so that it can only read from the `users` table and nothing else
- have added the password for this user to both GitHub Actions Secrets and to 1Password


gunar:
[Create script that syncs users from production into staging and pizzas](https://trello.com/c/rZ5AkV3F/2231-create-script-that-syncs-users-from-production-into-staging-and-pizzas)

.github/workflows/pull-request.yml
```diff
@@ -341,6 +341,7 @@ jobs:
             cat .env .env.temp .env.staging > .env.prod
 
             SSH_PASSWORD=${{ secrets.SSH_PASSWORD }} sh scripts/pullrequest/create.sh
+            PRODUCTION_PG_URL_FOR_USER_GITHUB_ACTIONS=${{secrets.PRODUCTION_PG_URL_FOR_USER_GITHUB_ACTIONS}} sh scripts/sync-users.sh
```
jessicamcinchak, line 344: thanks for this! 

one initial comment: if we only sync users from production onto pizzas, I believe editor authentication will still fail because pizza's use _staging_ google oauth. so I think we first need to sync production to staging (so I'd expect to see this command also called from `.github/workflows/sync-staging-db.yml`), and then production/staging to pizzas. 

does that make sense? or is there another reason i'm missing that you're able to skip the sync to staging here? 

once updated with a successful vult build, happy to test and approve!

gunar: The users table doesn't contain any oauth data so I think we're good to sync down from production into both staging and pizzas.

gunar:
Users successfully being sync'ed down from production to pizza.

Next up, sync from production to staging.

.github/workflows/pull-request.yml
gunar: Well then again, there's the issue that user ids need to match. You're right, best is to do prod->sta, then sta->pizza

scripts/sync-down-from-production.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env sh
```
gunar, line 1: Can I get an early review on this file?

scripts/sync-down-from-production.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# Check wheter `docker compose` or `docker-compose` is available
+if command -v docker-compose > /dev/null; then
+  DOCKER_COMPOSE=docker-compose
+elif command -v docker compose > /dev/null; then
+  DOCKER_COMPOSE="docker compose"
+else
+  echo "Neither docker compose nor docker-compose is available"
+  exit 1
+fi
```
gunar, line 17: Trying to make it as flexible as possible, so we can run this on dev env, pizzas, and staging

scripts/sync-down-from-production.sh
```diff
@@ -0,0 +1,70 @@
...
+
+REMOTE_PG="${PRODUCTION_PG_URL_FOR_USER_GITHUB_ACTIONS}"
+LOCAL_PG="postgres://${PG_USERNAME}:${PG_PASSWORD}@postgres/${PG_DATABASE}"
+
+# run postgres utils inside a container so that we don't have to install them on the host
+pg_exec() {
+  (cd "${SCRIPT_DIR}/.." && ${DOCKER_COMPOSE} exec -T postgres bash -c "${*}")
+}
```
gunar, line 25: Make it so devs don't have to install psql locally necessarily.

scripts/sync-down-from-production.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# truncate tables users and teams and cascade
+pg_exec "psql --command='TRUNCATE TABLE users, teams CASCADE;' ${LOCAL_PG}"
+# insert users from production
+pg_exec "psql --command=\"\\COPY (SELECT * FROM users) TO STDOUT (FORMAT CSV, DELIMITER ';')\" \"${REMOTE_PG}\" | psql --command=\"\\COPY users FROM STDIN (FORMAT CSV, DELIMITER ';')\" \"${LOCAL_PG}\""
```
gunar, line 30: Is this too _dark magic-y_?

scripts/sync-down-from-production.sh
```diff
@@ -0,0 +1,70 @@
...
+REMOTE_PG="${PRODUCTION_PG_URL_FOR_USER_GITHUB_ACTIONS}"
+LOCAL_PG="postgres://${PG_USERNAME}:${PG_PASSWORD}@postgres/${PG_DATABASE}"
+
+# run postgres utils inside a container so that we don't have to install them on the host
+pg_exec() {
+  (cd "${SCRIPT_DIR}/.." && ${DOCKER_COMPOSE} exec -T postgres bash -c "${*}")
+}
+
+# truncate tables users and teams and cascade
+pg_exec "psql --command='TRUNCATE TABLE users, teams CASCADE;' ${LOCAL_PG}"
```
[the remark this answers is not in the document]
jessicamcinchak: If we're truncating tables, have you thought at all about hwo that will work with changes that involve a hasura migration (especially thinking about local & pizza use case here I guess where ${LOCAL_PG} may have a different shape than prod)? 

I think an insert should still succeed even if a new column has been added, but what about removing columns / changing data types etc (not migrations we do often, but just trying to think through possible side effects with this truncation approach!)

scripts/sync-down-from-production.sh
```diff
@@ -0,0 +1,70 @@
...
+# insert users from production
+pg_exec "psql --command=\"\\COPY (SELECT * FROM users) TO STDOUT (FORMAT CSV, DELIMITER ';')\" \"${REMOTE_PG}\" | psql --command=\"\\COPY users FROM STDIN (FORMAT CSV, DELIMITER ';')\" \"${LOCAL_PG}\""
+# insert teams from production
+pg_exec "psql --command=\"\\COPY (SELECT * FROM teams) TO STDOUT (FORMAT CSV, DELIMITER ';')\" \"${REMOTE_PG}\" | psql --command=\"\\COPY teams FROM STDIN (FORMAT CSV, DELIMITER ';')\" \"${LOCAL_PG}\""
```
jessicamcinchak, line 32: rather than `select * from teams` can we please omit column `submission_email` from data syncs? basically, we don't want to accidently send live council contacts emails when we're testing from pizzas/non-prod envs! more context over in #1400

scripts/sync-down-from-production.sh
```diff
@@ -0,0 +1,70 @@
...
+)
+pg_exec "psql --command=\"${SQL}\" \"${LOCAL_PG}\""
+# insert published_flows (the last two)
+SQL="\\COPY (SELECT id, data, flow_id, summary, publisher_id FROM (SELECT id, data, flow_id, summary, publisher_id, ROW_NUMBER() OVER (PARTITION BY flow_id ORDER BY created_at DESC) as row_num FROM published_flows) as subquery WHERE row_num <= 2) TO STDOUT (FORMAT CSV);"
+pg_exec "psql --command=\"${SQL}\" \"${REMOTE_PG}\" | psql --command=\"\\COPY published_flows (id, data, flow_id, summary, publisher_id) FROM STDIN (FORMAT CSV)\" \"${LOCAL_PG}\""
+# (now with `users` sync, this might not be needed anymore) set published_flows.publisher_id=1 to prevent errors with non-existing ID constraint
+
+
+
+# TODO: Do all of this in a transaction ideally
```
jessicamcinchak, line 65: :heavy_plus_sign: would second wrapping in a transaction so we have a safe rollback on failure / don't end up in with a fully truncated but partially synced db etc

scripts/sync-down-from-production.sh
jessicamcinchak: overall seems like a good, more simplified direction to me - thanks for this! appreciate the explicit pg role for accessing the `REMOTE_DB` and containerized approach to script dependencies here.

main open questions for me: 
- can we wrap psql commands starting from truncate in a transaction
- double check it won't fail if you have a local hasura migration on one of the synced tables

scripts/sync-down-from-production.sh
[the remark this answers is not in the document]
gunar: Been thinking about this. I guess the PR that implements a migration will also have to update this sync'ing script at the same time. Possibly, we might need to disable sync'ing of specific tables until the migration reaches production but that's not optimal.

gunar:
This is now ready for review. It uses the same code, and the same container, on the development environment and on pizzas. I _believe_ it's okay to merge this first then tackle the sync'ing of staging on the next PR but let me know.

jessicamcinchak (commented):
This is overall looking really good, and feels much more straightforward than prior GraphQL approach! 

The tables on the pizza look good, but this script is not actually working for me locally yet - here's what I did & the console error I'm seeing:
- Pulled down your changes here to a local branch
- Emptied my local database `docker compose down --remove-orphans -v`
- Spun up docker triggering the seed script `docker compose up --build`
Screenshot from 2023-02-22 10-13-18
(also note that PG password is logged out by Docker here, we should confirm this doesn't also get exposed in action logs when Vultr is spinning up/updating docker!)

One more comment: 
- I'd also expect to see a change to [this file](https://github.com/theopensystemslab/planx-new/blob/main/.github/workflows/sync-staging-db.yml) as part of this PR (espsecially the line which runs `upsert_production_flows.js`)? Without also updating the staging db sync simultaneously, I think we could end up with pizzas that have production users and staging users that do _not_ match production users yet, and therefore Google OAuth failures? But maybe there's more to the order-of-syncs that I'm missing here?

jessicamcinchak:
@gunar can confirm this works for me locally now that I have fresh env vars, thanks for flagging that yesterday! Prior comment about also syncing to staging still outstanding. Thanks!

gunar:
This is ready for review. I _think_ the only way to test the staging script is by merging this to `main` (then dispatching the action manually).

jessicamcinchak (approved):

PR #1372, opened 2023-01-12
fix: when deleting a flow, cascade to its' associated analytics records

jessicamcinchak:
addresses https://john-opensystemslab-io.airbrake.io/projects/329753/groups/3448889502649653859

currently if you attempt to delete a flow in the editor that has associated analytics, it'll throw a postgres foreign key violation error in the console with no user feedback. this PR updates the relevant foreign key relationships to `cascade` on update/delete rather than `restrict`. 

open question: i'm not sure if the new data retention policy covers service content as well as user data yet in its first iteration, but this might be a good explicit note to add at some point regarding service content. Eg "Anonymized analytics for a flow are retained for the whole lifespan of the flow. When a flow is deleted, its' associated analytics are also deleted."


hasura.planx.uk/migrations/1673514721705_set_fk_public_analytics_flow_id/up.sql
```diff
@@ -0,0 +1,11 @@
...
+  add constraint "analytics_flow_id_fkey"
+  foreign key ("flow_id")
+  references "public"."flows"
+  ("id") on update cascade on delete cascade;
+
+alter table "public"."analytics_logs" drop constraint "analytics_log_analytics_id_fkey",
+  add constraint "analytics_logs_analytics_id_fkey"
+  foreign key ("analytics_id")
+  references "public"."analytics"
+  ("id") on update cascade on delete cascade;
```
jessicamcinchak, line 11: previously, when an `analytics` record was deleted, it's associated `analytics_log` was configured to set the `analytics_id` field to null, but this triggered another violation because `analytics_id` has a NOT NULL constraint. 

now, we'll simply cascade the flow deletion to _both_ `analytics` & `analytics_logs`.

gunar (approved):
Tested on pizza and cascading is working. I'm able to delete a flow that had analytics.

Note that E2E tests aren't passing.

PR #1374, opened 2023-01-13
fix: remove cypress

ghost:
I'll come back to port the 2 small cypress tests to playwright but keen to unblock CI with this quick fix.


jessicamcinchak (approved):
Sounds like a good plan to me ! Thank you for un-blocking :heavy_plus_sign:

PR #1375, opened 2023-01-16
fix: populate flow name in "Submission" email template and mark associated session as submitted

jessicamcinchak:
I expect Emily's going to be doing some testing on "Send to email" this week, so two quick tidy-ups: 
- Mark the associated lowcal_session as submitted on success to avoid any confusing expiry reminder emails later
- Populate the Planx flow name in the email template

How to test:
- Add `"sendToEmail":"<your email>"` to `team.settings` on this pizza (pick any un-assigned team for testing)
- Edit a flow within that team: make sure it has a Send component with "Email to planning office" selected and the content is published
- Submit a test application 
- Check that: you have received an email; the template shows the name of your flow without hyphens; and your lowcal_session has a `submitted_at` timestamp in Hasura

### DIFF
--- a/api.planx.uk/send/email.ts
+++ b/api.planx.uk/send/email.ts
@@ -3,6 +3,7 @@ import { stringify } from "csv-stringify";
 import { NextFunction, Request, Response } from "express";
 import fs from "fs";
 import { gql } from "graphql-request";
+import capitalize from "lodash/capitalize";
 import os from "os";
 import path from "path";
 
@@ -35,7 +36,7 @@ const sendToEmail = async(req: Request, res: Response, next: NextFunction) => {
       const config: EmailSubmissionNotifyConfig = {
         personalisation: {
           emailReplyToId: notify_personalisation.emailReplyToId,
-          serviceName: payload?.flowName[0]?.toUpperCase() || "PlanX",
+          serviceName: capitalize(payload?.flowName) || "PlanX",
           sessionId: payload.sessionId,
           applicantEmail: payload.email,
           downloadLink: `${process.env.API_URL_EXT}/download-application-files/${payload.sessionId}?email=${settings.sendToEmail}&localAuthority=${req.params.localAuthority}`,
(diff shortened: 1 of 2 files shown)


api.planx.uk/send/email.ts
```diff
@@ -44,9 +44,21 @@ const sendToEmail = async(req: Request, res: Response, next: NextFunction) => {
 
       // Send the email
       const response = await sendEmail("submit", settings.sendToEmail, config);
-      return res.json(response);
+      if (response?.message === "Success") {
+        // Mark session as submitted so that reminder and expiry emails are not triggered
+        markSessionAsSubmitted(payload.sessionId);
 
-      // TODO Mark lowcal_session as submitted, create/update audit table (and setup Slack notification trigger?)
+        // TODO create audit table entry? setup event trigger for slack notification on new row?
```
jessicamcinchak, line 51: Keeping this TODO comment around so we don't lose track of it, but no urgency for this step in my mind ! 

We have Hasura Event logs for auditing purposes in the meantime and can decide later if a Slack notification makes sense for discretionary email-only submission services (theoretically there isn't the same visibility concern here as Uniform connector for example).

api.planx.uk/send/email.ts
```diff
@@ -35,7 +35,7 @@ const sendToEmail = async(req: Request, res: Response, next: NextFunction) => {
       const config: EmailSubmissionNotifyConfig = {
         personalisation: {
           emailReplyToId: notify_personalisation.emailReplyToId,
-          serviceName: "Test",
+          serviceName: payload?.flowName[0]?.toUpperCase() || "PlanX",
```
DafyddLlyr, line 38: When I tried this, `flowName` wasn't an array. This resulted in the string being truncated to just the first letter.

image

api.planx.uk/send/email.ts
DafyddLlyr: Yeah this seems fine to me also - we can address this as and when it's required.

DafyddLlyr (changes requested):
As always, I really appreciate the testing notes!

One small issue to take a look at 👍

DafyddLlyr (approved):
Perfect!

PR #1377, opened 2023-01-17
feat: FindProperty component supports option to plot new address without a UPRN

jessicamcinchak:
Adds option to "Plot a new address" aka "Find the property on the map" for sites that do not yet have a known street address or UPRN. Currently, these applications exclusively get routed to Planning Portal via the `<ExternalPlanningSiteDialog />`.

What this changes:
- Editors can now control whether to allow users to plot new addresses. If this is not toggled on, we'll still default to using the existing `<ExternalPlanningSiteDialog />`. If it is toggled on, the same link will now switch pages to an interactive map where you can place a single point and enter a text description of your site (very similar to Planning Portal's UI for this).
- `SiteAddress` interface is updated to have a minimum set of fields shared between the two methods of setting an address. The minimum type still satisfies location-based graph dependences like fetching associated LAD to auto-answer if correct team, rendering Draw Boundary map, and fetching Planning Constraints based on a fallback point if necessary. If you propose a new address, it will _not_ set `property.type` meaning you will simply be asked this question later in the flow; this shouldn't interfere with our future ticket to allow this value to be "changed" (we'll only show "change" link if defined).

Testing (eg https://1377.planx.pizza/buckinghamshire/search):
- You can add/update a FindProperty component to "allow non-UPRN addresses" and add a title and description for the "second" page - all fields are populated as expected in the preview environment
- Find property always _starts_ on the page wtih the autocomplete component; you can switch pages to plot a new address; About the property page renders known details; Draw boundary, Planning constraints, and Review all reference/render address info as expected
  - `passport.data._address` should reflect the address "source" and at least the minimum fields
- From "Review" you can "change" the address (or you can go "back" from anywhere) and land on the right "page" of FindProperty with pre-populated values (I know the visual map marker is still missing!)

Open questions:
- I've had past verbal confirmation that BOPS will be able to accept an application without a UPRN (even though this isn't reflected in their [API docs](https://ripa.bops.services/api-docs/index.html) yet), and we know Uniform currently gets these apps from Planning Portal - but we'll definitely have to test these schema changes as this PR makes more address fields optional / possibly undefined - as to not confuse testing asks, I'm happy to wait to test this until until there's a bit more clarity on outstanding doc template tests, does that make sense? Or go for it now?


jessicamcinchak:
[Add 'find the property on a map instead' option to the first page of 'Find Property' component](https://trello.com/c/CHOmfbM3/855-add-find-the-property-on-a-map-instead-option-to-the-first-page-of-find-property-component)

editor.planx.uk/src/@planx/components/FindProperty/Editor.tsx
```diff
@@ -55,6 +60,44 @@ function FindPropertyComponent(props: Props) {
...
+                formik.setFieldValue(
+                  "allowNewAddresses",
+                  !formik.values.allowNewAddresses
+                );
+              }}
+            >
+              Allow users to plot new addresses without a UPRN
+            </OptionButton>
+          </InputRow>
+          {formik.values.allowNewAddresses ? (
```
jessicamcinchak, line 77: Question: this feature is already behind a toggle in the Editor, do we want to additionally wrap in a Feature Flag for good measure?

Also worth noting: all new Editor props (eg `allowNewAddresses`) are optional, meaning this change should _not_ require a content migration/component "update" when merged. It will default to `ExternalPlanningSiteDialog` same as now.

editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
```diff
@@ -0,0 +1,136 @@
...
+          id="plot-new-address-map"
+          data-testid="map-web-component"
+          zoom={14}
+          drawMode
+          drawType="Point"
+          resetControlImage="trash"
+          showScale
+          showNorthArrow
+          osVectorTilesApiKey={process.env.REACT_APP_ORDNANCE_SURVEY_KEY}
+        />
```
jessicamcinchak, line 101: Here's an interesting question: the starting viewport of this map is currently just the component default (London), but will need to reflect the "team" - so where do we want to store the center coordinate of each team?

Two ideas:
- Expose as fields in the Editor, fallback to component default if not provided. Pros: editors manage!; cons: this might be a more confusing content ask than normal (eg possibility of mixing up x/y, wrong projections, rounding, etc)
- Add to `teams.settings` in the database. Pros: we control data quality; cons: slow feedback loop if changes requested

Any strong opinions on which route? Any other ideas not on my list?

editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
```diff
@@ -0,0 +1,136 @@
...
+          <Input
+            name="newAddress"
+            bordered
+            onChange={(e) => {
+              setSiteDescription(e.target.value);
+            }}
+            value={siteDescription}
+            id="newAddressInput"
+          />
+        </InputLabel>
```
jessicamcinchak, line 132: I imagine a nice v2.0 improvement here would be to additionally make X & Y _inputs_ - allowing you to enter values and the map marker would respond. But I've defaulted to non-inputs for this MVP (it's not a need described in the Trello ticket yet), and only map interaction, to simplify component interactions and back/"change" behavior for now

editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
jessicamcinchak: Also, I'm totally expecting a piece of feedback here is going to be "can the map have search" rather than just zooming/panning interactions alone. 

In my mind, that's a v2.0 feature and the implementation would happen in the web component repo. I think the OS Places API `/find` endpoint would be a good fit for custom free-text search, returning a record with a lat/lng we could use to update viewport https://osdatahub.os.uk/docs/places/technicalSpecification

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -42,27 +45,48 @@ export const FETCH_BLPU_CODES = gql`
...
   let options = new URLSearchParams({
-    entries: "all", // includes historic
+    entries: "all", // includes historic for pre-merger LADs (eg Wycombe etc for Uniform connector mappings)
     geometry: `POINT(${address?.longitude} ${address?.latitude})`,
     geometry_relation: "intersects",
     limit: "100",
   });
   options.append("dataset", "local-authority-district");
-  options.append("dataset", "region");
+  options.append("dataset", "region"); // proxy for Greater London Authority (GLA) boundary
```
jessicamcinchak, line 89: I did check if Digital Land has a postcode boundary file we could additionally query here in cases where the address source is "proposed", but alas they do not. But I don't think this should be a blocker, just would have been nice to have.

editor.planx.uk/src/@planx/components/FindProperty/Editor.tsx
DafyddLlyr: Keeping `<ExternalPlanningSiteDialog />` as an option is a great idea 👍 

I'd love to understand how this change is going to work for Uniform submissions, in particular Southwark who have a different address set up to everybody else.

On this point - I think it's fine to expose this as is? I think it's unlikely someone will toggle this, publish, and not appreciate what it's doing.

editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
DafyddLlyr: I think of this as a nice v2 feature also...!

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -97,16 +121,30 @@ function Component(props: Props) {
     }
   }, [data]);
 
-  if (!address && Boolean(team)) {
+  if (!address && Boolean(team) && page === "os-address") {
     return (
-      <GetAddress
+      <PickOSAddress
```
DafyddLlyr, line 126: Sensible naming 👍

editor.planx.uk/src/@planx/components/FindProperty/model.ts
```diff
@@ -1,35 +1,62 @@
...
+  pao?: string;
+  street?: string;
+  town?: string;
+  postcode?: string;
+  single_line_address?: string;
+  planx_description?: string; // joined via table blpu_codes
+  planx_value?: string; // joined via table blpu_codes
 }
 
+export type NewAddressInputs = {
```
DafyddLlyr, line 51: nit: `UserDefinedAddress`? "New" could mean a few things to me.

Not a strongly held opinion!

editor.planx.uk/src/@planx/components/Send/bops/index.ts
```diff
@@ -249,16 +250,25 @@ export function getBOPSParams(
...
+      passport.data?.["property.localAuthorityDistrict"]?.join(", ");
     site.postcode = address.postcode;
 
     site.latitude = address.latitude;
     site.longitude = address.longitude;
 
+    site.x = address.x;
+    site.y = address.y;
+
+    site.source = address.source; // reflects "os" or "proposed"
```
DafyddLlyr, line 270: Could a string union or an enum be a more useful way of documenting this throughout the codebase?

`type source = "os" | "proposed"`

editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
```diff
@@ -167,19 +167,36 @@ function Question(props: ComponentProps) {
...
+    );
+  } else {
+    const { x, y, title } = props.passport.data?._address;
+    return (
+      <>
+        <dt>Proposed address</dt>
+        <dd>
+          {`${title}`}
+          <br />
+          {`${x.toFixed(5)} Easting (X), ${y.toFixed(5)} Northing (Y)`}
```
DafyddLlyr, line 195: Eastings and Northings are generally presented as whole numbers - I think that's probably what the OS API would return? I think this would apply on the map page also.

DafyddLlyr (changes requested):
Great stuff - apologies it's taken me so long to get here. 

The level of documentation and comments was great to see - you left the code base in a much better way than you found it 🧹 
 
I think there's a issue with what was the old `props.dataFieldArea` value - not quite spotted it in the code but it being set as `undefined`, despite your test flow having a passport value populated as expected. I tested the Bucks LDC flow on the Pizza and the value was set however so maybe it's a config thing somewhere? 
 
image

A few other small comments / ideas throughout but happy to approve once we can get to the bottom of the above issue 👍

editor.planx.uk/src/@planx/components/Send/bops/index.ts
jessicamcinchak: I think I added this comment as a reminder, but yes, `address.source` is already properly typed in the FindProperty/model ! 

Screenshot from 2023-02-02 14-36-43

editor.planx.uk/src/@planx/components/FindProperty/model.ts
jessicamcinchak: very fair point - updating to `ProposedAddress` as that best aligns to the new `_address.source: "os" | "proposed"` reference field I've also introduced. does that seem more clear?

editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  useEffect(() => {
+    const geojsonChangeHandler = ({ detail: geojson }: any) => {
+      if (geojson["EPSG:3857"]?.features && geojson["EPSG:27700"]?.features) {
+        setProposedAddress({
+          // only a single point can be plotted, so get first feature in geojson "FeatureCollection" per projection
+          longitude: geojson["EPSG:3857"].features[0]?.geometry?.coordinates[0],
+          latitude: geojson["EPSG:3857"].features[0]?.geometry?.coordinates[1],
+          x: geojson["EPSG:27700"].features[0]?.geometry?.coordinates[0],
+          y: geojson["EPSG:27700"].features[0]?.geometry?.coordinates[1],
```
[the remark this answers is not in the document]
jessicamcinchak: much nicer, updated!

editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
jessicamcinchak: you're right! i think this actually highlights a projection bug i need to look at in the map repo, because my whole numbers being returned by the geojson event look like X -171, Y 70 for London-ish, whereas I'd expect positive, six-digit whole numbers for Easting/Northing, right?

editor.planx.uk/src/@planx/components/FindProperty/model.ts
DafyddLlyr: Perfect 👍

editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
DafyddLlyr: That's correct - whole numbers, positive for eastings / northings. You could cross reference something like https://gridreferencefinder.com/os.php to sense check. 

Let me know if you want another pair of eyes on this one 👍

jessicamcinchak: projection fix queued up over here! https://github.com/theopensystemslab/map/pull/261

editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
```diff
@@ -0,0 +1,165 @@
...
+        <my-map
+          id="plot-new-address-map"
+          data-testid="map-web-component"
+          zoom={14}
+          drawMode
+          drawType="Point"
+          resetControlImage="trash"
+          showScale
+          showNorthArrow
+          osVectorTilesApiKey={process.env.REACT_APP_ORDNANCE_SURVEY_KEY}
```
DafyddLlyr, line 123: This should be pointing to the proxy - I'm actually not sure this is displaying OS tiles as this env var shouldn't be set. 

Looking into it now...!

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -353,26 +384,28 @@ function GetAddress(props: {
...
+        <ExternalPlanningSiteDialog
+          purpose={DialogPurpose.MissingAddress}
+          teamSettings={props.teamSettings}
+        ></ExternalPlanningSiteDialog>
+      ) : (
+        <Box sx={{ textAlign: "right" }}>
+          <Link
+            component="button"
+            onClick={() => props.setPage("new-address")}
+            disabled={Boolean(sanitizedPostcode) && Boolean(selectedOption)}
```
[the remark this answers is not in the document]
DafyddLlyr: Sounds sensible to me, happy to mark this one as resolved.

editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
jessicamcinchak: updated to point to proxy now - working as expected! think this was just a rebase problem on my side, this PRs been lingering too long :upside_down_face:

editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
jessicamcinchak: :bulb: had a nice realization here: we can simply store a GeoJSON URL from Digital Land representing the boundary we want for a given team. 

This gives us lots of flexibility to decide whether that's the `local-authority-district` boundary for this team (probably the best fit most often) or something else if it's an edge case (eg mergers, etc). As much as I'd love a good reason to use PostGIS column data types and methods, fetching and benefitting from someone else's maintained boundary does the trick here :upside_down_face: 

Example for Braintree: `teams.settings.boundary = "https://www.planning.data.gov.uk/entity/8600093.geojson"`

I simply searched Digital Land's datasets and landed on this page https://www.planning.data.gov.uk/entity/8600093, then clicked "Download geojson" from the bottom map view, and added the URL to Hasura :heavy_check_mark: 

If `teams.settings.boundary` isn't set, we'll simply default to London.

Will add an issue to the onboarding scripts repo to capture this step there too :+1:

jessicamcinchak: Result is a nicely centered map (geojson boundary is light grey for now, could easily be made more bold in future) 
Screenshot from 2023-02-07 18-47-32

jessicamcinchak:
@DafyddLlyr thanks for thorough review & conversation around this one! i think all updated throughout now. i'm still deciding if i'll merge before or after splitting FindProperty :see_no_evil: 

& regarding the one early question from below: this should just be a case of a DrawBoundary component needing to be "updated" - the flow is referencing an old version with different props than the current code, all prod flows have long been updated but not surprised there's a few legacy things floating around on non-live services still ! 
> I think there's a issue with what was the old props.dataFieldArea value - not quite spotted it in the code but it being set as undefined, despite your test flow having a passport value populated as expected. I tested the Bucks LDC flow on the Pizza and the value was set however so maybe it's a config thing somewhere?

editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
DafyddLlyr: This is a great solution - best of both worlds.
 - No maintenance (unless broken by Digital Land - unlikely!)
 - Boundary is richer and more useful than centroid
 - Can be handled pretty easily through onboarding, much more easily than pasting a geojson etc

Nice!

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -103,6 +105,21 @@ function Component(props: Props) {
...
+      props.allowNewAddresses && team?.settings?.boundary
+        ? team.settings?.boundary
+        : null,
+    fetcher,
+    {
+      shouldRetryOnError: true,
+      errorRetryInterval: 500,
+      errorRetryCount: 1,
+    }
+  );
```
DafyddLlyr, line 121: Super simple ✅ 

I've added a comment to this Trello ticket to also populate `team.settings.bounday` whilst we're in the business of updating this table - https://trello.com/c/mDxLZdjO/2216-link-council-logo-at-top-of-planx-pages-to-council-website-homepages

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -119,7 +136,11 @@ function Component(props: Props) {
         setRegions([...new Set(regions)]);
       }
     }
-  }, [data]);
+
+    if (geojson?.type) {
```
DafyddLlyr, line 140: Is there a reason we need to check `type` and not just `geojson`?

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -119,7 +136,11 @@ function Component(props: Props) {
...
+      setBoundary(geojson);
+    }
```
DafyddLlyr, line 142: I'm not as up to speed on React as I'd like to be (or should be!) but I think this would be best as it's own `useEffect()`. If not, it will re-set the boundary each time `data` changes - we just care about changes to `geojson`

editor.planx.uk/src/types.ts
```diff
@@ -36,6 +36,7 @@ export interface TeamSettings {
     url: string;
   };
   supportEmail?: string;
+  boundary?: string;
```
DafyddLlyr, line 39: nit: `boundaryURL`?

DafyddLlyr (approved):

DafyddLlyr:
Thanks for the updates and clear comments here - all looking good! Happy to merge as is, but a few questions / nits about the bounday.

>  i'm still deciding if i'll merge before or after splitting FindProperty 🙈

Yeah this is a bit of a tricky one, and I guess it depends on where the more painful rebase will be... I'd guess merging this first is probably easier?

editor.planx.uk/src/types.ts
jessicamcinchak: i also considered this but settled on plain "boundary" after reviewing other jsonb key names in `teams` - "logo" for example also expects a URL but we keep a simple key name, "homepage" etc !

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
jessicamcinchak: you're right! now it's own `useEffect()`. SWR was doing smart caching here though I think so we didn't actually see repeated requests in the network tab as far as I could tell

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
jessicamcinchak: nope - this is a remnant of an early thought that we might need to check for edge casey stuff like council boundaries that are multipolygons here - but `if (geojson)` should be fine here at this stage :+1:

editor.planx.uk/src/types.ts
DafyddLlyr: That makes perfect sense, thanks for clarifying that 👍

DafyddLlyr (approved):
✅ 

Thanks for much for your patience with all this feedback 😅

PR #1378, opened 2023-01-17
fix: don't hide external portal nodes in the editor if referenced flow has been deleted

jessicamcinchak:


editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Portal.tsx
```diff
@@ -43,9 +43,13 @@ const ExternalPortal: React.FC<any> = React.memo(
...
-    if (!data?.flows_by_pk) {
-      return null;
-    }
+    // If the flow referenced by an external portal has been deleted (eg !data),
+    //   still show a "Corrupted!" node so that editors have a visual cue to "delete".
+    //   The flow schema will still contain a node reference to this portal causing dataMerged to fail
+
+    // if (!data?.flows_by_pk) {
+    //   return null;
+    // }
```
jessicamcinchak, line 52: it felt most intuitive to me to add a comment here and comment out the seemingly-sensible faulty code (so the comment has context), rather than exclusively delete here - but open to other opinions here!

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Portal.tsx
```diff
@@ -43,8 +44,18 @@ const ExternalPortal: React.FC<any> = React.memo(
...
 
+    // If the flow referenced by an external portal has been deleted (eg !data),
+    //   still show a "Corrupted" node so that editors have a visual cue to "delete".
+    //   Until deleted, the flow schema will still contain a node reference to this portal causing dataMerged to fail
     if (!data?.flows_by_pk) {
-      return null;
+      return (
+        <Question
+          hasFailed
+          type="Error"
```
DafyddLlyr, line 54: Didn't know `type="Error"` was a thing - this is great

DafyddLlyr (approved):
This works as expected - thanks for jumping on this and writing up such a detailed and helpful PR - I really appreciate it!

Very happy with this solution but I am interested if this is a new bug or something that's been around for a long time and just not hit?

I think medium term here we ideally also have some sort of handing in `dataMerged()` also so that a `Corrupted` node was also built into the graph. The node wouldn't have any visual component and would just be skipped, but could trigger an error / feedback to let Editors know there's an issue?

Alternatively, updating parent flows when portals are deleted but that's a much tougher nut to crack!

jessicamcinchak:
thanks @DafyddLlyr !

> Very happy with this solution but I am interested if this is a new bug or something that's been around for a long time and just not hit?

I suspect this has always been this way (I'm first to edit this file in 2 years?), but never/rarely hit! I think extra flow methods (copy, rename, etc) in addition to a lot of recent content refactoring (see many new `permitteddevelopment-<something>` flows under osl team) are the new factors here.

> I think medium term here we ideally also have some sort of handing in `dataMerged()` also so that a `Corrupted` node was also built into the graph. The node wouldn't have any visual component and would just be skipped, but could trigger an error / feedback to let Editors know there's an issue?

I originally tried this approach, but throwing an explicit error in `dataMerged()` wasn't actually caught by any of our visual error boundary components and shown in the editor, `/unpublished`, or `/preview` routes as I'd hoped it might be. One to revisit another day - although `dataMerged` is a strong contender for refactoring/archiving all together in my mind, so could go many directions on this!

PR #1379, opened 2023-01-18
fix: make "application_type" dynamic in `getBOPSParams` for non-LDC services 

jessicamcinchak:
Replaces hardcoded instances of `application_type = "lawfulness_certificate"` with flow name when calling this method to generate CSV's for non-LDC services. Additionally adds BOPS unit tests to confirm we set `application_type` correctly in different scenarios. 

Still essentially replacing a hardcoded instance with a different version of hardcoding/relying on a flow name, and will need to be adjusted if we send different types of applications to BOPS in the future, but makes the CSV more usable for other services (eg Prior approvals) in the shortterm.

To test:
- Make a simple flow with some questions and either a Confirmation page or Send component (so S&R is enabled)
- Open `/preview` and go through flow, click "Save and return later" or reach Confirmation page and click "Download your application (.csv)" link, ensure that the row "application_type" reflects your flow's name

This PR originally included another change to the document packet for "send to email", but that's omitted until we've sorted the `decodeURI` bug and I'll plan to adjust in a future PR.

### DIFF
--- a/editor.planx.uk/src/@planx/components/Send/model.ts
+++ b/editor.planx.uk/src/@planx/components/Send/model.ts
@@ -25,7 +25,7 @@ export const USER_ROLES = ["applicant", "agent", "proxy"] as const;
 // See minimum POST schema for /api/v1/planning_applications
 // https://ripa.bops.services/api-docs/index.html
 interface BOPSMinimumPayload {
-  application_type: "lawfulness_certificate";
+  application_type: "lawfulness_certificate" | string;
   site: {
     uprn: string;
     address_1: string;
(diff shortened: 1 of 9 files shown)


editor.planx.uk/src/@planx/components/Send/model.ts
```diff
@@ -25,7 +25,7 @@ export const USER_ROLES = ["applicant", "agent", "proxy"] as const;
 // See minimum POST schema for /api/v1/planning_applications
 // https://ripa.bops.services/api-docs/index.html
 interface BOPSMinimumPayload {
-  application_type: "lawfulness_certificate";
+  application_type: "lawfulness_certificate" | string;
```
jessicamcinchak, line 28: i can never recall if there is a nicer way to indicate a "default" value in a type interface? this syntax feels clumsy.

DafyddLlyr (approved):
Nice and simple ✅ 

To me, this feels like another good reason to think about changing how we handle send and maybe moving the BOPS Payload construction to the backend.

If we hit a `submit/:session_id` route we could...
 - Based on session.flow_id, get associated templates
 - Based on session.flow_id, get flow name and data

PR #1380, opened 2023-01-18
fix: Reinstante port 4000 on API

DafyddLlyr:
Undoes change made here - https://github.com/theopensystemslab/planx-new/pull/1323/files#diff-9fee14c6790003e5ec78ba098a0e86794e0aeaac6a5ca302dc10ad9ae9d2aae8

Currently, API endpoint which end in `next()` seem to be throwing an instant 500 since the most recent deploy.

This seems to be the only significant change in the recent deploy (#1375

### DIFF
--- a/api.planx.uk/Dockerfile
+++ b/api.planx.uk/Dockerfile
@@ -14,6 +14,8 @@ RUN pnpm prune --production
 
 COPY . .
 
+EXPOSE 4000
+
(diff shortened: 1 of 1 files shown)


jessicamcinchak (approved):

DafyddLlyr:
GitHub actions cancelled, did not wait for anything beyond API tests to pass here.

PR #1332, opened 2022-12-14
chore: Delegate error handling to `reportError` in the API

ghost:
Similar to #1330   this PR delegates the responsibility of environment specific error handling but, this time, in the API.

PR #1381, opened 2023-01-18
fix: Revert reportError

DafyddLlyr:


api.planx.uk/server.ts
```diff
@@ -640,14 +640,25 @@ app.post("/webhooks/hasura/send-slack-notification", sendSlackNotification);
...
+  } catch (error) {
+    next(error)
+  }
+})
+
 const errorHandler: ErrorRequestHandler = (errorObject, _req, res, _next) => {
   const { status = 500, message = "Something went wrong" } = (() => {
-    if (errorObject.error) {
-      reportError(errorObject.error);
+    if (errorObject instanceof Error && airbrake) {
```
DafyddLlyr, line 653: If you look at #1332 there are actually some changes here I needed to make to hit this `if` block. `errorObject.error` is undefined, so this wasn't getting hit.

I'll also look at some tests for this tomorrow to make sure we're calling Airbrake when we think we are.

DafyddLlyr: There is some inconsistency with how we're currently handing errors I think - sometimes it's `next(error)` and other times we use `next({error, message: "something"})`.

DafyddLlyr:
Currently testing if I can send a submission to Uniform via the Pizza

jessicamcinchak (approved):
endpoints logging errors as expected/described on this pizza - thank you for investigating and identifying !

PR #1385, opened 2023-01-19
chore: Reinstate document review

DafyddLlyr:
This undoes the temporary changes made in #1383 and bumps the document review library to v1.1.2 which should resolve the issues previously described.


jessicamcinchak (approved):
thanks!

PR #1386, opened 2023-01-19
Fix: replace cypress tests

ghost:
Also includes some updates across the board as we aim to find a consistent testing style (feel free to disagree with my suggestions).


DafyddLlyr (approved):
Thank you for this one - Playwright is looking better and better. Going to make sure I read up on the docs and get some tests written myself shortly also!

PR #1387, opened 2023-01-19
chore: reduce e2e github action to run on "review_requested" only

jessicamcinchak:
was previously triggering duplicate runs! context and outcome of quick vote here https://opensystemslab.slack.com/archives/C01E3AC0C03/p1674053666637189

### DIFF
--- a/.github/workflows/end-to-end-tests.yml
+++ b/.github/workflows/end-to-end-tests.yml
@@ -8,7 +8,6 @@ on:
       - main
     types:
       - ready_for_review
-      - review_requested
   workflow_dispatch:
   schedule:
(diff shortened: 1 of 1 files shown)


jessicamcinchak:
actually, contrary to our quick vote, I realized just here that only "request_review" will make more sense than "ready_for_review" - because "ready_for_review" is only triggered when first opening as a draft and then converting to "ready", otherwise you'd open a regular PR, have to convert to draft, then convert to ready to trigger e2e. 

whereas I expect we actually _always_ request review, and it's easier to request review many times throughout the open lifespan of a PR if necessary to re-trigger action.

PR #1388, opened 2023-01-20
fix: replace map.html with application.html in zip sent by email

jessicamcinchak:
was previously only displaying the map on the HTML in the email packet, this additionally populates the "Data" column of the template.

together with #1379, ensures that folks who are testing Prior Approval will have an email packet that matches our demo from earlier this week. 

to test: 
- edit pizza db `teams.settings.sendToEmail` to add yourself
- make a simple flow with Draw, FileUpload, "Send to planning office" 
- go through flow: confirm you receive email, can download files, and zip packet contains expected files

### DIFF
--- a/api.planx.uk/send/email.ts
+++ b/api.planx.uk/send/email.ts
@@ -121,23 +121,26 @@ const downloadApplicationFiles = async(req: Request, res: Response, next: NextFu
         deleteFile(csvPath);
       }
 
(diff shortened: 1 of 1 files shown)


DafyddLlyr (approved):
Once I managed to actually send myself an email and not somebody else, this worked a charm 😅

PR #1390, opened 2023-01-20
fix: don't sync teams.settings data from production, use default column value

jessicamcinchak:
`teams.settings.sendToEmail` is now populated with council emails on production. We want to be able to freely test send to email locally and on pizzas without fear of accidently emailing someone, therefore we do want to include this value in our data sync. 

Since `teams.settings` column has a default value for `externalPlanningSite`, I think we can just fully omit syncing this column for right now. I don't know any clever GraphQL query syntax for omitting a jsonb key, only for selecting a specific key. By omitting settings altogether and only using the default column value, we'll lose `sendToEmail` (good!) and occassionally `homepage` properties (which means logos won't link to council websites on pizzas), but I think that's an okay trade-off right now!

Ideas/next steps for more flexibility are:
- add a new text column to `teams` (or flows even, but I don't think there's actually a use-case for this yet) to store `sendToEmail` rather than as a jsonb property on `settings`, so it's easier to omit from the data sync independently
- move `sendToEmail` into team or flow level settings in the Editor

Also worth noting: 
- It'd be most ideal to be able to sync `sendToEmail` from production to staging (more & more POs are testing on .dev), but _not_ to docker environments (local and pizza), but this data sync script is shared between all envs right now

To test:
- Open Hasura `teams` table and confirm settings column is only populated with default `externalPlanningSite` key


DafyddLlyr (approved):
Thanks for the PR description - super helpful. For now, I think this is a solid solution as the tradeoffs are small and well understood.

As far as I know, there isn't a native GraphQL way of doing an omit or a filter on a JSON. A Hasura blog posts suggests using a Postgres view to achieve something similar - https://hasura.io/blog/postgres-json-and-jsonb-type-support-on-graphql-41f586e47536/

I think making `sendToEmail` a field feels perfectly reasonable to me for a few reasons - 
 - It actually makes it safer to return `settings` to the client without exposing `sendToEmail` (the combination of email + sessionID can allow access to download files)
 - It's pretty quick/easy and there's no UI in team/settings we need to change in the editor
 - We've not had pushback at all (to the best of my knowledge) on a single email per team
 - It would make an eventual refactor into a sentToEmail per flow simpler, if the requirement ever comes up

jessicamcinchak:
thanks @DafyddLlyr ! i agree and can pick up the new column migration early next week. is it OK with you if I merge this to main in the meantime (recognizing e2e tests have not run here, and will fail when merging to main - but this one-liner shouldn't effect that upcoming fix!)

jessicamcinchak:
[Migrate `teams.settings.sendToEmail` to own text column so we can sync production data to other envs safely](https://trello.com/c/TBSYqBOS/2255-migrate-teamssettingssendtoemail-to-own-text-column-so-we-can-sync-production-data-to-other-envs-safely)

DafyddLlyr:
Yep totally @jessicamcinchak - should have been clearer there, that was certainly my plan also! Please merge this 😄

PR #1395, opened 2023-01-20
fix: replace cypress tests

builditben:
Take 2 - previously failing since some assumptions made about db state but now fixed.


jessicamcinchak (approved):
thanks for quick fix!

PR #1396, opened 2023-01-22
fix: Correctly rename queries for ReconciliationRequests

DafyddLlyr:
**Problem**
Since merging https://github.com/theopensystemslab/planx-new/pull/1325 the `sanitiseReconciliationRequests` operation has been failing. This can be seen in the Hasura logs on `.dev` and Slack errors in the `#planx-notifications-internal` channel.

**Cause**
This is caused by refactoring `sanitiseReconciliationRequests` to delete instead of sanitising, without properly updating how the request was being destructured.

**Solution**
Update queries and variables to better reflect what's happening here - `sanitiseReconciliationRequests` consistently becomes `deleteReconciliationRequests`, and queries are updated accordingly.

### DIFF
--- a/api.planx.uk/webhooks/sanitiseApplicationData/operations.ts
+++ b/api.planx.uk/webhooks/sanitiseApplicationData/operations.ts
@@ -20,7 +20,7 @@ export const getRetentionPeriod = () => subMonths(new Date(), RETENTION_PERIOD_M
   sanitiseSessionBackups,
   sanitiseUniformApplications,
   sanitiseBOPSApplications,
-  sanitiseReconciliationRequests,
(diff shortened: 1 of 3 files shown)


api.planx.uk/webhooks/sanitiseApplicationData/operations.ts
```diff
@@ -181,7 +181,7 @@ export const sanitiseReconciliationRequests: Operation = async () => {
       } 
     }
   `;
-  const { update_reconciliation_requests: {
+  const { delete_reconciliation_requests: {
```
DafyddLlyr, line 184: Here's the line I missed previously - this did not correspond with the query on line 173

PR #1399, opened 2023-01-24
fix: Data sanitation cronjob environment

DafyddLlyr:
## What does this PR do?
 - Correctly set the environment for Slack notifications
 - Drop cronjob retries - a single failure and notification in Slack should be fine

**Comments**
I think a better way forward here in the short / medium term is probably to meaningfully set `process.env.NODE_ENV` to one of Production / Staging / Pizza / Development - I think this has come up before. I've added it to the agenda for the dev call tomorrow 👍

### DIFF
--- a/api.planx.uk/helpers.test.ts
+++ b/api.planx.uk/helpers.test.ts
@@ -1,0 +1,34 @@
+import { getEnvironment } from "./helpers";
+
+describe("getEnvironment function", () => {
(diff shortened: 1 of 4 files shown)


DafyddLlyr:
Closed in favour of https://github.com/theopensystemslab/planx-new/pull/1411

I'll tackle setting `NODE_ENV` correctly once as opposed to merging this and re-working.

PR #1401, opened 2023-01-25
feat: split off new `PropertyInformation` component from `FindProperty` and add option to override "property.type"

jessicamcinchak:
These changes cover steps 1 & 2 in allowing users to change their `"property.type"` if they think it's wrong. I'll address the final 3rd step about content migration in a separate PR.

1) Split `FindProperty`, previously a "two-page" component, into two individual components:
- `FindProperty` is responsible for entering a postcode, selecting an OS address, mapping OS classification code to planx BLPU code, fetching supplementary address boundaries from Digital Land (eg `property.localAuthorityDistrict` and `property.region`), and rendering the `ExternalPlanningSiteDialog`
   - Next, we'll introduce a new "second-page" option to plot a new address (see #1377 for the full propery-related code review experience :sparkles:)
- `PropertyInformation` is reponsible for displaying address data currently in state and handling "override" logic via a "change" link for specific pieces of data. Now that we have a "change" link, the "Report an inaccuracy" input is redundant and removed from this component

2) Create a way for a user to "override" an answer, aka answer a question again that was previously auto-answered (hidden)
- Adds method `overrideAnswer` to the PreviewStore
- Adjusts styling of `PropertyInformation` to display a "change" link if an Editor toggles on the option to show option to override
- In the passport, the "original" property type fetched from OS will always be stored under the `_address` key, while the user-change-able property type will be reflected in `"property.type"`; no new passport variables are added here. 

**Open questions:** 
- How should the BOPS & Uniform field mappings be adjusted to communicate the property type if it has been changed? Currently, the `"property.type"` field is sent to both - which reflects the OS value if the user hasn't "changed" and the user-defined value if they have.

**Two testing asks:** 
- Add `PropertyInformation` node to a flow directly after FindProperty with override option **toggled off**. Confirm that the flow behaves with full feature parity to how a production flow with a single FindProperty node behaves now. Meaning: same passport data structure, error handling, back button behavior & initial data population, presentation and "change" behavior in Review, etc

- Use `PropertyInformation` with the override option **toggled on** here https://1401.planx.pizza/buckinghamshire/search (this is realistic content mocked up by Alastair). "Change" your property type once or many times and ensure the passport stays in sync. Go "back" between nodes and "change" from Review and confirm behavior is consistent with the first feature parity test.
Screenshot from 2023-01-31 10-31-32


jessicamcinchak:
[Split up FindProperty component](https://trello.com/c/p8wggZ46/2252-split-up-findproperty-component)

jessicamcinchak:
[[Build] ways to allow users to override the property type if they think its wrong](https://trello.com/c/9ig1C8j9/1958-build-ways-to-allow-users-to-override-the-property-type-if-they-think-its-wrong)

editor.planx.uk/src/@planx/components/FindProperty/Public/FindProperty.stories.tsx
```diff
@@ -1,7 +1,8 @@
 import { Meta, Story } from "@storybook/react/types-6-0";
 import React from "react";
 
-import FindProperty, { PropertyInformation } from "./";
+import PropertyInformation from "../../PropertyInformation/Public";
+import FindProperty from "./";
```
jessicamcinchak, line 5: I admittedly took a shortcut here by continuing to combine these components in the same "story" - will revisit/clean up in the future if that's okay (lots of bits of storybook are out of date, more FindProperty updates in the pipeline, etc)

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -195,15 +102,7 @@ const AutocompleteWrapper = styled(Box)(({ theme }) => ({
...
-function GetAddress(props: {
-  setAddress: React.Dispatch<React.SetStateAction<SiteAddress | undefined>>;
-  title?: string;
-  description?: string;
-  initialPostcode?: string;
-  initialSelectedAddress?: Option;
-  teamSettings?: TeamSettings;
-  id?: string;
-}) {
+function PickOsAddress(props: PickOsAddressProps) {
```
jessicamcinchak, line 105: this re-naming aligns to future FindProperty "pages" over in #1377

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -300,153 +201,50 @@ function GetAddress(props: {
...
-          <Input
-            required
-            bordered
-            name="postcode"
-            id="postcode-input"
-            value={postcode || ""}
-            errorMessage={
+    <AutocompleteWrapper>
+      <InputLabel label="Postcode" htmlFor="postcode-input">
+        <Input
```
jessicamcinchak, line 206: future extra credit: replace this input with our `<postcode-search />` web component [over here](https://github.com/theopensystemslab/map/tree/main/src/components/postcode-search), which will further simplify this file and let us drop `postcode` package dependcy from editor. I expect the web component may need some styling overrides first though and double-checks that we have feature parity with especially change/keyUp/blur events and error handling.

editor.planx.uk/src/@planx/components/PropertyInformation/Public.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+  );
+
+  expect(screen.getByTestId("error-summary-invalid-graph")).toBeInTheDocument();
+});
+
+test.skip("renders correctly", async () => {
+  const handleSubmit = jest.fn();
+
+  const useStore = jest.fn();
+  useStore.mockReturnValue([
```
jessicamcinchak, line 51: anyone have guidance/experience on best way to mock `useStore`? i fiddled with a few variations of this that aren't quite working how I expect them to. we do something similar in other components by mocking "previously submitted data", but in this case I want to check state on first render, not after coming "back".

jessicamcinchak:
[Implement change link and passport handling](https://trello.com/c/Lfve5XEX/2254-implement-change-link-and-passport-handling)

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -596,6 +598,42 @@ export const previewStore = (
...
+        overrideNodeId = nodeId;
+        return;
+      }
+    });
+
+    if (overrideNodeId) {
+      // Travel backwards to the "override" nodeId to manually re-answer this question, therefore re-setting the passport value onSubmit
+      changeAnswer(overrideNodeId);
+    }
+  },
```
jessicamcinchak, line 636: felt a little clumsy coming up with clear names here - very open to suggestions for more clear ways to describe "originalNodeId" (this may also have to become plural, still testing repeated "change" clicks to see if there's any edge cases) & "overrideNodeId" !

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -819,7 +857,11 @@ export const removeNodesDependentOnPassport = (
   breadcrumbsWithoutPassportData: Store.breadcrumbs;
   removedNodeIds: string[];
 } => {
-  const DEPENDENT_TYPES = [TYPES.PlanningConstraints, TYPES.DrawBoundary];
+  const DEPENDENT_TYPES = [
+    TYPES.PlanningConstraints,
+    TYPES.DrawBoundary,
+    TYPES.PropertyInformation,
+  ];
```
jessicamcinchak, line 864: This says that "PropertyInformation" should be cleared & shown fresh anytime "FindProperty" changes, rather than retain previously submitted data like other components do when navigating backwards/forwards.

editor.planx.uk/src/@planx/components/PropertyInformation/Editor.tsx
```diff
@@ -0,0 +1,85 @@
...
+            selected={formik.values.showPropertyTypeOverride}
+            onClick={() => {
+              formik.setFieldValue(
+                "showPropertyTypeOverride",
+                !formik.values.showPropertyTypeOverride
+              );
+            }}
+          >
+            Show users a "change" link to override the property type
+          </OptionButton>
```
jessicamcinchak, line 68: In practice, I think we'll always use this component with this option toggled "on" but I like the idea of introducing it behind a flag here because it gives us a bit more flexibility in deciding our content migration plan. 

We could do the "simple" migration that inserts it after every FindProperty with the option off to satisfy initial content-parity, and in a separate migration we could toggle it on and insert the whole tree of property type questions in-between.

DafyddLlyr (approved):
This is very solid work - changes are clear and meaningful, and it's a pretty straightforward change all things considered. Tests explain what's happening, and the new store method is well documented 👍 

I feel like I've just left a ton of nitpick comments sorry - as always, please consider / disregard / resolve as you see fit!

I've followed the testing asks in the PR description with no issues, and also had a play around - everything working as expected.

I think there's really one outstanding issue to be addressed here - 

> How should the BOPS & Uniform field mappings be adjusted to communicate the property type if it has been changed? Currently, the "property.type" field is sent to both - which reflects the OS value if the user hasn't "changed" and the user-defined value if they have.

A few thoughts on this one...
 - There's a lot of change in this PR already. If we can be careful about merging and come up with a plan for how we get this feature to `main` we can handle this as a small separate PR.
 - It makes sense to me that this responsibility lives in `overrideAnswer()` so it's always handled in a consistent way.

Lets's talk this through, but here's one way I could see this working...
 
A breadcrumb has `auto: boolean` currently. We use this for a number of purposes throughout the code base. The value of this is that it currently lets us work out how a question was answered - by the user, or by data.
 
The concept of "override" means that a question can now be answered in one of three meaningful ways - 
 - Manually
 - Automatically
 - Automatically, but overridden

We could change `auto: boolean` to `method: "auto" | "manual" | "override"`. We could make `"manual"` the default assumption here. We could then use this to structure the data meaningfully in our Passport, and on consumer payloads.

I'll have a think about how best to structure the Passport. My first inclination would be that the user submitted overwrite value should stay as `passport.data["property.type"][0]` but we could check if `passport.auto["property.type"][0]` existed if we wanted to structure a payload differently. In future, the passport could be a class with a `get()` method to abstract some of this.
 
**Uniform**
 - We don't send `property.type` to Uniform so there's no way of us communicating that change in a direct and meaningful way. Of all the fields we send to them, I don't think any can be affected by this change (e.g. if a user changes property type, no subsequent answer will be different on the XML as it mainly concerns personal contact details, type of application, etc).
 - I imagine they get this data from the UPRN somehow, so there's a chance applicants will answer one way on PlanX, and Uniform will log another.
 - For now, I'd be inclined to do nothing on the Uniform front. If it becomes an issue we can append the `description` field and hope that Validators respond to it.
   
**BoPS** 
 - If they tell us how they want it, we could use breadcrumbs / passport to package it up

I'll keep thinking on this one, the above might be a bit unstructured sorry!

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
jessicamcinchak: went for inline comments here & added more explicit types!

editor.planx.uk/src/@planx/components/PropertyInformation/Public.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+});
+
+test("hides 'change' link if the editor has not selected the override option", async () => {
+  const handleSubmit = jest.fn();
+  const overrideAnswer = jest.fn();
+
+  const { user } = setup(
+    <MockedProvider mocks={teamMock} addTypename={false}>
+      <Presentational
+        title="About the property"
```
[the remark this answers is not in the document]
jessicamcinchak: much cleaner, now updated!

editor.planx.uk/src/@planx/components/PropertyInformation/Public.tsx
```diff
@@ -0,0 +1,255 @@
...
+              >
+                Change
+                <span style={visuallyHidden}>
+                  your {heading || "this value"}
+                </span>
+              </Link>
+            </Box>
+          ) : (
+            <Box component="dd">
+              {/** ensure there's always a third column to not break styling, even if "Change" is hidden */}
```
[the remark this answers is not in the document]
jessicamcinchak: good catch, properly updated this here :+1:

editor.planx.uk/src/@planx/components/PropertyInformation/Public.tsx
```diff
@@ -0,0 +1,255 @@
...
+            <Box component="dd">
+              <Link
+                component="button"
+                onClick={(event) => {
+                  event.stopPropagation();
+                  // Specify the passport key (eg data.fn, data.val) that should be overwritten
+                  overrideAnswer(fn);
+                }}
+              >
+                Change
```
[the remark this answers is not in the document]
jessicamcinchak: I had spotted this too but not dug into, it's a weird one - I explicitly set `style={{ font: "inherit" }}` on the <Link /> here which isn't an explanation of why in the first place, but does visually fix this!

jessicamcinchak:
@DafyddLlyr thanks for thoughtful review of this one and patience as I've been working through feedback across related PRs.

I think it's properly "ready" now. Pizza has rebuilt from scratch.

Here's a few final thoughts on broader questions raised: 
- We're planning to merge this with the "override" feature "off" at first, so let's _not_ address BOPS & Uniform payload changes here. We can capture this separate follow-up PR, and the passport should still always have _a_ value for "property.type" whether or not it's been overwritten in the meantime, meaning no schemas should break based on the changes here.
> We could change auto: boolean to method: "auto" | "manual" | "override". We could make "manual" the default assumption here. We could then use this to structure the data meaningfully in our Passport, and on consumer payloads.
- I feel very hesitant to change "auto" to an enum at this time because this boolean is relied on in _so_ many store methods, conditional preview displays (eg show/hide from Review, Result, etc), and we have exactly one still-quite-bespoke case of "override" right now. And I'm thinking about auto-answering & overriding as possibly simultaneous actions we want to record, rather than always distinct ones which an enum would necessitate. 

BUT, this thinking absolutely raised the important note that we should have some type of initial representation on the breadcrumb "userData" to denote overrides. I've settled on an optional "override" key which is an object: the key will be the `fn` of the passport variable we are overriding, and the value will be the original value pre-override. This will be set on the "original" node (which previously would have had this `fn` in it's "data"). The passport will always reflect the post-override value of `fn`, derived from the one (or many in "property.type" `s case) nodes that allow the user to set it again. 

You can inspect this in-action on this pizza flow! Please do shout if feels too unclear/confusing. https://1401.planx.pizza/buckinghamshire/search

jessicamcinchak:
@DafyddLlyr thanks for final review! 

> One final thing (sorry!) I've noticed that's specific to `FindProperty` is that we're capturing some of these values under `_address` in the Passport also which never get touched after a change -

Yes, exactly -- I'm very aware of these `_address` keys, they are why I hadn't thought of adding the "override" data to the breadcrumb in the first place! Like you point out, FindProperty uniquely already has a way of handling/keeping the original value (I think this will become even more explicit when we add `_address.source: "os"` via #1377), but this obviously wouldn't have scaled to other future component types which the "override" userData should do!

PR #1400, opened 2023-01-24
fix: add new column `teams.submission_email` so we can safely sync `teams.settings` across environments

jessicamcinchak:
Better solution to quick bug fix last week #1390 &larr; full context in this PR

When merged to production, I'll manually migrate the handful of values in `teams` table to populate `submission_email` and remove `settings.sendToEmail` so that prior approval/send to email testing is uninterrupted.

PR #1402, opened 2023-01-25
Production deploy

jessicamcinchak:
reminder #1400 will require a quick content update once live, I'll plan to take care of it!


DafyddLlyr (approved):
✅

PR #1403, opened 2023-01-25
fix: Ensure dev/prod logger is properly configured in editor

ghost:
This PR adds tests for the env-specific logger function and updates the behavior to match the API.

Also, the env variable `DEBUG` has been replaced with `SUPPRESS_LOGS` and the logic inverted since this is less surprising and still enables the option of having "quiet" runs.


DafyddLlyr (changes requested):
Thank you for picking this up.

Could you also please add the new variable to the `.example.env` file?

DafyddLlyr (approved):

PR #1409, opened 2023-01-30
ci: enable deployments on a schedule

gunar:
Enables the scheduled automatic deployments, as per discussions before the end of the year.


gunar:
First, open a PR that extends the current production deployment actioin so that:
- if deployment failed, roll back
- healthcheck, if fails, roll back

PR #1415, opened 2023-02-01
test: Increase API test coverage

DafyddLlyr:
Something I picked up in the background whilst watching the show & tell this morning.

The commits might make the changes in `api.planx.uk/s3/s3.test.ts` more easy to understand.

### DIFF
--- a/api.planx.uk/s3/controller.ts
+++ b/api.planx.uk/s3/controller.ts
@@ -52,10 +52,6 @@ export const publicDownloadController = async (
 ) => {
   const filePath = buildFilePath(req.params.fileKey, req.params.fileName);
 
-  if (!filePath) {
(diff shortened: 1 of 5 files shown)


api.planx.uk/s3/controller.ts
```diff
@@ -52,10 +52,6 @@ export const publicDownloadController = async (
 ) => {
   const filePath = buildFilePath(req.params.fileKey, req.params.fileName);
 
-  if (!filePath) {
-    return next({ status: 404, message: "file not found" });
-  }
```
DafyddLlyr, line 57: It's not possible to hit this block - if we don't have `req.params.fileKey` and `req.params.fileName`, we don't hit this API route.

PR #1416, opened 2023-02-01
feat: Share domain code between projects

ghost:
Sharing files between projects is tricky without a proper monorepo setup (workspaces, etc). 

This change is to introduce a (hopefully simple) way of sharing isomorphic code between projects. It allows us to gather all of our core domain logic into modules that describe operations in domain language (i.e. "duplicateFlow") and hide the implementation details of those operations from the caller (mainly graphql operations and data transformations).

There were a few necessary changes to make this work well. While this is not the ideal long-term solution, it should be a lightweight enough substitute for a wholesale monorepo setup in the short-to-mid term.

### DIFF
--- a/api.planx.uk/package.json
+++ b/api.planx.uk/package.json
@@ -95,7 +95,11 @@
     "peerDependencyRules": {
       "allowedVersions": {
         "nock": "13"
-      }
+      },
+      "allowAny": [
+        "react",
+        "react-dom"
+      ]
     },
     "overrides": {
       "passport-oauth2@<1.6.1": ">=1.6.1",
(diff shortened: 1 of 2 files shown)


DafyddLlyr (approved):
This is a neat solution!

In terms of the workflow here, can I confirm that this is how it would work...
 - I make a change in `planx-client` which I'd like to pick up elsewhere
 - I run `pnpm distribute` from `/planx-client`
 - That's it! I don't need another `pnpm i` in the API or Editor as the filename / version number hasn't changed, just the `.tgz` replaced.
 
If the above is correct, would it make sense to run `pnpm distribute {dir}` as a postinstall on a `pnpm i` within that directory?

DafyddLlyr:
Ah yeah - you're right.

 - Once it gets to building the app, the GitHub workflows will always sync the client
 - But if we postinstall, we can run one version in API and another in Editor locally
 - Therefore, postinall = bad idea!
 
 Thanks for clearing that up - helpful to get a clear picture in my head.

PR #1422, opened 2023-02-02
chore: mass prettify

ghost:
Not sure if there's a better way to go about this but it'd be good to get all the code into a standard format. One big PR might be too much so maybe worth splitting this out into a few (boring) PRs.

I've run `npx prettier -w {target}` across the api and editor folders.

### DIFF
--- a/.github/workflows/end-to-end-tests.yml
+++ b/.github/workflows/end-to-end-tests.yml
@@ -4,8 +4,7 @@ on:
     branches:
       - main
   workflow_run:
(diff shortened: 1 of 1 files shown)


DafyddLlyr:
Mass prettify is the way to go here - I trust you're not sneaking anything in 😉 

In terms of maintaining this, can we also run this as a post-commit hook (or similar)?

DafyddLlyr (approved):

PR #1424, opened 2023-02-03
feat: Session XML endpoint

DafyddLlyr:
Wish I'd done this sooner...!

This takes away the pain and time of getting a Uniform submission XML from Hasura, formatting it, and checking it out.

We could add an explicit validation check here, but the browser will generally show this if there's an issue. If it's something we hit often we can add it in to get errors which are more nicely formatted. Theoretically, as we're using a library to build the XML this shouldn't really be an issue.

Example url (must be logged in) - https://api.1424.planx.pizza/admin/session/dummy_submission_ref_123/xml

### DIFF
--- a/api.planx.uk/admin/session/oneAppXML.test.ts
+++ b/api.planx.uk/admin/session/oneAppXML.test.ts
@@ -5,7 +5,7 @@ import { authHeader } from "../../tests/mockJWT";
 
 const endpoint = (strings: TemplateStringsArray) => `/admin/session/${strings[0]}/xml`;
 
-describe("Download feedback CSV endpoint", () => {
+describe("OneApp XML endpoint", () => {
   beforeEach(() => {
     queryMock.mockQuery({
       name: "GetUniformApplicationBySessionID",
--- a/api.planx.uk/admin/session/oneAppXML.test.ts
+++ b/api.planx.uk/admin/session/oneAppXML.test.ts
@@ -1,0 +1,62 @@
+import supertest from "supertest";
+import app from "../../server";
+import { queryMock } from "../../tests/graphqlQueryMock";
+import { authHeader } from "../../tests/mockJWT";
+
+const endpoint = (strings: TemplateStringsArray) => `/admin/session/${strings[0]}/xml`;
+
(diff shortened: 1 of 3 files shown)


api.planx.uk/admin/session/oneAppXML.test.ts
```diff
@@ -0,0 +1,62 @@
+import supertest from "supertest";
+import app from "../../server";
+import { queryMock } from "../../tests/graphqlQueryMock";
+import { authHeader } from "../../tests/mockJWT";
+
+const endpoint = (strings: TemplateStringsArray) => `/admin/session/${strings[0]}/xml`;
+
+describe("Download feedback CSV endpoint", () => {
```
jessicamcinchak, line 8: nit: this description is still copied from the other endpoint!

api.planx.uk/admin/session/oneAppXML.ts
```diff
@@ -0,0 +1,42 @@
...
+    res.set("content-type", "text/xml")
+    return res.send(xml)
+  } catch (error) {
+    return next({ message: "Failed to get OneApp XML: " + (error as Error).message });
+  }
+};
+
+const fetchSessionXML = async (sessionId: string) => {
+  try {
+    const client = adminGraphQLClient;
```
jessicamcinchak, line 21: nit: I've been slowing migrating to and appreciating this pattern which feels a bit cleaner to me: `import { adminGraphQLClient as client } from "../../hasura"`

jessicamcinchak (approved):
super handy, thank you! also a nice reminder that we're in a good spot lately to very quickly spin up little utility endpoints like this that add a lot of value :raised_hands:

api.planx.uk/admin/session/oneAppXML.test.ts
DafyddLlyr: 🤦

api.planx.uk/admin/session/oneAppXML.ts
DafyddLlyr: That's much neater - thanks.

PR #1425, opened 2023-02-03
feat: create a content migration endpoint to add new `PropertyInformation` component to flows in correct position

jessicamcinchak:
depends on #1401 for testing, sorry it makes for such a large diff! please just focus on **api.planx.uk/** changes here

i've noted some known outstanding to-do's below, but would appreciate a preliminary code review here to sense-check the direction this is going! 

todo: 
- make this a batch operation, rather than single flow
- more meaningful response & error handling - eg if flow doesn't have FindProperty (silently skips for now), etc
- figure out why API tests are passing locally, but failing on CI and linking my query mock to the BOPS test file path??

testing: 
- create or open an existing flow that has a FindProperty component, publish it
- run GET `/admin/migrate/splitFindProperty/:flowId` in a browser tab (must be signed into editor)
- reload your flow in the editor
  - confirm there is now a PropertyInformation node displaying directly after FindProperty
  - click "Check for changes to publish" - it should not pick up any changes
  - add/move/remove a node from the graph, confirm you can keep editing the flow without error (this will confirm it's okay to skip adding an associated operation for the migration) 
- edge casey stuff: make sure this works regardless of _where_ FindProperty is positioned in the graph (eg for new prior approval services it's on a branch); make sure it works with and without a published version, etc


api.planx.uk/admin/migrations/splitFindProperty.ts
```diff
@@ -0,0 +1,165 @@
...
+ * 
+ * Considerations:
+ *   - nodeIds for the new PlanningInformation must be unique among flows to prevent clone behavior if used as portals
+ *   - a flow and its' currently published version need to share a nodeId so that it's not picked up in the publishing diff
+ *   - we assume one FindProperty per flow
+ *   - we assume no flow has PropertyInformation yet
+ *   - we ignore external portal nodes when scanning a flow schema (external portals are flows and can be migrated independently)
+ *   - we do not insert a corresponding operation, as "migrations" don't have actors, shouldn't need to be rolled-back, etc
+ * 
+ */
```
jessicamcinchak, line 24: since migrations are essentially one-off and fairly bespoke jobs, I figured I'd leave context & assumptions directly in the file itself. open to suggestions for this if other documentation preferences!

api.planx.uk/admin/migrations/splitFindProperty.ts
```diff
@@ -0,0 +1,165 @@
...
+    "description": "This is the information we currently have about the property",
+    "showPropertyTypeOverride": false
+  }
+};
+
+// aligns with editor.planx.uk/src/@planx/graph
+const uniqueId = customAlphabet(en)(
+  "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789",
+  10
+);
```
jessicamcinchak, line 41: i'm not fully caught up on `planx-client` work yet, but I think this may be a good future candidate?? it's repeated a couple times in the editor, and now I've added these dependencies to the API - which feels a like a lot of added overhead for a migration we'll run once, but node id generation feels important to do consistently.

api.planx.uk/admin/migrations/splitFindProperty.test.ts
```diff
@@ -0,0 +1,117 @@
...
+            data: flowDataBeforeMigration
+          }
+        ]
+      }
+    },
+  });
+
+  queryMock.mockQuery({
+    name: "UpdateFlow",
+    matchOnVariables: false, // TODO mock uniqueId() and set to 'true'
```
jessicamcinchak, line 74: the "success" test is really only meaningful if I can match on variables, but that's tricky with a randomly generated new nodeId! 

i was having a hard time figuring out how to mock the uniqueId() caller - any tips?

jessicamcinchak:
[Migrate existing flows to have findProperty and FindPropertyReview](https://trello.com/c/wkj7gb1q/2253-migrate-existing-flows-to-have-findproperty-and-findpropertyreview)

jessicamcinchak:
[[Build] ways to allow users to override the property type if they think its wrong](https://trello.com/c/9ig1C8j9/1958-build-ways-to-allow-users-to-override-the-property-type-if-they-think-its-wrong)

DafyddLlyr:
> depends on https://github.com/theopensystemslab/planx-new/pull/1401 for testing, sorry it makes for such a large diff! please just focus on api.planx.uk/ changes here

Would pointing this branch to merge towards `jess/split-findproperty-component` make this a more meaningful diff for now, even if it's then repointed (and merged) to `main` once the original branch is merged?

api.planx.uk/admin/migrations/splitFindProperty.test.ts
DafyddLlyr: So something along these lines should work if you just want a single random node Id for the entire test file - 

```ts
jest.mock("nanoid-good", () => ({
  customAlphabet: jest.fn().mockImplementation(() => () => () => "randomID123"),
}));
```

The currying done by customAlphabet and uniqueId means the syntax a little weird looking here.

If you require different random ids in different tests it gets a little more tricky as Jest hoists mocks - I've found this a little troublesome in the past. This means it's not quite as simple as just declaring and reassigning a variable outside of the scope of the mock. 

Let me know if the above works or if you!

jessicamcinchak: this works, thank you!!

jessicamcinchak:
Closing this one, wrote up summary notes in associated Trello ticket here https://trello.com/c/wkj7gb1q/2253-migrate-existing-flows-to-have-findproperty-and-findpropertyreview

PR #1431, opened 2023-02-06
chore: Standardise API imports, gql tags

DafyddLlyr:
A few tidy ups which were flagged by the following comment - https://github.com/theopensystemslab/planx-new/pull/1424#discussion_r1095754112


DafyddLlyr:
Thanks for the comments here @builditben - you're right, explicit naming is better 👍

PR #1432, opened 2023-02-06
fix: Update getAddressForPerson() - address is required

DafyddLlyr:
# What does this PR do?
 - Ensures that `getAddressForPerson()` always returns an `ExternalAddress`
 - Due to a content bug (now fixed), it was previously possible to submit an application without an applicant address when submitted by an agent
 - This requires a site address in the passport. We'll need to revisit this assumption when an applicant has a site without an address.

This type of issue raises a whole load of questions for me about how we validate and understand passports, and how we handle this. The approach implemented here is still prone to content bugs (for example, if we don't ask for the agents address it will provide the site address).

## Why is this important?
A number of recent Uniform submissions have failed, all of which did not have an applicant address. The following message from Idox suggests that this is the cause of the issues we've been facing - 

> There is an issue where Applicant or Agent Address Details are not being passed in the appication.xml

## Context
Copied from `#planx` Slack thread - https://opensystemslab.slack.com/archives/C5Q59R3HB/p1675676050572899

> We’ve worked under the following assumptions -
>  - If an applicant is a resident, don’t ask extra questions in PlanX
>    - Derive address from site location, pass this along to Uniform as the applicant’s address
> - If they are not resident, ask the applicant for address details in PlanX
>   - Pass `applicant.address` to Uniform as the applicant’s address
> - If we don’t have an address, that’s fine it’s not required in the OneApp schema anyway
>   - ⚠️ I’m guessing this is where the issue is with Uniform - it looks like they may require this?
>
> What I think this means in the realm of content is that -
> - We always need to know an applicant’s address
> - We probably always need to ask the `applicant.resident` question to determine this

### DIFF
--- a/editor.planx.uk/src/@planx/components/Send/uniform/UniformPayload/model.ts
+++ b/editor.planx.uk/src/@planx/components/Send/uniform/UniformPayload/model.ts
@@ -206,13 +206,7 @@ export class UniformPayload implements IUniformPayload {
 
   private getAddressForPerson = (
     person: "applicant.agent" | "applicant"
-  ): ExternalAddress | undefined => {
-    const isResident =
-      this.passport.data?.["applicant.resident"]?.[0]?.toLowerCase() === "true";
-    // Passports of resident applicants will not have an "application.address" value
-    // This must be generated from their siteAddress
-    if (person === "applicant" && isResident)
-      return this.getAddressFromSiteAddress();
+  ): ExternalAddress => {
     const address: Address = this.passport.data?.[`${person}.address`];
     if (address)
       return {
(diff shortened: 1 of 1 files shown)


jessicamcinchak (approved):
thanks for investigating and updating this one! 

really strange to be learning about XML requirements like this now (and hard to believe we missed in testing and so many live application up until now??) - but that's a bigger thing we'll figure out later!

DafyddLlyr:
> really strange to be learning about XML requirements like this now (and hard to believe we missed in testing and so many live application up until now??) - but that's a bigger thing we'll figure out later!

@jessicamcinchak  I know right! I went back and checked how we were previously handling this, see diff here - https://github.com/theopensystemslab/planx-new/pull/1181/files#diff-af710d14f82dd4403843eb9e0515ec4d08041121df7c0e47810d8c27e0dca082L149-L170

We would have always ran afoul of this issue - if we didn't have an `applicant.address`, there would not have been values here. It's very strange to me that we're only now hitting this for the first time. 

Some possibilities...
 - There's a slim chance that we never had many "agent" applications before but it seems very odd to have gone from zero to many so quickly.
 - The other option is that empty nodes work but missing nodes don't? Our XML builder have an option in the config to suppress empty nodes or show them which we could toggle.
  - Some changes behind the scenes at the Uniform or connector level. We don't have any way of assessing on understanding this.
  
I have asked Emily and Majad if they can provide the XML schema again (and will also ask Tascomi) as the Planning Portal hosted one originally provided is down. I'm pretty sure that this field is an optional one on the schema but I'd love to be able to check that.

PR #1434, opened 2023-02-06
chore: Convert `api.planx.uk/send/bops.js` to TypeScript

DafyddLlyr:
**What does this PR do?**
 - Convert `bops.js` -> `bops.ts`
 - Use `fixRequestBody()` instead of custom function (https://github.com/theopensystemslab/planx-new/pull/1434#discussion_r1100022772)
 - Bump test coverage for this file to 100%

**To do**
 - [x] Confirm I can still send a test application to BOPS staging from Pizza
 
image

### DIFF
--- a/api.planx.uk/send/bops.test.ts
+++ b/api.planx.uk/send/bops.test.ts
@@ -19,11 +19,22 @@ import app from "../server";
     beforeEach(() => {
       queryMock.mockQuery({
         name: "FindApplication",
-        matchOnVariables: false,
+        matchOnVariables: true,
         data: {
           bops_applications: []
         },
-        variables: { sessionId: 123 },
+        variables: { session_id: 123 },
+      });
+
+      queryMock.mockQuery({
+        name: "FindApplication",
+        matchOnVariables: true,
+        data: {
+          bops_applications: [
+            { response: { message: "Application created", id: "bops_app_id" } }
+          ]
+        },
+        variables: { session_id: "previously_submitted_app" },
       });
 
       queryMock.mockQuery({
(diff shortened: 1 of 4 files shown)


api.planx.uk/send/bops.ts
```diff
@@ -35,21 +47,21 @@ const sendToBOPS = async (req, res, next) => {
...
       headers: {
-        ...req.headers,
+        ...(req.headers as Record<string, string | string[]>),
         Authorization: `Bearer ${process.env.BOPS_API_TOKEN}`,
       },
       pathRewrite: () => "",
       target,
       selfHandleResponse: true,
-      onProxyReq: (proxyReq, req, res) => {
+      onProxyReq: (proxyReq, req, _res) => {
```
DafyddLlyr, line 56: @jessicamcinchak It seems like we could replace this function with `fixRequestBody` from `http-proxy-middleware`. This is what was previously used, and is used on the Pay proxy currently.

Replacing this doesn't cause any tests to fail currently.

Can you recall why we took this approach, or give an example where it's needed over `fixRequestBody` so I can write a test for it?  I couldn't find any explicit discussions around this on PR https://github.com/theopensystemslab/planx-new/pull/1127

api.planx.uk/send/bops.test.ts
```diff
@@ -83,5 +94,42 @@ import app from "../server";
...
+        .send({ payload: { applicationId: 123, planx_debug_data: { session_id: "previously_submitted_app" } } })
+        .expect(200)
+        .then((res) => {
+          expect(res.body).toEqual({
+            sessionId: "previously_submitted_app",
+            bopsId: "bops_app_id",
+            message: "Skipping send, already successfully submitted",
+          });
+        });
+    });
```
jessicamcinchak, line 133: :+1:

api.planx.uk/send/bops.ts
jessicamcinchak: so I think `onProxyReq` came about here while debugging how the payload was being passed/parsed and following example solutions in the issue thread linked here that favored `onProxyReq` over `fixRequestBody`: https://github.com/theopensystemslab/planx-new/pull/1127#discussion_r958371210 

but i don't have any further reasons / think it was a convenience rather than requirement! as long as `fixRequestBody` doesn't break payload parsing here then I'd say good to update for consistency with Pay :+1: 

just about to put through a test submission to BOPS now on the pizza to double check.

jessicamcinchak: sorry i started reviewing this one yesterday & then didn't come back - i tried submitting two test apps to BOPS staging (Lambeth) via this pizza that haven't succeeded. 

see failed events here: https://hasura.1434.planx.pizza/console/events/one-off-scheduled-events/processed (returning general 500 internal server error, structure of payload seems fine/unchanged on quick look!)

used this test flow: https://1434.planx.pizza/lambeth/bops-test

just want to flag to you! will be curious if investigating leads you back to `onProxyReq` / jogs anything in memory?

DafyddLlyr: No need to apologise - thanks for the solid example, looking into it now 👍

DafyddLlyr: Right, so I got to the bottom of this one.

The issue was that we were trying to forward the entire `req.body` to BOPS, but their API just expects the payload we've constructed with `getBOPSParams()`. This is why `fixRequestBody` was fine before we migrated to Hasura events, but didn't work after. 

The solution previously in place was just a re-implementation of `fixRequestBody()` but handling `req.body.payload` instead of just `req.body` as the library function does (source: https://github.com/chimurai/http-proxy-middleware/blob/master/src/handlers/fix-request-body.ts).

I've pushed up a solution which I think clarifies the intention here.

BOPS submissions on the Pizza fail both before and after this change, so it looks like there's something else in this PR which isn't quite right. Currently trying to narrow it down...

PR #1436, opened 2023-02-06
feat: submission templates

ghost:
This change introduces document templates which can be attached to individual flows. I originally added this functionality in #1373 but the scope of the change there included moving XML document generation to the API. This PR keeps the XML generation where it is and adds functionality to both the frontend and backend.

Moving responsibility for data manipulation to the API is the right approach in the long-run. Without that refactor,  this change requires keeping two separate sets of filename conventions in sync (the names in the proposal.xml need to match the generated documents).

In order to fetch the document templates attached to a flow (to add to proposal.xml), we need to make an async call on the frontend. This data then needs to be passed to the document generation functions before the payload is sent to the hasura event queue. Unfortunately, the document generation functions needed refactoring to enable this workflow due to an issue with React hooks (`useCurrentRoute` was being called within the document generation functions so these functions could not be used within a `useAsync` hook). I tried to avoid too much change in this PR but this change was unavoidable (please correct me if you spot an alternative approach).

I also have added additional data to the event payload which will eventually just pass the sessionId. If I didn't do this, I would have to have made an additional API call to fetch the same data. Since this payload is due a refactor to just support sessionId, I didn't work on any additional refactors here.

I have tested this locally but I'm still to QA this via the pizza. I've added a "code complete" tag to indicate that this is ready for a code-only review so please don't worry about QA testing just yet (I'll put some time into that on Monday).

### DIFF
--- a/api.planx.uk/send/uniform.js
+++ b/api.planx.uk/send/uniform.js
@@ -282,7 +282,8 @@ export async function createUniformSubmissionZip({
       });
     } catch (e) {
       console.log(`Template "${templateName}" could not be generated so has been skipped`);
-      if (e.message) console.log(e.message)
+      console.log(e)
+      continue
     }
     if (isTemplateSupported) {
       const templatePath = path.join(tmpDir, `${templateName}.doc`);
--- a/api.planx.uk/pnpm-lock.yaml
+++ b/api.planx.uk/pnpm-lock.yaml
@@ -10,7 +10,7 @@ specifiers:
   '@airbrake/node': ^2.1.7
   '@babel/core': ^7.19.6
   '@babel/preset-typescript': ^7.18.6
-  '@opensystemslab/planx-document-templates': git://github.com/theopensystemslab/planx-document-templates#f41f371
+  '@opensystemslab/planx-document-templates': git://github.com/theopensystemslab/planx-document-templates#main
   '@types/adm-zip': ^0.5.0
   '@types/body-parser': ^1.19.2
   '@types/cookie-parser': ^1.4.3
@@ -82,7 +82,7 @@ specifiers:
 
 dependencies:
   '@airbrake/node': 2.1.7
-  '@opensystemslab/planx-document-templates': github.com/theopensystemslab/planx-document-templates/f41f371_@babel+core@7.19.6
+  '@opensystemslab/planx-document-templates': github.com/theopensystemslab/planx-document-templates/f689b3f193cef300a10024266c50c2078db93452_@babel+core@7.19.6
   adm-zip: 0.5.9
   aws-sdk: 2.1180.0
   axios: 0.27.2
@@ -7918,9 +7918,9 @@ packages:
       - encoding
     dev: false
 
-  github.com/theopensystemslab/planx-document-templates/f41f371_@babel+core@7.19.6:
-    resolution: {tarball: https://codeload.github.com/theopensystemslab/planx-document-templates/tar.gz/f41f371}
-    id: github.com/theopensystemslab/planx-document-templates/f41f371
+  github.com/theopensystemslab/planx-document-templates/f689b3f193cef300a10024266c50c2078db93452_@babel+core@7.19.6:
+    resolution: {tarball: https://codeload.github.com/theopensystemslab/planx-document-templates/tar.gz/f689b3f193cef300a10024266c50c2078db93452}
+    id: github.com/theopensystemslab/planx-document-templates/f689b3f193cef300a10024266c50c2078db93452
     name: '@opensystemslab/planx-document-templates'
     version: 1.0.0
     engines: {node: ^16, pnpm: ^7.8.0}
--- a/api.planx.uk/send/uniform.test.ts
+++ b/api.planx.uk/send/uniform.test.ts
@@ -1,4 +1,3 @@
-import { queryMock } from "../tests/graphqlQueryMock";
 import { createUniformSubmissionZip } from "./uniform";
 
 jest.mock("fs");
@@ -19,9 +18,10 @@ const mockPipe = {
     },
   })),
 };
+const mockHasRequiredDataForTemplate = jest.fn(() => true);
 jest.mock("@opensystemslab/planx-document-templates", () => {
   return {
-    hasRequiredDataForTemplate: jest.fn().mockImplementation(() => true),
+    hasRequiredDataForTemplate: jest.fn(() => mockHasRequiredDataForTemplate()),
     generateDocxTemplateStream: jest.fn().mockImplementation(() => mockPipe),
     generateHTMLOverviewStream: jest.fn().mockImplementation(() => mockPipe),
     generateHTMLMapStream: jest.fn().mockImplementation(() => mockPipe),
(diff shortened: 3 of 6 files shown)


editor.planx.uk/src/@planx/components/Send/uniform/UniformPayload/UniformPayload.test.ts
```diff
@@ -590,64 +562,74 @@ describe("File handling", () => {
...
+      expect.arrayContaining(expectedBoundaryFileDeclarations)
     );
   });
 
-  it("does not include a boundary geojson file when not possible", () => {
-    const expectedBoundaryFileDeclaration = {
-      "common:FileName": "boundary.geojson",
-      "common:Reference": "Other",
-    };
+  it("does not include a boundary geojson file when 'property.boundary.site' is not present in the passport", () => {
```
DafyddLlyr, line 616: Much clearer and more meaningful 👌

e2e/src/context.ts
```diff
@@ -87,12 +89,39 @@ export function generateAuthenticationToken(userId) {
...
+  );
+  const SECRET = process.env.HASURA_GRAPHQL_ADMIN_SECRET!;
+  return new GraphQLClient(API, {
+    headers: { "X-Hasura-Admin-Secret": SECRET },
+  });
+}
+
 export async function findSessionId(
-  client: Client,
+  admingGQLClient: GraphQLClient,
```
DafyddLlyr, line 120: typo: `admingGQLClient` -> `adminGQLClient`

e2e/src/context.ts
```diff
@@ -58,17 +59,18 @@ export async function setUpTestContext(
   return context;
 }
 
-export async function tearDownTestContext(client: Client, context: Context) {
+export async function tearDownTestContext(context: Context) {
+  const adminGQLClient = getGraphQLClient();
```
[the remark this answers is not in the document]
DafyddLlyr: Yeah this is a fair point.

It doesn't really make sense to put test specific logic into the planx-client, so we end up needing a specific client to handle test only procedures. Would just `testClient` or `testEnvironmentClient` make any sense here?

I think the naming isn't too bad as is, maybe we could do with more explicit docs on `getGraphQLClient()` to explain this though?

DafyddLlyr (approved):
Helpful comments throughout as always 👍 

A few comments / nits here and there - please disregard as you see fit...!

I'm going to do a pass of manual testing myself on the branch (just locally, no real submissions as I know you're handling that). If there's any feedback from there I'll pass along here as another review. 

Code looks great 👍

api.planx.uk/send/email.ts
```diff
@@ -125,19 +124,25 @@ const downloadApplicationFiles = async(req: Request, res: Response, next: NextFu
...
         zip.addFile("boundary.geojson", geoBuff);
+
+        // generate and add an HTML boundary document
+        const boundaryPath = path.join(tmpDir, "boundary.html");
+        const boundaryFile = fs.createWriteStream(boundaryPath);
+        const boundaryStream =
+          generateHTMLMapStream(geojson).pipe(boundaryFile);
+        await resolveStream(boundaryStream);
+        zip.addLocalFile(boundaryPath);
+        deleteFile(boundaryPath);
```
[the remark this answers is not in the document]
jessicamcinchak: adding map file, but waiting on doc templates seems right to me since we know Send to Email is being mostly used for Prior Approval testing right now, where these docs don't apply yet :+1:

api.planx.uk/send/uniform.js
```diff
@@ -224,47 +237,68 @@ export async function createZip({
...
+  const overviewPath = path.join(tmpDir, "Overview.htm");
+  const overviewFile = fs.createWriteStream(overviewPath);
+  const overviewStream = generateHTMLOverviewStream(csv).pipe(overviewFile);
+  await resolveStream(overviewStream);
+  zip.addLocalFile(overviewPath);
+  deleteFile(overviewPath);
 
-  // build an optional GeoJSON file for validators
+  // add an optional GeoJSON file to zip
+  const geojson = passport.data["property.boundary.site"];
```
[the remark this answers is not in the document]
jessicamcinchak: DrawBoundary is a unique Editor component in that it has default passport variables - technically it can still be changed by content editors, but I agree it's fair to assume it'll always work here. See here for reference https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/src/%40planx/components/DrawBoundary/model.ts#L30

api.planx.uk/send/uniform.js
```diff
@@ -224,47 +237,68 @@ export async function createZip({
...
   if (geojson) {
     const geoBuff = Buffer.from(JSON.stringify(geojson, null, 2));
-    zip.addFile("boundary.geojson", geoBuff);
+    zip.addFile("BoundaryGeoJSON.geojson", geoBuff);
```
jessicamcinchak, line 264: nit: `LocationPlanGeoJSON.geojson` for consistency with HTML boundary doc?

api.planx.uk/send/uniform.js
```diff
@@ -54,8 +59,16 @@ const sendToUniform = async (req, res, next) => {
     const { clientId, clientSecret } = getUniformClient(
       req.params.localAuthority
     );
+
     // Setup - Create the zip folder
-    const zipPath = await createZip(payload);
+    const zipPath = await createUniformSubmissionZip({
+      sessionId: payload.sessionId,
+      templateNames: payload.templateNames,
```
jessicamcinchak, line 66: passing these along in the payload seems like a fine first approach while we're still generating XML client side :ok_hand: 

one bigger picture question: do you imagine templateNames should always be stored on the `flows` table, or are these actually something we could infer from the Passport going forward? (Eg similar to how we decide to include or not boundary files, but using "application.type" = "ldc.existing" || "ldc.proposed" passport key/values). After that first passport check to determine which template(s) apply, then you could still proceed with the required fields validation and so on? 

once confirmed we can successfully send doc files to Uniform, I'm just wondering if there will be any cases where one LDC service would _not_ want to always try to generate the relevant doc(s) and if the passport-check approach would save repeated configs at the database-field level?

recognize the null-able database field is a nice proxy for a feature-flag though right now! not suggesting any changes here, just still a little unsure if templates at the flow-level make complete sense to me.

jessicamcinchak (approved):
This looks really good, smaller PR is much appreciated here :+1:

editor.planx.uk/src/@planx/components/shared/hooks.ts
```diff
@@ -14,6 +14,11 @@ export const useTeamSlug = () => {
   return route?.data?.team;
 };
 
+export const useFlowName = () => {
+  const route = useCurrentRoute();
+  return route?.data?.flowName;
+};
```
DafyddLlyr, line 20: This is a great little refactor, should be easy now to replace with a store function in future if required 👍

DafyddLlyr (approved):

api.planx.uk/send/uniform.js
jessicamcinchak: yes totally happy for this to merge as-is right now (sorry meant for that to be clear in original comment), just meant to raise these as questions to keep in mind as we come onto those future tasks and actually hook documents up to flows/passport going forward

PR #1441, opened 2023-02-08
infra: sandbox environment

gunar:
Creates a Sandbox cloud environment so that we can test, experiment, and develop the Infrastructure-as-Code.

https://trello.com/c/671FoXJ1/2047-create-a-sandbox-infrastructure-environment


api.planx.uk/helpers.ts
```diff
@@ -1,5 +1,5 @@
-import { gql } from 'graphql-request';
-import { capitalize } from 'lodash';
+import { gql } from "graphql-request";
```
gunar, line 1: This diff is mostly prettification .The only change happened on line 213.

api.planx.uk/helpers.ts
```diff
@@ -188,16 +209,19 @@ const makeUniqueFlow = (flowData: Flow["data"], replaceValue: string): Flow["dat
   return flowData;
 };
 
-const isLiveEnv = () => (["production", "staging", "pizza"].includes(process.env.NODE_ENV || ""));
+const isLiveEnv = () =>
+  ["production", "staging", "pizza", "sandbox"].includes(
```
gunar, line 213: Added `"sandbox"` here.

infrastructure/application/index.ts
```diff
@@ -100,14 +100,6 @@ export = async () => {
...
       new awsx.ec2.SecurityGroup("metabase-custom-port", {
         vpc,
-        ingress: [
-          {
-            protocol: "tcp",
-            cidrBlocks: ["0.0.0.0/0"],
-            fromPort: 443,
-            toPort: 443,
-          },
-        ],
```
gunar, line 110: Had an [issue with a duplicate security group rule being set](https://github.com/terraform-aws-modules/terraform-aws-security-group/issues/225), which made `pulumi up` fail. I've removed this explicitly set ingress rule from here and metabase is still accessible on [metabase.planx.in](https://metabase.planx.in).

If this change ends up disrupting our production metabase service, we'll get a UptimeRobot alert and I (we) will know this is probably the culprit, so we can fix it.

infrastructure/application/index.ts
```diff
@@ -472,7 +464,7 @@ export = async () => {
           { name: "PORT", value: String(SHAREDB_PORT) },
           {
             name: "JWT_SECRET",
-            value: config.require("jwt-secret"),
+            value: config.requireSecret("jwt-secret"),
```
gunar, line 467: Using `config`.requireSecret` make it so that pulumi forces you to encrypt this config value using `pulumi config set config-name --secret`. It's just nice so we don't end up storing secrets as plain-text.

infrastructure/application/index.ts
```diff
@@ -629,7 +621,7 @@ export = async () => {
   const cdn = createCdn({ domain: DOMAIN, acmCertificateArn: sslCert.arn });
 
   const frontendDnsRecord = new cloudflare.Record("frontend", {
-    name: tldjs.getSubdomain(DOMAIN) ?? "@",
+    name: tldjs.getSubdomain(DOMAIN) || "@",
```
gunar, line 624: The `getSubdomain` method returns `""` instead of `null`, so needed to fix this here as the sandbox environment is not using a subdomain at this moment (i.e. production hasura is `hasura.editor.planx.uk` but sandbox's is just `hasura.planx.in`)

infrastructure/application/utils/generateTeamSecrets.ts
```diff
@@ -19,21 +19,30 @@ const name = (name: string) => name.replace(regex, "_").toUpperCase();
...
     });
-    team.uniformInstances?.forEach(instance => {
+    team.uniformInstances?.forEach((instance) => {
       secrets.push({
         name: `UNIFORM_CLIENT_${name(instance)}`,
-        value: config.require(`uniform-client-${value(instance)}`),
+        value:
+          env === "sandbox"
+            ? "sandbox"
+            : config.require(`uniform-client-${value(instance)}`),
```
gunar, line 41: I've added this ternary operator / conditional here just so that we don't have to set up pulumi configs for uniform secrets for each team on sandbox, as sandbox isn't going to need them but otherwise they'd be required in order to call `pulumi up`.

infrastructure/data/index.ts
```diff
@@ -12,8 +12,9 @@ const networking = new pulumi.StackReference(`planx/networking/${env}`);
 
 const db = new aws.rds.Instance("app", {
   engine: "postgres",
-  // Available versions: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts.General.version12
-  engineVersion: "12.3",
+  // Available versions:
+  // $ aws rds describe-db-engine-versions --default-only --engine postgres
```
gunar, line 16: Makes the comment evergreen.

infrastructure/data/index.ts
```diff
@@ -12,8 +12,9 @@ const networking = new pulumi.StackReference(`planx/networking/${env}`);
...
+  engineVersion: "14.6",
```
gunar, line 17: Can't use `12.3` anymore as it's been deprecated. This diff isn't going to be a problem for staging nor production as we usually only deploy the `application` stack.

PR #1444, opened 2023-02-10
feat: Delete uploaded files on session sanitation

DafyddLlyr:


api.planx.uk/webhooks/sanitiseApplicationData/operations.ts
```diff
@@ -52,6 +55,47 @@ export const operationHandler = async (
   return operationResult;
 };
 
+/**
+ * Return list of session IDs which are now ready for sanitation
+ */
+export const getExpiredSessionIds = async (): Promise<string[]> => {
```
DafyddLlyr, line 61: This will find all sessions submitted more than 6 months ago, and iterate over them to delete files.

`s3.deleteObjects()` will not fail if it's passed an array of already deleted files (e.g. when this runs the second time).

Once this is live and has succeeded once, we can push a small change here to also check the `sanitised_at` column to ensure this operation only runs on...
 - Sessions submitted more than 6 months ago
 - Session not already flagged as sanitised (handled by the `sanitiseLowcalSessions` operation)

api.planx.uk/webhooks/sanitiseApplicationData/index.ts
```diff
@@ -15,9 +15,12 @@ export const sanitiseApplicationData = async (
...
 ): Promise<Response | void> => {
   const operations = getOperations();
-  let results: OperationResult[] = [];
+  const results: OperationResult[] = [];
   try {
-    results = await Promise.all(operations.map(operationHandler));
+    for (const operation of operations) {
+      const result = await operationHandler(operation)
+      results.push(result)
+    }
```
DafyddLlyr, line 23: Speed of operation doesn't really matter as it's a cron job, but we can no longer run these operations concurrently as we rely on `lowcal_session.data` for the `deleteApplicationFiles` operation to work.

api.planx.uk/webhooks/sanitiseApplicationData/operations.ts
```diff
@@ -52,6 +55,47 @@ export const operationHandler = async (
...
+  const query = gql`
+    query GetSessionIds($retentionPeriod: timestamptz) {
```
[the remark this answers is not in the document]
DafyddLlyr: Thanks for catching this - I made the function name more explicit one one point, but didn't rename the query.

Unique query names is the way to go 👍

DafyddLlyr:
Agreed to merge to `main` w/c 06 March so that it can run against staging S3 bucket for a few days before merging to `production`.

DafyddLlyr:
Over 6 months old - 
- https://s3.console.aws.amazon.com/s3/buckets/user-data-62abb74?region=eu-west-2&prefix=04ug41re/&showversions=false
- https://s3.console.aws.amazon.com/s3/buckets/user-data-62abb74?region=eu-west-2&prefix=9wne59qi/&showversions=false

Less than 6 months old - 
- https://s3.console.aws.amazon.com/s3/buckets/user-data-62abb74?region=eu-west-2&prefix=7fl0uxss/&showversions=false
- https://s3.console.aws.amazon.com/s3/buckets/user-data-62abb74?region=eu-west-2&prefix=9tpamzwy/&showversions=false

PR #1447, opened 2023-02-10
fix: log corrupted node errors to Airbrake in future

jessicamcinchak:
extends #1446, further context over here https://opensystemslab.slack.com/archives/C5Q59R3HB/p1676032127570869

since this is an error that stems from content editing, seems sensible to log it via Airbrake so we get visibility as soon as possible as we're less likely to catch them in local code changes & testing

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
@@ -657,7 +657,7 @@ export const removeOrphansFromBreadcrumbs = ({
   | Store.cachedBreadcrumbs
   | Store.breadcrumbs => {
   const idsToRemove =
-    flow[id].edges?.filter(
+    flow[id]?.edges?.filter(
       (edge) => !(userData?.answers ?? []).includes(edge)
     ) ?? [];
 
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
@@ -1,8 +1,7 @@
-import { gql } from "@apollo/client";
 import tinycolor from "@ctrl/tinycolor";
 import { TYPES } from "@planx/components/types";
 import { sortIdsDepthFirst } from "@planx/graph";
-import { client } from "lib/graphql";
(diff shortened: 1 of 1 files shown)


DafyddLlyr (approved):
One question - if the idea here is to notify us via Airbrake when something goes wrong, can we now revert to back to `flow[id].edges` instead of `flow[id]?.edges`?

If not, because we don't have enough certainty here, how do we get there? Can we check for this condition on "Publish"?

None of this needs to happen on this PR of course, just interested in what a longer term fix here would look like.

jessicamcinchak:
> One question - if the idea here is to notify us via Airbrake when something goes wrong, can we now revert to back to `flow[id].edges` instead of `flow[id]?.edges`?
> 
> If not, because we don't have enough certainty here, how do we get there? Can we check for this condition on "Publish"?

Good questions, sorry I didn't explain this so well. Here's my best explanation of where we're at / where we should go: 
- Currently, flows with corrupted nodes can be successfully published, as well as be successfully referenced as external portals and flattened/published by the parent flow (what happened last Friday)
  - Flows with corrupted external portals can NOT currently be published because the flattening function (dataMerged) will fail to fetch the corrupt flowId - this error is now shown to editors when they click "Publish" based on changes here: #1435 
- If we reverted back to `flow[id].edges` here right now, then a user would be prevented from "Continuing" through the first card that belongs the flow (or external portal) containing the corruped node. I'm still using `flow[id]?` to allow users to "continue", with the hope that most journeys won't actually cross paths with the corrupted node, which would then still fail / block them from proceeding. We'll log in the error in every case
- In order to check for this condition on publish (then be able to safely assume `flow[id]!` in /preview), we'd have to do something like: gather every nodeId that appears in `edges`, ensure it has a corresponding key in the outer level of the schema with a valid `"type"`. If it doesn't, fail to publish and show an error, or actually just auto-delete the nodeId from `edges` and proceed with publishing (which matches the change I manually made, because afaik there isn't actually a valid Editor operation for corrupted nodes currently, eg no "delete"/"update" panel). At point of publish, we're looking at potentially millions of nodes for something like LDC, so this validation would require a bit of performance testing.

Does that all make sense? I'll make a Trello ticket to capture this last point now too :heavy_plus_sign:

PR #1448, opened 2023-02-10
feat: stub out MultipleFileUpload component and update how-to-add docs

jessicamcinchak:
noticed these docs were quite out of date while working on property components recently, decided to stub out a new `MultipleFileUpload` while I was here as a quick exercise to make sure I caught everything.

component doesn't do anything yet, but all necessary files are here which should help make future PRs easier to start and more contained.


editor.planx.uk/docs/adding-a-new-component.md
```diff
@@ -1,10 +1,18 @@
...
+## Core directory & files 
 
-`model.ts`
+1. `src/@planx/components/types.ts`
+
+```typescript
+SetValue = 380,
+```
+
+1. `mkdir src/@planx/components/SetValue`
```
DafyddLlyr, line 13: nit: Markdown auto incrementing fails when there's a block between points, making this document actually display as 1., 1., 1. etc.

editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
```diff
@@ -54,6 +54,7 @@ const NodeTypeSelect: React.FC<{
       <optgroup label="Inputs">
         <option value={TYPES.TextInput}>Text Input</option>
         <option value={TYPES.FileUpload}>File Upload</option>
+        <option value={TYPES.MultipleFileUpload}>Multiple File Upload</option>
```
DafyddLlyr, line 57: nit: Could we comment this out for now, juuuuuust incase, as we don't have validation to catch these "in progress" nodes on Publish.

DafyddLlyr (approved):
Thanks for this and for improving docs 🥇

editor.planx.uk/docs/adding-a-new-component.md
jessicamcinchak: you're right - thanks for catching that ! manually numbered now.

editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
jessicamcinchak: done!

PR #1450, opened 2023-02-13
feat: Zod Schema for Uniform Payload

DafyddLlyr:
# What does this PR do?
Adds a schema for the Uniform payload - this means we can validate before passing things over to Uniform. This validation is done when the `parse()` function is called.

This actually caught a few issues with our current mapping and schema which is a positive sign. Commit https://github.com/theopensystemslab/planx-new/pull/1450/commits/cb1b0355c88ecf3213b17c58f584bd3110ec4535 shows these changes.

This isn't totally complete, but I think it's best to push what I have now for discussion before I take it too much further.

## Principles
 - Default values should be handled by the schema
 - `mapPassportToUniformPayload()` should be a straight mapping of values from our passport to the schema
 - This will always produce a `PartialDeep` object, with the missing default values subbed in when `parse()` is called

## Potential next steps...
- Reframe tests - just test our mapping functions, rather than a conversion to/from XML each time. We can still have tests to cover the full passport -> xml conversion.
- Test different passports - LDC Proposed, LDC Existing, Agents, etc
- Rename UniformPayload -> OneApp
- Move to `planx-client`, call from backend
- This could be called by a `PassportTransformer` (or something with a better name) that just takes a passport / session and can return BOPS Payload / OneApp XML / CSV etc. The OneApp logic can live in it's own class still.
- `Passport` class for handling passport transformations and getters (see comments)


editor.planx.uk/src/@planx/components/Send/uniform/UniformPayload/model.ts
```diff
@@ -52,29 +52,32 @@ export class UniformPayload implements IUniformPayload {
...
+    this.payload = this.mapPassportToUniformPayload();
+  }
 
-    this["portaloneapp:Proposal"] = {
-      ...this.getNamespaces(),
-      _Version: "1.3",
-      "portaloneapp:SchemaVersion": 1.3,
+  private stringToBool = (value: string): boolean | undefined => {
+    if (value) return value.toLowerCase() === "true";
+  };
```
DafyddLlyr, line 60: I think what I'd ideally like to put in place is a `Passport` class that just had a getter than handles this sort of thing, e.g. - 

```ts
// Passport.ts

class Passport {
  userPassport: Store.passport

  constructor(userPassport: Store.passport) {
    this.userPassport = userPassport
  }


  get = (value: string): boolean | string[] | string | number => {
    // Logic here to correctly access passport variables
  }
}
```

```ts
// model.ts
const passport = new Passport(userPassport)
...
"common:HaveSoughtAdvice": passport.get("application.preAppAdvice")
```

There's enough change in this PR as is, so this is just a helper function for now, but comments on the above most welcome.

editor.planx.uk/src/@planx/components/Send/uniform/UniformPayload/model.ts
```diff
@@ -52,29 +52,32 @@ export class UniformPayload implements IUniformPayload {
...
-          ...this.getGeneratedFiles(),
-          ...this.getUserUploadedFiles(),
+          ...(this.getGeneratedFiles() as FileAttachment[]),
+          ...(this.getUserUploadedFiles() as FileAttachment[]),
         ],
       },
-      "portaloneapp:Applicant": this.getApplicantOrAgent("applicant"),
-      "portaloneapp:Agent": this.getApplicantOrAgent("applicant.agent"),
+      "portaloneapp:Applicant": this.getApplicant(),
+      "portaloneapp:Agent": this.getAgent(),
```
DafyddLlyr, line 80: This was just solving some type problems.

editor.planx.uk/src/@planx/components/Send/uniform/UniformPayload/schema.ts
```diff
@@ -0,0 +1,247 @@
+import { z } from "zod";
+
+// Models the XML payload required by Uniform to process PlanX applications
+
+export const paymentSchema = z.object({
```
DafyddLlyr, line 5: The contents of this file were generated by just pasting the existing types into https://transform.tools/typescript-to-zod

I think there's a viable (if slightly annoying) method where we could enrich this by going XSD -> JSON Schema -> Zod

editor.planx.uk/src/@planx/components/Send/uniform/UniformPayload/model.ts
DafyddLlyr: I think a nice thing about this approach also is that if we only expose a `get()` interface we can more easily change the structure of the passport in future without breaking too much.

editor.planx.uk/src/@planx/components/Send/uniform/UniformPayload/model.ts
```diff
@@ -296,34 +299,28 @@ export class UniformPayload implements IUniformPayload {
...
       format: true,
       suppressEmptyNode: true,
     };
-    const builder = new XMLBuilder(buildOptions);
-    const xml = builder.build(payload);
+    return new XMLBuilder(buildOptions);
+  };
+
+  public buildXML = (): string => {
+    const validatedPayload = iUniformPayloadSchema.parse(this.payload);
```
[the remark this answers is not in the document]
DafyddLlyr: This is totally fair and a good catch thanks - I'll take a look at this 👍

editor.planx.uk/src/@planx/components/Send/uniform/UniformPayload/model.ts
```diff
@@ -178,32 +189,40 @@ export class UniformPayload implements IUniformPayload {
...
-        _TelMobile: "yes",
+      "common:OrgName": this.passport.data?.[`${person}.company.name`],
+      "common:ContactDetails": {
+        "common:Email": {
+          "apd:EmailAddress": this.passport.data?.[`${person}.email`],
+        },
+        "common:Telephone": {
+          "apd:TelNationalNumber":
+            this.passport.data?.[`${person}.phone.primary`],
+        },
```
jessicamcinchak, line 221: nice to see that defaults like `_TelUse`, `_TelPreferred`, and `_TelMobile` are defined on the ZOD schema now, that feels a lot more readability :+1:

jessicamcinchak (approved):
super valuable, thanks for this !

DafyddLlyr:
> This function can throw and error and that error wouldn't be gracefully handled by it's caller at the moment.

@builditben So error handling ended up being slightly more tricky than anticipated, but a rough implementation is in place in https://github.com/theopensystemslab/planx-new/pull/1450/commits/e28ed92aa9fc7f02bb846d8b3814228d0ffd6550

We want the following to be true -
 - PlanX devs know there's a problem
 - Applicant does not need to know there's a problem
 - Submission should not sent to Uniform

As a result, I'm catching and logging any errors when parsing, and failing silently in both the frontend and API so as to not alert the applicant. I expect that this becomes a simpler process when this logic moves wholesale to the API.

PR #1452, opened 2023-02-14
doc: re-submission steps for orphan payment scenario

jessicamcinchak:
Thanks for adding this doc yesterday! 

I think there's perhaps a case to split it into two sections/scenarios though (often partners are using same language for both/either scenario, yet technical steps required are different): 
1.  We're re-submitting because there was a previous attempt that failed. This is most common, and makes good case for re-submission API
2.  We're re-submitting beause a partner notified us of a payment without a submission. This is increasingly rare, but can still occassionally happen and anyone should know how to investigate/proceed. These are very uncommon edge cases that will probably never justify scripting or wrapping in the re-submissions API

Second section comes from this pinned Slack message in #planx, which is hard to find and collectively update: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1661244241319919 

If we don't want these steps spelled out in the doc, we could also just link to this pinned message though.

### DIFF
--- a/doc/how-to/how-to-resubmit-an-application.md
+++ b/doc/how-to/how-to-resubmit-an-application.md
@@ -1,12 +1,16 @@
 # How to re-submit an application to BoPS or Uniform
 This guide will explain how to manually re-submit a PlanX application which has not been successfully processed. 
 
-An application may fail to be submitted to a single destination, or to multiple destinations. This guide will explain how both BoPS and Uniform re-submissions are handled. We will re-create the "send" event generated in Hasura which dispatches a payload to our Express API. There is not currently a way to simply retry an event trigger manually in Hasura.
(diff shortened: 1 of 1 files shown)


doc/how-to/how-to-resubmit-an-application.md
```diff
@@ -1,12 +1,16 @@
 # How to re-submit an application to BoPS or Uniform
 This guide will explain how to manually re-submit a PlanX application which has not been successfully processed. 
 
-An application may fail to be submitted to a single destination, or to multiple destinations. This guide will explain how both BoPS and Uniform re-submissions are handled. We will re-create the "send" event generated in Hasura which dispatches a payload to our Express API. There is not currently a way to simply retry an event trigger manually in Hasura.
+Most commonly, we re-submit cases that have at least one previous submission (which was triggered normally during the applicant journey). In very rare cases, a council may request re-submission because they have received a Gov Payment which is not connected to any attempted submissions.
```
DafyddLlyr, line 4: 👍

DafyddLlyr (approved):
This is great - thank you...!

PR #1454, opened 2023-02-14
fix: Fetch only most recent submission

DafyddLlyr:
Adding an `order_by` here means we just display the most recently submitted payload, which is really the intention.

Multiple records will be returned when an application has been re-submitted.

### DIFF
--- a/api.planx.uk/admin/session/oneAppXML.ts
+++ b/api.planx.uk/admin/session/oneAppXML.ts
@@ -19,10 +19,11 @@ export const getOneAppXML = async (
 const fetchSessionXML = async (sessionId: string) => {
   try {
     const query = gql`
-      query GetUniformApplicationBySessionID($submission_reference: String) {
+      query GetMostRecentUniformApplicationBySessionID($submission_reference: String) {
         uniform_applications(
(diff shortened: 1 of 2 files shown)


api.planx.uk/admin/session/oneAppXML.ts
```diff
@@ -21,7 +21,8 @@ const fetchSessionXML = async (sessionId: string) => {
     const query = gql`
       query GetUniformApplicationBySessionID($submission_reference: String) {
         uniform_applications(
-          where: { submission_reference: { _eq: $submission_reference } }
+          where: { submission_reference: { _eq: $submission_reference } },
+          order_by: { created_at: desc }
```
jessicamcinchak, line 25: nit: `limit: 1` might also help communicate explicit intention here, even though we `return uniform_applications[0]` later?

jessicamcinchak (approved):

api.planx.uk/admin/session/oneAppXML.ts
DafyddLlyr: No this is a great comment - I agree that `created_at` is not particularly meaningful on its own. Updated!

PR #1458, opened 2023-02-15
chore: code checks

builditben:
This PR introduces CI code checks for formatting and linting. It also adds some pre-commit hooks for eslint and prettier checks in the main directories.

Code changes for the actual task of running prettier are in #1460 which will need merged here to pass the build (I thought the review might be easier that way).


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+SCRIPT_DIR=$(dirname "${0}")
+. ${SCRIPT_DIR}/_/husky.sh
+npx lint-staged
```
DafyddLlyr, line 4: Can we now lose Husky as a dependency in the Editor if the pre-commit hook is running here?

builditben:
A few recent PRs introduce a "lint:fix" command which mostly solves this problem.

PR #1462, opened 2023-02-16
chore: Minor MUI bump

DafyddLlyr:
# What's the problem?
Builds and tests on `main` are currently failing, throwing MUI errors.

The recently introduced `@opensystemslab/planx-document-templates` dependency also shares MUI dependencies, so it appears that a mismatch in versions may be causing issues.

Discussion here on PlanX Slack - https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1676533754515749

## Solution
Fixing MUI versions across both repos means all tests pass, and build can be completed. Clearly this isn't an ideal fix but this should let us get staging back up on it's feet.


editor.planx.uk/pnpm-lock.yaml
```diff
@@ -197,12 +197,12 @@ dependencies:
   '@emotion/styled': 11.10.4_jo6cenodoy3dipn53p6fs42f4m
   '@feedback-fish/react': 1.2.1_react@18.2.0
   '@material-ui/core': 4.12.4_63dfw6rfgd64rl77ptfma4cvt4
-  '@mui/icons-material': 5.10.6_q2zmiwrqmz5kmhrrmwgs4bxmzq
-  '@mui/material': 5.10.6_vbxcdxy3bftsefaguotlgalomu
+  '@mui/icons-material': 5.10.6_uiebhasrup4ygf5w5v6m4s6s6e
+  '@mui/material': 5.11.9_vbxcdxy3bftsefaguotlgalomu
```
jessicamcinchak, line 201: :eyes: this is helpful to confirm that pnpm is doing what's expected and pulling latest minor release when package.json specifies `^5.10.16`

but i suspect this means lock files can easily fall out of sync and we might need to lock in at a version (eg drop `^`)? 

looks like planx-document-template lock file would still be behind even though package.json matches now: https://github.com/theopensystemslab/planx-document-templates/blob/main/pnpm-lock.yaml#L1669

jessicamcinchak (approved):

PR #1464, opened 2023-02-16
feat: Write flowName to store

DafyddLlyr:
Quick refactor based on [comments here](https://github.com/theopensystemslab/planx-new/pull/1436) which means we can more readily get the `flowName` and `flowSlug` without having to call a hook.

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/shared.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/shared.ts
@@ -8,10 +8,20 @@ export interface SharedStore extends Store.Store {
   breadcrumbs: Store.breadcrumbs;
   childNodesOf: (id?: Store.nodeId) => Store.node[];
   flow: Store.flow;
+  flowSlug: string;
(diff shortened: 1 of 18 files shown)


editor.planx.uk/src/pages/FlowEditor/lib/store/shared.ts
```diff
@@ -55,8 +69,9 @@ export const sharedStore = (
     });
   },
 
-  setFlow(id, flow) {
-    set({ id, flow });
+  setFlow({ id, flow, flowSlug }) {
+    set({ id, flow, flowSlug });
+    set({ flowName: flowSlug.replaceAll?.("-", " ") });
```
jessicamcinchak, line 74: nit: I think this use of `replaceAll` to format the slug as `flowName` will return slightly different casing than `convertSlugToName` in the API saveAndReturn/utils - would be good to keep consistency!

jessicamcinchak (approved):
long overdue convenience method & simplification, thank you!

PR #1467, opened 2023-02-16
feat: query OS Features for classified roads that intersect with a buffered site boundary

jessicamcinchak:


api.planx.uk/gis/classifiedRoads.js
```diff
@@ -0,0 +1,60 @@
...
+) => {
+  try {
+    if (!req.query.geom)
+      return next({ status: 401, message: "Missing required query param `?geom=`" });
+
+    // Create an OGC XML filter parameter value which will select the road features (lines)
+    //   that intersect with the buffered site boundary (polygon) coordinates
+    //   ref https://labs.os.uk/public/os-data-hub-examples/os-features-api/wfs-example-intersects#maplibre-gl-js
+    const xml = `
+      <ogc:Filter">
```
DafyddLlyr, line 18:
```suggestion
      <ogc:Filter>
```

Typo?

jessicamcinchak: :bulb: :pray: thank you :sweat_smile: you'd really think I'd have a better xml IDE extension by now for spotting these

jessicamcinchak:
[[Implement] Pull in road classifications](https://trello.com/c/EvEuV0lm/2105-implement-pull-in-road-classifications)

api.planx.uk/gis/index.test.ts
```diff
@@ -32,3 +38,71 @@ describe("locationSearchWithTimeout", () => {
...
+        .then((res) => {
+          expect(res.body["constraints"]["article4"]).toBeDefined();
+          expect(res.body["constraints"]["listed"]).toBeDefined();
+          expect(
+            res.body["constraints"]["designated.conservationArea"]
+          ).toBeDefined();
+        });
+    }, 20_000); // 20s request timeout
+  });
+});
```
jessicamcinchak, line 108: these nock tests were previously in server.test.js, i've just colocated them to the /gis subdirectory here for better discoverability.

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.test.tsx
```diff
@@ -62,3 +67,7 @@ it("should not have any accessibility violations", async () => {
   const results = await axe(container);
   expect(results).toHaveNoViolations();
 });
+
+it.todo("fetches classified roads only when we have a siteBoundary"); // using expect(spy).toHaveBeenCalled() ??
+
+it.todo("fetches planning constraints when we have lng,lat or siteBoundary");
```
jessicamcinchak, line 73: our preview tests for PlanningConstraints leave a lot to be desired, but feel a bit hard to get "right" without a bit more refactoring of this general component (eg to enable easier mocking of siteBoundary & lng, lat content dependencies) - which I'd prefer to revisit as a followup task! stubbed out todos in the meantime to help clarify expected functionality.

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
```diff
@@ -95,13 +99,45 @@ function Component(props: Props) {
...
+      bufferRadiusInMeters,
+      { units: "meters", steps: 0 }
+    );
+
+    // Flip the buffered polygon coordinates from [x, y] to [y, x].
+    const flipped: Feature<Polygon, Properties> = flip(bufferedSiteBoundary);
+
+    // Format the flipped coordinates as a space-delimited string for the OGC XML filter param
+    const ogcCoordinates: string = flipped.geometry.coordinates[0].join(" ");
+    classifiedRoadsEndpoint += `?geom=${ogcCoordinates}`;
```
jessicamcinchak, line 118: made the decision to buffer the siteBoundary on the frontend rather than in the API for a couple of reasons:
- `PlanningConstraints` already handles other geo transformations - eg GeoJSON to WKT for Digital Land, GeoJSON to coordinate rings for ESRI, and now GeoJSON to OGC coordinates for OS Features WFS
- If we ever wanted to expose the buffer number to editors as a configurable input, it'd be straightforward to do with this approach
- keeps all @turf dependences in the editor

api.planx.uk/gis/classifiedRoads.js
```diff
@@ -0,0 +1,73 @@
...
+        <ogc:PropertyName>SHAPE</ogc:PropertyName>
+          <gml:Polygon srsName="EPSG:4326">
+            <gml:outerBoundaryIs>
+              <gml:LinearRing>
+                <gml:coordinates>${req.query.geom}</gml:coordinates>
+              </gml:LinearRing>
+            </gml:outerBoundaryIs>
+          </gml:Polygon>
+        </ogc:Intersects>
+      </ogc:Filter>
```
DafyddLlyr, line 26: I was really hoping that we could just buffer here with a `DWITHIN` filter but the OS doesn't support it currently.

From their `GetCapabilities` document - 

```xml
<fes:SpatialOperators>
  <fes:SpatialOperator name="BBOX"/>
  <fes:SpatialOperator name="Equals"/>
  <fes:SpatialOperator name="Disjoint"/>
  <fes:SpatialOperator name="Intersects"/>
  <fes:SpatialOperator name="Touches"/>
  <fes:SpatialOperator name="Crosses"/>
  <fes:SpatialOperator name="Within"/>
  <fes:SpatialOperator name="Contains"/>
  <fes:SpatialOperator name="Overlaps"/>
</fes:SpatialOperators>
```

api.planx.uk/gis/classifiedRoads.js
```diff
@@ -0,0 +1,73 @@
...
+
+    const url = `https://api.os.uk/features/v1/wfs?${new URLSearchParams(params).toString()}`;
+    const features = await fetch(url)
+      .then(res => res.json())
+      .then(data => {
+        if (!data.features?.length) return;
+        
+        // Filter out any intersecting roads that are not classified
+        const classifiedFeatures = data.features.filter((feature) => feature.properties["RoadClassification"] !== "Unclassified");
+        return classifiedFeatures;
```
DafyddLlyr, line 49: We can write this as part of the OGC Filter - 

```xml
<ogc:Filter>
  <ogc:And>
    <ogc:Intersects>
      <ogc:PropertyName>SHAPE</ogc:PropertyName>
      <gml:Polygon srsName="EPSG:4326">
        <gml:outerBoundaryIs>
          <gml:LinearRing>
            <gml:coordinates>${req.query.geom}</gml:coordinates>
          </gml:LinearRing>
        </gml:outerBoundaryIs>
      </gml:Polygon>
    </ogc:Intersects>
    <ogc:PropertyIsNotEqualTo>
      <ogc:PropertyName>RoadClassification</ogc:PropertyName>
      <ogc:Literal>'Unclassified'</ogc:Literal>
    </ogc:PropertyIsNotEqualTo>
  </ogc:And>
</ogc:Filter>
```

api.planx.uk/server.ts
```diff
@@ -471,6 +472,8 @@ app.get("/gis", (_req, res, next) => {
 
 app.get("/gis/:localAuthority", locationSearch);
 
+app.get("/roads", classifiedRoadsSearch);
```
DafyddLlyr, line 475: nit: I think this could be a bit more meaningful with a route like `/gis/roads` or `/gis/road-classification`

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
DafyddLlyr: Thanks for the context here - this is really helpful.

I agree that it's not ideal to have the logic split across domains, I'd be in favour or pulling it together if we can.

The value we're currently getting from including the API seems to be...
 - Hide the OS key
 - Format response in a standard way
 - Facilitate testing
 
The OS key can handled by pointing at the API proxy endpoint which is nice. The other two could be handled (I think!) by encapsulating this in a function/hook/library outside this component.

Happy to go with your judgement here if this feels like it's not worth the effort for what would be gained in consolidating this logic in one place.

api.planx.uk/gis/classifiedRoads.js
```diff
@@ -0,0 +1,73 @@
...
+
+export const classifiedRoadsSearch = async (req, res, next) => {
+  try {
+    if (!req.query.geom)
+      return next({ status: 401, message: "Missing required query param `?geom=`" });
+
+    // Create an OGC XML filter parameter value which will select the road features (lines)
+    //   that intersect with the buffered site boundary (polygon) coordinates
+    //   ref https://labs.os.uk/public/os-data-hub-examples/os-features-api/wfs-example-intersects#maplibre-gl-js
+    const xml = `
```
DafyddLlyr, line 14: OpenLayers has some nice APIs for reading and writing OGC Filters. This would handle little quirks like the x, y -> y, x flip.

 I think this is a simple enough example to have written as a string, but something to be aware of 👍

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
DafyddLlyr: An alternative could be to wrap this all up in the `/gis/:team` API and make a single request for everything.

DafyddLlyr (approved):
Some great work here - very happy for this to get merged as is 👍 

Test cases work as expected and it meets all requirements as I see them.

I think it's worth thinking about the splitting up of the road classification logic here, and the fact that it's split from the current planning constraints request - it would be nice (but not 100% necessary) imho to have this business logic co-located.

api.planx.uk/gis/classifiedRoads.js
```diff
@@ -0,0 +1,73 @@
+import "isomorphic-fetch";
+
+// Passport key comes from Digital Planning Schemas googlesheet
+export const PASSPORT_FN = "road.classified";
```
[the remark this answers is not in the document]
jessicamcinchak: `"fn"` is commonly used across the Editor as the field name which stores the passport key name for a given component type. 

so, in a flow schema, it's common an individual node has a representation like: 
```json
  "nodeId": {
    "data": {
      "fn": "property.constraints.planning",
      "title": "Planning constraints",
      "description": "<p>Things that might affect your project</p>"
    },
    "type": 11
```

`"road.classified"`, like other planning constraint dataset passport keys, isn't one we expose to Editors, but I chose the same naming here for consistency. If too confusing/unclear to reference "fn" in the API, happy for this to just be `"PASSPORT_KEY"` or something similar.

api.planx.uk/gis/classifiedRoads.js
jessicamcinchak: Yep, I had checked this as well (thanks for explicit doc link which I forgot), hence introducing `@turf/buffer` - which still doesn't feel too complicated, but agree `DWITHIN` would have been most ideal here if supported.

api.planx.uk/gis/classifiedRoads.js
jessicamcinchak: clever, i did not think of doing it this way! will update

api.planx.uk/gis/classifiedRoads.js
jessicamcinchak: Yeah, I am very much trying to push off introducing openlayers as a dependency here until we really have a strong case for it - but definitely aware of this if our filters become too complex to easily handle/understand as string templates!

api.planx.uk/server.ts
DafyddLlyr: Ah yeah that's a vary valid point - `/gis/roads` would match `/gis/:localAuthority`. Let's revisit this in the future at some point 👍

api.planx.uk/gis/classifiedRoads.test.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+  ];
+
+  loadOrRecordNockRequests("fetching-classified-roads", locations);
```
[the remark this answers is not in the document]
jessicamcinchak: So, my understanding here is that `supertest()` is still calling this live endpoint, and expects to receive a 200, _then_ the nock is loaded to validate the shape of the response. So even though there's indeed a saved nock, the uptime of external GIS services (which was historically much worse than lately) is still a factor here as I understand it.

The `.skip` came from annoying instances in the past where Digital Land or a GIS server goes down and CI fails/blocks un-related code changes.

Unless I'm misunderstanding this, I'd vote we leave skipped for now and re-evaluate this approach to GIS testing more widely in a separate PR?

api.planx.uk/gis/classifiedRoads.js
```diff
@@ -0,0 +1,73 @@
...
+
+export const classifiedRoadsSearch = async (req, res, next) => {
+  try {
```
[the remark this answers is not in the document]
jessicamcinchak: This makes sense, thanks for spotting! I've updated to tighten up the `try` block to exclude the initial setup and removed the fetch `catch` as that'll already be caught.

jessicamcinchak:
@builditben the `null` road name you caught was a case of the query in that location picking up two different road lines and only displaying the first - which happened to have type "Unknown" - different than "Unclassified" - which I was already filtering out. Now updated if you try again! 

Confirmed all possible values that can be returned by `RoadClassification` using this document https://www.ordnancesurvey.co.uk/documents/os-mastermap-highways-network-roads-technical-specification.pdf (section 3.11.2 on page 43) and updated filters appropriately!

api.planx.uk/gis/classifiedRoads.js
jessicamcinchak: so, after much tinkering with different quoting and not-quoting and google searching, i couldn't actually get this to work! was consistently just getting empty responses.

the list of values we want to filter out has also grown from 1 to 3, so keeping those in an array for now - but on my radar to try and make this work later!

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
jessicamcinchak: A frontend library function/hook/library is great food for thought, going to table right here and address comprehensively in a future PR - since other planning constraint logic in this component has the same concern of being split across front & back currently :+1: 

Reasons for forgoing a single request for everying at this stage  (eg different extents of available data, different spatial query methodogolgies, flexibility to call separately if decided, etc) connected to explanation over here: https://github.com/theopensystemslab/planx-new/pull/1467#discussion_r1112754073

api.planx.uk/gis/classifiedRoads.js
```diff
@@ -4,58 +4,59 @@ import "isomorphic-fetch";
...
     const features = await fetch(url)
       .then(res => res.json())
       .then(data => {
         if (!data.features?.length) return;
         
         // Filter out any intersecting roads that are not classified
-        const classifiedFeatures = data.features.filter((feature) => feature.properties["RoadClassification"] !== "Unclassified");
+        //   find all possible values on page 43 https://www.ordnancesurvey.co.uk/documents/os-mastermap-highways-network-roads-technical-specification.pdf 
+        //   XX in future consider doing this directly in XML using <ogc:PropertyIsNotEqualTo>
+        const classifiedFeatures = data.features.filter((feature) => !["Unclassified", "Not Classified", "Unknown"].includes(feature.properties["RoadClassification"]));
```
DafyddLlyr, line 50: I actually think this is clearer and more readable than a big XML filter now that there's more than one condition 👍

DafyddLlyr (approved):
Thanks for the follow up and helpful comments here. 

The point raised this morning at standup about displaying this info somewhere other than the Planning Constraints component is a solid reason not to merge this with any other endpoint.

jessicamcinchak:
closing in favor of #1558

PR #1477, opened 2023-02-21
feat: add new component type for sections

jessicamcinchak:
Screenshot from 2023-02-21 08-27-39

Stubs out a basic "Section" component !

To test: 
- The nodeId is added to your breadcrumbs as you navigate through a flow, we can cross reference the flow schema to get the "type" fo this nodeId
- The passport updates `"currentSection"` property when it passes a section node - the passport isn't ultimately a sensible place to track this so totally expect this will change and be replaced by Navigation store methods in the near future, but gives us something simple to work with immediately! 
- Sections are not visible from preview environments, they are always recorded as `"auto": true` in the breadcrumbs


jessicamcinchak:
[01. Add new component type for "Section"](https://trello.com/c/msDJ3SeJ/2292-01-add-new-component-type-for-section)

api.planx.uk/pnpm-lock.yaml
```diff
@@ -7906,7 +7906,7 @@ packages:
     dev: true
 
   file:shared/core-1.0.0.tgz:
-    resolution: {integrity: sha512-nCa6FB4SqTxM0IN5OlM8QYlVEQJlpEZ6DG0esoFzmRNBxeNAsMwYbOi5puH8oIUiyw3Sz8gl+Htpppa/u+P3YA==, tarball: file:shared/core-1.0.0.tgz}
+    resolution: {integrity: sha512-G2n/IbC7IEwKxMoE+aTHXwhTFGSvwNd9YydWGeOVy2vnArpKakBXFsJSj0KWbfkpaVn8ZraWpUH3jAVWfm0Jjg==, tarball: file:shared/core-1.0.0.tgz}
```
jessicamcinchak, line 7909: API & Editor lock file changes and `.tgz` files are a result of running `pnpm distribute` for the first time since pulling down re-named planx-client. It still feels a little odd/wrong to me to have them checked in here, is it okay?

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Node.tsx
```diff
@@ -93,6 +93,10 @@ const Node: React.FC<any> = (props) => {
       );
     case TYPES.Review:
       return <Question {...allProps} text={node?.data?.title ?? "Review"} />;
+    case TYPES.Section:
+      return (
+        <Question {...allProps} text={node?.data?.title ?? "Section marker"} />
+      );
```
jessicamcinchak, line 99: No special styling here yet, but this is where we'll want to edit if we want sections to have a unique Editor style representation. See `hasFailed` and `wasVisited` props on this existing `Question` component as example approaches to styling that we could follow, or we could roll an alternative to this `Question` altogether.

I quickly prototyped this style with an emphasized bottom border, but wasn't sold & figured this is better revisited later.
Screenshot from 2023-02-21 08-45-28

DafyddLlyr: I think this is great for now - one for Alastair an Ian I think!

editor.planx.uk/src/@planx/components/Section/Public.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { useEffect } from "react";
+
+import type { Section } from "./model";
+
+export type Props = PublicProps<Section>;
+
+export default function Component(props: Props) {
+  useEffect(() => {
+    props.handleSubmit?.({
+      ...makeData(props, props.title, "currentSection"), // updates 'currentSection' passport key with this section's title; proxy until we have a Navigation store method!
```
DafyddLlyr, line 12: Helpful comment for the next person here to write this function - I like it.

api.planx.uk/pnpm-lock.yaml
DafyddLlyr: Yeah I'm hitting this also on a lot of PRs - I _think_ we can add them to the `.gitignore` as they'll always get generated by the CI. Is this correct @builditben ?

DafyddLlyr (approved):
Perfect! 👌

PR #1478, opened 2023-02-21
feat: Sections navigation bar

DafyddLlyr:
# What does this PR do?
- Adds a new navigation bar, which displays the current section and total section count
- This is hidden behind the `NAVIGATION_UI` feature flag
- Works when going forwards, backwards, on resume (local storage) and on refresh. Purposefully not tested against Save & Return for now.

## Testing
 - Enable the `NAVIGATION_UI` feature flag
 - Please try to play with this and break things! I'm particularly interested in edge case behaviour when going "Back" or navigating via "Change answer" from a review component
 - There's an example flow with sections here - https://1478.planx.pizza/buckinghamshire/apply-for-a-lawful-development-certificate/preview

https://user-images.githubusercontent.com/20502206/220890745-296e1798-e938-45c3-a8b4-a0fa4ca7cfd7.mov


DafyddLlyr:
[03. User facing feature flag](https://trello.com/c/Yc0CcRvg/2295-03-user-facing-feature-flag)

DafyddLlyr:
[04. Navigation "panel" UI component beneath header (where there are sections)](https://trello.com/c/uWsdjk16/2294-04-navigation-panel-ui-component-beneath-header-where-there-are-sections)

editor.planx.uk/src/components/Header.tsx
```diff
@@ -27,7 +28,7 @@ import { useStore } from "../pages/FlowEditor/lib/store";
 import { rootFlowPath } from "../routes/utils";
 import AnalyticsDisabledBanner from "./AnalyticsDisabledBanner";
 
-export const HEADER_HEIGHT = 75;
+export const HEADER_HEIGHT = 74;
```
DafyddLlyr, line 31: This matches Figma

editor.planx.uk/src/lib/featureFlags.ts
```diff
@@ -2,6 +2,7 @@
 const AVAILABLE_FEATURE_FLAGS = [
   "DISABLE_SAVE_AND_RETURN",
   "ALT_THEME",
+  "NAVIGATION_BAR",
```
jessicamcinchak, line 5: nit: could we name this a bit more generally, eg `SECTIONS_NAVIGATION`, since it will also likely control other UI-bits like showing sections on the Review page or not?

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -125,12 +125,10 @@ export const editorStore: StateCreator<
     (window as any)["doc"] = doc;
 
     await connectToDB(doc);
-    // set the ID of the flow to assist with deciding what to render
-    set({ id });
 
     const cloneStateFromShareDb = () => {
       const flow = JSON.parse(JSON.stringify(doc.data));
-      set({ flow });
+      get().setFlow({ id, flow, flowSlug: get().flowSlug });
```
DafyddLlyr, line 131: `setFlow()` is now being used across all routes to set up the flow, which triggers `initNavigationStore()`

editor.planx.uk/src/pages/FlowEditor/lib/store/navigation.ts
```diff
@@ -1,16 +1,106 @@
...
+    const sectionIds = Object.keys(sectionNodes);
+    const mostRecentSectionId = findLast(
+      breadcrumbIds,
+      (breadcrumbId: string) => sectionIds.includes(breadcrumbId)
+    );
+
+    // This should not happen - if we have sections, we should have a most recent section
+    // Failing noisily for now will help identify issues whilst this is being developed & tested
+    // TODO: Fail more gracefully once feature is complete
+    if (!mostRecentSectionId) throw Error("Error finding mostRecentSectionId");
```
DafyddLlyr, line 84: This condition gets hit if a Section is not the first component in a flow, which I believe is one of the conditions we'll be validating against.

editor.planx.uk/src/pages/FlowEditor/lib/store/navigation.ts
```diff
@@ -1,16 +1,106 @@
...
+   * Get a subset of the full flow, by type
+   * Returned in correct order, based on _root node's edges
+   */
+  filterFlowByType: (type: TYPES): Store.flow => {
+    const flow = get().flow;
+    const rootEdges = flow._root.edges || [];
+    const filteredFlow = Object.fromEntries(
+      Object.entries(flow)
+        .filter(([_key, value]) => value.type === type)
+        .sort(([idA], [idB]) => rootEdges.indexOf(idA) - rootEdges.indexOf(idB))
```
DafyddLlyr, line 102: A flow's node keys are not sorted, but the `edges` array is.

This shouldn't be an expensive operation, LDC flows have 33 items in `_root.edges` currently.

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -344,6 +346,7 @@ export const previewStore: StateCreator<
         });
       }
     }
+    updateSectionData();
```
DafyddLlyr, line 349: I'm not massively keen on adding more complexity and logic to `record()`, but it's consistently called on 'forwards', 'back' and 'change' navigation, which is when we want to check our section.

editor.planx.uk/src/pages/FlowEditor/lib/store/navigation.ts
jessicamcinchak: This is a good shout and wasn't on our initial rules list, but is totally possible. It's a bit tricky because I'd naturally want to check for this on publish, but we actually need it valiated as soon as a section is added to the graph right? And I guess we'd also have to check that when a section is moved or deleted, if there are other sections, one of them is still satisfying the start position?

DafyddLlyr: So without this error, if we just say `if (!mostRecentSectionId) return` things work perfectly fine - whichever your first section is will get displayed first and will stay there until you pass into another section. This means we can chose not to enforce this if we don't want to, but it's just a bit of weird behaviour where the view the editor sees doesn't exactly match the behaviour which I'd rather avoid.

Ideally we'd catch this in the Editor (before publishing). For a plain `add()` I think it's not to bad....
- When adding a section, check if there's a section already at `_root.edges[0]`
  - If so, proceed
  - If not, but this component is being placed there, proceed
  - Otherwise, error
  
I think it's a little more complex on move or delete though like you said, as we'd need to run these checks when other node types and modified.

For now, can we get away with just checking on publish "if there's a section, is there a section at `_root.edges[0]`"?

DafyddLlyr: Sounds perfect 👍

jessicamcinchak: Validating we have a node in first root edge position on publish addressed here https://github.com/theopensystemslab/planx-new/pull/1481/commits/caaab6a37e165094ad320649605a3e2078f2ea07

jessicamcinchak (commented):
These are super solid changes - forward, back, & change from "review" or "result" are working veryy smoothly & exactly as expected for me :tada: 

But I am hitting one issue when testing on the pizza: 
- I have the feature flag toggled on
- I've created & published a test flow with sections https://1478.planx.pizza/braintree/search
- Sections header is _NOT_ appearing in /preview window
- If I add a Send component to my flow, then I can see sections

Only showing sections on "save & return" flows doesn't appear to be encoded business logic here though, and it shouldn't be because we expect sections to be used in FOIYNPP as well as LDCs right? Your mock flow data for unit testing also has section nodes without send, so I'm a little confused what gives and explains the behavior I'm getting on that test link! Curious if you can re-recreate it.

DafyddLlyr:
> But I am hitting one issue when testing on the pizza:

Good catch thank you - this is due to a mistake in types when destructuring the store which I didn't think was possible,  and then another assumption in tests that there was a send component (which you're right, there isn't).

jessicamcinchak (approved):
Working as expected now across all flows !

PR #1480, opened 2023-02-21
fix: Fix storybook pizza

### LINKED ISSUE #1293: Pizza link for storybook is broken
The [Production Storybook](https://storybook.planx.uk/) instance is available but I am unable to find a working instance of a staging or pizza storybook.

Note: this may be an issue with the link rather than the Pizza storybook instance itself.

ghost:
I removed the build job for storybook and added a single command to the react build job (should also mean faster builds :tada:)

Storybooks in pizzas are now exposed as https://storybook.1480.planx.pizza/

This resolves #1293

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -183,7 +183,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ./${{ env.EDITOR_DIRECTORY }}/build
-          key: ${{ runner.os }}-${{ hashFiles('editor.planx.uk/**', '!editor.planx/build/**') }}-react
+          key: ${{ runner.os }}-${{ hashFiles('editor.planx.uk/**') }}
       - uses: pnpm/action-setup@v2.2.2
         if: steps.cache-react-build-assets.outputs.cache-hit != 'true'
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request.yml
```diff
@@ -203,58 +203,19 @@ jobs:
...
         working-directory: ${{ env.EDITOR_DIRECTORY }}
-        # same env as above, if it's job.env it can't access existing env.[variable]
         env:
           REACT_APP_AIRBRAKE_PROJECT_ID: ${{ secrets.AIRBRAKE_PROJECT_ID }}
           REACT_APP_AIRBRAKE_PROJECT_KEY: ${{ secrets.AIRBRAKE_PROJECT_KEY }}
           REACT_APP_API_URL: https://api.${{ env.FULL_DOMAIN }}
           REACT_APP_FEEDBACK_FISH_ID: 65f02de00b90d1
           REACT_APP_HASURA_URL: https://hasura.${{ env.FULL_DOMAIN }}/v1/graphql
           REACT_APP_SHAREDB_URL: wss://sharedb.${{ env.FULL_DOMAIN }}
+          # needed because there's no API to change google's allowed OAuth URLs
```
jessicamcinchak, line 216: :+1: thanks for comments like this

jessicamcinchak (approved):
Link works! Thanks for this fix! :tada: 

On a slightly related note, I was checking deploy settings for another project in the opensystemlab Netlify account this morning & noticed these failed storybook build warnings for what I think was our pre-vultr way of deploying Storybook - these are safe to cleanup, right? (I can do this if so!)
Screenshot from 2023-02-22 09-01-57

jessicamcinchak:
aha right of course - the Netlify site domain is `http://storybook.planx.uk` (so prod, and we don't have a staging equivalent I guess?) - I won't touch it for now! Let me know if you need access to the org on Netlify, I think I can invite you as a member using Github login.

Seems like we can probably still clean up build settings at some point there to not care about PRs since those are handled outside of Netlify by your change here now, but won't remove or change otherwise.

PR #1481, opened 2023-02-22
feat: add checks in the editor for adding, moving, and cloning section nodes

jessicamcinchak:
Rules for working with Sections in the Editor:
- [x] only allowed to add sections on the "center of graph" aka "root" (eg `_root: { edges: ["sectionNodeId"] }`)
  - [x] not allowed to add inside an internal portal (implicitly enforced because parent is portal node type, never `_root`)
  - [x] not allowed to add inside an external portal (this is validated at point of publish because that is when external portals are "flattened" and merged with the outer graph)
- [x] only allowed to move to other positions on the `_root`; not allowed to move onto branches
- [x] cannot clone sections
- [x] when using sections, a section should be in the first position of the graph (validated at point of publish only for now, see discussion here https://github.com/theopensystemslab/planx-new/pull/1478#discussion_r1115510847)

Other changes here:
- [x] show the editor some basic feedback when they violate the rules
- [x] add unit tests

To test: 
- Spend some in test flows on this pizza adding and dragging and cloning sections! (to clone = right click existing node, then click on "open circle" on another branch to attempt to place)
- Confirm that you are seeing error messages when appropriate
- Keep your console open and confirm that there is _not_ an operation logged when errors are thrown

### DIFF
--- a/api.planx.uk/editor/publish.ts
+++ b/api.planx.uk/editor/publish.ts
@@ -4,7 +4,6 @@ import { adminGraphQLClient as adminClient } from "../hasura";
 import { dataMerged, getMostRecentPublishedFlow } from "../helpers";
 import { gql } from "graphql-request";
 import intersection from "lodash/intersection";
-import { flattenDeep } from "lodash";
 
 const diffFlow = async (
   req: Request,
--- a/api.planx.uk/editor/publish.ts
+++ b/api.planx.uk/editor/publish.ts
@@ -4,6 +4,7 @@ import { adminGraphQLClient as adminClient } from "../hasura";
 import { dataMerged, getMostRecentPublishedFlow } from "../helpers";
 import { gql } from "graphql-request";
 import intersection from "lodash/intersection";
+import { flattenDeep } from "lodash";
 
 const diffFlow = async (
   req: Request,
@@ -50,19 +51,23 @@ const publishFlow = async (
   try {
     const flattenedFlow = await dataMerged(req.params.flowId);
 
-    // Check the flattened flow to ensure that all sections (type 360) are still in valid positions (eg within `_root: { edges: [...]}`, not the edge of a portal)
-    //   If any are not, don't proceed with publishing and return a message to display in the Editor
-    const sectionTypeNodeIds = Object.entries(flattenedFlow).filter(([_nodeId, nodeData]) => nodeData?.type === 360).map(([nodeId, _nodeData]) => nodeId);
-    const intersectingNodeIds = intersection(flattenedFlow["_root"].edges, sectionTypeNodeIds);
-    if (intersectingNodeIds.length !== sectionTypeNodeIds.length) {
-      return res.json({
-        alteredNodes: null,
-        message: "Error publishing: found Sections in one or more External Portals"
-      });
+    // If the flattened flow (including external portals) has sections, handle validations
+    if (getSectionNodeIds(flattenedFlow)?.length > 0) {
+      if (!sectionIsInFirstPosition(flattenedFlow)) {
+        return res.json({
+          alteredNodes: null,
+          message: "Error publishing: when using Sections, your flow needs to start with a Section"
+        });
+      } 
+      if (!allSectionsOnRoot(flattenedFlow)) {
+        return res.json({
+          alteredNodes: null,
+          message: "Error publishing: found Sections in one or more External Portals, but Sections are only allowed in main flow"
+        });
+      }
     }
 
     const mostRecent = await getMostRecentPublishedFlow(req.params.flowId);
-
     const delta = jsondiffpatch.diff(mostRecent, flattenedFlow);
 
     if (delta) {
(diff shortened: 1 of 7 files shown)


jessicamcinchak:
[04. Implement editor placement logic](https://trello.com/c/Sr1N5o3o/2301-04-implement-editor-placement-logic)

editor.planx.uk/src/@planx/graph/index.ts
```diff
@@ -158,6 +161,13 @@ const _add = (
 
   draft[id] = sanitize(nodeData);
 
+  if (isSectionNodeType(id, draft) && parent !== ROOT_NODE_KEY) {
+    alert(
+      "cannot add sections on branches or in portals, must be on center of main graph. close this window & try again"
+    );
+    throw new Error("cannot add sections on branches or in portals");
+  }
```
jessicamcinchak, line 169: when calling `move()` or `clone()`, thrown errors are immediately raised as user-facing alert dialogs - but `_add()` only logs them, so I've additionally added an alert here. 

it'd be most ideal to automatically close the section dialog in the Editor on error, but i think this should be fine for now!

editor.planx.uk/src/@planx/graph/index.ts
```diff
@@ -209,6 +220,8 @@ export const clone =
       else if (!draft[toParent]) throw new Error("toParent not found");
       else if (draft[toParent].edges?.includes(id))
         throw new Error("cannot clone to same parent");
+      else if (isSectionNodeType(id, graph))
+        throw new Error("cannot clone sections");
```
jessicamcinchak, line 224: cloning is a bit of an interesting case here because as long as our `_add()` & `move()` rules for sections are working as expected, then cloning would be implicitly prohibited because you can never clone to the same parent (eg always `_root` in the case of a valid section node) - but feels good to be explicit about intended behavior here to me

api.planx.uk/editor/publish.ts
```diff
@@ -48,6 +49,18 @@ const publishFlow = async (
...
+
+    // Check the flattened flow to ensure that all sections (type 360) are still in valid positions (eg within `_root: { edges: [...]}`, not the edge of a portal)
+    //   If any are not, don't proceed with publishing and return a message to display in the Editor
+    const sectionTypeNodeIds = Object.entries(flattenedFlow).filter(([_nodeId, nodeData]) => nodeData?.type === 360).map(([nodeId, _nodeData]) => nodeId);
+    const intersectingNodeIds = intersection(flattenedFlow["_root"].edges, sectionTypeNodeIds);
+    if (intersectingNodeIds.length !== sectionTypeNodeIds.length) {
+      return res.json({
+        alteredNodes: null,
+        message: "Error publishing: found Sections in one or more External Portals"
+      });
```
jessicamcinchak, line 61: I opted to not actually return a true HTTP error here, but rather return a message that is shown to editors. Once you adjust your content, clicking "publish" will succeed again
Screenshot from 2023-02-23 11-37-39

I also imagine we'll definitely want to rethink how/where "validation checks" happen during publishing once we have > 1 ! But checking directly in the publish method feels okay to me for this first, rather simple, check.

But open to suggestions here!

DafyddLlyr (approved):
Tests are super helpful here, clearly documenting the expected behaviour.

Tested all behaviours on the Pizza and go expected results 🎉

api.planx.uk/editor/publish.ts
```diff
@@ -48,6 +49,18 @@ const publishFlow = async (
 
   try {
     const flattenedFlow = await dataMerged(req.params.flowId);
+
+    // Check the flattened flow to ensure that all sections (type 360) are still in valid positions (eg within `_root: { edges: [...]}`, not the edge of a portal)
+    //   If any are not, don't proceed with publishing and return a message to display in the Editor
+    const sectionTypeNodeIds = Object.entries(flattenedFlow).filter(([_nodeId, nodeData]) => nodeData?.type === 360).map(([nodeId, _nodeData]) => nodeId);
```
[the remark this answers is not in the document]
jessicamcinchak: Yep, good call! Now refactored to be two separate validation functions `sectionIsInFirstPosition` & `allSectionsOnRoot` which each return a boolean. These are still living in the `publish.ts` file since they're quite specific, not wider "helpers" etc. But easier to imagine how this might evolve into a growing list of validations this way :+1:

api.planx.uk/editor/publish.ts
jessicamcinchak: On the HTTP error part, returning success with null `alteredNodes` & a `message` is currently consistent with what happens when the publish endpoint doesn't detect any changes to publish (which is also a sort of fuzzy success or error scenario in that the actual publish mutation is skipped!)

If/when we start throwing actual errors, we'll have to rework how `setLastPublishedTitle` works in the preview browser which controls messages shown to editors, which feels a bit bigger scope than here but a worthwhile future rethink & todo for sure! More related to publishing/editor work in general than sections in my opinion

editor.planx.uk/src/@planx/graph/index.ts
jessicamcinchak: Going to note this one in a future card, and merge as-is - but appreciate these ideas!

My quick attempt at Introducing `navigate()` into graph/index methods is throwing invalid hooks errors and feels a little wrong still (like _should_ navigation/routing be a responsibility of these methods?). And I don't feel quite ready to figure out how to render the error in the Editor modal yet!

PR #1483, opened 2023-02-22
docs(infra): create application/README.md

gunar:
Trying to keep this simple, and link to other places instead of writing down all the details, as to avoid duplicate information spread across the stack. That being said, feedback is welcome.

PR #1484, opened 2023-02-23
fix: add type declaration for planx-document-templates for successful CI

jessicamcinchak:
Recent PRs & local docker "up"s have been failing on the Build React App step (see #1482, #1483 etc)

Based on the timing and error, I think this was likely introduced on this merge - https://github.com/theopensystemslab/planx-document-templates/pull/11 ? I'm not entirely sure if intentional or not (I definitely missed during code review!), so this might be best addressed as a two-part/two-repo fix, but this should clear up our planx-new CI in the meantime :checkered_flag:

### DIFF
--- a/api.planx.uk/@types/planx-document-templates.d.ts
+++ b/api.planx.uk/@types/planx-document-templates.d.ts
@@ -1,0 +1,1 @@
+declare module "@opensystemslab/planx-document-templates";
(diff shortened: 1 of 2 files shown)


DafyddLlyr (approved):
Cheers for looking into this one 👍

PR #1485, opened 2023-02-23
feat: Save session on payment responses

builditben:
Save a lowcal session before redirecting to GovPay.

This PR introduces a change which adds a lowcal session entry on pay interactions.

I considered only updating the lowcal session on new payments or only on success but decided to update on every request to resolve the payment response from the proxy. This means that the lowcal session is always up to date with the latest payment information (including status). GovPay will only allow one payment at a time so keeping track of the latest payment requested will always result in tracking any payments that actually succeed.

To test this in detail, you would need to set up some additional logging to validate the payment_id tracks correctly. The audit table is also a helpful tool for testing here. Have a look at the E2E test scenarios which should cover all relevant test cases (let me know if you think something is missing).

**EDIT**
Also, a user should not be able to query the API to retrieve stored session information (unless this is also part of a save-and-return journey and they provide their email address).

### DIFF
--- a/editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
@@ -1,10 +1,11 @@
 import { logger } from "airbrake";
 import axios from "axios";
 import DelayedLoadingIndicator from "components/DelayedLoadingIndicator";
+import { setLocalFlow } from "lib/local.new";
 import { useStore } from "pages/FlowEditor/lib/store";
 import { handleSubmit } from "pages/Preview/Node";
 import React, { useEffect, useReducer } from "react";
-import type { GovUKPayment } from "types";
+import type { GovUKPayment, Passport, Session } from "types";
 import { PaymentStatus } from "types";
 
 import { useTeamSlug } from "../../shared/hooks";
@@ -47,6 +48,7 @@ function Component(props: Props) {
   const [
     flowId,
     sessionId,
+    breadcrumbs,
     govUkPayment,
     setGovUkPayment,
     passport,
@@ -54,26 +56,14 @@ function Component(props: Props) {
   ] = useStore((state) => [
     state.id,
     state.sessionId,
+    state.breadcrumbs,
     state.govUkPayment,
     state.setGovUkPayment,
     state.computePassport(),
     state.previewEnvironment,
   ]);
-
-  const fee = props.fn ? Number(passport.data?.[props.fn]) : 0;
-
   const teamSlug = useTeamSlug();
-
-  const getGovUkPayUrlForTeam = (paymentId?: string | undefined) => {
-    const baseURL = useStagingUrlIfTestApplication(passport)(
-      `${GOV_UK_PAY_URL}/${teamSlug}`
-    );
-    const queryString = `?sessionId=${sessionId}&flowId=${flowId}`;
-    if (paymentId) {
-      return `${baseURL}/${paymentId}${queryString}`;
-    }
-    return `${baseURL}${queryString}`;
-  };
+  const fee = props.fn ? Number(passport.data?.[props.fn]) : 0;
 
   // Handles UI states
   const reducer = (_state: ComponentState, action: Action): ComponentState => {
@@ -124,7 +114,7 @@ function Component(props: Props) {
       handleSuccess();
     } else {
       dispatch(Action.IncompletePaymentFound);
-      refetchPayment(govUkPayment.payment_id);
+      refetchPayment();
     }
   }, []);
 
@@ -136,12 +126,8 @@ function Component(props: Props) {
   const normalizePaymentResponse = (responseData: any): GovUKPayment => {
     if (!responseData?.state?.status)
       throw new Error("Corrupted response from GOV.UK");
-    let payment: GovUKPayment = responseData;
-    if (payment.amount)
-      payment = {
-        ...payment,
-        amount: toDecimal(payment.amount),
-      };
+    let payment: GovUKPayment = { ...responseData };
+    payment.amount = toDecimal(payment.amount);
     return payment;
   };
 
@@ -150,15 +136,29 @@ function Component(props: Props) {
   ): Promise<GovUKPayment> => {
     const payment = normalizePaymentResponse(responseData);
     setGovUkPayment(payment);
+    // save a record of the session with the latest payment for debugging purposes
+    await saveSession({
+      breadcrumbs,
+      id: flowId,
+      passport,
+      sessionId,
+      govUkPayment: payment,
+    });
     return payment;
   };
 
-  const refetchPayment = async (id: string) => {
+  const refetchPayment = async () => {
     try {
       const {
         data: { state },
       } = await axios.get<Pick<GovUKPayment, "state">>(
-        getGovUkPayUrlForTeam(id)
+        getGovUkPayUrlForTeam({
+          sessionId,
+          flowId,
+          teamSlug,
+          passport,
+          paymentId: govUkPayment?.payment_id,
+        })
       );
 
       // Update local state with the refetched payment state
@@ -185,7 +185,7 @@ function Component(props: Props) {
     dispatch(Action.ResumePayment);
 
     if (!govUkPayment) {
-      startNewPayment();
+      await startNewPayment();
       return;
     }
 
@@ -193,7 +193,7 @@ function Component(props: Props) {
       case PaymentStatus.cancelled:
       case PaymentStatus.error:
       case PaymentStatus.failed: {
-        startNewPayment();
+        await startNewPayment();
         break;
       }
       case PaymentStatus.started:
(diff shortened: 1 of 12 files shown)


editor.planx.uk/src/lib/lowcalStorage.ts
```diff
@@ -145,7 +146,9 @@ const getPublicContext = (sessionId: string) => ({
   context: {
     headers: {
       "x-hasura-lowcal-session-id": sessionId,
-      "x-hasura-lowcal-email": useStore.getState().saveToEmail?.toLowerCase(),
+      "x-hasura-lowcal-email":
```
DafyddLlyr, line 149: Email address cannot and should not be unique in this table as it's perfectly valid for one email address to be associated with multiple sessions.

If we allow `""` as an email address, the `public` role can perform `SELECT` and `UPDATE` operations on any record where the email is `""` if they know the id, via our GraphQL API endpoint. The permissions for this are defined in `tables.yaml`

Example from Pizza using request headers - 

Screenshot 2023-02-24 at 18 20 34

I know we did speak about this a while back sorry, and I don't remember the conclusion we came to exactly - I might be treading over ground we've already covered. Is there a current use case for having a "Pay" component in a flow without a "Send" component? 

If there is, could we solve this by also checking for "Pay" components in `isSaveReturnFlow()` as opposed to loosening this restriction? This would mean we'd always have an email address to associate with a session, but there are implications in how we'd handle a session being completed as we'd need to sub the `submitted_at` logic with a `completed_at` / `paid_at` or similar.

DafyddLlyr: Actually, another (simpler and more flexible?) option here is to update the permissions to allow `SELECT` and `UPDATE` if... 
 - You know the session_id
 - You know the email address
 - The email address is not `""`

This would allow a non Save and Return flow to allow payments, keep sessions and audit logs, and not risk exposing session data.

These rules are set in the "Permission" tab of a table in Hasura (Data > lowcal_sessions > Permission > Public > Update / Select > Column / Row update rules).

If the above makes sense, I'd suggest going for this and trying it out 👍

editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
```diff
@@ -268,6 +271,35 @@ function Component(props: Props) {
       ) : (
         <DelayedLoadingIndicator text={state.displayText || state.status} />
       )}
+      {/* session id exposed for testing purposes */}
```
DafyddLlyr, line 274: This can also be accessed via the store (exposed on `window` if that can be accessed via Playwright) or the URL which might be a more repeatable solution that potentially adding `data-sessionid` across multiple elements in future (I could see a lot of tests relying on it).

editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
```diff
@@ -185,15 +185,15 @@ function Component(props: Props) {
     dispatch(Action.ResumePayment);
 
     if (!govUkPayment) {
-      startNewPayment();
+      await startNewPayment();
```
DafyddLlyr, line 188: The `await` here seems a good spot and something we've missed before 👍

DafyddLlyr (commented):
Didn't manage to finish an in depth review of this before I finished up sorry, but I've added a few comments on the permissions for the `lowcal_session` table which may be worth looking at 🙂

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -206,6 +206,8 @@
                 _is_null: true
             - deleted_at:
                 _is_null: true
+            - email:
+                _neq: ""
```
DafyddLlyr, line 210: Perfect 👌 

So that the full behaviour of this table and associated permissions are documented in one place, can we add a very simple test in `hasura.planx.uk/tests/lowcal_sessions.test.js` that verifies that the Public user cannot access an session with an email of `""`?

DafyddLlyr (changes requested):

DafyddLlyr:
One small thing I flagged with regards to tests, but this is looking great and happy to approve this once sorted.

The approach used here makes sense to me 👍

hasura.planx.uk/.env.test.example
```diff
@@ -1,4 +1,5 @@
 # Used in local testing
 
 HASURA_ADMIN_SECRET=👻
+HASURA_HOST=0.0.0.0
```
builditben, line 4: I also updated this in the pizza secrets on S3

hasura.planx.uk/tests/lowcal_sessions.test.js
```diff
@@ -193,10 +254,42 @@ describe("lowcal_sessions", () => {
...
+        const res2 = await gqlPublic(updateByPK, { sessionId: anon1, data: { y: 2 } }, headers);
+        expect(res2.data.update_lowcal_sessions_by_pk).not.toBeNull();
+        expect(res2.data.update_lowcal_sessions_by_pk.id).toEqual(anon1);
+        expect(res2.data.update_lowcal_sessions_by_pk.data).toHaveProperty("y", 2);
+
+        // update 2
+        const res3 = await gqlPublic(updateByPK, { sessionId: anon1, data: {} }, headers);
+        expect(res3.data.update_lowcal_sessions_by_pk).not.toBeNull();
+        expect(res3.data.update_lowcal_sessions_by_pk.data).toEqual({});
+      });
```
builditben, line 289: I was surprised by the behavior here. I would have expected not to be able to read back updated data but I can only read what I have just written (and the deleted_at timestamp) and these tests show that is the case.

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -181,12 +181,14 @@
...
       permission:
-        check: {}
         columns:
           - data
           - email
           - flow_id
           - id
+        check:
+          email:
+            _eq: x-hasura-lowcal-email
```
builditben, line 191: I've added an additional insert check (and associated test)

hasura.planx.uk/tests/utils.js
```diff
@@ -1,7 +1,7 @@
 const fetch = require("isomorphic-fetch");
 
 async function gqlAdmin(query, variables = {}) {
-  const res = await fetch(`http://0.0.0.0:${process.env.HASURA_PORT}/v1/graphql`, {
+  const res = await fetch(`http://${process.env.HASURA_HOST}:${process.env.HASURA_PORT}/v1/graphql`, {
```
builditben, line 4: I was trying to run this locally and have updated the environment variables here to enable that.

hasura.planx.uk/tests/lowcal_sessions.test.js
```diff
@@ -207,6 +300,16 @@ describe("lowcal_sessions", () => {
...
+
+      test("Alice can update her session with an empty email", async () => {
+        const headers = {
+          "x-hasura-lowcal-session-id": alice1,
+          "x-hasura-lowcal-email": ""
+        };
+        const res = await gqlPublic(updateByPK, { sessionId: alice1, data: { x: 1 } }, headers);
+        expect(res).not.toHaveProperty("errors");
+        expect(res.data.update_lowcal_sessions_by_pk).toBeNull();
+      });
```
builditben, line 312: This is unintended behavior but worth noting. This change means that existing sessions can be updated (but not read back) without an email.

hasura.planx.uk/tests/lowcal_sessions.test.js
DafyddLlyr: So I think I worked this one out - we've modified `select_permissions` by adding the ` email: _neq: ""` check but not the `update_permissions` for this table.

This isn't a public user action, so I think it's fine to allow users to insert records without an email, but not allow them to update? This would mean that it's not possible for the public role to overwrite these audit records. Does that make sense?

builditben: @DafyddLlyr if I remove permissions to update, we would only be able to insert a record capturing the payment information for the session once (because of the pk constraint). This would be an issue for cancelled and retried payments where the use also abandons the payment (the original scenario that triggered the work in this ticket) - there would be no way to know when the payment is going to be successful beforehand so no way to know which session to save before a potential abandonment scenario.

We've circled round this a few times and it sounds like you're uncomfortable with these public permissions for this particular edge case. My take is that a potential bad actor tampering with debug data in a Pay flow without a send component is not really a reasonable threat but I appreciate your zero-trust approach.

This could be an API feature if this approach isn't right. I'd also be happy to pass the baton to someone else in the team on this work if there's a better solution that I'm not seeing :runner: :runner:

DafyddLlyr: No thanks for clarifying that, you're right that the update is a requirement here.

I think we've both taken a detailed deep-dive here and understand the approach and with it's associated (very minor downsides).

I'm more than happy for this to proceed as is - the behaviour is well understood and documented with tests, and it's not something which can result in any practical risks. I just wanted to be sure we weren't missing anything here 👍 

✅ ✅ ✅

DafyddLlyr (approved):

PR #1486, opened 2023-02-23
chore: Lock planx-document-templates to a fixed version

ghost:
We normally update packages explicitly and I think that should be the case for planx-document-templates too. It does add a little bit of additional work but it means that updates have a chance to be sense-checked and released explicitly.

The ideal solution for managing multiple interdependent packages would be to use some monorepo tooling. Maybe its worth exploring those instead of some of these less-than-ideal workflows (or we just put up with the wonky workflow for the time-being).

If we do want to just track main for this pacakge, do we also need a testing workflow where we link the module in one of the consumers, install and build to avoid directly breaking main in planx-document-templates?.

Are there other solutions to avoiding issues and laborious workflows here?

### DIFF
--- a/api.planx.uk/README.md
+++ b/api.planx.uk/README.md
@@ -22,6 +22,10 @@ Start the development server `pnpm dev`
 
 Run tests `pnpm test`
 
+Development notes:
+
+ - if you need to test or pull new changes from @opensystemslab/planx-document-templates, make sure to update the commit hash in package.json first
+
 ## Prior art
 
 https://github.com/theopensystemslab/planx-api
--- a/editor.planx.uk/README.md
+++ b/editor.planx.uk/README.md
@@ -12,6 +12,10 @@ Start the development server `pnpm start`
 
 Run tests `pnpm test`
 
+Development notes:
+
+ - if you need to test or pull new changes from @opensystemslab/planx-document-templates, make sure to update the commit hash in package.json first
+
 ### Disabling type-checking
 
 If your IDE does type checking for you and you want to save some CPU resources,


api.planx.uk/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@airbrake/node": "^2.1.7",
-    "@opensystemslab/planx-document-templates": "git://github.com/theopensystemslab/planx-document-templates#main",
+    "@opensystemslab/planx-document-templates": "git://github.com/theopensystemslab/planx-document-templates#e0642be",
```
jessicamcinchak, line 7: this approach totally works for me for now until we have bigger picture convos about monorepo tooling or a testing module in place like you describe. i'm on board with the wonky workflow in the shorterm, especially as document templates requests hopefully slow down soon and won't be expected to change that often.

i am wondering though if it's worth adding a sentence about this to each the API & Editor READMEs in planx-new? something like: 

- `pnpm i` installs dependencies
  - if you need to test or pull new changes from `@opensystemslab/planx-document-templates`, make sure to update the commit hash in package.json first

editor.planx.uk/types.d.ts
```diff
@@ -9,4 +9,3 @@ declare module "draftjs-to-html";
 declare module "mathjs";
 declare module "@opensystemslab/map";
 declare module "wkt";
-declare module "@opensystemslab/planx-document-templates";
```
jessicamcinchak, line 12: :+1: thanks for cleaning this up

jessicamcinchak (approved):

PR #1488, opened 2023-02-23
chore: Drop unused ModalCard component

DafyddLlyr:
Would appreciate another pair of eyes here - came across this component which doesn't appear to be used or imported anywhere.

There are no other instances of "ModalCard" in the codebase (e.g. as a component name, or in an import path).


jessicamcinchak (approved):
Thanks, agree this is unused & helpful to have one less "card" around!

PR #1489, opened 2023-02-23
fix: Handle missing node id in Editor route

DafyddLlyr:
# What does this PR do?
- Adds an additional validation step in the Editor to ensure that users cannot navigate to an invalid route

## Context
See PlanX Slack here - https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1677159875899169

To recreate - 
- Delete a node
- Hit back in the browser so your URL is now `…nodes/<DELTED_NODE_ID>/edit`
- Invalid URL - Error! ❌ 

The above steps cannot be recreate on the Pizza, invalid routes will just redirect to the main flow.

### DIFF
--- a/editor.planx.uk/src/routes/flow.tsx
+++ b/editor.planx.uk/src/routes/flow.tsx
@@ -2,7 +2,17 @@ import { gql } from "@apollo/client";
 import { TYPES } from "@planx/components/types";
 import ErrorFallback from "components/ErrorFallback";
 import natsort from "natsort";
-import { compose, lazy, mount, route, withData, withView } from "navi";
+import {
+  compose,
+  lazy,
+  map,
+  Matcher,
+  mount,
+  redirect,
+  route,
+  withData,
+  withView,
+} from "navi";
 import mapAccum from "ramda/src/mapAccum";
 import React from "react";
 import { ErrorBoundary } from "react-error-boundary";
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/routes/flow.tsx
```diff
@@ -86,58 +96,69 @@ const newNode = route(async (req) => {
...
-            title: category.title,
-            children: childNodes.slice(index, index + category.count),
-          },
-        ],
-        0,
-        node.data.categories
-      )[1];
-    } else {
-      extraProps.options = childNodes;
+const editNode = validateEdit(
```
DafyddLlyr, line 108: The diff here is not super helpful - the only change to this function is that it's now wrapped in `validateEdit()`

editor.planx.uk/src/routes/flow.tsx
```diff
@@ -86,58 +96,69 @@ const newNode = route(async (req) => {
...
-  if (node.type === TYPES.ExternalPortal)
-    extraProps.flows = await getExternalPortals();
+// If nodeId is invalid (e.g. a deleted node), fall back to flow
+const validateEdit = (route: Matcher<object, object>) =>
+  map((req) => {
+    const { team, flow, id } = req.params;
+    const node = useStore.getState().getNode(id);
+    if (!node) return redirect(`/${team}/${flow}`);
+    return route;
+  });
```
jessicamcinchak, line 106: nit: is `validateNodeRoute` slightly more descriptive here? the original name `editNode` feels easy to confuse with graph/index methods to me, and I think this original function is really describing the action `openNode` (eg `getState()`) not the actual editing or mutating (eg click "Update <node>", call `setState()`) right?

jessicamcinchak (approved):
Thanks for this one, works as expected! 

Also tested the scenario of: 
- Add an external portal to flow "A"
- Delete that external portal flow from the teams/services list
- Navigate back to flow "A", confirm that portal node shows as corrupted but it CAN still be clicked/opened so you have option to "delete" (which is good!)
- Then use browser "back" from here on and get behavior consistent wilth your testing instructions :heavy_plus_sign:

editor.planx.uk/src/routes/flow.tsx
DafyddLlyr: Very helpful comment thanks - will make the change before I merge.

PR #1505, opened 2023-03-03
feat(wip): save & return happy path reconciliation with sections

jessicamcinchak:
Changes: 
- ability to show section status tag on SummaryLists (eg Review/Reconciliation)
Screenshot from 2023-03-03 16-05-20

Open questions / very much not-working yet:
- should we really just show a Sections Overview page (like the Section node) when resuming rather than full SummaryList?
- when I originally wrote `sortedBreadcrumbsBySections` I was assuming there should be a chunk of breadcrumbs for _every_ id in `sectionNodes`, which works well on the Review page when all section markers would be _before_ review in the graph, but not for Reconciliation - so need to re-work this to not include sections that a save&return user hasn't seen yet (you'll see buggy empty tags/headers in Reconciliation page currently that break accessible heirarchy)
- handling content changes / introducting "Needs updated" status (to be handled in separate PR/ticket! https://trello.com/c/EBcNOMg8/2310-10-in-sr-flows-how-does-reconciliation-display-if-content-changes)


jessicamcinchak:
[09. In S&R flows, how does reconciliation display if no content changes](https://trello.com/c/OI6u5662/2309-09-in-sr-flows-how-does-reconciliation-display-if-no-content-changes)

jessicamcinchak:
Closing this one in favor of #1535 after talking through in dev call! 

"Needs updated" status applies to section, not individual questions, so we decided this display is more visually confusing & cluttered than the refined overview display.

PR #1506, opened 2023-03-03
feat: Confirmation page "application type" reflects service name

jessicamcinchak:
Originally thought we should just make this configurable in the Editor, but that's annoyingly tricky currently because "Application type" is one row within a table of data points generated by `props.details`.

So, instead, trying this which I think still satisfies the original problem! 

**"Application type" will reflect the `flowName` with a few rules applied:**
- If the flowName is phrased as "apply-for-", we'll display the application type as "Application for..." (this service name pattern is agreed upon by content team according to govuk standards, so unlikely to change I think!)
- If there's a "work status" in the passport (eg "existing" or "proposed"), it'll be tacked onto the end of the flow name

On production, on an LDC Confirmation page, now: "Application for a Certificate of Lawfulness - Proposed"
After this change: "Application for a lawful development certificate - proposed"

Casing changes, but will actually be _more_ consistent with what is sent via emails etc now since it's calling flowName from state, so hopefully fine?

### DIFF
--- a/editor.planx.uk/src/pages/Preview/Node.tsx
+++ b/editor.planx.uk/src/pages/Preview/Node.tsx
@@ -53,6 +53,7 @@ const Node: React.FC<any> = (props: Props) => {
     resetPreview,
     sessionId,
     cachedBreadcrumbs,
+    flowName,
   ] = useStore((state) => [
     state.childNodesOf,
     state.resultData,
@@ -62,6 +63,7 @@ const Node: React.FC<any> = (props: Props) => {
     state.resetPreview,
     state.sessionId,
     state.cachedBreadcrumbs,
+    state.flowName,
   ]);
 
   const handleSubmit = isFinalCard ? undefined : props.handleSubmit;
@@ -111,14 +113,12 @@ const Node: React.FC<any> = (props: Props) => {
         passport.data?.[GOV_PAY_PASSPORT_KEY];
 
       const details = {
-        "Planning Application Reference": payment?.reference ?? sessionId,
+        "Planning application reference": payment?.reference ?? sessionId,
 
-        "Property Address": passport.data?._address?.title,
+        "Property address": passport.data?._address?.title,
 
         "Application type": [
-          // XXX: application type currently hardcoded as it's the only one being
-          //      tested, but this will need to become dynamic.
-          "Application for a Certificate of Lawfulness",
+          flowName.replace("Apply", "Application"),
           getWorkStatus(passport),
         ]
           .filter(Boolean)


jessicamcinchak:
[Make 'application type' editable in the confirmation page component](https://trello.com/c/fB4FFCIh/2313-make-application-type-editable-in-the-confirmation-page-component)

builditben (commented):
Could we fix the flowName in the store (i.e. make it a function call) rather than just amending the confirmation page? That way, we document the applied rule somewhere that's easier to find and we also enable this to be reused consistently.

Do we also need to consider exported data where the application type is set?

jessicamcinchak:
@builditben - thanks for taking a look at this one! 

a couple responses to your questions: 
- `setFlow` is the shared store function that calls & sets the `flowName` (& other flow-identifiers like id & slug) when a flow is loaded, therefore I don't think `flowName` needs to be a function call itself? (it should never change in a given session after that initial load)
  - `flowName` at the store-level handles replacing `-` & casing, but I think the application page is the only place where we'll want "Application" with the concatenated "work status", so that why that's handled directly & uniquely here. Honestly I'd guess it's also fine to just use `flowName` directly on this page too, but was aiming for feature parity closer to existing content since folks pointed that out during planning
- export data formats are a good question but they're already handled / aren't effected by this! `applicationType.test.ts` in Send/BOPS is a clear example of how this was & continues to work after this change
  - See `getBOPSParams` (called when generating CSV data structure too) where `application_type` defaults to `"lawfulness_certificate"` (the _only_ valid value BOPS accepts) or will use the `flowName` from state already for non-LDCs. Again, similar to Notify email "service names", I think it's correct for right now that these continue consistently reflecting the active `flowName` (eg "Apply" not "Application") from state until content team says otherwise

does that help clarify decisions here?

builditben:
@jessicamcinchak Thanks for clarifying. You're right, `flowName` probably doesn't need to be a function call. What I was really getting at was trying to encapsulate the transformation logic in one place (perhaps `setFlow`?) so that there's a clear split between data/content rules and presentation/layout.

It also makes sense that BOPs isn't impacted here but the CSVData is now used for the Overview.html which may be included in send-to-email applications. Since this has the potential to be used outside of LDC services, should we consider not hard-coding the value there too?

It might be my misunderstanding but this change looks like it is still only going to present the right text for LDC applications. I thought the ticket was about enabling editors to set the name of a service so that this confirmation page will work for other things like prior approval. Is that fair?

builditben:
Okay, I see now that the `getWorkStatus` is going to be empty and filtered out so this will work for prior approval. Is it still the case that we need to consider CSVData?

builditben (approved):
I've followed this through a little more and I see that `makeCsvData` calls `getBOPSParams` which does, like you mentioned, allow for other flowName values. You're right, this is covered already (though a little hard to follow!)

PR #1510, opened 2023-03-07
feat: Add warning text to Pay component

DafyddLlyr:
Screenshot 2023-03-07 at 09 18 36

### DIFF
--- a/editor.planx.uk/src/@planx/components/Pay/Public/Confirm.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Public/Confirm.tsx
@@ -31,6 +31,15 @@ interface Props {
   error?: string;
 }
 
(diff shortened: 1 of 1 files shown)


DafyddLlyr:
[Add a content warning when user leaves to GOV.UK Pay asking user not to close the browser when returning to this service](https://trello.com/c/FPattzmt/2250-add-a-content-warning-when-user-leaves-to-govuk-pay-asking-user-not-to-close-the-browser-when-returning-to-this-service)

builditben (approved):
I saw that this was Alice's suggestion so approving as is. :+1: 

The text here implies that the correct journey is to close the browser after completing the GovPay payment which isn't quite right.

Here's my suggestion:
"Once you have made a payment on GovPay you will be redirected here to complete and send your application."

PR #1512, opened 2023-03-07
feat: Handle sections in flow reconciliations

builditben:
This adds handling for section reconciliation. My understanding of how this should behave is described in the automated tests - please review those tests and also let me know if there are any additional behaviors that need testing here. 

This depends on a change to planx-core here: https://github.com/theopensystemslab/planx-core/pull/2

### DIFF
--- a/api.planx.uk/saveAndReturn/validateSession.test.ts
+++ b/api.planx.uk/saveAndReturn/validateSession.test.ts
@@ -157,10 +157,123 @@ describe("Validate Session endpoint", () => {
       });
   });
 
+  // TODO this may be an existing issue or it may be my misunderstanding of how this should work
+  it.skip("returns changed nodes and invalidated breadcrumbs when a flow has updated only nested nodes (answers but not questions)", async () => {
+    const oldFlow: Flow["data"] = {
+      _root: {
+        edges: ["question1", "question2"],
+      },
+      question1: {
+        data: { fn: "question", text: "Is it 'one' or 'two'" },
+        type: 100,
+        edges: ["one", "two"],
+      },
+      one: {
+        data: { val: "answer", text: "One" },
+        type: 200,
+      },
+      two: {
+        data: { val: "answer", text: "Two" },
+        type: 200,
+      },
+      question2: {
+        data: { fn: "question", text: "Is it 'A' or 'B'" },
+        type: 100,
+        edges: ["a", "b"],
+      },
+      a: {
+        data: { val: "answer", text: "A" },
+        type: 200,
+      },
+      b: {
+        data: { val: "answer", text: "B" },
+        type: 200,
+      },
+    };
+    const newFlow: Flow["data"] = {
+      ...oldFlow,
+      one: {
+        data: { val: "answer", text: "One (1)" },
+        type: 200,
+      },
+      two: {
+        data: { val: "answer", text: "Two (2)" },
+        type: 200,
+      },
+      b: {
+        data: { val: "answer", text: "B or Other" },
+        type: 200,
+      },
+    };
+    const breadcrumbs: Breadcrumb = {
+      question1: {
+        auto: false,
+        answers: ["two"],
+      },
+      question2: {
+        auto: false,
+        answers: ["b"],
+      },
+    };
+    mockQueryWithFlowDiff({ oldFlow, newFlow, breadcrumbs });
+
+    const data = {
+      payload: {
+        sessionId: mockLowcalSession.id,
+        email: mockLowcalSession.email,
+      },
+    };
+
+    const expectedDiff = [
+      {
+        id: "one",
+        data: { val: "answer", text: "One (1)" },
+        type: 200,
+      },
+      {
+        id: "two",
+        data: { val: "answer", text: "Two (2)" },
+        type: 200,
+      },
+      {
+        id: "b",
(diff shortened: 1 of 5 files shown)


builditben:
[10. In S&R flows, how does reconciliation display if content changes](https://trello.com/c/EBcNOMg8/2310-10-in-sr-flows-how-does-reconciliation-display-if-content-changes)

jessicamcinchak (commented):
This is looking super solid, left a couple initial nit picky naming things but tests are very explainative and helpful and doing exactly what I'd expect.

Going to properly look through associated planx-core change and spend a little time on this pizza before giving it a green tick if that's alright

api.planx.uk/saveAndReturn/validateSession.ts
```diff
@@ -54,6 +56,11 @@ export async function validateSession(
         });
       }
 
+      const sortedBreadcrumbs = sortBreadcrumbs(
+        savedFlow,
+        sessionData.data.breadcrumbs
+      );
```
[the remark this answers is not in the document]
builditben: Eventually, it'd be nice if this API returned ordered changes but, for now, I've clarified the naming a little more here and also added an additional comment.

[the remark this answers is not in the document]
jessicamcinchak: new naming here feels a lot more clear to me, thanks :+1:

api.planx.uk/saveAndReturn/validateSession.ts
```diff
@@ -62,16 +65,31 @@ export async function validateSession(
...
                 sessionData.data.breadcrumbs[node.id!];
               delete sessionData.data.breadcrumbs[node.id!];
+
+              // also remove the associated section
+              const affectedSectionId = affectedBreadcrumb.sectionId;
+              if (affectedSectionId) {
+                removedBreadcrumbs[affectedSectionId] =
+                  sessionData.data.breadcrumbs[affectedSectionId];
+                delete sessionData.data.breadcrumbs[affectedSectionId];
+              }
```
DafyddLlyr, line 92: How does this hold up to multiple changes nodes within the same section? I _think_ `sessionData.data.breadcrumbs[affectedSectionId]` will be `undefined` on the second pass.

api.planx.uk/saveAndReturn/validateSession.ts
DafyddLlyr, line 1: Changes here are clear - the naming conventions really help. 

The `validateSession()` function is pretty big and unwieldy for my tastes for a refactor of this can 100% come at a later date.

jessicamcinchak (approved):
thanks for minor naming adjustments. tested more on pizza and no showstoppers. happy for this to merge whenever! expecting it to link up nicely with with the overview page and statuses over in #1535 :+1:

DafyddLlyr (commented):
One small question about multiple changes within the same section. Taking a look at the planx-core PR just now 👍

api.planx.uk/saveAndReturn/validateSession.ts
builditben: @DafyddLlyr Thanks for pointing out the bug in these changes. 

I resisted the urge to refactor this but I find this imperative style very hard to follow. A refactor would be very welcome from my point of view.

I've now added an additional test with what I think is a pre-existing bug with nested nodes: content change for just answers doesn't bubble up to remove question breadcrumbs. Did I get the test right or did I misunderstand how this works?

api.planx.uk/saveAndReturn/validateSession.ts
builditben: Good catch :+1:

api.planx.uk/saveAndReturn/validateSession.test.ts
```diff
@@ -155,16 +156,410 @@ describe("Validate Session endpoint", () => {
         expect(response.body).toHaveProperty("removedBreadcrumbs", {});
       });
   });
+
+  // TODO this may be an existing issue or it may be my misunderstanding of how this should work
+  it.skip("returns changed nodes and invalidated breadcrumbs when a flow has updated only nested nodes (answers but not questions)", async () => {
```
builditben, line 161: Did I get this test right? Is this a pre-existing bug?

DafyddLlyr: Your test looks correct to me - this looks like a pre-existing bug.

If editors update the text an of an answer (e.g. "Yes" to "No"), then we would require users to re-visit this question and provide a new answer.

builditben:
Thanks for the feedback. There is a pre-existing issue here that it would be good to squash (along with a little refactor) but I'll merge this work in for now and we can follow-up with a fix for the skipped test later.

api.planx.uk/saveAndReturn/validateSession.test.ts
jessicamcinchak: This is a good catch! It seems reconciliation is only properly catching cases where the responses have changed order or changed in number (because that bubbles up as a diff in the `edges` of the parent question), but not the case where an existing response attribute changes alone (`alteredNodes` is properly populated in the validate-session response, but `removedBreadcrumbs` is not)

Addressing in separate PR makes sense to me, thanks for stubbing out this failing test :+1:

PR #1513, opened 2023-03-07
refactor: Move section chunking to store

DafyddLlyr:
This is a quick refactor to move the section chunking logic to the store for reuse.

The plan was to add tests and modify the export data structure so that we have access to these chunks in the document template data structure, but I'll handle that as a follow on PR so that this can get shared / reviewed sooner.

Please see PlanX Slack discussion here for context - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1678185974650719


builditben (approved):
Thanks for this :+1:

PR #1515, opened 2023-03-11
chore: update seed users

jessicamcinchak:
- Removes prior or inactive team members: Gunar, John, & Sarah
- Updates all seed emails to OSL emails (need to update this directly on prod/staging still for Ian to login successfully)


DafyddLlyr (approved):
I'm not 100% sure how foreign keys are set up here - I think there may be some tables that still reference user id 1 (John) as we see his name as the "ghost in the machine" on Pizzas. We can keep an eye out for this one - hopefully not a big issues, especially with upcoming user / permission updates.

jessicamcinchak:
@DafyddLlyr right - on production, I'd expect that we don't want to delete `users` records for john & gunar, but rather update the `users.email` value to be "REMOVED" or similar to preserve existing relationships to `operations` & flow content while preventing ability to sign-in. But seeds _should_ be able to be totally deleted here I think!

DafyddLlyr:
@jessicamcinchak Yep, that makes perfect sense thanks!

builditben (approved):

PR #1519, opened 2023-03-13
fix: update Article 4 references for St Peter's Basin in Newcastle

jessicamcinchak:
Jame's in Newcastle noted that a couple variables weren't getting picked up as expected during testing. This was a typo on my end - the references for St Peter's Basin are a bit of a different pattern than other records in Newcastle. 

Example address: 
- NE6 1TR &rarr; 6 MERCHANTS WHARF
- Should set `article4.newcastle.saintPetersBasin1`


DafyddLlyr (approved):

DafyddLlyr:
Just spotted this after sending the prod deployment on its way sorry. We can go for another prod deploy to pick this up today though also 👍

PR #1523, opened 2023-03-13
fix: Treat section card as first of next section

DafyddLlyr:
## Problem
Currently, the section banner does not display the correct data when reaching a section card. It should treat the section card as the first card of the _next_ section. 

## Solution
As we rely on breadcrumbs to calculate the current section and update this on `record()`, it's always backwards looking. By using `currentCard()` we can always display the correct section title and index.

## Testing
TODO: Link to example on Pizza

The section banner displays the correct information...

  - On the initial section card
  - On subsequent sections when navigating forwards
  - On previous sections when navigating via the "Back" button
  - On previous sections when navigating via the links on the section page


jessicamcinchak (approved):
Working as expected for me on the pizza!

PR #1524, opened 2023-03-13
fix: Link to first question from section page

DafyddLlyr:
## What does this PR do?
Instead of linking back to the section card, we should navigate back to the first question of that section.

https://user-images.githubusercontent.com/20502206/224757599-082e15cf-1861-4a31-b8ac-54b8b0ea078a.mov


jessicamcinchak (approved):
Also working as expected for me on the pizza! Feels like much smoother UI, and helps negate some of those "which status" questions we were coming up on when going "back" directly to the overview page :sweat_smile:

PR #1527, opened 2023-03-14
feat: Add `section_name` to BoPS Payload

DafyddLlyr:


editor.planx.uk/src/@planx/components/Send/bops/index.ts
```diff
@@ -215,6 +227,7 @@ export const makePayload = (
         ];
       }
       metadata = addPortalName(id, flow, metadata);
+      metadata = addSectionName(id, metadata);
```
DafyddLlyr, line 230: This pattern felt a bit cleaner if `addPortalName()` is not going to be around for much longer.

jessicamcinchak (approved):
Works exactly as expected, thanks for this one! 

Good shout too that this will be an easy key to "group by" in any data exports like overview.html :+1:

editor.planx.uk/src/@planx/components/Send/bops/index.ts
builditben: Is portal name used by BOPS currently?

builditben (approved):

PR #1528, opened 2023-03-14
fix: Update user pk, use port env var

DafyddLlyr:
A few little issues hit with getting Ian up and running on PlanX.

- Some Mac versions use port 7000 for Airdrop, but this needs to be updated in multiple places in the `.env` to work correctly. Alternatively, Airdrop can be disabled.
- Ian wasn't listed on prod and staging dbs, likely overwritten by the new data sync

### DIFF
--- a/hasura.planx.uk/config.yaml
+++ b/hasura.planx.uk/config.yaml
@@ -1,4 +1,4 @@
 version: 2
-endpoint: http://localhost:7000
+endpoint: http://localhost:${HASURA_PROXY_PORT}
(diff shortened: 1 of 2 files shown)


jessicamcinchak (approved):
thank you! sorry I switched seed emails and forgot to update id as well!

DafyddLlyr:
No apologies needed! There was another issue as well which I'm looking into - he had to manually run the `pnpm run dist` command for the API to spin up on Docker.

PR #1531, opened 2023-03-14
fix: Fetch current node without recursive loop

DafyddLlyr:
Quick fix using `useAnalyticsTracking()` to get current node in a more stable way - this is still calling `getCurrentCard()` behind the scenes. As it's using React's `useContext()` it's probably a bit wiser about rendering.

Ideally, we'd be able to track the current node in the store instead of having to call `currentCard()` which iterates over the flow. I'll take a look at this tomorrow 👍


builditben (approved):
This works :+1: 
:lady_beetle: :hammer:

PR #1533, opened 2023-03-14
fix: Re-enable Minio locally

DafyddLlyr:
## Problem
I'm currently getting the following error when running the app locally and trying to upload files - 

`{"error":"Inaccessible host: '127.0.0.1' at port '9000'. This service may not be available in the 'eu-west-2' region."}`

Is anybody else having this issue?

## Solution
- Always run Minio on local dev
- Use docker networking address

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -164,6 +164,7 @@ services:
       FILE_API_KEY: ${FILE_API_KEY}
       SLACK_WEBHOOK_URL: ${SLACK_WEBHOOK_URL}
       ORDNANCE_SURVEY_API_KEY: ${ORDNANCE_SURVEY_API_KEY}
+      MINIO_PORT: ${MINIO_PORT}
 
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -185,7 +186,6 @@ services:
   # used as an S3 service mock
   minio:
     image: minio/minio:RELEASE.2021-08-31T05-46-54Z
-    profiles: ["mock-services"]
```
builditben, line 188: Could this change mean that we could be running mino where we don't mean to be (i.e. outside of a test context)? Do the other change work without changing this approach to only running mino in a test context?

builditben (commented):
Will this change

DafyddLlyr:
@builditben Updated to now just use the `mock-services` profile when running locally which I think is a more accurate description of what's intended 👍

PR #1535, opened 2023-03-14
feat(wip): save & return happy path reconciliation with sections option two

jessicamcinchak:
opening for comparison against #1505 - this PR displays sections overview list instead of individual questions summary list on reconciliation page.

i think there's pros & cons to each! need to double check figma to see if this was considered & run by design team.

a few initial thoughts:
- sections overview list is most straightforward summary of changes when resuming,
- "NEW INFO NEEDED" status will apply to whole _section_ not individual questions which best matches the sections overview list, rather than summary list with status tags
- if you leave off on a section, then resume and see overview page, then click "continue", your first node is the overview page again and it feels redundant & wrong. this is less likely to happen with summary list because it only ever exists one other place in a flow (Review page)


editor.planx.uk/src/pages/FlowEditor/lib/store/navigation.ts
```diff
@@ -115,21 +118,39 @@ export const navigationStore: StateCreator<
...
+   * Calculate the status of each section node based on the current position in a flow (eg state.cachedBreadcrumbs) or provided breadcrumbs (eg on reconciliation)
    *   ** will need to become aware of all breadcrumbs within a section in the near future for "back" rather than only currentCard & upcomingCardIds
    */
-  sectionStatuses: (): Record<string, SectionStatus> => {
-    const { sectionNodes, currentCard, upcomingCardIds, cachedBreadcrumbs } =
-      get();
+  sectionStatuses: (
+    breadcrumbs?: Store.breadcrumbs,
+    updatedNodeIds?: string[]
+  ): Record<string, SectionStatus> => {
```
jessicamcinchak, line 127: this is a big change: `sectionStatuses()` now accepts optional arguments - if provided, it will calculate the section status based off the breadcrumbs object provided. if no arguments are provided, it will calculate the section status based off of cached breadcrumbs in state just as before.

this let's us calculate section statuses on the reconciliation page using the breadcrumbs object returned by /validate-session endpoint that is _not_ in state yet. 

open to other approaches for accomplishing this, but this felt most straightforward & flexible for now!

editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
```diff
@@ -78,11 +80,19 @@ const components: {
 
 interface SummaryListsBySectionsProps extends SummaryListProps {
   sectionComponent: React.ElementType<any> | undefined;
+  showStatusTag?: boolean;
```
jessicamcinchak, line 83: will cleanup this optional prop and the "Tag" export depending if we decide design in #1505 is useful or not!

editor.planx.uk/src/pages/Preview/ReconciliationPage.tsx
```diff
@@ -76,16 +92,24 @@ const ReconciliationPage: React.FC<Props> = ({
...
-          passport={data?.reconciledSessionData?.passport}
-          changeAnswer={changeAnswer}
-          showChangeButton={false}
-          sectionComponent="h3"
-        />
+        {hasSections ? (
+          <SectionsOverviewList
+            sectionNodes={sectionNodes}
+            sectionStatuses={reconciledSectionStatuses}
+            showChange={false}
```
jessicamcinchak, line 101: I'm choosing to _not_ link the section title names on the reconciliation page because I think we want to highlight/emphasize the "Continue" button as the singular action (consistent with hiding "change" links on SummaryList on reconciliation, but not review page) - 

On clicking "Continue" from reconciliation, we'll already have any changed nodes queued up to be answered/seen first.

jessicamcinchak:
[09. In S&R flows, how does reconciliation display if no content changes](https://trello.com/c/OI6u5662/2309-09-in-sr-flows-how-does-reconciliation-display-if-no-content-changes)

editor.planx.uk/src/pages/FlowEditor/lib/store/navigation.ts
```diff
@@ -115,19 +118,28 @@ export const navigationStore: StateCreator<
...
+    if (!breadcrumbs) breadcrumbs = cachedBreadcrumbs;
+
     const sectionStatuses: Record<string, SectionStatus> = {};
     Object.keys(sectionNodes).forEach((sectionId) => {
-      if (
+      if (updatedSectionNodeIds?.includes(sectionId)) {
+        // We only expect to receive updatedSectionNodeIds argument on reconciliation, therefore
+        //   this status should never apply to regular forwards/back/change navigation
+        sectionStatuses[sectionId] = SectionStatus.NeedsUpdated;
+      } else if (
```
jessicamcinchak, line 140: Ben's work will keep building on this by showing this status for all sections containing updated questions - but a basic way to _already_ trigger this status is to: save, directly edit your section node in your flow (eg give it a new title), re-publish, resume! 
Screenshot from 2023-03-16 08-46-08

editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
jessicamcinchak: closed #1505, removed this unused prop!

editor.planx.uk/src/pages/Preview/ReconciliationPage.tsx
```diff
@@ -35,16 +37,37 @@ const ReconciliationPage: React.FC<Props> = ({
...
+  // Assumes that reconciliation API has handled removing the associated section breadcrumb for each updated node
+  //   and returned them as part of `removedBreadcrumbs` so we can accurately set sectionStatuses
+  const updatedSectionNodeIds: string[] = [];
+  if (data?.removedBreadcrumbs) {
+    Object.keys(data.removedBreadcrumbs)?.forEach((nodeId: string) => {
+      if (flow[nodeId]?.type === TYPES.Section) {
+        updatedSectionNodeIds.push(nodeId);
+      }
+    });
+  }
```
builditben, line 63: nit/suggestion: as mentioned above, you could pass in an array of all affected node ids instead of filtering for sectionIds specifically (finding section ids can be `sectionStatuses` job). But filtering to only the nodeIds you care about makes sense too.

```suggestion
const affectedNodeIds =  data.removedBreadcrumbs.map((crumb) => crumb.id)
```

builditben (approved):
This looks good and makes sense. I've a few small optional suggestions but I think this is good to go.

editor.planx.uk/src/pages/FlowEditor/lib/store/navigation.ts
```diff
@@ -115,19 +118,28 @@ export const navigationStore: StateCreator<
...
   /**
-   * Calculate the status of each section node based on the current position in a flow
-   *   ** will need to become aware of all breadcrumbs within a section in the near future for "back" rather than only currentCard & upcomingCardIds
+   * Calculate the status of each section node based on the current position in a flow (eg state.cachedBreadcrumbs) or provided breadcrumbs (eg on reconciliation)
+   *   Pass updatedSectionNodeIds on reconiliation to ensure we set an accurrate status for every previously seen section because updated sections are removed from breadcrumbs during reconciliation
    */
-  sectionStatuses: (): Record<string, SectionStatus> => {
+  sectionStatuses: (
+    breadcrumbs?: Store.breadcrumbs,
+    updatedSectionNodeIds?: string[]
```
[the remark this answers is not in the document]
jessicamcinchak: i had been going back on forth on this one but i think you're right that `updatedNodeIds` is much clearer here - and yes, because it's checking against SectionNodes anyways it's self-filtering in that sense. 

and it simplifies the Reconciliation Page logic a bit to just directly pass `Object.keys(data.removedBreadcrumbs)`

editor.planx.uk/src/pages/Preview/ReconciliationPage.tsx
jessicamcinchak: addressed this is https://github.com/theopensystemslab/planx-new/pull/1535/commits/76c388f2bab1fa8efca4281bc4fa87037e49e937 !

editor.planx.uk/src/@planx/components/Section/Public.tsx
```diff
@@ -68,52 +71,85 @@ export default function Component(props: Props) {
...
+
 const getTagBackgroundColor = (theme: Theme, title: string): string => {
   const backgroundColors: Record<string, string> = {
-    [SectionStatus.NotStarted]: "#F3F2F1",
-    [SectionStatus.InProgress]: theme.palette.primary.main,
-    [SectionStatus.Completed]: "#00703C",
+    [SectionStatus.NotStarted]: theme.palette.grey[200],
+    [SectionStatus.InProgress]: lighten(theme.palette.primary.main, 0.9),
+    [SectionStatus.Completed]: theme.palette.success.main,
+    [SectionStatus.NeedsUpdated]: theme.palette.action.focus, // GOV UK YELLOW for now, check Figma
```
DafyddLlyr, line 139: This is much nicer / more sustainable. 

We could also add these as discrete status colours in the theme which would mean they get calculated once once the theme is generated and MUI will provide us with `contrastText` values automatically. MUI has defaults like warning, info, success etc that follow this pattern. I think this is probably only worth doing if we think these values will have meaning outside of this component though.

Docs: https://mui.com/material-ui/customization/palette/#adding-new-colors

editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
```diff
@@ -81,8 +82,15 @@ interface SummaryListsBySectionsProps extends SummaryListProps {
 }
 
 function SummaryListsBySections(props: SummaryListsBySectionsProps) {
-  const [hasSections, getSortedBreadcrumbsBySection] = useStore((state) => [
+  const [
+    hasSections,
+    sectionNodes,
```
DafyddLlyr, line 87: `sectionNodes` and `sectionStatuses()` look unused?

editor.planx.uk/src/pages/Preview/ReconciliationPage.tsx
```diff
@@ -76,16 +89,24 @@ const ReconciliationPage: React.FC<Props> = ({
             </Box>
           )}
         <Typography variant="h3" component="h2">
-          Review your answers so far
+          Review your {hasSections ? "progress" : "answers"} so far
```
DafyddLlyr, line 92: This is clever! 👏

editor.planx.uk/src/theme.ts
```diff
@@ -37,6 +37,9 @@ const DEFAULT_PALETTE: Partial<PaletteOptions> = {
   error: {
     main: "#E91B0C",
   },
+  success: {
+    main: "#00703C",
+  },
```
DafyddLlyr, line 42: nit: Is this used currently?

DafyddLlyr (approved):

editor.planx.uk/src/theme.ts
jessicamcinchak: `theme.palette.success.main` is not referenced anywhere explicitly in our code, and i did click around a bit after setting this to ensure it wasn't being referenced under the hood by MUI for default component states and didn't spot any - so I think this is safe to introduce! it felt a bit cleaner and easier to reference in future than hardcoding the "completed" green shade exclusively in the Sections component.

editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
jessicamcinchak: correct, good spot - now cleaned up!

PR #1536, opened 2023-03-15
chore: rotate `db-password` across all AWS environments

jessicamcinchak:
same methodology as #1526, but Pulumi CLI commands ran in `infrastructure/data` rather than `infrastructure/application`

### DIFF
--- a/infrastructure/data/Pulumi.staging.yaml
+++ b/infrastructure/data/Pulumi.staging.yaml
@@ -1,4 +1,4 @@
 config:
   aws:region: eu-west-2
   data:db-password:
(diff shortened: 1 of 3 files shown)


infrastructure/data/Pulumi.staging.yaml
```diff
@@ -1,4 +1,4 @@
 config:
   aws:region: eu-west-2
```
jessicamcinchak, line 2: while doing this, noticed that region is only explicitly set on staging - should we set this consistently?

DafyddLlyr (approved):

PR #1546, opened 2023-03-16
doc: How to setup custom domains

DafyddLlyr:


doc/how-to/how-to-setup-custom-subdomains.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Context 🖼️
+Teams can setup custom subdomains for PlanX so that the service can be accessed via their domain (e.g. https://planningservices.medway.gov.uk/).
+
+This guide will walk through the process of setting a custom domain for a new team. It will require actions by both PlanX developers and council IT teams.
+
+## Process ⚙️
+1. **PlanX** - Provide IT team with documentation listing what's required. This is part of the onboarding process and briefly outlines the process. Documentation - https://www.notion.so/opensystemslab/IT-systems-services-d4f8b88fb9694f33a24411801150c793
+
+2. **PlanX** - Provide IT team with Certificate Signing Request (CSR) if requested. The IT team will need to tell us what to put in [in each of the fields](https://en.wikipedia.org/wiki/Certificate_signing_request#Procedure).
```
jessicamcinchak, line 11: tiny typo: `...in [in...`

doc/how-to/how-to-setup-custom-subdomains.md
```diff
@@ -0,0 +1,87 @@
...
+
+    **How to generate a CSR**
+    ```shell
+    openssl req -new -newkey rsa:2048 -nodes -keyout <TEAM_NAME>.key -out <TEAM_NAME>.csr
```
jessicamcinchak, line 15: could we also add a quick note under this section of why `rsa:2048` is required based on the error earlier this week?

doc/how-to/how-to-setup-custom-subdomains.md
```diff
@@ -0,0 +1,87 @@
...
+        cat council.key | pulumi config set ssl-{team}-key --stack production --secret
+
+        cat council.cert | pulumi config set ssl-{team}-cert --stack production --secret
+
+        cat chain.cert | pulumi config set ssl-{team}-chain --stack production --secret
+        ```
+
+        To validate the above steps, check `infrastructure/application/Pulumi.production.yaml` which will list the above secrets for the team in an encrypted format.
+
+        At this stage we can execute `pulumi preview` from the application layer to check for any issues with the config.
```
jessicamcinchak, line 74: is it true that CI will execute this when you open a PR? or were you running this command manually? (or both: eg doing manually is quicker feedback loop for any issues than CI, but CI also always runs this to confirm?)

difference of which CI jobs are running `pulumi preview` versus `pulumi up` is usually one I need to double check so being explicit in docs is helpful for me!

jessicamcinchak (approved):
thanks for pulling this more together and thoroughly explaining :star2:

doc/how-to/how-to-setup-custom-subdomains.md
DafyddLlyr: I've clarified the requirements for this in our docs (I had also previously updated the public docs when going back and forth with James).

doc/how-to/how-to-setup-custom-subdomains.md
DafyddLlyr: Thanks for raising this one. It wasn't actually part of my process, but something I actually found in Gunar's docs when pulling this together and thought it might be helpful. I kicked myself when I saw it as I thought it could have saved some time / frustration.

I checked this out and it turns out it isn't really very useful here so I've dropped it from the docs for now.

`pulumi preview` can be run locally and does run on all PRs but it just gives a diff of what will change. Here's [a live preview](https://app.pulumi.com/planx/application/staging/previews/dda561c8-f96f-4f7b-829f-1048c2d2bb5e) from a PR which later failed `pulumi up` when deploying to prod. On the PR into `main` it's running `pulumi preview --stack staging` which isn't helpful when making a change to production.

Running `pulumi preview --stack production` is failing for me locally on both the application and data layer.

Here's what I propose - 
 - Work out and fix `pulumi preview --stack production` locally
 - Test locally with invalid certs and see if this is helpful. If so, add back to docs.
 - Run this as part of CI when opening a PR into `production` - there are parts of our IaC which treat production and staging stacks differently so this is something we would want to be doing

jessicamcinchak: totally makes sense, thanks for clearly outlining here - I confirmed same local behavior on `pulumi preview --stack production` :heavy_plus_sign:

doc/how-to/how-to-setup-custom-subdomains.md
```diff
@@ -0,0 +1,93 @@
...
+        cat chain.cert | pulumi config set ssl-{team}-chain --stack production --secret
+        ```
+
+        To validate that you have successfully added the secrets to Pulumi, check `infrastructure/application/Pulumi.production.yaml` which will list the above secrets for the team in an encrypted format.
+
+7. **PlanX** - Deploy to production in order to create ACM records and CloudFront CDN
+
+8. **PlanX** - Provide IT team with domain of CloudFront CDN. This will be in format `<abc123>.cloudfront.net` and is listed on the CloudFront dashboard (via AWS Console).
+  
+9. **IT Team** - Create CNAME DNS record pointing at `<abc123>.cloudfront.net`
```
builditben, line 86: I guess the reason this isn't done asynchronously and up-front is that we haven't created the CloudFront distribution until step 7.
Just an idea - would it be possible to create the distribution up front (alongside step 3?) and then edit it to include certs afterwards? If so, this may enable less back and forth and allow us to test the end-to-end (perhaps with a holding page instead of the application).

builditben (approved):
Thanks for the detailed write-up. :+1:

PR #1547, opened 2023-03-16
chore: Remove section navbar from section components

DafyddLlyr:
From @ianjon3s - 

> the navigation pages should not have the navigation banner present as they are intermediate states that mark milestone positions in the journey (they also give a more detailed breakdown of navigation status in the table). Scroll position isn't a concern here as we are jumping out of the user-flow of repetitive question patterns to signify the milestone has been reached.


builditben:
I'm glad to see this change. I think this feels a lot better. 
Working as expected :+1:

builditben (approved):
It works :+1:

PR #1553, opened 2023-03-17
refactor: Move OneApp generation to planx-core

DafyddLlyr:
## What does this PR do?

- Remove OneApp XML generation from frontend, in favour of generating it via planx-core on the backend

## Why?
 - This felt like a sensible refactor to tackle so that template file generation could be handled in a single way for sendToEmail and sendToUniform


editor.planx.uk/src/@planx/components/Send/Public.tsx
```diff
@@ -19,19 +18,6 @@ import {
 
 export type Props = PublicProps<Send>;
 
-const useStyles = makeStyles((theme) => ({
```
DafyddLlyr, line 22: This was unused, this is just a tidy up.

builditben (approved):
Really great to see this getting cleaned-up. Thanks :+1:

DafyddLlyr:
Converted to draft - `templateNames` is not correctly refactored into the API here currently.

jessicamcinchak (approved):

PR #1554, opened 2023-03-20
fix: Stable Pizza databases

DafyddLlyr:
**Current behavior**
 - Postgres database re-seeded on each commit to an open PR
 
**Desired behavior**
 - Postgres database only seeded when PR is opened


builditben (approved):
Thanks for this :+1:

PR #1561, opened 2023-03-22
fix: update valid GovPay referrer links on resume checks

jessicamcinchak:
**What's happening & what should happen:**
While doing a test application earlier, I noticed that I was prompted to enter my email when I was redirected back to Planx from GovPay. This is unintended behavior and in the past prevented folks from returning to submit their applications. This can be re-created on staging & production currently.

When a user specifically returns to Planx with both the sessionId & email query params in the URL and the referrer matches Gov Pay, we should bypass email entry & reconciliation.

**Why it's happening:**
As best I can tell GovPay is now using a subdomain that doesn't match the referrer we were checking for! I haven't been able to track down a technical changelog to confirm this, but I found this mismatch by calling `document.referrer` in my console when I was redirected back to Planx after making a payment on staging. I haven't removed the previous referrer because I'm not 100% sure if/how this may differ between GovUK Sandbox & Production envs; I've only redirected back after a test payments, but _connecting to_ the first page of Production GovPay does display the same `card.payment.gov.uk` subdomain now so I think it's safe to assume this will also be its' referrer.

**Testing this fix:**
- On a Save&Return flow, you can navigate back from Pay and not be prompted to confirm your email: https://1561.planx.pizza/lambeth/pay-test/preview
- On a non-S&R flow, you can navigate back from Pay as normal (no regressions): https://1561.planx.pizza/lambeth/pay-test-copy/preview

I believe this change successfully fixes HTTPS environments (pizzas, staging & prod), but my e2e test fails because the referrer is stripped when going HTTPS to HTTP (ie localhost) and I'm a little stumped on how to automatically test - see comment below!

### DIFF
--- a/api.planx.uk/saveAndReturn/validateSession.ts
+++ b/api.planx.uk/saveAndReturn/validateSession.ts
@@ -170,7 +170,7 @@ const findSession = async (
 ): Promise<LowCalSession | undefined> => {
   const query = gql`
     query FindSession($sessionId: uuid!) {
-      lowcal_sessions_by_pk(id: $sessionId) {
+      lowcal_sessions(where: { id: { _eq: $sessionId } }, limit: 1) {
         data
         updated_at
       }
@@ -178,7 +178,7 @@ const findSession = async (
   `;
   const headers = getSaveAndReturnPublicHeaders(sessionId, email);
   const response = await publicClient.request(query, { sessionId }, headers);
-  return response?.lowcal_sessions_by_pk;
+  return response.lowcal_sessions?.[0];
 };
 
 const updateLowcalSessionData = async (


api.planx.uk/saveAndReturn/utils.ts
```diff
@@ -157,7 +157,7 @@ const validateSingleSessionRequest = async (
   try {
     const query = gql`
       query ValidateSingleSessionRequest($sessionId: uuid!) {
-        lowcal_sessions(where: { id: { _eq: $sessionId } }, limit: 1) {
+        lowcal_sessions_by_pk(id: $sessionId) {
```
jessicamcinchak, line 160: this is just minor cleanup while trying to narrow down where this bug was coming from - `_by_pk` implies `limit: 1` since it's the unique, primary key on our table

e2e/src/pay-with-resume.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    await page.getByText("Pay using GOV.UK Pay").click();
+    await fillGovUkCardDetails({
+      page,
+      cardNumber: cards.successful_card_number,
+    });
+    await page.locator("#confirm").click();
+    const { paymentId: actualPaymentId } = await waitForPaymentResponse(page);
+    await expect(page.getByText("Application sent")).toBeVisible();
+    await expect(page.getByText(actualPaymentId)).toBeVisible();
```
jessicamcinchak, line 67: I thought an e2e test was going to be the right way to test this intended redirect/resume behavior and proactively catch in the future, but then I naively realized after spinning up the test that the referrer is stripped from the browser when navigating from HTTPS to HTTP (ie localhost), therefore this e2e test fails for the same reason this behavior can't be tested in local development.

Any other ideas for how to test??

jessicamcinchak:
[Applicants should not be prompted to confirm their email address when coming back from Gov Pay](https://trello.com/c/6fU0AYht/2341-applicants-should-not-be-prompted-to-confirm-their-email-address-when-coming-back-from-gov-pay)

editor.planx.uk/src/pages/Preview/ResumePage.tsx
```diff
@@ -151,8 +151,11 @@ const InvalidSession: React.FC<{
  */
 const getInitialEmailValue = () => {
   const emailQueryParam = useCurrentRoute().url.query.email;
-  const isRedirectFromGovPay =
-    document.referrer === "https://www.payments.service.gov.uk/";
+  const isRedirectFromGovPay = [
+    "https://www.payments.service.gov.uk/",
+    "https://card.payments.service.gov.uk/",
+  ].includes(document.referrer);
```
DafyddLlyr, line 157: Perfect! Works for me 👌

e2e/src/pay-with-resume.spec.ts
DafyddLlyr: Yeah this is a tricky one!

The only thing I can really see in the docs is setting this via the `Page.goto()` API (docs: https://playwright.dev/docs/api/class-frame#frame-goto-option-referer), but this doesn't really help us as we would need GovPay to set this.

I'd be ok to merge without the test and concentrate ideas on ways of avoiding `document.referrer` if possible.

DafyddLlyr (approved):

e2e/src/pay-with-resume.spec.ts
jessicamcinchak: Thanks for thinking on this one too! Agree that sounds like ideal way forward and lets us fix bug soonest.

I've just updated this branch to remove the e2e commit and I'll just merge the single line change to add the new gov pay subdomain once CI finishes :+1:

PR #1562, opened 2023-03-22
refactor: update validateSession interface and implementation

builditben:
This is a refactor of the reconciliation API with a bug-fix for removing breadcrumbs for updated/added responses when the question/checklist hasn't also been updated.

The API has been simplified to only expose the required fields and the editor has been updated to include type definitions for the `ReconciliationResponse` interface.

If an answer within a flow has updated, the question should be represented on reconciliation.

This PR removes some sorting logic from the API (but not the in editor - I still have not got to the bottom of why objects are sorted alphabetically by keys) but this change should be safe since sort order is not affected by the `delete` operations performed by the API.

This change also introduces a stored procedure for diff, which may improve the performance some (though not much in this particular API endpoint).

Previous iterations of this change had performance issues from a slow function planx-core which has now been patched.

Testing:
 - Does reconciliation work as expected?
 - What happens if the content change is a new, moved, or deleted section?
Here's a save-and-return flow to test with: https://1562.planx.pizza/testing/sections-test/preview?analytics=false


api.planx.uk/saveAndReturn/utils.ts
builditben, line 1: These utils are used (as far as I can tell) to deterministically sort the breadcrumbs. I can't see a good reason why this is needed in this context (it is used in the editor to enable memoization).

editor.planx.uk/src/lib/lowcalStorage.ts
```diff
@@ -7,17 +7,6 @@ import { client } from "./graphql";
...
-  // readonly length: number;
-
-  // [name: string]: any;
-
-  // /** Returns the name of the nth key, or null if n is greater than or equal to the number of key/value pairs. */
-  // async key(index: number) {
-  // }
-
-  // async clear() {}
-
```
builditben, line 20: These comments are not helpful so cleaning this up.

api.planx.uk/saveAndReturn/validateSession.ts
```diff
@@ -32,130 +42,78 @@ export async function validateSession(
...
-              });
-            });
-          }
+    const sessionData = fetchedSessionData!;
 
-          // TODO: FUTURE RECONCILIATION CHECKS
-          //   **what about collected flags? what about `auto: true`? component dependencies like FindProp/Draw/PlanningConstraints?
-
-          // update the lowcal_session.data to match our updated in-memory sessionData.data
-          //   XX: apply sorting to match the original get/set methods used in editor.planx.uk/src/lib/lowcalStorage.ts
```
builditben, line 123: These comments have been shortened and added to the top of the function.

builditben:
Performance issues causing a timeout for large flows (the issue you noticed @jessicamcinchak) have been addressed in https://github.com/theopensystemslab/planx-core/pull/10 which this PR depends on.

api.planx.uk/saveAndReturn/validateSession.ts
```diff
@@ -1,22 +1,32 @@
...
+export interface ValidationResponse {
+  message: string;
+  reconciledSessionData: LowCalSession["data"];
+  alteredNodes?: Node[];
+  removedBreadcrumbIds?: string[];
+}
+
+// TODO - ensure reconciliation handles:
+//  * collected flags
+//  * auto-answered component dependencies like FindProp/Draw/PlanningConstraints
```
jessicamcinchak, line 29: thanks for cleaning up these stray code comments

tiny nit: this last line refers to separate things: auto answers AND component dependencies, so might be best split onto a second line for clarity

builditben:
The API is behaving as it should but the editor doesn't always present the right questions. I'm putting this back into draft until I understand why this is.

jessicamcinchak:
@builditben related to breadcrumb order - these PRs & linked Trello tickets might be helpful context ! 
- https://github.com/theopensystemslab/planx-new/pull/1034
- https://github.com/theopensystemslab/planx-new/pull/919

if you have a specific flow & path that you're creating a bug on, i'd be happy to try to recreate and see if i can help narrow down issue/get a fix in place (if not sorting issue, maybe auto-answering etc)

editor.planx.uk/src/lib/lowcalStorage.ts
```diff
@@ -37,8 +26,7 @@ class LowcalStorage {
     try {
       const session: Session = data.lowcal_sessions_by_pk?.data;
       if (!session) return null;
-      current = stringifyWithRootKeysSortedAlphabetically(session) || null;
```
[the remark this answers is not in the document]
builditben: I've still not fully understood how/why alphabetical sort order of object keys is necessary but it seems that some ordering logic functions rely on this behavior. I've reverted this change for now.

builditben:
[99. In S&R flows, what happens if the content change is a new, moved, or deleted section?](https://trello.com/c/CaK460fr/2311-99-in-sr-flows-what-happens-if-the-content-change-is-a-new-moved-or-deleted-section)

jessicamcinchak:
> This is a refactor of the reconciliation API with a bug-fix for removing breadcrumbs for updated/added responses when the question/checklist hasn't also been updated.
>
> If an answer within a flow has updated, the question should be represented on reconciliation.

Working through reviewing newest changes here which have definitely mostly fixed the bug you describe but I'm still hitting some unexpected behavior that I'd like to clarify! 

Here's what I'm doing:
- Start a session, answer first two questions, save (I'm using test flow: https://1562.planx.pizza/testing/reconciliation-answers-test) 
- Edit first question in flow by changing the text of an answer _only_, re-publish
- Resume my session, I correctly see message "something has changed" on reconciliation overview page and my first question/answer is not listed, click "continue" (:tada: Your bug-fix works; this definitely breaks at this step on prod: https://editor.planx.uk/testing/reconciliation-answers-test)
- I expect to be presented with the first question AGAIN, which I am in the small test flow, but am **not** when testing in a full-sized LDCs
  - for example, start this flow https://1562.planx.pizza/buckinghamshire/apply-for-a-lawful-development-certificate/preview, proceed through to beginning of about-the-applicant questions, save
  - Edit answer option to question "What are you applying about?", republish
  - Resume (no longer timing out, that is definitely fixed!), reconciliation overview page is correct, click "continue", expect to see 'What are you applying about?" but instead land back to prior position in about-the-applicant questions (and "What are you applying about" nodeId is not queued up in "upcomingCardIds")

Can you recreate? Again, I'm just getting this when testing on a full LDC, not small isolated flows - so some guesses at where bugs might still be creeping in are around node ordering when a flow has clones and portals?

If you can't recreate, happy to write this off as an edge case until we can consistently recreate.

builditben:
Thanks for the detailed testing here @jessicamcinchak 

I did see the same behavior you described too, but I thought this was how auto-answering is implemented rather than reconciliation itself so I didn't flag this as an issue (at least not for this work). Am I right in thinking what you described only impacts auto-answered questions?

My view would be that auto-answering should be out of scope for this change since I don't believe this is a regression (is that right?  there was already a TODO comment about auto answering) and the requirements for that fix might increase the scope of this PR to be too broad. Is that fair?

If this isn't an auto-answering issue, I'll definitely investigate further.

DafyddLlyr:
I can recreate the above - the simple flow works, the LDC does not.

If I continue on from the reconciliation page and then go "Back" I can eventually hit the changed question and see that my answer is still selected by value (despite the text changing). It's not in `upcomingCardIds()` and isn't encountered on any forward journey.

jessicamcinchak:
@builditben double checking `removePassportValue` function in `validateSession` would be my best guess of where something might be going wrong here?

The question I'm changing in the flow between my save & resume is one that I originally _manually_ answered (eg its breadcrumb will read `auto: false` and it _does_ set an associated passport variable), therefore I expect it to be presented again to manually re-answer. In order for that to happen, it's breadcrumb needs to be removed from session data _as well as_ the assoicated passport key/value.

The "what about auto-answers" outstanding/bigger todo's for me are more about any _future_ questions that were immediately auto-answered based on this first manual anwer but not seen yet by the user (and should never be seen, but could still theoretically determine your path). This falls a lot more in the filter/auto-answers logic work and connects to some questions we've raised like "what if we only auto-answer _when a node is reached_ rather than immediately when we pick up the variable?" Those questions are definitely bigger than this PR, but I'd expect the existing "if breadcrumb sets a passport var, remove it too" logic to still work here as that should ensure changed questions are shown first when resuming!

jessicamcinchak (approved):
Based on the testing last week, happy for this one to merge as-is (great work fixing the known answers bug and your refactor makes the main reconciliation method a lot more clear) & we can plan to keep investigating the local state behavior bug separately !

PR #1563, opened 2023-03-22
feat: add accessibility tests for Section & SectionsOverviewList components

jessicamcinchak:
This is an outstanding sections EPIC trello ticket bullet point without a broken out card, apologies if I've forgotten if anyone previously said they wanted to pick this one up! 

We'll likely have to update these once we disable/remove (?) the feature flag when the feature goes live (open to other suggestions for how to work around this now if better ideas!). 

I think these cover all the major scenarios for now, and a11y axe tests didn't flag any new issues :sparkle:


DafyddLlyr (approved):
Perfect ✅ Thanks for picking this up

PR #1564, opened 2023-03-23
chore: Update Hasura proxy port

DafyddLlyr:
Follow up from https://github.com/theopensystemslab/planx-new/pull/1528

`config.yaml` doesn't seem to support the use of variables for `endpoint` (likely why it was hardcoded previously?).

I've also changed to port 7100 which works for me on Mac - would appreciate somebody on another OS checking the availability of this port (`lsof -i:7100` returns an empty response if open).

Once this has been reviewed / approved I'll push up the new `.env` var file, I suspect that the Pizza will fail until I do that?


jessicamcinchak (approved):
This looks good to me! 

To be totally clear: these port changes are for Docker envs (local plus pizza) only, and the `HASURA_PROXY_PORT` in our AWS envs was and is remaining `80`, correct?

DafyddLlyr:
PR merged and updated `.env` with `HASURA_PROXY_PORT=7100` pushed to S3.

PR #1565, opened 2023-03-23
feat: Add an error page

builditben:
Adds a simple error page and updates the cloudflare config to use this page.

---
image

### DIFF
--- a/editor.planx.uk/public/error.html
+++ b/editor.planx.uk/public/error.html
@@ -22,7 +22,7 @@
         body{ padding-top: 160px; }
       }
       h1 {
-        font-size: 50px;
+        font-size: 40px;
       }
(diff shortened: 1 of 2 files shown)


builditben:
[Create a PlanX maintenance page for planned downtime](https://trello.com/c/ZurqVLyH/2335-create-a-planx-maintenance-page-for-planned-downtime)

jessicamcinchak (approved):
Nice & simple! 

Do you think it's worth adding one more brief sentence to re-assure no application data is lost and you can resume any in-progress apps when the service is back up? I noticed that was suggested copy in Gov UK patterns like here: https://design-system.service.gov.uk/patterns/service-unavailable-pages/

PR #1568, opened 2023-03-24
fix: Handle file uploads in standard format

DafyddLlyr:
I hit this issue when looking at attach files for sendToEmail. Because the passport variable assigned in DrawBoundary is not the same as normal file upload we need some additional code to get the file based on the passport variable as opposed to relying on the structure of the object.

Before we jump into multiple file upload I'd be interested in wider refactor of some of this shared upload logic, but this feels like a decent light touch for now.

**What this fixes**
 - Manually having to find `DrawBoundary` files when parsing the passport
 - This file isn't being currently handled in `planx-core` when generating a OneApp payload, or when we delete files by parsing the passport

### DIFF
--- a/api.planx.uk/send/email.ts
+++ b/api.planx.uk/send/email.ts
@@ -157,12 +157,7 @@ const downloadApplicationFiles = async(req: Request, res: Response, next: NextFu
               files.push(url);
             });
           });
-        
-        // Additionally check if they uploaded a location plan instead of drawing
-        if (sessionData.passport?.data?.["proposal.drawing.locationPlan"]) {
-          files.push(sessionData.passport.data["proposal.drawing.locationPlan"]);
-        }
-  
+
         // Download files from S3 and add them to the zip folder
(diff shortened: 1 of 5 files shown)


jessicamcinchak (approved):
Code looks good to me! 

One thing I noticed that I'm not entirely sure is introduced here or not: because we have two entries for this file in the passport (under `proposal.drawing.locationPlan` & `property.uploadedFile`) - we'll have two entries for it in the BOPS `files` payload with different tags, which might cause confusion! Can we safely de-dupe?

DafyddLlyr:
> One thing I noticed that I'm not entirely sure is introduced here or not: because we have two entries for this file in the passport (under proposal.drawing.locationPlan & property.uploadedFile) - we'll have two entries for it in the BOPS files payload with different tags, which might cause confusion! Can we safely de-dupe?

Yeah this is true - I'll open a new PR that handles this. We can actually remove a few file things now that other PRs have been merged!

PR #1574, opened 2023-03-26
refactor: Simplify Editor file upload logic

DafyddLlyr:
# What does this PR do?
 - Generates list of files on the backend, no longer needs a list sent via `getUniformParams()`
 - Simplifies duplicated keys set by `DrawBoundary`
   - See comment here - https://github.com/theopensystemslab/planx-new/pull/1568#pullrequestreview-1356945031
   - `cachedFile` removed
   - `uploadedFile` removed
   - `proposal.drawing.locationPlan` kept in format which more closely matches other file upload data objects.

## Testing
There's certainly more scope for unit testing or E2E testing here, but I'd be a bit more keen to pick that up once we move more Submission logic to the backend. If you feel it's more appropriate to increase coverage at this stage to describe the current Editor behaviour I'm happy to pick that up.

**To test**
 - Upload file via `DrawBoundary`
   - If I navigate "Back" to this component file still persists (`cachedFile` behaviour)
 - If I submit a `DrawBoundary` upload to BOPS...
   - It is correctly added to the payload
   - It is correctly tagged in the payload
   - It is only listed once in the payload (and passport)
- I can submit a Uniform submission and files are correctly added to Zip file and `payload.xml`

### DIFF
--- a/editor.planx.uk/src/@planx/components/DrawBoundary/Public/Public.test.tsx
+++ b/editor.planx.uk/src/@planx/components/DrawBoundary/Public/Public.test.tsx
@@ -19,18 +19,6 @@ test("recovers previously submitted files when clicking the back button", async 
         url: "http://localhost:7002/file/private/slb56xfv/placeholder.png",
       },
     ],
-    cachedFile: [
-      {
-        file: {
-          path: "placeholder.png",
-          size: 6146,
-        },
-        status: "success",
-        progress: 1,
-        id: "43sDL_JNJ6JgYxd_WUYW-",
-        url: "http://localhost:7002/file/private/slb56xfv/placeholder.png",
-      },
-    ],
   };
 
   const { user } = setup(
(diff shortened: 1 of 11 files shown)


editor.planx.uk/src/@planx/components/DrawBoundary/Public/Public.test.tsx
```diff
@@ -7,21 +7,7 @@ import DrawBoundary from "./";
...
-          size: 6146,
-        },
-        status: "success",
-        progress: 1,
-        id: "43sDL_JNJ6JgYxd_WUYW-",
-        url: "http://localhost:7002/file/private/slb56xfv/placeholder.png",
-      },
-    ],
-    cachedFile: [
+    "proposal.drawing.locationPlan": [
```
jessicamcinchak, line 10: :+1: this feels much clearer!

editor.planx.uk/src/@planx/components/Send/uniform/index.ts
```diff
@@ -22,38 +17,9 @@ export function getUniformParams({
...
-
-  // applicants may upload the same file in multiple slots,
-  //  but we only want to send a single copy of each file to Uniform
-  const uniqueFiles: string[] = [];
-  files.forEach((file) => {
-    if (!uniqueFiles.includes(file.url)) {
-      uniqueFiles.push(file.url);
-    }
-  });
-
```
jessicamcinchak, line 52: satisfying to see whole chunks like this move to the API :heavy_plus_sign:

jessicamcinchak (approved):
Thanks for tidying this up, all working as expected for me per your testing notes. I support your testing plan & reasoning !

PR #1576, opened 2023-03-27
feat(wip): Basic routing for Invite to Pay

DafyddLlyr:


editor.planx.uk/src/routes/views/preview.tsx
```diff
@@ -0,0 +1,97 @@
...
+      team={flow.team}
+      footerContent={globalSettings?.footerContent}
+      settings={flow.settings}
+      context="Preview"
+      globalSettings={globalSettings}
+      flow={flow}
+    >
+      <SaveAndReturnLayout>
+        <View />
+      </SaveAndReturnLayout>
```
DafyddLlyr, line 94: Nice bonus from some of this re-organisation work - only `/published` is wrapped in the Save&Return layout, not `/unpublished` which addresses issues raised in this ticket - https://trello.com/c/0Uw8ViZJ/2324-make-editor-sidebar-preview-behave-same-as-the-unpublished-preview-increase-granularity-of-previewenvironment

editor.planx.uk/src/components/Header.tsx
```diff
@@ -259,17 +272,21 @@ const PublicToolbar: React.FC<{
...
+          <IconButton
+            color="secondary"
+            onClick={handleRestart}
+            aria-label="Restart Application"
+            size="large"
+          >
+            <Reset color="secondary" />
+          </IconButton>
+        ) : (
+          <Box />
```
DafyddLlyr, line 286: This is a bad hack to handle keeping the title centred - planning to revisit this!

editor.planx.uk/src/pages/Preview/Context.tsx
```diff
@@ -1,10 +1,10 @@
 import React from "react";
 
-import type { Flow, GlobalSettings } from "../../types";
+import type { FlowSettings, GlobalSettings } from "../../types";
 
 export const PreviewContext = React.createContext<
```
DafyddLlyr, line 5: I think we should remove this.

We're currently using three methods of state management -

- React Context (here)
 - Zustand stores
- Route data (using withData() in routes)

I'd be in favour of simplifying this down to just Zustand. There's a few places here that this would simplify things but I'd not touched them in order to keep this PR as focussed as possible.

editor.planx.uk/src/components/Header.tsx
```diff
@@ -383,15 +415,38 @@ const EditorToolbar: React.FC<{
...
+  headerRef: RefObject<HTMLDivElement>;
+  team?: Team;
+}
+
+const Toolbar: React.FC<ToolbarProps> = ({ headerRef, team }) => {
+  const route = useCurrentRoute();
+  const path = route.url.pathname.split("/").slice(-1)[0];
+  const flowSlug = useStore((state) => state.flowSlug);
+
+  switch (path) {
```
DafyddLlyr, line 428: This isn't a great solution - clearly it's not very scalable.

I think ideally when setting a view doing some route handling, we could set a context of some sort (e.g. "Published", "Standalone", "Editor") which the UI could then reference in various ways.

I did try using `previewEnvironment` for this exact purpose but it's being used pretty inconsistently and a wider refactor is outside the scope of this PR. For example, `previewEnvironment = "editor"` when in `/unpublished` routes, and is set to `"standalone"` when on the list of teams.

editor.planx.uk/src/components/Header.tsx
```diff
@@ -383,15 +415,41 @@ const EditorToolbar: React.FC<{
...
+    case flowSlug: // Custom domains
+    case "preview":
+    case "unpublished":
+      return <PublicToolbar team={team} route={route} />;
+    default:
+      return (
+        <PublicToolbar team={team} route={route} showResetButton={false} />
+      );
+  }
```
DafyddLlyr, line 445: This isn't a great solution - clearly it's not very scalable.

I think ideally when setting a view doing some route handling, we could set a context of some sort (e.g. "Published", "Standalone", "Editor") which the UI could then reference in various ways.

I did try using `previewEnvironment` for this exact purpose but it's being used pretty inconsistently and a wider refactor is outside the scope of this PR. For example, `previewEnvironment = "editor"` when in `/unpublished` routes, and is set to `"standalone"` when on the list of teams.

editor.planx.uk/src/routes/pay.tsx
```diff
@@ -0,0 +1,101 @@
...
+      (req) => `../../../preview/pages/${req.params.page}`
+    ),
+    "/invite/pages/:page": redirect(
+      (req) => `../../../../preview/pages/${req.params.page}`
+    ),
+  })
+);
+
+const getPaymentRequest = async (req: NaviRequest) => {
+  const paymentRequestId = req.params["payment-request-id"];
```
jessicamcinchak, line 65: nit / question: I was manually navigating to these URLs based on a database record and realised I was defaultting to `paymentRequestId=` rather than hyphens because that's how we format S&R magic link params (eg `sessionId=`) - do we care about consistency like that here? I'm not sure it really matters, just a very minor observation!

jessicamcinchak (approved):
This all looks really good ! Appreciate the thought & refactor into routes, layouts, and views - feels much easier to navigate and, i agree, better positions us for a one-day refactor away from ReactNavi. 

All working as expected for me: 
- I can navigate to `/pay` & `/pay/invite` only when I have a valid payment request id param and the feature flag on
- Neither pay or invite pages show reset button in header 
- Unpublished / published routes for non S&R and S&R flows continue to work as expected (and _better_ for unpublished S&R flows :raised_hands:) 
- Sections and nav header displaying as expected across all routes when both feature flags are enabled

Left a few minor naming questions, but happy for this to merge whenever since new routes are properly feature flagged etc

editor.planx.uk/src/routes/pay.tsx
```diff
@@ -0,0 +1,101 @@
...
+        title: makeTitle("Make a payment"),
+        view: <MakePayment {...paymentRequest} />,
+      };
+    }),
+    "/invite": route(async (req) => {
+      const paymentRequest = await getPaymentRequest(req);
+      return {
+        title: makeTitle("Invite to pay"),
+        view: <InviteToPay {...paymentRequest} />,
+      };
```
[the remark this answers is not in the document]
DafyddLlyr: This is a really good catch.

I'd be in favour of a single naming convention internally and externally and there's no time like the present to standardise this.

I'll try and get feedback from Alastair and George on this one 👍

editor.planx.uk/src/routes/pay.tsx
DafyddLlyr: It's worth being consistent here, I've updated this to camelCase 👍

PR #1577, opened 2023-03-27
feat: Initial Invite to pay API

builditben:
Adds a simple endpoint to initiate a payment request for invite to pay.

Once called, the session will be locked and a payment request will be created.

The caller is responsible for setting which `sessionPreviewKeys` are used to store a `sessionPreviewData` object. The API (via planx-core) looks up those values from a locked session to ensure the stored preview data is available and an accurate reflection of what has been stored on the session.

Change depends on: https://github.com/theopensystemslab/planx-core/pull/7

Notes:
* The endpoint is public and anyone can lock a session that meets the invite to pay criteria.
* A session that has been locked already is currently showing as "session not found". Locked sessions cannot use this feature.
* Currently this doesn't validate breadcrumbs beyond checking for the `sessionPreviewKeys`. Adding a check for completion of the pay component would be a good idea and a nice quick update for a follow-on PR.

To test:
```
curl -X POST -H "Content-Type: application/json" -H "Accept: application/json" --data '{ "payeeEmail": "benjamin@opensytemslab.io", "payeeName": "Benjamin", "sessionPreviewKeys": [["_address", "title"]] }' https://api.1577.planx.pizza/invite-to-pay/$SESSION_ID
```

### DIFF
--- a/api.planx.uk/inviteToPay/inviteToPay.ts
+++ b/api.planx.uk/inviteToPay/inviteToPay.ts
@@ -1,0 +1,102 @@
+import { NextFunction, Request, Response } from "express";
+import { validate as validateUUID, version as getUUIDVersion } from "uuid";
+import type {
+  Session,
+  PaymentRequest,
+} from "@opensystemslab/planx-core/types/types";
+
+import { ServerError } from "../errors";
+import { _admin } from "../client";
+
+export async function inviteToPay(
+  req: Request,
+  res: Response,
+  next: NextFunction
+) {
+  const sessionId = req.params.sessionId;
+  const payeeEmail = req.body.payeeEmail;
+
+  if (!isValidIdFormat(sessionId)) {
+    return next(
+      new ServerError({
+        message: "invalid sessionId parameter",
+        status: 400,
+      })
+    );
+  }
+
+  if (!payeeEmail) {
+    return next(
+      new ServerError({
+        message: "JSON body must contain payeeEmail",
+        status: 400,
+      })
+    );
+  }
+
+  const session: Session = await _admin.getSessionById(sessionId);
+
+  if (!session) {
+    return next(
+      new ServerError({
+        message: "session not found",
+        status: 404,
+      })
+    );
+  }
+
+  // posts to this endpoint are expected to occur only for valid scenarios
+  // any validation errors are logged but undisclosed to the caller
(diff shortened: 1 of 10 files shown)


builditben:
[[API] Add /invite-to-pay/:sessionId route](https://trello.com/c/kUIM2ivH/2345-api-add-invite-to-pay-sessionid-route)

api.planx.uk/tests/mocks/inviteToPayData.ts
```diff
@@ -0,0 +1,41 @@
+import type { Session, PaymentRequest } from "@opensystemslab/planx-core";
+
+export const validEmail = "the-payee@opensystemslab.io";
+
+export const payee = {
+  name: "Pay Ee",
```
DafyddLlyr, line 6: 👏

api.planx.uk/server.ts
```diff
@@ -680,13 +684,14 @@ app.get("/error", async (res, req, next) => {
...
   const { status = 500, message = "Something went wrong" } = (() => {
-    if (errorObject instanceof Error && airbrake) {
+    if (airbrake && (errorObject instanceof Error || errorObject instanceof ServerError)) {
       airbrake.notify(errorObject);
       return {
         ...errorObject,
         message: errorObject.message.concat(", this error has been logged"),
       };
     } else {
+      console.log(errorObject);
```
builditben, line 694: This makes test output very noisy but it also makes errors visible in dev.

DafyddLlyr (approved):
Works as expected - thanks for also improving error handling along the way 👍 

It's good to explicitly state in the PR as well that locking is a public action which I don't think we've covered. I think this is totally fine for now - we could always harden it by including the session email in the request and connecting as the public user.

api.planx.uk/inviteToPay/inviteToPay.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (locked === false) {
+      throw new Error("session could not be locked");
+    }
+    paymentRequest = await _admin.createPaymentRequest({
+      sessionId,
+      payeeName,
+      payeeEmail,
+      sessionPreviewKeys,
+    });
+  } catch (e: unknown) {
```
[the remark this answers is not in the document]
builditben: Good point. I've now updated this to two try blocks with a session unlock on the second `catch`.

PR #1333, opened 2022-12-15
chore: Split up docker files and upgrade docker compose

ghost:
This PR:
- upgrades pizza and workflow OS to Ubuntu 22
- updates docker compose to use the plugin instead of the standalone (`docker compose` instead of `docker-compose`)
- creates a separate docker-compose.yml for development (no more restarting everything to pick up an api change. **Note**: this means that all devs will need to have this plugin and be able to run `docker compose`).
- adds a root set of pnpm scripts for starting and stopping containers
- prefer bash instead of shell for all scripts

PR #1579, opened 2023-03-30
build: revert to old update script using docker-compose for tpx demo env

jessicamcinchak:
the TPX pizza was created 2 months ago when we were still using `docker-compose`, so when I rebased and tried to push GIS changes today it errored because now we're using `docker compose`.

 i want to avoid destroying & re-creating the whole vultr instance because they've been editing content in at least ~10 flows.

how i expect this to work: 
- reverts pull request update script to use `docker-compose` like here: https://github.com/theopensystemslab/planx-new/pull/1333/files#diff-e82530b26a3296f5accde8f5cb20952d0b84ab13d94eb03f32b8477be20926bc
- docker-compose.yml still matches main and has `["seed"]` profile, so update _should not_ re-seed the database

anything i might be missing here? 

i tried running this locally and got this error, but I'm hoping that already in use port might just be a local quirk and will behave differently on vultr. 
Screenshot from 2023-03-30 20-25-13

if this doesn't work, i have a copy of the edited content and could manually recreate/update flows if necessary.

### DIFF
--- a/scripts/pullrequest/update.sh
+++ b/scripts/pullrequest/update.sh
@@ -1,16 +1,5 @@
-#!/usr/bin/env bash
-set -o errexit -o pipefail
-
-# run from project root
-cd "$(dirname $0)/../.."
-
-# set env for this shell
-set -o allexport
-source .env.pizza
-DOCKER_BUILDKIT=1
-set +o allexport
-
-docker compose \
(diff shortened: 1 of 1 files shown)


DafyddLlyr (approved):
✅

PR #1580, opened 2023-03-30
fix: on reconciliation, previously "completed" sections should still be tagged as "completed" not "ready to continue"

jessicamcinchak:
testing out a quick new thought I had around how we can correct section statues on reconciliation without needing to calculate whole paths using the breadcrumbs.

because we have distinct statuses and only allow continuous forward travel (eg no jumping ahead to non-consecutive sections), I think we can simply check:
- does more than one section currently have the status "READY TO CONTINUE"?
- if yes, keep that status for the most recently seen section node only (the one latest in order), but update each prior to be "COMPLETED" (because of forward travel rule, it _must_ have all been seen, and if it hasn't because of a content change it will have been already tagged as "NEW INFORMATION NEEDED")

i'm not sure this would hold up as neatly if we introduced all of the original "middle statuses" (eg "READY TO CONTINUE", "IN PROGRESS", and "READY TO START" that are trickier to distinguish), but that hasn't come up in feedback yet and i think this satisfies our current context :crossed_fingers: 

poking holes in this logic very welcome though!

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/navigation.test.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/navigation.test.ts
@@ -218,3 +218,48 @@ test("sectionStatuses() applies the NEW INFORMATION NEEDED status if a section w
     thirdSection: "CANNOT CONTINUE YET",
   });
 });
+
+test("sectionStatuses() corrects multiple in-progress statuses on reconciliation to reflect completed sections", () => {
+  setState({ flow: flowWithThreeSections });
+  initNavigationStore();
+
+  // Navigate forwards
+  record("firstSection", { auto: false });
+  record("firstQuestion", { answers: ["firstAnswer"] });
+  record("secondSection", { auto: false });
+  record("secondQuestion", { answers: ["secondAnswer"] });
+
+  // Mimic "reconciliation" by explicitly passing breadcrumbs instead of relying on cachedBreadcrumbs, assume no content has changed
+  const breadcrumbs = getState().breadcrumbs;
+  const statuses = sectionStatuses(breadcrumbs);
+
+  expect(statuses).toEqual({
+    firstSection: "COMPLETED",
+    secondSection: "COMPLETED",
+    thirdSection: "READY TO CONTINUE",
+  });
+});
+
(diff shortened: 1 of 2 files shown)


jessicamcinchak:
[Correctly tag "Completed" sections when resuming from Save & Return](https://trello.com/c/Jy7U2V3n/2337-correctly-tag-completed-sections-when-resuming-from-save-return)

editor.planx.uk/src/pages/FlowEditor/lib/store/navigation.ts
```diff
@@ -133,7 +133,7 @@ export const navigationStore: StateCreator<
 
     const sectionStatuses: Record<string, SectionStatus> = {};
     Object.keys(sectionNodes).forEach((sectionId) => {
-      if (updatedSectionNodeIds?.includes(sectionId)) {
+      if (updatedNodeIds?.includes(sectionId)) {
```
jessicamcinchak, line 136: this is just a minor argument naming update that I failed to properly commit/propagate through from previous work here https://github.com/theopensystemslab/planx-new/pull/1535#discussion_r1138668346)

editor.planx.uk/src/pages/FlowEditor/lib/__tests__/navigation.test.ts
```diff
@@ -218,3 +218,48 @@ test("sectionStatuses() applies the NEW INFORMATION NEEDED status if a section w
...
+  record("secondQuestion", { answers: ["secondAnswer"] });
+
+  // Mimic "reconciliation" by explicitly passing breadcrumbs instead of relying on cachedBreadcrumbs, assume no content has changed
+  const breadcrumbs = getState().breadcrumbs;
+  const statuses = sectionStatuses(breadcrumbs);
+
+  expect(statuses).toEqual({
+    firstSection: "COMPLETED",
+    secondSection: "COMPLETED",
+    thirdSection: "READY TO CONTINUE",
```
jessicamcinchak, line 239: "Mimicing reconciliation" is admittedly a bit strange for navigation store tests, but I think it's our best option until we have a full e2e with "resume" functionality. 

This test specifically fails against our prod code & sets each section status to `"READY TO CONTINUE"` which is confidence that my mocked sectionStatuses() arguments are accurate!

PR #1581, opened 2023-03-30
feat: Add works dates to BOPS payload

DafyddLlyr:
Please see discussions on ODP Slack here - https://opendigitalplanning.slack.com/archives/C0182MVK4AW/p1680010700510649

I've not added the `immune` key as I think what we're doing in `result` should be enough and duplicating it doesn't make a whole load of sense to me but happy to get feedback from BoPS and we can take it from there.


jessicamcinchak (approved):
Perfect, thanks! Agree there wasn't a strong case for repeating the flag under `works` key and that `result` should already satisfy what they need :+1:

DafyddLlyr:
This doesn't break anything as no date was being passed to BOPS currently, but I'll send a test payload from staging once merged and then let them know.

PR #1582, opened 2023-03-31
fix: Use `address.title` as fallback

DafyddLlyr:
Something that came up on the dev call the other day - `address.title` should always be set, `address.single_line_address` may not be set.

As `address.single_line_address` includes postcode, I've used this as the first option.

Going forward it will be nice to roll this logic up in the `Passport()` class in `planx-core` 👍


jessicamcinchak (approved):
:100:

Meant to clarify that we'd want this fallback logic in order to display postcode after we talked about this on the fly the other day, thanks for already reflecting that here :slightly_smiling_face:

PR #1584, opened 2023-03-31
fix: Uniform submission issue

DafyddLlyr:
Tidying up a bit of leftover logic that was missed when moving OneApp generation to `planx-core`

### DIFF
--- a/api.planx.uk/send/createSendEvents.ts
+++ b/api.planx.uk/send/createSendEvents.ts
@@ -25,17 +25,13 @@ const createSendEvents = async (req: Request, res: Response, next: NextFunction)
     }
 
     if ("uniform" in req.body) {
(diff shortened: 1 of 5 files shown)


jessicamcinchak (approved):

api.planx.uk/send/createSendEvents.ts
```diff
@@ -25,17 +25,13 @@ const createSendEvents = async (req: Request, res: Response, next: NextFunction)
     }
 
     if ("uniform" in req.body) {
-      // Only send valid payloads to Uniform
-      // Fail silently - error already caught in frontend by Airbrake
-      if (req.body.uniform.body.xml) {
```
DafyddLlyr, line 30: Removing this condition is a leftover from moving OneApp generation out of the frontend.

PR #1590, opened 2023-04-03
fix: Show toast on GraphQL error

DafyddLlyr:
Hit an issue earlier when rotating db connection details - Hasura returns as `200 OK` if the db connection string is invalid, alongside an error message.


jessicamcinchak (approved):
thanks for picking this up!

PR #1593, opened 2023-04-04
feat: Toggle maintenance page GitHub action

DafyddLlyr:
GitHub action which toggles the default root object in CloudFront in order to swap in the `error.html` maintenance page and `index.html` page as we're organising planned downtime.

Script tested locally and works using my (staging) AWS credentials, action not tested yet but keen to give it a go once merged against staging.


jessicamcinchak (approved):
Very handy, was nice to discover this yesterday - thanks for scripting!

PR #1595, opened 2023-04-04
fix: Minio S3 pathname

DafyddLlyr:
Hit this issue when trying to deal with turning question buttons to radios, file upload was failing. Would appreciate if somebody could also test this out locally 👌


jessicamcinchak (approved):
checked out this branch, was able to upload a file locally !

PR #1600, opened 2023-04-05
prod deploy: rotate postgres root user and metabase roles

jessicamcinchak:
See notes in #1591


DafyddLlyr (approved):

jessicamcinchak:
Closing without merging as we're reverting the original change here #1601

PR #1602, opened 2023-04-06
feat: Radio button questions

DafyddLlyr:
## What does this PR do?
 - Introduce radio buttons instead of plain buttons for answers on the `Question` component
 - This is hidden behind the `ALT_THEME` feature flag
 - Test flow - https://1602.planx.pizza/testing/radio-button-test
 - Styling is pretty rough atm sorry..!

## Next steps...
 - Pass to Ian for final styling (layout and positioning changes, styling of MUI Radio to be more like GovUK radio)
 - E2E tests of a flow with questions (for a11y)
 - Turn off feature flag, remove button questions


editor.planx.uk/src/@planx/components/shared/Buttons/ImageRadio.tsx
```diff
@@ -0,0 +1,144 @@
...
+export interface Props {
+  id: string;
+  title: string;
+  description?: string;
+  responseKey?: string | number;
+  img?: string;
+  onChange: RadioProps["onChange"];
+}
+
+const FallbackImage = styled(Box)({
```
DafyddLlyr, line 18: I spent way too long trying to work out a way to avoid duplicating these styles. As the `Image` is using `component="img"` it's more tricky than it should be.

editor.planx.uk/src/theme.ts
```diff
@@ -236,6 +236,18 @@ const getAltThemeOptions = (primaryColor: string): ThemeOptions => {
...
+      MuiRadio: {
+        defaultProps: {
+          disableFocusRipple: true,
+          sx: {
+            "& .MuiSvgIcon-root": {
+              fontSize: 32,
+            },
+          },
+        },
+      },
```
DafyddLlyr, line 249: I'm hoping that we can convert the radios to GovUK looking ones entirely here 🤞

jessicamcinchak (approved):
Working as expected (including keyboard-only nav & selection!) and seems ready for pass-off to me :+1: :art:

PR #1608, opened 2023-04-11
chore: Upgrade xml2js

DafyddLlyr:
Addresses https://github.com/theopensystemslab/planx-new/security/dependabot/6

`pnpm why xml2js` shows that it's a dependency of `aws-sdk`. This is directly updated in the API, but Pulumi haven't yet updated their dependencies so needs an overwrite in IaC package.json files. `pnpm audit --fix` has also picked up another few issues in these directories.


jessicamcinchak (approved):
thanks for addressing this one! this will replace #1604, correct?

DafyddLlyr:
> this will replace https://github.com/theopensystemslab/planx-new/pull/1604, correct?

Yep, I'll close that one now with a reference to this PR.

PR #1611, opened 2023-04-12
fix: add transition group props to Card fade

jessicamcinchak:
something that came up in Sections & Navigation feedback is slow navigation between cards, especially when going "back" to a previous section. this has also been a known issue with rendering the long "Project type" Checklist and when using the "change" link from Review ([ticket](https://trello.com/c/zoYs9u8U/2204-text-input-takes-1s-to-respond-to-keyboard-input)).

i think adding these props very mildly helps, but i honestly didn't come up with a great way to measure this and am definitley still experiencing some lag in my own test runs through LDCs. i think most of our lag is likely still in `handleSubmit` method on "continue", rather than the transition/rendering of the Card itself, which we can turn more attention to in a near-future logic sprint.

https://mui.com/material-ui/transitions/#performance-amp-seo
https://reactcommunity.org/react-transition-group/transition#Transition-prop-mountOnEnter

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
@@ -43,7 +43,12 @@ const Card: React.FC<Props> = ({
     path === ApplicationPath.SaveAndReturn && handleSubmit;
 
   return (
-    <Fade in={true} timeout={theme.transitions.duration.enteringScreen}>
+    <Fade
+      in={true}
+      timeout={theme.transitions.duration.enteringScreen}
+      mountOnEnter={true}
+      unmountOnExit={true}
+    >
       <Container maxWidth="md">
         <Box
           className={classes.container}


editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
```diff
@@ -43,7 +43,12 @@ const Card: React.FC<Props> = ({
     path === ApplicationPath.SaveAndReturn && handleSubmit;
 
   return (
-    <Fade in={true} timeout={theme.transitions.duration.enteringScreen}>
+    <Fade
```
jessicamcinchak, line 46: my first instinct was to wrap this fade in `<Suspense fallback={<DelayedLoadingIndicator />}>`, but I couldn't actually ever get to the loading indicator to render between slow Card transitions - I think because Fade is not actually a React lazily-loaded component (even though the `mountOnEnter` prop is used to "lazy mount" the component), and therefore Suspense is never "activated" ? https://react.dev/reference/react/Suspense

very open to feedback if I'm missing something obvious here!

jessicamcinchak:
other ideas from dev call about this: 
- trigger some variable at the beginning and end of the `record` method in the preview store (called by `handleSubmit` on "continue") to hook a loading indicator into
- is `currentCard()` ever undefined during these long transitions? if so, could that trigger a loading indicator?

jessicamcinchak:
[Add loading indicator during slow "section", "change" and "back" navigation](https://trello.com/c/FytfpaPK/2379-add-loading-indicator-during-slow-section-change-and-back-navigation)

builditben (approved):
I didn't notice a huge difference between this flow and what's on staging but it does seem smooth enough. Let's merge it.

There are real performance issues in the logic function. I think the best fix would be to encode more information about the structure of a flow in data structures rather than computing state on every interaction. We can optimize the state computation for sure (there's a lot of redundant calls which could be avoided/cached) but computing state ahead of time would make everything significantly faster (and a lot easier to reason about!).

PR #1616, opened 2023-04-12
update: all font sizes

ianjon3s:
Font size update across theme file (alt theme used).

Font weight update across component files.

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUpload/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUpload/Public.tsx
@@ -17,6 +17,7 @@ import { Store } from "pages/FlowEditor/lib/store";
 import type { handleSubmit } from "pages/Preview/Node";
 import React, { useEffect, useRef, useState } from "react";
 import { FileWithPath, useDropzone } from "react-dropzone";
+import { FONT_WEIGHT_STRONG } from "theme";
 import ErrorWrapper from "ui/ErrorWrapper";
 import { array } from "yup";
 
@@ -323,8 +324,8 @@ function Dropzone(props: any) {
             </Box>
             <Box flexGrow={1}>
               <Box
-                fontSize="caption.fontSize"
-                fontWeight={700}
+                fontSize="body2.fontSize"
+                fontWeight={FONT_WEIGHT_STRONG}
                 color="text.secondary"
               >
                 File {index + 1}
--- a/editor.planx.uk/src/@planx/components/FileUpload/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUpload/Public.tsx
@@ -17,7 +17,7 @@ import { Store } from "pages/FlowEditor/lib/store";
 import type { handleSubmit } from "pages/Preview/Node";
 import React, { useEffect, useRef, useState } from "react";
 import { FileWithPath, useDropzone } from "react-dropzone";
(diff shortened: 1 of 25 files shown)


jessicamcinchak (commented):
Spent some time on this pizza going through changes & it's looking really good :muscle: 

One minor thing I spotted so far: "change" links are looking a bit small I think on the "Result"/"ResultReason" component? 
Screenshot from 2023-04-19 15-17-44

Let's catch up soon to make a merge plan as some of this is behind the feature flag and some isn't!

ianjon3s:
Thanks for the review @jessicamcinchak , I've now updated the PR to include a fix for the small "change" text in the result component. I've also moved all of the styles out of the alt theme feature flag.

jessicamcinchak:
@ianjon3s - apologies it's taken me a moment to come back to this one - working through your most recent changes now! 

i'm working off a small test flow here where I'm adding one of each component type/state in case useful to others too https://1616.planx.pizza/buckinghamshire/styles-test

overall looking very good, here's a few small minor notes of what I'm still noticing (i'll keep adding here as I find em!): 
- "change" link on Result and PropertyInformation (property type override) is now sized properly, but Review page "change" link still looks quite small (perhaps can be refactored to better share code!)
Screenshot from 2023-05-04 11-02-28
- new font sizes displaying without the `ALT_THEME` feature flag :tada:, while this is great for the frontend I'm realizing it introduces some side-effects in the Editor that look a bit wonky - but I'm not sure how much this matters yet. I attached a few screenshot examples though, as a few of these might be really quick cleanups (eg small team names in entry nav list (flow names within team route seem propertly sized), really big button text and overlapping inputs in Question modals, etc)
Screenshot from 2023-05-04 10-32-25

Screenshot from 2023-05-04 10-37-11

DafyddLlyr:
Apologies for the delay in getting back to this one @ianjon3s !

A few little issues I spotted. A fair number of these are in the Editor which I appreciate you've not covered much probably so far. 

Going forward we might find we need to create some variants of components for the Editor (e.g. buttons and inputs). Docs: https://mui.com/material-ui/customization/theme-components/#creating-new-component-variants

**Question cards are now grey/transparent** 
These are actually `li`s, which might explain why they've picked up a style from elsewhere?

**More padding needed on ellipses on flow list**
This should break line a bit sooner so as to not overlap
image

**Editor modal buttons**
The correct styling / sizing isn't applied across these buttons. This is the modal you see when creating or editing any nodes in the editor
image
image

**Flow settings page**
Grey background is not covering the entire page, I guess as a result of larger fonts? This happens across all tabs here.
Also an issue with text size in dropdowns as flagged above by @jessicamcinchak 
image

**Back button spacing**
The focus style clips the text here
image

**Input component**
I think the input here needs to be shrunk a little to accommodate for "square meters" spilling onto two lines.
image

**Clipped placeholder on date input**
"MM" is clipped here now the text size is increased
image


**File upload**
Some of the text here seems a little small maybe? e.g. "File1"
image

ianjon3s:
[Update component styles to new design system](https://trello.com/c/KJYB8cPv/2177-update-component-styles-to-new-design-system)

jessicamcinchak (approved):
Thanks for a couple rounds of updates here @ianjon3s, all looking really good! 

Happy for this one to squash&merge whenever you're feeling content with wider feedback loop :tada:

PR #1621, opened 2023-04-14
feat(ci): Action failure notification

DafyddLlyr:
This introduces a new action which should run when an action fails. 

For now, I'm running on `pull_request` but I think this will actually just be unnecessary noise - we want it just to run on failed staging and prod deploys.

Next steps...
 - Merge this to `main` so we can run it
 - Open a PR with a failing test to force failure, and triggering of this action
 - If this action is triggered, update workflow to run on staging and prod deploys

### DIFF
--- a/.github/workflows/action-failure.yml
+++ b/.github/workflows/action-failure.yml
@@ -1,0 +1,22 @@
+name: Action failure notification
+
+on: 
+  workflow_call:
+    inputs:
+      environment:
+        required: true
+        type: string
+
+jobs:
+  notifications:
+    name: Failure notifications
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Slack Notification
+        uses: rtCamp/action-slack-notify@v2
+        env:
+          SLACK_CHANNEL: planx-notifications-internal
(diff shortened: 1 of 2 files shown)


.github/workflows/action-failure.yml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  notifications:
+    name: Failure notifications
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Slack Notification
+        uses: rtCamp/action-slack-notify@v2
+        env:
+          SLACK_CHANNEL: planx-notifications-internal
+          SLACK_USERNAME: PlanX Action Failure
```
jessicamcinchak, line 19: nit: since this is already an internal notification channel, is "Github Actions" a more meaningful username here - eg it more immediately points us to the source?

jessicamcinchak (approved):

.github/workflows/action-failure.yml
DafyddLlyr: This is a good shout - I'll make this change 👍

PR #1624, opened 2023-04-14
fix: Add correct dependency for notification failure job

DafyddLlyr:


jessicamcinchak (approved):

DafyddLlyr:
Merged without vultr completing, should've used `[skip pizza]` tag in last commit.

PR #1628, opened 2023-04-14
chore: manage Gov Notify template uuids consistently

jessicamcinchak:
Notify template uuids aren't actually secrets.

We publicly exposed them in Pulumi, but then read them from environment variables in the API and Docker config.

This simplifies our env files and stores them directly as variables in the API. We could take this a step further in the future and make a proper database table to read them from, but this feels like a solid first step in simplification.


jessicamcinchak:

DafyddLlyr:
When this is merged we'll also want to update the remote `.env` files on S3 👍

jessicamcinchak:
@DafyddLlyr - yep haven't forgotten about changing S3 files, but will purposely wait a couple days to do it as to not break existing pizzas !

PR #1630, opened 2023-04-14
chore: Update passport to v0.6.0

DafyddLlyr:
Follows on from https://github.com/theopensystemslab/planx-new/pull/1185

Neither `passport` not `cookie-session` have yet had an update that addresses this compatibility issue.

This solution was sourced [here](https://github.com/jaredhanson/passport/issues/904#issuecomment-1307558283) and is worth testing on staging as it will allow us to resolve this long running issue if it works.


jessicamcinchak (approved):
This seems really promosing, thanks for further researching! 

Happy to test once on staging.

PR #1626, opened 2023-04-14
chore: share `getHumanReadableProjectTypes` among email templates

jessicamcinchak:
Changes:
- Updates `getHumanReadableProjectTypes` utils method to accept a passport-like dictionary of key/values, rather than an entire session, allowing us to re-use it between Save & Return (by passing `lowcal_sessions.data.passport.data`) and Invite to Pay (by passing `payment_requests.session_preview_data`)
- Tidies up payment request emails & validation to reference new db fields `applicant_name` & `payment_amount`
- Updates test data & query mocks accordingly

Next up:
- Move Notify-related utils to their own subdirectory in the API

PR #1633, opened 2023-04-17
feat: Populate ITP email templates

DafyddLlyr:
| Agent Email | Nominee Email |
|--------|--------|
| image | image |

Relies on - 
 - https://github.com/theopensystemslab/planx-core/pull/16
 - https://github.com/theopensystemslab/planx-core/pull/17

### DIFF
--- a/api.planx.uk/inviteToPay/sendPaymentEmail.ts
+++ b/api.planx.uk/inviteToPay/sendPaymentEmail.ts
@@ -22,7 +22,7 @@ const sendSinglePaymentEmail = async (
 ) => {
   try {
     const { session, paymentRequest } = await validatePaymentRequest(paymentRequestId, template);
-    if (!session || !paymentRequest || Boolean(paymentRequest?.paidAt)) throw Error(`Invalid payment request: ${paymentRequestId}`);
+    if (!session || !paymentRequest) throw Error(`Invalid payment request: ${paymentRequestId}`);
     const config = await getInviteToPayNotifyConfig(session, paymentRequest);
     const recipient = template.includes("-agent") ? session.email : paymentRequest.payeeEmail;
     return await sendEmail(template, recipient, config);
--- a/api.planx.uk/inviteToPay/sendPaymentEmail.ts
+++ b/api.planx.uk/inviteToPay/sendPaymentEmail.ts
@@ -1,37 +1,30 @@
+import { PaymentRequest } from '@opensystemslab/planx-core';
 import { gql } from "graphql-request";
 import { Template, calculateExpiryDate, convertSlugToName, getClientForTemplate, getHumanReadableProjectType, sendEmail } from "../saveAndReturn/utils";
 import { InviteToPayNotifyConfig } from "../types";
+import { Team } from '../types';
+
+interface SessionDetails {
+  email: string;
+  flow: {
+    slug: string;
+    team: Team;
+  }
+};
+
+interface ValidatePaymentRequest {
+  query: PaymentRequest & { session: SessionDetails }
+};
 
 const sendSinglePaymentEmail = async (
   template: Template,
   paymentRequestId: string,
 ) => {
   try {
-    const payment = await validatePaymentRequest(paymentRequestId, template);
-    if (!payment) throw Error("Unable to send payment email, cannot find payment request");
-    
-    const humanReadableProjectType = await getHumanReadableProjectType(payment.session_preview_data);
-    const config: InviteToPayNotifyConfig = {
-      personalisation: {
-        id: payment.session_id,
-        serviceName: convertSlugToName(payment.session.flow.slug),
-        emailReplyToId: payment.session.flow.team.notify_personalisation.emailReplyToId,
-        helpEmail: payment.session.flow.team.notify_personalisation.helpEmail,
-        helpOpeningHours: payment.session.flow.team.notify_personalisation.helpOpeningHours,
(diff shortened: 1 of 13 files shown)


scripts/seed-database/container.sh
```diff
@@ -14,7 +14,7 @@ REMOTE_PG="$2"
 psql --command="\\COPY (SELECT * FROM users) TO '/tmp/users.csv' (FORMAT CSV, DELIMITER ';')" "${REMOTE_PG}"
 
 # fetch teams
-psql --command="\\COPY (SELECT id, name, slug, theme, settings, domain, notify_personalisation FROM teams) TO '/tmp/teams.csv' (FORMAT CSV, DELIMITER ';')" "${REMOTE_PG}"
+psql --command="\\COPY (SELECT id, name, slug, theme, settings, notify_personalisation FROM teams) TO '/tmp/teams.csv' (FORMAT CSV, DELIMITER ';')" "${REMOTE_PG}"
```
DafyddLlyr, line 17: Having `domain` populated outside of prod means that email links don't point at the correct environment.

editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
```diff
@@ -148,7 +143,7 @@ const InviteToPayForm: React.FC<InviteToPayFormProps> = ({
         <ReactMarkdownOrHtml source={nomineeDescription} openLinksOnNewTab />
       </Typography>
       <StyledForm onSubmit={formik.handleSubmit}>
-        <InputLabel label="Full name (optional)" htmlFor="payeeName">
```
DafyddLlyr, line 151: This is required in the form and db so I've removed "(optional)" from the label. We don't currently use this field anywhere though so this is something we could make optional if we wanted to.

api.planx.uk/saveAndReturn/utils.ts
```diff
@@ -296,11 +296,9 @@ const markSessionAsSubmitted = async (sessionId: string) => {
  * Get formatted list of the session's project types
  */
 const getHumanReadableProjectType = async (
```
DafyddLlyr, line 298: This function is also required in the frontend, so I'll move this (in this form) to `planx-core` in another PR.

jessicamcinchak: This is related to a small change I opened here: https://github.com/theopensystemslab/planx-new/pull/1626

We made slightly different decisions in refactoring: I updated the argument to be a passport-like dict, whereas you've gone for the list of passport variable strings. 

In my mind, a small benefit to the first is then that we have fewer instances of the actual passport variable `proposal.projectType` to manage & update if it continues to live directly in the function rather than at the call-level.

But not strongly tied to one way or the other. Just flagging as an expected merge conflict.

DafyddLlyr: I think your instinct is correct here - a function related to project types is best responsible for knowing which key to fetch as opposed to multiple callers 👍 We'll merge your PR, then I'll rebase and resolve conflicts.

DafyddLlyr:
Rebased now that https://github.com/theopensystemslab/planx-new/pull/1626 has been merged.

api.planx.uk/inviteToPay/sendPaymentEmail.ts
```diff
@@ -1,37 +1,30 @@
...
-        fee: Number(payment.payment_amount).toLocaleString("en-GB", { style: "currency", currency: "GBP" }),
-        projectType: humanReadableProjectType || "Project type not submitted",
-        address: payment.session_preview_data?._address.title,
-        expiryDate: calculateExpiryDate(payment.created_at),
-      }
-    };
-
-    const recipient = template.includes("-agent") ? payment.session.email : payment.payee_email;
+    const { session, paymentRequest } = await validatePaymentRequest(paymentRequestId, template);
+    if (!session || !paymentRequest || Boolean(paymentRequest?.paidAt)) throw Error(`Invalid payment request: ${paymentRequestId}`);
```
jessicamcinchak, line 25: I think we can remove `... || Boolean(paymentRequest?.paidAt` here since after rebasing it's also below on L72 in the validate method and will throw a more specific "already paid" error.

api.planx.uk/inviteToPay/sendPaymentEmail.ts
```diff
@@ -71,17 +66,53 @@ const validatePaymentRequest = async (
...
+      team: {
+        domain,
+        slug: teamSlug,
+      },
+    },
+  } = session
+  // Use custom domain if available or fall back to PlanX URL
+  const serviceURL = domain ? `https://${domain}/${flowSlug}` : `${process.env.EDITOR_URL_EXT}/${teamSlug}/${flowSlug}`;
+  return `${serviceURL}/pay?paymentRequestId=${paymentRequest.id}`
+};
```
jessicamcinchak, line 116: thanks for adding correct fallback logic for magic links here and fixing the initial assumption I had around the `domain`!

jessicamcinchak (approved):
Looks good ! A couple minor final comments, but happy for this to merge whenever

api.planx.uk/inviteToPay/sendPaymentEmail.ts
DafyddLlyr: Thank you - missed this in the rebase.

PR #1642, opened 2023-04-18
chore: add a notify/ subdirectory in API, move email-related methods out of saveAndReturn/

jessicamcinchak:
- Adds a `/notify` subdirectory to the API for general email-sending-related methods
- Reorganises general email methods from `/saveAndReturn` into `/notify`


jessicamcinchak:

api.planx.uk/notify/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  try {
+    await notifyClient.sendEmail(templateId, emailAddress, config);
+    const returnValue: {
+      message: string;
+      expiryDate?: string;
+    } = { message: "Success" };
+    if (template === "expiry") softDeleteSession(config.personalisation.id!);
+    if (template === "save")
+      returnValue.expiryDate = config.personalisation.expiryDate;
```
DafyddLlyr, line 72: This is an example of the sort of tight coupling I'm now seeing with hindsight - email and save and return logic closely entangled.

A pattern to follow might be similar the the middleware pattern used by @builditben  in the recent pay proxy PR where we hand off responsibility to certain handlers as part of the process.

No action needed here, just wanted to flag an example.

DafyddLlyr (approved):
Great start in logically breaking these two concepts apart 🙌

PR #1646, opened 2023-04-18
fix: hide fee banner on second page of Pay component and on MakePayment page

jessicamcinchak:
- Shows the fee banner by default on the "Pay" page, always hidden by default on the "Invite someone else to pay" page
- Adds optional prop `hideFeeBanner` to override default "Pay" page behavior so we can reuse the component in `MakePayment` page with a hidden banner

chrome-capture-2023-3-18

https://trello.com/c/45whNb0u/2385-99-outstanding-itp-tasks


editor.planx.uk/src/@planx/components/Pay/Public/Confirm.tsx
```diff
@@ -61,7 +62,10 @@ const PayBody: React.FC<PayBodyProps> = (props) => (
     {!props.error ? (
       <Card>
         <PayText>
-          <Typography variant="h3">
+          <Typography
+            variant="h3"
+            component={props.hideFeeBanner ? "h2" : "h3"}
+          >
```
jessicamcinchak, line 68: This was a trickier than I anticipated - the banner is integral to the heading heirarchy of the "Pay" page and a11y tests failed when rendering the `MakePayment` implementation with a hidden banner without this adjustment

DafyddLlyr (approved):
One thing to consider here - see comment.

It may make more sense to wait a short while until https://github.com/theopensystemslab/planx-new/pull/1651 is ready to pick this one up? Still aiming for later today!

editor.planx.uk/src/@planx/components/Pay/Public/Confirm.tsx
jessicamcinchak: Ah yep, good shout - I hadn't even appreciated that's the same `<Banner />` yet actually - like you suggest below let's pick that one up separately after #1651 :+1:

PR #1647, opened 2023-04-19
fix: make sure `teams` seed scripts references same columns

jessicamcinchak:
followup to #1633 

`pnpm docker:seed` was throwing this error because `domain` was only removed from the `container.sh` script, not `container.sql`  - I admittedly am not sure why both are still needed, seems like something we should look into simplifying - but this fixes local dev environment in short term!

Screenshot from 2023-04-19 12-10-04


DafyddLlyr (approved):

DafyddLlyr:
Thanks for catching this 👍

PR #1649, opened 2023-04-20
fix: update Pay content copy, add secondary page title to Editor fields

jessicamcinchak:
A small tweak that came up in conversation with Ian yesterday ! Additionally better aligns other default text to the newest Figma. I expect this is still one of few default content iterations, but will give me a slightly more realistic heirarchy of headers to expect for #1646 

Screenshot from 2023-04-20 09-26-43


DafyddLlyr (approved):
✅ Perfect!

PR #1651, opened 2023-04-20
feat: E2E invite to pay journey (minus submission)

DafyddLlyr:
# What does this PR do?
- Wires up end to end invite to pay journey (without submissions)
- Adds data to the nominee page

## Invite to Pay - Nominee journey overview
This is a basic happy path that describes the changes journey implemented in this PR

```mermaid
sequenceDiagram
  participant Frontend
  participant API
  participant GovPay API 
  participant Hasura

  Frontend->>+API: fetchPayment()
  API->>+GovPay API: Proxy request
  GovPay API->>-API: Return GovUkPayment
  API->>-Frontend: Return GovUkPayment
  Frontend->>+Frontend: Redirect to GovPay, make payment, return
  Frontend->>+API: fetchPayment()
  API->>+GovPay API: Proxy request
  GovPay API->>-API: Return GovUkPayment
  API->>+Hasura: Update session data and paymentRequest
  Hasura->>-API: Success
  API->>-Frontend: Return GovUKPayment
```


api.planx.uk/pay/proxy.ts
```diff
@@ -8,6 +8,7 @@ export const usePayProxy = (options: Partial<Options>, req: Request) => {
     onProxyReq: fixRequestBody,
     headers: {
       ...(req.headers as NodeJS.Dict<string | string[]>),
+      "content-type": "application/json",
```
DafyddLlyr, line 11: This is required by GovPay APIs - a `P0920` error response is returned if it's missing. Docs: https://docs.payments.service.gov.uk/api_reference/#gov-uk-pay-api-error-codes

I think it's likely being stripped by one of the proxy layers as the request initially has this, but as it's mandatory it seems to make more sense to set here than pass through.

editor.planx.uk/src/pages/Pay/MakePayment.tsx
```diff
@@ -207,3 +265,24 @@ function computePaymentState(govUkPayment: GovUKPayment | null): PaymentState {
   // PaymentStatus.cancelled, PaymentStatus.error, PaymentStatus.failed,
   return PaymentState.Failed;
 }
+
+const parseSessionPreviewData = (sessionPreviewData: unknown) => {
```
DafyddLlyr, line 269: We could just cast this value to a typed `SessionPreviewData`, or update the type of `PaymentRequest` in `planx-core` but this approach which will make sure we have what's required.

editor.planx.uk/src/pages/Pay/MakePayment.tsx
```diff
@@ -44,137 +50,189 @@ enum PaymentState {
...
-  const [payment, setPayment] = useState(
-    sessionPreviewData.govUkPayment || undefined
-  );
+  const [isLoading, setIsLoading] = useState(true);
+  const [payment, setPayment] = useState<GovUKPayment | undefined>(undefined);
+  const flowName = useStore((state) => state.flowName);
+  const theme = useTheme();
 
+  // Pass async errors up to ErrorBoundary
+  const [errorMessage, setErrorMessage] = useState<string | undefined>();
```
DafyddLlyr, line 69: Errors thrown inside an async function (e.g. `updatePaymentState()`) don't get caught by the `ErrorBoundary` so we need a bit of manual handling for this here.

api.planx.uk/pay/pay.ts
```diff
@@ -19,40 +20,86 @@ export async function makePaymentViaProxy(
...
+          });
+          return responseBuffer;
+        }
+      ),
+    },
+    req
+  )(req, res, next);
+}
+
+export async function makeInviteToPayPaymentViaProxy(req: Request, res: Response, next: NextFunction) {
```
DafyddLlyr, line 58: I did spend a bit of time attempting to follow the pattern of creating a function which just accepted a callback instead of being invite to pay specific without much luck.

The complexity here (and one of the reasons that I suspect `fetchPaymentViaProxyWithCallback()` exists) is that `http-proxy-middleware` must be the _last_ middleware layer (as it calls `res.send()`). 

There's a good thread on this here alongside a few possibly more generic solutions - https://github.com/chimurai/http-proxy-middleware/issues/248

editor.planx.uk/src/pages/Pay/MakePayment.tsx
```diff
@@ -44,137 +50,189 @@ enum PaymentState {
...
-            handleSuccess();
-            break;
-        }
-      });
-    };
+    if (errorMessage) throw Error(errorMessage);
+  }, [errorMessage]);
+
+  useEffect(() => {
+    // If payment is completed, we don't need to fetch data from GovPay
```
DafyddLlyr, line 75: Also - it would fail. We would throw an error when trying to update the session and paymentRequest a second time as no `affected_rows` would be returned.

editor.planx.uk/src/pages/Pay/MakePayment.tsx
```diff
@@ -44,137 +50,189 @@ enum PaymentState {
...
+        term: "Project type",
+        details: rawProjectTypes.join(", "),
+      },
+    ];
+
+    if (paidAt) {
+      data.push({
+        term: "Paid at",
+        details: format(Date.parse(paidAt), "dd MMMM yyyy"),
+      });
```
DafyddLlyr, line 181: I don't think we've really decided what happens when a journey is completed and a payee revisits the payment link. 

This is just one simple option - they still see the success page and some basic details, until we sanitise the `payment_requests` records.

hasura.planx.uk/migrations/1681897578801_alter_table_public_payment_requests_add_column_govpay_payment_id/up.sql
```diff
@@ -0,0 +1,4 @@
+alter table "public"."payment_requests" add column "govpay_payment_id" text
+ null unique;
+
+comment on column "public"."payment_requests"."govpay_payment_id" is E'ID returned by GovPay when payment is initialised';
```
jessicamcinchak, line 4: :+1: thanks for comment here

i noticed auto-generated Hasura changelog comment on this PR reads `lowcal_sessions` not `payment_requests` which is a bit strange - maybe just not getting properly updated?

editor.planx.uk/src/pages/Pay/MakePayment.tsx
```diff
@@ -44,138 +50,190 @@ enum PaymentState {
...
+
+  const PaymentDetails = () => {
+    const data = [
+      { term: "Application type", details: flowName },
+      {
+        term: "Fee",
+        details: formattedPriceWithCurrencySymbol(toDecimal(paymentAmount)),
+      },
+      {
+        term: "Address",
```
jessicamcinchak, line 168: nit: this is referred to as "Property address" on the Confirmation page, rather than just "Address". Happy to bookmark and revisit in a future finalisation of copy (I'd guess we want to double check consistency of Notify templates too)

jessicamcinchak (approved):
Pizza working end-to-end for both invite to pay & "normal" pay journeys :tada: very very satisfying :sparkles: 

Thanks for tackling this and explaining the tricky bits you hit. A few minor comments, but I don't think any effect functionality so happy for this merge whenever as I know plenty of tidy up tasks will still build on it!

hasura.planx.uk/migrations/1681897578801_alter_table_public_payment_requests_add_column_govpay_payment_id/up.sql
DafyddLlyr: Yeah I think you're right - it's not picking up this change, but is picking up the change to `lowcal_sessions` made in `tables.yml`.

Something that's worth being aware of in future PRs!

api.planx.uk/inviteToPay/paymentRequest.ts
```diff
@@ -85,28 +85,76 @@ export const fetchPaymentRequestViaProxy = fetchPaymentViaProxyWithCallback(
...
+export const addGovPayPaymentIdToPaymentRequest = async (
+  paymentRequestId: string,
+  govUKPayment: GovUKPayment,
+) => {
+  const query = gql`
+    mutation AddGovPayPaymentIdToPaymentRequest($paymentRequestId: uuid!, $govPayPaymentId: String) {
+      update_payment_requests(
+        where: {
+          _and: {
+            id: { _eq: $paymentRequestId }
```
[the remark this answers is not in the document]
DafyddLlyr: Thanks - this is a good improvement! 💪

editor.planx.uk/src/ui/DescriptionList.tsx
```diff
@@ -0,0 +1,45 @@
...
+  data: DescriptionListItem[];
+}
+
+const List = styled("dl")(({ theme }) => ({
+  display: "grid",
+  gridTemplateColumns: "1fr 2fr",
+  marginTop: theme.spacing(4),
+  marginBottom: theme.spacing(4),
+  "& > *": {
+    borderBottom: "1px solid lightgrey",
```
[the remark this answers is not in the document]
DafyddLlyr: Yeah this was pretty much pinched `SummaryList` with the exact intention of picking this exact task up! 😄 Originally I thought this PR would be just styles / populating this table but I got sidetracked onto wiring up the journey.

I'd hope that we can use this `DescriptionList` component as the definitive one now - `PropertyInformation` should be quick/easy and I think there will be a bit more work in fitting it into `Review`

builditben (commented):
Thanks for finishing putting this together :+1:

PR #1653, opened 2023-04-20
chore: add "lock session" warning text to invite to pay page

jessicamcinchak:
Screenshot from 2023-04-20 15-39-15

https://trello.com/c/45whNb0u/2385-99-outstanding-itp-tasks

### DIFF
--- a/editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
@@ -1,3 +1,4 @@
+import ErrorOutline from "@mui/icons-material/ErrorOutline";
 import Box from "@mui/material/Box";
 import Button from "@mui/material/Button";
 import Link from "@mui/material/Link";
@@ -54,6 +55,17 @@ const StyledForm = styled("form")(({ theme }) => ({
   display: "flex",
   flexDirection: "column",
   gap: theme.spacing(3),
+}));
+
+const StyledWarningBox = styled(Box)(({ theme }) => ({
+  border: `solid 2px ${theme.palette.grey[400]}`,
+  backgroundColor: theme.palette.background.paper,
+  padding: theme.spacing(2),
+  marginTop: theme.spacing(2),
+  marginBottom: theme.spacing(2),
+  display: "flex",
+  flexDirection: "row",
+  alignItems: "center",
 }));
 
 const SubmitButton: React.FC = () => (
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
```diff
@@ -220,6 +232,13 @@ const InviteToPayForm: React.FC<InviteToPayFormProps> = ({
             }}
           />
         </InputLabel>
+        <StyledWarningBox>
+          <ErrorOutline />
```
jessicamcinchak, line 236: `<ErrorOutline />` is the same icon used in the Notice component

editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
```diff
@@ -220,6 +232,13 @@ const InviteToPayForm: React.FC<InviteToPayFormProps> = ({
...
+          <Typography variant="body2" ml={2} fontWeight="bold">
+            Selecting "continue" locks your application and you'll no longer be
+            able to make changes.
```
jessicamcinchak, line 239: any strong opinions if this text should be editor-configurable? it's quite general and vague (eg no expiration date, etc) so i figured for now it's not likely to change?

DafyddLlyr: No I would say that letting editors configure this is not necessary 👍

editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
```diff
@@ -56,6 +57,17 @@ const StyledForm = styled("form")(({ theme }) => ({
   gap: theme.spacing(3),
 }));
 
+const StyledWarningBox = styled(Box)(({ theme }) => ({
```
DafyddLlyr, line 60: nit: I imagine we'll be using this again in other parts of @ianjon3s' redesign - could we make it a component that lives in `editor.planx.uk/src/ui`?

DafyddLlyr (approved):

editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
jessicamcinchak: yep this is a fair shout - moving over now :+1: I _think_ `editor.planx.uk/@planx/components/shared/Preview/` is actually the more fitting location based on other styled container-like components there (but the difference of this subdirectory to `src/ui` has always been a bit vague for me!)

PR #1665, opened 2023-04-24
fix: Capture XML payload for Uniform applications audit

DafyddLlyr:


api.planx.uk/send/uniform.ts
```diff
@@ -414,13 +414,16 @@ const createUniformApplicationAuditRecord = async ({
   localAuthority: string,
   submissionDetails: UniformSubmissionResponse,
   }): Promise<UniformApplication> => {
+  const xml = await _admin.generateOneAppXML(payload?.sessionId);
```
DafyddLlyr, line 417: Regenerating this here shouldn't be too expensive and it's an async action so time is not of the essence. The alternative of generating it once and passing along to where it's required felt like quite a lot of code pollution for a small fix.

hasura.planx.uk/migrations/1682351174497_alter_table_public_uniform_applications_add_column_xml/up.sql
```diff
@@ -0,0 +1,4 @@
+alter table "public"."uniform_applications" add column "xml" xml
```
DafyddLlyr, line 1: I'm not super familiar with the `xml` type in Postgres, but it seems a suitable choice? 

Docs: https://www.postgresql.org/docs/current/datatype-xml.html

jessicamcinchak (approved):
thanks for catching this !

PR #1666, opened 2023-04-25
chore: update planx-core

builditben:
Updates planx-core and imports to work with [@theopensystemslab/planx-core#15](https://github.com/theopensystemslab/planx-core/pull/15)

### DIFF
--- a/api.planx.uk/.eslintrc
+++ b/api.planx.uk/.eslintrc
@@ -16,7 +16,7 @@
         "argsIgnorePattern": "^_"
       }
     ],
-    "@typescript-eslint/no-non-null-assertion": "warn",
+    "@typescript-eslint/no-non-null-assertion": "off",
     "jest/expect-expect": [
       "error",
       {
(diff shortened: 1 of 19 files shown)


jessicamcinchak (approved):

api.planx.uk/.eslintrc
```diff
@@ -16,7 +16,7 @@
         "argsIgnorePattern": "^_"
       }
     ],
-    "@typescript-eslint/no-non-null-assertion": "warn",
+    "@typescript-eslint/no-non-null-assertion": "off",
```
DafyddLlyr, line 19: I've got somewhat of a solution for this (it's generally used for env vars which we can handle a bit better). Inspired by this... https://www.youtube.com/watch?v=q1im-hMlKhM

DafyddLlyr (approved):

api.planx.uk/.eslintrc
DafyddLlyr: Another benefit would be that we can validate all env vars up front and lose the dispersed `assert()` statements across the API

builditben: In general, I'm finding that the non-null assertion is an easy way to avoid more complicated casts and type checks. It's a bit ugly but, from my point of view, hits the sweet spot between being strictly typed and easy to follow.

PR #1668, opened 2023-04-25
chore: update e2e tests and add initial tests for sections

builditben:
Updates to planx-core required some updates to e2e tests which was also an opportunity for a tidy-up of the typechecking which was failing. I've added code checks to the CI workflow and fixed broken typing issues. I also did a small refactor to share code while adding some tests for sections.

In the process, I found some tests that were not set-up correctly and should actually be failing. I've skipped them out in this PR so I can fix them in the next (to avoid too big a change).

### DIFF
--- a/e2e/src/flows/sections-flow.ts
+++ b/e2e/src/flows/sections-flow.ts
@@ -1,5 +1,5 @@
-import { ComponentType } from "@opensystemslab/planx-core";
-import type { FlowGraph } from "@opensystemslab/planx-core/types/types";
+import { ComponentType } from "@opensystemslab/planx-core/types";
+import type { FlowGraph } from "@opensystemslab/planx-core/types";
 
 export const flow: FlowGraph = {
   _root: {
--- a/e2e/src/flows/sections-flow.ts
+++ b/e2e/src/flows/sections-flow.ts
@@ -58,7 +58,7 @@ export const flow: FlowGraph = {
     edges: ["NUFNkSkxqR"],
   },
   T1X91EICsC: {
-    data: { text: "Mulit-select", allRequired: false },
+    data: { text: "Multi-select", allRequired: false },
     type: ComponentType.Checklist,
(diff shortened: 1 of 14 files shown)


e2e/src/flows/sections-flow.ts
DafyddLlyr, line 1: Elsewhere in E2E tests we have flows as `json` files being imported (we also do this in some `planx-new` tests).

I massively prefer this version with a typed object as a pattern we should be using - it means that as types change, we have to update and check tests and we reduce the drift between the expected and actual behaviour.

e2e/src/helpers.ts
```diff
@@ -1,25 +1,26 @@
...
-  process.env.DEBUG
-    ? console.log(...args)
-    : () => {
-      // silent
-    };
+  process.env.DEBUG_LOG && console.log(...args);
 }
 
 // a collection of useful playwright actions
+// these could evolve into fixtures: https://playwright.dev/docs/test-fixtures
```
DafyddLlyr, line 15: Helpful thank you - this is the bit of documentation I needed pointed towards when we were talking yesterday. I'll read up 👍

e2e/src/helpers.ts
DafyddLlyr, line 1: This is gold - exactly what I had in mind and had started on basically - great to see.

DafyddLlyr (approved):
Great work, and great to see the E2E toolkit expanding

e2e/src/helpers.ts
```diff
@@ -57,8 +58,147 @@ export async function getTeamPage({
...
+}: {
+  page: Page;
+  title: string;
+  answer: string;
+}) {
+  const group = await findQuestion({ page, title });
+  await group.getByRole("button", { name: answer }).click();
+}
+
+export async function answerChecklist({
```
DafyddLlyr, line 150: I'd be curious to know if this works as-is for expandable checklists - I think the labels are always in the DOM and just hidden when the "parent" is closed. No changes needed - just a thought / comment!

PR #1674, opened 2023-05-03
feat: add all section statuses

builditben:
This PR changes and adds some new section states: 

Change:
 :heavy_minus_sign:  `NotStarted: "CANNOT CONTINUE YET"`
 :heavy_plus_sign:  `NotStarted: "CANNOT START YET"`
 
Change:
:heavy_minus_sign:  `InProgress: "READY TO CONTINUE"`
:heavy_plus_sign:   `ReadyToContinue: "READY TO CONTINUE"`
  
 Additions:
 :heavy_plus_sign: `Started: "CANNOT CONTINUE YET"`
 :heavy_plus_sign:  `ReadyToStart: "READY TO START"`
 
These may not be the final text for each sections but the focus of this PR is the functionality for each of these scenarios.
This ticket took a stab at styling the new statuses but any suggestions welcome (or that could be part of the clickable tags PR).

The PR adds some end-to-end tests and updates some unit tests for a `computeSectionStatuses` function. If you spot any scenarios not covered, please do suggest them.

Computing sections no longer requires a call to `upcomingCardIds()` which is a computationally expensive function so this _may_ resolve some performance issues.

Testing: 

This flow should allow all test scenarios (feel free to make edits if not): https://1674.planx.pizza/testing/sections-test
image

This ticket is branched from #1562 so testing save and return journeys with both states of `featureFlags.toggle("NAVIGATION_UI")` on and off should test most of the scenarios in that ticket too.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -4,8 +4,8 @@ on:
     types:
       - opened
       - synchronize
-    branches:
-      - main
+    branches-ignore:
+      - production
 
 env:
   DOMAIN: planx.pizza
--- a/e2e/src/helpers.ts
+++ b/e2e/src/helpers.ts
@@ -14,8 +14,20 @@ export function log(...args: any[]) {
 // a collection of useful playwright actions
 // these could evolve into fixtures: https://playwright.dev/docs/test-fixtures
 
-export async function debugPageConsole(page: Page) {
+export function debugPageConsole(page: Page) {
   page.on("console", (msg) => console.log(msg.text()));
+}
+
+// used to detect `{ "setItem": ... }`, `{"getItem": ... }`
+// and "section state updated" debug messages on state transitions
+export function waitForDebugLog(page: Page) {
+  return new Promise((resolve) => {
+    page.on("console", (msg) => {
+      if (msg.type() == "debug") {
+        resolve(true);
+      }
+    });
+  });
 }
 
 export async function createAuthenticatedSession({
@@ -75,9 +87,6 @@ export async function saveSession({
   page: Page;
   context: Context;
 }): Promise<string | undefined> {
-  // ensure store has had time to update any previous answer before saving
-  await new Promise((resolve) => setTimeout(resolve, 500));
-
   const pageResponsePromise = page.waitForResponse((response) => {
     return response.url().includes("/send-email/save");
   });
@@ -99,11 +108,11 @@ export async function returnToSession({
   page: Page;
   context: Context;
   sessionId: string;
-  shouldContinue: boolean;
+  shouldContinue?: boolean;
 }) {
   const returnURL = `/${context.team?.slug}/${context.flow?.slug}/preview?analytics=false&sessionId=${sessionId}`;
   log(`returning to http://localhost:3000/${returnURL}`);
-  await page.goto(returnURL);
+  await page.goto(returnURL, { waitUntil: "load" });
   await page.locator("#email").fill(context.user?.email);
   if (shouldContinue) {
     const waitPromise = page.waitForResponse((response) => {
@@ -116,24 +125,30 @@ export async function returnToSession({
 
 export async function clickContinue({
   page,
+  waitForLogEvent = false,
   waitForResponse = false,
 }: {
   page: Page;
   waitForResponse?: boolean;
+  waitForLogEvent?: boolean;
 }) {
-  const waitPromise = waitForResponse
-    ? page.waitForResponse(
-        (response) =>
-          response.url().includes("graphql") && response.status() === 200
-      )
-    : new Promise((resolve) => setTimeout(resolve, 500));
-  await page.getByTestId("continue-button").click();
-  await waitPromise;
+  if (waitForLogEvent || waitForResponse) {
+    const waitPromise = waitForResponse
+      ? page.waitForResponse((response) => {
+          return response.url().includes("/graphql");
+        })
+      : waitForDebugLog(page); // assume debug message is triggered on state transition
+    await page.getByTestId("continue-button").click();
+    await waitPromise;
+  } else {
+    await page.getByTestId("continue-button").click();
+  }
 }
 
 export async function clickBack({ page }: { page: Page }) {
-  await new Promise((resolve) => setTimeout(resolve, 500));
+  const waitPromise = waitForDebugLog(page); // assume debug message is triggered on state transition
   await page.getByRole("button", { name: "Back", exact: true }).click();
+  await waitPromise;
 }
 
 export async function fillInEmail({
--- a/e2e/src/helpers.ts
+++ b/e2e/src/helpers.ts
@@ -2,7 +2,7 @@ import { expect } from "@playwright/test";
 import type { Page, Browser, Locator } from "@playwright/test";
 import { findSessionId, generateAuthenticationToken } from "./context";
 import type { Context } from "./context";
-import type { GraphQLClient } from "graphql-request";
+import { getGraphQLClient } from "./context";
 
 // utility functions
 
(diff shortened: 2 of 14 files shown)


builditben:
[Can the first section only be labelled 'Ready to start' instead of ready to 'continue'?](https://trello.com/c/NPQBRPOR/2382-can-the-first-section-only-be-labelled-ready-to-start-instead-of-ready-to-continue)

.github/workflows/pull-request.yml
```diff
@@ -4,8 +4,8 @@ on:
     types:
       - opened
       - synchronize
-    branches:
-      - main
+    branches-ignore:
+      - production
```
builditben, line 8: This change was to allow this PR to target a non-master branch. That branch has since merged but I think this is still a useful addition.

e2e/src/helpers.ts
```diff
@@ -14,6 +14,22 @@ export function log(...args: any[]) {
 // a collection of useful playwright actions
 // these could evolve into fixtures: https://playwright.dev/docs/test-fixtures
 
+export function debugPageConsole(page: Page) {
+  page.on("console", (msg) => console.log(msg.text()));
+}
```
builditben, line 19: This is a handy function for printing editor console messages in a playwright run for debugging.

editor.planx.uk/src/@planx/components/Section/model.ts
```diff
@@ -10,3 +13,71 @@ export const parseSection = (
...
+    // check for unseen sections
+    if (allSeenNodeIds.includes(sectionId)) {
+      sectionStatuses[sectionId] = SectionStatus.Started;
+    } else {
+      sectionStatuses[sectionId] = SectionStatus.NotStarted;
+    }
+  }
+
+  return sectionStatuses;
+}
```
builditben, line 83: This function should ideally live in planx-core where it could also be used to compute the section statuses of exported data.

editor.planx.uk/src/@planx/components/Section/model.ts
```diff
@@ -10,3 +13,71 @@ export const parseSection = (
   title: data?.title || "",
   ...parseMoreInformation(data),
 });
+
+export function computeSectionStatuses({
```
jessicamcinchak, line 17: could you share about decision to move this function out of navigation store and into the component model instead? my first impression is that it feels out a bit out of place here!

editor.planx.uk/src/@planx/components/Section/model.ts
jessicamcinchak: Do we have an existing case of an export data format that will need the section _status_ ? I've been assuming that for exported application data, we always assume it's a finished app so all sections have been completed and we only care about sending the section _names_ as a grouping mechanism for the rest of the flow nodes? I figure statuses are really only valuable to applicants, not validation officers. 

Therefore I'd picture this function in the navigation store and not planx-core longterm!

jessicamcinchak (approved):
Thanks for this, overall looks good! The subtle differences between "READY TO START" and "READY TO CONTINUE" do read a lot more naturally now as your navigating.

Spent some time on the pizza and noticed a couple minor edge cases if you can recreate, but don't think either are blockers to this initial piece of work going in:
- On reconciliation, if I have a section that "needs updated", it correctly tags the section status on the reconciliation overview page, but then when I hit "continue" if the first node shown to me is also a Section, it will have different statuses (eg no longer "needs updated"), which is pretty confusing. Ideally, I think the goal here is to click "continue" and go immediately to the first non-section node that needs updated though? 
- Editors have added sections to Prior Approval already (https://1674.planx.pizza/lambeth/apply-for-prior-approval/preview) and I did hit one performance/timeout bug when testing on the fullsized content. I navigated forwards all the way to the first quesiton of the "About you" (section 3), and hit "back" in the upper left to go back to the section overview node and my browser window timed out eventually prompting me to wait or kill the page - I didn't have a chance to grab anything meaningful from the console. Navigating via the section header links _does_ seem a bit speadier though than previously!

editor.planx.uk/src/@planx/components/Section/model.ts
builditben: Yes, fair point. It's not helpful for exported data but it could still be helpful to determine the status of an application for pre-processing steps (for example, checking that a flow is complete before locking a session for invite-to-pay) but this isn't an immediate concern so this can stay here for now.

editor.planx.uk/src/@planx/components/Section/Public.tsx
```diff
@@ -133,23 +150,27 @@ export function SectionsOverviewList(props: SectionsOverviewListProps) {
 
 const getTagBackgroundColor = (theme: Theme, title: string): string => {
   const backgroundColors: Record<string, string> = {
+    [SectionStatus.NeedsUpdated]: theme.palette.action.focus, // GOV UK YELLOW for now, check Figma
+    [SectionStatus.ReadyToContinue]: lighten(theme.palette.primary.main, 0.9),
+    [SectionStatus.ReadyToStart]: lighten(theme.palette.primary.main, 0.9),
+    [SectionStatus.Started]: lighten(theme.palette.primary.main, 0.8),
```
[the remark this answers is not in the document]
builditben: I'll gladly defer to others for decisions on styles. Any concrete suggestions on what these should be? @ianjon3s any recommendations?

editor.planx.uk/src/@planx/components/Section/model.ts
builditben: It's a pure function that is only being used by this one component so I thought the model makes the most sense. The store should ideally be for managing state rather than grouping/sharing functionality. That's my view anyway.

editor.planx.uk/src/@planx/components/Section/model.ts
```diff
@@ -10,3 +13,71 @@ export const parseSection = (
...
+  currentCard: Store.node | null;
+  breadcrumbs: Store.breadcrumbs;
+  cachedBreadcrumbs?: Store.cachedBreadcrumbs;
+  isReconciliation?: boolean;
+  alteredSectionIds?: string[];
+}): Record<string, SectionStatus> {
+  const seenPastNodeIds = Object.keys({ ...breadcrumbs });
+  const seenUpcomingNodeIds = cachedBreadcrumbs
+    ? Object.keys({ ...cachedBreadcrumbs })
+    : [];
```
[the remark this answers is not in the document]
builditben: Good to know what "seen" implicitly means for others in the team. I've updated this to the (hopefully) less ambiguous "visited" which doesn't imply anything about end-user visibility, just graph behavior.

editor.planx.uk/src/@planx/components/Section/Public.tsx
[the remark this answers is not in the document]
ianjon3s: Here are the specs for colour:

NEW INFO NEEDED
Background: #FAFF00
Text: theme.palette.text.primary

COMPLETED
Background: theme.palette.success.dark
Text: #FFFFFF

READY TO START
READY TO CONTINUE
Background: #E8F1EC
Text: theme.palette.success.dark

CANNOT START YET
CANNOT CONTINUE YET
Background: theme.palette.background.paper
Text: theme.palette.text.secondary

builditben:
Colours updated to match designs (thanks @ianjon3s!):

image

@jessicamcinchak thanks for pointing out the additional edge cases. I've created a [trello ticket](https://trello.com/c/VIAFGA4D/2447-skip-first-section-card-on-reconciliation) for the reconciliation status page workflow since it's not directly related to this PR.
It is a shame that this doesn't solve broader performance issues but it should speed up section statuses themselves. I guess overall performance should be a question we look into more during logic work, especially if it's causing pages to hang.

PR #1675, opened 2023-05-03
feat: Create `TeamStore`

DafyddLlyr:
## What does this PR do?
 - Creates a `TeamStore` which is populated when routes are loaded
 - This replaces relying on the `react-navi` `useCurrentRoute()` hook as appropriate - this makes testing simpler and makes it a little simpler to reliably get team details without prop-drilling
 - Populates `InviteToPay` with team details instead of hardcoding them

### Testing
- I'm currently part way through some E2E tests which will check this once merged/rebased


editor.planx.uk/src/@planx/components/PlanningConstraints/Public.test.tsx
```diff
@@ -22,27 +17,16 @@ jest.spyOn(SWR, "default").mockImplementation((url: any) => {
...
-        fn="property.constraints.planning"
-        handleSubmit={handleSubmit}
-      />
-    </MockedProvider>
+    <PlanningConstraints
+      title="Planning constraints"
+      description="Things that might affect your project"
+      fn="property.constraints.planning"
+      handleSubmit={handleSubmit}
+    />
```
jessicamcinchak, line 29: :scissors: this diff is such a nice demo of changes / how much these store methods simplify !

editor.planx.uk/src/pages/FlowEditor/lib/store/settings.ts
```diff
@@ -4,25 +4,24 @@ import { client } from "lib/graphql";
...
-    flowSlug: string,
-    newSettings: FlowSettings
-  ) => Promise<number>;
+  updateFlowSettings: (newSettings: FlowSettings) => Promise<number>;
   updateGlobalSettings: (newSettings: { [key: string]: TextContent }) => void;
 }
 
 export const settingsStore: StateCreator<
-  SettingsStore,
+  SettingsStore & TeamStore & SharedStore,
```
builditben, line 20: I'm not quite understanding where the dependency for the `SharedStore` is coming from. Could you explain what this does?
Is it perhaps to get the `flowSlug` from `get()` below? 
Does including the shared store have any meaningful overhead?

jessicamcinchak (approved):
This reduces lots of complexity, really nice to see! :tada: 

Spent some time on the pizza and all components are behaving as expected.

Only noticed one thing that didn't behave as expected: 
- from this flow https://1675.planx.pizza/buckinghamshire/search/preview?analytics=false, i was running console commands like `window.api.getState().teamName` to confirm the store was setting details correctly
- all work as expected except `...notifyPersonalisation`, which doesn't seem to be getting set even though its populated in the database - can you recreate? 
- i didn't try sending an email yet, but would guess these details aren't get getting properly set/passed along to the templates at the moment? 

otherwise looks good & happy for this to merge whenever you're happy with testing status.

editor.planx.uk/src/pages/Pay/InviteToPay.tsx
```diff
@@ -60,19 +62,18 @@ const InviteToPay: React.FC<PaymentRequest> = ({ createdAt }) => {
             if you have any questions about your application or this service
           </li>
         </List>
-        {/* TODO: Get this from team data (store?) */}
         <Box pt={2}>
           <Typography variant="body2" sx={{ margin: 0 }}>
-            <b>Telephone</b> 0300 131 6000
+            <b>Telephone</b> {team.notifyPersonalisation?.helpPhone}
```
builditben, line 67: Should all of these be in a conditional block if `team.notifyPersonalisation` isn't set?

DafyddLlyr:
>from this flow https://1675.planx.pizza/buckinghamshire/search/preview?analytics=false, i was running console commands like window.api.getState().teamName to confirm the store was setting details correctly
>
>all work as expected except ...notifyPersonalisation, which doesn't seem to be getting set even though its populated in the database - can you recreate?
>
>i didn't try sending an email yet, but would guess these details aren't get getting properly set/passed along to the templates at the moment?

So this is actually intentional but I can change if it makes more sense?

I've made `notifyPersonalisation` on the `Team` type optional in the public client and I'm not querying (and adding to `TeamStore`) on routes other than `/pay`. To fix this we just need to add this field to the query on other routes.

I didn't include it as it's not something that's currently required outside of `/pay` routes, but if you think it's clearer / cleaner to always have the same `TeamStore` contents it's a quick fix 👍

builditben (approved):
Nice to see a change towards more consistent handling of state like this. I've a few questions about how this works but the changes look good to me.
I tested the standard Pay flow and all looks good there.

DafyddLlyr:
I've actually gone ahead and made that change - I think it's more consistent (and expected?) to just always have it there.

editor.planx.uk/src/pages/FlowEditor/lib/store/settings.ts
DafyddLlyr: That's correct - it's so that we can just grab `flowSlug` as opposed to passing it into the function.

My understanding is as follows - 
 - All the stores are merged in `editor.planx.uk/src/pages/FlowEditor/lib/store/index.ts`
 - This change just allows TypeScript to know it has access to these variables which are controlled in another store - it's the type of the "getter" function essentially

PR #1676, opened 2023-05-03
fix: Standard error handling on `InviteToPayForm`

DafyddLlyr:
Removes browser validation, in favour of validation via Formik

### DIFF
--- a/editor.planx.uk/src/@planx/components/Pay/Public/Pay.test.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Public/Pay.test.tsx
@@ -165,7 +165,7 @@ describe("Confirm component with inviteToPay", () => {
     ).toBeInTheDocument();
   });
 
-  it("displays an error if you submit an applicant display name", async () => {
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/@planx/components/Pay/Public/Pay.test.tsx
```diff
@@ -149,6 +149,42 @@ describe("Confirm component with inviteToPay", () => {
...
+    await user.type(
+      await screen.findByLabelText("Email"),
+      "test@opensystemslab.io{enter}"
+    );
+    expect(
+      await screen.findByText("Enter the full name of the person paying")
+    ).toBeInTheDocument();
+  });
+
+  it("displays an error if you submit an applicant display name", async () => {
```
jessicamcinchak, line 168: nit: should be "displays an error if you _do not_ submit an..." I think

jessicamcinchak (approved):
working as expected now, thanks !!

PR #1679, opened 2023-05-04
feat: Clean-up ITP emails on changes to `paid_at`

builditben:
This trigger and stored procedure cleans up hasura events when `paid_at` is updated to a non-null value.

Testing via direct db manipulation:

1. create a lowcal session
2. create a payment requests for that session
3. Note the creation of pending events at https://hasura.1679.planx.pizza/console/events/one-off-scheduled-events/pending
4. set the `paid_at` field in the payment request to `now()`
5. Note the removal of pending events for that payment request at https://hasura.1679.planx.pizza/console/events/one-off-scheduled-events/pending

### DIFF
--- a/hasura.planx.uk/migrations/1683197635599_delete_payment_request_events_on_paid/down.sql
+++ b/hasura.planx.uk/migrations/1683197635599_delete_payment_request_events_on_paid/down.sql
@@ -1,0 +1,2 @@
+DROP TRIGGER IF EXISTS delete_payment_request_scheduled_email_events_trigger ON payment_requests;
+DROP FUNCTION IF EXISTS delete_payment_request_scheduled_email_events;
--- a/hasura.planx.uk/migrations/1683197635599_delete_payment_request_events_on_paid/up.sql
+++ b/hasura.planx.uk/migrations/1683197635599_delete_payment_request_events_on_paid/up.sql
@@ -1,0 +1,21 @@
+DROP FUNCTION IF EXISTS delete_payment_request_scheduled_email_events;
+CREATE OR REPLACE FUNCTION delete_payment_request_scheduled_email_events()
+RETURNS TRIGGER AS $$
+BEGIN
+  DELETE FROM hdb_catalog.hdb_scheduled_events
+  WHERE comment IN (
+    'payment_reminder_' || OLD.id,
+    'payment_expiry_' || OLD.id
+  );
+  RETURN NULL;
+END
+$$ LANGUAGE plpgsql VOLATILE;
+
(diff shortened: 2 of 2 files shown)


builditben:
[[Data retention / event cleanup] Figure out how session/payment invitation expiration timelines work & statuses](https://trello.com/c/UnmsnVzy/2349-data-retention-event-cleanup-figure-out-how-session-payment-invitation-expiration-timelines-work-statuses)

hasura.planx.uk/migrations/1683197635599_delete_payment_request_events_on_paid/up.sql
```diff
@@ -0,0 +1,28 @@
...
+    'payment_expiry_agent_' || OLD.id
+  );
+  RETURN NULL;
+END
+$$ LANGUAGE plpgsql VOLATILE;
+
+-- trigger to set up function
+DROP TRIGGER IF EXISTS delete_payment_request_scheduled_email_events_trigger ON payment_requests;
+
+CREATE TRIGGER delete_payment_request_scheduled_email_events_trigger
```
jessicamcinchak, line 21: nit: could this be `CREATE OR REPLACE TRIGGER...` and therefore negate the need for the `DROP...` command first? Similar question with the function definition!

hasura.planx.uk/migrations/1683197635599_delete_payment_request_events_on_paid/down.sql
```diff
@@ -0,0 +1,2 @@
+DROP TRIGGER IF EXISTS delete_payment_request_scheduled_email_events_trigger ON payment_requests;
```
jessicamcinchak, line 1: not that I ever expect this down migration to run, but I think because there's a `COMMENT` on the trigger this would technically fail without a `CASCADE` to drop the corresponding comment as well ?

jessicamcinchak (approved):
Works exactly as expected on pizza, thank for this !

jessicamcinchak (approved):
Works exactly as expected on pizza, thank for this !

hasura.planx.uk/migrations/1683197635599_delete_payment_request_events_on_paid/down.sql
builditben: Good thinking but [comments are automatically dropped when their object is dropped](https://www.postgresql.org/docs/current/sql-comment.html#:~:text=Comments%20are%20automatically%20dropped%20when%20their%20object%20is%20dropped). I did run some up/down commands and both work fine :+1:

PR #1681, opened 2023-05-05
feat: Add `SaveResumeButton` to pay component

DafyddLlyr:
Simpler than expected!

### DIFF
--- a/editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
@@ -13,7 +13,7 @@ import { useStore } from "pages/FlowEditor/lib/store";
 import React from "react";
 import { useNavigation } from "react-navi";
 import useSWRMutation from "swr/mutation";
-import { PaymentStatus } from "types";
+import { ApplicationPath, PaymentStatus } from "types";
 import ErrorWrapper from "ui/ErrorWrapper";
 import Input from "ui/Input";
 import InputLabel from "ui/InputLabel";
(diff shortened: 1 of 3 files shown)


DafyddLlyr:
[Wire up "Save and continue later" button on Pay component](https://trello.com/c/gFsBj2vH/2433-wire-up-save-and-continue-later-button-on-pay-component)

jessicamcinchak (commented):
If we go with this approach (which I agree is very simple!), the Pay component with InviteToPay enabled will always show the `<SaveResumeButton />` even if the flow doesn't have a Send component (eg no other cards will show `<SaveResumeButton />`). 

For example, view this Pay component with feature flag on: https://1681.planx.pizza/buckinghamshire/temporary-pay-test/preview?analytics=false
Screenshot from 2023-05-05 09-59-58
(The link defaults to "Resume..." because we don't have a `sessionId` available). 

I think there's perhaps an argument to be made that this is still better than the current hard-coded-ness, and we know we don't _currently_ have any live flow content that supports Pay without Send so this scenario shouldn't be possible in production, but this might still lead to confusion in testing flows? Basically I got to this exact spot too and then made that checklist item to rethink :sweat_smile:

DafyddLlyr:
Ha! Good catch 😅 Thought this was a bit too simple!

I'll take another look at it 👍

editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
```diff
@@ -258,12 +259,7 @@ const InviteToPayForm: React.FC<InviteToPayFormProps> = ({
       >
         {"I want to pay for this application myself"}
       </Button>
-      <Typography variant="body2">or</Typography>
-      <Link component="button">
-        <Typography variant="body2">
-          Save and return to this application later
-        </Typography>
-      </Link>
+      <SaveResumeButton />
```
jessicamcinchak, line 262: Maybe one way to address the bug of currently showing on _all_ services is to check if the current route has a `?sessionId=` param, and if it does conditionally show the `<SaveResumeButton />`? That feels a little hacky, but far less complicated/tricky than getting into `handleSubmit`/`Card` props?

jessicamcinchak (approved):
Works exactly as expected now, thanks for figuring out the `ApplicationPath` solution - it feels clean and simple! 

I'll go ahead and merge this since you're out until next week and update the corresponding Trello checklist item :heavy_check_mark:

PR #1682, opened 2023-05-05
chore: refactor NotifyConfig types to be more consistent

jessicamcinchak:
Key changes:
- Adds a `MinimumNotifyPersonalisation` interface to be more explicit about the values _all_ Notify template access, then domain-specific personalisations extend the minimum
- Renames `id` to `sessionId` for clarity
- Populates `serviceLink` for Pay templates (was only on Save&Return previously) per request from Hannah

Future cleanup: 
- Tidy up which personalisation variables are required  versus optional _per_ domain/template - this feels a little flakey still, I think we're using `?` more generously than ideal to account for possibly undefined query results ? 

Testing https://1682.planx.pizza/buckinghamshire/email-testing
- Check key templates like "save", "resume", "invite-to-pay", "send-to-email", "confirmation" can all be generated & sent as expected


jessicamcinchak:

api.planx.uk/types.ts
```diff
@@ -61,53 +61,50 @@ export interface LowCalSession {
...
-    emailReplyToId: string;
-    expiryDate?: string;
-    id?: string;
-    serviceName: string;
-    sessionId: string;
-  };
+  personalisation: EmailSubmissionNotifyPersonalisation;
+}
+
+interface InviteToPayNotifyPersonalisation extends MinimumNotifyPersonalisation {
```
builditben, line 95: nit: Since this is an internal data structure (one we control), and since this is the exact set of data fields required for the template, would this be better modeled as a `type` (with fewer optional fields if possible). An `interface` is a bit looser than a `type` but I think we're aiming at quite strong typing for email templates.

builditben (approved):

api.planx.uk/types.ts
jessicamcinchak: Thanks for highlighting this distinction - updated `MinimumNotifyPersonalisation` to be a `type`, the rest are still interfaces to make use of `extend` :+1:

PR #1686, opened 2023-05-09
feat: Make next section tag clickable

builditben:
This is a response to user research where people sometimes attempt to click the tag instead of the continue button.

Rather than making the tag a full-fledged button (i.e. with accessibility features) my approach here is to retain the emphasis on the "Continue" button as the primary navigation but to also allow the tag to be clickable. I think this adds the least confusion and complexity while allowing user expectations to be met. The tag has a hover state when clickable but otherwise doesn't advertise itself as a button.

**Testing**:

This flow contains sections and a send component: https://1686.planx.pizza/testing/new-sections-test/preview?analytics=false

The "NEEDS UPDATED", "READY TO CONTINUE" and "READY TO START" tags should link to the same location as the "Continue" button. The other tags should not have any changed styles or behavior.

### DIFF
--- a/editor.planx.uk/src/@planx/components/Section/Public.test.tsx
+++ b/editor.planx.uk/src/@planx/components/Section/Public.test.tsx
@@ -96,8 +96,8 @@ describe("SectionsOverviewList component", () => {
   it("renders correctly when the NAVIGATION_UI feature flag is toggled on", () => {
     setup(<SectionsOverviewList {...defaultProps} />);
 
-    expect(screen.getByText("Section one")).toBeInTheDocument();
-    expect(screen.getByRole("button")).toHaveTextContent("Section one");
+    const changeLink = screen.getByText("Change Section one");
+    expect(screen.getByText("Section one")).toContainElement(changeLink);
     expect(screen.getByText(SectionStatus.Completed)).toBeInTheDocument();
 
     expect(screen.getByText("Section two")).toBeInTheDocument();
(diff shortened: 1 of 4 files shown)


builditben:
[Make the 'Ready to continue' tag also serve as a continue button](https://trello.com/c/rryqnoNR/2370-make-the-ready-to-continue-tag-also-serve-as-a-continue-button)

builditben:
> Unfortunately I don't think the hover-state-only approach will work here because it means the status button isn't picked up in the page heirarchy during keyboard-only navigation. As I understand our accessibility requirements, I believe all interactive elements have to be keyboard reachable - I can't think of any exceptions to this besides in-map interactions!

This was an intentional decision. I think the accessibility support isn't lost here since this is a redundant means to navigating to the same place as "Continue". I actually think the page gets a little more confusing if this is included as a distinct navigation option for screen readers. Is there a strong reason to ensure this particular route is keyboard navigable rather than just a "bonus" navigation feature? 

@ianjon3s, what do you think. Is this an accessibility concern or is it right to think of this as a "bonus" navigation enhancement for keen mouse users?

ianjon3s:
I'd agree with @jessicamcinchak on the above, the replication of the button is in the same way illogical (or at least a replication) to both sighted and non-sighted users, so following the standard that tab order should follow the flow of the visual link structure would be the best route.

builditben:
@jessicamcinchak @ianjon3s - Section status tags are now buttons!

Tab based navigation works. Styles are _almost_ identical. Do let me know if this needs any more attention.

jessicamcinchak (approved):
I love the smooth tabbing :keyboard: :tada: thank you!

PR #1496, opened 2023-02-27
refact(ci): extract reusable workflow

gunar:
This PR DRYes the CI job that builds the React App. Similar code was being repeated in `pull-request.yml`, `push-main.yml`, and `push-production.yml`, and according to the rule of three it's time to refactor it.

The refactoring was not the sole goal though. This is a step towards #1442. In order to roll back a failed deployment, we have to actually roll forward. This means re-build (or possibly pull from cache) the frontend and any other dependencies. So really this PR came into existence to avoid adding yet another place (it'd be the 4th place) where CI builds the react app.

Rolling back failed deployments (#1442)  will unblock scheduled production deployments (#1409).

PR #1688, opened 2023-05-10
chore: Add manually triggered CI test suite

builditben:
To validate the state of a particular branch, it is helpful to be able to trigger just the automated tests for that branch.

This PR adds a CI workflow to run automated tests on demand.

### DIFF
--- a/.github/workflows/automated-tests.yml
+++ b/.github/workflows/automated-tests.yml
@@ -1,0 +1,162 @@
+name: Automated Tests
+on:
+  workflow_dispatch:
+
+env:
+  EDITOR_DIRECTORY: editor.planx.uk
+  PNPM_VERSION: 7.8.0
+  NODE_VERSION: 16.13.1
+
+jobs:
+  integration_tests:
+    name: Run Integration tests
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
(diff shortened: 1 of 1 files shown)


builditben:
Next steps after this is merged is to revive #1496 and break down these steps into reusable workflows.

.github/workflows/automated-tests.yml
```diff
@@ -0,0 +1,162 @@
...
+jobs:
+  integration_tests:
+    name: Run Integration tests
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.PIZZA_AWS_ACCESS_KEY_ID }}
```
jessicamcinchak, line 19: will these `PIZZA_` variables actually work if this is triggered on the `main` branch do you think? Just thinking about recent scenarios of broken Notify templates where it'd be convenient to check/compare state of `main`, not a local PR branch? 

testing actions is tricky though so happy for this to merge & fix forward if needed :slightly_smiling_face:

jessicamcinchak (approved):

.github/workflows/automated-tests.yml
builditben: These aren't specific to a Vultr deployment but general AWS keys. I was aiming to change as little as possible to make future de-duping easier later and I _think_ this should work :crossed_fingers:

PR #1690, opened 2023-05-10
feat: validate flow content on publish if Invite to Pay is enabled

jessicamcinchak:
**Validation rules introduced:**
If Invite to Pay is enabled, check that:
- There is at least one Send component
- There is at least one FindProperty (this component type sets `_address` by default)
- There is at least one Checklist that sets the passport variable `proposal.projectType` (we could be more flexible about component type here, but I think Checklist actually makes sense because a) it matches all existing content patterns and b) we expect to parse a _list_ of one or many project types answers)

This validation checks for the _existance_ of these components, but doesn't actually do any path calculations to ensure that all paths that reach the Pay node also are guaranteed to pass through the prior required nodes. That would be an interesting challenge to solve next! 

**Other changes:**
- Moves validation checks from `publishFlow` to `diffFlow` so we can check for & show validation errors earlier, _before_ being prompted to review altered nodes
- Refactors `lastPublishedTitle` & the publish dialog content in the `PreviewBrowser` to more clearly display validation warning messages if there are any (previously these showed up under publish button only)
Screenshot from 2023-05-10 22-15-45
- Renames `diffFlow` to `validateAndDiffFlow` to be explicit about what it's doing now - splitting these into two separate functions is obviously an appealing option too, but keeping as one right now allows for a more minimal refactor of how the publish change button & dialog UI behave in the Editor

### DIFF
--- a/api.planx.uk/editor/publish.test.ts
+++ b/api.planx.uk/editor/publish.test.ts
@@ -123,16 +123,82 @@ it("does not update if there are sections, but there is not a section in the fir
     });
 });
 
-it.todo("does not update if invite to pay is enabled, but there is not a Send component");
+it("does not update if invite to pay is enabled, but there is not a Send component", async () => {  
+  let { "Send": _, ...invalidatedFlow } = flowWithInviteToPay;
+  invalidatedFlow["_root"].edges?.splice(invalidatedFlow["_root"].edges?.indexOf("Send"));
+  
+  queryMock.mockQuery({
+    name: "GetFlowData",
+    matchOnVariables: false,
+    data: {
+      flows_by_pk: {
+        data: invalidatedFlow,
+      },
+    },
+  });
 
-it.todo("does not update if invite to pay is enabled, but there is not a FindProperty (`_address`) component");
+  await supertest(app)
+    .post("/flows/1/diff")
+    .set(authHeader())
+    .expect(200)
+    .then((res) => {
+      expect(res.body.message).toEqual("Cannot publish an invalid flow");
+      expect(res.body.description).toEqual("When using Invite to Pay, your flow must have a Send");
+    });
+});
 
-it.todo("does not update if invite to pay is enabled, but there is not a Checklist that sets `proposal.projectType`");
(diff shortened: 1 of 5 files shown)


jessicamcinchak:
[Add on-publish validation to check that flows using Invite to Pay have a Send component](https://trello.com/c/mFqoDvs6/2448-add-on-publish-validation-to-check-that-flows-using-invite-to-pay-have-a-send-component)

api.planx.uk/editor/publish.test.ts
```diff
@@ -77,13 +77,14 @@ it("does not update if there are sections in an external portal", async () => {
   });
 
   await supertest(app)
-    .post("/flows/1/publish")
+    .post("/flows/1/diff")
```
builditben, line 80: should this be it's own new test or are we intentionally dropping a test for `/publish`?

api.planx.uk/editor/publish.test.ts
```diff
@@ -184,26 +258,26 @@ it("updates published flow and returns altered nodes if there have been changes"
 const mockFlowData: Flow["data"] = {
```
builditben, line 258: Thanks for adding node names, that's helpful. 
nit: Could this data live in its own file to make the test file more focused?

api.planx.uk/editor/publish.test.ts
builditben: related nit: could `/publish` and `/diff` tests live each in their own `describe` block?

builditben (approved):
Looks good and it's great to see that you've also improved the visibility of the validation messages.

Pizza wasn't live so I haven't test driven this yet but the code changes look good.

jessicamcinchak:
@builditben rebuilding a pizza from scratch now - sorry I hit "ready" & walked out for lunch before noticing that step had failed ! thanks for initial look through already, should be ready for editor testing soon

api.planx.uk/editor/publish.test.ts
jessicamcinchak: this is simply a result of moving all validation steps (including prior validation work for sections) out of `/publish` and into `/diff` - happy to update these to be describe blocks for clarity!

PR #1693, opened 2023-05-11
fix: remove useEffect in Section now that feature flag is removed

jessicamcinchak:
should fix failed React build on staging deploy https://github.com/theopensystemslab/planx-new/actions/runs/4948901909


builditben (approved):
Sorry, bad merge conflict resolution on my part. Thanks for the fix.

PR #1695, opened 2023-05-15
feat: generate CSV data on the backend so that we can send to Uniform using only a sessionId

jessicamcinchak:
depends on https://github.com/theopensystemslab/planx-core/pull/30 & builds on #1671

Changes:
- [x] adds new endpoint GET `/admin/session/:sessionId/csv` (use with optional param `?download=true`)
- [x] updates POST `/uniform/:localAuthority` to only require a sessionId to be passed through from createSendEvents, no longer the full `csv` data string

To test:
- confirm you can submit an LDC app to Uniform staging successfully!
  - flow: https://1695.planx.pizza/buckinghamshire/apply-for-a-lawful-development-certificate/preview
  - event should only have sessionId in payload and succeed: https://hasura.1695.planx.pizza/console/events/one-off-scheduled-events/processed
   - audit table should have a success record: https://hasura.1695.planx.pizza/console/data/default/schema/public/tables/uniform_applications/browse
  
Future cleanup/considerations:
- Replace all instances of `makeCsvData` & `getBOPSParams` methods in Editor using `$public` client methods - will require headers to access session data first! 
- When we make the switch to `$public.generateCSVData`, we'll only be able to generate CSVs for sessions using `lowcal_sessions`, not local storage. Do we have any existing instances of offering CSVs for FOIYNPP or elsewhere for non-S&R services? I don't think there's any Confirmation components used in those services?


api.planx.uk/send/uniform.test.ts
```diff
@@ -113,45 +137,19 @@ describe("createUniformSubmissionZip", () => {
...
     expect(mockWriteZip).toHaveBeenCalledTimes(1);
   });
 
   test("geojson and location plan is excluded when not present", async () => {
     const payload = {
-      sessionId: "1234",
-      passport: { data: {} },
-      csv: [],
-      files: [],
+      sessionId: "noGeoJSON",
```
DafyddLlyr, line 150: I think a cleaner / clearer solution here is as follows.

- Make `getSessionById` it's own `jest.fn()` at the top of this file that always returns a promise of `mockLowcalSession`

```ts
const mockGetSessionById = jest
  .fn()
  .mockResolvedValue(mockLowcalSession);

jest.mock("../client", () => ({
  _admin: {
    getSessionById: () => mockGetSessionById(),
    ...
  }
})
```

- When we hit a test that has different behaviour, explicitly just overwrite the mock here inside this one test

```ts
const lowcalSessionWithoutBoundary: { ...mockLowcalSession };
delete lowcalSessionWithoutBoundary.data.passport.data["property.boundary.site"];

mockGetSessionById.mockResolvedValue(lowcalSessionWithoutBoundary);
```

This means that the logic is a little closer to the test instead of separated at the top of the file and can be more easily expanded on a test-by-test basis than just adding to the mock created at the top of the file 👍 

An example of something similar would be how `mockGenerateOneAppXML()` is being used in this file - it gives different responses in different tests.

DafyddLlyr (approved):
Thanks for picking this up and resolving - I also hadn't appreciated that this was outstanding initially.

One small suggestion on the tests but looking very good - some satisfying tidy ups here 🎉

DafyddLlyr:
Testing - 

- confirm you can submit an LDC app to Uniform staging successfully! ✅ 
 - event should only have sessionId in payload and succeed ✅ 
 - audit table should have a success record ✅ 
 
 > When we make the switch to $public.generateCSVData, we'll only be able to generate CSVs for sessions using lowcal_sessions, not local storage. Do we have any existing instances of offering CSVs for FOIYNPP or elsewhere for non-S&R services? I don't think there's any Confirmation components used in those services?
 
This is a great question which I don't think we've considered yet - I think we're ok atm, but I can see this ask coming up for sure. Sessions are the same type regardless of location so I'd like to think it's just a case of a new API endpoint which handles a POST of the session body? So `planx-core` has a `generateCSVDataBySessionId()` and a `generateCSVDataBySession()` or something but under the hood it's literally the same function but one has a lookup of `getSessionById()`?

PR #1697, opened 2023-05-15
feat: Add confirmation emails for invite-to-pay

ghost:
Confirmation emails are triggered by updating the `submitted_at` on lowcal sessions. If there is also a `locked_at` time on the session, invite-to-pay confirmation emails will be triggered instead of the standard save-and-return confirmation email.

**Code review notes**: 
- There are some changes adding missing email templates to the auth-middleware check. Is this correct?
- The new confirmation templates aren't generated directly like other templates - this breaks the current convention but keeps the interface simple. The API receives a request to send confirmation emails for the session and is responsible for working out which template to use and who to send it to.

**Testing notes**: You can manipulate the `submitted_at` and `locked_at` fields in the [database](https://hasura.1697.planx.pizza/console/data/default/schema/public/tables/lowcal_sessions/browse) and then check for successful [submission events](https://hasura.1697.planx.pizza/console/events/data/email_user_submission_confirmation/processed) and a set of emails in your inbox.

- Save and return confirmation emails should continue to work as normal when the session has not been locked
- Confirmation emails should be sent to both the payee and the applicant (note, different templates are used so the content should differ slightly)

### DIFF
--- a/api.planx.uk/auth/index.ts
+++ b/api.planx.uk/auth/index.ts
@@ -50,7 +50,6 @@ const useSendEmailAuth = (
     // Requires authorization - can only be triggered by Hasura scheduled events
     case "reminder":
     case "expiry":
-    case "submit":
     case "confirmation":
     case "invite-to-pay":
     case "invite-to-pay-agent":
@@ -63,6 +62,9 @@ const useSendEmailAuth = (
       return useHasuraAuth(req, res, next);
     // Public access
     case "save":
+      return next();
+    // Handled by other routes
+    case "submit":
     case "resume":
       return next();
     default: {
--- a/api.planx.uk/server.ts
+++ b/api.planx.uk/server.ts
@@ -591,12 +591,7 @@ app.post(
   useSendEmailAuth,
   routeSendEmailRequest
 );
-app.post(
-  "/resume-application",
-  sendEmailLimiter,
-  useSendEmailAuth,
-  resumeApplication
-);
+app.post("/resume-application", sendEmailLimiter, resumeApplication);
 app.post("/validate-session", validateSession);
 
 app.post("/invite-to-pay/:sessionId", inviteToPay);


hasura.planx.uk/metadata/tables.yaml
```diff
@@ -294,7 +294,8 @@
             {
               "payload": {
                 "sessionId": {{$body.event.data.new.id}},
-                "email": {{$body.event.data.new.email}}
+                "email": {{$body.event.data.new.email}},
+                "isLocked": {{inverse(empty($body.event.data.new.locked_at))}}
```
builditben, line 298: The lowcal submission webhook payload now includes the locked status from which we can infer if this is an invite-to-pay submission or not.

api.planx.uk/notify/notify.ts
```diff
@@ -32,21 +32,23 @@ const hybridEmailTemplates = {
 const privateEmailTemplates = {
   reminder: "43be4c11-a406-4381-b2be-056a1127455d",
   expiry: "9619f89d-5d33-4cb0-a365-42c431ea9db3",
-  confirmation: "8b82b606-defa-4daa-8fdb-e78b852b8ffb",
   submit: "7e77bdae-7379-4dd8-a8cc-086a0029163c",
+  confirmation: "8b82b606-defa-4daa-8fdb-e78b852b8ffb",
+  "confirmation-agent": "c7202e07-08cf-468e-a6a4-ac528d60d2f7", // TODO update template
+  "confirmation-payee": "c7202e07-08cf-468e-a6a4-ac528d60d2f7", // TODO update template
```
builditben, line 38: TODO: actually add the templates in Gov Notify

api.planx.uk/auth/index.ts
```diff
@@ -37,26 +39,34 @@ const useSendEmailAuth = (
...
     // Public access
     case "save":
+    case "resume":
       return next();
     default: {
-      return next({
-        status: 400,
-        message: "Invalid template",
-      });
+      return handleInvalidTemplate(template);
```
builditben, line 69: I added exhaustive checking here (for parity with a similar function I was introducing) and found missing cases for "submit" and "resume" templates. Are my assumptions about how these should be handled correct?

api.planx.uk/notify/routeSendEmailRequest.ts
```diff
@@ -1,44 +1,106 @@
...
+        } else {
+          return await handleSingleApplicationEmail();
+        }
+      }
+      case "confirmation-agent":
+      case "confirmation-payee":
+        // templates that are already handled by other routes
+        return invalidTemplate();
+      default:
+        return invalidTemplate(template);
```
builditben, line 94: Rather than routing based on the params, I'm explicitly routing to named templates and then delegating the responsibility to check the params. Exhaustive checking here also helps validate that new emails are routed correctly.

api.planx.uk/auth/index.ts
DafyddLlyr: This is a good catch thanks - some of this is handled inconsistently and it would be nice to sort out.

Your assumptions on the permissions for both templates are correct here 👍 

`resume` is left out as there's a `/resume-application` endpoint - I'd suggest we tidy that and update the endpoints so it goes through this middleware and then on to `resumeApplication()`.

`submit` is the same - it has it's own endpoint at `app.post("/email-submission/:localAuthority")`. Untangling that may be a little more involved as we need to calculate the correct local authority for a session but that's something that will be required here also - https://github.com/theopensystemslab/planx-new/pull/1709

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -294,7 +294,8 @@
             {
               "payload": {
                 "sessionId": {{$body.event.data.new.id}},
-                "email": {{$body.event.data.new.email}}
+                "email": {{$body.event.data.new.email}},
+                "lockedAt": {{$body.event.data.new.locked_at}}
```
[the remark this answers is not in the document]
DafyddLlyr: Perfect!

api.planx.uk/notify/routeSendEmailRequest.ts
```diff
@@ -1,44 +1,106 @@
...
+        return await handlePaymentEmails();
+      case "confirmation": {
+        // if the session is locked we can infer that a payment request has been initiated
+        const paymentRequestInitiated = Boolean(lockedAt);
+        if (paymentRequestInitiated) {
+          return await handleInviteToPayConfirmationEmails();
+        } else {
+          return await handleSingleApplicationEmail();
+        }
+      }
```
DafyddLlyr, line 88: Great - just what I had in mind here.

api.planx.uk/inviteToPay/sendPaymentEmail.test.ts
```diff
@@ -133,6 +141,8 @@ describe("Send email endpoint for invite to pay templates", () => {
 
   describe("'Payment Expiry' templates", () => {
     const templates = ["payment-expiry", "payment-expiry-agent"];
-    it.todo("soft deletes the payment_request when a payment expiry email is sent");
+    it.todo(
+      "soft deletes the payment_request when a payment expiry email is sent"
```
DafyddLlyr, line 145: We have `payment_request.paid_at` which serves as a proxy for this - would that work?

api.planx.uk/notify/routeSendEmailRequest.ts
```diff
@@ -1,44 +1,106 @@
...
+      const response = await sendAgentAndPayeeConfirmationEmail(sessionId);
+      return res.json(response);
+    };
+
+    switch (template) {
+      case "reminder":
+      case "expiry":
+      case "resume":
+      case "save":
+      case "submit":
```
DafyddLlyr, line 71: See previous comment - I believe "resume" and "submit" need to be handled slightly differently here.

DafyddLlyr (commented):
Super clear as always - a pleasure to review!

All looks great and works as expected, both emails received 👍 

Not hit approve just now due to the outstanding templates issue mentioned below. As is, hitting `/send-email/submit` or `/send-email/resume` won't work.

api.planx.uk/inviteToPay/sendPaymentEmail.test.ts
builditben: This was a linting fix and I wasn't planning on adding this test as part of this PR. Does this fit better into #1709 ?

api.planx.uk/auth/index.ts
builditben: @DafyddLlyr I've updated how this is handled but I'm not clear on what is needed around "submit". If there is still some remaining work to refactor, could it be a separate PR since it's not directly related to confirmation emails?  I'd be happy to catch-up about this to better understand the requirements.

builditben:
@DafyddLlyr I'm now explicitly skipping "submit" and "resume" templates in both switch statements. If these need handling differently, I'd be happy to pick that up in a separate PR to keep this one focused if that works (also I'm not clear how they should be handled differently but happy to discuss).

api.planx.uk/inviteToPay/sendPaymentEmail.test.ts
DafyddLlyr: Apologies - I missed that this wasn't a new additional. Happy to ignore and pick it up in another PR 👍

DafyddLlyr (approved):
Looking great 👍

PR #1698, opened 2023-05-16
FIX: font import weights

ianjon3s:
The current app imports the following font weights for Inter:

400 regular
600 semi-bold
800 extra-bold

This means that any bold text is being rendered at 800 weight, despite being referred to at bold or 700 weight in the CSS.

PR to correct the imports to be 400, 600, 700.


jessicamcinchak (approved):
Thanks for spotting this!

Issue #1699, opened 2023-05-16
S&R validation fails when "proposal.projectType" not in passport


builditben:
When attempting save and return, I get the error `Error: Unable to validate request. expected a list, but found null` which looks like it's coming from the `LookupHumanReadableProjectType` query.

Here's the full error from the API (pretty printed): 
```ts
{
  response: {
    errors: [
      {
        extensions: {
          path: "$.selectionSet.project_types.args.where.value._in",
          code: "validation-failed",
        },
        message: "expected a list, but found null",
      },
    ],
    status: 200,
    headers: {},
  },
  request: {
    query: `
      query LookupHumanReadableProjectType($rawList: [String!]) {
        projectTypes: project_types(where: { value: { _in: $rawList } }) {
          description
        }
      }
    `,
    variables: {},
  },
};
 ```
 
I suspect this is due to the fact that we are expecting "proposal.projectType" to exist and are otherwise attempting to format an empty list which fails (see `getSessionDetails` in [api.planx.uk/saveAndReturn/utils.ts](https://github.com/theopensystemslab/planx-new/blob/7116a4d0d6f335801777a149bbb294754b5be131/api.planx.uk/saveAndReturn/utils.ts#L141)):

PR #1700, opened 2023-05-16
fix: Handle empty project types

builditben:
Fix for #1699

### DIFF
--- a/api.planx.uk/saveAndReturn/utils.ts
+++ b/api.planx.uk/saveAndReturn/utils.ts
@@ -1,12 +1,10 @@
 import { SiteAddress } from "@opensystemslab/planx-core/types";
 import { format, addDays } from "date-fns";
 import { gql } from "graphql-request";
-import {
-  adminGraphQLClient as adminClient,
-} from "../hasura";
+import { adminGraphQLClient as adminClient } from "../hasura";
 import { LowCalSession, Team } from "../types";
 import { Template, getClientForTemplate, sendEmail } from "../notify/utils";
-import { _admin } from "../client";
+import { _admin as $admin } from "../client";
 
 const DAYS_UNTIL_EXPIRY = 28;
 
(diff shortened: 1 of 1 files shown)


api.planx.uk/saveAndReturn/utils.ts
```diff
@@ -1,12 +1,10 @@
...
 import { format, addDays } from "date-fns";
 import { gql } from "graphql-request";
-import {
-  adminGraphQLClient as adminClient,
-} from "../hasura";
+import { adminGraphQLClient as adminClient } from "../hasura";
 import { LowCalSession, Team } from "../types";
 import { Template, getClientForTemplate, sendEmail } from "../notify/utils";
-import { _admin } from "../client";
+import { _admin as $admin } from "../client";
```
jessicamcinchak, line 7: mind sharing reasoning for this syntax change?

jessicamcinchak (approved):
Thanks for spotting & fixing this one - all emails sending on pizza before & after address and projectType have been set :100: 

Worth noting that I'm seeing "Project type not submitted" in each email I've received though, even when my session passport has `"proposal.projectType" = ["alter.repair"]` - but I'm guessing that is a query match issue unrelated to this fix?

api.planx.uk/saveAndReturn/utils.ts
builditben: As @DafyddLlyr mentioned in the change that introduced `_admin` the underscore convention is generally used to indicate that a variable isn't referenced. I was working towards changing this shared planx-core client to `$admin` and `$public` instead if that works for everybody.

builditben:
> Thanks for spotting & fixing this one - all emails sending on pizza before & after address and projectType have been set 100
> 
> Worth noting that I'm seeing "Project type not submitted" in each email I've received though, even when my session passport has `"proposal.projectType" = ["alter.repair"]` - but I'm guessing that is a query match issue unrelated to this fix we can investigate separately

@jessicamcinchak thanks for testing this.  The issue you mentioned might just be a data issue. I don't see any matches for `alter.repair` in the db: https://hasura.editor.planx.dev/console/data/default/schema/public/tables/project_types/browse?filter=value%3B%24eq%3Balter.repair

PR #1701, opened 2023-05-16
fix: Fix resume state on reconciliation

builditben:
This change allows sections to behave as expected during reconciliation.

The ResumePage component is no longer competing with the Question component to update state after reconciliation. 

I attempted to solve some of the strange behavior we had noticed around state updating in unexpected ways but it looks like some weird behaviours are rooted in how `upcomingCards` and/or `computePassport` work. 

Here's my current theory about outstanding issues with reconciliation:
It seems that the preemptive auto-answering in `upcomingCards` can create new breadcrumb entries for auto-answerable cards. Reconciliation removes altered breadcrumbs but it doesn't remove breadcrumbs with matching "automations". When `upcomingCards` runs again (and it is called many times) it reconstructs and re-adds the removed breadcrumbs based on the computed passport (which is a core to the auto-answer feature). If I'm correct, an update to the reconciliation API could be the simplest way to resolve this (see #1704 ).

TLDR; This change does not fix issues with auto-answering in big LDC flows but follow-on PR #1704 does.

### DIFF
--- a/api.planx.uk/saveAndReturn/validateSession.ts
+++ b/api.planx.uk/saveAndReturn/validateSession.ts
@@ -96,13 +96,6 @@ export async function validateSession(
     const { reconciledSessionData, alteredSectionIds } =
       await reconcileSessionData({ sessionData, alteredNodes });
 
-    // store reconciled session data
-    await updateLowcalSessionData({
-      sessionId,
-      sessionData: reconciledSessionData,
-      email,
-    });
-
     const responseData: ValidationResponse = {
       message:
         "This service has been updated since you last saved your application." +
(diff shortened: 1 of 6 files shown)


editor.planx.uk/src/pages/Preview/Questions.tsx
```diff
@@ -105,9 +105,7 @@ const Questions = ({ previewEnvironment }: QuestionsProps) => {
       setPreviewEnvironment(previewEnvironment);
       if (isStandalone) {
         NEW.getLocalFlow(sessionId).then((state) => {
-          if (state) {
-            resumeSession(state);
-          }
+          // session data is resumed by ./ResumePage.tsx
```
builditben, line 108: I believe this is was the cause of the strange reconciliation behavior. The Question layout was resetting the state to the pre-reconciliation state. 

Instead, I think the ResumePage should be responsible for updating state (calling `resumeSession`) after reconciliation.

api.planx.uk/saveAndReturn/validateSession.ts
```diff
@@ -96,13 +96,6 @@ export async function validateSession(
     const { reconciledSessionData, alteredSectionIds } =
       await reconcileSessionData({ sessionData, alteredNodes });
 
-    // store reconciled session data
-    await updateLowcalSessionData({
-      sessionId,
-      sessionData: reconciledSessionData,
-      email,
-    });
```
builditben, line 104: I think session validation should be idempotent - it shouldn't change state and it should return the same response if called multiple times for the same account. 

If a user refreshes the resume page before continuing, they should see the same validation message and their session state should only be updated after they continue answering questions.

editor.planx.uk/src/pages/Preview/ResumePage.tsx
```diff
@@ -178,6 +178,7 @@ const ResumePage: React.FC = () => {
   }, [email]);
 
   const teamSlug = useStore((state) => state.teamSlug);
+  const resumeSession = useStore((state) => state.resumeSession);
```
builditben, line 181: I think `useStore` should only be called at the root of the component (see [React's rules of hooks](https://legacy.reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level))

api.planx.uk/saveAndReturn/validateSession.ts
jessicamcinchak: This makese a lot of sense :+1:

jessicamcinchak (approved):
Thanks for this fix, working smoothly for me!

PR #1705, opened 2023-05-18
chore: Tidy up e2e test context and add test annotations

builditben:
* Shares context defaults
    - adds a new planx testing logo
    - uses smoke testing email address (as recommended [here](https://docs.notifications.service.gov.uk/rest-api.html#smoke-testing))
    - reduces duplication
* Adds the "@integration" test annotation for test calling 3rd party services
* Adds commands for `test:ours` and `test:integrations` to run annotation sensitive tests

Whether we run every e2e test on every PR change or only a subset is still up for discussion but this PR provides a methodology for setting-up selective test runs.

### DIFF
--- a/e2e/src/context.ts
+++ b/e2e/src/context.ts
@@ -28,6 +28,23 @@ export interface Context {
   sessionIds?: string[];
 }
 
+export const contextDefaults = {
+  user: {
+    firstName: "Test",
+    lastName: "Test",
+    // Gov.uk Notify requests testing service use smoke test email addresses
+    // see https://docs.notifications.service.gov.uk/rest-api.html#smoke-testing
+    email: "simulate-delivered@notifications.service.gov.uk",
+  },
+  team: {
+    name: "E2E Test Team",
+    slug: "e2e-test-team",
+    logo: "https://raw.githubusercontent.com/theopensystemslab/planx-team-logos/main/planx-testing.svg",
+    primaryColor: "#444444",
+    homepage: "planx.uk",
+  },
+};
+
 export async function setUpTestContext(
   initialContext: Context
 ): Promise<Context> {
--- a/e2e/src/create-flow.spec.ts
+++ b/e2e/src/create-flow.spec.ts
@@ -1,22 +1,15 @@
 import { test, expect, Browser } from "@playwright/test";
-import { setUpTestContext, tearDownTestContext } from "./context";
+import {
+  contextDefaults,
+  setUpTestContext,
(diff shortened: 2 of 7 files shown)


e2e/src/context.ts
```diff
@@ -28,6 +28,23 @@ export interface Context {
   sessionIds?: string[];
 }
 
+export const contextDefaults = {
+  user: {
+    firstName: "Test",
+    lastName: "Test",
+    // Gov.uk Notify requests testing service use smoke test email addresses
+    // see https://docs.notifications.service.gov.uk/rest-api.html#smoke-testing
+    email: "simulate-delivered@notifications.service.gov.uk",
```
DafyddLlyr, line 37: I noticed this this morning looking into the email submission audit table - good to have this in place, should have spotted this sooner sorry!

Screenshot 2023-05-23 at 10 40 24

e2e/src/create-flow.spec.ts
```diff
@@ -1,22 +1,15 @@
...
-      email: "e2etest@test.com",
-    },
-    team: {
-      name: "E2E Navigation Test",
-      slug: "e2e-navigation-test",
-      logo: "https://placedog.net/250/250",
-      primaryColor: "#000000",
-      homepage: "example.com",
-    },
+    ...contextDefaults,
```
DafyddLlyr, line 12: 👍

e2e/src/pay.spec.ts
```diff
@@ -6,21 +6,18 @@ import { gql, GraphQLClient } from "graphql-request";
...
-  },
+  ...contextDefaults,
   team: {
     name: "Buckinghamshire",
     slug: "buckinghamshire",
-    logo: "https://placedog.net/250/250",
+    logo: "https://raw.githubusercontent.com/theopensystemslab/planx-team-logos/main/planx-testing.svg",
     primaryColor: "#F30415",
     homepage: "example.com",
   },
```
DafyddLlyr, line 23: Are we relying on Bucks to pass the `isSupported` check in `api.planx.uk/pay/pay.ts` and for a valid API key?

If we could have all tests running again an independent team it could help with DX not always having to destroy / recreate test containers if all content was encapsulated within a single team.

DafyddLlyr (approved):
Some quality improvements here - thank you!

e2e/src/pay.spec.ts
builditben: Exactly, a test team with credentials would be a nice improvement here.

PR #1708, opened 2023-05-21
feat: send to email using only a sessionId

jessicamcinchak:
Builds on #1695

Allows us to send to email & download application files using only a sessionId, no longer relying on formatting the CSV data on the frontend and passing it through the payload or storing on the `lowcal_sessions.data` object. 

To test:
- confirm you can submit a flow via email successfully!
  - update `teams.submission_email` for a team
  - create & publish a flow with "Send to email" selected
  - event should only have a `sessionId` in the payload & succeed
  - audit table should have a sucess record
  - you can click the download link in the email, all files are in zip as expected


builditben (approved):
Works perfectly :+1:

PR #1709, opened 2023-05-21
feat: create send events when a payment request is marked as `paid_at`

jessicamcinchak:
Submits Invite to Pay applications when the `payment_request` is marked as `paid_at`. Relies on each Send component destination being able to send based on solely a `sessionId`. 

To test:
- go through a published flow with invite to pay feature flag enabled & toggled on in the Pay component
  - when you reach pay, click "invite someone else" to generate a payment request
  - pay for your application, confirm your `payment_request` now shows `paid_at`
- check that Hasura successfully "setup-payment-send-events" (1 event record independent of number of Send destinations): https://hasura.1709.planx.pizza/console/events/data/setup_payment_send_events/processed
- check that "One-off scheduled events" succeeded (1 event record _per_ Send destination): https://hasura.1709.planx.pizza/console/events/one-off-scheduled-events/processed


api.planx.uk/inviteToPay/createPaymentSendEvents.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    // TODO Get this session's published flow data, find it's Send component, determine which "destinations" we need to queue up events for
+    const session = await $admin.getSessionById(req.params.sessionId);
+    const publishedFlowData = await getMostRecentPublishedFlow(session.flowId);
+    const destinations: Destination[] = [];
+
+    // TODO Determine which local authority it should send to based on team & destination    
+
+    const eventPayload = { sessionId: req.params.sessionId };
+    if ("bops" in destinations) {
```
builditben, line 29: I think this should be:

```suggestion
    if (destinations.includes("bops")) {
```
and similar for other `if` statements below.

The [`in` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in) looks for properties in an object rather than elements in an array (well it works on the index of the array rather than the value).

Sorry if this is actually just pseudo-code while this is still in draft. I was just being a little nosy :nose:

api.planx.uk/inviteToPay/createPaymentSendEvents.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// Create "One-off Scheduled Events" in Hasura when a payment request is paid
+const createPaymentSendEvents = async (req: Request, res: Response, next: NextFunction): 
+  Promise<NextFunction | Response | void> => {
+  try {
+    const now = new Date();
+    const combinedResponse: CombinedResponse = {};
+
+    // TODO Get this session's published flow data, find it's Send component, determine which "destinations" we need to queue up events for
```
DafyddLlyr, line 21: This feels more and more to me like a valuable piece of data that's worth exposing on "publish" of a flow maybe?

api.planx.uk/inviteToPay/createPaymentSendEvents.ts
```diff
@@ -0,0 +1,68 @@
...
+    const session = await $admin.getSessionById(req.params.sessionId);
+    const publishedFlowData = await getMostRecentPublishedFlow(session.flowId);
+    const destinations: Destination[] = [];
+
+    // TODO Determine which local authority it should send to based on team & destination    
+
+    const eventPayload = { sessionId: req.params.sessionId };
+    if ("bops" in destinations) {
+      const bopsEvent = await createScheduledEvent({
+        webhook: `{{HASURA_PLANX_API_URL}}/bops/${req.body.bops.localAuthority}`,
```
DafyddLlyr, line 31: Not a strong opinion, but there might be scope here for simplification here by dropping the payload and just having a route of  something like `{{HASURA_PLANX_API_URL}}/submit/${sessionId}/bops/${localAuthority}`

api.planx.uk/inviteToPay/createPaymentSendEvents.ts
jessicamcinchak: you're right! this is just pseudo code still to stub out general logic, will update and properly run/test soon :sweat_smile:

api.planx.uk/inviteToPay/createPaymentSendEvents.ts
jessicamcinchak: Yes! I agree storing this on the table makes sense. We can populate/update on the publishing step? 

Was thinking about this and also realised another on-publish validation step we'll want: currently, when Invite to Pay is enabled via the Pay component, we check for the existence of a Send component. But, I think we actually need to check for _exactly one_ Send component. 

Otherwise, the flow could have many Send components with different destinations selected each on branches. This is fine if it's a non-Invite to Pay flow where we know all applicants stay in the graph, but I don't think we want to have to calculate which branched Send/destinations the Invite to Pay applicant _should_ have reached when queuing up these payment request events (eg API has no knowledge of upcoming node ids). 

It doesn't seem like more than one Send is a pattern that shows up in flow content anyways right now so hopefully won't ever actually be an issue, but adding a validation step will make sure we know if/when that changes!

builditben: `findNextNodeOfType` in [planx-core](https://github.com/theopensystemslab/planx-core/blob/main/src/models/session/logic.ts#L104) is able to find the right Pay component using the flow and breadcrumbs. It should work for Send too.

jessicamcinchak: In this case, there won't be a Send component in my `breadcrumbs` though - so I wrote a simple `.find()` instead that just iterates over a flattened flow looking for a match on component type & extracts it's `destinations` from node data.

I've also updated the invite to pay validation rules to be a stricter: rather than just checking for the existance of any Send component(s), I'm now checking that there's exactly one. That'll save us having to collate possibly many Send components that each set different "destinations" like explained above. I don't think this will invalidate any existing flow content.

api.planx.uk/inviteToPay/createPaymentSendEvents.ts
```diff
@@ -0,0 +1,122 @@
...
+    const publishedFlowData = await getMostRecentPublishedFlow(session.flowId);
+    if (!session || !publishedFlowData) {
+      return next({
+        status: 400,
+        message: `Cannot fetch session or flow data to create payment send events`,
+      });
+    }
+
+    // Find this sessions Send component, determine which "destinations" we need to queue up events for
+    const sendNode: [string, Node] | undefined = Object.entries(publishedFlowData).find(([_nodeId, nodeData]) => nodeData.type === ComponentType.Send);
```
jessicamcinchak, line 47: Question: do we care that after these events successfully run, we'll have a submitted `lowcal_session` that is marked as `submitted_at` but does not have a `Send` component in its' breadcrumb data? 

My instinct is this is exactly the _expected_ behavior for a session with an associated `payment_request` because it has "left" the graph at the Pay node, but want to checkin/confirm that feels right to others too. If not, we could in theory add this node to the `lowcal_session.data.breadcrumbs` at this stage as well with `auto: true` or similar.

builditben: That's interesting. I've imagined that breadcrumbs are a temporary data structure used to construct something more canonical (the submission itself, and perhaps the passport as the distillation of the session data). The flows can change (perhaps invalidating breadcrumbs) but the output data (exported submission data/passport) is more stable and that's what we want to persist, ideally in a registry somewhere external, as the canonical submission. The breadcrumbs would ideally be disposable once the submission is complete.

api.planx.uk/inviteToPay/createPaymentSendEvents.ts
builditben: I think `findNextNodeOfType` does what you wanted to do - it looks for the next component of the type on the flow (naively at the moment, but it should evolve to handling the edge cases of traversing edges properly soon). It's using breadcrumbs to find what's upcoming but it's checking the flow to find the upcoming nodes.
Since it's not robust enough to handle every edge case yet, restricting editor functionality for the time-being makes sense but I'd be keen to come back to remove these kinds of limitations where possible.

api.planx.uk/inviteToPay/createPaymentSendEvents.ts
jessicamcinchak: Yes - on the same page here - but I think it's a v2 enhancement if that's alright! I'd actually like to get to the point where each submission endpoint _only_ requires the `:sessionId` param and `:localAuthority` logic is moved to each respective submission domain endpoint rather than at this outer level (eg `createPaymentSendEvents` and in frontend `Send` component model) like now.

DafyddLlyr: Totally! No issues here 😄

jessicamcinchak:
Seeing this CORS issue when testing on this pizza when attempting to generate the payment request / email the nominee. Any ideas? I can't re-create on .dev or pizza 1721. Could it be a strange Vultr quirk / can we assume it will be fine when merged?
Screenshot from 2023-05-24 15-45-46

If I manually create a `payment_request` record and mark as `paid_at` on the pizza - all events work as expected. No issues in the full end-to-end test locally.

DafyddLlyr (approved):
Tested by creating a payment request in Hasura, and then populating it's `paid_at` column.

Events created as expected! Submissions failed as there wasn't correct dummy data in the test session I used 👍

PR #1710, opened 2023-05-22
chore: fix mocks

ghost:
@jessicamcinchak, this is a fix for the mocks on your branch.

~~In this case, the `_admin` client needs mocking rather than `planx-core`. Hope that helps :smile:~~

Actually, I think this was just because of a nested callback that didn't need to be nested.

This isn't a super-tidy solution but it should highlight how to get tests passing on your branch just now.

### DIFF
--- a/api.planx.uk/send/uniform.test.ts
+++ b/api.planx.uk/send/uniform.test.ts
@@ -52,7 +52,16 @@ jest.mock("../client", () => {
       getDocumentTemplateNamesForSession: jest
         .fn()
         .mockResolvedValue(["X", "Y"]),
-      getSessionById: jest.fn(() => mockLowcalSession),
+      getSessionById: jest.fn((id: string) => {
+        if (id === "noGeoJSON") {
+          const mockLowcalSessionCopy: any = { ...mockLowcalSession };
+          delete mockLowcalSessionCopy.data.passport.data[
+            "property.boundary.site"
+          ];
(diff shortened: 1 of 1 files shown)


api.planx.uk/send/uniform.test.ts
```diff
@@ -33,20 +33,42 @@ jest.mock("@opensystemslab/planx-document-templates", () => {
...
+        }
+        return mockLowcalSession;
+      }),
+      generateCSVData: jest
+        .fn()
+        .mockResolvedValue([
+          { question: "Test", responses: [{ value: "Answer" }], metadata: {} },
+        ]),
+    },
+  };
```
jessicamcinchak, line 71: :bulb: super helpful to see/understand the `client` mocked like this, thank you for writing out !

api.planx.uk/send/uniform.test.ts
```diff
@@ -125,7 +147,7 @@ describe("createUniformSubmissionZip", () => {
 
   test("geojson and location plan is excluded when not present", async () => {
     const payload = {
-      sessionId: "1234",
+      sessionId: "noGeoJSON",
```
jessicamcinchak, line 150: thanks for spotting this wrong sessionId too

jessicamcinchak (approved):
Appreciate this!! Makes more sense now, I'll go ahead and merge this into my branch :+1:

PR #1711, opened 2023-05-22
chore: Insert production data with optional truncation on sync

builditben:
Rather than clobbering the db and then re-inserting data, I have updated the seed scripts to perform upserts for flows, users and teams.

What this means is that local changes are persisted and new additions are not removed on staging sync or after running `pnpm docker:add-seeds`. This means pizzas and local data can be preserved between syncs. Data can be reset by deleting the record and re-syncing.

This PR should enable send-to-email to work on staging without testing data being overridden (i.e. configured `teams.submission_email` values in staging should be retained after a sync).

**Testing**

This is best tested locally running `pnpm docker:add-seeds` to trigger the sync script:
 - Insert a new team, user and/or flow
 - Amend an existing team, user and/or flow
 - Delete an existing team, user and/or flow
 - Run `pnpm docker:add-seeds` or `PG_USERNAME=${} PG_PASSWORD=${} PG_DATABASE=${} ./scripts/seed-database/sync.sh`
 - Ensure that the new and amended records have persisted
 - Ensure that the deleted record has been restored


scripts/seed-database/container.sh
```diff
@@ -1,26 +1,29 @@
...
+  echo ${table} downloaded
+done
 
-# fetch flows
-psql --command="\\COPY (SELECT * FROM flows) TO '/tmp/flows.csv' (FORMAT CSV, DELIMITER ';')" "${REMOTE_PG}"
+echo beginning write transaction
+psql --quiet ${LOCAL_PG} -f write.sql
 
-# fetch published_flows (the last two)
-psql --command="\\COPY (SELECT id, data, flow_id, summary, publisher_id FROM (SELECT id, data, flow_id, summary, publisher_id, ROW_NUMBER() OVER (PARTITION BY flow_id ORDER BY created_at DESC) as row_num FROM published_flows) as subquery WHERE row_num <= 2) TO '/tmp/published_flows.csv' (FORMAT CSV);" "${REMOTE_PG}"
```
builditben, line 23: Published flows take a long time to download and I'm not certain it's particularly valuable.

The practice I've been following is to ensure a flow is published before testing. If we continue to do this, syncing published flows should be something we can avoid. Is that right?

builditben:
[Allow 'send to email' to work on staging / test flows](https://trello.com/c/THwzD3nq/2450-allow-send-to-email-to-work-on-staging-test-flows)

scripts/seed-database/container.sh
jessicamcinchak: My understanding of why we want to pull down the most recent published flow per flow is that we'd like to be able to share stable `/preview` flow links to testers (eg user research team) who do not have editor logins. If these flows aren't published, then key features like save & return can behave unexpectedly.

scripts/seed-database/container.sh
```diff
@@ -1,26 +1,29 @@
...
-# fetch users
-psql --command="\\COPY (SELECT * FROM users) TO '/tmp/users.csv' (FORMAT CSV, DELIMITER ';')" "${REMOTE_PG}"
+tables=(flows users teams flow_document_templates)
 
-# fetch teams
-psql --command="\\COPY (SELECT id, name, slug, theme, settings, notify_personalisation FROM teams) TO '/tmp/teams.csv' (FORMAT CSV, DELIMITER ';')" "${REMOTE_PG}"
+# run copy commands on remote  db
+for table in "${tables[@]}"; do
+  target=/tmp/${table}.csv
+  cmd="\\copy (SELECT * FROM ${table}) TO ${target} WITH (FORMAT csv, DELIMITER ';');"
```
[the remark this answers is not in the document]
jessicamcinchak: I believe it's intentional to not copy `teams.domain` - as that would result in incorrect magic links being sent in emails from staging !

jessicamcinchak (commented):
This looks good - left a few answers to your questions! 

I agree `upsert` is the perfect solution to our `teams.submission_email` bug, but I'm not actually convinced other tables like `flows` should `upsert` instead of drop/re-create as before. For example, if I make one edit to an LDC flow on staging as an experiment, my change will persist and the flow content will fall out of sync week after week? 

We've definitely communicated to editors that flow content is overwritten/fully synced from production to staging each Sunday night independent of any edits they do in the meantime, and we've never gotten any complaints about this (if anything I think it's good to have room to temporarily experiment and know you don't have to cleanup after yourself?), so I think we should aim to maintain that behavior for `flows`? I don't have strong opinions about `users` one way or the other.

scripts/seed-database/container.sh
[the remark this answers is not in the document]
jessicamcinchak: I think we also want to continue to omit `teams.submission_email` from the `COPY` statement, else we'd end up with emails synced to local envs and pizzas again, right? We intentionally adjusted this after unintentionally spamming a few emails.

scripts/seed-database/write.sql
```diff
@@ -0,0 +1,135 @@
...
+
+  INSERT INTO teams (
+    id,
+    name,
+    slug,
+    theme,
+    settings,
+    notify_personalisation,
+    domain,
+    submission_email
```
jessicamcinchak, line 56: follow-on comment from above: I think we want to omit `domain` and `submission_email` from this list of fields!

scripts/seed-database/container.sh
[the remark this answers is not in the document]
builditben: Thanks. I've removed `team.domain` from the insert command.

package.json
```diff
@@ -3,9 +3,10 @@
     "docker:logs": "docker compose logs --tail 10 -f",
     "docker:up": "docker compose -f ./docker-compose.yml -f ./docker-compose.local.yml --profile mock-services up --quiet-pull --renew-anon-volumes -d",
     "docker:down": "docker compose -f ./docker-compose.yml -f ./docker-compose.local.yml --profile mock-services --profile seed down",
-    "docker:recreate": "docker compose -f ./docker-compose.yml -f ./docker-compose.local.yml --profile mock-services up -d --build --force-recreate",
+    "docker:recreate": "docker compose -f ./docker-compose.yml -f ./docker-compose.local.yml --profile mock-services up -d --build --force-recreate --remove-orphans",
     "docker:destroy": "docker compose -f ./docker-compose.yml -f ./docker-compose.local.yml down --remove-orphans -v",
     "docker:add-seeds": "docker compose --profile seed run seed-database",
+    "docker:clean-data": "docker compose --profile seed run seed-database reset_all",
```
builditben, line 9: New command to locally reset user, team and flow data and pull a fresh set from prod.

.github/workflows/sync-staging-db.yml
```diff
@@ -33,6 +33,6 @@ jobs:
         with:
           files: ".env, .env.staging, api.planx.uk/.env.test, hasura.planx.uk/.env.test"
           fail: true
-      - run: docker run --rm -v "$(pwd)/scripts/seed-database:/app" --workdir="/app" postgis/postgis:12-3.0-alpine "./container.sh" "${STAGING_PG_URL}" "${PRODUCTION_PG_URL_FOR_USER_GITHUB_ACTIONS}"
+      - run: docker run --rm -v "$(pwd)/scripts/seed-database:/app" --workdir="/app" postgis/postgis:12-3.0-alpine "./container.sh" "${STAGING_PG_URL}" "${PRODUCTION_PG_URL_FOR_USER_GITHUB_ACTIONS}" "reset_flows" "include_published_flows"
```
builditben, line 36: Staging continues to reset flows on every sync. 
Only the latest published flow is synced (previously synced the last 2).
Users and teams are inserted where the records do not conflict.

scripts/seed-database/container.sh
builditben: Thanks. I've added the explicit option to include this table and that will still be part of the staging sync workflow.

DafyddLlyr (approved):

PR #1712, opened 2023-05-22
feat: add GIS metadata from planning.data.gov.uk to constraints API

jessicamcinchak:
Fetches metadata using the `Get dataset` endpoint from [planning.data.gov.uk](https://www.planning.data.gov.uk/docs) and appends it to our overall planning constraints response. This enables re-styling the component to be grouped by category and show additional information when expanded.

Splitting off the planning constraints API changes into their own PR for a couple of reasons:
- Appending `metadata`  & `category` properties to the `/gis` endpoint is only additive, it doesn't break the current public display of the component
- Once merged & rebased, this will simplify the original branch (PR #1296) to only include frontend file changes so it's easier for @ianjon3s to come in and share design work of the new accordions
- With this new API structure in place in the background, we can already start to think ahead to this ticket & followup from our conservation with BOPS last Friday https://trello.com/c/6vTXkdNR/1547-send-additional-a4-information-to-bops

### DIFF
--- a/api.planx.uk/gis/classifiedRoads.ts
+++ b/api.planx.uk/gis/classifiedRoads.ts
@@ -34,6 +34,7 @@ type PlanningConstraintBody = {
   value: boolean;
   text: string;
   data?: OSFeatures["features"];
+  category?: string;
 };
 
 // Passport key comes from Digital Planning Schemas googlesheet
@@ -91,6 +92,7 @@ export const classifiedRoadsSearch = async (
           value: true,
           text: `is on a Classified Road (${features[0].properties["RoadName1"]} - ${features[0].properties["RoadClassification"]})`,
           data: features,
+          category: "General policy",
         }
       } as PlanningConstraintResponse)
     } else {
@@ -98,6 +100,7 @@ export const classifiedRoadsSearch = async (
         [PASSPORT_FN]: {
           value: false,
           text: "is not on a Classified Road",
+          category: "General policy",
         }
       } as PlanningConstraintResponse)
     }
--- a/api.planx.uk/gis/digitalLand.js
+++ b/api.planx.uk/gis/digitalLand.js
@@ -98,6 +98,7 @@ async function go(localAuthority, geom, extras) {
           value: true,
           text: baseSchema[key].pos,
           data: [omitGeometry(entity)],
+          category: baseSchema[key].category,
         };
       }
     });
@@ -111,7 +112,11 @@ async function go(localAuthority, geom, extras) {
       baseSchema[key]["active"] && !Object.keys(formattedResult).includes(key)
   );
   nots.forEach((not) => {
-    formattedResult[not] = { value: false, text: baseSchema[not].neg };
+    formattedResult[not] = { 
+      value: false, 
+      text: baseSchema[not].neg, 
+      category: baseSchema[not].category, 
+    };
   });
 
   // --- DESIGNATED LAND ---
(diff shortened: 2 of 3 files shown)


api.planx.uk/gis/digitalLand.js
```diff
@@ -185,7 +190,23 @@ async function go(localAuthority, geom, extras) {
...
+      .catch(error => console.log(error))
+  )).then((responses) => {
+    responses.forEach((response) => {
+      // get the planx variable that corresponds to this 'dataset', should never be null because we only requested known datasets
+      const key = Object.keys(baseSchema).find((key) => baseSchema[key]["digital-land-datasets"].includes(response.dataset));
+      metadata[key] = response;
+    });
+  }).catch(error => console.log(error));
+
+  return { url: url, constraints: formattedResult, metadata: metadata };
```
builditben, line 209: Extracting this new code out into its own function would be nice but that could also be a task for a full refactor of this large function.

api.planx.uk/gis/classifiedRoads.ts
```diff
@@ -34,6 +34,7 @@ type PlanningConstraintBody = {
   value: boolean;
   text: string;
   data?: OSFeatures["features"];
+  category?: string;
```
builditben, line 37: Are categories actually an enum? 

Since the plan sounds like it would be to use these strings as semantic tags in the constraints component, then I think a string enum or constant object would be the way to go since it describes the range of possibilities and allows for exhaustive checking. Types can be great as documentation for things like this.

builditben (commented):
Code-only review: Looks good. I've made a few optional suggestions.

api.planx.uk/gis/digitalLand.js
jessicamcinchak: Yes, good shout - this file is definitely a near future candiate for js > ts conversion and breaking down into smaller methods. I'll come back to in a followup for the sake of getting the frontend PR ready for Ian first.

api.planx.uk/gis/classifiedRoads.ts
jessicamcinchak: I _think_ so, but actually need to confirm this with content team! Similar to above, will address when the main planning constraints API file migrates to typescript (it'd be a bit of a false enum if only referenced in the "roads" file here because the roads response should always be the _same_ category).

api.planx.uk/gis/digitalLand.js
```diff
@@ -185,7 +190,23 @@ async function go(localAuthority, geom, extras) {
     }
   }
 
-  return { url: url, constraints: formattedResult };
+  // --- METADATA ---
+  // additionally fetch metadata from Digital Land's "dataset" endpoint for extra context
+  let metadata = {};
+  const urls = activeDatasets.map((dataset) => `https://www.planning.data.gov.uk/dataset/${dataset}.json`);
+  await Promise.all(urls.map(url =>
```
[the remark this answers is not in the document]
jessicamcinchak: Two thoughts: 
- yes, all requests could be wrapped in the `Promise.all()` to run concurrently, but actually, we probably shouldn't proceed with fetching metadata at all if our constraints request throws an error? I'm not doing this currenty, but will adjust
- caching makes sense here (although unfortunately when profiling this component, I think `/roads` is still by far the slowest culprit because of the USRN string matching even with 10+ new metadata requests in play). is it okay if i similarly punt introducing this dependency until I convert to ts? planning to pick that up this week still once Ian is set on styling front

PR #1713, opened 2023-05-22
fix: Support 'test test' BOPS submissions from prod to staging

DafyddLlyr:
## Problem
Production payloads sent using "test test" are not supported on the staging API currently. As `payload.sessionId` is undefined on production payloads, the helper function `checkBOPSAuditTable()` returns a positive response, and the application is _not_ submitted to BOPS.

## How was this spotted?
Sam contacted me as a Southwark test failed earlier today. Logs show that the event had the response "Skipping send, already successfully submitted" and the Hasura query log on AWS showed that `""` was used as a sessionId for the query.

## Solution
This is pretty quick / dirty - it might just be simpler to go for a prod deploy and close this PR first thing tomorrow if there's no reason not to do this. Sam and Freya have a demo tomorrow so I'm opening this PR to hedge our bets a little on the quickest / most practical solution here.

I'm pretty sure we've been lucky so far with API changes to not hit an issue with "test test" submissions so it's good to keep this in mind for Submission API work going forward 👌

### DIFF
--- a/api.planx.uk/send/bops.ts
+++ b/api.planx.uk/send/bops.ts
@@ -10,6 +10,9 @@ import { _admin } from "../client";
 interface SendToBOPSRequest {
   payload: { 
     sessionId: string; 
+    planx_debug_data?: {
+      session_id: string
+    };
   }
 }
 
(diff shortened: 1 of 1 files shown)


jessicamcinchak (approved):
Thanks for picking this one up & apologies I forgot to consider this scenario when this merged to `main`! 

I don't think there's any reason we can't just go for a prod deploy here first? BOPS applications from staging to staging are working smoothly. If there's other PRs that shouldn't go up yet though, no issues with this merging and then coming back to cleanup/revert later.

DafyddLlyr:
Closed as this is a bit more complex than first thought - to be discussed at dev call tomorrow.

PR #1714, opened 2023-05-22
test(e2e): Invite to Pay

DafyddLlyr:
## What does this PR do?
 - Adds basic E2E tests in Playwright for both Agent and Nominee "invite to pay journeys"
 - Some tests are still marked as `fixme` - rather than expand this large PR I intend to branch off and fix these as their own issues


e2e/src/context.ts
```diff
@@ -26,6 +27,7 @@ export interface Context {
     data?: object;
   };
   sessionIds?: string[];
+  paymentRequest?: Partial<PaymentRequest>;
```
builditben, line 30: nit: This isn't wrong but it doesn't communicate what the interface expects. My preference is for types as documentation that state the expected structure explicitly.

e2e/src/helpers.ts
```diff
@@ -242,3 +252,141 @@ export async function expectSections({
...
+  await page.getByLabel("Town or city").fill("Test");
+  await page.getByLabel("Postcode").fill("HP111BB");
+  await page
+    .getByLabel("Email")
+    .fill("simulate-delivered@notifications.service.gov.uk");
+  await page.locator("button#submit-card-details").click();
+}
+
+export async function answerFindProperty(page: Page) {
+  await setupOSMockResponse(page);
```
builditben, line 290: Nice to see our fist API mock here :+1:

e2e/src/invite-to-pay/agent.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+  makePaymentRequest,
+  navigateToPayComponent,
+} from "./helpers";
+import { mockPaymentRequest, modifiedInviteToPayFlow } from "./mocks";
+
+let context: Context = {
+  user: {
+    firstName: "test",
+    lastName: "test",
+    email: "e2epaytest@opensystemslab.com",
```
builditben, line 28: Should we be using `simulate-delivered@notifications.service.gov.uk` here?

e2e/src/invite-to-pay/agent.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+    expect(toggleInviteToPayButton).toBeVisible();
+    await toggleInviteToPayButton.click();
+    const inviteToPayFormHeader = await page.getByText(
+      "Invite someone else to pay for this application"
+    );
+    expect(inviteToPayFormHeader).toBeVisible();
+
+    await answerInviteToPayForm(page);
+    await page.getByText("Send invitation to pay").click();
+    await page.waitForLoadState("networkidle");
```
builditben, line 79: nit: It's worth noting that waiting for `networkidle` is discouraged in the Playwright docs: https://playwright.dev/docs/api/class-page#page-wait-for-load-state
I don't think this needs to change now but we should review an approach to notifying event completion (listening for debug logs isn't ideal either but that's my best stab at this so far).

e2e/src/invite-to-pay/nominee.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+import { mockPaymentRequest, mockSessionData } from "./mocks";
+import { GraphQLClient, gql } from "graphql-request";
+import { PaymentRequest, Session } from "@opensystemslab/planx-core/types";
+import { getPaymentRequestBySessionId } from "./helpers";
+
+let context: Context = {
+  user: {
+    firstName: "test",
+    lastName: "test",
+    email: "e2epaytest@opensystemslab.com",
```
builditben, line 20: again, I think we want to use mock emails here.

builditben (approved):
Looks good so far. Thanks for your work on this :+1:

e2e/src/context.ts
DafyddLlyr: No this is totally fair - it's actually ended up as not used, so this has been tidied up.

PR #1716, opened 2023-05-23
fix: Correct type for sessionId in CreateEmailApplication

DafyddLlyr:


builditben (commented):
Should these mutation arguments all be required without default empty strings? Are the defaults redundant?

builditben (approved):

PR #1726, opened 2023-05-26
chore: Retire 'test test' workaround

DafyddLlyr:
## What does this PR do?
 - Removes `useStagingUrlIfTestApplication()` and associated tests
 - Adds form validation to stop "Test Test" being a valid submission name
 - Bumps data sync frequency to nightly
 
 image


editor.planx.uk/src/@planx/components/ContactInput/model.ts
```diff
@@ -23,6 +23,22 @@ export const userDataSchema: SchemaOf<Contact> = object({
...
+  name: "Test Test is not used for applications",
+  test: ({ firstName, lastName }, context) => {
+    const isValid =
+      [firstName, lastName]
+        .map((x) => String(x).toLowerCase().trim())
+        .join("|") !== "test|test";
+
+    if (isValid) return true;
+
+    return context.createError({
```
DafyddLlyr, line 36: Due to some Formik / Yup integration issues, we need to manually throw an error here instead of just returning false and including a `message` in the object. A good number of issues on the Formik repo related to this, solution was found here - https://github.com/jaredpalmer/formik/issues/2146#issuecomment-720639988

DafyddLlyr:
[Retire "Test Test" and do user-testing on staging](https://trello.com/c/QnN3DJ7i/2471-retire-test-test-and-do-user-testing-on-staging)

PR #1728, opened 2023-05-26
fix: Do not display empty section titles

DafyddLlyr:
## What does this PR do?
- Fixes a bug where "empty" sections have their title displayed on the `Review` component

## How is this fixed?
- A section can be "empty" if there are either no components within it (this probably shouldn't happen) or no "presentational" components within in (components without a visual representation on the `Review` page such as auto-answered nodes, or the `Review` component itself)
- By pre-filtering, we can iterate over a list of valid breadcrumbs and only display those which meet the required conditions. - Previously the title and validation logic were in separate components, so these could not be controlled together

**"Check your application" sections header no longer displays on Pizza**
image


jessicamcinchak:
oh thanks for picking this one up and thinking through the proper code fix! i had also flagged this with the content team earlier this week too just fyi (because the "Check your answers" section is already in a more sensible place in prior approvals that doesn't lead to the empty section title, just not on LDCs!) https://trello.com/c/HfAbzK3G/1250-adjust-check-your-application-section-position-on-ldc-services-to-prevent-stray-review-page-header

editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
```diff
@@ -76,56 +76,119 @@ const components: {
...
-    <SummaryList
-      breadcrumbs={props.breadcrumbs}
-      flow={props.flow}
-      passport={props.passport}
-      changeAnswer={props.changeAnswer}
-      showChangeButton={props.showChangeButton}
-    />
-  );
-}
+    const isPresentationalComponent = Boolean(Component);
```
builditben, line 116: nit: This was helpful for understanding what `components` represents. Maybe a sign that `components` should actually be called `presentationalComponents`?

builditben (approved):
Looks good.

I've left a few minor comments, the main one being the need for a rational for writing e2e tests vs. unit tests. It's not a huge deal so happy to approve :+1:

editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
DafyddLlyr: Good shout - updated 👌

e2e/src/sections.spec.ts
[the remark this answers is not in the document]
DafyddLlyr: No this is totally fair - I was taking the quick / easy road here... 😬 

These have now been ported across to unit tests with a bit of setup of state around them. As a bonus there were some other section unit tests which were marked as `skip` which I've picked up here also.

PR #1730, opened 2023-05-30
Fix: Limit ITP to one Pay node

builditben:
Invite to Pay validation is timing-out and/or running out of memory and a patch for simple validation is currently in place (https://github.com/theopensystemslab/planx-core/pull/36)

This PR applies a different workaround which limits flow that use the ITP feature to only define one Pay node in that flow (corresponding planx-core update here: https://github.com/theopensystemslab/planx-core/pull/38).

Like the Send node restriction, this limit is temporary until we can efficiently traverse the flow graph to find the next node of the matching type.

### DIFF
--- a/api.planx.uk/editor/publish.test.ts
+++ b/api.planx.uk/editor/publish.test.ts
@@ -280,37 +280,6 @@ describe("invite to pay validation on diff", () => {
       });
   });
 
-  it("does not update if invite to pay is enabled, but there is no Pay component", async () => {
-    const invalidFlow = {
-      ...flowWithInviteToPay,
-      Pay: undefined,
-      _root: {
-        edges: flowWithInviteToPay._root.edges!.filter((id) => id !== "Pay"),
-      },
-    };
-
-    queryMock.mockQuery({
-      name: "GetFlowData",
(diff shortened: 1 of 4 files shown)


jessicamcinchak (approved):
Thanks! Confirmed I can successfully generate a payment request for a full LDC flow on this pizza.

Agree we should be aiming for these "exactly one" restrictions to be temporary, but could you please still add a quick unit test to cover this scenario in `publish.test.ts` so it stays in sync? Curently there's a test per rule except for this new addition.

api.planx.uk/editor/publish.test.ts
```diff
@@ -280,37 +280,6 @@ describe("invite to pay validation on diff", () => {
...
-      .set(authHeader())
-      .expect(200)
-      .then((res) => {
-        expect(res.body.message).toEqual("Cannot publish an invalid flow");
-        expect(res.body.description).toEqual(
-          "When using Invite to Pay, your flow must have a Pay"
-        );
-      });
-  });
-
```
jessicamcinchak: sorry seeing this late, but why did this existing test get removed?

PR #1735, opened 2023-05-31
refactor: Remove duplicated file upload logic

DafyddLlyr:
This turned out to be fairly straightforward following on from #1731 🤞 

So this is how it works...
 - In order for a component to have the ability to upload private files it needs all three of `Dropzone`, `FileStatus` and an array of `UploadedFileCards`
 - These are currently grouped within `PrivateFileUpload` in the layout currently used by both `DrawBoundary` and `FileUpload`
 - The `slots` are always controlled by the parent component of these three as validation logic will change from use case to use case and is best controlled at this level I believe.

I've added a commit [e9f6b02](https://github.com/theopensystemslab/planx-new/pull/1735/commits/e9f6b02c6358850849a40be16ee45f43d251957b) to very roughly demonstrate/test this with a mock `MultipleFileUpload` which has all 3 components working together in a totally different layout. Happy to drop this commit or leave as a rough skeleton for now but it is very much a proof of concept.

**Next steps...**
- There's probably a more sophisticated way of managing this disconnect between layout and components via the use of slots in React, but I think a simple/dumb approach here may be right for now until we have `MultipleFileUpload` slightly better understood. I'm not keen on implementing something to have to undo/change it very shortly if not required...!
 - We need a callback to open a modal when `onDrop()` is called. I'm sure this can be managed by an event.

**To test**
- Quick dummy flow here with all three upload components - https://1735.planx.pizza/testing/daf-upload-test


builditben (commented):
Thanks for explaining this change so clearly.

Question: for "Upload a location plan" should I be able to upload multiple files? If so, this seems to be broken at the moment. If not, should the option not be presented.

Other than this, testing looks good for the demo flow you shared :+1:

DafyddLlyr:
Sorry - just tested this a bit more, you're right - it's failing and throwing an error. Looking at this right now thanks.

jessicamcinchak (approved):
Slot validation all working as expected here and getting expected behavior when uploading, navigating backwards (files retained), etc :raised_hands: 

Only thing I still noticed is what Ben previously flagged - after uploading a location plan a second upload slot is still shown, even though disabled. Can that be easily hidden? Appreciate this behaves the same on prod though/wasn't introduced here so happy for this to merge to fix known bug and come back to!

DafyddLlyr:
> Only thing I still noticed is what Ben previously flagged - after uploading a location plan a second upload slot is still shown, even though disabled. Can that be easily hidden? Appreciate this behaves the same on prod though/wasn't introduced here so happy for this to merge to fix known bug and come back to!

Good shout - said yesterday I'd take a look at this - should be a quick fix 👌

builditben (approved):
That single file upload feels much more intuitive now. Thanks for the update :+1:

PR #1739, opened 2023-05-31
UPDATE: styles for radio + checkbox

ianjon3s:
PR adds styles for radio and checkbox inputs.

Requires the following feature flag to see radio buttons:
`window.featureFlags.toggle("ALT_THEME")`


ianjon3s:
@DafyddLlyr here are the styles for radio buttons and checkboxes as discussed.

The only thing that is missing is the logic to toggle the border of the radio card (with image / description) when selected (ImageRadio.tsx). I tried to replicate the logic used for the checkbox cards but was unable to do this.

e2e/src/helpers.ts
```diff
@@ -179,7 +179,7 @@ export async function findQuestion({
   page: Page;
   title: string;
 }): Promise<Locator> {
-  const group = await page.getByRole("group", { name: title });
+  const group = await page.getByRole("radiogroup", { name: title });
```
DafyddLlyr, line 182: Such a simple and pain free fix - thanks @builditben !

editor.planx.uk/src/@planx/components/Question/Public/Question.tsx
```diff
@@ -33,12 +43,116 @@ export enum QuestionLayout {
...
+                  formik.setFieldValue("selected.a", response.responseKey);
+                };
+                const buttonProps = {
+                  onChange,
+                };
 
-  return isUsingAltTheme ? (
-    <RadioQuestion {...props} />
-  ) : (
-    <ButtonQuestion {...props} />
```
DafyddLlyr, line 41: This file has a very noisy diff, but all that's happened here is that what was previously `RadioQuestion.tsx` is now just `Question.tsx` - there's no logical changes apart from the dropping of this ternary and the child becoming the parent.

editor.planx.uk/src/@planx/components/shared/Radio/DescriptionRadio.tsx
```diff
@@ -12,27 +13,45 @@ export interface Props {
...
-    <Box sx={{ paddingBottom: 2, display: "flex", alignItems: "center" }}>
-      <Radio value={id} onChange={onChange} />
-      <Typography>{title}</Typography>
-    </Box>
-    <Typography variant="body2">{description}</Typography>
-  </StyledFormLabel>
-);
+}) => {
+  const radioGroupState = useRadioGroup();
+  const isSelected = radioGroupState?.value === id;
```
DafyddLlyr, line 44: @ianjon3s (For future reference) This is how I resolved the issues you were facing with getting the selected radio in order to apply a "selected" style. Docs: https://mui.com/material-ui/react-radio-button/#useradiogroup

builditben:
I spotted on bug that I can't work out where the fix would be in the code.

The checkbox with image doesn't register as selected when the click is off-centre:

image

There's a div around the checkbox that would ideally have the css `pointer-events: none` to prevent this but I can't see where that might be applied in the code. Sorry I can't be more helpful finding the fix for this.

DafyddLlyr:
> The checkbox with image doesn't register as selected when the click is off-centre:

Good find - I can recreate this, let me take a look at it.

DafyddLlyr:
@builditben Not a perfect fix, but I believe the issue is now resolved if you're happy to quick cross-check and test again?

builditben (approved):
Sorry I missed your mention @DafyddLlyr.

This looks great, nice work both of you!

I think this is good but I still wanted to point out a couple of accessibility issues I hit:
 - Tab navigation for radio buttons only selected the first or the chosen option and I couldn't tab navigate through the other radio options
 - For some reason both checkbox and radio buttons couldn't be selected by pressing enter on the keyboard while highlighted (but spacebar worked??). The other buttons work with the enter key but I've no idea why this wouldn't also work for checkboxes and radio buttons.

I wonder if these issues are best resolved (if they are actually a problem) in a follow-up PR? I've approved because the work here looks good to me other than these minor issues.

ianjon3s:
Thanks @builditben,

Both of the points you raise are expected behaviour.

- Radio button groups are classed as a single element, as they submit only one value, and once tabbed to are navigated using the arrow keys (native HTML functionality). Checkboxes should be tabbed individually.
- The expected input to select both radio buttons and checkboxes is the spacebar, I believe this is because the enter key is reserved for form submission (this is what happens on the W3 examples https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_type_radio).

builditben:
Good to know, thanks @ianjon3s!

PR #1740, opened 2023-06-01
chore: Standardise import of `styled()` MUI utility function

DafyddLlyr:
Quick fix for an issue I've hit a few times and then taken way to long to work out the cause of the issue 😅 

Importing from "@mui/styles" (or elsewhere) does not allow us to use the `sx` prop, or to pass custom props into styled components easily.

Very good summary of the benefits of importing from "@mui/material/styles" here - https://mui.com/system/styled/#what-problems-does-it-solve


jessicamcinchak (approved):
had not appreciated this difference before, it's an important one!! thanks for straightening out

PR #1741, opened 2023-06-01
feat: add another reminder the day before expiry for sessions & payment requests

jessicamcinchak:
Changes / how this works:
- I've introduced `const REMINDER_DAYS_FROM_EXPIRY = [7, 1]` in `/saveAndReturn/utils`, the same place as we control the length of a session. It should make it easy to add or subtract new reminder intervals going forward.
- I loop through the reminder days array when the database trigger is called to `createScheduleEvents` to create one reminder event per day, tacking on the `reminderDays` number to the end of the event comment so they remain unique. 
- The methods to send emails and the Notify config/personalisation types per template don't change at all

Notify templates: 
- I realised that Invite to Pay reminder templates _only_ reference the `((expiryDate))` while the Save & Return reminder template has a hardcoded sentence "Your application is due to expire within the next 7 days". 
- Rather than introduce a new dynamic variable like `((reminderDays))`, I'd vote we simply only reference `((expiryDate))` in both templates and update [this template](https://www.notifications.service.gov.uk/services/012e65af-0eb2-45d5-87bd-4248354c4c22/templates/43be4c11-a406-4381-b2be-056a1127455d) to read: 

> You have until ((expiryDate)) to complete this application. Use the link to return to it when you are ready to continue. Your application will be deleted if you do not complete it by this date.

Does that seem reasonable?


api.planx.uk/webhooks/lowcalSessionEvents.ts
```diff
@@ -15,12 +15,14 @@ const createReminderEvent = async (req: Request, res: Response, next: NextFuncti
...
-      schedule_at: addDays(Date.parse(createdAt), (DAYS_UNTIL_EXPIRY - 7)),
-      payload: payload,
-      comment: `reminder_${payload.sessionId}`,
-    });
+    const response = await Promise.all(REMINDER_DAYS_FROM_EXPIRY.map((day: number) => (
+      createScheduledEvent({
+        webhook: "{{HASURA_PLANX_API_URL}}/send-email/reminder",
+        schedule_at: addDays(Date.parse(createdAt), (DAYS_UNTIL_EXPIRY - day)),
+        payload: payload,
+        comment: `reminder_${payload.sessionId}_${day}day`,
```
jessicamcinchak, line 23: As is, this comment will break our event cleanup procedures (eg PG functions `delete_lowcal_session_scheduled_events` and `delete_payment_request_scheduled_email_events`) which look for an exact string match on the comment with the prefix & sessionId only.

Would appreciate any opinions on possible approaches here - I see main options as: 
1. We don't care about unique comments on the events, but can determine which reminder it is (7 or 1 day) by it's scheduled date. I revert these changes to `comment` throughout, we don't have to replace existing PG functions.
2. We do care about unique comments on the events. We keep this structure, I `CREATE OR REPLACE` both PG functions to do fuzzy matching on comments using `reminder_sessionId%` so we can still easily add/subtract other reminder intervals in the future

DafyddLlyr (approved):
Super simple and elegant solution here - I love it 🥇

api.planx.uk/webhooks/lowcalSessionEvents.ts
jessicamcinchak: Yep, agree the comment is a really handy debugging context here :+1: Just wanted to sense check that direction before creating a migration

jessicamcinchak:
[Expiry email bug on non-LDC services?](https://trello.com/c/Jx7NM6wM/2380-expiry-email-bug-on-non-ldc-services)

jessicamcinchak:
confirmed that reminder & expire events are being properly cleaned up on this pizza on `submitted_at` with the most recent updated pg functions. going to merge to staging & move [this related ticket](https://trello.com/c/Jx7NM6wM/2380-expiry-email-bug-on-non-ldc-services) into "in review" - if staging is working, then would expect bug to be fixed when deployed to prod & we can do a one-time manual cleanup of lingering events for sessions that have been submitted.

PR #1743, opened 2023-06-01
tests: Add api e2e test suite and submission tests

builditben:
Goals of this PR:
 - to test submissions for the invite-to-pay journey
 - to allow e2e tests to run using the "e2e" test team
 - to create a pattern for setting up api driven end-to-end tests

corresponding change in planx-core: https://github.com/theopensystemslab/planx-core/pull/39


e2e/tests/api-driven/src/invite-to-pay/submissions.feature
DafyddLlyr, line 1: The separation here between behaviour and implementation is great to see - super clear to understand. It feels like a really solid way to describe business logic and rules.

I can't immediately think of a huge number of features that currently fall into this category that isn't user-driven (and could be tested by Playwright), but there are certainly some like cron-jobs (cleanup operations), future flow views, or other stored procedures  that could benefit from this. 

Have you thought about likely future candidates for other API-driven features to test?

DafyddLlyr (approved):

api.planx.uk/send/helpers.ts
```diff
@@ -156,13 +156,23 @@ async function insertPaymentStatus({
       status,
     }
   );
+  console.log(response);
```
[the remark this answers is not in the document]
builditben: I didn't spot this. thanks

e2e/tests/api-driven/src/invite-to-pay/submissions.feature
builditben: Exactly like you say, I think the main candidates are the bits of event processing logic and anywhere where the business logic is managed by passing data between systems (e.g.  webhooks, stored procedures, etc).  Clear articulation of what should happen in tests like these are valuable both as automated regression tests and as documentation of the intended behaviour.

DafyddLlyr (approved):

DafyddLlyr:
From dev call 07/06/23: We're going to run these tests daily as part of a cronjob, will hold off merging until this is in place.

builditben:
Depends on #1767

PR #1744, opened 2023-06-02
fix: Header link style

DafyddLlyr:
Fix for visual regression introduced in #1740

### DIFF
--- a/editor.planx.uk/src/components/Header.tsx
+++ b/editor.planx.uk/src/components/Header.tsx
@@ -50,10 +50,10 @@ const BreadcrumbsRoot = styled(Box)(() => ({
   fontSize: 20,
 }));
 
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/components/Header.tsx
```diff
@@ -50,10 +50,10 @@ const BreadcrumbsRoot = styled(Box)(() => ({
   fontSize: 20,
 }));
 
-const BreadcrumbLink = styled(ReactNaviLink)(() => ({
+const StyledLink = styled(ReactNaviLink)(() => ({
   color: "#fff",
   textDecoration: "none",
-}));
+})) as typeof Link;
```
DafyddLlyr, line 56: Casting like this is required whenever we're using the component={XXX} prop (see line 194).

jessicamcinchak (approved):

PR #1745, opened 2023-06-02
feat(wip): Multiple file upload form

DafyddLlyr:


editor.planx.uk/src/@planx/components/MultipleFileUpload/Editor.test.tsx
DafyddLlyr, line 1: I feel like a few additional tests here would have been nice (in particular a test of create conditional rule > change to simple rule > check additional properties `operator`, `val` and `fn` have been removed).

I didn't go for this as...
 - We don't have a great test setup worked out for the `FormModal` component yet
 - There isn't generally visual feedback on the form if something is invalid
 - It's basically going to boil down to "are we validating the schema on submit?" which is more easily testable in `schema.test.ts`

editor.planx.uk/src/@planx/components/MultipleFileUpload/Editor.tsx
```diff
@@ -71,3 +119,159 @@ function MultipleFileUpload(props: Props) {
...
+              setCondition(
+                Condition[e.target.value as keyof typeof Condition],
+                props.value
+              )
+            )
+          }
+        >
+          {Object.entries(Condition).map(([key, value]) => (
+            <MenuItem key={key} value={value}>
+              {upperFirst(lowerCase(value))}
```
jessicamcinchak, line 166: nit: does lodash `capitalize` accomplish the same here but in one method instead of two? https://lodash.com/docs/4.17.15#capitalize

jessicamcinchak (approved):
Super solid foundation, thanks for starting this :raised_hands:

PR #1749, opened 2023-06-06
feat: show global and per-file help text in MultipleFileUpload component

jessicamcinchak:
Conditionally renders the more information icon & sidebar for both the overall component and individual files which have help text defined in the editor.
Screenshot from 2023-06-06 13-17-33

### DIFF
--- a/editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
@@ -3,10 +3,19 @@ import Typography from "@mui/material/Typography";
 import { PublicProps } from "@planx/components/ui";
 import React, { useState } from "react";
 import { FONT_WEIGHT_BOLD } from "theme";
+import MoreInfoIcon from "ui/icons/MoreInfo";
+import ReactMarkdownOrHtml from "ui/ReactMarkdownOrHtml";
+import { emptyContent } from "ui/RichTextInput";
 
 import { FileUploadSlot } from "../FileUpload/Public";
+import { MoreInformation } from "../shared";
 import Card from "../shared/Preview/Card";
-import QuestionHeader from "../shared/Preview/QuestionHeader";
+import MoreInfo from "../shared/Preview/MoreInfo";
+import MoreInfoSection from "../shared/Preview/MoreInfoSection";
+import QuestionHeader, {
+  Image,
+  StyledIconButton,
+} from "../shared/Preview/QuestionHeader";
 import { Dropzone } from "../shared/PrivateFileUpload/Dropzone";
 import { FileStatus } from "../shared/PrivateFileUpload/FileStatus";
 import { UploadedFileCard } from "../shared/PrivateFileUpload/UploadedFileCard";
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
```diff
@@ -40,7 +56,10 @@ function Component(props: Props) {
               Required files
             </Typography>
             {props.fileTypes.map((fileType) => (
-              <p key={fileType.key}>{fileType.key}</p>
+              <InteractiveFileListItem
+                name={fileType.key}
```
jessicamcinchak, line 60: worth noting that React throws this error when trying to pass the full `fileType` as props to a child component: 
```
"key" is not a prop. Trying to access it will result in "undefined" being returned. If you need to access the same value within the Child component, you should pass it as a different value
```
we might want to consider renaming `key` to `name` in the `FileType` model too??

editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
```diff
@@ -24,7 +34,14 @@ function Component(props: Props) {
...
     <Card handleSubmit={props.handleSubmit} isValid>
-      <QuestionHeader title={props.title} description={props.description} />
+      <QuestionHeader
+        title={props.title}
+        description={props.description}
+        info={props.info}
+        policyRef={props.policyRef}
+        howMeasured={props.howMeasured}
+        definitionImg={props.definitionImg}
+      />
```
DafyddLlyr, line 44:
```suggestion
<QuestionHeader {...props}/>
```

I think this would work?

editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
DafyddLlyr: Ah this is interesting and a good catch - I'd certainly be in favour of just changing to `name` if we end up having to do an awkward half mapping of key -> name to work around this 👍

editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
```diff
@@ -63,3 +83,61 @@ function Component(props: Props) {
...
 }
+
+interface FileListItemProps {
+  name: string;
+  moreInformation?: MoreInformation;
+}
+
+const InteractiveFileListItem = (props: FileListItemProps) => {
+  const [open, setOpen] = React.useState(false);
+  const { trackHelpClick } = useAnalyticsTracking();
```
DafyddLlyr, line 94: This is a great shout and will probably be super useful 👏

editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
```diff
@@ -63,3 +83,61 @@ function Component(props: Props) {
...
+  const { info, policyRef, howMeasured, definitionImg } =
+    props.moreInformation || {};
+
+  const handleHelpClick = () => {
+    setOpen(true);
+    // TODO: track granularity of file name in analytics, currently only knows help was clicked for this overall component type/node id
```
DafyddLlyr, line 100: I've added this as a TODO on the Trello list 👍

editor.planx.uk/src/@planx/components/shared/Preview/QuestionHeader.tsx
```diff
@@ -28,13 +28,13 @@ const Description = styled(Box)(({ theme }) => ({
   },
 }));
 
-const StyledIconButton = styled(IconButton)(() => ({
+export const StyledIconButton = styled(IconButton)(() => ({
```
DafyddLlyr, line 31: nit: This might be best moving to the `editor.planx.uk/src/ui`folder (and having a `size` props?)

DafyddLlyr (approved):
Great stuff! 🎨

PR #1751, opened 2023-06-06
feat: adds basic validation and back/change behavior to MultipleFileUpload

jessicamcinchak:
Two changes here for some foundational functionality we can keep building on:
1. Adds a basic error wrapper around the file dropzone if you try to continue before uploading anything - and introduces a basic yup validation schema we can add "real" rules to next
Screenshot from 2023-06-06 16-19-41

2. Retains previously uploaded files when navigating fowards / back (& on "change" from Review page in future)
Screenshot from 2023-06-06 16-20-04


DafyddLlyr (approved):
This is great 👍 😄

PR #1752, opened 2023-06-06
chore: Bump up API timeout

DafyddLlyr:
An attempt at an immediate fix for recent submission timeouts. This value is set to 60s by default, bumping it to 120s may give us a bit of breathing space to get to the root of the problem.

It's likely that fetching and processing the entire flow on the API is taking longer than it did on the frontend where we already had the flow upfront.

### DIFF
--- a/infrastructure/application/index.ts
+++ b/infrastructure/application/index.ts
@@ -260,6 +260,7 @@ export = async () => {
     external: true,
     vpc,
     subnets: networking.requireOutput("publicSubnetIds"),
+    idleTimeout: 120,
   });
   // XXX: If you change the port, you'll have to make the security group accept incoming connections on the new port
   const API_PORT = 80;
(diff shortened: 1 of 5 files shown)


infrastructure/application/index.ts
```diff
@@ -260,6 +260,7 @@ export = async () => {
     external: true,
     vpc,
     subnets: networking.requireOutput("publicSubnetIds"),
+    idleTimeout: 120,
```
jessicamcinchak, line 263: question: since this error is quite specific so far, should we maybe think about route-specific timeout bumps instead of our overall server? 

admittedly not super clear on when to update config in pulumi versus express though - but was thinking something like `server.setTimeout(60 * 1000)` overall then `req.setTimeout(120 * 1000)` directly in applicable endpoints? i'd guess that this `idleTimeout` still needs to match our highest possible route timeout.

appreciate this is just intended to be a quick interim fix though so trust your judgement here, thanks for opening something today :+1:

jessicamcinchak (approved):

infrastructure/application/index.ts
DafyddLlyr: No this is a good shout! 👌 

You're right that this would need to match our highest value so that sounds like a much more sensible and safe approach - I'll try and implement that just now.

PR #1753, opened 2023-06-06
feat: Create `FileList` from passport and parsed `FileType` rules

DafyddLlyr:
# What does this PR do?
 - Build a `FileList` object based on the current passport and the `FileType` rules built by the Editor
 - This structure can (hopefully) do the job of driving the logic for the public facing component

## Comments
 - Not totally sold on a few names here - `FileList` feels pretty vague and generic - thoughts welcome!

### DIFF
--- a/editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
@@ -3,6 +3,7 @@ import { styled } from "@mui/material/styles";
 import Typography from "@mui/material/Typography";
 import { PublicProps } from "@planx/components/ui";
 import { useAnalyticsTracking } from "pages/FlowEditor/lib/analyticsProvider";
+import { useStore } from "pages/FlowEditor/lib/store";
 import React, { useEffect, useRef, useState } from "react";
 import { FONT_WEIGHT_BOLD } from "theme";
 import ErrorWrapper from "ui/ErrorWrapper";
@@ -24,11 +25,9 @@ import { Dropzone } from "../shared/PrivateFileUpload/Dropzone";
 import { FileStatus } from "../shared/PrivateFileUpload/FileStatus";
 import { UploadedFileCard } from "../shared/PrivateFileUpload/UploadedFileCard";
 import { getPreviouslySubmittedData, makeData } from "../shared/utils";
-import { MultipleFileUpload } from "./model";
+import { createFileList, FileList, MultipleFileUpload } from "./model";
 
 type Props = PublicProps<MultipleFileUpload>;
-
-export default Component;
 
 const DropzoneContainer = styled(Box)(({ theme }) => ({
   display: "grid",
@@ -106,6 +105,18 @@ function Component(props: Props) {
       setValidationError(undefined);
     }
   }, [slots]);
+  
+  const [fileList, setFileList] = useState<FileList>({
+    required: [],
+    recommended: [],
+    optional: [],
+  });
+
+  useEffect(() => {
+    const passport = useStore.getState().computePassport();
+    const fileList = createFileList({ passport, fileTypes: props.fileTypes });
+    setFileList(fileList);
+  }, []);
 
   return (
     <Card
@@ -114,6 +125,7 @@ function Component(props: Props) {
     >
       <QuestionHeader {...props} />
         <DropzoneContainer>
+          <FileStatus status={fileUploadStatus} />
           <ErrorWrapper error={validationError} id={props.id}>
           <Dropzone
             slots={slots}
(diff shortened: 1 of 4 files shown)


editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
```diff
@@ -10,17 +11,26 @@ import QuestionHeader from "../shared/Preview/QuestionHeader";
...
 function Component(props: Props) {
   const [slots, setSlots] = useState<FileUploadSlot[]>([]);
   const [fileUploadStatus, setFileUploadStatus] = useState<string | undefined>(
     undefined
   );
+  const [fileList, setFileList] = useState<FileList>({
+    required: [],
+    recommended: [],
+    optional: [],
+  });
```
DafyddLlyr, line 27: This is the data structure which _I think_ can drive this component...
 - It can display the list of files on right of the initial page
 - It can populate the modal
 - It can map a slot to a `FileType` instance for parsing on `handleSubmit()`

editor.planx.uk/src/@planx/components/MultipleFileUpload/model.ts
```diff
@@ -82,3 +86,82 @@ export const newFileType = (): FileType => ({
...
+  slot?: FileUploadSlot;
+}
+
+export interface FileList {
+  required: UserFiles[];
+  recommended: UserFiles[];
+  optional: UserFiles[];
+}
+
+export const createFileList = ({
```
DafyddLlyr, line 100: I'm not totally sold that this is the best place for this logic, but it's good for now!

editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
```diff
@@ -39,8 +49,20 @@ function Component(props: Props) {
...
+              Recommended files
+            </Typography>
+            {fileList.recommended.map((file) => (
+              <p key={file.key}>{file.key}</p>
+            ))}
+            <Typography fontWeight={FONT_WEIGHT_BOLD}>
+              Optional files
+            </Typography>
+            {fileList.optional.map((file) => (
+              <p key={file.key}>{file.key}</p>
```
DafyddLlyr, line 65: I'm not actually sure if we need these here based on the Figma wireframes, but they will be needed in the modal.

editor.planx.uk/src/@planx/components/MultipleFileUpload/model.ts
```diff
@@ -82,3 +86,82 @@ export const newFileType = (): FileType => ({
...
+
+const isRuleMet = (
+  passport: Store.passport,
+  rule: ConditionalRule<Condition.RequiredIf | Condition.RecommendedIf>
+) => {
+  return (
+    passport.data?.[rule.fn] === rule.val ||
+    passport.data?.[rule.fn]?.includes(rule.val)
+  );
+};
```
jessicamcinchak, line 167: these logic methods feel super clear :raised_hands:

jessicamcinchak (approved):
Looks good!

editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
jessicamcinchak: :100: I think this also covers everything!

#1751 introduces a `slotsSchema` and local `handleSubmit()` to get us started around your last point

editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
jessicamcinchak: Guessing the `optional` list will be particularly long (especially if it needs to include more than the explicitly configured `Not required` files per my other question), so best contained to the modal makes sense to me :ok_hand:

editor.planx.uk/src/@planx/components/MultipleFileUpload/model.ts
```diff
@@ -82,3 +86,82 @@ export const newFileType = (): FileType => ({
...
+      fileList.required.push(fileType);
+      break;
+    case Condition.AlwaysRecommended:
+      fileList.recommended.push(fileType);
+      break;
+    case Condition.RequiredIf:
+      if (isRuleMet(passport, fileType.rule)) {
+        fileList.required.push(fileType);
+      }
+      break;
```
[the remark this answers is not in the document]
DafyddLlyr: This is a great question - your understanding here is correct of what's currently happening. I'll cross check this with @Alastairparvin just now to check what the intended behaviour should be.

editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
DafyddLlyr: Cool this matches what I think also so I've now dropped this from here.

PR #1758, opened 2023-06-07
feat: add a modal for tagging files in MultipleFileUpload

jessicamcinchak:
Has been feeling good to have small, incremental PRs for MultipleFileUpload to avoid big rebases as we work in parallel, so opening this "starter Dialog" for review now 

What it does so far:
- Opens on clicking "Change" from any uploaded file card, closes on clicking action dialog buttons or outside dialog
- Renders each uploaded file in the dialog
- Very basic initial styling to match Figma like square corners and top border
Screenshot from 2023-06-07 14-34-41

What it still needs to do: 
- Also open when a file is dropped in the dropzone
  - What about on file deletion or validation errors? Quickly realised this was trickier than hooking into existing `useEffect` hook that listens for changes to `slots`, so just started with straightforward option to open on "Change" for now
- Render a dropdown with grouped checklists of `fileTypes`
  - MUI docs https://mui.com/material-ui/react-select/#checkmarks && https://mui.com/material-ui/react-select/#grouping
- Show the selected `fileTypes` as "chips" in the Dialog select
  - MUI docs https://mui.com/material-ui/react-select/#chip
- `handleSubmit` or similar on "Done"

PR #1759, opened 2023-06-08
feat: display grouped checklists in MultipleFileUpload modal

jessicamcinchak:
building on #1758 

key changes:
- dynamically populates the lists of required, recommended and optional files specified by an editor on both main page next to dropzone and in the modal select menu
- displays checklists with subheaders in the select menu, saves the selection to local state (likely need a more robust data structure here, just a simple list of tags per file to start!)
- renames `fileType.key` to `fileType.name` to avoid colliding with React prop errors

future scope: 
- dialog "done" button should hook into `handleSubmit` or similar
- clean up styles
- accessibility tests

Screenshot from 2023-06-12 16-40-59


editor.planx.uk/src/@planx/components/MultipleFileUpload/Modal.tsx
```diff
@@ -60,3 +72,77 @@ export const FileTaggingModal = (props: FileTaggingModalProps) => {
...
+              <ListSubheader
+                key={`subheader-${fileListCategory}-files-${name}`}
+              >
+                {`${capitalize(fileListCategory)} files`}
+              </ListSubheader>
+              {fileList[fileListCategory].map((fileType) => (
+                <MenuItem
+                  key={`menuitem-${fileType.key}-${name}`}
+                  value={fileType.key}
+                >
```
jessicamcinchak, line 131: While troubleshooting a `handleChange` bug where checklist clicks weren't registering anything, I inspected `MenuItem` and noticed it appears that `key` is only reflecting `menuitem-${name}` which indicates that `fileType.key` is behaving as `undefined` even though it's not ! 

I think it's this error bubbling up again: https://github.com/theopensystemslab/planx-new/pull/1749#discussion_r1219461737

Can someone checkout this branch & double check my hypothesis? If we're seeing this again, shall I rename & update the type throughout to `fileType.name` ?

editor.planx.uk/src/@planx/components/MultipleFileUpload/model.ts
```diff
@@ -108,9 +108,9 @@ export const createFileList = ({
   const sortedFileTypes = sortFileTypes(fileTypes);
   const uniqueKeys: string[] = [];
```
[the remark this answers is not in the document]
jessicamcinchak: thanks for spotting, updated as well!

editor.planx.uk/src/@planx/components/MultipleFileUpload/Modal.tsx
```diff
@@ -60,3 +72,78 @@ export const FileTaggingModal = (props: FileTaggingModalProps) => {
...
+        id="select-multiple-file-tags"
+        labelId="select-multiple-file-tags-label"
+        variant="standard"
+        multiple
+        value={tags}
+        onChange={handleChange}
+        IconComponent={ArrowIcon}
+        input={<Input />}
+        inputProps={{ name }}
+        renderValue={(selected) => selected.join(", ")}
```
DafyddLlyr, line 110: I've not actually looked at the docs for this component in a while, but I think this is the prop that will allow us to display neat little chips here 💪

editor.planx.uk/src/@planx/components/MultipleFileUpload/Modal.tsx
```diff
@@ -60,3 +72,78 @@ export const FileTaggingModal = (props: FileTaggingModalProps) => {
...
+  };
+
+  return (
+    <FormControl key={name} sx={{ display: "flex", flexDirection: "column" }}>
+      <InputLabel id="select-mutliple-file-tags-label">
+        What does this file show?
+      </InputLabel>
+      <Select
+        id="select-multiple-file-tags"
+        labelId="select-multiple-file-tags-label"
```
DafyddLlyr, line 102: I would guess both of these need to be unique to make React (and a11y tests) happy, e.g. ``id={`select-multiple-file-tags-${name}`}``

editor.planx.uk/src/@planx/components/MultipleFileUpload/Modal.tsx
```diff
@@ -60,3 +72,78 @@ export const FileTaggingModal = (props: FileTaggingModalProps) => {
...
+                    <ListItemText
+                      key={`listitemtext-${fileType.name}-${name}`}
+                      primary={fileType.name}
+                    />
+                  </MenuItem>,
+                ];
+              }),
+            ];
+          })}
+      </Select>
```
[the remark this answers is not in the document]
DafyddLlyr: I've hit this exact same painful issue in `PermissionSelect` which has a similar setup of multiple groups within the select - maybe worth a look to see how it's handled there for reference, but this is pretty much the same 👍

editor.planx.uk/src/@planx/components/MultipleFileUpload/Modal.tsx
```diff
@@ -60,3 +72,78 @@ export const FileTaggingModal = (props: FileTaggingModalProps) => {
...
+  const [tags, setTags] = React.useState<string[]>([]);
+
+  const handleChange = (event: SelectChangeEvent<typeof tags>) => {
+    const {
+      target: { value },
+    } = event;
+    setTags(
+      // On autofill we get a stringified value.
+      typeof value === "string" ? value.split(",") : value
+    );
```
DafyddLlyr, line 92: This is actually a nice setup - we get a discrete change event here each time a checkbox is checked / unchecked. 

Maybe we simplify things a bit and we can just assign slots -> fileList directly here and not map a list of tags on modal submit / close?

So something like...
 - When we check a box...
 - We know the slot, we know the tag, we know the category (required/recommended/optional)
 - Assign the slot to the filetype, call `setFileState()`

Maybe something like...
 
 ```ts
const updatedFileList = {...fileList};
updatedFileList[category].find(fileType => fileType.name === name).slot = slot;
setFileList(updatedFileList);
 ```
 
 I guess unchecking would be just the same but `slot === undefined`

editor.planx.uk/src/@planx/components/MultipleFileUpload/Modal.tsx
jessicamcinchak: Yes exactly ! There's a nice example of just this in the docs here :nerd_face: https://mui.com/material-ui/react-select/#chip

editor.planx.uk/src/@planx/components/MultipleFileUpload/Modal.tsx
jessicamcinchak: updated!

jessicamcinchak:
> I've been assuming that each time we checked a file name, it would drop off the remaining checklists. So I couldn't tag two files as "heritage statement" because once I did the first one, that option would be disabled on the second dropdown. Does this sound right or wrong?

@DafyddLlyr  - so I've been thinking that multiple uploads _can_ share the same tag, therefore the tags would NOT drop off or become disabled after first use. Currently with single FileUpload, we know that it may require 2+ actual files to fulfill the single passport variable asked for, so I've been assuming we need to continue holding that same assumption with this new component, even though we expect it to happen less in practice with this UI.

But it does sound like we might want some visual indicator in checklists to show that the _requirement_ has been satisfied already, even though you can still use the tag _again_ if you want. Although this means the various lists need to become aware of one another which might get tricky? Maybe we can use the same :heavy_check_mark: icons that we need to still introduce in the list next to the dropzone ?? One to double check with content & design for sure.

DafyddLlyr:
> so I've been thinking that multiple uploads can share the same tag, therefore the tags would NOT drop off or become disabled after first use. Currently with single FileUpload, we know that it may require 2+ actual files to fulfill the single passport variable asked for, so I've been assuming we need to continue holding that same assumption with this new component, even though we expect it to happen less in practice with this UI.

Yeah - I do think that this is correct 👍 It will need a few small changes like `userFile.slot: FileUploadSlot` -> `userFile.slots: FileUploadSlot[]` and some tests updated but it's the way to go here.

editor.planx.uk/src/@planx/components/MultipleFileUpload/Modal.tsx
DafyddLlyr: 🎉 👏

PR #1760, opened 2023-06-08
feat: MFU `handleSubmit()` and basic validation logic

DafyddLlyr:


editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
```diff
@@ -40,52 +45,52 @@ const DropzoneContainer = styled(Box)(({ theme }) => ({
...
-  );
-  const [slots, setSlots] = useState<FileUploadSlot[]>(recoveredSlots ?? []);
+  useEffect(() => {
+    const passport = useStore.getState().computePassport();
+    const fileList = createFileList({ passport, fileTypes: props.fileTypes });
+    setFileList(fileList);
+  }, []);
+
+  useEffect(() => {
+    // TODO: Re-map slots to userfiles also?
```
DafyddLlyr, line 62: This is a genuine `TODO` to address once the modal is doing it's thing

editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
```diff
@@ -40,52 +45,52 @@ const DropzoneContainer = styled(Box)(({ theme }) => ({
...
-      .validate(slots)
+    // This is a temp cheat to bypass tagging
+    // The first slot is mapped to a single required file
+
+    // const fileListWithTaggedFile = {...fileList}
+    // fileListWithTaggedFile.required[0].slot = slots[0]
+    // setFileList(fileListWithTaggedFile);
+
+    Promise.all([
+      slotsSchema.validate(slots),
```
DafyddLlyr, line 88: One thing I'm not 100% certain on just now is how we validate that all uploaded files are tagged. I expect this will be part of the modal logic, but it might be nice at this stage to validate that the length of `slots` is equal to the filtered `fileList`. This is something I'll come back to shortly once the modal tagging logic is in place.

jessicamcinchak (approved):
Very happy with this starting point, we can keep wiring up once tags are working!

PR #1764, opened 2023-06-08
feat: Make new property label editable

DafyddLlyr:
image


DafyddLlyr:
[Allow text in 'Find property on a map instead' to be changed in editor](https://trello.com/c/MyRi46i6/2368-allow-text-in-find-property-on-a-map-instead-to-be-changed-in-editor)

jessicamcinchak (approved):

PR #1766, opened 2023-06-08
fix: Make e2e tests pass first time round

builditben:
Playwright test assertions can wait until the assertion is met (or a timeout) when `await`'ed: https://playwright.dev/docs/test-assertions

Adding `await` before each expectation fixes the issue of these e2e tests failing on the first run through.


DafyddLlyr (approved):
This is great - thanks for taking the time to look into this one.

jessicamcinchak (approved):

PR #1767, opened 2023-06-08
chore: Set-up regression test suite

builditben:
This PR introduces the notion of a regression test suite.

Running `pnpm test` in the e2e directory will run all tests **not** marked as a regression test. If `@regression` is in the name (Playwright) of the test or above the test (Cucumber), then the test will only run with the `pnpm test:regression` command.

This PR also introduces a cron job to run the regression test suite every day in the morning.


.github/workflows/automated-tests.yml
DafyddLlyr, line 1: I think a notification on failure here might be a good step

DafyddLlyr (approved):

PR #1769, opened 2023-06-09
fix: Correctly format conditions

DafyddLlyr:
Something I missed in the previous review - I was using `lowerCase()` as a way of splitting the `Condition` enum between words with a space

Docs: https://lodash.com/docs/4.17.15#lowerCase

| Before | After |
|--------|--------|
| image | image |


jessicamcinchak (approved):
did not realize `lowerCase` split words - good to know!

PR #1771, opened 2023-06-12
fix: Result reasons shouldn't have hardcoded green band

jessicamcinchak:
Fixes https://opensystemslab.slack.com/archives/C01E3AC0C03/p1686565382480159

Quickly updated to use `paper` color for all "Reasons". In the future, the band color should match the flag color if the question has an associated flag, or fallback to `paper` if it doesn't have an associated flag? Currently, the `ResultReason` doesn't have access to the flag info; it'll need to join to flow data on `id` in order to set the color. 
Screenshot from 2023-06-12 12-38-41


DafyddLlyr (approved):

DafyddLlyr:
> the band color should match the flag color if the question has an associated flag, or fallback to paper if it doesn't have an associated flag?

This sounds correct to me!

PR #1773, opened 2023-06-12
fix: Guard against multiple payments for ITP sessions

DafyddLlyr:
Addresses issue raise here - https://trello.com/c/2n0EJ3wA/1635-invite-to-pay-epic#comment-6481a3db0d31432020c497bf

Relies on https://github.com/theopensystemslab/planx-core/pull/41

This PR implements the following two checks - 
 - Do not allow payment request to be generated for a session where a payment is already initialised
 - Do not allow payment for sessions where a payment request has been generated


api.planx.uk/pay/index.test.ts
```diff
@@ -0,0 +1,171 @@
...
+        description: "New application",
+        return_url: "https://editor.planx.uk",
+      })
+      .expect(200)
+      .then((res) => {
+        expect(res.body).toEqual(govUKResponse);
+      });
+  });
+
+  it.skip("succeeds with missing query string", async () => {
```
DafyddLlyr, line 61: I've skipped this test just to flag it in review - it looks very intentional, but a `sessionId` is going to be required for this API call going forward. I think it's likely that all other query params are also required (they're used for the audit log).

Is there any context around this test or why query params should be optional here @builditben  - is it to do with how proxies are nested for ITP and query params not necessarily passed on?

If so, I'm sure I can do a check for `sessionId` in either body or query params but I just wanted to cross check before making assumptions here.

api.planx.uk/pay/pay.ts
```diff
@@ -27,16 +29,38 @@ export async function makePaymentViaProxy(
...
+        status: 400,
+      })
+    );
   }
 
   const flowId = req.query?.flowId as string | undefined;
   const sessionId = req.query?.sessionId as string | undefined;
   const teamSlug = req.params.localAuthority;
 
+  if (!flowId || !sessionId || !teamSlug ) {
```
DafyddLlyr, line 44: Please see comment above on optional query params - this might be an overcorrection.

api.planx.uk/pay/index.test.ts
DafyddLlyr: Thanks for the sense check that I'm not missing something silly here - I'll take a closer look 🔍

builditben (approved):
Looks good to me :+1: 

For an attempted payment after inviting someone to pay, the feedback looks like an application error and there's no obvious way to get back to a happy path but I think that should be fine for an edge-case like this (or until we see more people hitting this error). As a future improvement, maybe we suggest that payment can still be made via the invite to pay link and present the link on the error page.

When I try to re-send the invite to pay link, which I might do if I'm confused and have hit the back button, I don't get the best validation response (something like "could not generate a payment request") but again, this is a edge case. Preventing the additional request is good enough for now in my opinion.

PR #1774, opened 2023-06-12
feat: Automatically open tagging modal when file is uploaded

DafyddLlyr:
## What does this PR do?
Adds the following behaviour to the `MultipleFileUpload` public component - 
 - `FileTaggingModal` does not open on component mount
 - `FileTaggingModal` opens when a single file is uploaded
 - `FileTaggingModal` opens when multiple files are uploaded
 - `FileTaggingModal` is not opened when files are deleted

Also a quick a11y fix, and some basic tests for the above cases.

### DIFF
--- a/editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
@@ -143,24 +143,22 @@ function Component(props: Props) {
         <List disablePadding sx={{ width: "100%" }}>
           {(Object.keys(fileList) as Array<keyof typeof fileList>)
             .filter((fileListCategory) => fileList[fileListCategory].length > 0)
-            .map((fileListCategory) => (
-              <Box key={`wrapper-${fileListCategory}-files`}>
-                <ListSubheader
-                  key={`subheader-${fileListCategory}-files`}
-                  disableGutters
-                >
-                  {`${capitalize(fileListCategory)} files`}
-                </ListSubheader>
-                {fileList[fileListCategory].map((fileType) => (
-                  <ListItem key={fileType.name} disablePadding>
-                    <InteractiveFileListItem
-                      name={fileType.name}
-                      moreInformation={fileType.moreInformation}
-                    />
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
```diff
@@ -70,6 +72,13 @@ function Component(props: Props) {
 
   const [slots, setSlots] = useState<FileUploadSlot[]>([]);
 
+  // Track number of slots, and open modal when this increases
+  const previousSlotCount = usePrevious(slots.length);
+  useEffect(() => {
+    if (previousSlotCount === undefined) return;
+    if (slots.length > previousSlotCount) setShowModal(true);
+  }, [slots.length]);
```
jessicamcinchak, line 80: nice use of `usePrevious` here :heavy_plus_sign: 

also didn't think of passing `[slots.length]` in the callback like this - super handy !

editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
```diff
@@ -131,14 +140,17 @@ function Component(props: Props) {
...
+          <List disablePadding>
+            {fileList.required.map((fileType) => (
+              <ListItem key={fileType.key} disablePadding>
+                <InteractiveFileListItem
+                  name={fileType.key}
+                  moreInformation={fileType.moreInformation}
+                />
+              </ListItem>
+            ))}
+          </List>
```
jessicamcinchak, line 153: :+1: for this a11y fix - I also added an outer `List` and replaced `Typography` with a `ListSubheader` over here https://github.com/theopensystemslab/planx-new/pull/1759/files#diff-36665c16b9d4531b049313c7577008ca7ec35d1a34915ae16f1fc9e55803d3f4R135 

just as heads up so we can keep an eye on this one while rebasing!

jessicamcinchak (approved):
Works perfect! Thanks for starting `Public.test.tsx` too :100:

PR #1775, opened 2023-06-13
feat: Page container / layout restructure

ianjon3s:
PR aims to simplify and unify the container system.

- Adds two new media queries, `contentWrap` for page-level containers and `formWrap` to use as a max-width for form content.
- Adds MUI `<Container>` to all page level items to unify padding and max-widths.
- Sets global padding for `<Container>` at theme level.
- Removes `getHeaderPadding` property as this is inherited from containers.
- Uses wildcard selector (`& > *`) to apply max-width to form items in `Card.ts`, and targeted styles to set the following to full-width: Upload and label, maps, radio and checkbox with description and/or images.
- Adds targeted max-widths to invite-to-pay hardcoded content.

### DIFF
--- a/editor.planx.uk/src/@planx/components/Checklist/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/Checklist/Public.tsx
@@ -1,8 +1,10 @@
 import Box from "@mui/material/Box";
 import Grid from "@mui/material/Grid";
+import { styled } from "@mui/material/styles";
 import type { Checklist, Group } from "@planx/components/Checklist/model";
 import ImageButton from "@planx/components/shared/Buttons/ImageButton";
 import Card from "@planx/components/shared/Preview/Card";
+import { fullWidthContent } from "@planx/components/shared/Preview/MapContainer";
 import QuestionHeader from "@planx/components/shared/Preview/QuestionHeader";
 import { useFormik } from "formik";
 import React, { useState } from "react";
@@ -43,6 +45,15 @@ function getFlatOptions({
   }
   return [];
 }
+
+const FullWidthContainer = styled(Box)(({ theme }) => ({
+  ...fullWidthContent(theme),
+}));
+
+const FormWrapper = styled(Box)(({ theme }) => ({
+  width: theme.breakpoints.values.formWrap,
+  maxWidth: "100%",
+}));
 
 const ChecklistComponent: React.FC<Props> = ({
   allRequired,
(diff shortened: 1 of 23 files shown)


editor.planx.uk/src/@planx/components/Checklist/Public.tsx
```diff
@@ -44,6 +46,15 @@ function getFlatOptions({
...
 }
 
+const FullWidthContainer = styled(Box)(({ theme }) => ({
+  ...fullWidthContent(theme),
+}));
+
+const FormWrapper = styled(Box)(({ theme }) => ({
+  width: theme.breakpoints.values.formWrap,
+  maxWidth: "100%",
+}));
```
DafyddLlyr, line 56: These two components and their styles are repeated a few time across various files. There a are a few ways we could reduce this - 

- Make a variant (e.g. `<Container variant="formWrapper">`)
- Make a standalone UI component (e.g. `<FormWrapper>`)

I don't really think there's a compelling difference between the two - it's much more about semantics and what's clearer when reading the code. Generally we're favouring standalone components over variants and I think that's probably the only deciding factor here. It's worth having a look through [the docs](https://mui.com/material-ui/customization/theme-components/#creating-new-component-variants) about variants to get a good understanding of the concept though.

Here's how I'd approach making a standalone UI component here - 

 - Make new components for both of these in `src/ui` directory (there should be a few simple examples here)
 - Import these components into the files where they're needed
 
If this isn't clear lets jump on a call and do some pairing on it 👨‍💻 👨‍💻

editor.planx.uk/src/theme.ts
```diff
@@ -27,6 +27,18 @@ export const FONT_WEIGHT_BOLD = "700";
...
+  interface BreakpointOverrides {
+    xs: true;
+    sm: true;
+    md: true;
+    lg: true;
+    xl: true;
+    formWrap: true;
+    contentWrap: true;
+  }
+}
```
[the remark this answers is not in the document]
ianjon3s: Agreed, I'll keep that file in mind for future overrides, updated.

editor.planx.uk/src/@planx/components/Checklist/Public.tsx
ianjon3s: I had looked into using variants for `<Container>` before and MUI doesn't support variants for this, so a standalone UI seems like the best route. I've had a go at setting one up based on the other examples, seems to work as expected.

DafyddLlyr: Ah yeah - not all components allow variants - that's a bit of an annoying restriction but it's not a real issue here!

PR #1707, opened 2023-05-19
refactor: Remove `makeStyles` classes in favour of styled components

DafyddLlyr:
Boring and repetitive work which helped the long train journey pass well yesterday!

There may be a few concerns with conflicts against some of Ian's work but I'm happy to rebase this, or even cherry-pick out commits that we can take forward a bit at a time. I appreciate there's a decent amount here for one PR sorry.

PR #1780, opened 2023-06-13
feat: Styled `StatusPage`

DafyddLlyr:
Cherry-picked from https://github.com/theopensystemslab/planx-new/pull/1707 🍒

### DIFF
--- a/editor.planx.uk/src/pages/Preview/StatusPage.tsx
+++ b/editor.planx.uk/src/pages/Preview/StatusPage.tsx
@@ -3,7 +3,6 @@ import Button from "@mui/material/Button";
 import Link from "@mui/material/Link";
 import { useTheme } from "@mui/material/styles";
 import Typography from "@mui/material/Typography";
-import makeStyles from "@mui/styles/makeStyles";
 import Card from "@planx/components/shared/Preview/Card";
 import { useStore } from "pages/FlowEditor/lib/store";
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/pages/Preview/StatusPage.tsx
```diff
@@ -57,7 +50,6 @@ const StatusPage: React.FC<Props> = ({
   });
 
   const theme = useTheme();
```
jessicamcinchak, line 52: question: as we're moving away from `useStyles()` should we also be aiming to move away from `useTheme()` in favor of `sx` with the `theme` callback value?

jessicamcinchak (approved):

editor.planx.uk/src/pages/Preview/StatusPage.tsx
DafyddLlyr: This is a good question!

`useTheme()` is not deprecated in MUI v5 like `useStyles()` is, so we can safely continue using it. There are some cases where we likely can't avoid it just now - for example in this component we're using `theme` to pass a value into another component as a custom prop (so `sx` with a callback wouldn't work).

If there are places we're using `useTheme()` just to pass a value where we we could use `sx`, I'd probably favour `sx` as being a bit neater and and terser, but I think it's really just a personal preference.

PR #1783, opened 2023-06-13
UPDATE: sections styling

ianjon3s:
PR updates styling for sections to better match Figma designs.

Also modifies question header to use redundant 'subtitle1' property, to make 'subtitle2' a style for smaller headings.

Screenshot below shows before (left) and after (right):
image


DafyddLlyr:
@ianjon3s If you're able to please put a before/after screenshot in the PR description that would be really helpful in just visually identifying the changes here. Doesn't need comments or annotation or anything like this, just straight screenshots, but it's helpful to just cross check that things are as they should be on the Pizza.

ianjon3s:
@DafyddLlyr affirmative, I've updated the description with screenies 👍

editor.planx.uk/src/@planx/components/Section/Public.tsx
```diff
@@ -163,29 +169,37 @@ export function SectionsOverviewList({
...
-    paddingTop: theme.spacing(2),
+    borderBottom: `1px solid ${theme.palette.secondary.main}`,
+    paddingBottom: theme.spacing(1.5),
+    paddingTop: theme.spacing(1.5),
     verticalAlign: "top",
     margin: 0,
   },
+  "& > dt:first-of-type, & > dd:first-of-type": {
+    borderTop: `1px solid ${theme.palette.secondary.main}`,
+  },
```
jessicamcinchak, line 185: Is this top border for the first list item GovUK standard? 

It's the only Sections change here that stands out as a bit odd to me! Mostly because it feels inconsistent with how Checklists & the Review page use borders between list items (eg only bottom).
Screenshot from 2023-06-13 17-31-19
Screenshot from 2023-06-13 17-32-16
Screenshot from 2023-06-13 17-34-13

ianjon3s: GOV.UK are also inconsistent with this (you caught them, HA!). Their task list (our sections) has borders top & bottom and summary list (our review) has only bottom. They also do accordions very differently so we can disregard that.

For simplicity I'll remove the top border so that _our_ components are consistent.

editor.planx.uk/src/@planx/components/Section/Public.tsx
```diff
@@ -138,21 +138,27 @@ export function SectionsOverviewList({
...
+                <Link
+                  onClick={() => changeFirstAnswerInSection(sectionId)}
+                  component="button"
+                  sx={{
+                    fontFamily: "inherit",
+                    fontSize: "inherit",
+                    textAlign: "left",
+                  }}
+                >
+                  {sectionNode.data.title}
```
DafyddLlyr, line 153:
```suggestion
            <Typography variant="body2">
                  {sectionNode.data.title}
            </Typography>
```

This approach (`Link` wrapping a `Typography`) is more consistent with the code base and allows you to drop the `Link`'s `sx` prop entirely (the awkward workaround to inherit font details).

Appreciate this isn't your change here but seems worth tidying up 👍

DafyddLlyr (approved):
Looking good 👌

editor.planx.uk/src/@planx/components/Section/Public.tsx
jessicamcinchak: Thanks for looking into this one & adjusting !

jessicamcinchak (approved):

PR #1784, opened 2023-06-13
fix: Support multiple slots per `userFile`

DafyddLlyr:
## What does this PR do?
- Changes the `UserFile` type to support have `FileUploadSlot[]` instead of just a single `FileUploadSlot`
- This matches our current file upload data structure, and should work with submissions as-is 🤞


jessicamcinchak (approved):
Thanks for this one! Happy for you to merge this first then I'll rebase & bring the logic up to date in #1777 to account for many `slots` :+1:

PR #1800, opened 2023-06-15
UPDATE: multiple file upload styling

ianjon3s:
PR further builds on multiple upload styling.

Summary of changes:

- Further adjustment of type sizes, padding etc
- Styling of dropzone to prevent text jumping on drag state
- Styling of select input
- Styling and positioning of select input label + shrink state (I've used `"&[data-shrink=true]": {` for this to do it with CSS, there may be a better way to do this programatically)
- Update of chip layout to allow for wrapping onto several lines (thinking about smaller devices), screenshot below:

image


jessicamcinchak:
Noticing this pizza failed to sync content, going to manually destroy the server & rebuild CI steps before reviewing just FYI ! You don't have to do anything / didn't do anything wrong - this is just flakey once in awhile :ok_hand:

jessicamcinchak (approved):
Okay back in business with a fresh :pizza: now: https://1800.planx.pizza/braintree/file-test/preview?analytics=false

These changes all look great ! 

Two minor things I'm still noticing / no strong preference if they're picked up here or in a future PR: 
1. Question header help text & per-file help text are still a bit misaligned 
2. Error border doesn't match dropzone height / dropzone should become shorter when the erorr is visible rather than shift downwards? 
Screenshot from 2023-06-16 09-47-12
Screenshot from 2023-06-16 09-47-19

ianjon3s:
Thanks for picking this up @jessicamcinchak ,

I've pushed up a couple of changes that fix the above.

- Add property to error wrapper so that it expands rather than overflows content
- Shift alignment of heading help icon so that it aligns with columns within content

image

editor.planx.uk/src/ui/ErrorWrapper.tsx
```diff
@@ -15,6 +15,8 @@ const useClasses = makeStyles((theme) => ({
     width: "100%",
     paddingLeft: theme.spacing(1.5),
     borderLeft: `5px solid ${theme.palette.error.main}`,
+    display: "flex",
+    flexDirection: "column",
```
jessicamcinchak, line 19: :raised_hands: perfect

editor.planx.uk/src/@planx/components/MultipleFileUpload/Modal.tsx
```diff
@@ -147,7 +148,17 @@ const SelectMultiple = (props: SelectMultipleProps) => {
     >
       <InputLabel
         id={`select-mutliple-file-tags-label-${uploadedFile.id}`}
-        sx={{ top: "15%", color: (theme) => theme.palette.text.primary }}
+        sx={{
+          top: "16%",
+          textDecoration: "underline",
+          color: (theme) => theme.palette.primary.main,
+          "&[data-shrink=true]": {
```
DafyddLlyr, line 155: @ianjon3s @jessicamcinchak Just for future reference (no changes needed just now!), I think I would have solved this one like this - 

```ts
const StyledInputLabel = styled(InputLabel)(({ theme, shrink }) => ({
    top: "16%",
    textDecoration: "underline",
    color: theme.palette.primary.main,
    ...(shrink && {
      textDecoration: "none",
      color: theme.palette.text.primary,
      top: "0",
      transform: "translate(14px, -5px) scale(0.85)",
    })
  }));
```

And then using it later...

```jsx
<StyledInputLabel
  id={`select-mutliple-file-tags-label-${uploadedFile.id}`}
  >
  What does this file show?
</StyledInputLabel>
```

The benefits of this would be...
 - Type safety - you could do `&[data-shrink="some string"]`, `&[data-shrink>6]` or `&[data-someMadeUpProp=true]` which would be invalid, but in the example above we know that `shrink` is a `Boolean | undefined` and it's a valid prop for an `InputLabel`
 - We could do more JS-ey stuff with the prop rather than a straight comparison, e.g. `someFunctionThatCalculatesSomething(shrink)`

ianjon3s: Thanks @DafyddLlyr , I'm keen to use the best method for doing this kind of thing, and also to understand how this is working. I'll probably bother you for a quick chat at some point next week to talk this over.

DafyddLlyr: Yeah send me a calendar invite for next week and lets make it happen. It's 100% no bother and upskilling is a key part of all our roles 😄 

If you have a set of questions/problems you can prep/send in advance that would be super helpful, even if it's just very rough or basic.

PR #1801, opened 2023-06-16
chore: add tests for multiple file upload tagging logic

jessicamcinchak:
Adds tests for adding & removing tags as defined in model.ts. Public UI tests related to tagging & syncing the display of tags across components coming next in a separate PR.


editor.planx.uk/src/@planx/components/MultipleFileUpload/schema.ts
```diff
@@ -73,7 +73,9 @@ export const fileListSchema = object({
     message: "Please upload and tag all required files",
     test: (userFile?: UserFile[]) => {
       const isEverySlotFilled = Boolean(
-        userFile?.every((userFile) => userFile?.slots)
+        userFile?.every(
+          (userFile) => userFile?.slots && userFile.slots.length > 0
+        )
```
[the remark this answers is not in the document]
DafyddLlyr: Thanks for picking this up - agree that this more explicit and a simpler solution 👍

editor.planx.uk/src/@planx/components/MultipleFileUpload/model.ts
```diff
@@ -265,7 +265,9 @@ export const addOrAppendSlots = (
         const updatedFileType = updatedFileList[category][index];
         if (
           updatedFileType.slots &&
-          !updatedFileType.slots.includes(uploadedFile)
+          !updatedFileType.slots
+            .map((slot) => slot.id)
+            .includes(uploadedFile.id)
```
DafyddLlyr, line 270: 👍

editor.planx.uk/src/@planx/components/MultipleFileUpload/model.test.ts
```diff
@@ -360,3 +372,129 @@ describe("getRecoveredSlots function", () => {
     expect(result[0]).toMatchObject(mockCachedSlot);
   });
 });
+
+describe("getTagsForSlot function", () => {
```
DafyddLlyr, line 376: Could we add a test case here for a slot without any tags - I'm relying on this function to handle this in #1811 so it would be good to cover this scenario.

DafyddLlyr (approved):
Tests perfectly describe the test scenarios - super clear 👌

editor.planx.uk/src/@planx/components/MultipleFileUpload/model.test.ts
jessicamcinchak: Good shout, now covered ! Will return `[]` in cases where a slot does not have any tags fyi !

PR #1802, opened 2023-06-16
feat: Styled `Calculate`

DafyddLlyr:
Cherry-picked from https://github.com/theopensystemslab/planx-new/pull/1707 🍒

### DIFF
--- a/editor.planx.uk/src/@planx/components/Calculate/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/Calculate/Editor.tsx
@@ -19,7 +19,7 @@ import { evaluate, getVariables, parseCalculate } from "./model";
 
 export type Props = EditorProps<TYPES.Calculate, Calculate>;
 
-const Were = styled("span")(() => ({
+const ConditionLabel = styled("span")(() => ({
   alignSelf: "center",
   margin: "auto 1rem",
   textAlign: "center",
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/@planx/components/Calculate/Editor.tsx
```diff
@@ -123,10 +120,10 @@ export default function Component(props: Props) {
           {variables?.length ? (
             <>
               <p>If</p>
-              {variables.map((variable, i) => (
+              {variables.map((variable) => (
                 <InputRow key={variable}>
                   <Input value={variable} disabled />
-                  <span className={classes.were}>were</span>
+                  <Were>were</Were>
```
jessicamcinchak, line 126: nit: totally appreciate this isn't introduced here, but this language feels a little awkward ! 

in the future, perhaps we can simplify to "If {x} **'is** {y}, then {fn} **will be** set to {output}" - but happy for this to be revisited, no need to open that can of worms now

jessicamcinchak (approved):

editor.planx.uk/src/@planx/components/Calculate/Editor.tsx
DafyddLlyr: Yeah this is totally fair - I felt this was awkward and couldn't hit on a good name. Will go for `ConditionLabel` thanks 👍

PR #1810, opened 2023-06-19
fix: Missing margin on `FileUpload`

DafyddLlyr:
| Before | After |
|--------|--------|
| image | image | 

A more robust solution for all components here may be having the wrapping `Card` component managing this spacing and using `margin-trim` (docs: https://developer.mozilla.org/en-US/docs/Web/CSS/margin-trim) to ensure nothing gets doubled up?

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUpload/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUpload/Public.tsx
@@ -1,3 +1,4 @@
+import Box from "@mui/material/Box";
 import { MoreInformation } from "@planx/components/shared";
 import Card from "@planx/components/shared/Preview/Card";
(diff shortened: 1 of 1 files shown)


ianjon3s:
Hey @DafyddLlyr this issue appears to be more of global regression. The card component contains a wildcard (` > * + * `) margin-top that is applied to all children. Something has changed within the button component which is causing it to override this with a margin of 0. I'll take a look at this tomorrow first thing.

DafyddLlyr:
Thanks @ianjon3s - you're correct, it looks like this is tied to this recent change https://github.com/theopensystemslab/planx-new/pull/1796

The selector on the card is not being correctly applied any longer. I'll also take a look at it to try and work out what's causing this.

DafyddLlyr:
Closed in favour of https://github.com/theopensystemslab/planx-new/pull/1814

PR #1811, opened 2023-06-19
fix: MFU "Back" behaviour

DafyddLlyr:
## What does this PR do?
- Allows an applicant to navigate "back" to an MFU component and retain their state
  - This was a regression I missed when changing from a single `userFile.slot` to `userFiles.slots` in https://github.com/theopensystemslab/planx-new/pull/1784
- Fixes a few schema validation issues to enforce the following - 
  - All user uploaded files must be tagged
  - All required fileTypes must have associates slots (also fixed here - https://github.com/theopensystemslab/planx-new/pull/1801#discussion_r1233842550)

### DIFF
--- a/editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
@@ -98,7 +98,7 @@ function Component(props: Props) {
   const handleSubmit = () => {
     Promise.all([
       slotsSchema.validate(slots),
-      fileListSchema.validate(fileList),
+      fileListSchema.validate(fileList, { context: { slots } }),
     ])
       .then(() => {
         const payload = generatePayload(fileList);
--- a/editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
@@ -7,12 +7,10 @@ import { styled } from "@mui/material/styles";
 import Typography from "@mui/material/Typography";
 import { PublicProps } from "@planx/components/ui";
 import capitalize from "lodash/capitalize";
-import merge from "lodash/merge";
 import { useAnalyticsTracking } from "pages/FlowEditor/lib/analyticsProvider";
 import { useStore } from "pages/FlowEditor/lib/store";
 import React, { useEffect, useRef, useState } from "react";
 import { usePrevious } from "react-use";
-import { FONT_WEIGHT_BOLD } from "theme";
 import ErrorWrapper from "ui/ErrorWrapper";
 import MoreInfoIcon from "ui/icons/MoreInfo";
 import ReactMarkdownOrHtml from "ui/ReactMarkdownOrHtml";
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/@planx/components/MultipleFileUpload/model.ts
```diff
@@ -1,5 +1,7 @@
+import cloneDeep from "lodash/cloneDeep";
```
DafyddLlyr, line 1: TIL: The native `structuredClone()` is not yet supported in our version of Jest / JSDOM https://github.com/jsdom/jsdom/issues/3363

editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.tsx
```diff
@@ -65,23 +63,26 @@ function Component(props: Props) {
...
-    setSlots(recoveredSlots);
-  }, [props.previouslySubmittedData, fileList]);
+    if (props.previouslySubmittedData) {
+      const recoverredData = getRecoveredData(
+        props.previouslySubmittedData,
+        fileList
+      );
+      setSlots(recoverredData.slots);
+      setFileList(recoverredData.fileList);
+      setIsUserReturningToNode(true);
```
DafyddLlyr, line 74: This feels a bit awkward but it works - I couldn't find another way using existing state and stores to track if a user has just clicked back. The modal needs to still behave as normal when uploading files after clicking "back" (e.g. to upload more files)

editor.planx.uk/src/@planx/components/MultipleFileUpload/schema.test.ts
```diff
@@ -149,3 +156,97 @@ describe("multipleFileUploadSchema", () => {
...
+      recommended: [],
+      optional: [],
+    };
+
+    await expect(() =>
+      fileListSchema.validate(mockFileList, { context: { slots: mockSlots } })
+    ).rejects.toThrow(/Please tag all files/);
+  });
+
+  it("rejects if any slots are untagged", async () => {
```
jessicamcinchak, line 221: Requiring _all_ uploaded files to have tags, rather than just satisfying the "required" tags is interesting / not totally intuitive to me! 

We currently allow `FileUpload` component to be created without setting a passport variable - and I'd be curious if there's any legitimate/regular uses for that in "real" content that might counteract this rule? But happy to go into testing with this assumption and see if it comes up :+1:

jessicamcinchak (approved):
Works smoothly :raised_hands:

editor.planx.uk/src/@planx/components/MultipleFileUpload/schema.test.ts
DafyddLlyr: This is a fair question! Here's my reasoning - 

 - Unless a file is tagged, we don't know what it is and where to put it in the passport.
 - If you upload a file on staging atm via `MultipleFileUpload` it will not get added to the passport or breadcrumbs unless it is tagged - in that context this validation makes sense.
 
You are right though, `FileUpload` files do still pass along "anonymous" files to BOPS (untagged and missing description)/Uniform/Email if we don't have a passport variable for them. It does this by assigning them to `passport.nodeId` if there isn't a value set for the component. We could implement this in `generatePayload()` and drop the validation here?

I think it's a good question to raise - let's see what come back from testing. I think we can add it to a list of questions to flush out with Alastair/POs.

PR #1812, opened 2023-06-19
chore: add tests for multiple file upload tagging UI 

jessicamcinchak:
Builds on #1801 

Adds tests to `Public.test.tsx` that cover: uploading a file, opening the modal, adding a tag, closing the modal, checking state has synced (eg uploaded card shows chips and complete/incomplete icons are correct), clicking "Continue" and either passing/failing validation.

Plus one small one to check that we render help/more info icons as expected.

Hopefully a good starting point to keep building on (eg "change" & "remove" tags, "delete" an uploaded file, etc)


editor.planx.uk/src/@planx/components/MultipleFileUpload/Modal.tsx
```diff
@@ -152,8 +152,10 @@ const SelectMultiple = (props: SelectMultipleProps) => {
         What does this file show?
       </InputLabel>
       <Select
+        native={true}
```
jessicamcinchak, line 155: This prop was brought to my attention when inspecting hte DOM for testing (there wasn't actually a `<select>`!). While I ended up using `data-testid` for test selectors, I think we'll still want this prop set to `true` for accessibility/screen reading devices. See this thread: https://stackoverflow.com/a/55185837

editor.planx.uk/src/@planx/components/MultipleFileUpload/Modal.tsx
```diff
@@ -163,6 +163,7 @@ const SelectMultiple = (props: SelectMultipleProps) => {
         What does this file show?
       </InputLabel>
       <Select
+        native={false}
```
jessicamcinchak, line 166: This is a bit of a silly change (it's just explicitly defining the default prop value) but one i think we're going to need to address in the future and this feels like a helpful way to remember.

This prop was brought to my attention when inspecting the DOM for testing (there wasn't actually a `<select>`!). While I ended up adding `data-testid` to `inputProps` to get test selectors working, I think we'll eventually want this prop set to `true` for accessibility/screen reading devices. See this thread: https://stackoverflow.com/a/55185837

Whenever we come back to this (wait to see if flagged during a11y audit?), there were a couple immediate things I noticed when locally switching to `native={true}`: console throws error that our `options` are invalid because they include list subheaders, the "what does this file show" label/placeholder becomes disconnected, and styles overall get really wonky.

DafyddLlyr: Thanks for the investigation and explanation 👌 

Happy with the "wait and see" approach here as we know that an audit will be around shortly.

editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.test.tsx
```diff
@@ -208,3 +222,133 @@ describe("Modal trigger", () => {
...
+    const fileTaggingModal = await within(document.body).findByTestId(
+      "file-tagging-dialog"
+    );
+
+    // The number of selects in the modal matches the number of uploaded files
+    const selects = await within(document.body).findAllByTestId("select");
+    expect(selects).toHaveLength(1);
+
+    // Apply multiple tags to this file
+    fireEvent.change(selects[0], { target: { value: "Roof plan" } });
```
DafyddLlyr, line 263: I'm guessing this is a necessary compromise as without `native={true}` we can't use something like `user.selectOptions()`?

DafyddLlyr (approved):
Helpful comments walking through the tests and describing what's happening👌

editor.planx.uk/src/@planx/components/MultipleFileUpload/Public.test.tsx
jessicamcinchak: exactly - even with `native={true}` which I fiddled with locally for awhile `user.selectOptions()` still won't work exactly as expected unless we refactor our <options> to be a more true list of just checkboxes without list subheaders which was a pretty crucial part of the design / we've already comprised a bit there going from 2-lines to one.

PR #1814, opened 2023-06-20
FIX: content flow margins

ianjon3s:
PR fixes a visual regression that was causing elements to not inherit the `margin-top` added with a wildcard selector.

The regression appears to have been caused by the move to `styled` as a method, which has changed the specificity of styles in the CSS output. Any elements that bring default styling (in this case `margin: 0` for `<button>` and `<p>`) are overriding the selector used on the container with these styles.

To fix this I have created the margin as a global property that can be applied to elements (`<button>`, `<p>` etc) that carry their own default styling.


DafyddLlyr (approved):

DafyddLlyr:
Thanks very much for this and for the explanation here 💡

PR #1815, opened 2023-06-20
UPDATE: tidy up styles

ianjon3s:


editor.planx.uk/src/@planx/components/Notice/Public.tsx
```diff
@@ -112,6 +115,7 @@ const NoticeComponent: React.FC<Props> = (props) => {
             size="large"
             type="submit"
             onClick={props.resetPreview}
+            sx={{ ...contentFlowSpacing(theme) }}
```
DafyddLlyr, line 118:
```suggestion
            sx={contentFlowSpacing}
```

This is a bit more terse, and allows us to avoid an extra import of `useTheme()`.

We can pass a callback, `functionName` which is just a shorthand for `(x) => functionName(x)`. 

The `sx` prop allows us to natively use the theme in callbacks so in this example `sx={(theme) => contentFlowSpacing(theme) }` is the same as `sx={contentFlowSpacing}`

editor.planx.uk/src/@planx/components/Question/Public/Question.tsx
```diff
@@ -89,7 +90,7 @@ const Question: React.FC<IQuestion> = (props) => {
         definitionImg={props.definitionImg}
         img={props.img}
       />
-      <FormControl sx={{ width: "100% " }}>
+      <FormControl sx={{ ...contentFlowSpacing(theme), width: "100%" }}>
```
DafyddLlyr, line 93: In this example, we can't use the shorthand version as we're using the [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) (`...`) to also add in the `width` property.

DafyddLlyr (approved):

editor.planx.uk/src/@planx/components/Notice/Public.tsx
ianjon3s: Thanks @DafyddLlyr, will update.

editor.planx.uk/src/@planx/components/Question/Public/Question.tsx
ianjon3s: Appreciate this context 👍

PR #1816, opened 2023-06-20
chore: Rename MFU to "Upload and label"

DafyddLlyr:


editor.planx.uk/src/@planx/components/types.ts
```diff
@@ -14,7 +14,7 @@ export enum TYPES {
   AddressInput = 130,
   ContactInput = 135,
   FileUpload = 140,
-  MultipleFileUpload = 145,
+  FileUploadAndLabel = 145,
```
jessicamcinchak, line 17: I think this will require a corresponding planx-core change, right?

jessicamcinchak (approved):

editor.planx.uk/src/@planx/components/types.ts
DafyddLlyr: https://github.com/theopensystemslab/planx-core/pull/45 Here we go 😄

PR #1820, opened 2023-06-21
feat: Granular MFU analytics

DafyddLlyr:
## What does this PR do?
 - Add analytics metadata each time the help text on a fileType is clicked

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
@@ -9,11 +9,13 @@ import { Store, useStore } from "./store";
 export type AnalyticsType = "init" | "resume";
 type AnalyticsLogDirection = AnalyticsType | "forwards" | "backwards";
 
+export type HelpClickMetadata = Record<string, string>;
+
 let lastAnalyticsLogId: number | undefined = undefined;
 
 const analyticsContext = createContext<{
   createAnalytics: (type: AnalyticsType) => Promise<void>;
-  trackHelpClick: () => Promise<void>;
+  trackHelpClick: (metadata?: HelpClickMetadata) => Promise<void>;
   node: Store.node | null;
 }>({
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -9,11 +9,13 @@ import { Store, useStore } from "./store";
 export type AnalyticsType = "init" | "resume";
 type AnalyticsLogDirection = AnalyticsType | "forwards" | "backwards";
 
+export type HelpClickMetadata = Record<string, string>;
```
DafyddLlyr, line 12: This wasn't exactly what I had wanted here initially - I thought something like this would be a bit clearer - 

```ts
export interface HelpClickMetadata {
  fn: string;
  name: string;
}
```

And we could then keep an array of these in the `metadata` column.

The native `_append` in Hasura is a little limited, and I couldn't simply do this without either - 
 - always overwriting the previous value
 - tracking `HelpClickMetadata` in state locally and modifying this object to overwrite the previous value
 
I've gone for this as it's the simplest option for now.

DafyddLlyr: Docs: https://hasura.io/blog/postgres-json-and-jsonb-type-support-on-graphql-41f586e47536/

jessicamcinchak (commented):
Looking good but hitting one bug: 

Testing on a pizza (https://1820.planx.pizza/braintree/file-test/preview) and can see my per-file help interactions being recorded in `metadata` as expected. But, console throws this error when I click help associated with the question header! If i do that first (before any per-file help clicks), it doesn't toggle `has_clicked_help` to `true` like I'd expect.
Screenshot from 2023-06-21 13-32-40

Can you recreate?

DafyddLlyr:
Thanks for spotting this - I'll take a look!

jessicamcinchak (approved):

PR #1823, opened 2023-06-21
fix: Account for granularity when matching rules

DafyddLlyr:


editor.planx.uk/src/@planx/components/FileUploadAndLabel/model.ts
```diff
@@ -159,14 +159,21 @@ const populateFileList = ({
...
-    passport.data?.[rule.fn]?.includes(rule.val)
-  );
+): Boolean => {
+  const passportVal = passport.data?.[rule.fn];
+  if (!passportVal) return false;
+
+  const isExactMatch = passportVal === rule.val;
+  const isArray = Array.isArray(passportVal);
+  const isExactMatchInArray = isArray && passportVal.includes(rule.val);
+  const re = new RegExp(`^${rule.val}(\\..+| $)`);
```
DafyddLlyr, line 172: I think this describes what we want here - the test cases may serve to better illustrate this.

jessicamcinchak: This is a handy piece of regex :bookmark: one day it would cool to split into its own little util function that could be shared between `computePassport()` and here and the store method responsible for auto-answering (`record()` off the top of my head?) - but totally appreciate that's huge !

jessicamcinchak (approved):
Tests are clear & works as expected for me :muscle:

PR #1827, opened 2023-06-22
feat: send BOPS enhanced constraints data and update PlanningConstraint breadcrumb shape

jessicamcinchak:
Relies on https://github.com/theopensystemslab/planx-core/pull/46

**Planning constraints breadcrumb data before:**
- `fn: [...intersectingPassportValues]`
- `_nots: { fn: [...nonIntersectingPassportValues] }`
- `digitalLandRequest: https://planning.data.gov.uk...` - never used afaik, just available as reference

**After:**
- `fn: [...intersectingPassportValues]`
- `_nots: { fn: [...nonIntersectingPassportValues] }`
  - I think we _should_ be able to get rid of this, but it's heavily entwined in `upcomingCardIds()` still and I was about to easily lose an afternoon untangling. So I'll pick it up in another PR as it's no longer used to format what we send to BOPS (see `planx-core` change), but only referenced internally for automations
- `_constraints: [{...gisResponse}, {...roadsResponse}]` 
  - This is the new record of everything checked plus it's metadata and urls to re-fetch via the source or planx APIs. This does _not_ know about what the PlanningConstraint component's `prop.fn` was set to, but I don't think that matters for BOPS or other end consumers?

**Testing with BOPS**:
- I'll generate and share a test payload with BOPS devs that:
  - Populates existing `constraints` key as normal
  - Populates new `constraints_proposed` key with `_constraints` for comparison / review - I think this can be a good first trial of testing the usefulness of our raw internal response with an end consumer in the spirit of them pulling "full fat" data one day
   - https://api.1827.planx.pizza/admin/session/89420b67-21e4-4c1e-96df-5cd0d9394672/bops
- We'll wait to for feedback and if approved, retire old contraints format for the new proposed one 

**Internal testing**:
- Confirm that automations still work as expected for sites with 0 or many constraints: https://1827.planx.pizza/opensystemslab/digital-land-api-search-test/preview
- Double check formatting of other export formats (XML, CSV, application.html, Review page) - but constraints are already omitted from a number of these


jessicamcinchak:
[Send additional A4 information to BOPS](https://trello.com/c/6vTXkdNR/1547-send-additional-a4-information-to-bops)

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
```diff
@@ -30,8 +35,10 @@ function Component(props: Props) {
     (state) => state.computePassport().data?.["property.boundary.site"]
   );
   const { x, y, longitude, latitude, usrn } =
-    useStore((state) => state.computePassport().data?._address) || {};
-  const showGraphError: boolean = !x && !y && !longitude && !latitude;
+    (useStore(
+      (state) => state.computePassport().data?._address
+    ) as SiteAddress) || {};
+  const showGraphError = !x && !y && !longitude && !latitude;
```
DafyddLlyr, line 41: We show the graph error if all of these are missing - should it be if any of them are missing?

DafyddLlyr (approved):
Just one small question!

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
jessicamcinchak: Fair question - I think you're right that `||` makes more sense than `&&` here even though these fields are all required on our minimum address type. Either case should still only happen when: 1) planning constraints isn't positioned after find property, or 2) there's been an error in OS source data :ok_hand:

PR #1832, opened 2023-06-22
feat: Add 'Start new application' option on ITP page

DafyddLlyr:
image


jessicamcinchak (approved):
Is this a work around for not easily being able to access the restart application arrow button in the header? But looks good & this approach works for me :+1:

DafyddLlyr:
> Is this a work around for not easily being able to access the restart application arrow button in the header? But looks good & this approach works for me 👍

So this was just a more quick / pragmatic fix to be perfectly honest. The header "restart" is just a browser page refresh and gives the following alert - 
image

We'd need a custom alert and to route to a custom URL in the header and the effort / reward ratio here seemed a little off.

PR #1835, opened 2023-06-23
feat: add a warning banner to non-production environments

jessicamcinchak:
New councils are onboarding to the Editor and we want to avoid hiccups around creating content on staging https://opensystemslab.slack.com/archives/C5Q59R3HB/p1687518232056569

This banner will display on the following pages: 
Screenshot from 2023-06-23 13-54-58
Screenshot from 2023-06-23 13-48-44
Screenshot from 2023-06-23 13-55-10

### DIFF
--- a/editor.planx.uk/src/components/TestEnvironmentBanner.tsx
+++ b/editor.planx.uk/src/components/TestEnvironmentBanner.tsx
@@ -1,0 +1,42 @@
+import ReportIcon from "@mui/icons-material/Report";
+import Box from "@mui/material/Box";
+import Button from "@mui/material/Button";
+import { styled } from "@mui/material/styles";
+import Typography from "@mui/material/Typography";
+import React, { useState } from "react";
+
+import { getHeaderPadding } from "./Header";
+
+const TestEnvironmentWarning = styled(Box)(({ theme }) => ({
+  display: "flex",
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/components/TestEnvironmentBanner.tsx
```diff
@@ -0,0 +1,42 @@
...
+  color: theme.palette.text.primary,
+  justifyContent: "space-between",
+  alignItems: "center",
+  ...getHeaderPadding(theme),
+}));
+
+const TestEnvironmentBanner: React.FC = () => {
+  const [showWarning, setShowWarning] = useState(true);
+
+  const isTestEnvironment = () => !window.location.href.includes(".uk");
```
jessicamcinchak, line 22: this might be too naive? but i think it should work! 

intended behavior is: 
- do _not_ show on `editor.planx.uk` or a custom domain which should always be `...gov.uk` for now
- do show on `editor.planx.dev`, `...planx.pizza`, and `localhost`

DafyddLlyr: I can't see any issues with this - let's go with it 👌

DafyddLlyr (approved):

PR #1844, opened 2023-06-26
fix: ensure radio buttons display properly on "back" & "change"

jessicamcinchak:
Notes on how to recreate bug in linked Trello ticket.

Now properly updating `useRadioGroup()` state.

### DIFF
--- a/editor.planx.uk/src/@planx/components/Question/Public/Question.tsx
+++ b/editor.planx.uk/src/@planx/components/Question/Public/Question.tsx
@@ -105,6 +105,7 @@ const Question: React.FC<IQuestion> = (props) => {
         <RadioGroup
           aria-labelledby={`radio-buttons-group-label-${props.id}`}
           name={`radio-buttons-group-${props.id}`}
+          value={previousResponseId}
         >
           <Grid
             container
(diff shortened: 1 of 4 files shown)


editor.planx.uk/src/@planx/components/Question/Public/Question.tsx
```diff
@@ -103,53 +103,53 @@ const Question: React.FC<IQuestion> = (props) => {
...
           aria-labelledby={`radio-buttons-group-label-${props.id}`}
           name={`radio-buttons-group-${props.id}`}
+          value={previousResponseId}
         >
           <Grid
             container
             spacing={layout === QuestionLayout.Basic ? 0 : 2}
             alignItems="stretch"
           >
-            {!props.text?.startsWith("Sorry") &&
```
jessicamcinchak, line 112: this is the key change here that triggers reformatting many lines below - "sorry" was used a long time ago to behave like a Notice with "reset", but Alastair confirmed should no longer be around in current content: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1687773842654919

DafyddLlyr (approved):
Thanks for the additional tests and tidy up here 🙌

PR #1852, opened 2023-06-26
fix(analytics): Grant public role `UPDATE` permission on `analytics_logs.metadata` column

DafyddLlyr:


jessicamcinchak (approved):
I always forget that I'm usually signed into the editor when I'm testing this stuff, these column-level permissions are tricky to catch!

PR #1854, opened 2023-06-27
wip: try fetching a subset of flow data when preparing export data

jessicamcinchak:
Mocking up two options here: 
1. Function `get_visited_flow_nodes` returns jsonb and is only accessible via psql commands like:
```sql
select get_visited_flow_nodes('your-session-id');
```
2. Function `get_visited_flow_nodes_tracked` returns one full row of `published_flows` where the `data` column is a filtered subset of all nodes. This function is ["tracked" in Hasura terms](https://hasura.io/docs/latest/schema/postgres/custom-functions/#pg-track-custom-sql-functions), meaning it can be accessed via our GraphQL API like: 
```gql
query MyQuery {
  get_visited_flow_nodes_tracked(args: { session_id: "your-session-id" }) {
    data
  }
}
```

Todo: 
- Decide which format works best for our use case / are there any performance tradeoffs? adjust migrations accordingly
- Does the `select` need to be revised at all? eg published flow within a time frame rather than most recent
- Test with planx-core


jessicamcinchak:
Closing this out after dev call chat with conclusions:
- the flow is the source of truth for _order_, not the breadcrumbs, and this data structure alone doesn't necessarily help us solve that easily
- we think we'll get the biggest boost from writing/storing a new sorted, enriched breadcrumb structure that will simplify `sortBreadcrumbs` - that in itself will contain the filtered nodes returned here. this can be stored in a new table that we'll run parallel to our current session data for ~2 months

Alternative SQL for consideration: https://gist.github.com/builditben/3aab86e7fab828b8ddd99f5ba2f4aad1

If we come back to this, note that it:
- Needs to account for flow nodes that match breadcrumb answers as well as keys
- Needs to be ordered (include `_root` and do depth first search/sort againt it?); consider storing as `json` or `array` instead of `jsonb`?

PR #1855, opened 2023-06-27
fix(itp): Handle routing for custom domains

DafyddLlyr:
Tested by adding `localhost` to the list of preview only domains, and setting it as domain for Bucks locally as well - getting the expected results now for both domain types.


editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
```diff
@@ -85,7 +85,15 @@ const InviteToPayForm: React.FC<InviteToPayFormProps> = ({
...
-  const navigation = useNavigation();
+  const { navigate } = useNavigation();
+  const {
+    data: { mountpath },
+  } = useCurrentRoute();
+
+  // Scroll to top when loading component
+  useEffect(() => {
+    window.scrollTo(0, 0);
+  }, []);
```
DafyddLlyr, line 96: Small QoL improvement when loading this page!

jessicamcinchak (approved):

PR #1858, opened 2023-06-27
feat: Login screen style update

ianjon3s:
PR adds a styled login screen.

Also updates default theme colour to #0010A4

### DIFF
--- a/editor.planx.uk/src/pages/Login.tsx
+++ b/editor.planx.uk/src/pages/Login.tsx
@@ -4,14 +4,14 @@ import { styled } from "@mui/material/styles";
 import Typography from "@mui/material/Typography";
 import React from "react";
 
-const Wrapper = styled(Box)(() => ({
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/pages/Login.tsx
```diff
@@ -1,15 +1,51 @@
...
+  width: "100vw",
+  height: "100vh",
+  lineHeight: "1.333",
+  display: "flex",
+  alignItems: "center",
+  justifyContent: "center",
+  background: theme.palette.primary.main,
+}));
+
+const Container = styled(Box)(() => ({
```
DafyddLlyr, line 17: We'll cover this later, but I think what you're describing here is essentially this MUI component https://mui.com/material-ui/react-container/

It might make more sense to use this, and we can maybe make some variants which describe our column layouts which can be shared as opposed to just sharing a style snippet?

DafyddLlyr (approved):
💅 Looking good!

PR #1860, opened 2023-06-27
feat: Styled `OptionButton`

DafyddLlyr:


ianjon3s (commented):
When testing the Pizza link, in the editor the OptionButton outputs without a value (blank bulleted buttons below):

image

DafyddLlyr:
Thanks for the review @ianjon3s - missed that sorry but now pushed up a fix 👍

PR #1865, opened 2023-06-28
chore: add e2e mocks

ghost:
Regression tests don't need to hit 3rd party services (BOPS, Uniform, etc) and mocking the responses allows for testing failure scenarios.

Environment variables have been added to allow redirecting traffic to mock submission services in e2e tests. Local authority variables have been grouped (these vars might eventually become team settings). Also a `APP_ENVIRONMENT` variable has been added to determine if notifications should be sent (this may be useful elsewhere too).

This PR adds Smocker which is an HTTP mocking service which can be configured by posting YAML to describe the mock but also has a UI interface to create and inspect mocks:

image


DafyddLlyr (approved):
One small questions (maybe something I've missed) but that's all.

Really nice implementation though - I think it will prove super helpful 👏

PR #1868, opened 2023-06-28
feat: Setup dependabot

DafyddLlyr:
Next steps...
 - [ ] Merge to `main`, test this works as expected
 - [ ] Configure other PlanX repos similarly
 - [ ] Work on updating dependencies so dependabot isn't working through a big backlog bit by bit

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,187 @@
+version: 2
+updates:
+  # Root
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - "root"
+      - "docker"
+    open-pull-requests-limit: 1
+    reviewers:
+      - "theopensystemslab/planx"
+
+  # Editor
+  - package-ecosystem: "npm"
+    directory: "/editor.planx.uk"
+    schedule:
+      interval: "daily"
+    labels:
+      - "editor"
+    open-pull-requests-limit: 1
+    reviewers:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,169 @@
+version: 2
+updates:
+  # Root
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
DafyddLlyr, line 7: I think this will be quite noisy for a while, but then we can bump the `interval` and `open-pull-requests-limit` up

.github/dependabot.yml
```diff
@@ -0,0 +1,181 @@
...
+    labels:
+      - "root"
+      - "docker"
+    open-pull-requests-limit: 1
+    reviewers:
+      - "theopensystemslab/planx"
+
+  # Editor
+  - package-ecosystem: "npm"
```
DafyddLlyr, line 16: This should be set to `npm` when using pnpm

image

jessicamcinchak: I hadn't caught this when updating the map repo today, opening a PR over there shortly to fix :+1:

jessicamcinchak (approved):

.github/dependabot.yml
DafyddLlyr: Yeah my linter flagged it locally and I ignored it initially as being outdated - kind weird!

PR #1880, opened 2023-06-29
fix: Handle radio group state

DafyddLlyr:
This fixes a visual regression when taking the following actions - 
 - Answer a question
 - Go back
 - Attempt to change answer

This journey results in the correct, new, answer being selected but there's no visual feedback for the user (the radio button remains unchecked).

This was something missed as part of #1844 - `previousResponseId` is static and calculated on initial of the component, whilst `formik.values.selected.id` is dynamic and updates as users interact with the radio group.

### DIFF
--- a/editor.planx.uk/src/@planx/components/Question/Public.test.tsx
+++ b/editor.planx.uk/src/@planx/components/Question/Public.test.tsx
@@ -96,17 +96,15 @@ describe("Question component", () => {
 
       expect(screen.getByRole("heading")).toHaveTextContent("Best food");
 
-      let celeryRadio: HTMLElement | undefined;
-      if (QuestionLayout[type] === "Basic") {
-        celeryRadio = screen.getByRole("radio", { name: "Celery" });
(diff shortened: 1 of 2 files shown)


jessicamcinchak (approved):
thanks for catching this !

editor.planx.uk/src/@planx/components/Question/Public.test.tsx
```diff
@@ -96,18 +96,16 @@ describe("Question component", () => {
...
-      }
+      const celeryRadio = screen.getByRole("radio", { name: /Celery/ });
+      const pizzaRadio = screen.getByRole("radio", { name: /Pizza/ });
+
+      // State is preserved...
       expect(celeryRadio).toBeChecked();
 
+      // ...and can be updated
+      await user.click(pizzaRadio);
+      expect(pizzaRadio).toBeChecked();
```
jessicamcinchak, line 107: :raised_hands: perfect, thanks!

PR #1885, opened 2023-06-29
feat: Styled `ExpandableList`

DafyddLlyr:
image

### DIFF
--- a/editor.planx.uk/src/ui/ExpandableList.tsx
+++ b/editor.planx.uk/src/ui/ExpandableList.tsx
@@ -1,37 +1,27 @@
 import MuiButtonBase from "@mui/material/ButtonBase";
+import List from "@mui/material/List";
+import ListItem from "@mui/material/ListItem";
+import { styled } from "@mui/material/styles";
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/ui/ExpandableList.tsx
```diff
@@ -41,20 +31,20 @@ export function ExpandableListItem(props: {
...
   };
 
   return (
-    <li
-      className={classnames(classes.root, props.expanded && classes.expanded)}
+    <ListItem
+      sx={{
+        borderBottom: (theme) => `1px solid ${theme.palette.secondary.main}`,
+        display: "block",
+      }}
```
ianjon3s, line 43: Just missing `padding: "0",` to prevent default padding on `<li>`

ianjon3s (approved):

editor.planx.uk/src/ui/ExpandableList.tsx
DafyddLlyr: Very good spot thank you 🏅

PR #1890, opened 2023-06-29
fix: Incorrect MUI classes syntax

DafyddLlyr:
Fixes https://trello.com/c/Edty1cU8/2502-text-in-text-input-showing-outside-of-box

It appears that under certain conditions (possibly when the parent is not a styled component?) the syntax for applying classes cannot have a space. `$ .classname` is invalid but `$.classname` is.


DafyddLlyr:
[Text in 'text input' showing outside of box](https://trello.com/c/Edty1cU8/2502-text-in-text-input-showing-outside-of-box)

jessicamcinchak (approved):

PR #1901, opened 2023-07-03
fix: paper drawer width

ianjon3s:
Over the weekend I noticed the width on the help drawer dialog had stopped applying, and tracked it down to a regression introduced in https://github.com/theopensystemslab/planx-new/pull/1890

This PR reverts that and adds a `max-width` to simplify and prevent overlap at smaller viewports.


DafyddLlyr (approved):
Thanks for catching this!

PR #1902, opened 2023-07-03
docs(adr): Remove `GovUkPayment` from session data

DafyddLlyr:
A few thought on our recent conversation around restructuring/enriching breadcrumbs.


DafyddLlyr:
> The lowcal session table has a one-to-many relationship with payment statuses so we might want to write a stored procedure (or planx-core request function) to find the entry related to a successful payment (or just find the latest).

Yep - this is a good shout - generally I think we always need this alongside the session data, so we'll need to add this relationship to Hasura (currently it's just one way, `payment_status` -> `lowcal_session`). 

> Do downstream services need to know about the GovPay reference? Would it be a problem to stop sending this reference to BOPS?

The GovPay ID is added to the BOPS payload as `payment_reference` which we can get from `payment_status`, and actually isn't included in the debug data so I'll update the doc to remove this.

doc/architecture/decisions/0006-remove-govukpayment-from-session-data.md
```diff
@@ -0,0 +1,40 @@
...
+This object can then be stored temporarily in the application's memory during the user's active session (Zustand store).
+
+## Consequences
+
+### Required code changes
+We will need to make a series of changes to how we currently interact with this object, listed below - 
+
+| Interaction | Context | Required change | Consequence |
+|--|--|--|--|
+| Pay component | `GovUkPayment` is required here to get status and `next_url` links | On component load, fetch GovPay ID via audit table and query GovPay API. This could be wrapped in a `planx-core` function `session.get`GovUkPayment`(sessionId)`. <br/><br/> We would need to handle Hasura permissions here to ensure a public user can access `payment_status.gov_pay_id` for their session ([Hasura docs](https://hasura.io/docs/latest/auth/authorization/permissions/row-level-permissions/#relationships-in-permissions)) | Increased load time on pay component for additional queries |
```
jessicamcinchak: Question: in this in-flow Pay scenario, are we still mutating session data upon successful payment? Not to store the `GovUkPayment` object, but to update the session `breadcrumbs` to reflect the Pay node? 

Is it worth noting any specific mutations for this scenario as well as queries?

jessicamcinchak (approved):
This looks good & I think simplifies a lot :100: 

One big picture question / clarification point: once _implemented_, will existing "open" sessions & new ones both work? 

I think so, right? Since `payment_status` table already exists, the only possible exception I can think of is maybe needing a fallback value for looking up the payment amount (eg if not included in payment_status table yet, then by passport variable)?

DafyddLlyr:
> One big picture question / clarification point: once _implemented_, will existing "open" sessions & new ones both work?

Yes, open sessions will still work as is because...
 - All changes listed above just need a payment status, which is already in the audit table for all sessions
 - The exception is submit, which also needs an amount. Submit events happen very shortly after pay events - there should be no open sessions which have been paid but not yet submitted (i.e. without a payment amount in the audit table). 
 
Just as a precaution, I'll make a small PR which adds payment amount to the audit table first and we can get that merged and on prod before any other changes.

doc/architecture/decisions/0006-remove-govukpayment-from-session-data.md
DafyddLlyr: I'll add this in to be explicit 👍 

In this scenario the Pay component would still generate a breadcrumb exactly as it currently does, it would simply not be adding the additional `session.data.govUkPayment` to the passport.

PR #1909, opened 2023-07-04
feat: Boilerplate next steps component

ianjon3s:
PR adds the boilerplate setup for the Next Steps component, based on the work of Jess https://github.com/theopensystemslab/planx-new/pull/1448/files

Bits I'm not sure about:
- Classification of component, I've added under "Question" for now
- Numbering system used in editor.planx.uk/src/@planx/components/types.ts, I've added to the end of the list with sequential number

Pizza preview here:
https://1909.planx.pizza/testing/next-steps-test/preview?analytics=false

### DIFF
--- a/editor.planx.uk/src/@planx/components/NextSteps/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/NextSteps/Public.tsx
@@ -1,0 +1,25 @@
+import Box from "@mui/material/Box";
+import Typography from "@mui/material/Typography";
+import { PublicProps } from "@planx/components/ui";
+import React from "react";
+
+import Card from "../shared/Preview/Card";
+import QuestionHeader from "../shared/Preview/QuestionHeader";
+import { NextSteps } from "./model";
+
+type Props = PublicProps<NextSteps>;
+
+export default Component;
+
+function Component(props: Props) {
+  return (
+    <Card handleSubmit={props.handleSubmit} isValid>
+      <QuestionHeader title={props.title} description={props.description} />
+      <Box>
+        <Typography variant="h2" style={{ color: "orangered" }}>
+          UNDER DEVELOPMENT
+        </Typography>
+      </Box>
+    </Card>
+  );
+}
--- a/editor.planx.uk/src/@planx/components/NextSteps/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/NextSteps/Public.tsx
@@ -14,7 +14,7 @@ export default Component;
 function Component(props: Props) {
   return (
     <Card handleSubmit={props.handleSubmit} isValid>
-      <QuestionHeader title={props.title} description={props.description} />
+      <QuestionHeader {...props} />
       <Box>
         <Typography variant="h2" style={{ color: "orangered" }}>
           UNDER DEVELOPMENT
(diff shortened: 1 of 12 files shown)


editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
```diff
@@ -46,6 +46,7 @@ const NodeTypeSelect: React.FC<{
       <optgroup label="Question">
         <option value={TYPES.Statement}>Question</option>
         <option value={TYPES.Checklist}>Checklist</option>
+        {/* <option value={TYPES.NextSteps}>Next steps</option> */}
```
jessicamcinchak, line 49: This grouping makes sense to me too :heavy_plus_sign:

As long as the Public display shows "Under development", I'd actually be in favor of this _not_ being commented out in this list so we can easily test/preview on the pizza! And that should be enough of a signal to Editors to not use it yet

editor.planx.uk/src/@planx/components/types.ts
```diff
@@ -29,4 +29,5 @@ export enum TYPES {
   Send = 650,
   Calculate = 700,
   Confirmation = 725,
+  NextSteps = 730,
```
jessicamcinchak, line 32: This works :heavy_plus_sign: There's no obvious method here, but as long as it's a new unique number it'll be good to go! 

One new thing the original docs don't cover I think is that this list of Component types is now also maintained in planx-core and will need to be updated here as well: https://github.com/theopensystemslab/planx-core/blob/main/src/types/component.ts

editor.planx.uk/src/@planx/components/NextSteps/Public.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { NextSteps } from "./model";
+
+type Props = PublicProps<NextSteps>;
+
+export default Component;
+
+function Component(props: Props) {
+  return (
+    <Card handleSubmit={props.handleSubmit} isValid>
+      <QuestionHeader title={props.title} description={props.description} />
```
jessicamcinchak, line 17: A more flexible option is to use `<QuestionHeader {...props} />` here which will then also automatically pick up help text if it's defined. Otherwise you'll have to keep adding those prop fields manually like `howMeasured={props.howMeasured}` and so on

jessicamcinchak (commented):
Looking very solid :tada: 

A few small comments, but no show stoppers - this'll be ready to merge soon !

editor.planx.uk/src/@planx/components/types.ts
ianjon3s: Thanks, I've added as a new PR to planx-core:
https://github.com/theopensystemslab/planx-core/pull/53

editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
ianjon3s: Thanks, I've updated so this is unhidden, also updated the description with a pizza preview link for the component.

editor.planx.uk/src/@planx/components/NextSteps/Public.tsx
ianjon3s: Thanks, updated.

DafyddLlyr (approved):

jessicamcinchak (approved):
:tada:

PR #1910, opened 2023-07-04
chore: update Storybook components (Inputs & Calculate)

jessicamcinchak:
Opening for review / sense check before I tackle the whole list (incremental PRs feel best here for fewer file changes at once)!

This updates stories for all "input" type components plus "Calculate" because I wanted to try one "editor-only" story which we typically haven't done in the past. Feedback welcome about any organizational/formatting choices I've making so far please :slightly_smiling_face: 

**General changes:**
- Auto-migrated all story files to CSF3 by running `npx storybook@latest migrate csf-2-to-3 --glob="**/*.stories.tsx"` ([docs here](https://storybook.js.org/blog/storybook-csf3-is-here)) - still requires a bit of manual cleanup as I reach them
- Added dependency `@storybook/testing-library` so we can use the `play` function to automatically fill out forms and trigger validation errors and more 
- Updated "With Editor" `Wrapper` styles to better mimic actual editor modal / not blend into Storybook background etc
- Renaming `Combined.stories.tsx` or other to `{ComponentName}.stories.tsx` as I reach that file

**Working through all components:** 
- [x] AddressInput
- [x] Calculate
- [ ] Checklist
- [ ] Confirmation
- [x] ContactInput
- [ ] Content
- [x] DateInput
- [ ] DrawBoundary
- [ ] ExternalPortal
- [ ] FileUpload
- [ ] FileUploadAndLabel
- [ ] Filter
- [ ] FindProperty
- [ ] InternalPortal
- [ ] Notice
- [x] NumberInput
- [ ] Pay
- [ ] PlanningConstraints
- [ ] PropertyInformation
- [ ] Question
- [ ] Result
- [ ] Review
- [ ] Section
- [ ] Send
- [ ] SetValue
- [ ] TaskList
- [x] TextInput


jessicamcinchak:

editor.planx.uk/src/@planx/components/Calculate/Calculate.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    <>
+      <Wrapper Editor={Editor} Public={Public} />
+      <WarningContainer>
+        <ErrorOutline />
+        <Typography variant="body2" ml={2}>
+          This component is only available in the Editor when designing
+          services, it does <strong>not</strong> display in the Public form.
+        </Typography>
+      </WarningContainer>
```
jessicamcinchak, line 26: In editor-only cases, there won't be any user feedback when you click "Update" on the `<Wrapper />` unless there's a validation error in the modal itself. I think manually adding our existing `<WarningContainer />` here is a simple solution to communicate expectations?

editor.planx.uk/src/@planx/components/Calculate/Public/index.tsx
```diff
@@ -24,5 +24,5 @@ export default function Component(props: Props) {
     });
   }, []);
 
-  return <p>Calculating…</p>;
+  return null;
```
jessicamcinchak, line 27: `null` is now more consistent with other "hidden" components like `SetValue`; rendering an un-Card-formatted "Calculating..." was a bit confusing in Storybook etc

editor.planx.uk/src/@planx/components/AddressInput/AddressInput.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+import React from "react";
+
+import Wrapper from "../fixtures/Wrapper";
+import Editor from "./Editor";
+import Public from "./Public";
+
+export default {
+  title: "PlanX Components/AddressInput",
+  component: Public,
+} as Meta;
```
jessicamcinchak, line 12: This file results in these stories, which is generally the pattern I've followed for all inputs: 
Screenshot from 2023-07-05 13-45-02

`Filled Form` will automatically trigger validation errors. I think with another addon or two we could directly incorporate `expect` testing in here in the future too, but I haven't done that yet - and those types of scenarios are already typically tested via Jest in `Public.test.tsx` for each component. 

`With Editor` renders a `Wrapper` that allows you to fill out the editor modal, click "Update", then render the live "Preview" below. Storybook 7's new UI felt better suited to top-bottom layout here, but an easy one to come back to if side-by-side is preferred?

ianjon3s:
The updated component preview here is brilliant! Seeing the component in a 'frame' is a big improvement. Also the zoom controls are more prominent (I hadn't noticed they were there before) and useful in the isolated frame.

editor.planx.uk/src/@planx/components/AddressInput/AddressInput.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+export const EmptyFormWithHelpText: StoryObj = {
+  args: {
+    ...EmptyForm.args,
+    howMeasured: "This is an example definition.",
+  },
+};
+
+export const FilledForm: StoryObj = {
```
DafyddLlyr, line 28: This raises an interesting question about who Storybook is for / what do we want to get out of it (doesn't need an answer right away!).

I'd be inclined to see this more as a design example of error handling / error state, and then use the `play` API for more complex components that would maybe benefit from this visual demonstration (e.g. Upload and label, draw boundary). Appreciate this is just a demo of what v7 can do, but these are questions it would be good to cover on dev calls 👍

DafyddLlyr: On the `expect()` point - lets look into this properly 🕵️ 

If Storybook can give us a more robust environment for a11y and component level unit testing then that would be great.

editor.planx.uk/src/@planx/components/AddressInput/AddressInput.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+import Wrapper from "../fixtures/Wrapper";
+import Editor from "./Editor";
+import Public from "./Public";
+
+export default {
+  title: "PlanX Components/AddressInput",
+  component: Public,
+} as Meta;
+
+export const EmptyForm: StoryObj = {
```
DafyddLlyr, line 14: What the TS support like here? Do we have an option to do anything like `StoryObj<AddressInputProps>` or maybe `StoryObj<typeof AddressInputProps>` for additional safety / future proofing here?

If there's a way of doing this it could be a neat way of ensuring we keep our stories up to date when component inputs and outputs change.

Happy for this to be a follow on task or I can pick it up also 👍

DafyddLlyr (approved):
This is awesome, really good to get things back in shape here 🥇 

A few questions on the code, just about types and how we plan on using this - this can follow on I think.

In terms of remaining stories / tasks, is there a naming convention we can maybe use to track what's outstanding? e.g. `button.stories.deprecated.tsx` or something?

editor.planx.uk/src/@planx/components/AddressInput/AddressInput.stories.tsx
jessicamcinchak: TS support & more proactive typing is as good question here - there's some docs in this direction here (be careful to sift out CSF3 format though which we're on now): https://storybook.js.org/docs/react/api/csf#args-story-inputs

I can update to this: 
```
export const EmptyForm: StoryObj<AddressInput> = {
  args: {
    title: "Enter your address",
    description: "It might be where you live",
  },
};
```

But, `args` property has type `args?: Partial<AddressInput> | undefined` - so it unfortunately doesn't complain or throw any errors/warnings if I, for example, remove `title` which we say is required.

One to keep thinking on?

editor.planx.uk/src/@planx/components/AddressInput/AddressInput.stories.tsx
jessicamcinchak: Yep - one for dev calls - I see this chat & decision around testing boundaries as it's own set of tasks ! Trying to keep focus to "a working, updated story file per component" here for starters.

Here's a nice example of `play` though that goes a couple steps further than I have: https://storybook.js.org/docs/react/api/csf#play-function (I've held off on lots of new dependencies for now, this would require at least `@storybook/jest` too)

editor.planx.uk/src/@planx/components/AddressInput/AddressInput.stories.tsx
DafyddLlyr: Ah I see - lets come back to this then - happy with it as it just now! 🙂

PR #1911, opened 2023-07-05
chore: Upgrade pnpm to v8.2.0

DafyddLlyr:
- Upgrades pnpm to v8.2.0
 - Removes pnpm engine
 
CI currently failing due to dependencies on https://github.com/theopensystemslab/planx-document-templates/pull/20 & https://github.com/theopensystemslab/planx-core/pull/51 which need to be merged first and commit SHAs updated here.

PR #1918, opened 2023-07-06
chore: Drop 'engines' from package.json

DafyddLlyr:
Incremental change to get us closer to resolving https://github.com/theopensystemslab/planx-new/pull/1911

I think it's wise to drop this for a few reasons - 
 - It makes managing version numbers more work (especially compared with repository variables)
 - We're not really tied to a version number in a totally hard way (the repo will probably be fine on Node 17)
 - We already have `.nvmrc` in the project root which essentially handles this for us, and details in the readme of required versions
  - It's potentially causing issues when upgrading across repositories


jessicamcinchak (approved):
Looks okay to me, seems to be working as expected !

PR #1923, opened 2023-07-06
chore: Use repo vars in GitHub actions

DafyddLlyr:
Version numbers not updated, just where we get them from.

This is managed in https://github.com/theopensystemslab/planx-new/settings/variables/actions

It seems like a sensible option to simplify things a little, and make it a bit easier to reliably bump version numbers in future.

image


DafyddLlyr:
There is probably scope for more "repository variables" for GitHub actions, but going with just these for now.

jessicamcinchak (approved):

PR #1924, opened 2023-07-06
chore: update Storybook components (FindProperty, PropertyInformation, DrawBoundary, PlanningConstraints)

jessicamcinchak:
Follows on from #1910


jessicamcinchak:

editor.planx.uk/src/@planx/components/PropertyInformation/PropertyInformation.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+import Editor from "./Editor";
+import { Presentational, PresentationalProps } from "./Public";
+
+export default {
+  title: "PlanX Components/PropertyInformation",
+  component: Presentational,
+} as Meta;
+
+// Would be nice to import/share these default props between tests & stories, but Storybook will throw error "jest is not defined" on overrideAnswer
```
DafyddLlyr, line 14: It's not perfectly neat but could we do something like this, which may express and expose the issue we're facing here.

```ts
const defaultPresentationalProps: PresentationalProps = merge(
   objectUsedByJestTests,
   { overrideAnswer: () => console.log("test") }
   // or make an object if there are multiple overrides maybe?
   // objectWithStorybookOverrides,
 )
```

DafyddLlyr (commented):
Sorry just seen this is still a draft!

editor.planx.uk/src/@planx/components/PropertyInformation/PropertyInformation.stories.tsx
jessicamcinchak: Addressed now, both tests & stories pulling from same mock file and each setting appropriate `overrideAnswer` :+1:

editor.planx.uk/src/@planx/components/FindProperty/Public/FindProperty.stories.tsx
```diff
@@ -1,27 +1,65 @@
...
   parameters: {
-    reactNavi: {
-      useCurrentRoute: () => {
+    swr: {
+      default: () => {
         return {
-          data: {
-            team: "canterbury",
-          },
+          data: localAuthorityMock,
```
jessicamcinchak, line 19: This `swr` mock is currently unused - I _think_ it's actually mocked correctly, but the address-autocomplete doesn't render in storybook because it's a custom web component so it's never actually called. Might be interesting to try and come back to these cases, but I wasn't quite ready to get into the react vs web component webpack settings in storybook config here just yet.

For components relying on web components, I've made a note in their descriptions & linked to the Pitsby docs for now

DafyddLlyr (approved):

PR #1936, opened 2023-07-07
feat: allow file upload and label to be used in an info-only mode

jessicamcinchak:
Editor gets a toggle option that is `false` by default:
Screenshot from 2023-07-07 10-32-36

Preview shows list only & "Continue" is always enabled, only writes breadcrumb on submit: 
Screenshot from 2023-07-07 10-36-13

Other changes:
- Updated test suite for "info-only mode"
- Upload & label is _only_ shown on the Review page SummaryList when the drop zone is available, otherwise it's omitted which is consistent with Content & Notice type components

Future PRs: 
- Story file ! This needs to be updated for regular FileUpload too, so I'll address in a followup PR


jessicamcinchak:
[Show 'upload and label' as list to user (without needing to upload)](https://trello.com/c/OhTk1o9R/2504-show-upload-and-label-as-list-to-user-without-needing-to-upload)

editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
```diff
@@ -118,8 +117,16 @@ function SummaryListsBySections(props: SummaryListsBySectionsProps) {
     const isPresentationalComponent = Boolean(Component);
     const doesNodeExist = Boolean(props.flow[nodeId]);
     const isAutoAnswered = userData.auto;
+    const isInfoOnlyMode =
+      node.type === TYPES.FileUploadAndLabel &&
+      props.flow[nodeId].data?.hideDropZone;
 
-    return doesNodeExist && !isAutoAnswered && isPresentationalComponent;
+    return (
```
DafyddLlyr, line 124: Good thought to check here 👌

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.test.tsx
```diff
@@ -28,6 +28,63 @@ describe("Basic state and setup", () => {
...
+        howMeasured="This is sample help text for the whole component"
+      />
+    );
+
+    const helpIcons = screen.getAllByTestId("more-info-icon");
+    expect(helpIcons).toHaveLength(2);
+  });
+});
+
+describe("Info-only mode with hidden drop zone", () => {
```
DafyddLlyr, line 67: 👍

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
```diff
@@ -130,19 +130,26 @@ function Component(props: Props) {
 
   return (
     <Card
-      handleSubmit={handleSubmit}
-      isValid={slots.every((slot) => slot.url && slot.status === "success")}
+      handleSubmit={props.hideDropZone ? props.handleSubmit : handleSubmit}
```
DafyddLlyr, line 133: This reads a little funny (but is perfectly correct!)

This is because `handleSubmit()` is not a very accurate / descriptive name - maybe we need something more meaningful like `validateAndSubmit()`. We could also destructure `props.handleSubmit` to a more meaningful name (maybe overkill?).

`handleSubmit={props.hideDropZone ? continueWithoutUploads : validateUploadsAndSubmit}`

DafyddLlyr (changes requested):
Perfect - simple and efficient solution 👏 

Hit one bug which I've not quite pinpointed in the code but can't recreate on staging - 

 - Create upload and label component with rules, set to info-only mode
 - Save
 - Edit node
 - Rules are missing ❌

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
jessicamcinchak: Totally agree this is awkward - renaming the custom `handleSubmit` we've defined in the component to `validateAndSubmit()` makes most sense to me :+1: 

For the sake of global search-ability, I think it's actually ideal _not_ to destructure/rename props.handleSubmit to something else!

DafyddLlyr: Yep - I agree here, this works!

jessicamcinchak:
@DafyddLlyr thanks for taking a look here - I can re-create that missing rules bug, it's a strange one. 

I thought at first it was only happening when an existing component was getting updated - so when the new prop `hideDropZone` was going from _undefined_ to defined and causing a hard refresh of everything. But that doesn't seem to hold and I can recreate on any change of that toggle. On a similar thread, I tried making `hideDropZone` required in the yup schema, but that doesn't seem to make a difference either.

Especially strange because the rules still exist at point of clicking "update" on the component modal in the editor, otherwise you'd get the required validation error. So my current best guess is that the Public component is unexpectedly overwriting/resetting `props.fileTypes` whenever `props.hideDropZone` updates? I'll keep thinking on this one, but I haven't isolated it yet!

DafyddLlyr:
If you want another pair of eyes on it just let me know 👍

jessicamcinchak:
Merging this one with known rules bug, since realising it's a wider pre-existing issue & getting addressed over here #1952

DafyddLlyr:
Now addressed in https://github.com/theopensystemslab/planx-new/pull/1958

PR #1937, opened 2023-07-07
feat: Basic validation in 'Upload and label' modal

DafyddLlyr:
https://github.com/theopensystemslab/planx-new/assets/20502206/23b7d2d1-4164-43e7-b40f-e15bab3d1d20

 - Add additional validating of `FileList` on "Done"
 - Still allow users to exit modal with errors (click away or `Esc`) for a11y - validation will catch any issues on "Continue"
 - As with the current component, validation of for the entire state and not "per file" so the warning is at a global level
 - `ErrorWrapper` is moved to a more prominent location on the main component

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
@@ -136,13 +136,11 @@ function Component(props: Props) {
       <QuestionHeader {...props} />
       <DropzoneContainer>
         <FileStatus status={fileUploadStatus} />
-        <ErrorWrapper error={validationError} id={props.id}>
-          <Dropzone
-            slots={slots}
-            setSlots={setSlots}
-            setFileUploadStatus={setFileUploadStatus}
-          />
-        </ErrorWrapper>
+        <Dropzone
+          slots={slots}
+          setSlots={setSlots}
+          setFileUploadStatus={setFileUploadStatus}
+        />
         <List
           disablePadding
           sx={{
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
```diff
@@ -136,13 +136,11 @@ function Component(props: Props) {
...
-            slots={slots}
-            setSlots={setSlots}
-            setFileUploadStatus={setFileUploadStatus}
-          />
-        </ErrorWrapper>
+        <Dropzone
+          slots={slots}
+          setSlots={setSlots}
+          setFileUploadStatus={setFileUploadStatus}
+        />
```
jessicamcinchak, line 143: nit: I actually much prefer the ErrorWrapper wrapping the Dropzone in the particular case of trying to continue before uploading anything because it draws attention to the action needed to proceed. Felt like it aligned well to the govuk guidance here: https://design-system.service.gov.uk/components/file-upload/#error-messages

The floating "Upload at least one file" looks a little odd to me, and if we keep this, I think would technically be a better fit for the govuk [error summary fully-bordered style](https://design-system.service.gov.uk/components/error-summary/) because it's not wrapping any particular element on the page?

But I guess this approach let's you avoid multiple ErrorWrappers that conditionally render based on the specific `validationError` message?

jessicamcinchak (approved):
Works as expected for me with one edge case: 
- Upload a file, "esc" the modal before tagging it
- Click "Continue", render the "Please tag all files" error as expected 
- Click "Change", add a tag in the modal, click "Done"
- Return to main card and error message hasn't been cleared even though my file has a tag now - I can "Continue" as expected so no functional bugs here, just a visual bug that might be confusing!

DafyddLlyr:
Thanks for the feedback - super helpful.

I've pushed up a change which does the following -
 - Adds an error wrapper around the dropzone for dropzone errors
 - Adds an error wrapper around the filelist for filelist errors
 - Clears appropriate errors on modal open, file upload, or file deletion
 
File list errors are still handled globally as opposed to per-file due to how validation is done - we'll see if there's any feedback on this from testing.
 
Demo below! If this works / feels ok let me know and I'll add tests covering the various states.
 

https://github.com/theopensystemslab/planx-new/assets/20502206/a7976b2e-9c9f-406f-9b4e-9cf5e7c4e659

jessicamcinchak:
Feels much better in my opinion :raised_hands:

PR #1946, opened 2023-07-10
chore: Temporarily turn off dependabot for IaC

DafyddLlyr:
### Why I think we should do this...
 - IaC code has numerous dependencies major versions behind
 - Updating and testing these is difficult as it can only be done on staging / production
 - Testing here will be slow and could hold up CI pipelines whilst we work issues out

 
### I think a better approach here would be - 
 - Close open dependabot infra PRs
 - Document and use Sandbox environment (https://trello.com/c/H3mGuTjL/2497-document-and-turn-off-aws-sandbox-environment)
 - Update dependencies as it's own task, test on Sandbox (https://trello.com/c/U2Lmt4oK/2280-upgrade-infrastructure-dependencies)
 - Once looking good, merge to staging
 - Turn dependabot back on

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -127,53 +127,53 @@ updates:
       - "theopensystemslab/planx"
   
   # Infrastructure
-  - package-ecosystem: "npm"
-    directory: "/infrastructure/application"
-    schedule:
-      interval: "daily"
-    labels:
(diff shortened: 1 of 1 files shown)


jessicamcinchak (approved):
This makes sense to me !

PR #1948, opened 2023-07-10
chore: TS annotations for Storybook v7

DafyddLlyr:
Follow up from conversation here - https://github.com/theopensystemslab/planx-new/pull/1910#discussion_r1254053772

I did a bit of reading up on Storybook v7 and TypeScript integration and this pattern looks like it works. We end up inheriting the expected props without everything being optional or having `| undefined` appended to its types. Docs / blog post here - https://storybook.js.org/blog/improved-type-safety-in-storybook-7/

Might be worth pulling this branch down @jessicamcinchak and just double checking that this works as expected.

If so I'm happy to go through and update v7 stories, or it can be picked up on your branch - whatever works 🙂

### DIFF
--- a/editor.planx.uk/src/@planx/components/AddressInput/AddressInput.stories.tsx
+++ b/editor.planx.uk/src/@planx/components/AddressInput/AddressInput.stories.tsx
@@ -5,27 +5,32 @@ import React from "react";
 import Wrapper from "../fixtures/Wrapper";
 import Editor from "./Editor";
 import Public from "./Public";
+import AddressInputComponent from "./Public";
 
-export default {
+const meta = {
   title: "PlanX Components/AddressInput",
   component: Public,
-} as Meta;
+} satisfies Meta<typeof AddressInputComponent>;
 
-export const EmptyForm: StoryObj = {
+type Story = StoryObj<typeof meta>;
+
+export default meta;
+
+export const EmptyForm = {
   args: {
     title: "Enter your address",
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/@planx/components/AddressInput/AddressInput.stories.tsx
```diff
@@ -5,27 +5,32 @@ import React from "react";
 import Wrapper from "../fixtures/Wrapper";
 import Editor from "./Editor";
 import Public from "./Public";
+import AddressInputComponent from "./Public";
```
jessicamcinchak, line 8: nit: `AddressInputComponent` should be the same as `Public` on the line above in this case? We deal with this a bit inconsistently across Public component files and test/story files, so happy to go with whichever you think is most clear here as long as we only import/use one or the other!

jessicamcinchak (approved):
Thanks for this, works as expected for me locally :raised_hands: 

I'd say let's merge this then follow this pattern for all future Storybook changes / update as we go

editor.planx.uk/src/@planx/components/AddressInput/AddressInput.stories.tsx
DafyddLlyr: Thanks for spotting this - you're right and I'll remove this duplicated import.

I think `import Public from "./Public";` will be the more consistent pattern here.

PR #1957, opened 2023-07-11
feat: Next steps component build

ianjon3s:
PR adds a new component named Next Steps.

I've constructed this using examples from other components. 

Bits I'm not sure on so far:
- The component does not have a continue button and I've tried to set this up accordingly, using the Confirmation component as an example.
- ~~The URL field is currently a text field, I'm expecting this should have a `type="url"` property added with validation, but wasn't sure how to bring this in.~~ update: added `type="url"`
- Model.ts has the commented text `/* remove once migrated */`, which was copied over from another file, does this need addressing? https://github.com/theopensystemslab/planx-new/pull/1957/files#diff-99e03f9ab40b152de3dbfa281c5ae664df487266dee9594485d388d0b4ae221cR18

Working example:
https://1957.planx.pizza/lambeth/next-steps-component/preview

Storybook example:
https://storybook.1957.planx.pizza/?path=/docs/planx-components-nextsteps--docs

### DIFF
--- a/editor.planx.uk/src/@planx/components/NextSteps/NextSteps.stories.tsx
+++ b/editor.planx.uk/src/@planx/components/NextSteps/NextSteps.stories.tsx
@@ -1,0 +1,43 @@
+import { Meta, StoryObj } from "@storybook/react";
+import React from "react";
+
+import Wrapper from "../fixtures/Wrapper";
+import Editor from "./Editor";
+import type { Props as PublicProps } from "./Public";
+import Public from "./Public";
+
+const metadata: Meta = {
+  title: "PlanX Components/NextSteps",
+  component: Public,
+  argTypes: {
+    handleSubmit: { control: { disable: true }, action: false },
+  },
+};
+
+export const Frontend: StoryObj<PublicProps> = {
+  args: {
+    title: "What would you like to do next?",
+    description: "The decision is yours.",
+    steps: [
+      {
+        title: "First option",
+        description: "Get advice before submitting continuing your project.",
+        url: "https://www.planx.uk",
+      },
+      {
+        title: "Second option",
+        description: "Find out how to submit an application.",
+        url: "https://www.planx.uk",
+      },
+      {
+        title: "Third option",
+        description: "Your progress will be automatically saved.",
+        url: "https://www.planx.uk",
+      },
+    ],
+  },
+};
+
+export const WithEditor = () => <Wrapper Editor={Editor} Public={Public} />;
+
+export default metadata;
--- a/editor.planx.uk/src/@planx/components/NextSteps/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/NextSteps/Public.tsx
@@ -1,5 +1,3 @@
-import Box from "@mui/material/Box";
-import Typography from "@mui/material/Typography";
 import { PublicProps } from "@planx/components/ui";
 import React from "react";
 import NextStepsList from "ui/NextStepsList";
--- a/editor.planx.uk/src/@planx/components/NextSteps/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/NextSteps/Public.tsx
@@ -20,7 +20,7 @@ const NextStepsComponent: React.FC<Props> = (props) => {
         policyRef={props.policyRef}
         howMeasured={props.howMeasured}
       />
-      <NextStepsList items={props.steps} heading={"h2"} />
+      <NextStepsList items={props.steps} />
     </Card>
   );
 };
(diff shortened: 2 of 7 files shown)


editor.planx.uk/src/@planx/components/NextSteps/NextSteps.stories.tsx
```diff
@@ -0,0 +1,43 @@
+import { Meta, StoryObj } from "@storybook/react";
+import React from "react";
+
+import Wrapper from "../fixtures/Wrapper";
+import Editor from "./Editor";
+import type { Props as PublicProps } from "./Public";
+import Public from "./Public";
+
+const metadata: Meta = {
```
jessicamcinchak, line 9: :tada: yay stories

jessicamcinchak (commented):
took a quick first read through, looking really good - will come back to more thoroughly soon !

editor.planx.uk/src/@planx/components/NextSteps/Public.tsx
```diff
@@ -1,25 +1,26 @@
...
+const NextStepsComponent: React.FC<Props> = (props) => {
   return (
-    <Card handleSubmit={props.handleSubmit} isValid>
-      <QuestionHeader {...props} />
-      <Box>
-        <Typography variant="h2" style={{ color: "orangered" }}>
-          UNDER DEVELOPMENT
-        </Typography>
-      </Box>
+    <Card>
```
jessicamcinchak, line 13: to answer one of your questions - this line confirms you've set up the component correctly to not have a "Continue" button!

the props passed to the `<Card>` component determine how the "Continue" button is rendered:
- if `handleSubmit` is provided, then show the "Continue" button
- if `isValid` is falsey, then disable the button
- if both props are omitted like here, then no button at all

editor.planx.uk/src/ui/NextStepsList.tsx
```diff
@@ -0,0 +1,97 @@
...
+  width: "50px",
+  height: "50px",
+  flexShrink: "0",
+}));
+
+interface Item {
+  title: string;
+  description: string;
+  url: string;
+}
```
jessicamcinchak, line 60: nit: this typescript interface looks to be the same as `Step` defined in `@planx/components/NextSteps/model` - could you import that type instead of repeat a new definition here?

editor.planx.uk/src/@planx/components/NextSteps/model.ts
```diff
@@ -3,15 +3,21 @@ import { MoreInformation, parseMoreInformation } from "../shared";
...
+export interface Step {
+  title: string;
+  description: string;
+  url: string;
+}
+
+export const parseNextSteps = (
+  data: Record<string, any> | undefined
 ): NextSteps => ({
+  steps: /* remove once migrated */ data?.taskList?.steps || data?.steps || [],
```
jessicamcinchak, line 18: it looks to me like this could just be: `steps: data?.steps || [],` ? 

i don't see any other references to `taskList` & contrary to the code comment it looks like we definitely still want the `steps` property defined overall because it's required on the `NextSteps` interface defined above.

jessicamcinchak (commented):
Back with fresh eyes - a few last minor comments but overall in great shape !

editor.planx.uk/src/ui/NextStepsList.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+function ListItem(props: Item) {
+  return (
+    <Step>
+      <Inner href={props.url}>
```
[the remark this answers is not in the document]
ianjon3s: I usually work along the principal of _let the user have control_ with regards to opening new links, i.e. keep all links opening in the same tab unless the user decides they want to open a new tab.

However I acknowledge there may be other places where we open links in new tabs and it'd be better to be consistent. Do we have a house style for this?

editor.planx.uk/src/@planx/components/NextSteps/Public.tsx
ianjon3s: Appreciate the clarification on this.

editor.planx.uk/src/ui/NextStepsList.tsx
ianjon3s: I've updated this to import the type.

editor.planx.uk/src/@planx/components/NextSteps/model.ts
ianjon3s: Updated, thanks for the explanation here.

jessicamcinchak (approved):
I'll let you make final decide what to do about where link open based on govuk patterns / yesterday's dev chat, but I don't see any reason this can't merge! Looks & works great.

Since this is a brand new componet, sharing staging links rather than pizza ones for PO testing on Trello will probably save some rebase headaches and we can just keep fixing-forward if any feedback comes up :slightly_smiling_face:

ianjon3s:
Thanks @jessicamcinchak , I've updated the links to open in new tabs as discussed. Will merge and add for PO testing now.

PR #1962, opened 2023-07-12
chore: Add theme type definitions

ghost:
This PR adds an export for the `Theme` type and implements it in the remaining `makeStyles` calls. This keeps the type checker happy so we can upgrade typescript.


DafyddLlyr (approved):
Thanks for this one 👍 

I've been slowly working my way through removing `makeStyles` and there's now really just a handful remaining. I've got about 5 open PRs that drop a few more, which will them simplify this PR as there will be fewer changes.

If we merge this first I `think` git will pick up that these changes won't affect those PRs so it won't be a case of rebasing x5.

PR #1966, opened 2023-07-12
chore: update Storybook components [Design System / Atoms and Molecules]

jessicamcinchak:
Filling in a lot of gaps here, but haven't really sweated what's best categorised as a "Molecule", an "Atom" or otherwise yet - I figure we can revisit that more comprehensively later once we have better component coverage.

My general approach here is:
- Update type annotations & naming in existing stories
- Add new stories for un-documented components
- "Form elements" typically just get a "Basic" story, maybe an additional "With Error" story if applicable

### DIFF
--- a/editor.planx.uk/src/components/AnalyticsDisabledBanner.stories.tsx
+++ b/editor.planx.uk/src/components/AnalyticsDisabledBanner.stories.tsx
@@ -1,0 +1,18 @@
+import { Meta, StoryObj } from "@storybook/react";
+import React from "react";
+
+import AnalyticsDisabledBanner from "./AnalyticsDisabledBanner";
+
+const meta = {
+  title: "Design System/Molecules/AnalyticsDisabledBanner",
+  component: AnalyticsDisabledBanner,
+} satisfies Meta<typeof AnalyticsDisabledBanner>;
+
+type Story = StoryObj<typeof meta>;
+
+export default meta;
+
+/** Add `&analytics=false` to the end this page's URL to display the banner */
+export const Basic = {
+  render: () => <AnalyticsDisabledBanner />,
+} satisfies Story;
--- a/editor.planx.uk/src/components/TestEnvironmentBanner.stories.tsx
+++ b/editor.planx.uk/src/components/TestEnvironmentBanner.stories.tsx
@@ -1,0 +1,18 @@
+import { Meta, StoryObj } from "@storybook/react";
+import React from "react";
+
+import TestEnvironmentBanner from "./TestEnvironmentBanner";
+
+const meta = {
+  title: "Design System/Molecules/TestEnvironmentBanner",
+  component: TestEnvironmentBanner,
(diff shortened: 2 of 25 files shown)


editor.planx.uk/src/components/AnalyticsDisabledBanner.stories.tsx
```diff
@@ -0,0 +1,18 @@
...
+const meta = {
+  title: "Design System/Molecules/AnalyticsDisabledBanner",
+  component: AnalyticsDisabledBanner,
+} satisfies Meta<typeof AnalyticsDisabledBanner>;
+
+type Story = StoryObj<typeof meta>;
+
+export default meta;
+
+/** Add `&analytics=false` to the end this page's URL to display the banner */
```
jessicamcinchak, line 15: JSDoc comments translate to Storybook 7 "descriptions" :sparkles: see here for more https://storybook.js.org/docs/react/api/doc-block-description#writing-descriptions

jessicamcinchak:

editor.planx.uk/src/components/AnalyticsDisabledBanner.stories.tsx
DafyddLlyr: This is a nice feature!

editor.planx.uk/src/components/TestEnvironmentBanner.stories.tsx
```diff
@@ -0,0 +1,18 @@
...
+const meta = {
+  title: "Design System/Molecules/TestEnvironmentBanner",
+  component: TestEnvironmentBanner,
+} satisfies Meta<typeof TestEnvironmentBanner>;
+
+type Story = StoryObj<typeof meta>;
+
+export default meta;
+
+/** Will not display if the page URL includes .uk */
```
DafyddLlyr, line 15: 👍

editor.planx.uk/src/ui/CollapsibleInput.stories.tsx
```diff
@@ -1,27 +1,35 @@
...
   component: CollapsibleInput,
-};
+} satisfies Meta<typeof CollapsibleInput>;
 
-export const Basic: StoryObj<Props> = {
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+// TODO clarify use of args.children versus render here
```
DafyddLlyr, line 16: Yeah I also had this question whilst reviewing - it looks like `render()` is actually optional and will be defaulted using these `args` and the predefined `meta` right?

DafyddLlyr (approved):
Good to get these converted / updated, and lots of new additions also.

I've not necessarily been thinking in terms of atoms / molecules often when building components (often as many of the building blocks have already been created), but it's quite a helpful framework to have. I agree with the approach of revisiting (and defining) these categories later once we have Storybook updated as we want it.

PR #1972, opened 2023-07-13
fix: HTML or markdown detection

ianjon3s:
PR fixes a regression where markdown was been rendered as source code in the planning constraints component.

Not an ideal solution but could work as a patch for now.


jessicamcinchak (approved):
Thanks for picking this one up! Patch is behaving as expected for me: rich text is parsing links and other html tags and I can still add/display images in rich text as expected too :heavy_check_mark: 

One to keep on the radar to see if we can get a more flexible solution in the future, but very appreciative of this quick fix now!

PR #1973, opened 2023-07-14
chore: update Storybook [Design System / Pages]

jessicamcinchak:
We haven't captured "pages" at all in previous iterations of Storybook - but feels like a useful area to try to get some coverage. "Pages" will cover custom page layouts/templates associated with Save & Return, Resume, Reconciliation, Invite to Pay and other journeys.

### DIFF
--- a/editor.planx.uk/src/pages/Pay/InviteToPay.stories.tsx
+++ b/editor.planx.uk/src/pages/Pay/InviteToPay.stories.tsx
@@ -1,0 +1,15 @@
+import { Meta } from "@storybook/react";
+import React from "react";
+
+import InviteToPay from "./InviteToPay";
+
+const meta = {
+  title: "Design System/Pages/InviteToPay",
+  component: InviteToPay,
+} satisfies Meta<typeof InviteToPay>;
+
+export default meta;
+
+// export const Basic = {
+//   render: () => <InviteToPay {...paymentRequestMock} />,
+// };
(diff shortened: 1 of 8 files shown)


editor.planx.uk/src/pages/Pay/InviteToPay.stories.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+const meta = {
+  title: "Design System/Pages/InviteToPay",
+  component: InviteToPay,
+} satisfies Meta<typeof InviteToPay>;
+
+export default meta;
+
+// export const Basic = {
+//   render: () => <InviteToPay {...paymentRequestMock} />,
```
jessicamcinchak, line 14: I'm coming up on a few cases of needing quite fleshed out mock data for meaningful stories - I'm curious if this could be something we start storing in & importing from `planx-core` ? We have a detailed selection of mocks in `api.planx.uk/tests/mocks` that would be handy to share with Storybook here.

jessicamcinchak:

editor.planx.uk/src/pages/Pay/InviteToPay.stories.tsx
DafyddLlyr: This is something I've been wondering about also - we have a lot of mocks dotted about the place, often untyped as just JSON. There's a lot of repetition and a lot of large flows in being redefined.

A spike to look into this feels like a good shout - there must be some standard patterns and practices we could look at here. Importing from a single (typed) source of truth feels like a good option for data structures shared across services.

DafyddLlyr (approved):

PR #1979, opened 2023-07-18
chore: Split data into breadcrumbs and passport columns

ghost:
The new sessions table has a `data` column but will only store an array of `breadcrumbs` so this column has been renamed.

### DIFF
--- a/hasura.planx.uk/migrations/1689681979734_split_out_breadcrumbs_and_passport/up.sql
+++ b/hasura.planx.uk/migrations/1689681979734_split_out_breadcrumbs_and_passport/up.sql
@@ -1,7 +1,0 @@
-ALTER TABLE "public"."sessions" RENAME COLUMN "data" TO "breadcrumbs";
-
-ALTER TABLE "public"."sessions" ADD COLUMN "passport" jsonb
- NULL;
-
-COMMENT ON COLUMN "public"."sessions"."passport"
-  IS E'An representation of collected breadcrumb values';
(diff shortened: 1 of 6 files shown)


hasura.planx.uk/migrations/1689681979734_split_out_breadcrumbs_and_passport/up.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE "public"."sessions" RENAME COLUMN "data" TO "breadcrumbs";
+
+ALTER TABLE "public"."sessions" ADD COLUMN "passport" jsonb
+ NULL;
```
jessicamcinchak, line 4: Question: if the passport can always be derived from the breadcrumbs, what value do you think we get from storing it directly? In moving towards a new data model I've been curious if we can exclusively store breadcrumbs and flow graphs - and have shared methods (eg a new planx-core iteration of `computePassport()`?) to derive all other commonly used data formats? 

Just curious around your thoughts on this! Happy for this to merge as proposed here regardless.

jessicamcinchak (approved):

hasura.planx.uk/migrations/1689681979734_split_out_breadcrumbs_and_passport/up.sql
jessicamcinchak: Sounds good, thanks for sense checking this. If porting over proves tricky, very happy with a shared short versus long term goal here though ! Agree it would be easy enough to drop the column in the future if simplest to get a basic working example, then move `computePassport` later.

PR #1981, opened 2023-07-18
feat: mobile styles update

ianjon3s:
### Summary

PR updates and fixes a number of mobile specific styles, following extensive mobile QA.

### Updates

- Updates default "factor" for responsive font resizing
- Adds small button size
- Adds missing `<Typography>` where needed
- Adjusts margins and padding of numerous components to unify layout
- Changes hierarchy of `<MapContainer>` and `<MapFooter>` to prevent overlap on smaller screens
- Adds a `nowrap` span to red line size indicator to keep units (m²) on same line as number

### Previews

Size comparison before (left) & after (right)

image

image

### DIFF
--- a/editor.planx.uk/src/@planx/components/Checklist/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/Checklist/Public.tsx
@@ -4,7 +4,6 @@ import { styled } from "@mui/material/styles";
 import type { Checklist, Group } from "@planx/components/Checklist/model";
 import ImageButton from "@planx/components/shared/Buttons/ImageButton";
 import Card from "@planx/components/shared/Preview/Card";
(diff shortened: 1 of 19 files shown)


DafyddLlyr:
Pizza link for mobile 👇 

frame

DafyddLlyr (changes requested):
I'd suggest having a crack at converting the shared style in `fullWidthContent()` to it's own component - this will probably tidy a few things up and be a simpler way to understand structure and layout as we're building components I think.

DafyddLlyr (approved):
That's fantastic thanks - significant improvement to the readability of the code I think 👍

ianjon3s:
Thanks @DafyddLlyr , agreed it makes a lot more sense as a component.

I've just got a couple more style updates coming through that I noticed on final run through the service/components.

PR #1982, opened 2023-07-18
fix: Conditionally hide optional file list in `UploadAndLabel`

DafyddLlyr:
Addresses feedback [here](https://trello.com/c/yOjEJ3XT/2137-build-upload-and-label-component) - 

> The ‘Optional’ list makes it rather absurdly long! Maybe let’s only show Required and recommended here, and Optional only in the dropdown? Or show optional list only if nothing is set to required or recommended?

image

image


jessicamcinchak (commented):
This looks good - but was working through stories for this component and two new questions came up for me:

1. If we have only "Optional files", we're still requiring that at least one file be uploaded before "Continuing" and that file must be tagged - so it's not really "optional" after all? If there's only optional files, should we also adjust validation to allow the user to "Continue" and skip past without uploading anything?
Screenshot from 2023-07-23 10-46-20

2. When "hideDropzone" is true, would it make sense to show all file types regardless of "Required", "Recommended" or "Optional" ? If the aim of that view is to help applicants prepare, I don't think we'd want to hide any defined rules? And it's a single list at that point, so perhaps not so much a design issue like the original motivation here.

DafyddLlyr:
Hitting a few issues with "Back" and "Change" behaviour here - moved back to draft as I look into this further.

Possibly related to https://trello.com/c/aHHaTagP/2525-upload-and-label-throws-id-not-found-error-when-going-back-or-changing

jessicamcinchak (approved):

editor.planx.uk/src/@planx/components/FileUploadAndLabel/model.ts
```diff
@@ -226,9 +236,9 @@ export const generatePayload = (fileList: FileList): Store.userData => {
 const getCachedSlotsFromPreviousData = (
   userFile: UserFile,
   previouslySubmittedData: Store.userData | undefined,
-): FileUploadSlot =>
+): FileUploadSlot[] =>
```
DafyddLlyr, line 239: This incorrect type was the root cause of this bug https://trello.com/c/aHHaTagP/2525-upload-and-label-throws-id-not-found-error-when-going-back-or-changing

DafyddLlyr:
[Upload and Label throws `id not found` error when going "back" or "changing"](https://trello.com/c/aHHaTagP/2525-upload-and-label-throws-id-not-found-error-when-going-back-or-changing)

jessicamcinchak (approved):
Re-tested and back & change working as expected again for me, thanks for picking these ones up!

PR #1983, opened 2023-07-19
fix: Stop sending E2E emails to enquiries inbox

DafyddLlyr:
Told Mel earlier in the week I'd sort this...! 

An outstanding E2E test was sending emails to `testAgent@opensystemslab.io` which was then being forwarded to `enquiries@opensystemslab.io`.

### DIFF
--- a/e2e/tests/ui-driven/src/invite-to-pay/mocks.ts
+++ b/e2e/tests/ui-driven/src/invite-to-pay/mocks.ts
@@ -8,7 +8,7 @@ import inviteToPayFlow from "../flows/invite-to-pay-flow";
 import { TEST_EMAIL } from "../helpers";
 
 export const mockPaymentRequest: Partial<PaymentRequest> = {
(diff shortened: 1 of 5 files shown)


e2e/tests/ui-driven/src/invite-to-pay/mocks.ts
```diff
@@ -5,6 +5,7 @@ import {
   SessionData,
 } from "@opensystemslab/planx-core/types";
 import inviteToPayFlow from "../flows/invite-to-pay-flow";
+import { TEST_EMAIL } from "../helpers";
 
 export const mockPaymentRequest: Partial<PaymentRequest> = {
   payeeEmail: "testNominee@opensystemslab.com",
```
jessicamcinchak, line 11: Do we want to replace this one too with `TEST_EMAIL` ?

jessicamcinchak:
Thanks for fix & update - I've merged so this goes in before next week when you get back :sunny:

PR #1991, opened 2023-07-20
fix: better PlanningConstraints story with positive & negative mocks

jessicamcinchak:
A few PlanningConstraint questions came up during planning and it's always hard to find good example addresses on the spot - so updating this story to better reflect an example "With Intersections" and "Without Intersections".

Before: https://storybook.planx.uk/?path=/docs/planx-components-planningconstraints--docs
After: https://storybook.1991.planx.pizza/?path=/docs/planx-components-planningconstraints--docs


editor.planx.uk/src/@planx/components/PlanningConstraints/mocks/classifiedRoadsNegativeResponseMock.ts
```diff
@@ -0,0 +1,20 @@
+// 2, WALLER ROAD, BEACONSFIELD, BUCKINGHAMSHIRE, HP9 2HE
+// https://api.editor.planx.uk/roads?usrn=35200102
```
DafyddLlyr, line 2: 👍

DafyddLlyr (approved):
This is a really nice approach to feedback from planning - stories should reflect these types of questions. Great to see!

PR #1994, opened 2023-07-21
fix: don't validate that all required files are tagged yet when clicking modal "Done" 

jessicamcinchak:
This came out of PO testing feedback: 

Previously:
- Upload one file, the tagging modal opens, select 1 or many "Required files" checkboxes
- Click “Done” in modal and throws error “All required files must be tagged”. 
  - Clicking away from Modal still closes it successfully, but as a user I don't understand this is an option
  - Clicking "Done" throws error that confuses me because I want to close it and upload my second file. I have to select incorrect checkboxes to proceed through "Done"
 
This wasn't noticeable if you drag many files into dropzone at same time and tag all at same time, or if you upload a single file that satisfies all requirements in one go. But if you're uploading one-by-one and focused on navigating with buttons, it's confusing! 
 
Now:
- Only validate on "Continue", simply close the modal on "Done" - basically reverting back to original behavior here https://github.com/theopensystemslab/planx-new/pull/1937

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Modal.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Modal.tsx
@@ -28,7 +28,6 @@ import {
   removeSlots,
   resetAllSlots,
 } from "./model";
-import { fileListSchema, slotsSchema } from "./schema";
 
 interface FileTaggingModalProps {
   uploadedFiles: FileUploadSlot[];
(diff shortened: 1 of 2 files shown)


jessicamcinchak:
[Upload and Label tagging validation is too strict on modal](https://trello.com/c/ohHxzNOX/2526-upload-and-label-tagging-validation-is-too-strict-on-modal)

DafyddLlyr (approved):

PR #1995, opened 2023-07-23
chore: remaining Storybook components

jessicamcinchak:
Last of the outstanding ones per Trello checklist: 
- [x] Confirmation
- [x] FileUpload
- [x] FileUploadAndLabel
- [x] Result
- [x] Review
- [x] Section
- [x] Send
- [x] SetValue
- [x] TaskList
- [ ] ExternalPortal *
- [ ] Filter *
- [ ] InternalPortal *

`*` = Thesecomponent types DO NOT create a breadcrumb, therefore they have zero Public representation - which is different than "hidden" component types like Calculate or SetValue that still set a breadcrumb. I think I'd vote these are totally omitted from Storybook? I attemped to mock ExternalPortal & Filter but they look like a bit like misfits in my opinion.


jessicamcinchak:

editor.planx.uk/src/@planx/components/Review/Review.stories.tsx
```diff
@@ -1,276 +1,54 @@
...
-    },
-    type: 200,
-  },
-  wViIvfuIJC: {
-    data: {
-      text: "(No Result)",
-    },
-    type: 200,
-  },
+// TODO improve mock data to include one component of each type
```
DafyddLlyr, line 26: We should also add  this step to the "adding a component" doc I guess?

editor.planx.uk/src/@planx/components/SetValue/SetValue.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const WithEditor = () => {
+  return (
+    <>
+      <Wrapper Editor={Editor} Public={Public} />
+      <WarningContainer>
+        <ErrorOutline />
+        <Typography variant="body2" ml={2}>
+          This component is only available in the Editor when designing
+          services, it does <strong>not</strong> display in the Public form.
```
DafyddLlyr, line 24: I think this is a good approach - super clear.

editor.planx.uk/src/@planx/components/Section/Section.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+const meta = {
+  title: "PlanX Components/Section",
+  component: Public,
+} satisfies Meta<typeof Public>;
+
+type Story = StoryObj<typeof meta>;
+
+export default meta;
+
+// TODO needs internal state
```
DafyddLlyr, line 17: I'd be happy to pick this one up!

DafyddLlyr (approved):
Thanks for this! 

I agree that the outstanding components don't really make much sense - we could have an Editor modal only displayed but there's not much there really to display.

editor.planx.uk/src/@planx/components/Review/Review.stories.tsx
jessicamcinchak: Maybe, but it's really just one component _if_ shown on the Review page - not sure it's a meaningful step yet at that phase (still interested in thinking through less manual/duplicated mock data etc?) but will keep this one on the radar :ok_hand:

PR #1998, opened 2023-07-25
chore: move seeds from main docker file

ghost:
The `seed-database` command has been running during a `pnpm start` causing data to be truncated. 

This container has now been moved to its own dockerfile and must be included explicitly.

I have also added a few convinence pnpm scripts to allow `pnpm run up` and `pnpm run down`  along with `pnpm start` and `pnpm stop`.  The first two create and destroy data, the second two don't.

### DIFF
--- a/scripts/pullrequest/update.sh
+++ b/scripts/pullrequest/update.sh
@@ -13,4 +13,5 @@ set +o allexport
 docker compose \
   -f docker-compose.yml \
   -f docker-compose.pizza.yml \
+  -f docker-compose.seed.yml \
   up --build --renew-anon-volumes --force-recreate --remove-orphans --wait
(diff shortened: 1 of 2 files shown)


DafyddLlyr (approved):

scripts/pullrequest/update.sh
```diff
@@ -13,4 +13,5 @@ set +o allexport
 docker compose \
   -f docker-compose.yml \
   -f docker-compose.pizza.yml \
+  -f docker-compose.seed.yml \
```
DafyddLlyr, line 16: Will this overwrite all Pizza data when a new commit is pushed? 

I think the intention would be that the Pizza is just seeded once on creation and then data will persist until the Vultr instance is destroyed / recreated manually if we need to re-seed.

DafyddLlyr: The `ON CONFLICT DO NOTHING` bit is what I forgot! This looks good, thanks for explaining.

DafyddLlyr (approved):

PR #2003, opened 2023-07-25
feat: table of contents for admin endpoints

jessicamcinchak:
https://api.2003.planx.pizza/admin

Screenshot from 2023-07-25 18-15-07

### DIFF
--- a/api.planx.uk/admin/index.html
+++ b/api.planx.uk/admin/index.html
@@ -1,0 +1,100 @@
+<!DOCTYPE html>
+<head>
+  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Inter">
+  <style>
+    body {
+      font-family: "Inter", sans-serif;
+      font-size: 1em;
+      max-width: 850px;
+      margin: auto;
+      margin-top: 2em;
+    }
+
+    table {
+      width: 100%;
+      border-collapse: collapse;
+      border-width: 2px;
+      border-color: #0010a4;
+      border-style: none;
+      color: #000000;
+      font-size: 1em;
+    }
+
+    table td,
+    table th {
+      border-width: 2px;
+      border-color: #0010a4;
+      border-style: solid;
+      padding: 10px 5px;
+    }
+
+    table thead {
+      background-color: #0010a4;
+      color: #fff;
+      text-align: left;
+    }
+  </style>
+</head>
+<div>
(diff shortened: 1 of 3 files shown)


api.planx.uk/admin/index.html
```diff
@@ -0,0 +1,119 @@
...
+        <td><code>/session/:sessionId/zip</code></td>
+        <td><code>?includeXML=true</code></td>
+        <td>
+          Generates a zip file for Send to Email, or Uniform when XML is
+          included
+        </td>
+      </tr>
+    </tbody>
+  </table>
+</div>
```
jessicamcinchak, line 119: html doesn't feel like a particularly clever solution, but looks okay and is easy enough to update? open to other suggestions though!

DafyddLlyr (commented):
Appreciate this is still a draft atm...!

I agree that we want...
 - Better (any!) API docs
 - A convenient way to access some of this data as devs / admins
 
This implementation sort of falls between two stools for me as not quite meeting what we're looking for I think.

What I'd go for here is proper API documentation with Swagger (just for the `/admin` endpoints for now). I've not looked into exactly how to do this for TS / Express but it will be a case of installing a few libraries and defining a spec file either as JSDocs, a TS object, or a YAML file and letting Swagger to the rest. Happy to have a look at this together or review options here.

It would also be nice to actually have a very very simple frontend to access some of this (e.g. `admin.planx.uk` or `editor.planx.uk/admin`) which could just be a table with sessionIds and then links to these API endpoints for convenience, e.g. - 

| SessionId | Zip | JSON | XML | CSV |
|--------|--------|--------|--------|--------|
| abc123 | Download | Link | Link | Download |
| abc456 | Download | Link | Link | Download |
| abc789 | Download | Link | Link | Download |

Alternatively, just having something like a HATEOAS reponse at `admin/:sessionId` with links to the various views would be another very quick/lightweight way of accessing these endpoints. Something like - 

```json
{
  "sessionId": "abc123",
  "links": {
    "xml": "https://api.editor.planx.dev/admin/session/abc123/xml",
    "zip": "https://api.editor.planx.dev/admin/session/abc123/zip",
    "csv": "https://api.editor.planx.dev/admin/session/abc123/csv"
  }
}
```

jessicamcinchak:
Closing in favor of #2040

PR #2006, opened 2023-07-26
fix: missing input label on feedback

ianjon3s:
Accessibility fix: missing input label on collapsable feedback input.

PR adds an `aria-label` to the input, referencing the input button text.

image

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/FeedbackInput.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/FeedbackInput.tsx
@@ -18,7 +18,7 @@ const StyledReactMarkdown = styled(ReactMarkdown)(() => ({
 }));
 
 const FeedbackInput: React.FC<Props> = ({ text, ...componentProps }: Props) => (
-  <CollapsibleInput {...componentProps} name="feedback">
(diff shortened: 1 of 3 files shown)


DafyddLlyr (changes requested):
Thanks for catching this 👍 

This looks like the sort of thing that should have been caught by an a11y test. There is a test for the parent component where this is used (`editor.planx.uk/src/@planx/components/Result/Public.test.tsx`) but it looks like it's running without the button being hit (and therefore the input isn't in the DOM).

Could you please either - 
 - Improve the test case in the file referenced above, but toggle the input to display before checking for a11y violations
 - Add a test file for `CollapsibleInput`
 
The second would probably be my preference as we would be testing a smaller, simpler, component.

DafyddLlyr:
@ianjon3s Let me know if you want to pair on this or want a bit more detail on the above comment 👍

DafyddLlyr (approved):
Happy for this to be merged @ianjon3s !

Let's catch up when you're back from annual leave and pair on unit tests here 👍

ianjon3s:
Thanks @DafyddLlyr

PR #2007, opened 2023-07-26
fix: missing input label on dropzone

ianjon3s:
Accessibility fix: missing input label on dropzone component.

PR adds an `aria-labelledby` tag to the input, referencing the input button text.

image


ianjon3s:
Thanks @DafyddLlyr . It may be because the `<input>` is nested within a `<button>` element, resulting in invalid HTML (something I've marked to look into at a later date).

Also I'm using the Wave testing tool, which is notoriously picky when it comes to A11Y rules.

PR #2021, opened 2023-07-27
feat: improve usability of more info drawer

ianjon3s:
PR improves the usability of the 'more info' drawer.

Updates:

- Change position of close button to `fixed` so that it remains in place on scroll. This is especially needed for larger amounts of content on smaller screens, where there isn't the option to 'click outside' to close, and after reading content the user must scroll back to the top of the dialog to close.
- Increase contrast of close button.
- Bring `<Container>` into the dialog for more consistent and responsive gutters.
- Reduce heading font sizes so they form more readable content.
- Update `font-weight` of `<h3>`, this was missed when the numbers were incremented to make of use of the previously unused `<h2>` size.

Example to come once a pizza link is generated.


DafyddLlyr:
Again sorry - this should run through CI following a rebase!

DafyddLlyr (approved):

PR #2027, opened 2023-07-27
fix: Fix `pull-secrets.sh`

DafyddLlyr:
Currently causing CI to fail. I did re-run all jobs on https://github.com/theopensystemslab/planx-new/pull/2005 after making this change and all jobs passed - GitHub actions must have been using a cache from a previous run.

This works locally for me, and if this PR succeeds it also works on CI.


jessicamcinchak (approved):
:100: thanks for fix!

PR #2028, opened 2023-07-27
fix: a11y errors in upload dialog

ianjon3s:
PR fixes 3 issues identified in the dialog of the upload & label component.

Audited using [WAVE accessibility tool](https://wave.webaim.org/) in Chrome.

### Issue 1
Dialog does not have a heading or summary text that can describe it to non-sighted users. Due to this it is also missing an aria-label.

image

**Fix:** I've added the visually hidden text "What do these files show" and assigned as an `aria-label`.

### Issue 2
Multi-select input does not have an assigned label.

image

**Fix:** The ID of the label was also being assigned to the input, using the `labelId` property (probably done in error as this created a duplicate ID. I've moved this ID to be assigned to `aria-labelledby`.

### Issue 3
Checkbox inputs do not have correctly assigned labels.

image

Fix: I've added an ID to the text elements and associated these to the inputs as an `aria-label`.


DafyddLlyr (approved):

DafyddLlyr:
@ianjon3s CI should pass here once you rebase 👌

PR #2029, opened 2023-07-27
fix: revert omission of `designated` variable in GIS response

jessicamcinchak:
manually reverts #1821 

for testing by August & George, see this thread: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1690462520855369

this was only working for sites that have at least one true constraint, not if all constraint values are part of `_nots` - which seems to break our passport granularity checks


DafyddLlyr:
This will need a rebase following #2027

DafyddLlyr (approved):
Approved based on August's positive feedback on Slack 👍

PR #2033, opened 2023-07-27
fix: radio label spacing

ianjon3s:
PR fixes an issue with vertical spacing in radio buttons. Padding is applied to the radio button icon rather than the element itself, meaning if text labels are taller than the button icon there is no space between the text.

Small spacing adjustment also made to `ImageRadio` to compensate (no adjustment needed for `DescriptionRadio`).

Example before (left) and after (right):

image

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/Radio/BasicRadio.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Radio/BasicRadio.tsx
@@ -15,6 +15,7 @@ const BasicRadio: React.FC<Props> = ({ id, onChange, title }) => (
     onChange={onChange}
     control={<Radio />}
     label={title}
+    sx={{ pb: 1 }}
   />
 );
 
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/shared/Radio/BasicRadio.tsx
```diff
@@ -15,6 +15,7 @@ const BasicRadio: React.FC<Props> = ({ id, onChange, title }) => (
     onChange={onChange}
     control={<Radio />}
     label={title}
+    sx={{ pb: 1 }}
```
DafyddLlyr, line 18: This leads to all instances of `BasicRadio` (including the final one) having additional padding-bottom. Would it be better to control this at the parent container to avoid the double spacing above the "Continue" button?

ianjon3s: I had considered omitting the final spacing, but the extra 10px at the bottom of the component gives good vertical balance, when compared to the description and image version which have borders + margins on each element.

DafyddLlyr (approved):

editor.planx.uk/src/@planx/components/shared/Radio/BasicRadio.tsx
DafyddLlyr: Totally fair - it looks better to me - just wanted to check it was intentional!

PR #2040, opened 2023-07-31
feat: add Swagger documentation for API

jessicamcinchak:
replaces #2003.

:scroll: https://api.2040.planx.pizza/docs/ 

still figuring out / known todos:
- [x] configure / document `servers` 
- [x] document JWT authentication for admin endpoints
- [x] think about how we might derive/define ["components"](https://swagger.io/docs/specification/components/) based on types? not actually connected to typescript in meaningful way yet, but re-using parameters and auth definitions 
- [ ] document `responses` for each endpoint added so far

resources:
- https://swagger.io/specification/
- https://nerd-corner.com/import-swagger-in-node-typescript-project/
- https://blog.logrocket.com/documenting-express-js-api-swagger/

Screenshot from 2023-07-31 15-38-01


api.planx.uk/server.ts
```diff
@@ -251,6 +253,28 @@ passport.deserializeUser(function (obj: Express.User, cb) {
...
+      version: "0.1.0",
+      description:
+        "Plan✕ is a platform for creating and publishing digital planning services. Our REST API is built on Express and documented with Swagger.",
+      license: {
+        name: "MPL-2.0",
+        url: "https://github.com/theopensystemslab/planx-new/blob/main/LICENSE.md",
+      },
+    },
+    schemes: ["http", "https"],
+    servers: [{ url: "http://localhost:7002/" }],
```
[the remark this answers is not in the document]
DafyddLlyr: Not 100% sure this will work but worth a try!

api.planx.uk/server.ts
```diff
@@ -251,6 +253,28 @@ passport.deserializeUser(function (obj: Express.User, cb) {
 
 const app = express();
 
+// Swagger documentation config
+const options = {
```
DafyddLlyr, line 257: The `failOnErrors: true` option looks helpful and will probably help identify issues https://www.npmjs.com/package/swagger-jsdoc

DafyddLlyr (approved):
This is looking great 👍 

I agree that it's going to make an already big `server.ts` feel huge and unwieldy pretty quickly but I'm totally happy with this implementation as a really solid first step.

Going forward, it would be nice to split the API into modules more clearly (we know this is something we want to tackle).

```
server.ts
routes.ts    # Import all routes
docs.ts      # Import all docs, merge and convert to json/yaml
/gis
├── routes.ts
├── docs.ts
/pay
├── routes.ts
├── docs.ts
/etc...
```

We'd lose co-location of docs immediately next to routes, but hopefully this would be offset by some sort of linting / typing of the doc files (https://www.npmjs.com/package/@types/swagger-schema-official).

Just some quick thoughts - none of this is pressing and we can be can migrate module by module I'd hope.

api.planx.uk/server.ts
jessicamcinchak: Added!

jessicamcinchak:
[Build and document new API endpoint [EPIC]](https://trello.com/c/QYEzY0yi/2443-build-and-document-new-api-endpoint-epic)

jessicamcinchak:
@DafyddLlyr made a few organisational changes here since you're initial review if you want to give one more look before I merge this MVP! 

- moved JSDocs out of server.ts and into files where endpoints are defined, this should help keep server.ts more manageable until we do a re-org to modules
- started adding a few shared types for parameters and auth under `components` in the main config object 
- have _not_ started adding `responses` to each endpoint yet because a) requires lots of data mocking and b) the config object is going to start to grow very quickly for shared response schemas and I think we probably want to move out of server.ts to a `.yml` or `.json` in the API root directory and import only in server.ts ? 

I'm still coming up a bit short on how best to leverage TypeScript types, but feels like a question worth exploring next before we get too deep into this setup?

api.planx.uk/send/bops.ts
```diff
@@ -12,6 +12,26 @@ interface SendToBOPSRequest {
...
+ *    security:
+ *      - bearerAuth: []
+ *    requestBody:
+ *      description: This endpoint is only called via Hasura's scheduled event webhook, so body is wrapped in a `payload` key
+ *      required: true
+ *      content:
+ *        application/json:
+ *          schema:
+ *            $ref: '#/components/schemas/SessionPayload'
+ */
```
jessicamcinchak, line 34: was getting curious about documenting POSTs too - so just went for an MVP `submissions` section of docs too: 
Screenshot from 2023-08-02 20-39-03

next steps i imagine here:
- correct auth (JWT doesn't quite feel accurate for these?)
- mock response statuses and data
- much more detailed human `description` - eg our submission endpoints are responsible for third party auth, generating payload, updating audit tables, etc. A lot of this lives in code comments, but could easily move into swagger docs here

DafyddLlyr: Yeah this is correct - JWT is not correct as these requests come via Hasura so it's more like an API key or header than validates this.

DafyddLlyr (approved):
This is great!

PR #2055, opened 2023-08-01
fix: notice component text colour

ianjon3s:
PR fixes a regression where the text of the notice component is outputting as the primary theme colour instead of a black/white contrast to the background.

Before:
image

After:
image


DafyddLlyr:
Change looks good - looks like there are a few React unit tests based on the previous assumption that would need to be updated.

PR #2061, opened 2023-08-03
feat: Basic `Section` stories

DafyddLlyr:
I spent some time trying to resolve this using Zustand stores without much luck within the Storybook components. I think in order to get this to work we'd need to use a Provider approach, similar to what's outlined here - 

This would be a big refactor and I'm unsure of the real benefit for the effort involved.

I've gone for just a stateless wrapper component to simplify the visual testing of the "Section" component.

### DIFF
--- a/editor.planx.uk/src/@planx/components/Section/Section.stories.tsx
+++ b/editor.planx.uk/src/@planx/components/Section/Section.stories.tsx
@@ -1,9 +1,9 @@
 import { Meta, StoryObj } from "@storybook/react";
-import React from "react";
+import { ComponentProps } from "react";
+import { SectionNode } from "types";
 
-import Wrapper from "../fixtures/Wrapper";
-import Editor from "./Editor";
-import Public from "./Public";
+import { TYPES } from "../types";
+import { Root as Public } from "./Public";
 
 const meta = {
   title: "PlanX Components/Section",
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/@planx/components/Section/Section.stories.tsx
```diff
@@ -14,13 +14,73 @@ type Story = StoryObj<typeof meta>;
...
+      firstSection: {
+        auto: false,
+      },
+    },
   },
 } satisfies Story;
 
-export const WithEditor = () => {
-  return <Wrapper Editor={Editor} Public={Public} />;
-};
```
DafyddLlyr, line 26: One downside - I can't get this approach to work with the `Wrapper` component. Not a huge issue for this single component which has a simple Editor representation but it probably shows that this isn't an approach we want to repeat too much.

editor.planx.uk/src/@planx/components/Section/Section.stories.tsx
```diff
@@ -14,13 +14,73 @@ type Story = StoryObj<typeof meta>;
...
+export const SecondSectionCompleted = {
```
jessicamcinchak, line 73: nit: would a `WithNewInformationNeeded` story be more distinct/meaningful to mock than another "completed" tag?

jessicamcinchak (approved):
Thanks for picking this one up & sorting out !

editor.planx.uk/src/@planx/components/Section/Section.stories.tsx
DafyddLlyr: Good shout - I'll add this.

PR #2065, opened 2023-08-04
fix: Restore result flag colour

DafyddLlyr:
image


DafyddLlyr:
[Restore flag colours on result component](https://trello.com/c/j8zg0cQs/2509-restore-flag-colours-on-result-component)

jessicamcinchak (approved):

PR #2090, opened 2023-08-11
feat: Use cookieAuth strategy for API docs auth

DafyddLlyr:
Something I hit earlier - requests are sent with a cookie anyway currently, due to the domain.

We can just rely on this for now as it's simpler / quicker to use. Especially in the context of non-devs using the API docs (e.g. for find/replace or other admin operations without a UI) - I think it's a lot easier to explain "just log in and then use the docs" vs "log in, get bearer token from network tab, copy into Swagger" etc).

Swagger gives us the option to overwrite this cookie as well so if we had a way of locally producing one we could test API docs with various roles.

To test -
 - Try open API endpoint such as `/roads` on Staging - a cookie is sent with the request without any user interaction
 - Try `/me` endpoint on Pizza. It will fail if you are logged out, and succeed if you are logged in


api.planx.uk/server.ts
```diff
@@ -450,6 +468,46 @@ app.get("/hasura", async function (_req, res, next) {
...
+ *                first_name:
+ *                  type: string
+ *                  example: Albert
+ *                last_name:
+ *                  type: string
+ *                  example: Einstein
+ *                email:
+ *                  type: string
+ *                  example: albert@princeton.edu
+ *                is_admin:
```
jessicamcinchak, line 501: nit: `is_admin` is defunct now :tada:

api.planx.uk/server.ts
```diff
@@ -450,6 +468,46 @@ app.get("/hasura", async function (_req, res, next) {
   }
 });
 
+/**
+ * @swagger
+ * /me:
```
jessicamcinchak, line 473: should this also include ??
```yml
security:
  - cookieAuth: []
```

api.planx.uk/admin/feedback/downloadFeedbackCSV.ts
```diff
@@ -61,8 +61,6 @@ type ParsedFeedback = Feedback & {
  *    description: Downloads the FeedbackFish CSV
  *    tags:
  *      - admin
- *    security:
- *      - bearerAuth: []
```
jessicamcinchak, line 65: Not sure if I completely understand why the `security` schema key is being completely removed from admin endpoints, rather than replaced with `cookieAuth: []` ?

When an endpoint has this `security` key, swagger renders the :lock: icon which is a nice visual indicator of our authenticated versus public endpoints - I'm wondering if there's a way to keep that visual queue while using this simplified auth strategy? Not a dealbreaker, but just wondering!

jessicamcinchak (approved):

api.planx.uk/admin/feedback/downloadFeedbackCSV.ts
DafyddLlyr: So this is a good catch and an interesting one.

I removed this on all individual endpoints, but added to the root of the docs to apply it across all of them (`server.ts`, line 316)

This does apply the lock icon to most endpoints, but not all for some reason - 

image

The best way to handle this is to just tag each endpoint manually I think - I'll do that now

DafyddLlyr:
In latest push...
 - Rename `cookieAuth` to the more meaningful `userJWT`
 - Add `hasuraAuth` as security schema instead of a parameter
 - Fix `Unauthorised` response schema

PR #2091, opened 2023-08-11
docs: Find and Replace Swagger docs

DafyddLlyr:
Relies on #2090 

Adds Swagger docs for find and replace endpoint. Hitting a few DX pain points with YAML in JSDocs annotation (spacing, no linting/types, text legibility etc). I'm going to take a look sometime next week at splitting out docs into their own YAML / JSON files that can just be referred to in modules before we go too far down the path with the `@swagger` notation 👍

### DIFF
--- a/api.planx.uk/editor/findReplace.ts
+++ b/api.planx.uk/editor/findReplace.ts
@@ -49,6 +49,55 @@ const getMatches = (
   };
 };
 
+/**
+ * @swagger
+ * /flows/{flowId}/search:
+ *  post:
+ *    summary: Find and replace
+ *    description: Find and replace a data variable in a flow
+ *    tags:
+ *      - flows
+ *    parameters:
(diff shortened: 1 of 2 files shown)


jessicamcinchak (approved):
Thanks for this one :raised_hands:

An option I was trying out when documenting admin endpoints before we jump straight to modules is simply moving the `@swagger` JSDoc to where the actual function is defined (so in this case the file containing `findAndReplaceInFlow`) which keeps server.ts a bit less cluttered! Swagger will find & render the same - still doesn't help with the YAML in JS painpoints, but just a thought / would be nice to establish a consistent pattern for this

jessicamcinchak:
Woops noticed that you have a double-swagger definition here now (need to remove from server.ts too) - just in case that's causing the staging deploy failure !

PR #1958, opened 2023-07-11
fix: Handle arrays when updating node data

DafyddLlyr:
New PR cherry-picked from https://github.com/theopensystemslab/planx-new/pull/1952 to avoid rebase headaches! 🍒 

## What does this PR do?
- Fixes a bug in `_update()` where updates in an array in `node.data` is not handled correctly
- This bug results in arrays incorrectly getting a `remove` patch / operation when another property of the node is updated. See video below for demonstration of this on the `TaskList` component - 

https://github.com/theopensystemslab/planx-new/assets/20502206/05a1841a-14f0-4925-ad48-3ffc5fb9040c

## How are node updates handled?
Helpful model to have in mind here - I wasn't super familiar with this aspect of the codebase.

```mermaid
sequenceDiagram
    participant Editor
    participant Immer
    participant convert as convertPatchesToOps()
    participant ShareDB Client
    participant ShareDB Service
    participant Postgres

    Editor->>Immer: Add /remove / update node
    Immer->>convert: Produces patches
    convert->>ShareDB Client: Produces operations
    ShareDB Client->>ShareDB Service: Websockets API
    ShareDB Service->>Postgres: INSERT operations
    ShareDB Service->>Postgres: UPDATE flow

```

PR #2119, opened 2023-08-15
fix: Resolve non-persistent state in `FileUploadAndLabel`

DafyddLlyr:
## Context
So this was an interesting / crazy one...

Before this fix - 
 - When I modified a `FileType` it would persist if I had just added it
 - Subsequent changes to the same `FileType` do not persist, but appear to do so as the local flow is updated

This actually matched behaviour George has described a few times (and maybe Emily?) and we've put it down to ShareDB I think.

## What was happening?
- The local flow was updating _immediately_ as soon as I modified my `FileType` in the modal, no need to "Confirm"
- When saving and exiting the modal, the ShareDB side of things should start to persist flow changes to the db (see https://github.com/theopensystemslab/planx-new/pull/1958 for context)
- However, as the local flow as _already_ up to date, there was no difference between old and existing data and no `ops` were generated 🤯 

## What was the cause?
A pass by value / pass by reference issue I believe. The `FileTypes[]` being passed into Formik wasn't a copy, but a reference to the flow (as it as deeply nested). This meant as your made individual changes in the modal, the local flow was instantaneously updating.

## What's the fix?
- Explicitly copy the data using `cloneDeep()`
- We probably want to watch out for this one elsewhere if we have more complex data types in the Editor Modal in future - it might be safer to explicitly use copies every time we use Formik / `parseContent()`

https://github.com/theopensystemslab/planx-new/assets/20502206/a739d9a5-cca4-4326-b733-2f056ed26a5d

### DIFF
--- a/editor.planx.uk/src/@planx/graph/index.ts
+++ b/editor.planx.uk/src/@planx/graph/index.ts
@@ -408,12 +408,7 @@ const _update = (
 
       // eslint-disable-next-line no-prototype-builtins
       const isRemoved = !isSomething(v) && existingData.hasOwnProperty(k);
-      const isUpdatedArray =
-        Array.isArray(v) && !isEqual(v, (existingData as any)[k]);
-      const isUpdatedPrimitive =
-        typeof v !== "object" && v !== (existingData as any)[k];
-      const isUpdated =
-        isSomething(v) && (isUpdatedPrimitive || isUpdatedArray);
+      const isUpdated = isSomething(v) && !isEqual(v, (existingData as any)[k]);
 
       if (isRemoved) delete (existingData as any)[k];
       if (isUpdated) (existingData as any)[k] = v;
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/graph/index.ts
```diff
@@ -408,12 +408,7 @@ const _update = (
 
       // eslint-disable-next-line no-prototype-builtins
       const isRemoved = !isSomething(v) && existingData.hasOwnProperty(k);
-      const isUpdatedArray =
-        Array.isArray(v) && !isEqual(v, (existingData as any)[k]);
-      const isUpdatedPrimitive =
-        typeof v !== "object" && v !== (existingData as any)[k];
-      const isUpdated =
-        isSomething(v) && (isUpdatedPrimitive || isUpdatedArray);
+      const isUpdated = isSomething(v) && !isEqual(v, (existingData as any)[k]);
```
DafyddLlyr, line 411: This is a non-functional change, just a tidy up I spotted when looking for the cause of the issue.

`isEqual()` can compare anything, so we don't need to worry about separating out primitives and arrays.

PR #2125, opened 2023-08-17
fix: Handle constraints errors more gracefully

DafyddLlyr:
See https://github.com/theopensystemslab/planx-core/pull/98 for more context

### DIFF
--- a/api.planx.uk/package.json
+++ b/api.planx.uk/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@airbrake/node": "^2.1.8",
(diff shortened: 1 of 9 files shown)


jessicamcinchak (approved):
Thanks for debugging here - errors only popped up last week in cases of Scot GIS, correct? Or Planning Data too?

DafyddLlyr:
This error actually only happened on a failed BOPS submission - the GIS query failed, and the payload ended up with an error object in it.

I retried the exact same query to Planning Data and it worked fine - likely a temporary issue on their side.

I'll DM you the data now - have a look at `passport.data._constraints[1]`

PR #2127, opened 2023-08-18
doc(adr): Proposal for modularised Express structure, and refactor of Auth module

DafyddLlyr:
## What does this PR do?
- New ADR - Proposal for modularised Express structure (see ADR for further context!)
- Refactor of auth module as an example
- PR is pointing at `dp/re-implement-team-members-table-and-relationships` as this PR was the spun out of me picking up some changes to the JWT creation. That will follow along in another PR now.


hasura.planx.uk/migrations/1692347940126_drop_table_public_team_members/down.sql
DafyddLlyr, line 1: These two migration files have picked up line ending changes from rebases but there's no content change here.

api.planx.uk/modules/auth/routes.ts
```diff
@@ -0,0 +1,20 @@
...
+router.get("/auth/google", Middleware.useGoogleAuth);
+
+//After authorization, Google will redirect the user back to this route
+router.get(
+  "/auth/google/callback",
+  Middleware.useGoogleCallbackAuth,
+  Controller.handleSuccess,
+);
+
+export default router;
```
jessicamcinchak, line 20: this file feels super clear, definitely seeing the benefits of this approach over our current method of all route defs in server.js and only breaking out service logic :100:

api.planx.uk/modules/auth/strategy/google.ts
```diff
@@ -0,0 +1,23 @@
...
+    if (!jwt) {
+      return done({
+        status: 404,
+        message: `User (${email}) not found.Do you need to log in to a different Google Account?`,
+      } as any);
+    }
+
+    done(null, { jwt });
+  },
+);
```
jessicamcinchak, line 23: Can you share a bit about why you chose to organise this into `strategy/google` rather than adding into controller or service.ts ? Do you imagine `strategy/` will apply to more modules besides auth going forward?

api.planx.uk/server.ts
```diff
@@ -421,11 +247,20 @@ app.use(
...
+
+passport.deserializeUser(function (obj: Express.User, cb) {
+  cb(null, obj);
+});
 app.use(passport.initialize());
 app.use(passport.session());
 app.use(urlencoded({ extended: true }));
 
-app.use("/auth", router);
+app.use(authRoutes);
```
jessicamcinchak, line 263: really satisying diff in this file :muscle:

jessicamcinchak (approved):
Apologies it's taken me awhile to get to this one - but I'm very on board. The ARD is clear and there aren't any sticking points for me, agree it's a structure we'll keep chipping away at rather than one big rewrite which feels realistic given capacity/priorities. 

One small question here but no blockers to merging!

api.planx.uk/server.ts
DafyddLlyr: 💯

api.planx.uk/modules/auth/strategy/google.ts
DafyddLlyr: This is a great question.

It was really moved here just for organisation - it's part of the `auth` module, but not really a middleware or controller, just a "strategy" used by `Passport`. I'm not 100% sure but I think we'll end up with another [Passport Strategy](https://www.passportjs.org/concepts/authentication/strategies/) here for Microsoft SSO.

It might be simpler / more meaningful to group all `passport` logic into `/auth/passport` or something in future as opposed to having some `app.use()` functions using `passport` in `server.ts`

jessicamcinchak: That's helpful context, thanks ! Let's leave as this for now but maybe revisit organisation in future as needed ?

PR #2128, opened 2023-08-18
chore: Setup husky and lint-staged for API and E2E projects

DafyddLlyr:
## What does this PR do?
- Add pre-commit hooks (using husky and lint-staged) to the API and E2E projects

I did consider configuring this at the root level of the project, [which lint-staged supports](https://github.com/okonet/lint-staged#how-to-use-lint-staged-in-a-multi-package-monorepo), but for now I'm happy steering away from any mono-repo tooling unnecessarily.


jessicamcinchak (approved):
Thanks for this one! :broom:

PR #2164, opened 2023-08-23
feat: Auth module documentation

DafyddLlyr:
## What does this PR do?
 - Demonstrates how we could use separate Swagger documents within a modularised structure
 - Adds basic API docs for auth module

### Why keep `swagger-jsdoc`?
- It allows us to migrate the existing docs slowly
- It's actually doing a bit of helpful heavy lifting - without this we'd be need to combine and parse multiple YAML files on build which will probably lead to poor DX (slower to work on files and then see changes in browser)
- The small downside is that each `docs.yaml` needs to be a full OpenAPI file. In reality that just means it needs the following minimal heading - 

```yaml
openapi: 3.1.0
info:
  title: Plan✕ API
  version: 0.1.0
paths: {}
```

### Why YAML and not JSON?
Not a strong preference! Being able to comment in YAML files is really the only real difference which usually proves really helpful.

Linting tools / plugins in your IDE can help make life easy regardless on the format we settle on, and the `failOnErrors: true` param from `swagger-jsdoc` should also be keeping us straight here.

### Alternatives
 - Tspec looks like a nice option that lets us leverage TypeScript a little more here https://ts-spec.github.io/tspec/


DafyddLlyr:
Moved to draft - going to have a closer look at an implementation using Tspec as it looks like a solid option here

DafyddLlyr:
Verdict - Tspec looks fantastic as an option, but the project doesn't yet have the right level of maturity for us - a few show stopping bugs, no tests in the repo, and limited options in taking some types/annotations to OpenAPI. 

One to keep an eye on for sure 👍

api.planx.uk/docs/schemas.yaml
```diff
@@ -0,0 +1,18 @@
...
+      type: object
+      properties:
+        payload:
+          type: object
+          properties:
+            sessionId:
+              type: string
+      example:
+        payload:
+          sessionId: "123"
```
jessicamcinchak, line 18: Similar to above, it seems like the "Schemas" section isn't actually rendering on the pizza yet: https://api.2164.planx.pizza/docs/ versus https://api.editor.planx.dev/docs/#/

api.planx.uk/modules/auth/docs.yaml
```diff
@@ -0,0 +1,46 @@
...
+        "200":
+          description: OK
+  /auth/google/callback:
+    get:
+      summary: Generate a JWT for an authenticated user
+      description: After authentication, Google will redirect the user back to this route which generates a JWT for the user
+      tags: ["auth"]
+      responses:
+        "200":
+          description: OK
```
jessicamcinchak, line 46: I'd expect to see these on the pizza `/docs` but I don't?

jessicamcinchak (commented):
Solid re-organisation plan, but don't think all the dots are quite connecting / rendering correctly yet - let me know if you need extra eyes !

DafyddLlyr:
Thanks for taking a look here - I'm sure this was working locally at last push but I'll rebase and test - sorry for that..!

DafyddLlyr:
So this does work as expected locally, but something in the build step is breaking this which I can't replicate currently. Looking into it!

DafyddLlyr:
Fixed! 

TIL: The build step for the API only complies `.ts` -> `.js` so any other files will be excluded - a bit of a head scratcher! I'm not sure if this explains some of the HTML headaches you were previously having but it's something to bear in mind.

DafyddLlyr:
@jessicamcinchak Now ready for re-review - schemas, responses, and parameters no longer YAML but encoded in the config object directly.

jessicamcinchak (approved):

PR #2139, opened 2023-08-21
feat: Add `platformAdmin` role to Hasura

DafyddLlyr:
## What does this PR do?
 - Adds the `platformAdmin` role to Hasura
 - Adds new tables, `user_roles` and `user_role_enum` and sets up relationships

## Next steps...
 - Grant all users the `platformAdmin` role (essentially the current "god mode")
 - Add role to JWT - probably with helpers in `planx-core`, e.g. `admin.user.getById(123).roles`
 - Ensure JWT is used for all frontend and backend calls to Hasura (frontend should be correct currently, backend uses `HASURA_ADMIN_SECRET` frequently)

PR #2167, opened 2023-08-24
feat: Add `user.is_platform_admin` and `team.role` columns

DafyddLlyr:
A simplified re-working of https://github.com/theopensystemslab/planx-new/pull/2139

 - PlatformAdmin is a type of user
 - Roles are team based


hasura.planx.uk/migrations/1692609979701_create_table_public_user_role_enum/up.sql
```diff
@@ -0,0 +1,23 @@
...
+  UNIQUE ("value")
+);
+
+COMMENT ON TABLE "public"."user_roles" IS E'Enum of possible user roles in PlanX';
+
+INSERT INTO
+  "public"."user_roles"("value")
+VALUES
+  (E'admin'),
+  (E'viewer');
```
jessicamcinchak, line 13: I'm a little confused - is `admin` here the same as Hasura's reserved `admin`? 

My original understanding was that Hasura reserved roles `admin` & `public` won't actually be included in this user_roles enum, but things like `Flow Editor`, `Flow Viewer` would be.

Can we perhaps align on an MVP of role names before this is merged and try to update originally proposed table accordingly? https://docs.google.com/document/d/1Vui62_7iHuT60-FrMJATrhqsj2nQZ71N5fEX1nfZnvU/edit

hasura.planx.uk/migrations/1692863825415_alter_table_public_users_add_column_isPlatformManager/up.sql
```diff
@@ -0,0 +1,4 @@
+alter table "public"."users" add column "is_platform_admin" boolean
+ not null default 'false';
+
+comment on column "public"."users"."is_platform_admin" is E'A Platform Admin is the highest level of permission in PlanX, and can operate across all teams';
```
jessicamcinchak, line 4: :+1:

hasura.planx.uk/tests/lowcal_sessions.test.js
```diff
@@ -428,4 +428,19 @@ describe("lowcal_sessions", () => {
...
+    });
+
+    test("cannot query lowcal_sessions", () => {
+      expect(i.queries).not.toContain("lowcal_sessions");
+    });
+
+    test("cannot create, update, or delete lowcal_sessions", () => {
+      expect(i).toHaveNoMutationsFor("lowcal_sessions");
+    });
+  });
```
jessicamcinchak, line 445: Updated introspection tests are very helpful to quickly understand intended access-level, thanks !  

One clarification though, when I sign into this pizza's Hasura console with my user, who is _not_ a Platform Admin according to the `teams` table, I can indeed query tables like `lowcal_sessions` - is that expected? I guess so because of these headers, which sort of still inherently contradict value of `teams.is_platform_admin` ?? 
```
"x-hasura-allowed-roles": ["platformAdmin"],
"x-hasura-default-role": "platformAdmin",
```

jessicamcinchak (approved):
Looks good ! A couple questions for clarification but no showstoppers

hasura.planx.uk/migrations/1692609979701_create_table_public_user_role_enum/up.sql
DafyddLlyr: Apologies - this is me being a bit clumsy with naming things. I've updated this to `teamAdmin` and `teamViewer`.

The intention was to later map admin -> teamAdmin, and viewer -> teamViewer when building the JWT but we can skip that step by just using consistent naming. The reasoning for this was that the `team` prefix was a bit superfluous in the context of this table but looking back it's just unclear / confusing.

If we end up with flow-level permissions in addition to team-level permissions, we would then be able to add a `flow_id` column here (or make another table for this?) and then have `flowViewer` and `flowAdmin` enums added.

hasura.planx.uk/tests/lowcal_sessions.test.js
DafyddLlyr: This is a good question.

When we log into the console, we always use the `admin` role - you're never logged in as your user and have "god mode".

However, using the GraphiQL playground in Hasura you can uncheck the `x-hasura-admin-secret` header, and add a `x-hasura-role` header (set to platformAdmin) and then queries and mutations will reflect this permission level.

The follow up step to this PR would be to make sure we always pass the correct headers to Hasura (it's hardcoded to the admin secret in `planx-core` currently). Requests made via Apollo in the frontend already correctly pass your JWT through to Hasura.

jessicamcinchak: Makes sense now, thanks !

PR #2171, opened 2023-08-25
fix: include new `teams.boundary` column in data sync

jessicamcinchak:
Updated a bunch of boundaries on production, but realised staging is still using default UK extent value until this column is explicitly synced. 

To test: check that this pizza builds with updated values !


jessicamcinchak:
Will close in favor of https://github.com/theopensystemslab/planx-new/pull/2164/commits/6063f07dd52f7894d77d2d072410c3a716d9aa27, this was likely erroring on `user.is_admin`

PR #2175, opened 2023-08-29
feat: Team module

DafyddLlyr:


api.planx.uk/shared/middleware/validate.ts
```diff
@@ -0,0 +1,39 @@
...
+  ) => {
+    try {
+      schema.parse({
+        params: req.params,
+        body: req.body,
+        query: req.query,
+      });
+      return next();
+    } catch (error) {
+      return res.status(400).json(error);
```
DafyddLlyr, line 23: We could do a lot more with this error, like format it in a more meaningful way with https://github.com/causaly/zod-validation-error but a raw Zod error is meaningful enough for our current API users.

api.planx.uk/shared/middleware/validate.ts
```diff
@@ -0,0 +1,39 @@
...
+import { AnyZodObject, ZodSchema, z } from "zod";
+
+/**
+ * Middleware to validate incoming requests to the API
+ * Takes a ZodSchema and returns a validated, and typed, Request object
+ */
+export const validate =
+  (schema: AnyZodObject) =>
+  async (
+    req: Request<z.infer<typeof schema>>,
```
DafyddLlyr, line 11: I think this may not be _perfectly_ type safe as it's not using a generic, but it's safe enough that we can trust it when we type out Express request handlers with `ValidatedRequestHandler<YourSchema, YourResponse>`👌 

I spent a bit of time trying to get this to use this type and it ended up being a bit of a rabbit hole so I'm happy with this for now.

api.planx.uk/modules/team/index.test.ts
```diff
@@ -0,0 +1,259 @@
...
+        userId: 123,
+      })
+      .expect(400)
+      .then((res) => {
+        expect(res.body).toHaveProperty("issues");
+        expect(res.body).toHaveProperty("name", "ZodError");
+      });
+  });
+
+  it("validates that role must one an accepted value", async () => {
```
jessicamcinchak, line 60: nit: typo here!

api.planx.uk/server.ts
```diff
@@ -188,6 +189,7 @@ app.use(passport.session());
 app.use(urlencoded({ extended: true }));
 
 app.use(authRoutes);
+app.use("/team", teamRoutes);
```
jessicamcinchak, line 192: I'd expect that we want `useJWT` here similar to admin endpoints defined like this ? 
```
app.use("/admin", useJWT);
app.get("/admin/feedback", downloadFeedbackCSV);
```

jessicamcinchak (approved):
Thanks for these !

api.planx.uk/server.ts
DafyddLlyr: Good catch - it's actually still there, I've just moved the responsibility to the routes module.

`router.use(AuthMiddleware.useJWT);` in `modules/team/routes.ts`

DafyddLlyr:
To confirm - Cloudflare is currently restricting access to `/team` endpoints also 👍

api.planx.uk/modules/team/index.test.ts
DafyddLlyr: Picked this up in the next PR - thank you

PR #2177, opened 2023-08-29
feat(ci): Throw error on sync script failure, wrap in single transaction

DafyddLlyr:
Taken a few attempts but it's working as expected now.

The issue was that psql was not throwing an error exit code when it was hitting an error, and also that the GitHub action was not looking for an error code.

From the psql docs - 

> psql returns 0 to the shell if it finished normally, 1 if a fatal error of its own occurs (e.g., out of memory, file not found), 2 if the connection to the server went bad and the session was not interactive, and 3 if an error occurred in a script and the variable ON_ERROR_STOP was set.

https://www.postgresql.org/docs/current/app-psql.html#ON-ERROR-STOP

Now, we're doing...
 - A try/catch in GitHub actions
 - Building a single SQL file to execute using `cat`...
 - ...and wrapping it in a single, error throwing, transaction with the `--single-transaction -v ON_ERROR_STOP=on"` flags which automatically rolls back on failure


scripts/seed-database/write/main.sql
```diff
@@ -1,6 +1,4 @@
-BEGIN;
 \include write/users.sql
 \include write/teams.sql
 \include write/flows.sql
-\include write/flow_document_templates.sql
-COMMIT;
+\include write/flow_document_templates.sql
```
jessicamcinchak, line 4: should `published_flows` be included on this list too ? i think it's a bit more clear to have a reference of all synced tables in same list and would mean we can remove this line (i think) https://github.com/theopensystemslab/planx-new/pull/2177/files#diff-ac583dce0ccc6237bde4c2e74af81ce849d30a30fc84beba82b143b6a2dc5617R43

scripts/seed-database/container.sh
```diff
@@ -33,20 +33,20 @@ for table in "${tables[@]}"; do
 done
 
 if [[ ${RESET} == "reset_flows" ]]; then
-  psql ${LOCAL_PG} --command="TRUNCATE TABLE flows CASCADE;"
+  cat write/truncate_flows.sql > sync.sql
 fi
 
 if [[ ${INCLUDE_PUBLISHED_FLOWS} == "include_published_flows" ]]; then
   psql --quiet ${REMOTE_PG} --command="\\copy (SELECT DISTINCT ON (flow_id) id, data, flow_id, summary, publisher_id, created_at FROM published_flows ORDER BY flow_id, created_at DESC) TO '/tmp/published_flows.csv' (FORMAT csv, DELIMITER ';');"
   echo published_flows downloaded
```
jessicamcinchak, line 41: while we're adjusting the sync, i personally don't agree that there's ever a good reason to _not_ include published flows and would be in favor of removing this conditional - which adds extra complexity like updating synced columns in two places for this particular table. 

understand you didn't introduce it here, but would help simplify concatenation i think / see other comment!

.github/workflows/sync-staging-db.yml
```diff
@@ -33,6 +33,23 @@ jobs:
         with:
           files: ".env, .env.staging, api.planx.uk/.env.test, hasura.planx.uk/.env.test"
           fail: true
-      - run: docker run --rm -v "$(pwd)/scripts/seed-database:/app" --workdir="/app" postgis/postgis:12-3.0-alpine "./container.sh" "${STAGING_PG_URL}" "${PRODUCTION_PG_URL_FOR_USER_GITHUB_ACTIONS}" "reset_flows" "include_published_flows"
+      - name: Run sync script
+        run: | 
+          {
+            docker run --rm -v "$(pwd)/scripts/seed-database:/app" --workdir="/app" postgis/postgis:12-3.0-alpine "./container.sh" "${STAGING_PG_URL}" "${PRODUCTION_PG_URL_FOR_USER_GITHUB_ACTIONS}" "reset_flows" "include_published_flows"
+          } || {
+            echo "Database sync failed!"
```
jessicamcinchak, line 41: :100:

jessicamcinchak (approved):

scripts/seed-database/container.sh
```diff
@@ -32,21 +29,19 @@ for table in "${tables[@]}"; do
   fi
 done
 
-if [[ ${RESET} == "reset_flows" ]]; then
-  psql ${LOCAL_PG} --command="TRUNCATE TABLE flows CASCADE;"
-fi
+psql --quiet ${REMOTE_PG} --command="\\copy (SELECT DISTINCT ON (flow_id) id, data, flow_id, summary, publisher_id, created_at FROM published_flows ORDER BY flow_id, created_at DESC) TO '/tmp/published_flows.csv' (FORMAT csv, DELIMITER ';');"
+echo published_flows downloaded
```
jessicamcinchak, line 33: are these lines still necessary? published_flows should be in concatened command below, and I _think_ fullly covered by write/published_flows.sql now?

jessicamcinchak: resolved this in slack!

PR #2182, opened 2023-08-31
feat: `teamEditor` role

DafyddLlyr:
# What does this PR do?
- Adds the `teamEditor` role in Hasura, and associated introspection tests (more on this below)
- Builds a JWT token on login which represent the user's role
- Adds a helper script in `/hasura.planx.uk/tests` to generate a token
- A few necessary changes in the Editor to allow `platformAdmin` and `teamEditor` users to access public data

It's certainly a lot more than I wanted to put in one PR sorry, but things would be broken without joining up the dots a little here.

## What's going on with the tests?
I hit an issue with the introspection tests - if a user has access to a query/mutation, but it's execution is conditional (i.e. permission based) then an introspection test doesn't mean much. We'd need to execute a mutation or query to really test if a user has permission - and to do that we'd need to insert records to mutate, and insert users, teams, team member records etc... It gets heavy pretty fast and it's basically repeating a lot of boilerplate from `/e2e/tests/api-driven`.

I've still added (coarse-grained) introspection tests for the new role for new measure, but it's important to state clearly that these aren't actually testing the code we care most about (the permission checks defined in `tables.yaml`).

## How are we going to test this?
I'd probably vote we create test suites `/e2e/tests/api-driven` that properly test roles at a deeper level (actually trying to execute queries and mutations). These should be regression tests that run nightly as opposed to every PR.

I think given the repetitive nature of them (query X table as Y role, mutate X table as Y role) the Cucumber tests Ben has already set up are probably well suited to this task and will allow to to get a large number of meaningful tests with a relatively small amount of code. I'm super happy to pick this up the week I'm back.

I consider not covering these high importance code areas very far from ideal, but honestly it's actually super low risk - introspection tests cover the show stoppers (mutations which delete) and we're moving from god-mode so everything is an improvement!

## Next steps...
 - I've asked @Alastairparvin to assign users to teams in a CSV. When I get this back I'll update records on prod.
 - Every team member will be granted `teamEditor` status
 - Stop using `x-hasura-admin-secret` in the API and `planx-core`
 - Frontend implementation

### DIFF
--- a/api.planx.uk/modules/auth/service.ts
+++ b/api.planx.uk/modules/auth/service.ts
@@ -29,7 +29,7 @@ const getAllowedRolesForUser = (user: User): Role[] => {
   const teamRoles = user.teams.map((teamRole) => teamRole.role);
   const allowedRoles: Role[] = [
     "public", // Allow public access
-    "teamAdmin", // Least privileged role for authenticated users - required for Editor access
+    "teamEditor", // Least privileged role for authenticated users - required for Editor access
     ...teamRoles, // User specific roles
   ];
   if (user.isPlatformAdmin) allowedRoles.push("platformAdmin");
@@ -45,5 +45,5 @@ const getAllowedRolesForUser = (user: User): Role[] => {
  * This is the role of least privilege for the user
  */
 const getDefaultRoleForUser = (user: User): Role => {
-  return user.isPlatformAdmin ? "platformAdmin" : "teamAdmin";
+  return user.isPlatformAdmin ? "platformAdmin" : "teamEditor";
 };
--- a/api.planx.uk/modules/auth/service.ts
+++ b/api.planx.uk/modules/auth/service.ts
@@ -1,34 +1,49 @@
 import { sign } from "jsonwebtoken";
-import { adminGraphQLClient as adminClient } from "../../hasura";
-import { gql } from "graphql-request";
+import { $admin } from "../../client";
+import { User, Role } from "@opensystemslab/planx-core/types";
 
-export const buildJWT = async (email: string | undefined) => {
-  const { users } = await adminClient.request(
-    gql`
-      query ($email: String!) {
-        users(where: { email: { _eq: $email } }, limit: 1) {
-          id
-        }
-      }
-    `,
-    { email },
-  );
-
-  if (!users.length) return;
-
-  const { id } = users[0];
-
-  const hasura = {
-    "x-hasura-allowed-roles": ["platformAdmin", "public"],
-    "x-hasura-default-role": "platformAdmin",
-    "x-hasura-user-id": id.toString(),
-  };
+export const buildJWT = async (email: string): Promise<string | undefined> => {
+  const user = await $admin.user.getByEmail(email);
+  if (!user) return;
 
   const data = {
-    sub: id.toString(),
-    "https://hasura.io/jwt/claims": hasura,
+    sub: user.id.toString(),
+    "https://hasura.io/jwt/claims": generateHasuraClaimsForUser(user),
   };
 
   const jwt = sign(data, process.env.JWT_SECRET!);
   return jwt;
+};
+
+const generateHasuraClaimsForUser = (user: User) => ({
+  "x-hasura-allowed-roles": getAllowedRolesForUser(user),
+  "x-hasura-default-role": getDefaultRoleForUser(user),
+  "x-hasura-user-id": user.id.toString(),
+});
+
+/**
+ * Get all possible roles for this user
+ * Requests made outside this scope will not be authorised by Hasura
+ */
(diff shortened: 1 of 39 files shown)


editor.planx.uk/src/lib/graphql.ts
```diff
@@ -87,3 +88,13 @@ export const client = new ApolloClient({
   link: from([retryLink, errorLink, authLink, httpLink]),
   cache: new InMemoryCache(),
 });
+
+/**
+ * Explicitly connect to Hasura using the "public" role
+ * Allows authenticated users with a different x-hasura-default-role (e.g. teamEditor, platformAdmin) to access public resources
+ */
+export const publicContext: DefaultContext = {
```
DafyddLlyr, line 96: What's going on here?

In order to keep things a little simpler, the newly created roles do not have the same set of permissions as the public role. This would mean any change to the public role (e.g. new tables and columns) would need to get propagated across all other roles. Maintenance would be hard and not much fun! I had hoped by just adding the `public` role to a user's `x-hasura-allowed-role` it would automatically fall back to this, but there's no concept of role "hierarchy" like this.

In a perfect would it would be nice to user [inherited roles](https://hasura.io/docs/latest/auth/authorization/inherited-roles/) (meaning the actual `teamAdmin` role could be made of `public` and an abstract `_teamAdmin`. I tried getting this to work but changes made in Hasura weren't persisting and updating any metadata files. We might need to [update Hasura to the metadata v3 config](https://hasura.io/docs/latest/migrations-metadata-seeds/legacy-configs/upgrade-v3/) for this to work.

I'm happy with this solution for now - we always know what's a public query.

hasura.planx.uk/seeds/1595528762802_teams_and_users.sql
DafyddLlyr, line 1: Just realised - I think this might be redundant now that we just sync prod data?

editor.planx.uk/src/lib/graphql.ts
DafyddLlyr: I think the approach in https://github.com/theopensystemslab/planx-new/pull/2195 of just using an `authLink` and 
`publicLink` is a little simpler than modifying the context.

Once the other PR is merged I'll rebase this to adopt that approach.

hasura.planx.uk/tests/scripts/buildJWT.js
```diff
@@ -0,0 +1,17 @@
...
+ *   Useful for testing permissions in the Hasura console at a finer-grained level than the introspection tests currently allow.
+ * @usage pnpm build-jwt <ROLE> <USER_ID>
+ * @example pnpm build-jwt teamEditor 123
+ * @returns A JWT which can be copy/pasted to the Hasura admin console as a the "authorization" header
+ */
+const buildJWT = () => {
+  const [role, userId] = process.argv.slice(2);
+  const jwt = buildJWTForRole(role, userId);
+  console.log(`Bearer ${jwt}`)
+}
```
jessicamcinchak, line 15: this is a handy utility method, thanks :raised_hands:

jessicamcinchak (approved):
Changes here all look good to me - thanks for thorough PR description and explaining limitation of introspection tests here, a nightly regression suite "next" makes sense !

api.planx.uk/modules/auth/service.ts
```diff
@@ -1,34 +1,49 @@
...
+ * Get all possible roles for this user
+ * Requests made outside this scope will not be authorised by Hasura
+ */
+const getAllowedRolesForUser = (user: User): Role[] => {
+  const teamRoles = user.teams.map((teamRole) => teamRole.role);
+  const allowedRoles: Role[] = [
+    "public", // Allow public access
+    "teamEditor", // Least privileged role for authenticated users - required for Editor access
+    ...teamRoles, // User specific roles
+  ];
```
jessicamcinchak, line 34: is `teamViewer` intentionally omitted here? Not sure I totally understand why

DafyddLlyr: This is a good question and something I don't think I've spelt out anywhere really.

`teamViewer` is intentionally left out here as it's not currently backed up by a Hasura role. We could do with talking about how we handle `teamViewer` and what this actually means.

 - `teamEditor` grants read-only permissions to all flows/operations/settings etc, but only insert, update, and delete when you're a member of that team
 - This pretty much describes what a `teamViewer` should be? Read only access to all teams - we don't need a relationship to a specific team. Having a record for every user and every team in `team_members` wouldn't be necessary.
 
 So here's my proposal - 
  - Drop `teamViewer` from the `user_roles` enum table - it doesn't really make sense as it doesn't require a relationship to a team
  - Make a `readOnly` role for people like DLUHC users
  - This would be the new default for logged in users

PR #2183, opened 2023-08-31
fix: Correct `cat` syntax (again...)

DafyddLlyr:
Apologies for the churn here....

Tested and works as expected locally. Pizza not skipped so we can see check the db.


jessicamcinchak (approved):
Pizza looks good !

PR #1748, opened 2023-06-06
chore: add Doncaster's newest A4 to GIS schema

jessicamcinchak:
:tada: This one's finally ready! 

In adding the new A4, it also seems Digital Land has updated (duplicated?) the reference to the original HMO A4 too - so I've gone with the most recent references available for each: https://www.planning.data.gov.uk/entity/?dataset=article-4-direction-area&geometry_curie=statistical-geography%3AE08000017&entries=all&entry_date_day=&entry_date_month=&entry_date_year=

Here's the [A4 spreadsheet for reference too](https://docs.google.com/spreadsheets/d/1SRS6EGl5f4uw1cvzkm65jEBna4w6IwGSeaE3v8L84to/edit#gid=1017787825).

PR #2184, opened 2023-08-31
Production deploy

DafyddLlyr:


jessicamcinchak (approved):
Can we get #1748 in here too please before merge?

PR #2195, opened 2023-09-01
feat: Explicitly use `publicClient` in frontend

DafyddLlyr:
Please see https://github.com/theopensystemslab/planx-new/pull/2190 and https://github.com/theopensystemslab/planx-new/pull/2182#discussion_r1312254263 for context.

This PR introduces the following change - 
 - Explicitly use a `publicClient` - this means that the `public` role will be used for these requests

### DIFF
--- a/editor.planx.uk/src/lib/graphql.ts
+++ b/editor.planx.uk/src/lib/graphql.ts
@@ -4,7 +4,6 @@ import {
   from,
   InMemoryCache,
 } from "@apollo/client";
-import { setContext } from "@apollo/client/link/context";
 import { onError } from "@apollo/client/link/error";
 import { RetryLink } from "@apollo/client/link/retry";
 import { toast } from "react-toastify";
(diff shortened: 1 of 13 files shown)


editor.planx.uk/src/lib/graphql.ts
```diff
@@ -83,7 +72,19 @@ const retryLink = new RetryLink({
   },
 });
 
+/**
+ * Client used to make all requests by authorised users
+ */
 export const client = new ApolloClient({
```
DafyddLlyr, line 78: I felt keeping the simple `client` name was probably more appropriate and concise compared to `authorisedClient` or similar. This will be used for almost all frontend calls as most of the complexity is within the Editor.

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -221,7 +221,7 @@ export const editorStore: StateCreator<
   deleteFlow: async (teamId, flowSlug) => {
     const response = await client.mutate({
       mutation: gql`
-        mutation MyMutation($team_id: Int, $flow_slug: String) {
+        mutation DeleteFlow($team_id: Int, $flow_slug: String) {
```
jessicamcinchak, line 224: thanks for tidying these names up along the way!

jessicamcinchak (approved):

PR #2198, opened 2023-09-03
feat: Use roles in API

DafyddLlyr:


api.planx.uk/editor/publish.ts
```diff
@@ -107,7 +101,7 @@ const publishFlow = async (
         {
           data: flattenedFlow,
           flow_id: req.params.flowId,
-          publisher_id: parseInt(req.user.sub, 10),
+          publisher_id: parseInt(req.user!.sub!, 10),
```
DafyddLlyr, line 104: The next PR will handle this a bit more elegantly than forcing `!`

api.planx.uk/modules/auth/middleware.ts
```diff
@@ -109,3 +110,58 @@ export const useGoogleCallbackAuth: RequestHandler = (req, res, next) => {
...
+/**
+ * Validate that an incoming request is using the role required for an endpoint
+ * Wrapped by the useJWT middleware to ensure token is valid, available, and decoded
+ *
+ * This does not check if a user can ultimately access a resource, only that they can access this route
+ * Hasura will validate this on a row and column basis when the query or mutation is made
+ */
+export const useRoleAuth: UseRoleAuth =
+  (authRoles) => async (req, res, next) => {
+    useJWT(req, res, () => {
```
DafyddLlyr, line 125: I spent a while trying to get these two middlewares a little less tightly coupled. The Express docs have a nice example of using an array of middleware - https://expressjs.com/en/guide/using-middleware.html

The issue I was hitting was keeping the return types for `req, res, next` outside this file when passing in the `role` argument to the middleware / function which returns a middleware array.

In reality they are tightly coupled - we can't check roles without decoding and validating the JWT, but I don't think this reads particularly clearly.

api.planx.uk/modules/auth/middleware.ts
```diff
@@ -109,3 +110,58 @@ export const useGoogleCallbackAuth: RequestHandler = (req, res, next) => {
...
+
+      next();
+    });
+  };
+
+// Convenience methods
+export const useTeamViewerAuth = useRoleAuth([
+  "teamViewer",
+  "teamEditor",
+  "platformAdmin",
```
DafyddLlyr, line 164: These arrays indicate that any of these roles will be valid here. See comment on inherited roles here - https://github.com/theopensystemslab/planx-new/pull/2182#discussion_r1312254263

In future, PlanX roles could be less hierarchical so this approach would still work there also.

api.planx.uk/modules/auth/middleware.ts
```diff
@@ -109,3 +110,58 @@ export const useGoogleCallbackAuth: RequestHandler = (req, res, next) => {
...
+
+      const userId = req.user.sub;
+      // Check if a user has any of the roles required for this route
+      const isAuthorised = userRoles.some((role) => authRoles.includes(role));
+
+      if (!isAuthorised) {
+        console.error(
+          `Authentication error: User ${userId} does have have any of the roles [${authRoles.join(
+            ", ",
+          )}] which are required to access ${req.path}`,
```
DafyddLlyr, line 148: We can probably drop this in future but it could prove handy for debugging any initial issues we have implementing roles.

api.planx.uk/server.ts
```diff
@@ -248,11 +250,7 @@ app.get("/hasura", async function (_req, res, next) {
  *                  type: string
  *                  example: 2023-08-11T11:28:38.237493+00:00
  */
-app.get("/me", useJWT, async function (req, res, next) {
-  // useJWT will return 401 if the JWT is missing or malformed
-  if (!req.user?.sub)
-    next({ status: 401, message: "User ID missing from JWT" });
-
+app.get("/me", usePlatformAdminAuth, async function (req, res, next) {
```
jessicamcinchak, line 253: will any signed in user (including non-platform admins) still be able to call `/me` if we're using this auth method here? if not, do we want that to be possible? thinking forward to the `/me` endpoint perhaps underpinning a frontend user store & ability to display your use rdetails in the editor?

jessicamcinchak (approved):

api.planx.uk/server.ts
DafyddLlyr: Yeah this is a good catch - I added `usePlatformAdminAuth` as I was viewing this as a nice way for devs to check their status.

I'll make it accessible to any logged on user again on a follow up PR 👍 

For a user store, we could populate it a number of ways - 
- Calling `/me`
- Calling `$admin.user.getByEmail()` directly
- Enriching the JWT with all required user details

I think you're right here - `/me` may be the simplest method

PR #2209, opened 2023-09-06
fix: `FileUploadAndLabel` modal should error if each file is not tagged, but not validate that all required files are uploaded yet

jessicamcinchak:
Splits up the validation checks for "each file should be tagged" & "all required files must be tagged" so that we can call _only_ the first check from the Modal when clicking "Done". This hopefully makes for any easier to understand user journey than this original feedback that we went back on forth on: https://github.com/theopensystemslab/planx-new/pull/1994

Other changes:
- Disables outside clicks to close the modal, ESC key still okay
- Changes all user-facing instances of "tag" to "label" in hardcoded error messages, etc

Behavior before: 
- Modal opens, you can click "Done" to close without tagging any files, no error
- Modal opens, you can click away to close without tagging any files, no error
- Modal opens, you can press ESC to close without tagging any files, no error
- You only see errors on "Continue" from main component

After: 
- Modal opens, you can click "Done" to close without tagging any files, you'll see error "Please tag all files" 
- Modal opens, you can click "Cancel" to close without tagging any files, you will not see a modal error
- Modal opens, you cannot click away to close whether or not you've tagged files yet
- Modal opens, you can press ESC to close without tagging any files, no error
- You continue to see errors on "Continue" from the main component
  - Order of error messages is: upload at least one file, all files need tags (for cases where modal is closed by "Cancel" or ESC), upload & tag all _required_ files

Testing: https://2209.planx.pizza/testing/test-upload-and-label/preview

Outstanding feedback / suggested followup PR:
- More robost click-handling for "Cancel" button: can it fully remove the actual uploaded files that triggered that modal session rather than just close? See feedback
- Address label persistance bug - this is re-creatable on staging too, so not introduced here - made a followup Trello ticket: https://trello.com/c/Kc3ifwND/2632-upload-label-labels-occassionally-disapear-in-the-modal-when-changed-multiple-times-in-a-row

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Modal.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Modal.tsx
@@ -115,10 +115,18 @@ export const FileTaggingModal = ({
             <Button
               variant="contained"
               onClick={handleValidation}
-              sx={{ paddingLeft: 2 }}
               data-testid="modal-done-button"
             >
               Done
+            </Button>
+            <Button
+              variant="contained"
+              color="secondary"
+              sx={{ ml: 1.5 }}
+              onClick={closeModal}
+              data-testid="modal-cancel-button"
+            >
+              Cancel
             </Button>
           </Box>
         </ErrorWrapper>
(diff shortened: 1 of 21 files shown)


jessicamcinchak:
[[Investigate if we can] Make users label file before returning to main upload and label component](https://trello.com/c/IryaS9Yo/2582-investigate-if-we-can-make-users-label-file-before-returning-to-main-upload-and-label-component)

ianjon3s (approved):
Works as per specification, all good.

ianjon3s:
Further changes work good, with the note that we can't deploy this until the cancel button is in place (as we've removed the click outside to close).

I'm now getting an error with the validation (this may have already been in place), here's a video to demonstrate:

Uploading Screen Recording 2023-09-06 at 13.15.04.mov…

ianjon3s:
Yes @jessicamcinchak, makes sense to pick the already existent bug up in a separate task in order to move this forward.

PR #2214, opened 2023-09-08
feat: setup `UserStore` and basic visual feedback throughout editor based on access

jessicamcinchak:
## Changes
- Adds a `UserStore` with simple state management getters & setters for working with users & their roles within all authenticated routes
- Adds visual feedback throughout the editor UI to indicate if you have edit- or view-access for that team (this is crude and I'm not a designer! but hopefully gives us some starting ideas to improve on now :slightly_smiling_face:)

### Teams
- Shows an icon next to each team to indicate edit- (:pen:) or view- (:eye:) access (in the future, we may want to think about ordering so edit-access is at the top etc)
Screenshot from 2023-09-22 10-27-15

### Team
- Shows an icon in the team header to indicate edit- or view-access (we only restrict at team-level right now, not flows)
- If view-access, hides the copy/move/rename/delete menu & the "add a new service" button at the bottom of the list of services
- If edit-access, all options display as "normal" 
Screenshot from 2023-09-22 10-30-42

### Flow
- If view-access, disables the "check for changes to publish" button
- If view-access, hides all `<Hanger />` points in the graph which makes it impossible to move, add or clone nodes (in the UI at least, not the actual `operations` table!)
- If view-access, I can still navigate through external portals, and that external portal flow loads correctly depending on my permissions to its' team
- If edit-access, all options display as "normal"
- It shows an icon next to the flow name in the header regardless of access
- It shows my real initial (!) regardless of access
Screenshot from 2023-09-22 13-28-40

### Node
- If view-access, disables the delete/make unique/update buttons in the node dialog. You can still close using the upper "X" or clicking outside
- If view-access, disables the "image upload" icon preventing new images from being uploaded and disables the "remove" item in the menu of existing uploaded images
- If view access, disables the "add new", "drag/re-order" and "remove" icons for list options 
- If view access, inputs are not disabled to allow easier copy + paste, but no changes will persist or be written to the DB
Screenshot from 2023-09-26 14-11-49

** I learned our original hypothesis about not connecting to ShareDB altogether, rather than restricting per operation method, doesn't actually work very well here because the `connect` & `connectTo` Editor store methods are actually responsible for loading the whole flow - so if we don't proceed with the connection, it just shows an empty graph!

## Open questions / future improvements
- Where should the `TeamStore` be initiated ideally? For now, I often rely on `window.location.pathname` to get the `teamSlug` instead (which works totally fine, but doesn't feel the cleanest and assumes the editor will _never_ be hosted at a custom subdomain!)
- I haven't touched global settings or settings pages yet
- How to consistently show user initial menu across teams, team and flow pages? It's inconsistent for me now on teams and team (eg sometimes shows if I'm navigating _back_ from a flow, but not on initial load)
- TESTS - there's a few that got skipped here because we'll need to figure out how to mock an authenticated user with platform admin permissions for buttons to be visible etc


jessicamcinchak:
[Implement user permissions in frontend](https://trello.com/c/6ZN4VyNb/2623-implement-user-permissions-in-frontend)

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -91,6 +92,8 @@ export const editorStore: StateCreator<
     { id = undefined, type, data },
     { children = undefined, parent = ROOT_NODE_KEY, before = undefined } = {},
   ) => {
+    if (get().getUser().isViewOnly) return;
```
jessicamcinchak, line 95: this is just a proof of concept that based on user details from the store, we can prohibit adding/moving/updating/etc nodes! note there isn't any UI feedback yet, just silent failures.

i learned our original hypothesis about not connecting to ShareDB altogether, rather than restricting per operation method, doesn't actually work very well here because the `connect` & `connectTo` store methods here are actually responsible for _loading_ the whole flow. So if we don't proceed with the connection, it just shows an empty graph.

editor.planx.uk/src/pages/FlowEditor/lib/store/user.ts
```diff
@@ -0,0 +1,43 @@
...
+  getUser: () => ({
+    name: get().name,
+    email: get().email,
+    roles: get().roles,
+    isPlatformAdmin: get().isPlatformAdmin,
+    isViewOnly:
+      !get().isPlatformAdmin &&
+      !get()
+        .roles.map((r) => r.role)
+        .includes("teamEditor"),
```
jessicamcinchak, line 41: i'm not actually sure we want `isViewOnly` on the actual User type, rather it's probably better as it's own method within the user store here, but again this is just quick first go !

editor.planx.uk/src/routes/authenticated.tsx
```diff
@@ -11,14 +11,26 @@ import { makeTitle } from "./utils";
...
+        name: "Jess",
+        email: "jessica@opensystemslab.io",
+        isPlatformAdmin: false,
+        roles: [
+          {
+            teamSlug: "lambeth",
+            role: "teamViewer",
+          },
+        ],
+      });
```
jessicamcinchak, line 32: this outstanding todo is related to #2224

editor.planx.uk/src/lib/lowcalStorage.ts
```diff
@@ -136,7 +136,7 @@ const getSessionContext = (sessionId: string): DefaultContext => ({
     "x-hasura-lowcal-email":
       // email may be absent for non save and return journeys
       useStore.getState().saveToEmail?.toLowerCase() || "",
-  }
+  },
```
jessicamcinchak, line 139: not sure why I picked up these linting changes, sorry for the extra noise !

editor.planx.uk/src/pages/FlowEditor/lib/store/user.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  canUserEditTeam: (teamSlug) => {
+    return (
+      get().roles.filter(
+        (r) =>
+          (r.role === "teamEditor" && r.teamSlug === teamSlug) ||
+          get().isPlatformAdmin,
+      ).length > 0
+    );
+  },
```
jessicamcinchak, line 47: this is the most important bit of logic to doublecheck please ! 

intended behavior here is: 
- return `true` if I'm a `platformAdmin` independent of `teamSlug`
- return `true` if I'm a `teamEditor` of the given `teamSlug`
- return `false` otherwise

editor.planx.uk/src/ui/NextStepsList.tsx
```diff
@@ -103,17 +106,21 @@ const Step = ({ title, description, url }: ListItemProps) => (
...
+      {props.steps?.map((step, i) => (
+        <StyledListItem key={i}>
+          {step.url ? (
+            <LinkStep {...step} />
+          ) : (
+            <ContinueStep {...step} handleSubmit={props.handleSubmit} />
+          )}
+        </StyledListItem>
+      ))}
+    </Root>
```
jessicamcinchak, line 123: again just noisy linter changes in this file, sorry!

editor.planx.uk/src/routes/authenticated.tsx
```diff
@@ -11,14 +14,45 @@ import { makeTitle } from "./utils";
...
+                  name
+                  slug
+                  id
+                }
+              }
+            }
+          }
+        `,
+        variables: { email },
+      });
```
jessicamcinchak, line 47: I'm using the Apollo client for authenticated users here out of convenience, but this query is also duplicated in planx-core's `UserClient` as `getByEmail` - likely one to simplify going forward!

editor.planx.uk/src/pages/FlowEditor/lib/store/user.ts
```diff
@@ -0,0 +1,57 @@
...
+    set({
+      id: user.id,
+      firstName: user.firstName,
+      lastName: user.lastName,
+      email: user.email,
+      isPlatformAdmin: user.isPlatformAdmin,
+      teams: user.teams,
+    }),
+
+  getUser: () => ({
```
DafyddLlyr, line 39: I'm not 100% here (worth a try?) but I think we could simplify `setUser()` and `getUser()` here as we don't have any other stores in our `StateCreator` to just - 

```ts
setUser: (user) => set(user),
getUser: () => get(),
```

editor.planx.uk/src/pages/FlowEditor/lib/store/user.ts
DafyddLlyr: The logic here is correct 👍

DafyddLlyr (approved):
A few comments on PR, no show stoppers! Looking great and feels very solid 👏 I'm sure Alastair and Ian will have some design thoughts but that's 100% a separate issue.

A few unit tests for maybe some components and UserStore methods would probably be wise as a next step 👍 

Hitting a few issues small on the Pizza mainly related to the teamStore not being populated. I think this is related to an issue we've discussed before around how teams are handled within the editor (`window.api.getState().getTeam()` is perfect in `/preview` and `/unpublished` but inconsistent in the Editor as you've flagged in this PR.

I think we'll need to resolve this before going for a prod deploy - I'll queue on up just now so we can get the code changes from already merged to `main` pushed forward in the meantime.

DafyddLlyr:
Just realised in the demo...
 - We need to disable drag drop in the modal
 - We need to hide "delete" buttons in modal
 
 I'm not exactly sure when the "update" happens - I think it's on save of the modal but juuuuust in case? And I guess it's about expected behaviour also.

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Question.tsx
```diff
@@ -53,9 +53,16 @@ const Question: React.FC<Props> = React.memo((props) => {
...
   const iconTitleAccess = props.type === "Error" ? "Error" : undefined;
 
+  // useStore.getState().getTeam().slug undefined here, use window instead
+  const teamSlug = window.location.pathname.split("/")[1];
+
   return (
     <>
-      <Hanger hidden={isDragging} before={props.id} parent={parent} />
+      <Hanger
+        hidden={isDragging || !useStore.getState().canUserEditTeam(teamSlug)}
```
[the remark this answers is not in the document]
jessicamcinchak: Yep, good call - I hadn't realised I would have to touch _every_ `Hanger` when I started hiding, but I definitely did - so this is now updated to set the logic once directly in the component

editor.planx.uk/src/pages/FlowEditor/lib/store/user.ts
jessicamcinchak: so this pattern `setUser()` & `getUser()` matches exactly how the `TeamStore` is currently setup! I'd vote perhaps one to revisit in a followup PR, as this PR points out a few limitations of how the `TeamStore` is initiated and where it's available too?

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -365,7 +365,6 @@ export const editorStore: StateCreator<
         toBefore,
       })(get().flow);
       send(ops);
-      get().resetPreview();
```
jessicamcinchak, line 368: so this is a notable change! `moveNode` is the only graph store method that called this shared store method, which causes the window to reload, the user to no longer be set, and therefore the flow incorrectly reloads as if it's now view-only rather than edit-access as intended.

from what I can tell, `resetPreview` is intended to be called outside the Editor based on the variables it's setting, and in testing I can't see any in-editor issues with removing it (the operation is still successful, the flow order after move persists as expected, etc). But please do test this!

editor.planx.uk/src/pages/FlowEditor/lib/store/user.ts
DafyddLlyr: This works for me!

PR #2216, opened 2023-09-10
chore: hookup Gloucester to submit to BOPS staging

jessicamcinchak:
Will only for BOPS Staging, _not_ BOPS Production yet similar to Camden. We shouldn't need to make any further changes on our end though once BOPS enables prod environments for these councils.


jessicamcinchak:
[Hookup Gloucester to BOPS staging for testing submissions](https://trello.com/c/ET8N5zcy/2631-hookup-gloucester-to-bops-staging-for-testing-submissions)

PR #2217, opened 2023-09-15
fix: Handle multiple external portals as siblings

DafyddLlyr:
## What does this PR do?
 - Fixes bug described here - https://trello.com/c/23Wi60zZ/2551-blank-screen-shows-when-passing-through-external-portal
 - Adds tests
 - Small restructure/renaming refactor of `dataMerged()`

### When was this bug introduced?
It appears to have been there since `dataMerged()` was written as far as I can tell which is a little surprising - I don't have total confidence in this but it's my best guess.

### Why not moved `dataMerged()` to a shared location like `planx-core`?
We could do this, but there's not real need to - if more flow management / logic goes there then sure, but for now we can continue to live with the duplication here I think.

### Test flows
 - Simplified example - https://2217.planx.pizza/testing/external-portal-fixed-test/preview?analytics=false
 - Compex example (real flow described in ticket) - https://2217.planx.pizza/buckinghamshire/apply-for-building-regulations-applications/preview?analytics=false. Note - this required a "publish" in order to be re-flattened according to the changes in `dataMerged()`.

### DIFF
--- a/editor.planx.uk/src/lib/dataMergedHotfix.ts
+++ b/editor.planx.uk/src/lib/dataMergedHotfix.ts
@@ -20,28 +20,44 @@ const getFlowData = async (id: string) => {
   return data.flows_by_pk;
 };
 
-export const dataMerged = async (
-  id: string,
-  ob: Record<string, any> = {},
-): Promise<Record<string, any>> => {
-  const { slug, data }: { slug: string; data: Record<string, any> } =
-    await getFlowData(id);
+// Flatten a flow's data to include main content & portals in a single JSON representation
+// XXX: getFlowData & dataMerged are currently repeated in api.planx.uk/helpers.ts
+//        in order to load frontend /preview routes for flows that are not published
+export const dataMerged = async (id: string, ob: Record<string, any> = {}) => {
+  // get the primary flow data
+  const { slug, data }: { slug: string; data: Record<string, any> } = await getFlowData(id);
+
+  // recursively get and flatten internal portals & external portals
   for (const [nodeId, node] of Object.entries(data)) {
-    if (nodeId === "_root" && Object.keys(ob).length > 0) {
+    const isExternalPortalRoot = nodeId === "_root" && Object.keys(ob).length > 0;
+    const isExternalPortal = node.type === TYPES.ExternalPortal;
+    const isMerged = ob[node.data?.flowId];
+
+    // Merge portal root as a new node in the graph
+    if (isExternalPortalRoot) {
       ob[id] = {
         ...node,
         type: TYPES.InternalPortal,
         data: { text: slug },
       };
-    } else if (node.type === TYPES.ExternalPortal && !ob[node.data.flowId]) {
-      await dataMerged(node.data.flowId, ob);
+    }
+
+    // Merge as internal portal, with reference to flowId
+    else if (isExternalPortal) {
       ob[nodeId] = {
         type: TYPES.InternalPortal,
-        edges: [node.data.flowId],
+        edges: [node.data?.flowId],
       };
-    } else {
-      ob[nodeId] = node;
+
+      // Recursively merge flow
+      if (!isMerged) {
+        await dataMerged(node.data?.flowId, ob);
+      }
     }
+
+    // Merge all other nodes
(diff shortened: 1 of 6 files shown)


editor.planx.uk/src/lib/dataMergedHotfix.ts
```diff
@@ -20,28 +20,44 @@ const getFlowData = async (id: string) => {
...
+    const isFlowAlreadyMerged = isExternalPortal && !ob[node.data?.flowId];
+
+    // Flatten portal root as a new node in the graph
+    if (isExternalPortalRoot) {
       ob[id] = {
         ...node,
         type: TYPES.InternalPortal,
         data: { text: slug },
       };
-    } else if (node.type === TYPES.ExternalPortal && !ob[node.data.flowId]) {
```
DafyddLlyr, line 36: There's a little rearranging / renaming here but breaking up this condition is what's fixed the bug. As is, this flattens the first external portal it hits, but skips subsequent ones.

DafyddLlyr:
[Blank screen shows when passing through external portal](https://trello.com/c/23Wi60zZ/2551-blank-screen-shows-when-passing-through-external-portal)

editor.planx.uk/src/lib/dataMergedHotfix.ts
```diff
@@ -20,28 +20,44 @@ const getFlowData = async (id: string) => {
...
+export const dataMerged = async (id: string, ob: Record<string, any> = {}) => {
+  // get the primary flow data
+  const { slug, data }: { slug: string; data: Record<string, any> } = await getFlowData(id);
+
+  // recursively get and flatten internal portals & external portals
   for (const [nodeId, node] of Object.entries(data)) {
-    if (nodeId === "_root" && Object.keys(ob).length > 0) {
+    const isExternalPortalRoot = nodeId === "_root" && Object.keys(ob).length > 0;
+    const isExternalPortal = node.type === TYPES.ExternalPortal;
+    const isMerged = ob[node.data?.flowId];
```
jessicamcinchak, line 34: nit: is `isNotMerged` a more clear name here since `node.data?.flowId` indicates it's still an external portal reference, _not_ flattened representation of individual nodes?

jessicamcinchak (approved):
Thanks for this fix & tests - pizza working as expected for me! 

Still a bit hung up on where we've introduced this (not totally convinced beginning of `dataMerged` because flows published a couple months ago are ok?) - one half-baked thought while sifting through changes is perhaps that the newish `FlowGraph` type from planx-core is being misused somewhere to type unpublished flows (which it actually can't handle because it doesn't account for `data.flowId` in a Node) when it should only be used to type published flows? Just noticing the type right now, and haven't confirmed if any actual nodes were getting dropped/overwritten on sort or order, but just a thought! It's such a strange one.

DafyddLlyr:
> One half-baked thought while sifting through changes is perhaps that the newish `FlowGraph` type from planx-core is being misused somewhere to type unpublished flows (which it actually can't handle because it doesn't account for `data.flowId` in a Node) when it should only be used to type published flows? Just noticing the type right now, and haven't confirmed if any actual nodes were getting dropped/overwritten on sort or order, but just a thought! It's such a strange one.

I actually think something along these lines is more likely but I've also not been able to prove anything / work it out 🤔

editor.planx.uk/src/lib/dataMergedHotfix.ts
DafyddLlyr: So here `ob` is the returned item (the flattened flow). If `ob[node.data?.flowId] == true` then the flow has already been flattened and `ob[example-uuid-of-portal]` exists. Does that make sense?

DafyddLlyr: TBH I think some of the language here is a little fuzzy around merging / flattening

DafyddLlyr:
Thanks for the review! As anything graph related is high impact I'll request PO testing on the Pizza.

DafyddLlyr:
Merged - PO testing done by Silvia and Sam https://trello.com/c/23Wi60zZ/2551-blank-screen-shows-when-passing-through-external-portal

PR #2218, opened 2023-09-15
feat: Allow `NextSteps` without URL to continue flow

DafyddLlyr:
## What does this PR do?
- Makes the URL field optional in the `NextSteps` component
- If not populated, the next steps button acts as a "Continue"

### Questions
 - Does this achieve what we want, or if branching here essential from within the component?
 - Do we want to style the buttons and links differently (as I've done here) or is it better than they look visually consistent (even if semantic HTML describes the difference between the options).

https://github.com/theopensystemslab/planx-new/assets/20502206/16a01f39-0303-4105-9d44-4e92c852bac2


DafyddLlyr:
[Improve Next steps component in guidance services "If no URL, act as a Continue Button"](https://trello.com/c/XttkNFtQ/2625-improve-next-steps-component-in-guidance-services-if-no-url-act-as-a-continue-button)

jessicamcinchak (approved):
Nice & simple! Week before last we decided in pre-planning that branching was non-essential here, and could always be accomplished with content / `SetValue`'s if necessary in particular cases

Current styling choices make sense to me :heavy_plus_sign:

PR #2220, opened 2023-09-18
test: Setup E2E tests for user roles

DafyddLlyr:
## What does this PR do?
- Setup test harness for user permissions, with one example test in each category
- More details in README!

### DIFF
--- a/e2e/tests/api-driven/src/globalHelpers.ts
+++ b/e2e/tests/api-driven/src/globalHelpers.ts
@@ -1,32 +1,44 @@
 import { TEST_EMAIL } from "../../ui-driven/src/helpers";
 import { $admin } from "./client";
 
-export function createTeam(args?: Partial<Parameters<typeof $admin.team.create>[0]>) {
-  return safely(() => $admin.team.create({
-    name: "E2E Test Team",
-    slug: "E2E",
-    logo: "https://raw.githubusercontent.com/theopensystemslab/planx-team-logos/main/planx-testing.svg",
-    primaryColor: "#444444",
-    submissionEmail: TEST_EMAIL,
-    homepage: "planx.uk",
-    ...args,
-  }));
+export function createTeam(
+  args?: Partial<Parameters<typeof $admin.team.create>[0]>,
+) {
+  return safely(() =>
+    $admin.team.create({
+      name: "E2E Test Team",
+      slug: "E2E",
+      logo: "https://raw.githubusercontent.com/theopensystemslab/planx-team-logos/main/planx-testing.svg",
+      primaryColor: "#444444",
+      submissionEmail: TEST_EMAIL,
+      homepage: "planx.uk",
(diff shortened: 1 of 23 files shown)


e2e/tests/api-driven/src/globalHelpers.ts
```diff
@@ -0,0 +1,57 @@
...
+      data: { dummy: "flowData " },
+      ...args,
+    }),
+  );
+}
+
+/**
+ * Error handling boilerplate for client functions
+ */
+export function safely<T extends () => ReturnType<T>>(callback: T) {
```
DafyddLlyr, line 47: One thing this highlighted was that the `planx-core` client doesn't always return consistent values for success/failure. Sometimes it's a `null`, or `false`, or an error is thrown.

Taking a bit of inspiration from `useSWR` it would be super nice if it just always returned a consistent result object of some sort which we could check for errors.

e2e/tests/api-driven/src/jwt.ts
```diff
@@ -0,0 +1,51 @@
+import { sign } from "jsonwebtoken";
+import { User, Role } from "@opensystemslab/planx-core/types";
+import { $admin } from "./client";
+
+// This code is copied from api.planx.uk/modules/auth/service.ts
```
DafyddLlyr, line 5: Not an ideal solution! Another candidate for functionality that could move to `planx-core` possibly?

e2e/tests/api-driven/src/permissions/queries/index.ts
```diff
@@ -0,0 +1,7 @@
+import { INSERT_FLOW_QUERY } from "./flows";
+
+export const queries = {
```
DafyddLlyr, line 3: This PR is just setup. I envision the next PR will be just a bunch of queries (and maybe a few additional setup/cleanup steps)

e2e/tests/api-driven/src/permissions/teamAdmin.feature
```diff
@@ -0,0 +1,21 @@
+Feature: Testing Permissions for teamAdmin Role
+
+  @regression @team-admin-permissions
+  Scenario Outline: teamAdmin permissions
+    Given a teamAdmin is a member of a team
+    When they perform "<ACTION>" on "<TABLE>"
+    Then they have access
+
+    Examples:
+      | TABLE   | ACTION          |
```
DafyddLlyr, line 10: I did consider adding a `DATA` and `COLUMNS` column here and templating out standard CRUD GraphQL queries. 

At least initially hardcoding queries is simpler (if slightly more tedious).

jessicamcinchak (approved):
Looks good & seems like a very solid start - thanks for these initial tooling decisions, setup, and documentation ! 

I'm sure I'll have more thoughts once I start writing tests myself, but don't see any blockers to this going in now :+1:

PR #2224, opened 2023-09-19
feat: Enrich `/me` endpoint with user roles

DafyddLlyr:
## What does this PR do?
 - Adds user roles data to `/me`, makes this accessible to all logged in users
 - Moves `/me` and `/` (healthcheck) to a "misc" module in the API


api.planx.uk/modules/misc/docs.yaml
```diff
@@ -0,0 +1,72 @@
...
+                    example: albert@princeton.edu
+                  isPlatformAdmin:
+                    type: boolean
+                    example: true
+                  teams:
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        teamId:
```
jessicamcinchak, line 63: nit: it'd be super handy to have the `teamSlug` exposed here in addition to `teamId` - as I think we can then call this endpoint from the editor for signed-in users to populate their user menu in a meaningful way?

jessicamcinchak (approved):

api.planx.uk/modules/misc/docs.yaml
DafyddLlyr: On it!

DafyddLlyr:
Blocked by https://github.com/theopensystemslab/planx-new/pull/2226

PR #2225, opened 2023-09-20
feat: Improve upload + label select usability

ianjon3s:
### Summary

- Adds a 'done' button to the header of the select dropdown list.
- Header is sticky so will be visible at all times for longer lists that overflow the viewport

PR satisfies: https://trello.com/c/XCE7KiSR/2645-add-okay-or-done-button-to-labels-component-for-upload-drawings-section

Demo:
https://2225.planx.pizza/planx/test-upload-label-select/preview
There are two upload and label components in the demo flow, the first with a short file list and the second with a longer list (to demonstrate the effectiveness of the list header). You will need to upload and label a file to reach the second component.

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Modal.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Modal.tsx
@@ -39,8 +39,7 @@ interface FileTaggingModalProps {
   setShowModal: (value: React.SetStateAction<boolean>) => void;
 }
 
-
-const ListHeader = styled('li')(({ theme }) => ({
+const ListHeader = styled(Box)(({ theme }) => ({
   display: "flex",
   justifyContent: "space-between",
   alignItems: "center",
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/@planx/components/FileUploadAndLabel/Modal.tsx
```diff
@@ -267,14 +280,27 @@ const SelectMultiple = (props: SelectMultipleProps) => {
           },
         }}
       >
+        <ListSubheader disableGutters>
+          <ListHeader>
+            <Typography variant="h4" component="h3" pr={3}>
+              Select all document types that apply
```
jessicamcinchak, line 286: nit: "document types" feels like brand new language only used here - would "Select all labels that apply" better match the error messages that might be shown later on validation? If this copy came from content team/user research, obviously feel free to ignore my suggestion :slightly_smiling_face:

jessicamcinchak (approved):
Works for me !

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Modal.tsx
ianjon3s: Thanks @jessicamcinchak, I will ask August about this one.

ianjon3s: I've updated to this (Select all labels that apply) for now, will add as a separate PR if there is further feedback from content design.

PR #2228, opened 2023-09-20
fix: Incomplete multi-character sanitation

DafyddLlyr:
Please see https://github.com/theopensystemslab/planx-new/security/code-scanning/7

The same PR will shortly follow on `planx-core`

### DIFF
--- a/editor.planx.uk/package.json
+++ b/editor.planx.uk/package.json
@@ -83,6 +83,7 @@
     "rxjs": "^7.8.1",
     "scroll-into-view-if-needed": "^2.2.31",
     "sharedb": "^3.3.1",
+    "striptags": "^3.2.0",
     "swr": "^2.2.0",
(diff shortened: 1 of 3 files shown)


jessicamcinchak (approved):
This one is a good reminder that we should be able to strip out a lot of this editor "Send" logic now I think that planx-core is capturing both BOPS & CSV generation (but maybe still an outstanding question of public versus admin client if I remember right? let's briefly discuss at dev call later?)

DafyddLlyr:
> This one is a good reminder that we should be able to strip out a lot of this editor "Send" logic now I think that planx-core is capturing both BOPS & CSV generation (but maybe still an outstanding question of public versus admin client if I remember right? let's briefly discuss at dev call later?)

Yep - 100%....! I had this same question also - I think we can update this and remove it now.

PR #2229, opened 2023-09-20
fix: ensure unauthorised public users can select from `blpu_codes` table to populate PropertyInformation

jessicamcinchak:
This one came up in #help-issues: when testing Doncaster FOI _not signed into the editor_, Jenna was unexpectedly being asked property type questions manually. When testing signed in, the description from the BLPU codes lookup was correctly populating automatically. 

The public role can select from `blpu_codes`, so this ensures `useQuery()` now uses the `publicClient` rather than default authorised client.

To test:
- Search 45 Thealby Gardens, DN4 7EQ [here](https://2229.planx.pizza/doncaster/find-out-if-you-need-planning-permission/preview) - confirm "Property type" populates with "Detached" automatically

### DIFF
--- a/api.planx.uk/admin/feedback/downloadFeedbackCSV.ts
+++ b/api.planx.uk/admin/feedback/downloadFeedbackCSV.ts
@@ -62,7 +62,7 @@ type ParsedFeedback = Feedback & {
  *    tags:
  *      - admin
  *    security:
- *      - userJWT: []
+ *      - bearerAuth: []
  */
 export const downloadFeedbackCSV = async (
   req: Request,
--- a/api.planx.uk/admin/session/bops.ts
+++ b/api.planx.uk/admin/session/bops.ts
@@ -1,5 +1,5 @@
 import { Request, Response, NextFunction } from "express";
-import { $admin } from "../../client";
+import { getClient } from "../../client";
 
 /**
  * @swagger
@@ -12,7 +12,7 @@ import { $admin } from "../../client";
  *    parameters:
  *      - $ref: '#/components/parameters/sessionId'
  *    security:
- *      - userJWT: []
+ *      - bearerAuth: []
  */
 export const getBOPSPayload = async (
   req: Request,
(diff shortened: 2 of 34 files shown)


jessicamcinchak:
[Public users need to be able to query `blpu_codes` table](https://trello.com/c/BcIdUFqc/2644-public-users-need-to-be-able-to-query-blpucodes-table)

jessicamcinchak:
Switching back to draft - this is still failing with console error 
```
[GraphQL error]: Message: Could not verify JWT: JWSError (CompactDecodeError Invalid number of parts: Expected 3 parts; got 1), Location: undefined, Path: undefined
nn @ console.ts:15
```

DafyddLlyr (changes requested):
Thanks for picking this up so quickly and troubleshooting with Jenna and August 🏅 

This will certainly work as a solution but I think it's a bit of a misdiagnosis of the underlying issue here.

In order to keep the permission set of a role as small as possible we're aiming to not have to duplicate public permissions into each role. One of the biggest wins here would be that we wouldn't have to update every role each time the permissions set for "public" is updated. There's a little bit more context on this here - https://github.com/theopensystemslab/planx-new/pull/2182#discussion_r1312254263

What should be happening - 
 - This is a public query to fetch BLPU codes, so it should be using the `publicClient` (which sets the `x-hasura-user-role = "public"` header)
 
 Why this was missed - 
  - In #2195 I updated all calls using `client.request()` but didn't think about the `useQuery()` hook 🤦 
   - `useQuery()` uses the default client used by the `<ApolloClient>` wrapper (set to the private scope)
   
How we could fix this - 
 - Decide if `ApolloClient` should use private or public clients by default (public would be safer if slightly more annoying to deal with?)
  - Update all `useQuery()` calls to have the correct client, e.g. `useQuery(MY_QUERY, { client: privateClient });`
  
Frustratingly it looks like we can only use one `ApolloProvider` per app - so we can't conveniently wrap public / private layouts with the correct provider.

DafyddLlyr (approved):

PR #2230, opened 2023-09-21
feat(e2e): `teamAdmin` permission tests

DafyddLlyr:
- Add test cases for the teamAdmin role for all tables not covered by introspection tests
- Fix permissions on operations table (ShareDB does the writing to this as admin currently, but I thought it worthwhile to correctly model Hasura)
- Resolves CVE-2023-26144 for e2e/api tests, API & Editor by bumping `planx-core`
  - e2e/ui to follow - aiming to use `planx-core` here for teardown instead of the more manual `graphql-request` approach

✅  Regression test run against this branch - https://github.com/theopensystemslab/planx-new/actions/runs/6338450886


e2e/tests/api-driven/src/permissions/steps.ts
```diff
@@ -1,72 +1,101 @@
...
+  user2Id!: number;
+  user2Email!: string;
+  team2FlowId!: string;
+
+  activeUserId!: number;
+  activeUserEmail!: string;
 
   error?: Error = undefined;
-  activeUser!: TestUser;
+  result: any[] | Record<"returning", any[]> | null = null;
```
jessicamcinchak, line 28: Personally, would love to turn off the type warnings for `unexpected any` to make PRs a bit less noisy - I think we can trust each other to catch actually-unnecessary ones in code reviews ?

e2e/tests/api-driven/src/permissions/helpers.ts
```diff
@@ -24,26 +28,27 @@ export const cleanup = async () => {
...
 export const setup = async () => {
+  const user1Id = await createUser({ email: "e2e-user-1@opensystemslab.io" });
   const teamId1 = await createTeam({ name: "E2E Team 1", slug: "e2e-team1" });
+  const team1FlowId = await createFlow({
+    teamId: teamId1,
+    slug: "team-1-flow",
+  });
+
+  const user2Id = await createUser({ email: "e2e-user-2@opensystemslab.io" });
   const teamId2 = await createTeam({ name: "E2E Team 2", slug: "e2e-team2" });
```
jessicamcinchak, line 39: double checking i understand this: in overall setup, we create a user who is by default _not_ a platform admin and they also do not have roles on either `teamId1` or `teamId2`? then later in individual test suites, you're creating & cleaning up team memberships and toggling permissions?

i'm wondering a bit if it'd be clearer to setup more explicitly with something like `e2e-platform-admin@..` etc rather than vague `user-1` etc? definitely not a showstopper though.

e2e/tests/api-driven/src/permissions/helpers.ts
```diff
@@ -24,26 +28,27 @@ export const cleanup = async () => {
 };
 
 export const setup = async () => {
+  const user1Id = await createUser({ email: "e2e-user-1@opensystemslab.io" });
```
jessicamcinchak, line 31: nit: I think helen & mel previously reported an annoying issue where e2e test emails using the @opensystemslab.io domain were causing bounced errors in the enquiries inbox? can we use a harmless @example.com or something instead?

jessicamcinchak (approved):
Looks good to me, thanks for figuring this one out! 

Left a few minor questions, but no blockers.

e2e/tests/api-driven/src/permissions/helpers.ts
DafyddLlyr: Yeah I appreciate this is an unclear point - team1 and team2 are quite vague/meaningless names but I needed something to differentiate (I was also considering red/blue or something to make it slightly more distinct).

The intention is that there are -
 - Two teams
 - One TeamEditor of each team
 - We test that teamEditor1 can access team1 resources
 - We test that teamEditor2 cannot access team1 resources

I'll add comments to explicitly explain the setup here 👍

PR #2234, opened 2023-09-22
docs: Include details on WARP for API endpoints

DafyddLlyr:
image

### DIFF
--- a/api.planx.uk/modules/team/docs.yaml
+++ b/api.planx.uk/modules/team/docs.yaml
@@ -26,6 +26,10 @@ paths:
   /team/{teamId}/add-member:
     put:
       summary: Add user to team, and assign them a role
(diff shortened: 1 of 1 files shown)


api.planx.uk/modules/team/docs.yaml
```diff
@@ -26,6 +26,10 @@ paths:
   /team/{teamId}/add-member:
     put:
       summary: Add user to team, and assign them a role
+      description: 
+        "Requires authentication via the WARP platform
```
jessicamcinchak, line 30: nit: "via a Cloudflare WARP client" might be more accurate language and better match how we talk about this in IT policy etc?

jessicamcinchak (approved):
Thanks for adding this !

PR #2235, opened 2023-09-22
Production deploy

DafyddLlyr:


jessicamcinchak (approved):

jessicamcinchak:
Should we also consider getting https://github.com/theopensystemslab/planx-new/pull/2217 in here? Feels like a high-value one!

DafyddLlyr:
@jessicamcinchak Just merged! Silvia and Sam have both tested on Trello now ✅

PR #2236, opened 2023-09-26
feat: Create user endpoint, manage team members by email and teamSlugs

DafyddLlyr:
## What does this PR do?
- Adds a new create user endpoint to be used by Platform Admins
- Updates team endpoints to use user emails and team slugs
- Updates tests, standardises error handling in new modules
- Fixes a few missing permissions for the `platformAdmin` role

Depends on https://github.com/theopensystemslab/planx-core/pull/137

### DIFF
--- a/api.planx.uk/modules/team/service.test.ts
+++ b/api.planx.uk/modules/team/service.test.ts
@@ -1,0 +1,108 @@
+import { getJWT } from "../../tests/mockJWT";
+import { userContext } from "../auth/middleware";
+
+import {
+  getUserAndTeam,
+  addMember,
+  removeMember,
+  changeMemberRole,
+} from "./service";
+
+const getStoreMock = jest.spyOn(userContext, "getStore");
+
+const mockTeam = { id: 123 };
+const mockUser = { id: 456 };
+
+const mockAddMember = jest.fn();
+const mockRemoveMember = jest.fn();
+const mockChangeMemberRole = jest.fn();
+const mockGetTeamBySlug = jest.fn().mockResolvedValue(mockTeam);
+const mockGetUserByEmail = jest.fn().mockResolvedValue(mockUser);
+
+jest.mock("@opensystemslab/planx-core", () => {
+  return {
+    CoreDomainClient: jest.fn().mockImplementation(() => ({
+      team: {
+        getBySlug: () => mockGetTeamBySlug(),
+        addMember: () => mockAddMember(),
+        removeMember: () => mockRemoveMember(),
+        changeMemberRole: () => mockChangeMemberRole(),
+      },
+      user: {
+        getByEmail: () => mockGetUserByEmail(),
+      },
+    })),
+  };
+});
+
+describe("TeamService", () => {
+  beforeEach(() => {
+    getStoreMock.mockReturnValue({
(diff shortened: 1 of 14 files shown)


api.planx.uk/modules/team/service.test.ts
```diff
@@ -0,0 +1,108 @@
...
+const mockTeam = { id: 123 };
+const mockUser = { id: 456 };
+
+const mockAddMember = jest.fn();
+const mockRemoveMember = jest.fn();
+const mockChangeMemberRole = jest.fn();
+const mockGetTeamBySlug = jest.fn().mockResolvedValue(mockTeam);
+const mockGetUserByEmail = jest.fn().mockResolvedValue(mockUser);
+
+jest.mock("@opensystemslab/planx-core", () => {
```
DafyddLlyr, line 22: At some point soon I'm going to look at how Jest can automock modules - I think this will allow us much easier setup and less repetition.

hasura.planx.uk/metadata/tables.yaml
DafyddLlyr, line 1: Changes in this file were required as queries were failing which should have been allowed - apologies for the nosie.

api.planx.uk/modules/team/service.ts
```diff
@@ -0,0 +1,74 @@
...
+  if (!isSuccess) throw Error(`Failed to assign ${userEmail} to ${teamSlug}`);
+};
+
+export const changeMemberRole = async ({
+  userEmail,
+  teamSlug,
+  role,
+}: UpsertMember) => {
+  const $client = getClient();
+  const { user, team } = await getUserAndTeam({ userEmail, teamSlug });
```
DafyddLlyr, line 37: Doing multiple round trips to the DB to get these details feels pretty awkward. If / when we expose this functionality into the UI it would be worth having `userId` and `teamId` upfront to make these requests.

DafyddLlyr:
Moved (briefly!) back to draft - there's a few outstanding docs tidy ups needed I've just noticed on the Pizza sorry!

api.planx.uk/modules/user/controller.ts
```diff
@@ -0,0 +1,34 @@
...
+  try {
+    const $client = getClient();
+    await $client.user.create(req.body);
+    return res.send({ message: "Successfully created user" });
+  } catch (error) {
+    return next(
+      new ServerError({ message: "Failed to create user", cause: error }),
+    );
+  }
+};
```
jessicamcinchak, line 34: would we benefit from a "RemoveUser" here too that, rather than delete the actual record, nullifies the email field to lock them out? Probably not crucial yet, but it feels like the tiny missing piece from an otherwise very useful set of user & team management endpoints ! And it would probably encourage better house-keeping around this.

jessicamcinchak (approved):
Used the Swagger docs to test and truly a joy - all worked as expected! 

Can very much imagine a near-future workflow where others interact with this too :tada:

api.planx.uk/modules/user/controller.ts
DafyddLlyr: Good shout - I'll pick this up in a quick follow on PR.

PR #2239, opened 2023-09-27
feat: all users can edit "Templates" team and `moveFlows` to teams they're allowed to edit

jessicamcinchak:
- I added a "Templates" team to production & staging, this change ensures any user will have `teamEditor` access to "Templates" by default when they're added independent of records in the `team_members` table (if this feels too liberal, very happy to re-think how we should do this!)
- `moveFlow` Editor store method now checks if the given user has permission to move the flow into the requested team. If they don't, it'll display and error and return early. The copy/move menu is only visible in teams you can edit already, so I don't think we need to do anything similar with other store methods besides `moveFlow` yet.

Screenshot from 2023-09-27 09-46-52

**Testing:** 
- (If you're a platformAdmin, toggle this off first in https://hasura.2239.planx.pizza/console/login)
- I see an "edit" icon next to `Templates` even though I do not have a `team_members` entry
- Inside "Templates", I can create a flow, copy it, and move it - I can successfully move to a team that I have edit access to, but see an error message when I try to move it to a team I can only view

### DIFF
--- a/editor.planx.uk/src/components/Header.tsx
+++ b/editor.planx.uk/src/components/Header.tsx
@@ -480,7 +480,7 @@ const EditorToolbar: React.FC<{
                 <Edit />
               </ListItemIcon>
               <ListItemText>
-                {user.isPlatformAdmin ? `All teams` : user.teams.map((team) => team.team.name).join(", ")}
+                {user.isPlatformAdmin ? `All teams` : user.teams.map((team) => team.team.name).concat(["Templates"]).join(", ")}
               </ListItemText>
             </MenuItem>
           )}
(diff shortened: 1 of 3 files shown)


DafyddLlyr:
Ah actually I might be wrong here sorry...!

I thought it was a copy + move we needed, but I think this would work for just a copy. I'm going to get a coffee... ☕ 😅

DafyddLlyr (approved):

jessicamcinchak:
**Next steps:**
- This currently works because the API routes for moving & copying aren't scoped to user permissions yet
- We'll need to update this approach once that is implemented, but we're happy to merge this in the meantime because the UI will remain consistent

PR #2241, opened 2023-09-28
fix: Improve FileUpload text

DafyddLlyr:


DafyddLlyr:
[Users often aren't aware that a folder can be dragged into upload and label](https://trello.com/c/nDdiL43B/2630-users-often-arent-aware-that-a-folder-can-be-dragged-into-upload-and-label)

jessicamcinchak (approved):

PR #2242, opened 2023-09-28
fix: Init user store across all routes

DafyddLlyr:
- Init user store via router context
 - This means that we can refresh at any point / any page and the store will re-init


jessicamcinchak (approved):
Much tidier & works for me!

PR #2243, opened 2023-09-28
fix: Persist `TeamStore`

DafyddLlyr:
## What does this PR do?
 - Resolves an issue described by August [here on OSL Slack](https://opensystemslab.slack.com/archives/C5Q59R3HB/p1695652749495649?thread_ts=1695633553.425559&cid=C5Q59R3HB)
 - Ensures that the team store is not just loaded and populated on initial run through, but works on page refresh
 - Tidies up the `Header` component to use the teamStore as opposed to the route data

## How does this work?
 - The logic inside a `mount()` function on a route only gets called for a specific route
 - Using `withView()` matches all child routes, e.g. `/team/settings`

### DIFF
--- a/editor.planx.uk/src/components/Header.test.tsx
+++ b/editor.planx.uk/src/components/Header.test.tsx
@@ -27,14 +27,9 @@ const mockTeam2: Team = {
   slug: "closedsystemslab",
 };
 
-jest.spyOn(ReactNavi, "useNavigation").mockImplementation(
-  () =>
-    ({
-      data: {
-        team: mockTeam1,
-      },
-    }) as any,
-);
+jest.spyOn(ReactNavi, "useNavigation").mockReturnValue(({
+  navigate: jest.fn()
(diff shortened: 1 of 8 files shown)


jessicamcinchak (commented):
Getting a bit of strange behavior on this pizza - /preview links are redirecting to a login page? Can you recreate? Happening both when I navigate to a preview link directly and when using the icon in the editor flow sidepanel.

Eg https://2243.planx.pizza/barking-and-dagenham/find-out-if-you-need-planning-permission

jessicamcinchak (approved):

jessicamcinchak:
Re-tested latest changes, all working as expected :+1:

DafyddLlyr:
Thanks very much for this! On my agenda shortly is a very small / quick e2e test (or maybe a modification of an existing one) that just adds a refresh once the page has loaded to cover this behaviour with a test.

PR #2246, opened 2023-09-28
fix: Use view to init user store

DafyddLlyr:
Follow up / fix for https://github.com/theopensystemslab/planx-new/pull/2242

Currently on staging, non-authenticated routes will fail for authenticated users. An E2E test does cover this scenario, but is currently skipped (I hit this issue whilst fixing it - this PR will be incoming very shortly).

This should resolve this by following the patterns used elsewhere (recently in #2243) of getting a view to handle this for us.


DafyddLlyr:
Thanks for testing and reviewing so quickly! 🙌

PR #2247, opened 2023-09-29
fix(e2e): Update E2E tests for recent permissions changes

DafyddLlyr:
## What does this PR do?
- Unskips and fixes E2E tests
- Resolves a few user store issues hit along the way

## What was the issue?
`email` was added to the JWT in the API, but not in the e2e tests - commit [56e91c9](https://github.com/theopensystemslab/planx-new/pull/2247/commits/56e91c9bb3dd6993ad2fe84c92550c898939e223) has a quick fix for this.

`email` was only added to the token to make things less noisy in another PR - I've just made the update to rely on the ID by using the `user.getById()` method introduced here - https://github.com/theopensystemslab/planx-core/pull/143

Apologies for the commit with linting issues - I'll take a look at Husky today and try to work out what the issues is here.


editor.planx.uk/.eslintrc
```diff
@@ -73,5 +72,11 @@
   },
   "settings": {
     "jest": { "version": 27 }
-  }
+  },
+  "overrides": [
```
DafyddLlyr, line 76: This config update means that these linting rules will only apply to test files - I was getting a linting error from `testing-library/react` for using the `client.user.getById()` which shares a name with a `testing-library` function.

editor.planx.uk/src/pages/FlowEditor/lib/store/user.ts
```diff
@@ -1,20 +1,14 @@
...
 import type { StateCreator } from "zustand";
 
 export interface UserStore {
-  id: number;
-  firstName: string;
-  lastName: string;
-  email: string;
-  isPlatformAdmin: boolean;
-  teams: UserTeams[];
+  user?: User
```
DafyddLlyr, line 8: I consolidated these into a single user because of a side effect of merging our Zustand stores I was hitting up against.

FlowId is also called `id` in the stores and then overwrites userId which we wouldn't expect.
image

editor.planx.uk/src/pages/FlowEditor/lib/store/user.ts
```diff
@@ -23,69 +17,30 @@ export const userStore: StateCreator<UserStore, [], [], UserStore> = (
...
-                slug
-                id
-              }
-            }
-          }
-        }
-      `,
-      variables: { email },
-    });
+    if (this.getUser()) return;
```
DafyddLlyr, line 39: Checking this means we now don't run this query on each route change, just on initial page load.

api.planx.uk/modules/misc/routes.test.ts
```diff
@@ -70,21 +68,21 @@ describe("/me endpoint", () => {
...
       });
   });
 
   it("returns an error for an invalid email address", async () => {
-    mockGetByEmail.mockResolvedValueOnce(null);
+    mockGetById.mockResolvedValueOnce(null);
 
     await supertest(app)
       .get("/me")
       .set(authHeader({ role: "teamEditor" }))
```
jessicamcinchak, line 81: :+1:

jessicamcinchak, line 81: :+1:

jessicamcinchak (approved):

PR #2249, opened 2023-09-29
feat(api): Remove user endpoint

DafyddLlyr:
image

- Allow `platformAdmins` to update user emails (for soft delete)
- Add delete user endpoint to API docs


jessicamcinchak (approved):
Thanks for this one !

PR #2251, opened 2023-10-01
feat: Add payment exemption status to Slack notifications

DafyddLlyr:
## What does this PR do?
- Refactors `webhooks` directory into modular structure
- Untouched files moved to an `_old` directory temporarily - I'll pick these up very shortly
- Appends "[Exempt]" to Slack messages where `application.fee.payable === 0`


api.planx.uk/modules/webhooks/_old/lowcalSessionEvents.test.ts
```diff
@@ -1,10 +1,10 @@
 import supertest from "supertest";
-import app from "../server";
-import { createScheduledEvent } from "../hasura/metadata";
+import app from "../../../server";
```
DafyddLlyr, line 2: If the new API structure leads to a lot of long/messy relative import paths we could look at resolving this with path aliases -  https://blog.logrocket.com/using-path-aliases-cleaner-react-typescript-imports/

api.planx.uk/modules/webhooks/docs.yaml
DafyddLlyr, line 1: ChatGPT generated this for me based on the Zod schema 🤖 It just needed a small tidy up after.

api.planx.uk/modules/webhooks/sendNotification/service.ts
```diff
@@ -0,0 +1,59 @@
...
+export const sendSlackNotification = async (
+  data: EventData,
+  type: EventType,
+) => {
+  const slack = SlackNotify(process.env.SLACK_WEBHOOK_URL!);
+  let message = getMessageForEventType(data, type);
+
+  const sessionId = getSessionIdFromEvent(data, type);
+  const feePayable = await getFeePayableForSession(sessionId);
+  if (!feePayable) message += " [Exempt]";
```
DafyddLlyr, line 21: This PR ended up being more refactor than feature sorry, but this is the core of the new work.

api.planx.uk/modules/webhooks/sendNotification/index.test.ts
```diff
@@ -0,0 +1,321 @@
...
+        .send(body)
+        .expect(200)
+        .then((response) => {
+          expect(SlackNotify).toHaveBeenCalledWith(
+            process.env.SLACK_WEBHOOK_URL,
+          );
+          expect(mockSend).toHaveBeenCalledTimes(1);
+          expect(response.body.message).toBe("Posted to Slack");
+          expect(response.body.data).toMatch(/abc123/);
+          expect(response.body.data).toMatch(/www.bops-production.com/);
```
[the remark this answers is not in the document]
DafyddLlyr: Alert dismissed, this is just a test case.

api.planx.uk/modules/webhooks/sendNotification/service.ts
jessicamcinchak: I'm a little hesitant about this `(!feePayable)` logic because our most recent bug around this was actually content incorrectly setting fee payable to 0 regardless of exempt/resubmission answers. I _think_ this change would have actually incorrectly labeled those cases `[Exempt]` and we perhaps wouldn't have caught that as quick? 

My hunch is something like this will be more reliable / robust:
- Regardless of `application.fee.payable` & `application.fee.calculated`...
- If `application.fee.exemption.disability` is `true` then `message += " [Exempt]"`
- If `application.fee.exemption.resubmission` is `true` then `message += " [Resubmission]"`
- ** Exemptions are asked about via a Question node so you'll want to do `Boolean({field}[0])` to evaluate

So, the only cases going forward where we'd manually look into a case are: 
- Missing payment notification && submission notification does _not_ have `[Exempt]` or `[Resubmission]` suffix

Does this make sense?

DafyddLlyr: This is a really helpful comment thanks 👍 I'll go ahead and make this change.

DafyddLlyr: PR updated with these changes @jessicamcinchak 👌

jessicamcinchak (approved):

PR #2252, opened 2023-10-01
refactor(api): Update remaining webhook routes to modular structure

DafyddLlyr:
# What does this PR do?
 - Refactors remaining `/webhook` routes [to format proposed here](https://github.com/theopensystemslab/planx-new/blob/main/doc/architecture/decisions/0007-modularise-express-application.md)
 - Adds documentation
 - Updates types and tests
 - No changes to business logic, just a refactor - the only exception is to update the return type of one API endpoint (see comment)

image


api.planx.uk/hasura/metadata/index.ts
```diff
@@ -25,13 +25,18 @@ type RequiredScheduledEventArgs = Pick<
   "webhook" | "schedule_at" | "comment" | "payload"
 >;
 
+export interface ScheduledEventResponse {
```
DafyddLlyr, line 28: Docs: https://hasura.io/docs/latest/api-reference/metadata-api/scheduled-triggers/#metadata-create-scheduled-event

DafyddLlyr: Typing this cascaded across a few functions and tests which is a bit noisy for a PR, but quite nice in practice.

api.planx.uk/inviteToPay/createPaymentSendEvents.ts
```diff
@@ -14,9 +17,9 @@ enum Destination {
 }
 
 interface CombinedResponse {
```
DafyddLlyr, line 19: This type is repeated in `api.planx.uk/send/createSendEvents.ts` but I'm ignoring this for now as this PR is already bigger than it should be.

api.planx.uk/modules/webhooks/service/lowcalSessionEvents/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  createdAt,
+  payload,
+}: CreateSessionEvent) => {
+  const response = await createScheduledEvent({
+    webhook: "{{HASURA_PLANX_API_URL}}/send-email/expiry",
+    schedule_at: addDays(createdAt, DAYS_UNTIL_EXPIRY),
+    payload: payload,
+    comment: `expiry_${payload.sessionId}`,
+  });
+  return [response];
```
DafyddLlyr, line 43: Returning `[response]` as opposed to just `response` gives us a consistent return type across all these APIs.

jessicamcinchak (approved):

PR #2271, opened 2023-10-02
fix: Reset team store on navigation

DafyddLlyr:


jessicamcinchak (approved):
Pizza is working for me! Can't recreate the out of sync route & team store :heavy_check_mark: 

Thanks for quick fix!

PR #2274, opened 2023-10-03
chore: Bump tiptap

DafyddLlyr:
Flagged by dependabot earlier this week - https://github.com/theopensystemslab/planx-new/pull/2261

Makes sense to keep these all in sync with each other where possible 👍 

Linting issues grouped into a single commit - will prioritise looking at Husky as this is getting a bit tedious to fix.... 🫠


DafyddLlyr:
Tests currently failing with bug described here - https://github.com/ueberdosis/tiptap/issues/4455

The proposed fix in the comments works, but as this update is of low importance we can hold out for the package to get fixed.

PR #2282, opened 2023-10-04
Track which url is navigated to from Next Steps component

Mike-Heneghan:
Draft PR for adding adding analytics for which url was navigated to from the Next Steps component.


editor.planx.uk/src/ui/NextStepsList.tsx
```diff
@@ -73,7 +76,16 @@ const ArrowButton = styled("span")(({ theme }) => ({
...
+  <InnerLink
+    href={props.url}
+    target="_blank"
+    rel="noopener"
+    onClick={() =>
+      props.handleNextStepsLinkClick &&
+      props?.url &&
+      props.handleNextStepsLinkClick({ url: props.url })
+    }
+  >
```
Mike-Heneghan, line 88: I'm not loving what's going on here although I was getting warnings until I added this. I think it suggests I've made a type error somewhere else 🤔

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -171,6 +176,30 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     }
   }
 
+  async function trackNextStepsLinkClick(metadata?: NextStepsLinkMetadata) {
+    if (shouldTrackAnalytics && lastAnalyticsLogId) {
+      await publicClient.mutate({
+        mutation: gql`
+          mutation UpdateHasClickNextStepsLink(
```
Mike-Heneghan, line 183: Can I just define mutations like this? In my previous experience this would have to match a mutation via Graphene although is there Hasura magic going on? It seems to work which honestly surpised me 😅

DafyddLlyr: Yep! This is allowed and should work just fine. 

We don't currently enforce an allow-list of queries/mutations though it is something Hasura supports (and recommends in production). Docs: https://hasura.io/docs/latest/security/allow-list/

editor.planx.uk/src/ui/NextStepsList.tsx
DafyddLlyr: Good question. This is happening because the `url` property is optional in  `ListItemProps` despite actually being required in `LinkStep`.

There's a few ways to expressing this in TypeScript - I'd probably recommend trying to reimplement `ListItemProps` as a discriminated union of two types - one where `url` and `handleNextStepsLinkClick` are not there, and one where they're required. 

There's a nice example of this here - https://www.totaltypescript.com/discriminated-unions-are-a-devs-best-friend

DafyddLlyr: Another option (which I think would work?) could be just to call the `useAnalyticsTracking()` hook within this component to save passing the function along as props.

No right or wrong answers here necessarily!

DafyddLlyr (commented):
Looking great! Meta-review of the PR - great to see this level of detail with the what / why / screenshots headers. Super helpful and great 🏅 

A few comments on the PR in response to your questions - very happy to jump on a call and talk them through tomorrow or leave it with you. Whatever you think would be best.

If you push a commit without the `[skip ci]` tag it'll spin up a Pizza where this can be tested and interacted with.

A few questions to consider - 
- What happens if a user clicks "Help" and an external link on the `NextSteps` component? I think the `_append` GraphQL operator actually handles this from memory but worth checking.

- How do you think we should handle a user who clicks a link (which opens in another tab) and then clicks comes back to the PlanX tab and clicks a different one? Should we care about this or are we fine with just the latest link they clicked?

editor.planx.uk/src/ui/NextStepsList.tsx
Mike-Heneghan: > Another option (which I think would work?) could be just to call the `useAnalyticsTracking()` hook within this component to save passing the function along as props.
> 
> No right or wrong answers here necessarily!

I had tried to do this but I think I was encountering issues around using the hook in the component. I suspect it might just be a syntax thing which I'm struggling to recall 😅

Mike-Heneghan: > Good question. This is happening because the `url` property is optional in `ListItemProps` despite actually being required in `LinkStep`.
> 
> There's a few ways to expressing this in TypeScript - I'd probably recommend trying to reimplement `ListItemProps` as a discriminated union of two types - one where `url` and `handleNextStepsLinkClick` are not there, and one where they're required.
> 
> There's a nice example of this here - https://www.totaltypescript.com/discriminated-unions-are-a-devs-best-friend

I don't think I've seen the union of types pattern before but it looks brilliant! I think it definitely solves for frustrations I've previously had 😌

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -10,16 +10,20 @@ export type AnalyticsType = "init" | "resume";
 type AnalyticsLogDirection = AnalyticsType | "forwards" | "backwards";
 
 export type HelpClickMetadata = Record<string, string>;
+export type NextStepsLinkMetadata = Record<string, string>;
```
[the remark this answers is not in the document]
Mike-Heneghan: What could tightening this type up look like? Would we explicitly expect a key of 'url' and a string which would be the value? I'm no too sure how I would improve this, it's been too long since I've worked with types 😂

Mike-Heneghan:
> Looking great! Meta-review of the PR - great to see this level of detail with the what / why / screenshots headers. Super helpful and clear 🏅
> 
> A few comments on the PR in response to your questions - very happy to jump on a call and talk them through tomorrow or leave it with you. Whatever you think would be best.
> 
> If you push a commit without the `[skip ci]` tag it'll spin up a Pizza where this can be tested and interacted with.
> 
> A few questions to consider -
> 
> * What happens if a user clicks "Help" and an external link on the `NextSteps` component? I think the `_append` GraphQL operator actually handles this from memory but worth checking.
> * How do you think we should handle a user who clicks a link (which opens in another tab) and then clicks comes back to the PlanX tab and clicks a different one? Should we care about this or are we fine with just the latest link they clicked?

I'm glad the PR format itself has been useful, I'm always happy to take of feedback with this sort of thing. 

Would you generally recommend pushing WIP work with `[skip ci]` until it's at a point where the Pizza env / other CI checks would be useful?

- I just double checked and if the user clicks "Help" and an external link on the `NextSteps` component it does indeed update the record as we would intend so the `_append` is covering that 😌 

- That's an excellent question re how to handle users selecting multiple next step links. I had wondered about this and what would make sense I think some potential options could be:

  - Overwrite the URL selected as is currently being done (we'll lose the previous link clicks)
  - Insert a new analytics_log per click (we'll create duplicates but preserve user selections)
  - Update the value of the `url` key to instead be an array of selected urls (preserves data without creating duplicates although _might_ add complexity to queries with the more complex data structure)

Mike-Heneghan:
> > Looking great! Meta-review of the PR - great to see this level of detail with the what / why / screenshots headers. Super helpful and clear 🏅
> > A few comments on the PR in response to your questions - very happy to jump on a call and talk them through tomorrow or leave it with you. Whatever you think would be best.
> > If you push a commit without the `[skip ci]` tag it'll spin up a Pizza where this can be tested and interacted with.
> > A few questions to consider -
> > 
> > * What happens if a user clicks "Help" and an external link on the `NextSteps` component? I think the `_append` GraphQL operator actually handles this from memory but worth checking.
> > * How do you think we should handle a user who clicks a link (which opens in another tab) and then clicks comes back to the PlanX tab and clicks a different one? Should we care about this or are we fine with just the latest link they clicked?
> 
> I'm glad the PR format itself has been useful, I'm always happy to take of feedback with this sort of thing.
> 
> Would you generally recommend pushing WIP work with `[skip ci]` until it's at a point where the Pizza env / other CI checks would be useful?
> 
> * I just double checked and if the user clicks "Help" and an external link on the `NextSteps` component it does indeed update the record as we would intend so the `_append` is covering that 😌
> * That's an excellent question re how to handle users selecting multiple next step links. I had wondered about this and what would make sense I think some potential options could be:
>   
>   * Overwrite the URL selected as is currently being done (we'll lose the previous link clicks)
>   * Insert a new analytics_log per click (we'll create duplicates but preserve user selections)
>   * Update the value of the `url` key to instead be an array of selected urls (preserves data without creating duplicates although _might_ add complexity to queries with the more complex data structure)

After discussion the plan is to maintain an array of URLs which the user has selected. Having a look online it doesn't appear that we can append to an array directly with the mutation as the level of nesting we have within Metadata as far as I can see. 

As discussed another option would be to locally maintain the array of selected urls and append to this before firing the mutation.

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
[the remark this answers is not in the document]
Mike-Heneghan: As discussed in a call this was resolved with `Record<'selectedUrls', string[]` i.e. more specific key.

DafyddLlyr (approved):
Look good and working as expected on the Pizza 👍 

Please feel free to merge to main @Mike-Heneghan and we'll include this in the next production deploy.

image

PR #2284, opened 2023-10-05
feat: add option to use "Pay without pay" and show error in cases of undefined fee

jessicamcinchak:
**Changes:**
- Adds Editor option to "Hide the pay buttons and show fee for information only" (off by default)
  - Renders [like this in the public interface](https://storybook.2284.planx.pizza/?path=/story/planx-components-pay--for-information-only)
- Updates how `NaN` / `undefined` fees are handled
  - Current behavior: if the data field referenced by the Pay component (`props.fn`) is undefined when Pay is reached, then automatically skip the Pay component. This led to real applications being successfully submitted without paying because of a content error a few weeks ago.
  - New behavior: if `props.fn` is undefined when Pay is reached, display an error and do not let the applicant continue. An airbrake error including the `sessionId` is also thrown/should be picked up in Slack. 
  
**Testing:**
- Try the different paths in this flow and confirm you have expected behavior: https://2284.planx.pizza/lambeth/pay-without-pay-test/preview
- Confirm that teams that do not have GOV.UK Pay enabled still show the expected error: https://2284.planx.pizza/barking-and-dagenham/pay-test/preview (specifically with a "Continue" button that does not prevent you from proceeding with your test)

**Future scope / not addressed here:**
- The Pay component is still automatically skipped in cases where the fee = 0. My understanding is we don't want to "show" this case until we can properly design how to display "reasons" (eg resubmission or exemption or other calculation)
- Show a breakdown of how the fee was calculated ([currently on roadmap here](https://miro.com/app/board/uXjVPrebzwQ=/?moveToWidget=3458764558166771974&cot=14) :crystal_ball:)

### DIFF
--- a/editor.planx.uk/src/@planx/components/Pay/Public/Pay.test.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Public/Pay.test.tsx
@@ -161,7 +161,8 @@ describe("Confirm component without inviteToPay", () => {
 
   it("displays an error and continue-with-testing button if Pay is not enabled for this team", async () => {
     const handleSubmit = jest.fn();
-    const errorMessage = "No pay token found for this local authority";
+    const errorMessage =
+      "GOV.UK Pay is not enabled for this local authority (testing)";
 
     const { user } = setup(
       <Confirm
--- a/editor.planx.uk/src/@planx/components/Pay/Public/Pay.test.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Public/Pay.test.tsx
@@ -89,7 +89,7 @@ describe("Confirm component without inviteToPay", () => {
 
   it("displays an error and continue-with-testing button if Pay is not enabled for this team", async () => {
     const handleSubmit = jest.fn();
-    const errorMessage = "No pay token found for this team!";
+    const errorMessage = "No pay token found for this local authority";
 
     const { user } = setup(
       <Confirm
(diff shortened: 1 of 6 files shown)


jessicamcinchak:
[Display rather than hide undefined or zero Pay components (Pay without pay)](https://trello.com/c/KWSwZqEt/2544-display-rather-than-hide-undefined-or-zero-pay-components-pay-without-pay)

jessicamcinchak:
[Allow Pay component to display fee for information only without payment option](https://trello.com/c/CwFTY5nn/2652-allow-pay-component-to-display-fee-for-information-only-without-payment-option)

editor.planx.uk/src/@planx/components/Pay/Public/Pay.test.tsx
```diff
@@ -21,21 +22,92 @@ jest
...
+    setState({ flow: flowWithZeroFee, breadcrumbs: breadcrumbs });
+    expect(getState().computePassport()).toEqual({
+      data: { "application.fee.payable": ["0"] },
+    });
 
-it("should not have any accessibility violations", async () => {
-  const handleSubmit = jest.fn();
-  const { container } = setup(<Pay handleSubmit={handleSubmit} />);
-  const results = await axe(container);
-  expect(results).toHaveNoViolations();
```
jessicamcinchak, line 38: This accessibility level test at the top-level of Pay was actually meaningless (it was testing the "skipped scenario" :ghost:), so now there's 3 accessibility tests running against the `Confirm` child component container: 
Screenshot from 2023-10-06 09-44-45

editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
```diff
@@ -42,6 +43,9 @@ enum Action {
   Success,
 }
 
+export const PAY_API_ERROR_UNSUPPORTED_TEAM =
+  "GOV.UK Pay is not enabled for this local authority";
```
jessicamcinchak, line 47: Gentle reminder that Pay & Planning Constraints specifically have a dependency between their API error messages and frontend component states. API errors were recently updated here to append team name on the end and broke the frontend matching here, which I also didn't catch in previous code reviews! 

The default fronted ErrorWrapper that'll catch API errors doesn't render with a "Continue" button and fully stops a tester from proceeding, but in cases of Pay & Constraints we match on the specific "not enabled" error to show a warning while _still allowing testers to proceed_ past the Component.

DafyddLlyr: Thanks - I was hitting this yesterday afternoon whilst reviewing it and was also looking into it 🙌

DafyddLlyr: This is a good reminder that there are a few places we do this matching on error messages as opposed to statuses (I know there's a few in Save & Return / Invite to Pay also). This might be a sign that something like an enum of statuses may be a better pattern that we could consider using where it's not as simple as relying on a HTTP code.

```ts
{
  message: "Some changeable text",
  errorStatus: PayError.UnsuportedTeam
}
```

In future, it might also be nice to cover these instances with E2E tests so we get better visibility on breaking changes.

editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
```diff
@@ -101,11 +109,17 @@ function Component(props: Props) {
...
   useEffect(() => {
-    if (isNaN(fee) || fee <= 0) {
-      // skip the pay component because there's no fee to charge
+    // Auto-skip component when fee=0
+    if (fee <= 0) {
       return props.handleSubmit({ auto: true });
     }
 
+    // If props.fn is undefined, display & log an error
+    if (isNaN(fee)) {
```
DafyddLlyr, line 118: Whilst making this change could we also visually handle this a little more gracefully in the component which displays the value?

image

DafyddLlyr (approved):
Looking great! All behaviour working as expected - thanks for also adding more a11y tests which is always good to see.

A few small comments on the code, but no show stoppers.

editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
jessicamcinchak: Yeah I didn't think much of this because it should really _never_ be seen :sweat_smile:

Now updated to display "Unknown": 
Screenshot from 2023-10-06 10-43-28

editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
```diff
@@ -68,7 +69,11 @@ function Component(props: Props) {
   const reducer = (_state: ComponentState, action: Action): ComponentState => {
     switch (action) {
       case Action.NoPaymentFound:
-        return { status: "init" };
+        if (isNaN(fee)) {
+          return { status: "undefined_fee" };
```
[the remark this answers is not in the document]
jessicamcinchak: Yep that's totally fair, now dispatching a new unique event :+1:

DafyddLlyr (approved):

PR #2285, opened 2023-10-06
fix: Correctly parse and transform strings to date in `/webhook` APIs

DafyddLlyr:
Hit this issue when troubleshooting E2E test (this doesn't fix them yet! 😅)

The change introduced in https://github.com/theopensystemslab/planx-new/pull/2252 validates that incomming requests are valid (e.g. dates), and then hands over the request type to the `ValidatedRequestHandler` type.

What I missed is that we're never passing along the parsed request to the next middleware layer (but it does pass the type as mentioned above).

This means we end up with strings being passed into functions that expect `Date` objects. When running E2E tests a number of API requests fail as a result.

### DIFF
--- a/api.planx.uk/modules/webhooks/service/lowcalSessionEvents/schema.ts
+++ b/api.planx.uk/modules/webhooks/service/lowcalSessionEvents/schema.ts
@@ -4,7 +4,7 @@ import { ScheduledEventResponse } from "../../../../hasura/metadata";
 
 export const createSessionEventSchema = z.object({
   body: z.object({
-    createdAt: z.string().transform((val) => new Date(val)),
+    createdAt: z.string().pipe(z.coerce.date()),
     payload: z.object({
       sessionId: z.string(),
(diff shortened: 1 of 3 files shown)


api.planx.uk/modules/webhooks/service/lowcalSessionEvents/schema.ts
```diff
@@ -4,7 +4,7 @@ import { ScheduledEventResponse } from "../../../../hasura/metadata";
 
 export const createSessionEventSchema = z.object({
   body: z.object({
-    createdAt: z.string().transform((val) => new Date(val)),
+    createdAt: z.string().pipe(z.coerce.date()),
```
DafyddLlyr, line 7: This is functionally the same, but syntactically a little nicer https://github.com/colinhacks/zod/discussions/879#discussioncomment-4646183

jessicamcinchak (approved):

PR #2288, opened 2023-10-09
feat: Modularise and document analytics endpoints

DafyddLlyr:
## What does this PR do?
 - Converts analytics REST API endpoints to modular structure proposed for API
 - Adds test coverage and Swagger docs

I needed a break from E2E tests and it seemed worthwhile getting this quickly tidied up before @Mike-Heneghan may want to expand on this.

image

### DIFF
--- a/api.planx.uk/modules/analytics/docs.yaml
+++ b/api.planx.uk/modules/analytics/docs.yaml
@@ -1,0 +1,25 @@
+openapi: 3.1.0
+info:
+  title: Plan✕ API
+  version: 0.1.0
+tags:
+  - name: analytics
+paths:
+  /analytics/log-user-exit:
+    post:
+      summary: Log user exit
+      description: Capture an analytic event which represents a user exiting a service
+      tags:
+        - analytics
+      responses:
+        "204":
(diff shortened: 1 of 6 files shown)


api.planx.uk/modules/analytics/docs.yaml
```diff
@@ -0,0 +1,25 @@
...
+          description: Successful response - no awaited response from server
+  /analytics/log-user-resume:
+    post:
+      summary: Log user resume
+      description: Capture an analytic event which represents a user resuming a service
+      tags:
+        - analytics
+      responses:
+        "204":
+          description: Successful response - no awaited response from server
```
jessicamcinchak, line 25: nit: can either the post description or response description be tweaked to include "via the Beacon API" or similar so it's more clear why our "success" here is no awaited response?

jessicamcinchak (approved):

api.planx.uk/modules/analytics/docs.yaml
DafyddLlyr: Very good catch - change made! Will merge when CI is happy 👌

PR #2291, opened 2023-10-09
feat: Automatically grant users access to the "Templates" team

DafyddLlyr:
## What does this PR do?
 - Using a trigger, automatically adds new users to the "Templates" team
 - Removes the responsibility for handling this end from the frontend
 - Adds E2E test coverage of this new behaviour
 - Adds basic tests for `canUserEditTeam()` logic in `UserStore`

### DIFF
--- a/hasura.planx.uk/migrations/1696884217237_grant_new_user_template_team_access/up.sql
+++ b/hasura.planx.uk/migrations/1696884217237_grant_new_user_template_team_access/up.sql
@@ -1,0 +1,28 @@
+CREATE OR REPLACE FUNCTION grant_new_user_template_team_access() RETURNS trigger AS $$
+DECLARE
+    templates_team_id INT;
+BEGIN
+    SELECT id INTO templates_team_id FROM teams WHERE slug = 'templates';
+    IF templates_team_id IS NOT NULL THEN
+        INSERT INTO team_members (user_id, team_id, role) VALUES (NEW.id, templates_team_id, 'teamEditor');
+    END IF;
+    
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER grant_new_user_template_team_access AFTER INSERT ON users
+    FOR EACH ROW EXECUTE PROCEDURE grant_new_user_template_team_access();
+
+COMMENT ON TRIGGER grant_new_user_template_team_access ON users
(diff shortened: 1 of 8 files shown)


DafyddLlyr:
[Automatically add all new users to the "Templates" team](https://trello.com/c/IdH2fU07/2651-automatically-add-all-new-users-to-the-templates-team)

hasura.planx.uk/migrations/1696884217237_grant_new_user_template_team_access/up.sql
```diff
@@ -0,0 +1,28 @@
...
+
+-- Insert a record to team_members for all existing users
+INSERT INTO
+  team_members (user_id, team_id, role)
+SELECT
+  id,
+  29,
+  'teamEditor'
+FROM
+  users;
```
jessicamcinchak, line 28: I expected the pizza's `team_members` table to include a record for each existing user, but the only member of  `team_id` 29 in the pizza is my freshly added user - so I don't think this final `insert` has actually ran? https://hasura.2291.planx.pizza/console/data/default/schema/public/tables/team_members/browse?filter=team_id%3B%24eq%3B29

jessicamcinchak (commented):
Working as expected for me when adding a _new_ user, but left a comment below about _existing_ users !

hasura.planx.uk/migrations/1696884217237_grant_new_user_template_team_access/up.sql
DafyddLlyr: Ah this is interesting. Thanks for spotting - will take a look!

DafyddLlyr: So this is happening on Pizzas because the sync script is dropping the `team_members` table - `TRUNCATE TABLE team_members CASCADE;`

This happens after all Hasura migrations are run (which adds the new `team_members` records). What I think this means is...
 - This is ok as is
 - We'll probably want to get this to prod fairly sharpish so that the new team_members can cascade from prod -> staging and prod -> pizzas and not get truncated

jessicamcinchak: That explanation makes perfect sense! We can confirm this works for existing users on staging first then before a deploy :+1:

jessicamcinchak (approved):

PR #2294, opened 2023-10-11
feat: Create new `api` role in Hasura

DafyddLlyr:
## What does this PR do?
 - Adds a new `api` role in Hasura
 - Adds introspection test coverage for user role

## What's the motivation behind this?
I'm working towards https://trello.com/c/qSCoDRAO/2650-implement-editor-permissions-throughout-the-api which has two main aims - 
 - When a request to the API is initiated by a user, use their permission level throughout the call stack
 - When a request to the API is initiated by an internal service (e.g. Hasura), use appropriate permissions (not admin token with full "God mode")

The first step has been pretty well started in previous PRs. Adding an `api` role will allow us to scope service → API requests appropriately when a service has authenticated to the API (e.g. through the `useHasuraAuth()` middleware).

The permission set outlined here is based on queries currently made by the two admin GraphQL clients currently used in the API.

## Next steps...
- Drop `$admin` client from API
- Drop `adminGraphQLClient` from API
- Replace them with a role-scoped client. Either a user's role, or the new `api` role for service → API requests.
- Where a user-initiated request requires a "side effect" (such as adding an audit log), I think it would be appropriate to call the API-scoped client. An alternative could be using [backend only mutations](https://hasura.io/docs/latest/auth/authorization/permissions/backend-only/) but this feels like a more complex solution.

I'll likely split the above into a few PRs just to make them a bit more manageable.

## Nice to have?
Porting the introspection tests from JavaScript / Jest to Gherkin could be a really nice tidy up. I think it would be approx 10% of the number of lines of code that we currently have and be a lot easier to maintain and parse. This is not a very high priority tbh!

### DIFF
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -62,6 +62,52 @@
 - table:
     schema: public
     name: bops_applications
+  insert_permissions:
+    - role: api
+      permission:
+        check: {}
+        columns:
+          - id
+          - req_headers
+          - request
+          - response
+          - response_headers
+          - bops_id
+          - destination_url
+          - session_id
+          - created_at
+          - sanitised_at
+  select_permissions:
+    - role: api
+      permission:
+        columns:
+          - id
+          - req_headers
+          - request
+          - response
+          - response_headers
+          - bops_id
+          - destination_url
+          - session_id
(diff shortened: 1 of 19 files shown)


jessicamcinchak (approved):
All really clear, thanks for continuing to pick this up !

PR #2295, opened 2023-10-11
fix: Left align 'Start new application' link

DafyddLlyr:
Spotted this one today!

| Before | After |
|--------|--------|
| image | image |


ianjon3s (approved):
Thanks for catching this one!

PR #2299, opened 2023-10-12
chore: add/update logic tests for auto-answers

jessicamcinchak:
see also existing filters test `"Filter path nodes are auto-answered correctly when a lower order flag is picked up first"`

goal here is to add clear test cases that fail and outline expected behavior, then have a baseline to compare against before I start editing store functions!


jessicamcinchak:
Replaced by #2510

PR #2300, opened 2023-10-12
feat: Drop `publicGraphQLClient` in favour of `$public` client

DafyddLlyr:
## What does this PR do?
 - Replaces `publicGraphQLClient` in favour of `$public`
 - No functionality change, just a tidy up and simplification
 - Small permission change with `getClientForTemplate()` now conditionally using the `api` role instead of the `admin` role

### DIFF
--- a/api.planx.uk/inviteToPay/sendPaymentEmail.test.ts
+++ b/api.planx.uk/inviteToPay/sendPaymentEmail.test.ts
@@ -6,16 +6,11 @@ import {
   validatePaymentRequestNotFoundQueryMock,
   validatePaymentRequestQueryMock,
 } from "../tests/mocks/inviteToPayMocks";
+import { CoreDomainClient } from "@opensystemslab/planx-core";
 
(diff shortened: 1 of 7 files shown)


api.planx.uk/inviteToPay/sendPaymentEmail.test.ts
```diff
@@ -6,16 +6,11 @@ import {
...
-jest.mock("@opensystemslab/planx-core", () => {
-  return {
-    CoreDomainClient: jest.fn().mockImplementation(() => ({
-      formatRawProjectTypes: jest
-        .fn()
-        .mockResolvedValue(["New office premises"]),
-    })),
-  };
-});
+jest
```
DafyddLlyr, line 11: Using spies like this actually feels much nicer than mocks as it keeps the function type safe - it forces us to pass in valid return types of functions.

jessicamcinchak (approved):

PR #2302, opened 2023-10-12
feat: query & display Flood Risk Zones in PlanningConstraints

jessicamcinchak:
Changes:
- Adds values `flood`, `flood.zone.1`, `flood.zone.2` and `flood.zone.3` to the planning constraints check (only adds granular entries if `flood` is true, consistent with Article 4 granular variables)

I hit two quirks with Planning Data that I've raised with Matt L, and expect to be updated soon on their side. When it is, the dropdown will populate as expected, no extra code changes required (so no need to block initial merge I think!):
- Planning data hasn't populated the "text" for `flood-risk-zone` metadata (which is used to render the contextual paragraphs in the dropdown)
- Planning data hasn't populated the "name" for `flood-risk-zone` entities (which is used to render the bullet-pointed list in the dropdown)

Address in a flood zone for testing: 47, COBOURG ROAD, SOUTHWARK, SE5 0HU

### DIFF
--- a/editor.planx.uk/src/@planx/components/PlanningConstraints/List.tsx
+++ b/editor.planx.uk/src/@planx/components/PlanningConstraints/List.tsx
@@ -23,6 +23,7 @@ const CATEGORY_COLORS: Record<string, string> = {
   "Heritage and conservation": "#EDDCD2",
   Ecology: "#E0EFCC",
   Trees: "#DBE7E4",
+  Flooding: "#ECECEC",
 };
 
 interface StyledConstraintProps extends BoxProps {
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/PlanningConstraints/List.tsx
```diff
@@ -23,6 +23,7 @@ const CATEGORY_COLORS: Record<string, string> = {
   "Heritage and conservation": "#EDDCD2",
   Ecology: "#E0EFCC",
   Trees: "#DBE7E4",
+  Flooding: "#ECECEC",
```
jessicamcinchak, line 26: Open to better suggestions than grey here! Easy to adjust later though.

DafyddLlyr (approved):

PR #2305, opened 2023-10-13
chore: Convert `/download-schema` endpoint to `$public` client

DafyddLlyr:
There may be a few of these incoming - keeping them small and easy should it make it more readable and manageable. There's not much test coverage for some of these steps (and I'm not sure how valuable they would be?) so hopefully small PRs would mean they're a little more easily testable as we go along.


jessicamcinchak (approved):
Sorry I missed this one last week!

PR #2306, opened 2023-10-13
feat: track a user restarting their application via a new api route

Mike-Heneghan:
## What

- Add a new analytics endpoint route to handle a updating the `"flow_direction"` of an application restart to a new type `"reset"`. 
- Hit the new endpoint when a user clicks "confirm" when restarting via the `Header` restart before the page reloads. 
- As per: https://trello.com/c/spEYlVk0/2566-track-which-node-users-are-on-when-they-click-to-restart-application-or-start-over

## Why

- We want to be able to log on which nodes users decide to restart their applications. 
- We already have a concept of `"flow_direction"` with `"reset"` being a descriptive name for the direction change being to the start of the application.  
- I was having issues with the mutation being called in the FE before the page reloaded. 
- Using Beacon guarantees we can update the `analytics_log` record.

## Screen Recording 

https://github.com/theopensystemslab/planx-new/assets/36415632/011362a0-a250-4a2b-b82c-bfb1898e6f1f

## Related PRs

- This might supersede https://github.com/theopensystemslab/planx-new/pull/2301

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
@@ -6,7 +6,7 @@ import React, { createContext, useContext, useEffect, useState } from "react";
 
 import { Store, useStore } from "./store";
 
-export type AnalyticsType = "init" | "resume";
+export type AnalyticsType = "init" | "resume" | "reset"
 type AnalyticsLogDirection = AnalyticsType | "forwards" | "backwards";
 
 export type HelpClickMetadata = Record<string, string>;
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -6,7 +6,7 @@ import React, { createContext, useContext, useEffect, useState } from "react";
 
 import { Store, useStore } from "./store";
 
-export type AnalyticsType = "init" | "resume";
+export type AnalyticsType = "init" | "resume" | "reset"
```
Mike-Heneghan, line 9: Although the [ticket](https://trello.com/c/spEYlVk0/2566-track-which-node-users-are-on-when-they-click-to-restart-application-or-start-over) suggests using `"reset"` I think `"restart"` might be more descriptive and consistent.

Mike-Heneghan:
Superseded

PR #2307, opened 2023-10-13
feat: track user resetting flow

Mike-Heneghan:
## What

- Add a new mutation to update the `flow_direction` of an `analytics_log` to a new type `reset`. 
- Call this mutation on confirming a reset when clicking the reset icon in the header.
- Call this mutation when a user clicks the "reset" of a `Notice` with a reset.
- As per: 
  - https://trello.com/c/spEYlVk0/2566-track-which-node-users-are-on-when-they-click-to-restart-application-or-start- 
  - https://trello.com/c/G1bTFyVm/2564-track-notice-reset-status-in-node-metadata

## Why
- `{flow_direction: "reset"}` will show when a user has decided to restart. 
- We can use this to infer data about user behaviour such as how often users are hitting dead ends or from which nodes they're deciding to restart. 

## Screen Recording

Shows:
- Change to `reset` when resetting via the header.
- No change on regular `Notice`  button press.
- Change on `reset` `Notice` button press.

https://github.com/theopensystemslab/planx-new/assets/36415632/bdf161fd-b052-41fd-9850-780d60a2c988

## Supersedes
- https://github.com/theopensystemslab/planx-new/pull/2306
- https://github.com/theopensystemslab/planx-new/pull/2301

## Note

- This still shows the bug that when restarting from a notice the first node re-renders unnecessarily.

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
@@ -6,7 +6,7 @@ import React, { createContext, useContext, useEffect, useState } from "react";
 
 import { Store, useStore } from "./store";
 
-export type AnalyticsType = "init" | "resume";
+export type AnalyticsType = "init" | "resume" | "reset"
 type AnalyticsLogDirection = AnalyticsType | "forwards" | "backwards";
 
 export type HelpClickMetadata = Record<string, string>;
@@ -18,11 +18,13 @@ const analyticsContext = createContext<{
   createAnalytics: (type: AnalyticsType) => Promise<void>;
   trackHelpClick: (metadata?: HelpClickMetadata) => Promise<void>;
   trackNextStepsLinkClick: (metadata?: SelectedUrlsMetadata) => Promise<void>;
+  trackResetFlow: () => Promise<void>;
   node: Store.node | null;
 }>({
   createAnalytics: () => Promise.resolve(),
   trackHelpClick: () => Promise.resolve(),
   trackNextStepsLinkClick: () => Promise.resolve(),
+  trackResetFlow: () => Promise.resolve(),
   node: null,
 });
 const { Provider } = analyticsContext;
@@ -104,6 +106,7 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
         createAnalytics,
         trackHelpClick,
         trackNextStepsLinkClick,
+        trackResetFlow,
         node,
       }}
     >
@@ -194,6 +197,31 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
         variables: {
           id: lastAnalyticsLogId,
           metadata,
+        },
+      });
+    }
+  }
+
+
+  async function trackResetFlow() {
+    if (shouldTrackAnalytics && lastAnalyticsLogId) {
+      await publicClient.mutate({
+        mutation: gql`
+          mutation UpdateHasResetFlow(
+            $id: bigint!
+            $flow_direction: String
+          ) {
+            update_analytics_logs_by_pk(
+              pk_columns: { id: $id }
+              _set: { flow_direction: $flow_direction }
+            ) {
+              id
+            }
+          }
+        `,
+        variables: {
+          id: lastAnalyticsLogId,
+          flow_direction: 'reset'
         },
       });
     }
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -44,6 +44,7 @@
     - role: public
       permission:
         columns:
+          - flow_direction
           - has_clicked_help
           - metadata
         filter: {}
--- a/editor.planx.uk/src/components/Header.tsx
+++ b/editor.planx.uk/src/components/Header.tsx
@@ -315,6 +315,8 @@ const PublicToolbar: React.FC<{
   const showCentredServiceTitle = useMediaQuery((theme: Theme) =>
     theme.breakpoints.up("md"),
   );
+  
+  const { trackResetFlow } = useAnalyticsTracking();
 
   const handleRestart = async () => {
     if (
@@ -322,6 +324,7 @@ const PublicToolbar: React.FC<{
         "Are you sure you want to restart? This will delete your previous answers",
       )
     ) {
+      trackResetFlow();
       if (path === ApplicationPath.SingleSession) {
         clearLocalFlow(id);
         window.location.reload();


editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -199,6 +202,31 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     }
   }
 
+  // Comment to trigger new pizza build.
```
Mike-Heneghan, line 205: Vultr isn't behaving itself still 😅 

I logged in and killed the instances and DNS records for this PR i.e. `2307` but it seems to be taking a very long time. 

I'll remove this comment and troubleshoot further next week.

jessicamcinchak: :eyes: pizza is looking okay now to me, is this resolved or do you have any outstanding vultr bugs/questions?

Mike-Heneghan: I think it's resolved 😅 

The situation was: 
  - Pizza failed to build
  - I deleted the instances and DNS records for it on Vultr
  - The pizza built but the vultr health check failed. 
  - I pushed a commit
  - The vultr health check passed 🤷 
  
  I'm pretty confident the PR behaves as expected.

Mike-Heneghan: If you're happy to review @jessicamcinchak I'll merge / update as per feedback 👍

editor.planx.uk/src/@planx/components/Notice/Public.tsx
```diff
@@ -75,6 +76,13 @@ const NoticeComponent: React.FC<Props> = (props) => {
     ? () => props.handleSubmit?.()
     : undefined;
 
+  const { trackResetFlow } = useAnalyticsTracking();
+
+  const handleNoticeResetClick = () => {
+    trackResetFlow()
+    props.resetPreview && props.resetPreview()
+  }
```
jessicamcinchak, line 84: nit: row end semi-colons in the body here would make this more readable for me? (i'm surprised linting/husky didn't pick this up actually!)

jessicamcinchak, line 84: nit: row end semi-colons in the body here would make this more readable for me? (i'm surprised linting/husky didn't pick this up actually!)

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -44,6 +44,7 @@
     - role: public
       permission:
         columns:
+          - flow_direction
```
jessicamcinchak, line 47: :heavy_check_mark:

jessicamcinchak (approved):
Working as expected for me and database structure looks good :+1: 

Super easy to understand, really nice analytics-value to number-of-line code change ratio here I think :slightly_smiling_face:

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
jessicamcinchak: A couple tips for future Vultru debugging:
- If only the healthcheck fails, Vultr might just be slow to warm up - wait a couple minutes and try re-running just the failed GH action job to save yourself commits (it usually passes on the next go!)
- Vultr fails to build entirely for a couple common reasons: timeouts (usually around 10m mark), commits pushed close together and an "update" trys to run before the "create" is finished. In these cases, again first trying to re-run just the failed GH action job before doing a full nuke can save a lot of time!

editor.planx.uk/src/@planx/components/Notice/Public.tsx
Mike-Heneghan: Fixed here: ba123628fbfe319d4d1f22391c498b6e0532a155

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -6,7 +6,7 @@ import React, { createContext, useContext, useEffect, useState } from "react";
 
 import { Store, useStore } from "./store";
 
-export type AnalyticsType = "init" | "resume";
+export type AnalyticsType = "init" | "resume" | "reset"
```
Mike-Heneghan, line 9: Hey @jessicamcinchak something I hadn't realised is that this `AnalyticsType` is related to the `analytics` model which tracks the overarching analytics session? It stores whether the user's session was an `init` or a `resume`?

I believe the log direction only applies to `analytics_log` model instances which are tied together with an instance of the `analytics` model?

Based on the way the `"reset"` is being recorded against an `analytics_log` if it make more sense to think of it as a log direction i.e. an extreme case of going `backwards` where you go back to the the start?

editor.planx.uk/src/@planx/components/Notice/Public.tsx
Mike-Heneghan: I re-ran `pnpm install` in `editor.planx.uk` and that installed husky and it appears to be working now. 

I'm not really sure why that was required although it's working now 🤷

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
jessicamcinchak: I'm not entirely sure I'm following this one yet, but my instinct here is that "reset" _ends the given session_ and therefore, at least in my mind, it's an analytics type rather than pure "direction" (eg "backwards" is always _within_ a session). But also totally trust your judgement on this!

Mike-Heneghan: I think that's a really good point although in this PR the `"reset"` is only ever applied within a session as a `flow_direction`

As in:
- The last `analytics_log` for that `analytic` session would have a `flow_direction` of `"reset"`. 
- The `type` of the aforementioned `analytic` session would remain whatever `type` it had been i.e. `"init"` or `"resume"`
- The new `analytic` session would default to `"init"`

So currently the `AnalyticType` is only applied as a `flow_direction`.  

Does that make sense? I'm maybe over focussing on the semantics but I'm just trying to get my head around who things have been working and what we're looking for with this new functionality.

Mike-Heneghan: That's a very helpful framing of the situation, thanks Jess! 

I'll have a look at "Where applicants drop-off" and see how it's being approached there and if there are any implications with this change.

Mike-Heneghan: I had a wee look at the SQL query:

```
SELECT date_trunc('week', analytics_logs.created_at) as "Week", count(*) as "Exits"
FROM analytics_logs
JOIN analytics ON (analytics.id = analytics_logs.analytics_id)
WHERE analytics_logs.user_exit = true
AND analytics_logs.node_type = 3
AND analytics.flow_id = {{flow_id}}::uuid
GROUP BY "Week"
ORDER BY "Week"
```

The `drop-off` uses the `user_exit` property which to my understanding is marked as `true` when a user stops viewing the page:

https://github.com/theopensystemslab/planx-new/blob/a6362685baeb3b3c5068cde3bb499dcc50d6ffc4/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx#L60-L86

I had thought of a user deciding to `reset` as not actually exiting the page and therefore not being a `user_exit: true`. 

Although in the context of drop off I'm not so sure 🤔

Mike-Heneghan: Actually on second thoughts I think we'll be able to handle the `reset` state in a similar fashion as the query above so we should be okay I think 👍

Mike-Heneghan:
[Track Notice "reset" status in node metadata](https://trello.com/c/G1bTFyVm/2564-track-notice-reset-status-in-node-metadata)

Mike-Heneghan:
[Track which node users are on when they click to "Restart application" or "Start over"](https://trello.com/c/spEYlVk0/2566-track-which-node-users-are-on-when-they-click-to-restart-application-or-start-over)

PR #2308, opened 2023-10-14
fix: Add security header for `/webhooks` API routes

DafyddLlyr:
Restores `hasuraAuth` option to submission and webhook endpoints

image


jessicamcinchak (approved):
Thanks for this!!

PR #2313, opened 2023-10-16
feat: Explicitly handle redaction

DafyddLlyr:
## What does this PR do?
- Updates `planx-core` to pick up the changes implemented in https://github.com/theopensystemslab/planx-core/pull/162
- This avoids repeated calls to the expensive `computeBOPSParams()` function. Local testing indicates that this approximately halves the time it takes to hit the `/admin` endpoints being updated here.

### DIFF
--- a/api.planx.uk/send/exportZip.ts
+++ b/api.planx.uk/send/exportZip.ts
@@ -39,7 +39,7 @@ export async function buildSubmissionExportZip({
   // add OneApp XML to the zip
   if (includeOneAppXML) {
     try {
-      const xml = await $api.generateOneAppXML(sessionId);
+      const xml = await $api.export.oneAppPayload(sessionId);
       const xmlStream = str(xml.trim());
       await zip.addStream({
         name: "proposal.xml", // must be named "proposal.xml" to be processed by Uniform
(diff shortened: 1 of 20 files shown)


api.planx.uk/send/exportZip.ts
```diff
@@ -65,7 +65,8 @@ export async function buildSubmissionExportZip({
   }
 
   // generate csv data
-  const { responses, redactedResponses } = await $api.export.csvData(sessionId);
+  const responses = await $api.export.csvData(sessionId);
+  const redactedResponses = await $api.export.csvDataRedacted(sessionId);
```
DafyddLlyr, line 69: This is the only place where both redacted and un-redacted payloads were used. This means that these two calls will now be _slightly_ less efficient as the `ExportClient` is making additional requests to the DB (e.g. to fetch session data).

I think this is a fair compromise as... 
- We only need both in one place (vs. all other methods touched in this PR)
- Fetching from the DB is the quick / "cheap" part of the function that's being called under the hood - and our GraphQL client (or DB / Hasura) is very likely doing some caching of these requests for us anyway.

jessicamcinchak (approved):

PR #2321, opened 2023-10-17
chore: update local development instructions in README

jessicamcinchak:
We got the repo up & running on an external fork today in less than an hour :tada: :fork_and_knife: 

Updating our docs with a couple gotchas we spotted along the way.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,10 +6,12 @@
 API_PORT=7002
 API_URL_EXT=http://localhost:${API_PORT}
 
+# JWT_SECRET must be at least 32 characters
 JWT_SECRET=👻
 
 SESSION_SECRET=👻
 
+# Google Cloud OAuth 2.0 Client ID
 GOOGLE_CLIENT_ID=👻
 GOOGLE_CLIENT_SECRET=👻
 
@@ -49,9 +51,9 @@ MINIO_ADMIN_PORT=9001
 
 # PostgreSQL
 PG_PORT=7001
-PG_DATABASE=👻
+PG_DATABASE=planxdb
+PG_USERNAME=dbuser
 PG_PASSWORD=👻
-PG_USERNAME=👻
 
 # PG user with permission to sync content between environments
 PRODUCTION_PG_URL_FOR_USER_GITHUB_ACTIONS=👻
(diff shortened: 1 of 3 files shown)


.env.example
```diff
@@ -49,9 +51,9 @@ MINIO_ADMIN_PORT=9001
 
 # PostgreSQL
 PG_PORT=7001
-PG_DATABASE=👻
+PG_DATABASE=planxdb
+PG_USERNAME=dbuser
```
jessicamcinchak, line 55: this is already exposed in infrastructure/data too

Mike-Heneghan: _nit_: I'm maybe a little bit paranoid but I'd lean towards an abundance of caution removing them from  infrastructure/data and not exposing them here. Although as I say it's more of a feeling than anything else 😅

README.md
```diff
@@ -48,6 +51,30 @@ planx-new is a monorepo containing our full application stack. Here's a quick su
...
+
+5. Move into the hasura directory `cd ../hasura.planx.uk` and install dependencies `pnpm i`.
+
+6. Open [Hasura's](https://hasura.io/) web console (`cd hasura.planx.uk` then `pnpm start`) and add your Google email address to the `users` table. You'll also likely want to create an initial `team`. This will eventually allow you to authenticate into the application as an admin.
+
+7. Follow steps 7-9 above to start the editor and login !
+
+At this point you'll be running the full Planx application locally in a docker container. See our Github Actions pull request workflow as an example of how to deploy via docker to a virtual linux server, or explore the `infrastructure` directory for how to deploy via Pulumi infrastructure-as-code to AWS. 
+
+We'd love to hear what you're building on Planx, don't hesitate to get in touch with questions.
```
Mike-Heneghan, line 76: _suggestion_: it might be nice to point at a preferred contact method if we have one?

Mike-Heneghan (approved):
Really nice to see this added and I'm excited to see what this could lead to 🥳 

I'm maybe being mega paranoid about DB names etc. so I don't think it's blocking this from going out 😊

.env.example
jessicamcinchak: This is a fair comment! A bit more context: this database name is only used in Docker envs (which will never have "real" session data), our staging & production databases on AWS RDS have different names. The username is shared across envs but has already been exposed for awhile and isn't sensitive on its own and was annoyingly required to successfully spin up the DB. 

Security-wise, I think working towards more restricted RDS access altogether will give us a much better peace of mind!

Mike-Heneghan: That makes a lot of sense, thanks for the extra context 🙌

PR #2328, opened 2023-10-18
fix: Handle unique file list correctly

DafyddLlyr:
## What's the problem?
- When constructing a `FileList` for display to the user, we want this to only contain unique files (by name)
- We're experiencing a bug where a file is missing from the `FileList`
  - There are two `FileType`s with the same name
  - Only one of their conditions is met
  - Neither appear in the final list ❌ 
- Described in detail with example flow [here on OSL Slack](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1697646369616819)

## What's the cause?
- The unique list of file names we maintain does not check if the file has been added or not to the `FileList`

## What's the solution?
- Check if a file is added by the `populateFileList()` function
- Only count it as a unique file if added to our `FileList`

## Demo
Broken flow on prod - https://editor.planx.uk/testing/upload-and-label-recommended-test ❌ 
Fixed flow on pizza - https://2328.planx.pizza/testing/upload-and-label-recommended-test ✅

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/model.test.ts
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/model.test.ts
@@ -164,6 +164,78 @@ describe("createFileList function", () => {
     expect(result).toEqual(expected);
   });
 
+  it("correctly handles unique file names", () => {
+    const fileTypes: FileType[] = [
+      {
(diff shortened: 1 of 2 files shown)


DafyddLlyr:
image

August has given this the green light 🟢

PR #2332, opened 2023-10-19
chore: Overwrite records on data-sync conflict

DafyddLlyr:
See context on OSL Slack - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1696942132535689

Currently, we don't overwrite DB records on conflict - we skip them. This can lead to inconsistencies (as on the above thread), could potentially lead to permissions diverging between staging and prod, and also causes this new issue raised here where the prod boundary has not been copied across to staging - https://trello.com/c/hreL5mn4/2665-site-doesnt-have-address-showing-london-map-extract-on-bucks-service

### DIFF
--- a/scripts/seed-database/write/flows.sql
+++ b/scripts/seed-database/write/flows.sql
@@ -34,7 +34,7 @@ SELECT
   settings,
   copied_from
 FROM sync_flows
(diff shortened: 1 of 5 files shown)


jessicamcinchak (approved):

PR #2335, opened 2023-10-20
feat: first pass reordering uploaded files to show most recent first

Mike-Heneghan:
## What

- Update the `FileUploadAndLabel` component to order the slots with the most recently uploaded at the top.
- This includes the main page and the labelling modal.
- As per: https://trello.com/c/YDoIk2tG/2663-upload-and-label-new-files-to-appear-at-top-of-modal

## Why

- In user testing it was found that users were not seeing their most recently uploaded files as they were not being clearly displayed as they were rendered at the bottom of the list.
- This caused issues for labelling and frustration for users. 

## How

- As per @jessicamcinchak's comment reversed the slot order in the `useEffect`
- Changing the order doesn't break the logic for uploading and labelling.

## Screenshots

The files uploaded and their labels as seen by the user:

Screenshot 2023-10-26 at 12 17 25

Subsequent update to the data with the labels assigned as expected:

Screenshot 2023-10-26 at 12 26 05

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
@@ -95,7 +95,7 @@ function Component(props: Props) {
   useEffect(() => {
     if (previousSlotCount === undefined) return;
 
-    //  *** Show most recent upload at the top ***
+    // Show most recent upload at the top
     if (slots.length) setSlots(slots.reverse());
 
     // Only stop modal opening on initial return to node
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
@@ -94,6 +94,10 @@ function Component(props: Props) {
   const previousSlotCount = usePrevious(slots.length);
   useEffect(() => {
     if (previousSlotCount === undefined) return;
+
+    //  *** Show most recent upload at the top ***
+    if (slots.length) setSlots(slots.reverse());
+
     // Only stop modal opening on initial return to node
     if (isUserReturningToNode) return setIsUserReturningToNode(false);
     if (slots.length && dropzoneError) setDropzoneError(undefined);


editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
```diff
@@ -89,6 +89,7 @@ function Component(props: Props) {
   }, []);
 
   const [slots, setSlots] = useState<FileUploadSlot[]>([]);
+  const orderByMostRecentlyUploaded = true;
```
Mike-Heneghan, line 92: I'm not sure about this pattern to be honest but I was reluctant to directly change the order of the `slots` and I liked the idea of having one variable at the parent level to determine the order.

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.test.tsx
Mike-Heneghan, line 1: The testing I've added here is super clunky. 

I was thinking about changing tact to move from static definitions like:

```TypeScript
const fileTaggingModal = await within(document.body).findByTestId(
  "file-tagging-dialog",
);
``` 

Which aren't reliable later on depending on rendering to instead be functions which wrap the finding so can be reused like: 

```TypeScript
async function findFileTaggingModal(): Promise<HTMLElement> {
  const modal = await within(document.body).findByTestId(
    "file-tagging-dialog",
  )
  return modal
};
```
I'm not sure if it adds much value though 🤔

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
Mike-Heneghan, line 1: I don't think the approach I've taken to reordering data will break things but I'm not sure 😅 

It'll be useful to check once the Pizza instance has been built.

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
jessicamcinchak: I'm totally on board with just wanting to define order in one high-level place, but personally I think a boolean prop to control this is at the component level is a bit unnecessary and would never be changed? 

When user research makes a suggestion for a design change based on feedback, and it's not a style attribute that needs to be exposed to editors, then we should just full out adopt it in my opinion and no need to add the internal complexity of making it configurable.

jessicamcinchak (commented):
(For later this week when you're back!!)

This might be a completely naive suggestion, but I'm curious if this one-line change to `/FileUploadAndLabel/Public.tsx` would solve the re-ordering with less complexity? 

So, in the `useEffect` that's listening for changes to `slots`:
```ts
  useEffect(() => {
    if (previousSlotCount === undefined) return;
    // Only stop modal opening on initial return to node
    if (isUserReturningToNode) return setIsUserReturningToNode(false);
    if (slots.length && dropzoneError) setDropzoneError(undefined);
    if (!slots.length && fileListError) setFileListError(undefined);
    if (slots.length > previousSlotCount) setShowModal(true);
    if (slots.length) setSlots(slots.reverse()); // Show most recent upload at the top
  }, [slots.length]);
```

By altering `slots` once at this level, that'll then get passed through as props to the list & modal. I'm not positive it should be the last statement to run in that block, might be worth experimenting with order?

If this works in practice, then let's talk tests more next!

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
jessicamcinchak: Looks like the data sync failed on this pizza, I'll remove & re-run soon !

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
Mike-Heneghan: That's a very good point @jessicamcinchak I think I still have the mindset of my last job where we were generally always A/B testing and tended to follow this sort of pattern. I'll get this update 😊

Mike-Heneghan:
> (For later this week when you're back!!)
> 
> This might be a completely naive suggestion, but I'm curious if this one-line change to `/FileUploadAndLabel/Public.tsx` would solve the re-ordering with less complexity?
> 
> So, in the `useEffect` that's listening for changes to `slots`:
> 
> ```ts
>   useEffect(() => {
>     if (previousSlotCount === undefined) return;
>     
>     //  *** Show most recent upload at the top ***
>     if (slots.length) setSlots(slots.reverse());
>     
>     // Only stop modal opening on initial return to node
>     if (isUserReturningToNode) return setIsUserReturningToNode(false);
>     if (slots.length && dropzoneError) setDropzoneError(undefined);
>     if (!slots.length && fileListError) setFileListError(undefined);
>     if (slots.length > previousSlotCount) setShowModal(true);
>   }, [slots.length]);
> ```
> 
> By altering `slots` once at this level, that'll then get passed through as props to both the list and modal without requiring any extra `.reverse()` calls etc.
> 
> If this works in practice for you too, then let's talk tests more next!

Thanks for this comment @jessicamcinchak it's a much simpler solution to achieve the same result. 

I'd been reticent to change the `slots` and focused on changing the presentation as I was a bit concerned it would break the logic for file uploads. Although I don't think this is the case with the way things have been implemented.  

This definitely looks like it's working for me locally with what I can test without having my file uploads working. I'll use the Pizza to double check.

Mike-Heneghan:
> (For later this week when you're back!!)
> 
> This might be a completely naive suggestion, but I'm curious if this one-line change to `/FileUploadAndLabel/Public.tsx` would solve the re-ordering with less complexity?
> 
> So, in the `useEffect` that's listening for changes to `slots`:
> 
> ```ts
>   useEffect(() => {
>     if (previousSlotCount === undefined) return;
>     
>     //  *** Show most recent upload at the top ***
>     if (slots.length) setSlots(slots.reverse());
>     
>     // Only stop modal opening on initial return to node
>     if (isUserReturningToNode) return setIsUserReturningToNode(false);
>     if (slots.length && dropzoneError) setDropzoneError(undefined);
>     if (!slots.length && fileListError) setFileListError(undefined);
>     if (slots.length > previousSlotCount) setShowModal(true);
>   }, [slots.length]);
> ```
> 
> By altering `slots` once at this level, that'll then get passed through as props to both the list and modal without requiring any extra `.reverse()` calls etc.
> 
> If this works in practice for you too, then let's talk tests more next!

In terms of testing the change if this works in the Pizza I'd be pretty tempted to not add an additional test where I had one before? I believe it'll very noticeably work or not and I'm not sure the amount of code recorded to test it is justified?

jessicamcinchak (approved):
Thanks for making this update, working as expected for me on the pizza :slightly_smiling_face: 

nit: the `***` are a little flashy for a regular code comment, I really just meant them as emphasis on the snippet - please feel free to remove before merging :sweat_smile: 

agree that testing isn't a priority for such a small change like this one, much more important to focus testing efforts on the known labelling bug :+1:

DafyddLlyr (commented):
One small comment ([see discussion here on Slack](https://opensystemslab.slack.com/archives/C5Q59R3HB/p1698220801038089)). 

It would be nice to keep this behaviour consistent across both `FileUpload` and `FileUploadAndLabel` if possible. I'm pretty sure there's a function somewhere that's just `push()`ing a file into a slot on upload - if we just `unshift()` onto the front of the array I think this will mean most recent is at the "top" of the list and it's ordered as expected.

Happy for this to be a follow on PR if easier!

Mike-Heneghan:
> One small comment ([see discussion here on Slack](https://opensystemslab.slack.com/archives/C5Q59R3HB/p1698220801038089)).
> 
> It would be nice to keep this behaviour consistent across both `FileUpload` and `FileUploadAndLabel` if possible. I'm pretty sure there's a function somewhere that's just `push()`ing a file into a slot on upload - if we just `unshift()` onto the front of the array I think this will mean most recent is at the "top" of the list and it's ordered as expected.
> 
> Happy for this to be a follow on PR if easier!

Thanks for this follow up @DafyddLlyr! As discussed I think the best path at the moment might be sticking with this and reviewing the `FileUpload` component behaviour with some more input down the line.

PR #2342, opened 2023-10-25
fix: Increase spacing above "Name the site" input

DafyddLlyr:
Quick fix for an outstanding issue I noticed when recently discussing this component.

The `MapContainer` currently does not encapsulate the `<MapFooter>` correctly as the `MyMap` element within it is styled to have `height: "100%"`. This means that any elements beneath it (such as the "Name the site" input) are not spaced correctly.

| | Before | After |
|--------|--------|--------|
|   **Spacing**   | image | image |
|   **Map Container**   | image | image |


jessicamcinchak (approved):
Thanks for picking this one up, much better!

PR #2343, opened 2023-10-26
chore: Fix Run API test type warnings

DafyddLlyr:
## What does this PR do?
- Address some eslint warnings in API tests
- Imports and uses types from `planx-core`

I've been staring at the message below a fair amount these past few days trying to sort E2E tests, so this is just a quick pass at addressing a few of these issues.

image

## Next steps...
The eventual aim would be to remove the Passport / Flow etc types from the API and just use the ones from `planx-core`. This isn't a small change though so I'll just try chipping away at them here and there as a maintenance task 👍

### DIFF
--- a/api.planx.uk/editor/publish.ts
+++ b/api.planx.uk/editor/publish.ts
@@ -4,8 +4,13 @@ import { adminGraphQLClient as adminClient } from "../hasura";
 import { dataMerged, getMostRecentPublishedFlow } from "../helpers";
 import { gql } from "graphql-request";
 import intersection from "lodash/intersection";
-import { ComponentType } from "@opensystemslab/planx-core/types";
+import {
+  ComponentType,
+  FlowGraph,
+  Node,
+} from "@opensystemslab/planx-core/types";
 import { userContext } from "../modules/auth/middleware";
+import type { Entry } from "type-fest";
 
 const validateAndDiffFlow = async (
   req: Request,
@@ -139,9 +144,9 @@ type ValidationResponse = {
   description?: string;
 };
 
-const validateSections = (flow: Record<string, any>): ValidationResponse => {
-  if (getSectionNodeIds(flow)?.length > 0) {
-    if (!sectionIsInFirstPosition(flow)) {
+const validateSections = (flowGraph: FlowGraph): ValidationResponse => {
+  if (getSectionNodeIds(flowGraph)?.length > 0) {
+    if (!sectionIsInFirstPosition(flowGraph)) {
       return {
         isValid: false,
         message: "Cannot publish an invalid flow",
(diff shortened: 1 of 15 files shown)


api.planx.uk/editor/publish.ts
```diff
@@ -241,27 +251,40 @@ const validateInviteToPay = (flow: Record<string, any>): ValidationResponse => {
   };
 };
 
-const inviteToPayEnabled = (flow: Record<string, any>): boolean => {
-  const payNodeStatuses = Object.entries(flow)
-    .filter(([_nodeId, nodeData]) => nodeData?.type === ComponentType.Pay)
-    ?.map(([_nodeId, nodeData]) => nodeData?.data?.allowInviteToPay);
+const inviteToPayEnabled = (flowGraph: FlowGraph): boolean => {
+  const payNodes = Object.entries(flowGraph).filter(
+    (entry): entry is [string, Node] =>
```
DafyddLlyr, line 256: The type narrowing here has to be repeated as it doesn't propagate as expected from the function which is being called.

https://github.com/microsoft/TypeScript/issues/12798

api.planx.uk/editor/publish.ts
```diff
@@ -241,27 +251,40 @@ const validateInviteToPay = (flow: Record<string, any>): ValidationResponse => {
...
     payNodeStatuses.every((status) => status === true)
   );
 };
 
+const isComponentType = (
+  entry: Entry<FlowGraph>,
+  type: ComponentType,
+): entry is [string, Node] => {
+  const [nodeId, node] = entry;
+  if (nodeId === "_root") return false;
```
DafyddLlyr, line 273: An alternative (and possibly better) solution to this would be to add type property to the root node (`ComponentType._Root`?).

api.planx.uk/types.ts
```diff
@@ -1,13 +1,19 @@
...
 export interface Node {
   id?: string;
   data?: Record<string, any>;
   edges?: Array<string>;
   type?: number;
 }
 
+/**
+ * @deprecated Migrating to Flow and FlowGraph from planx-core
+ */
```
jessicamcinchak, line 16: :+1: these are helpful reminders, thanks!

jessicamcinchak (approved):
Thanks for these tidy ups! I can try to revisit those lingering Digital Land ones soon - I think it's just a matter of typing the shape the expected response from their Entity & Metadata endpoints.

api.planx.uk/types.ts
DafyddLlyr: Yep - hopefully it just means when we next need to reach for a type we avoid these ones 👌

PR #2345, opened 2023-10-26
chore: Bump API task CPU

DafyddLlyr:
See https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_size

This should double the API vCPU from 0.5 → 1, which should help with Uniform submission issues.

image


jessicamcinchak (approved):

Mike-Heneghan (approved):
Looks good 👍 

The syntax feels weird for CPU but thanks for linking to the docs to explain.

PR #2351, opened 2023-10-27
fix: Improve flakey E2E tests

DafyddLlyr:
## What does this PR do?
 - Reduces / removes flakiness of `user data persists on page refresh @regression` E2E Playwright test
 - I've re-run this multiple times on CI (7-10 times?) without a failure so I've got a good degree of confidence it's better than the current implementation
 - This is achieved by wrapping actions in a `Promise.all()`

## Docs 
- https://github.com/microsoft/playwright/issues/5470
- https://www.workwithloop.com/blog/our-1-solution-to-playwright-flakiness-waitforresponse-waitforrequest-promises

### DIFF
--- a/e2e/tests/ui-driven/src/create-flow/helpers.ts
+++ b/e2e/tests/ui-driven/src/create-flow/helpers.ts
@@ -1,12 +1,10 @@
 import { Browser, Page, Request } from "@playwright/test";
 import { createAuthenticatedSession } from "../globalHelpers";
 
-export const getUserRequest = (req: Request) => {
+export const isGetUserRequest = (req: Request) => {
   const isHasuraRequest = req.url().includes("/graphql");
-  const isGetUserRequest = Boolean(
-    isHasuraRequest && req.postData()?.toString().includes("GetUserById"),
-  );
-  return isGetUserRequest;
+  const isGetUserOperation = req.postData()?.toString().includes("GetUserById");
+  return Boolean(isHasuraRequest && isGetUserOperation);
 };
 
 export async function getAdminPage({
--- a/e2e/tests/ui-driven/src/create-flow/helpers.ts
+++ b/e2e/tests/ui-driven/src/create-flow/helpers.ts
@@ -1,0 +1,39 @@
+import { Browser, Page, Request } from "@playwright/test";
+import { createAuthenticatedSession } from "../globalHelpers";
+
+export const getUserRequest = (req: Request) => {
+  const isHasuraRequest = req.url().includes("/graphql");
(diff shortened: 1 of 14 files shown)


e2e/tests/ui-driven/src/create-flow/helpers.ts
```diff
@@ -0,0 +1,39 @@
+import { Browser, Page, Request } from "@playwright/test";
+import { createAuthenticatedSession } from "../globalHelpers";
+
+export const getUserRequest = (req: Request) => {
+  const isHasuraRequest = req.url().includes("/graphql");
+  const isGetUserRequest = Boolean(
+    isHasuraRequest && req.postData()?.toString().includes("GetUserById"),
+  );
+  return isGetUserRequest;
+};
```
Mike-Heneghan, line 10: _Question / Naming:_

This function is `getUserRequest` but it seems like it takes in a request as an argument and checks whether it's a user request? Rather than literally getting the request? If this is the case would it make more sense to reflect that in the name?

Mike-Heneghan (commented):
Hey @DafyddLlyr 

I tried having wee look over this and although it looks good to me I'm very aware I'm not very familiar with our E2E tests. 

Maybe we could use this PR as a vehicle for discussing E2E?

e2e/tests/ui-driven/src/invite-to-pay/agent.spec.ts
```diff
@@ -15,9 +15,9 @@ import {
   navigateToPayComponent,
 } from "./helpers";
 import { mockPaymentRequest, modifiedInviteToPayFlow } from "./mocks";
-import { saveSession } from "../helpers";
-import { returnToSession } from "../helpers";
-import { clickContinue } from "../helpers";
+import { saveSession } from "../globalHelpers";
+import { returnToSession } from "../globalHelpers";
+import { clickContinue } from "../globalHelpers";
```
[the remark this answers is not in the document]
DafyddLlyr: Good spot! I think it's ended up like this as a result of some file renaming and reorganising.

e2e/tests/ui-driven/src/create-flow/helpers.ts
DafyddLlyr: Good catch thanks - renamed!

DafyddLlyr:
> Maybe we could use this PR as a vehicle for discussing E2E?

Thanks for the reminder - put a slot in the calendar tomorrow morning and we can cover this 👍

jessicamcinchak (approved):

PR #2353, opened 2023-10-30
fix: remove reset logic which cleared all slots

Mike-Heneghan:
## What

- Remove the `resetAllSlots` logic from `updateFileListWithTags`.
- Remove initial definition of `updatedFileList` from `updateFileListWithTags`.
- Refactor from `merge(fileList)` to `cloneDeep(fileList)`

## Why

- A bug was encountered where if a file had one tag which was then removed it would erroneously trigger the `resetAllSlots` which would wipe the tags applied to all files.
- As a tag can either be added or removed one at a time it was unclear what the use case for the reset was so it was removed to resolve the bug. 
- During debugging some opportunities for refactoring were found this included:
  - Moving from single argument `merge` to `cloneDeep` as `merge` was not a deep clone leading us to mutate state directly.
  - Removing the initial definition of `updatedFileList` from the `updateFileListWithTags` function as the initial value was never used. 

## Screen recordings 

Bug recording:

https://github.com/theopensystemslab/planx-new/pull/2209#issuecomment-1708237134

 Fix recording:

https://github.com/theopensystemslab/planx-new/assets/36415632/71aae0b5-bcad-427a-89d2-09ff01c4b5ab

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Modal.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Modal.tsx
@@ -16,7 +16,6 @@ import Select, { SelectChangeEvent, SelectProps } from "@mui/material/Select";
 import { styled } from "@mui/material/styles";
 import Typography from "@mui/material/Typography";
 import capitalize from "lodash/capitalize";
-import merge from "lodash/merge";
 import React, { useEffect, useState } from "react";
 import { usePrevious } from "react-use";
 import ErrorWrapper from "ui/ErrorWrapper";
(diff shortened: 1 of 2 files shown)


Mike-Heneghan:
[Error message in Upload & Label that prevents users from saving and moving to the next page (other similar bugs previously seen)](https://trello.com/c/Kc3ifwND/2632-error-message-in-upload-label-that-prevents-users-from-saving-and-moving-to-the-next-page-other-similar-bugs-previously-seen)

DafyddLlyr (approved):
Perfect! ✅ 

We can merge this to `main` (staging) and request PO testing there on the Trello ticket 👍

PR #2354, opened 2023-10-31
fix: Don't overwrite `req` with `parsedReq`

DafyddLlyr:
## What's the problem?
A "reminder" email failed this morning - the reason for this was that the required "email" field was missing from the event sent to Hasura.

## What caused this?
This field was missed when [typing the schema for this endpoint](https://github.com/theopensystemslab/planx-new/pull/2252). This happened as the `routeSendEmailRequest()` function is not yet typed - meaning the variable was missed.

## What's the solution?
 - Add "email" to schema
 - Add test cases
 - Also, don't overwrite `req` with `parsedReq` - this change means that non-updated endpoints shouldn't fail as the original `req` object is not touched. I'm now writing to `res.locals` ([docs](https://expressjs.com/en/api.html#res.locals)) which is a nice solution to the issue I was talking about in a previous dev call. We can now coerce types as part of validation in the middleware, and not break existing endpoints - this feels like a solid solution here.

## Next steps...
I think it's pretty likely that multiple reminder and expiry events setup since this change was made are affected by this problem of a missing email field. These events should be identifiable from the Hasura `hdb_catalog` tables, and we could then append emails to them. I'll update this thread as I troubleshoot this one.

This might also point towards considering a cron job for these events as an alternative to manage reminder and expiry events. Something to consider!

### DIFF
--- a/api.planx.uk/modules/webhooks/controller.ts
+++ b/api.planx.uk/modules/webhooks/controller.ts
@@ -16,7 +16,7 @@ import { SanitiseApplicationData } from "./service/sanitiseApplicationData/types
 import { sanitiseApplicationData } from "./service/sanitiseApplicationData";
 
 export const sendSlackNotificationController: SendSlackNotification = async (
-  req,
+  _req,
   res,
   next,
 ) => {
@@ -27,8 +27,9 @@ export const sendSlackNotificationController: SendSlackNotification = async (
     });
   }
 
-  const eventData = req.body.event.data.new;
-  const eventType = req.query.type;
+  const { body, query } = res.locals.parsedReq;
+  const eventData = body.event.data.new;
+  const eventType = query.type;
 
   try {
     const data = await sendSlackNotification(eventData, eventType);
@@ -46,7 +47,9 @@ export const sendSlackNotificationController: SendSlackNotification = async (
 export const createPaymentInvitationEventsController: CreatePaymentEventController =
   async (req, res, next) => {
     try {
-      const response = await createPaymentInvitationEvents(req.body);
+      const response = await createPaymentInvitationEvents(
+        res.locals.parsedReq.body,
+      );
       res.json(response);
     } catch (error) {
       return next(
@@ -61,7 +64,9 @@ export const createPaymentInvitationEventsController: CreatePaymentEventControll
 export const createPaymentReminderEventsController: CreatePaymentEventController =
   async (req, res, next) => {
     try {
-      const response = await createPaymentReminderEvents(req.body);
+      const response = await createPaymentReminderEvents(
+        res.locals.parsedReq.body,
+      );
       res.json(response);
     } catch (error) {
       return next(
@@ -76,7 +81,9 @@ export const createPaymentReminderEventsController: CreatePaymentEventController
 export const createPaymentExpiryEventsController: CreatePaymentEventController =
   async (req, res, next) => {
     try {
-      const response = await createPaymentExpiryEvents(req.body);
+      const response = await createPaymentExpiryEvents(
+        res.locals.parsedReq.body,
+      );
       res.json(response);
     } catch (error) {
       return next(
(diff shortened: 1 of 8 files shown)


DafyddLlyr:
Right, I've now got some SQL I'm happy with which is tested on staging and locally. 

We can run this manually on production once this fix is merged (and the SQL is cross-checked please...!).

Events without an `email` can be found with this query - 

```sql
SELECT
  *
FROM
  hdb_catalog.hdb_scheduled_events
WHERE
  webhook_conf::TEXT ILIKE '%expiry%'
  AND status = 'scheduled'
  AND payload->'email' IS NULL;
```

Records can be updated with - 

```sql
UPDATE
  hdb_catalog.hdb_scheduled_events
SET
  payload = jsonb_set(
    payload::JSONB,
    '{email}',
    ('"' || public.lowcal_sessions.email || '"')::JSONB
  )
FROM
  public.lowcal_sessions
WHERE
  webhook_conf::TEXT ILIKE '%expiry%'
  AND status = 'scheduled'
  AND payload -> 'email' IS NULL
  AND hdb_scheduled_events.payload->>'sessionId' = public.lowcal_sessions.id::TEXT;
```

`%expiry%` and `%reminder%` would both need to be run - I guess we could combine them into one query but I'd favour a step by step approach juuust to be safe.

api.planx.uk/modules/webhooks/controller.ts
```diff
@@ -16,7 +16,7 @@ import { SanitiseApplicationData } from "./service/sanitiseApplicationData/types
 import { sanitiseApplicationData } from "./service/sanitiseApplicationData";
 
 export const sendSlackNotificationController: SendSlackNotification = async (
-  req,
+  _req,
```
Mike-Heneghan, line 19: _Question:_

Am I right in saying `_req` is still being passed in despite not being used directly as down the chain it is required?

Mike-Heneghan (approved):
Looks good to me 👍

Mike-Heneghan:
> Right, I've now got some SQL I'm happy with which is tested on staging and locally.
> 
> We can run this manually on production once this fix is merged (and the SQL is cross-checked please...!).
> 
> Events without an `email` can be found with this query -
> 
> ```sql
> SELECT
>   *
> FROM
>   hdb_catalog.hdb_scheduled_events
> WHERE
>   webhook_conf::TEXT ILIKE '%expiry%'
>   AND status = 'scheduled'
>   AND payload->'email' IS NULL;
> ```
> 
> Records can be updated with -
> 
> ```sql
> UPDATE
>   hdb_catalog.hdb_scheduled_events
> SET
>   payload = jsonb_set(
>     payload::JSONB,
>     '{email}',
>     ('"' || public.lowcal_sessions.email || '"')::JSONB
>   )
> FROM
>   public.lowcal_sessions
> WHERE
>   webhook_conf::TEXT ILIKE '%expiry%'
>   AND status = 'scheduled'
>   AND payload->'email' IS NULL
>   AND hdb_scheduled_events.payload->>'sessionId' = public.lowcal_sessions.id::TEXT;
> ```
> 
> `%expiry%` and `%reminder%` would both need to be run - I guess we could combine them into one query but I'd favour a step by step approach juuust to be safe.

This looks as if it would do what you're intending as far as I can see. Can you remind me where you'd actually run this? Also, I'm happy to jump on a call to go through this 👍

jessicamcinchak (approved):
Changes & proposed SQL look good to me here :+1: 

One question about the SQL - the `email` param is only expected in payloads for _session_ reminders and expiry events, right? Not, `payment_` reminder & expiry too? Just double checking we're catching everything expected with this update.

DafyddLlyr:
> One question about the SQL - the `email` param is only expected in payloads for _session_ reminders and expiry events, right? Not, `payment_` reminder & expiry too? Just double checking we're catching everything expected with this update.

This is a good question! I've double checked this and invite to pay emails just need a `paymentRequestId` and nothing else 👌 

https://github.com/theopensystemslab/planx-new/blob/4e291379d59856b5bb6c1d38f82aedbf53d3f1ba/hasura.planx.uk/metadata/tables.yaml#L895-L902

DafyddLlyr:
I'm going to go ahead and merge this - I'll run the SQL once this has got to production and will then update this PR with a paper trail confirming it's been done successfully.

api.planx.uk/modules/webhooks/controller.ts
DafyddLlyr: That's correct, `_req` is always passed in, but no longer used in this instance. We're using the underscore prefix to indicate that this is an unused variable - there's a linting rule set up for this in `.eslintrc` under `@typescript-eslint/no-unused-vars`

PR #2356, opened 2023-11-01
chore: Drop `adminClient` (1/2)

DafyddLlyr:
## What does this PR do?
 - Belated follow on to https://github.com/theopensystemslab/planx-new/pull/2294
- Remove `adminClient` (still in "god mode" using the Hasura admin secret)
- Replace with permission scoped `$public` or `$api` clients
- I've split this in two to keep things simpler and smaller. [Part 2 here](https://github.com/theopensystemslab/planx-new/pull/2375)!

## Principles
 - Queries should be typed (by default `planx-core` is returning an `unknown` otherwise). For simplicity I've just used the mutation / query name as the interface name. If this isn't clear / meaningful I can append a `Query` / `Mutation` suffix.
 
 - Where possible, use queries as the place where we convert snake_case (from the db) to camelCase (in the application). This means our interfaces and types are all camelCase as expected. This has meant cascading (noisy) changes to code and tests but it seemed worthwhile over introducing more snake_case properties to types.
 
 - Allow `$api` client for side-effects triggered by permission-scoped users (such as writing audit logs). The alternative here would be to increase public client scope and add [backend-only mutations](https://hasura.io/docs/latest/auth/authorization/permissions/backend-only/) - see previous thoughts on this [here](https://github.com/theopensystemslab/planx-new/pull/2294).


hasura.planx.uk/metadata/tables.yaml
```diff
@@ -1116,6 +1116,16 @@
           - message
           - session_id
           - created_at
+  select_permissions:
```
DafyddLlyr, line 1119: This was required to give the API role access to the `insert_reconciliation_requests_one` mutation.

> _`insert_<object>_one` will only be available if you have select permissions on the table, as it returns the inserted row_

Docs:  https://hasura.io/docs/latest/mutations/postgres/insert/#insert-a-single-object

api.planx.uk/saveAndReturn/resumeApplication.test.ts
```diff
@@ -4,31 +4,33 @@ import app from "../server";
...
-      formatRawProjectTypes: () => mockFormatRawProjectTypes(),
-    })),
+    CoreDomainClient: class extends actualCoreDomainClient {
+      constructor() {
+        super();
+        this.formatRawProjectTypes = () => mockFormatRawProjectTypes();
+      }
+    },
   };
 });
```
DafyddLlyr, line 29: This structure is repeated a few times throughout this PR. I spent a good bit of time attempting to generalise it so that we could import it with an interface similar to - 

```ts
function mockCoreDomainClient(customMocks = {}) {
   const actualCoreDomainClient = jest.requireActual(
    "@opensystemslab/planx-core",
  ).CoreDomainClient;

  return {
    CoreDomainClient: class extends actualCoreDomainClient {
      constructor() {
        super();
        Object.assign(this, customMocks)
      }
    }
  }
}

...

mockCoreDomainClient({
  formatRawProjectTypes: () => mockFormatRawProjectTypes(),
});
```

Tried a few variations without much success - I suspect the problem is related to how Jest is hoisting mocks (through the `jest.mock("@opensystemslab/planx-core")` notation). I wanted to just move on as opposed to losing more time here so I'll open a GitHub issue describing this.

DafyddLlyr: This structure is repeated a few times throughout this PR. I spent a good bit of time attempting to generalise it so that we could import it with an interface similar to - 

```ts
function mockCoreDomainClient(customMocks = {}) {
   const actualCoreDomainClient = jest.requireActual(
    "@opensystemslab/planx-core",
  ).CoreDomainClient;

  return {
    CoreDomainClient: class extends actualCoreDomainClient {
      constructor() {
        super();
        Object.assign(this, customMocks)
      }
    }
  }
}

...

mockCoreDomainClient({
  formatRawProjectTypes: () => mockFormatRawProjectTypes(),
});
```

Tried a few variations without much success - I suspect the problem is related to how Jest is hoisting mocks (through the `jest.mock("@opensystemslab/planx-core")` notation). I wanted to just move on as opposed to losing more time here so I'll open a GitHub issue describing this.

api.planx.uk/inviteToPay/createPaymentSendEvents.test.ts
```diff
@@ -52,7 +52,7 @@ describe("Create payment send events webhook", () => {
       name: "GetTeamSlugByFlowId",
       matchOnVariables: false,
       data: {
-        flows_by_pk: {
+        flows: {
```
jessicamcinchak, line 55: nit: `flows_by_pk` returns a singular `flow` ?

api.planx.uk/saveAndReturn/resumeApplication.ts
```diff
@@ -42,6 +41,11 @@ const resumeApplication = async (
   }
 };
 
+interface ValidateRequest {
+  teams: Team[];
+  sessions: LowCalSession[] | null;
```
jessicamcinchak, line 46: nit: I think typing `lowcal_sessions` as `sessions` is quite confusing given we have a new, empty, to-be-eventually-populated table `sessions` - can we be wordy but clear here and use `lowcalSessions` ?

api.planx.uk/saveAndReturn/resumeApplication.ts
```diff
@@ -81,16 +85,19 @@ const validateRequest = async (
...
+        email: email.toLowerCase(),
+      },
+    );
 
     if (!teams?.length) throw Error;
 
     return {
       team: teams[0],
-      sessions: lowcal_sessions,
+      sessions: sessions || [],
```
jessicamcinchak, line 100: same comment as above here - this re-naming feels like it changes the whole table reference!

jessicamcinchak (approved):
Thanks for following through with all these updates and thorough explanations of how you've set up mocks here! 

Left a few naming suggestions for clarity, but overall looks good :+1:

api.planx.uk/saveAndReturn/resumeApplication.test.ts
DafyddLlyr: Issue logged here - https://github.com/theopensystemslab/planx-new/issues/2404

PR #2357, opened 2023-11-01
feat: track a user navigating backwards through flow by node id

Mike-Heneghan:
## What

- Users can initiate moving backwards through the flow by node id via the `back` or `change` button.
- Track users initiating this and store the initiator i.e `back` or `change` and store target node metadata i.e. `title` and `type`.

## Why

- Currently we can't clearly infer when a user initiated moving backwards through a flow via the `back` or `change` button.
- Added some basic data of the target node being navigated back to gain insight on which question are challenging and which are often changed

## Screen Recording

https://github.com/theopensystemslab/planx-new/assets/36415632/4d32450a-6bc3-46f5-a2ab-bde0398c85fb

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
@@ -312,10 +312,8 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     initiator: BackwardsNaviagtionInitiatorType,
   ) {
     const targetNodeMetadata = getTitleAndTypeFromFlow(nodeId);
-    console.log("Target node info: ", targetNodeMetadata);
     const metadata: Record<string, NodeMetadata> = {};
     metadata[`${initiator}`] = targetNodeMetadata;
-    console.log("Metadata payload:", metadata);
 
     if (shouldTrackAnalytics && lastAnalyticsLogId) {
       await publicClient.mutate({
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
@@ -21,6 +21,7 @@ type AnalyticsLogDirection =
 
 export type HelpClickMetadata = Record<string, string>;
 export type SelectedUrlsMetadata = Record<"selectedUrls", string[]>;
+export type BackwardsNaviagtionInitiatorType = "change" | "back";
 
 type NodeMetadata = {
   flagset?: FlagSet;
@@ -29,6 +30,8 @@ type NodeMetadata = {
     description?: string;
   };
   flag?: Flag;
+  title?: string;
+  type?: TYPES;
 };
 
 let lastAnalyticsLogId: number | undefined = undefined;
@@ -40,12 +43,17 @@ const analyticsContext = createContext<{
   trackFlowDirectionChange: (
     flowDirection: AnalyticsLogDirection,
   ) => Promise<void>;
+  trackBackwardsNavigationByNodeId: (
+    nodeId: string,
+    backwardsNavigationType: BackwardsNaviagtionInitiatorType,
+  ) => Promise<void>;
   node: Store.node | null;
 }>({
   createAnalytics: () => Promise.resolve(),
   trackHelpClick: () => Promise.resolve(),
   trackNextStepsLinkClick: () => Promise.resolve(),
   trackFlowDirectionChange: () => Promise.resolve(),
+  trackBackwardsNavigationByNodeId: () => Promise.resolve(),
   node: null,
 });
 const { Provider } = analyticsContext;
@@ -64,6 +72,7 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     resultData,
     previewEnvironment,
     flowId,
+    flow,
   ] = useStore((state) => [
     state.currentCard,
     state.breadcrumbs,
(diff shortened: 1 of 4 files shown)


Mike-Heneghan:
[Track clicks to "Back" & "Change" on each component](https://trello.com/c/szauqcNY/2565-track-clicks-to-back-change-on-each-component)

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -296,6 +307,39 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
...
   }
 
+  async function trackBackwardsNavigationByNodeId(
+    nodeId: string,
+    initiator: BackwardsNaviagtionInitiatorType,
+  ) {
+    const targetNodeMetadata = getTitleAndTypeFromFlow(nodeId);
+    console.log("Target node info: ", targetNodeMetadata);
+    const metadata: Record<string, NodeMetadata> = {};
+    metadata[`${initiator}`] = targetNodeMetadata;
```
Mike-Heneghan, line 317: I opted for this syntax as I was getting errors that the `initiator` wasn't being used when I was trying to apply is a key in the object creation.

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -21,6 +21,7 @@ type AnalyticsLogDirection =
 
 export type HelpClickMetadata = Record<string, string>;
 export type SelectedUrlsMetadata = Record<"selectedUrls", string[]>;
+export type BackwardsNaviagtionInitiatorType = "change" | "back";
```
Mike-Heneghan, line 24: I thought this might be an extensible option for capturing very similar user behaviour whilst maintaining what initiated it. I could see this expanding to include returning to a section via a section header link.

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -350,6 +394,15 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
         return {};
     }
   }
+
+  function getTitleAndTypeFromFlow(nodeId: string) {
```
Mike-Heneghan, line 398: I thought it made sense to look up the flow data in the analytics provider to reduce duplication of this step in the places where the tracking is called. If I'm missing the nuances of interacting with the store state please let me know.

Mike-Heneghan:
[Track clicks to "Back" & "Change" on each component](https://trello.com/c/szauqcNY/2565-track-clicks-to-back-change-on-each-component)

Mike-Heneghan:
[Track clicks to "Back" & "Change" on each component](https://trello.com/c/szauqcNY/2565-track-clicks-to-back-change-on-each-component)

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -296,6 +307,39 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     }
   }
 
+  async function trackBackwardsNavigationByNodeId(
+    nodeId: string,
+    initiator: BackwardsNaviagtionInitiatorType,
+  ) {
+    const targetNodeMetadata = getTitleAndTypeFromFlow(nodeId);
+    console.log("Target node info: ", targetNodeMetadata);
+    const metadata: Record<string, NodeMetadata> = {};
```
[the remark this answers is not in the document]
jessicamcinchak: This seems like the right call to me! Since you can only click "change" once, and always have to step forward again one by one after changing (and could theoretically drop off before you're back at Review/Result), individual records make sense here to me :+1:

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
jessicamcinchak: Makes sense, nice to have an enum for this! Another possible future extension in addition to section would be "override" I think - eg via PropertyInformation component, and maybe other node types in future

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -296,6 +307,39 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     }
   }
 
+  async function trackBackwardsNavigationByNodeId(
+    nodeId: string,
+    initiator: BackwardsNaviagtionInitiatorType,
+  ) {
+    const targetNodeMetadata = getTitleAndTypeFromFlow(nodeId);
+    console.log("Target node info: ", targetNodeMetadata);
```
jessicamcinchak, line 315: nit: looks like this `console.log` might be a remnant of local dev and can be dropped before merging? one more below on L318 too

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
jessicamcinchak: I like this approach! Technically possible to do using the store from each component it's called at, but nice to reduce component complexity and shouldn't interfere with applicant-facing response times/performance when done here since analytics requests route through Beacon :+1:

jessicamcinchak (approved):
Apologies it's taken me awhile to properly get to this one, all working as expected on pizza for me and tracking "back" and "change" from both Result & Review :raised_hands:

Feels like a simple structure to extend in the future like you mention for section navigation and "overrides" :100:

PR #2378, opened 2023-11-04
fix: Correctly parse `amount` as number

DafyddLlyr:
## What's the problem?
Regression tests are failing as an invalid body is being sent to GovPay to create a payment. The following error is returned from the GovPay sandbox API - 

```json
{
  "field": "amount",
  "code": "P0102",
  "description": "Invalid attribute value: amount. Must be a valid numeric format"
}
```

## What's causing this?
The bug was introduced [here](https://github.com/theopensystemslab/planx-new/blob/6e7b88f48d279b14a47021ad4cfbb21a171bf1ab/api.planx.uk/inviteToPay/paymentRequest.ts#L71) in this PR - https://github.com/theopensystemslab/planx-new/pull/2356.

`req.params` must be a string, so the amount was cast to as string but never back to a number.

This is only on staging currently so will not be affecting customers.

I did locally run E2E regression tests on this branch but must not have rebuilt the API after these changes were made.

## Next steps
This is fine and fixes the issue now, a few things I'd love to implement though would be - 
- GovPay types (maybe namespaced?)
- Convert pay to modules to introduce validation between middleware layers so we'd catch things like this.

### DIFF
--- a/api.planx.uk/inviteToPay/paymentRequest.ts
+++ b/api.planx.uk/inviteToPay/paymentRequest.ts
@@ -8,6 +8,14 @@ import {
 import { GovUKPayment } from "@opensystemslab/planx-core/types";
 import { $api } from "../client";
 
+// https://docs.payments.service.gov.uk/api_reference/create_a_payment_reference/#json-body-parameters-for-39-create-a-payment-39
+interface GovPayCreatePayment {
+  amount: number;
+  reference: string;
+  description: string;
+  return_url: string;
+}
+
 interface GetPaymentRequestDetails {
   paymentRequest: {
     sessionId: string;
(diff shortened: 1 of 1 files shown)


api.planx.uk/inviteToPay/paymentRequest.ts
```diff
@@ -83,17 +91,30 @@ export async function buildPaymentPayload(
...
         status: 400,
       }),
     );
   }
-  req.body = {
-    amount: req.params.paymentAmount,
-    reference: req.query.sessionId,
+
+  const createPaymentBody: GovPayCreatePayment = {
+    amount: parseInt(req.params.paymentAmount),
```
jessicamcinchak, line 110: can you remind me how `req.params.paymentAmount` relates to `payment_requests.payment_amount` here - are they the both formatted as pence? if so, does `parseInt()` actually work as expected here (eg `20600` to £206)? 

sorry if this is something obvious i'm missing!

DafyddLlyr: No worries! I'm actually really keen to come back to all this and do some typing / documentation to tidy up here as I've found the data flow of the PaymentRequest middlewares a little hard to follow at times.

GovPay always handles payment amount in pence, and this is also what we store in the DB (as a number).

The `fetchPaymentRequestDetails()` middleware gets these details from the db when the API endpoints are called. It gets `payment_requests.payment_amount` as a number (in pence). To pass this onto the next middleware layer it was adding this as a query param. As it was untyped TS wasn't flagging it as a problem that we were going -

`payment_requests.payment_amount` → `req.params.paymentAmount` → GovPay API `payment_amount`

Which should be - 

number → string → number

So now we're just converting the types correctly.

Honestly a better solution here (but a wider refactor) would be to use something like `res.locals` docs (https://expressjs.com/en/5x/api.html#res.locals) to pass these values between middleware layers.

If it seems worthwhile (and makes things easier to follow) I'd be happy to pick this up just now tbh!

jessicamcinchak: Thanks for clarification - prod `payment_requests` table has an outdated mix of `103` and `20600` from early testing (and no "real" requests since) which always confuses me a bit!

jessicamcinchak (approved):

api.planx.uk/inviteToPay/paymentRequest.ts
DafyddLlyr: Yeah I think this was a bug which we had in place during development of the feature which is now resolved, there's a PR somewhere where I added this comment on the column - 

image

PR #2387, opened 2023-11-06
Mh/analytics link to analytics in editor

Mike-Heneghan:


Mike-Heneghan:
[Add a link to each service's analytics page in the editor](https://trello.com/c/aFef5oDo/2636-add-a-link-to-each-services-analytics-page-in-the-editor)

editor.planx.uk/src/pages/FlowEditor/lib/store/shared.ts
```diff
@@ -13,6 +13,7 @@ export interface SharedStore extends Store.Store {
   flow: Store.flow;
   flowSlug: string;
   flowName: string;
+  flowMetabaseLink: string;
```
DafyddLlyr, line 16: As this is a nullable value (for flows where it's not set) it might be best to type it as `string | null` and set the initial value to `null`.

image

editor.planx.uk/src/routes/flow.tsx
```diff
@@ -204,8 +209,12 @@ const routes = compose(
 
   withView(async (req) => {
     const [flow, ...breadcrumbs] = req.params.flow.split(",");
-    const settings: FlowSettings = await getFlowSettings(flow, req.params.team);
+    const { settings, flowMetabaseLink } = await getFlowMetadata(
+      flow,
+      req.params.team,
+    );
     useStore.getState().setFlowSettings(settings);
+    useStore.setState({ flowMetabaseLink });
```
DafyddLlyr, line 217: You might be able to simplify things here and set both `flowSettings` and `flowMetabaseLink` in the same setter.

Something like `useStore.setState({ flowMetabaseLink, flowSettings });`

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -347,6 +347,7 @@
           - creator_id
           - data
           - id
+          - metabase_link
```
DafyddLlyr, line 350: This indicates that only the `platformAdmin` role can access this column, when I think what we really want is for all roles used by authorised users (i.e. all but `public`) to be able to access this.

DafyddLlyr (changes requested):

editor.planx.uk/src/pages/FlowEditor/components/PreviewBrowser.tsx
```diff
@@ -163,6 +171,19 @@ const PreviewBrowser: React.FC<{
             </Link>
           </Tooltip>
 
+          {!!flowMetabaseLink && (
+            <Tooltip arrow title="Open analytics page">
```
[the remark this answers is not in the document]
jessicamcinchak: nit: I personally think the unpublished & published icons should stay next to each other at the far right since these are closely related actions. Could the analytics icon shift one left in position on this row, so it's directly to the right of "Toggle debug console" instead ?

hasura.planx.uk/migrations/1699010412430_alter_table_public_flows_add_column_metabase_link/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."flows" add column "metabase_link" text
+ null;
```
jessicamcinchak, line 2: nit: could this column name be generalised as `analytics_link`? Just thinking: since it's an external link, there's no reason councils couldn't set their own non-metabase thing here in the future? And while we use metabase now, we know the wider ODP metrics & analytics team is also exploring other platforms.

Mike-Heneghan: Absolutely, that's a great point @jessicamcinchak!

editor.planx.uk/src/routes/flow.tsx
Mike-Heneghan: I'll check this out 👍

editor.planx.uk/src/pages/FlowEditor/components/PreviewBrowser.tsx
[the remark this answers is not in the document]
ianjon3s: Agree with Jess' point about maintaining the grouping of the two right-side related icons and placing the analytics one in the middle.

However I think there is a value in maintaining all of the icons in place, and displaying more of an active/inactive state, rather than showing/hiding the icon.

Something like this (example has the inactive icon at 40% opacity):
image

The logic being that is is easier/quicker to decipher whether an element is active/inactive through contrast than to work out if it is present or not.

hasura.planx.uk/migrations/1699010412430_alter_table_public_flows_add_column_metabase_link/up.sql
jessicamcinchak: I think the followup task of allowing editors to manage this link via service settings is already be captured by [this roadmap ticket](https://miro.com/app/board/uXjVPrebzwQ=/?moveToWidget=3458764548469369468&cot=14 ) :+1:  Since a future phase, maybe no need to make a Trello ticket yet?

editor.planx.uk/src/pages/FlowEditor/lib/store/shared.ts
Mike-Heneghan: Fixed here: 6ecabd07f9a64a721b55e274c44c714f8820c53c

editor.planx.uk/src/routes/flow.tsx
Mike-Heneghan: Fixed here: 7d604531d87fb0a147cedb199731daac885b8a40

I suspect that `.setFlowSettings` can probably be removed and associated files updated to use the altered syntax. 

I might have a go adding that as a commit once I've updated as per the other comments.

editor.planx.uk/src/pages/FlowEditor/components/PreviewBrowser.tsx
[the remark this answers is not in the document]
Mike-Heneghan: That's a great shout @ianjon3s I'll get this updated.

hasura.planx.uk/migrations/1699010412430_alter_table_public_flows_add_column_metabase_link/up.sql
Mike-Heneghan: Thanks for sharing that @jessicamcinchak, I had a look on Trello but I didn't think to check the roadmap.

hasura.planx.uk/metadata/tables.yaml
Mike-Heneghan: Fixed here: 1d5c3043c60aa3f63327967d5e677ec20ad46a13

editor.planx.uk/src/routes/flow.tsx
DafyddLlyr: Yep - sounds good - any chance to rationalise and remove code!

editor.planx.uk/src/pages/FlowEditor/components/PreviewBrowser.tsx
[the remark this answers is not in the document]
Mike-Heneghan: I've got two ways of adding this but I'm a little unsure which option has the best balance for accessibility etc. 🤔 

I think it would make sense to have the icon as a `Link` although there's not really a concept of a link you can't click? Although there's is a concept of a `disabled` button. 

I was thinking of two ways of potentially indicating that the analytics page is unavailable:

Option 1:

Show the icon with a reduced opacity 

```JavaScript
<Tooltip arrow title="Analytics page unavailable">
  <BarChart style={{ opacity: 0.4 }} />
</Tooltip>
```

Option 2:

Return the material UI `Link` as a disabled button.

```JavaScript
<Tooltip arrow title="Analytics page unavailable">
  <Link
    component={"button"}
    disabled
  >
     <BarChart />
  </Link>
 
</Tooltip>
```

I think Option 2 is maybe better as a disabled button has a consistent style and there's precedent for it not having any action when clicked on? Whereas Option 1 is a random icon which doesn't convey much to those using screen readers etc.?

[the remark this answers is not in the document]
Mike-Heneghan: I've got `Option 2` on this commit: b4b7e6602f79437443732c785ade50113ea44dfb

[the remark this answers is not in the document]
ianjon3s: I think the second option with `aria-disabled="true"` in place of the native `disabled` state would be ideal, in that it would allow for focus on the element, and therefore readability of the label for a screen reader.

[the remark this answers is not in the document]
Mike-Heneghan: I tried swapping out `disabled` for `aria-disabled="true"` directly but it changes the styling of the icon to being blue rather than opaque. Although adding it to the component I think maybe has the best of both worlds? 

```JavaScript
        <Tooltip arrow title="Analytics page unavailable">
              <Link component={"button"} disabled aria-disabled={true}>
                <BarChart />
              </Link>
            </Tooltip>
```

hasura.planx.uk/migrations/1699010412430_alter_table_public_flows_add_column_metabase_link/up.sql
Mike-Heneghan: Fixed here: b4e3f1de31c9aa578bd4322f8c61083d797a13c4

editor.planx.uk/src/routes/flow.tsx
Mike-Heneghan: If you don't mind @DafyddLlyr I'm going to do this in another branch to avoid muddying things in this PR

editor.planx.uk/src/pages/FlowEditor/components/PreviewBrowser.tsx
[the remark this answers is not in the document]
ianjon3s: Agreed @Mike-Heneghan , worth adding in addition 👍

As a more general comment, the editor side of the product has neither been built or scrutinised at the same level as the public interface from an accessibility perspective, so I think a wider review is in order (and of course keeping it in the forefront when rethinking/design the interface or adding further features).

hasura.planx.uk/migrations/1699010412430_alter_table_public_flows_add_column_metabase_link/up.sql
jessicamcinchak: nit: since this migration hasn't run against `main` yet, you can actually simplify this into a single migration with the correct name and skip the `alter... rename...` ! 

It's a bit annoying mid-PR because you'll have to destroy/re-create Vultr for a fresh DB for it to run against, but it makes for a tidier version history

jessicamcinchak (approved):
Looks good and all working as expected for me! I successfully added a (fake) `analytics_link` to [this flow](https://2387.planx.pizza/opensystemslab/flood-risk-assesment-template-keep-and-improve).

One minor comment about migration history, but not a particularly big deal to skip if you're ready to get this merged.

hasura.planx.uk/migrations/1699010412430_alter_table_public_flows_add_column_metabase_link/up.sql
Mike-Heneghan: Thanks for the comment @jessicamcinchak I should have added a note on this! 

I'd tried squashing the migration but it still started with the incorrect column name so it didn't really act as the tidy up I was hoping for. 

My plan is to do an interactive rebase to clean up the history to start off with `analytics_link` although when I'm reviewing I sometimes find it hard to see if feedback has been taken on if the commit history changes around. While I was waiting I thought I'd incrementally add commits as fixes then do a tidy up once they were approved.

Do you have any recommendations for better work flows to handle this sort of stuff?

jessicamcinchak: When I'm dealing with Hasura migrations on branches, I'd actually skip a Git workflow and just remove/edit the mgiration files directly. Eg, you can delete these two folders and their respective up/down sqls, create a new up migration to add the column name afresh and just make sure you've destroyed Vultr before you push your new commit. When merged, staging hasura won't ever know these first two iterations existed! 

Does that make sense?

jessicamcinchak: Hasura CLI also has some built-in commands for squashing migrations you could try: https://hasura.io/docs/latest/hasura-cli/commands/hasura_migrate_squash/

Mike-Heneghan: That does make sense! Thanks @jessicamcinchak 😁

Mike-Heneghan: I did have a go with squashing but I maybe didn't get the correct type of squash as it just combined the two i.e. start with the wrong name then rename in one `up.sql`.

jessicamcinchak: Yeah the Hasura CLI will always aim to preserve all commands I think, whereas in this case that's not actually necessary because none of these commands are on main yet!

Mike-Heneghan: I've just pushed up a commit which removes the old migrations and adds the new ones. 

If I've understood correctly, as I deleted the Vultr instance when the CI runs it'll create a new Pizza environments with the files that are there now i.e. the old migrations in the commit history will have no impact. 

The new pizza environment will only know the files as they appear in 4ed65920ad1abee9d12f9bf9d467fccf1d160e06

When it comes to merging everything gets squashed anyway so there will effectively be no record of the `metabase_link`?

Mike-Heneghan: Thanks for the guidance I was really keen to tidy up 🥳 I'm still getting my head around how to work with Hasura and the general vibe is it's too easy after working with Django/Rails migrations 😂

Mike-Heneghan:
> ## Pizza
> Deployed [d5e56ac](https://github.com/theopensystemslab/planx-new/commit/d5e56aceb025340630d977d532ca78383e7a7d60) to https://2387.planx.pizza.
> 
> Useful links:
> 
> * [Editor](https://2387.planx.pizza)
> * [Storybook](https://storybook.2387.planx.pizza)
> * [Hasura](https://hasura.2387.planx.pizza)
> * [API](https://api.2387.planx.pizza/docs)
> * [ShareDB](https://sharedb.2387.planx.pizza)

All seems good on this Pizza testing with this flow: https://2387.planx.pizza/testing/mikes-analytics-testing

PR #2392, opened 2023-11-07
fix: update audit table API permissions so unpublished routes can render PlanningConstraints

jessicamcinchak:
August flagged this morning that on `/unpublished` routes, PlanningConstraints are failing to render correctly & throwing the following console error:
```json
{
 "extensions": {
    "path": "$.selectionSet.insert_planning_constraints_requests_one",
    "code": "validation-failed"
  },
  "message": "field "insert_planning_constraints_requests_one" not found in type: 'mutation_root'"
}
```

Before: https://editor.planx.uk/barnet/find-out-if-you-need-planning-permission/unpublished
After: https://2392.planx.pizza/barnet/find-out-if-you-need-planning-permission/unpublished

**Changes**:
- Update `planning_constraints_requests` table to grant the `api` role select permissions, consistent with other audit tables, as this is necessary for the `insert_` mutation to return the new record's `id`
- Adjust frontend conditional rendering so in cases of an error, that's all we ever display - never a mix of error plus successful roads response like happened in this edge case

This bug is only effecting `/unpublished` routes, not `/preview`, so no particular rush to deploy

### DIFF
--- a/editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
@@ -263,7 +263,7 @@ export function PlanningConstraintsContent(
           refreshConstraints={refreshConstraints}
         />
       )}
-      {positiveConstraints.length > 0 && (
+      {!showError && positiveConstraints.length > 0 && (
         <>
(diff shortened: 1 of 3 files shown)


DafyddLlyr (approved):

DafyddLlyr:

PR #2394, opened 2023-11-07
Revert "chore: Add St.Albans certs (take 2)"

DafyddLlyr:
Pre-emptive revert juuuuust in case it fails again - CI should have run allowing us a quicker turnaround.

All being well this PR will just get closed if the current prod deploy succeeds.


DafyddLlyr:
Update: The prod deploy failed with the same Pulumi error 

```
InvalidViewerCertificate: The certificate that is attached to your distribution was not issued by a trusted Certificate Authority. For more details, see: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/CNAMEs.html#alternate-domain-names-requirements
```

jessicamcinchak (approved):

PR #2396, opened 2023-11-07
refactor: update the pattern for settings the flowSettings in state

Mike-Heneghan:
## What 

- Rationalise the syntax for managing the `flowSettings` in state
- Remove unnecessary custom setter

## Why

- Improves readability/maintainability

### DIFF
--- a/editor.planx.uk/src/routes/views/standalone.tsx
+++ b/editor.planx.uk/src/routes/views/standalone.tsx
@@ -10,7 +10,7 @@ import { Flow, GlobalSettings } from "types";
 import { getTeamFromDomain } from "../utils";
 
 interface StandaloneViewData {
-  flows: Pick<Flow, "team" | "settings">[];
+  flows: Pick<Flow, "team" | "flowSettings">[];
   globalSettings: GlobalSettings[];
 }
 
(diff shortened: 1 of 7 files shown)


editor.planx.uk/src/routes/views/standalone.tsx
```diff
@@ -25,16 +25,19 @@ const standaloneView = async (req: NaviRequest) => {
...
 
+  useStore.setState({ flowSettings });
+
   const state = useStore.getState();
   state.setFlowNameFromSlug(flowSlug);
   state.setGlobalSettings(globalSettings[0]);
-  state.setFlowSettings(flowSettings);
   state.setTeam(team);
 
+  console.log('Standalone state: ', state)
```
Mike-Heneghan, line 39: I've been trying to test the refactoring locally by logging out the state to ensure that the state is being set successfully in the relevant components. 

Although I'm struggling to figure out how to setup a `standalone` view 🤔 

I believe an example would be an invite to pay and I maybe just need to persevere to setup that up locally?

DafyddLlyr: This might add some context here - https://trello.com/c/0Uw8ViZJ/2324-make-editor-sidebar-preview-behave-same-as-the-unpublished-preview-increase-granularity-of-previewenvironment

It's a bit of a pain point (at least for me!) and something I'd love to have a crack at refactoring / removing / rationalising.

This is referring to the `PreviewEnvironment` variable in `shared.ts` (the SharedStore slice) - `standalone` is essentially all pages that are not the Editor 👌

Mike-Heneghan: Thanks for the info on this @DafyddLlyr!

I'm still a bit confused because based on this I thought I'd see the output from `console.log('Standalone state: ', state)` when I was on a `/preview` route? 

Although that seems to use the `published` view rather than the `standalone` view?

PR #2382, opened 2023-11-06
chore: Add St.Albans subdomain to frontend

DafyddLlyr:
Draft until St.Albans are pointing to our CDN

PR #2397, opened 2023-11-08
chore: St. Albans key, chain, and cert

DafyddLlyr:
Take 3...! 🤞 

St. Albans have now provided a chain certificate which should resolve the previous issues. Again this has been tested on AWS staging without issues.


infrastructure/application/index.ts
```diff
@@ -52,6 +52,10 @@ const CUSTOM_DOMAINS =
           domain: "planningservices.newcastle.gov.uk",
           name: "newcastle",
         },
+        {
+          domain: "stalbans.newcastle.gov.uk",
```
Mike-Heneghan, line 56: Is this right or a copy and paste error?

DafyddLlyr: 🤦

DafyddLlyr: Thank you very much! Will fix now

Mike-Heneghan (changes requested):
I think there's maybe a wee copy/paste error for the domain for St Albans.

DafyddLlyr:
Thanks very much for catching that @Mike-Heneghan - clearly time for a walk / coffee / both 😅 Fixed!

Mike-Heneghan (approved):
Looks good to me 👍 

Non-blocking but in a previous PR I think you included an [Airbrake config change](https://github.com/theopensystemslab/planx-new/pull/2390/files) but I don't think it blocks this.

DafyddLlyr:
That's correct - it's intentionally left out here. That change is already queued up here once the CDN is generated https://github.com/theopensystemslab/planx-new/pull/2382/files

PR #2398, opened 2023-11-08
docs: Update subdomain docs and `.gitignore`

DafyddLlyr:


.gitignore
```diff
@@ -20,3 +20,10 @@ hasura.planx.uk/.env.test
 /playwright-report/
 /playwright/.cache/
 api.planx.uk/tmp/
+
+# Ignore certificate files
```
DafyddLlyr, line 24: This is just due to my paranoia when working with these files, I'm always triple checking I'm not adding them to git.

This should prove a handy guard to avoid any major issues as we generally handle *.councilName.gov.uk certificates.

Mike-Heneghan (approved):
:shipit:

PR #2402, opened 2023-11-08
chore: Setup Camden frontend

DafyddLlyr:
Awaiting Camden CNAME record

### DIFF
--- a/hasura.planx.uk/migrations/1699449006468_setup_camden_domain/up.sql
+++ b/hasura.planx.uk/migrations/1699449006468_setup_camden_domain/up.sql
@@ -1,0 +1,3 @@
+UPDATE teams
+SET domain = 'planningservices.camden.gov.uk'
+WHERE slug = 'camden';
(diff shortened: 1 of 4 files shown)


hasura.planx.uk/migrations/1699449006468_setup_camden_domain/up.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE teams
+SET domain = 'planningservices.camden.gov.uk'
+WHERE slug = 'camden';
```
jessicamcinchak, line 3: Are there any data sync side effects of handling this via migration? Eg it means it will get set locally & on staging, whereas I think `domain` column is omitted from data sync script?

DafyddLlyr: Very good point thanks! Migration dropped from PR in favour of doing this manually on prod only.

jessicamcinchak (approved):

PR #2416, opened 2023-11-13
fix: show warning and skip Send event creation if `DISABLE_SAVE_AND_RETURN` feature flag is enabled

jessicamcinchak:
Addresses this error: https://opensystemslab.slack.com/archives/C01E3AC0C03/p1699868054247069

Testing:
- With no feature flags toggled, a published flow will create "Send" events as expected. There is no warning banner.
- With the `DISABLE_SAVE_AND_RETURN` feature flag toggled, you can still proceed through Send to the Confirmation page, but no events have been created in Hasura. The warning banner is also visible.

### DIFF
--- a/editor.planx.uk/src/components/FeatureFlagBanner.tsx
+++ b/editor.planx.uk/src/components/FeatureFlagBanner.tsx
@@ -1,0 +1,54 @@
+import ReportIcon from "@mui/icons-material/Report";
+import Box from "@mui/material/Box";
+import Button from "@mui/material/Button";
+import Container from "@mui/material/Container";
+import { styled } from "@mui/material/styles";
+import Typography from "@mui/material/Typography";
+import { hasFeatureFlag } from "lib/featureFlags";
+import React, { useState } from "react";
+
+const TestEnvironmentWarning = styled(Box)(({ theme }) => ({
+  display: "flex",
+  backgroundColor: theme.palette.background.paper,
+  color: theme.palette.text.primary,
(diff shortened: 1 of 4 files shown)


scripts/seed-database/write/flows.sql
```diff
@@ -46,7 +46,7 @@ SET
   version = EXCLUDED.version,
   settings = EXCLUDED.settings,
   copied_from = EXCLUDED.copied_from,
-  analytics_link = null;
+  analytics_link = NULL;
```
jessicamcinchak, line 49: Sorry for bundling, but this is a followup syntax fix to #2414

editor.planx.uk/src/components/FeatureFlagBanner.tsx
```diff
@@ -0,0 +1,54 @@
...
+              Hide
+            </Button>
+          </Container>
+        </TestEnvironmentWarning>
+      )}
+    </>
+  );
+};
+
+export default FeatureFlagBanner;
```
jessicamcinchak, line 54: This displays like this when condition is met:
Screenshot from 2023-11-13 15-40-01

Open to language suggestions here!

scripts/seed-database/write/flows.sql
jessicamcinchak: Confirmed this works here https://github.com/theopensystemslab/planx-new/actions/runs/6852179815

DafyddLlyr (approved):
Perfect! 👌

PR #2417, opened 2023-11-13
fix: bug with file requirements not changing on delete of file

Mike-Heneghan:
## What:

- Update the `removeFile` function to set the updated file list to state.

## Why:

- This was written when the state was being updated directly and was therefore broken when the fileList was deep cloned.

## Screen recording:

Bug:

https://github.com/theopensystemslab/planx-new/assets/36415632/1d8f28c8-accb-458e-94c8-361ff1b1f9d7

Post fix:

https://github.com/theopensystemslab/planx-new/assets/36415632/a5535588-7c6b-41b4-b3c9-8db7188068fc

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
@@ -240,11 +240,12 @@ function Component(props: Props) {
                       ),
                     );
                     setFileUploadStatus(`${slot.file.path} was deleted`);
-                    removeSlots(
+                    const updatedFileList = removeSlots(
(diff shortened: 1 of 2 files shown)


Mike-Heneghan:
[Error message in Upload & Label that prevents users from saving and moving to the next page (other similar bugs previously seen)](https://trello.com/c/Kc3ifwND/2632-error-message-in-upload-label-that-prevents-users-from-saving-and-moving-to-the-next-page-other-similar-bugs-previously-seen)

jessicamcinchak (approved):
Thanks for catching this! Would be great to have a test case to cover this scenario going forward, but happy for fix to go in first - your judgement / decision based on time to write test!

PR #2418, opened 2023-11-14
feat: File API module

DafyddLlyr:
## What does this PR do?
 - Group file related API endpoints into a module
 - Document endpoints in Swagger
 - Rename some routes for a more logical structure
 - Add validation and types
 - Adds auth to private file upload endpoint
 - `file_type` → `fileType`

image


api.planx.uk/modules/file/file.test.ts
```diff
@@ -41,39 +42,55 @@ describe("File upload", () => {
...
+    });
 
     it("should not upload without filename", async () => {
       await supertest(app)
         .post(ENDPOINT)
+        .set(auth)
         .field("filename", "")
         .attach("file", Buffer.from("some data"), "some_file.txt")
-        .expect(422)
+        .expect(400)
```
DafyddLlyr, line 65: One issue with the current simplified Zod error handling is that the validation checker in `api.planx.uk/shared/middleware/validate.ts` just returns a 400 so we lose some nuance. 

I'll open a GH issue describing this small improvement as well as nicer error formatting using something like https://github.com/causaly/zod-validation-error maybe.

api.planx.uk/modules/file/file.test.ts
```diff
@@ -41,39 +42,55 @@ describe("File upload", () => {
...
         .then((res) => {
           expect(mockPutObject).not.toHaveBeenCalled();
-          expect(res.body.error).toBe("missing filename");
+          expect(res.body).toHaveProperty("issues");
+          expect(res.body).toHaveProperty("name", "ZodError");
```
DafyddLlyr, line 69: See above - I'd like there to be more meaningful errors returned here. It does show that the error is the missing `filename` field but it's not in a super friendly format.

api.planx.uk/modules/file/service/getFile.ts
```diff
@@ -11,7 +11,7 @@ export const getFileFromS3 = async (fileId: string) => {
   const file = await s3.getObject(params).promise();
 
   return {
-    body: file.Body,
+    body: file.Body as Buffer,
```
DafyddLlyr, line 14: This should be `S3.Body` which is just `Buffer | Uint8Array | Blob | string | Readable` but this was causing Vultr to run out of heap memory when compiling with `tsc`. It was going from 350mb to over 1gb just for this change - presumably it was importing and compiling a ton of additional AWS types?

jessicamcinchak (approved):
Changes make this really clean & consistent, thanks!

PR #2419, opened 2023-11-14
feat: Flows API module

DafyddLlyr:
## What does this PR do?
 - Documents the flows module (move, copy, validate, publish etc)
 - Splits code into controllers (handle HTTP requests) and services (business logic)
 - Adds validation to the `/flows` endpoints
 - Standardise a few response types
 - Splits up validate and publish functionality into their own files
 - No functional changes!

image


api.planx.uk/modules/flows/copyFlow/controller.ts
```diff
@@ -0,0 +1,54 @@
...
+  replaceValue: string;
+  data: Flow["data"];
+}
+
+export const copyFlowSchema = z.object({
+  params: z.object({
+    flowId: z.string(),
+  }),
+  body: z.object({
+    replaceValue: z.string().length(5),
```
DafyddLlyr, line 19: Strictly controlling the length here to 5 matches the current frontend behaviour, and has meant that a few test cases need nodeIds tweaked below.

I'll double check the onboarding scripts to enforce this to 5 characters there as well once this is merged.

jessicamcinchak (approved):
Looks good & all working as expected on pizza, thanks again for working through all these !

DafyddLlyr:
Thanks for review - kept in draft as I hit a small issue with the download-schema endpoint. Looking at it again now and will push a fix to this branch shortly.

PR #2427, opened 2023-11-15
fix: Apply auth to public upload, not private

DafyddLlyr:
## What's the problem?
Reported here on ODP Slack - https://opendigitalplanning.slack.com/archives/C0241GWFG4B/p1700064607120349

Uploading files is not working from applicant facing flows.

## What's the solution?
The route guard `useTeamEditorAuth()` was incorrectly applied to the "public" and not the "private" - anybody can upload to private, only Editors can upload to public 🙃 

Maybe another sign that some E2E file upload tests may be a good idea to consider soon?

### DIFF
--- a/api.planx.uk/modules/file/file.test.ts
+++ b/api.planx.uk/modules/file/file.test.ts
@@ -43,6 +43,68 @@ describe("File upload", () => {
 
   describe("Private", () => {
     const ENDPOINT = "/file/private/upload";
+
+    it("should not upload without filename", async () => {
+      await supertest(app)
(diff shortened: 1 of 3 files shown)


jessicamcinchak (approved):
Counterintuitive! Sorry I didn't catch this one in review yesterday, thanks for quick fix!

DafyddLlyr:
> Counterintuitive! Sorry I didn't catch this one in review yesterday, thanks for quick fix!

I know right - no apologies needed - a quirky language thing where public upload is private, and private is public 🙃

PR #2428, opened 2023-11-15
refactor: update the duplication step info

Mike-Heneghan:
- Suggest only duplicating the dashboard not the cards
- This allows us to manage the cards across all dashboards and update once.


DafyddLlyr (approved):
Thanks for picking this up!

PR #2429, opened 2023-11-15
feat: enable transparent or custom background for content

ianjon3s:
## What does this PR do?

The current version of the Content component has the option for a custom background colour but by default is published with a white (transparent) background.

Horizontal and vertical padding is added to the content container by default to compensate for a background colour, meaning that the default behaviour is to push the content out of alignment with the parent container.

The PR introduces options for transparent and custom containers.

https://2429.planx.pizza/testing/content-colour-test/preview

### Preview
Before (above) & after (below):

image

### DIFF
--- a/editor.planx.uk/src/@planx/components/Content/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/Content/Public.tsx
@@ -13,8 +13,6 @@ export type Props = PublicProps<Content>;
 const Content = styled(Box, {
   shouldForwardProp: (prop) => prop !== "color",
 })<{ color?: string }>(({ theme, color }) => ({
-  padding: theme.spacing(2),
-  backgroundColor: color,
   color:
     mostReadable(color || "#fff", [
       "#fff",
@@ -25,10 +23,15 @@ const Content = styled(Box, {
   },
 }));
 
-const ContentComponent: React.FC<Props> = (props) => {
+const ContentComponent: React.FC<Props> = (props, theme) => {
   return (
     <Card handleSubmit={props.handleSubmit} isValid>
-      <Content color={props.color} data-testid="content">
+      <Content
+        color={props.customBackground === true ? props.color : theme.primary}
+        bgcolor={props.customBackground === true ? props.color : "transparent"}
+        data-testid="content"
+        p={props.customBackground === true ? 2 : 0}
+      >
         <ReactMarkdownOrHtml
           source={props.content}
           openLinksOnNewTab
--- a/editor.planx.uk/src/@planx/components/Content/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/Content/Public.tsx
@@ -13,6 +13,7 @@ export type Props = PublicProps<Content>;
 const Content = styled(Box, {
   shouldForwardProp: (prop) => prop !== "color",
 })<{ color?: string }>(({ theme, color }) => ({
+  backgroundColor: color,
   color:
     mostReadable(color || "#fff", [
       "#fff",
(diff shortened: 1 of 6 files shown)


editor.planx.uk/src/@planx/components/Content/model.ts
```diff
@@ -1,6 +1,7 @@
 import { MoreInformation, parseMoreInformation } from "../shared";
 
 export interface Content extends MoreInformation {
+  customBackground: boolean;
```
DafyddLlyr, line 4: Is there any way we could derive this information from the props already passed in as opposed to adding a new prop?

Is `customBackground` always the same as `color !== "#FFF"` or whatever the default that's set is?

DafyddLlyr (commented):
One small question which you might already have looked into on this one. Also happy to be another pair of eyes here if required!

I suspect that we can check if a color is set (or if it's set back to white) to get the same data point.

Either way - this is a great one to have picked up and fixed thank you 👏

editor.planx.uk/src/@planx/components/Content/model.ts
jessicamcinchak: Need the pizza to build to confirm this - but _I think_ if we introduce a new required prop like this, all existing Content components will break / need to be updated which is less an than ideal! Another reason to: try to derive from existing props, or at minimum introduce this as an optional prop!

ianjon3s: Good shout and I appreciate the logic introduced here may not be optimal (and I hadn't considered the implications of a new prop breaking existing components).

I'll try to rework so that the padding is introduced only if the background colour is changed from the default to see if that would be a simpler solution.

ianjon3s: @DafyddLlyr @jessicamcinchak I've now updated this to be much simpler, doing as Daf suggested and checking whether or not the background is white rather than introducing a new option.

I am however struggling to get it working as expected, if no background colour is selected the prop is passed through empty, however trying to target this with `undefined` is not working. 

I've also tried adding `#ffffff` (equivalent white output by the colour picker) as a default prop and this is still not working.

**Note:** I've also disabled the alpha/transparency slider in the colour picker as this appears to be non-functional, at least in Chrome and Safari.

DafyddLlyr: Thanks for the update - I'll take a look @ianjon3s - checking for a negative is always hard!

ianjon3s: Here's a pizza with three content components:

1. Content with default background (no background colour selected)
2. Content with white background selected
3. Content with other custom background colour selected

https://2429.planx.pizza/testing/content-colour-test/preview

editor.planx.uk/src/@planx/components/Content/Public.tsx
```diff
@@ -25,10 +24,18 @@ const Content = styled(Box, {
...
+};
+
 const ContentComponent: React.FC<Props> = (props) => {
   return (
     <Card handleSubmit={props.handleSubmit} isValid>
-      <Content color={props.color} data-testid="content">
+      <Content
+        color={props.color}
+        data-testid="content"
+        p={props.color === ("#ffffff" || undefined) ? 0 : 2}
```
DafyddLlyr, line 37:
```suggestion
        p={(props.color === "#ffffff" || !props.color) ? 0 : 2}
```

DafyddLlyr: It's just a little syntax issue here - we need to check `props.color` against both conditions. I've shorthanded here using `!` but the above is equal to - 

```js
props.color === "#ffffff" || props.color === undefined ? 0 : 2
```

ianjon3s: Ah gotcha, I need to revisit the basics and brush up on my syntax.

Thanks for the help with this.

editor.planx.uk/src/@planx/components/Content/model.ts
DafyddLlyr: Pizza looks great - thanks for the link ✅

DafyddLlyr (approved):

PR #2430, opened 2023-11-15
feat: create a `submission_services_summary` database view for analytics

jessicamcinchak:
As discussed in Slack and over dev call! Currently "select" permissions on this view are restricted to the "admin" role, which I think is all Metabase will need?


DafyddLlyr (approved):
Looks good! Yep - I think as Metabase has its own DB user / role it exists outside the Hasura permission model 👍

PR #2434, opened 2023-11-16
feat: enabled St Albans to query planning constraints

jessicamcinchak:
A4s not published yet, so frontend change only & no API metadata changes yet https://www.planning.data.gov.uk/entity/?dataset=article-4-direction-area&geometry_curie=statistical-geography%3AE07000240&entry_date_day=&entry_date_month=&entry_date_year=

### DIFF
--- a/editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
@@ -70,6 +70,7 @@ function Component(props: Props) {
     "medway",
     "newcastle",
     "southwark",
+    "st-albans",
   ];
 
   const digitalLandParams: Record<string, string> = {


jessicamcinchak:
[Hook up Planning Data constraints for St Albans](https://trello.com/c/JVlAfSp5/2701-hook-up-planning-data-constraints-for-st-albans)

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
```diff
@@ -70,6 +70,7 @@ function Component(props: Props) {
     "medway",
     "newcastle",
     "southwark",
+    "st-albans",
```
DafyddLlyr, line 73: I knew we'd hit this GIS meme sooner or later 😅 

image

This has made me realise that I used "stalbans" in infra code - I think if we switch it it might create a new CDN (and then they'll have to update CNAME records). We should aim to fix this when we renew their certs next year 👍 

This is correct though - our team slug is "st-albans". Might be another pointer towards using db flags for these types of values?

DafyddLlyr (approved):

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
jessicamcinchak: Actually I think it's our team slug that is wrong and should be aligned to https://www.stalbans.gov.uk/ ! 

That has fewer consequences of updating immediately I think since service URLs aren't live yet? 

But, yes, db flag for toggling constraints is the way forward here - I'll add a column tomorrow!

DafyddLlyr: Yep I think the slug is fine to update as they aren't advertising the editor.planx.uk links anywhere - we might just need to cross check and update thinks like any links we have in Notion.

PR #2435, opened 2023-11-16
feat: Reposition and style help text button

ianjon3s:
## What does this PR do?

User-testing feedback and analytics have told us that we need to rethink the position and style of the help text button.

**Summary of changes**

- Reposition button from right-side of viewport to within the content flow (maintaining sticky positioning)
- Update and use the 'outline' button variant
- Add dedicated width, height (min 44px in each dimension as WCAG AAA guidance), font size and drop shadow

**Live example**
https://2435.planx.pizza/lambeth/help-button-test/preview

**Visual example**

image

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
@@ -307,7 +307,7 @@ const InteractiveFileListItem = (props: FileListItemProps) => {
           aria-haspopup="dialog"
           size="small"
         >
-          <span>Help</span>
+          <span>Info</span>
         </InfoButton>
       )}
       <MoreInfo open={open} handleClose={() => setOpen(false)}>
--- a/editor.planx.uk/src/@planx/components/Notice/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/Notice/Public.tsx
@@ -91,10 +91,7 @@ const NoticeComponent: React.FC<Props> = (props) => {
           policyRef={props.policyRef}
           howMeasured={props.howMeasured}
         />
-        <Container
-          customColor={props.color}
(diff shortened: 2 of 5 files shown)


editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
```diff
@@ -307,7 +307,7 @@ const InteractiveFileListItem = (props: FileListItemProps) => {
           aria-haspopup="dialog"
           size="small"
         >
-          <span>Help</span>
+          <span>Info</span>
```
Mike-Heneghan, line 310: _Non-blocking Question_

I was really curious about this but I could never find an instances where it says "Info" as I played around in the editor/previews 😅 

If it's not too much of a hassle would you be able to point me at this @ianjon3s? It's a silly request I'm just curious about where this lives as I haven't encountered it and I feel like I'm missing something.

Mike-Heneghan (approved):
I think this looks great and in my opinion it strikes the balance of being very noticeable without getting into spammy pop-up territory 🥳

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
ianjon3s: Hey @Mike-Heneghan ,

The help/info text is displayed on the Upload + Label component for files that have further info text added.

Here's an example:
https://2435.planx.pizza/testing/upload-label-logic-example/preview?analytics=false

There was a decision a while back from user-testing that we should rename this to read "Info", I failed to mention this in the original PR description.

Mike-Heneghan: Amazing, thanks for the explanation and the example link 🙌

PR #2439, opened 2023-11-17
feat: add `teams.reference_code` to store Planning Data organisation

jessicamcinchak:
Followup from https://github.com/theopensystemslab/digital-planning-data-schemas/pull/71

Once merged to production, I'll populate for all existing council-teams based on https://files.planning.data.gov.uk/dataset/local-authority.json?limit=400


DafyddLlyr (approved):
Perfect - thanks for picking this up!

We should also add a step to the onboarding script to populate this.

PR #2444, opened 2023-11-17
feat: Pay and Invite to Pay docs

DafyddLlyr:
## What does this PR do?
 - Adds docs for "pay" and "invite to pay" routes
 - Adds validation layer to all routes
 - Does not refactor all current functionality into "controllers" and "services" - this PR is already too big sorry and in order to get the remaining routes across the line I've prioritised documentation over this refactoring.

I had initially split this into two "Pay" and "Invite to Pay" PRs but as they're quite tightly coupled it proved pretty tricky - the total lines of code changed in two PRs (even with one branching over the other) was significantly higher than a single PR.

Regression tests passing against this branch https://github.com/theopensystemslab/planx-new/actions/runs/6912983715 ✅


api.planx.uk/modules/pay/middleware.ts
```diff
@@ -82,7 +84,14 @@ export async function fetchPaymentRequestDetails(
   next();
 }
 
-// middleware used by /payment-request/:paymentRequest/pay?returnURL=...
+// https://docs.payments.service.gov.uk/api_reference/create_a_payment_reference/#json-body-parameters-for-39-create-a-payment-39
```
jessicamcinchak, line 87: appreciate these comments :+1:

jessicamcinchak (approved):
Prioritising docs over refactor makes sense ahead of audit, thanks for getting these through!

PR #2449, opened 2023-11-18
fix: Remove empty back bar from confirmation component

DafyddLlyr:
## What does this PR do?
- Hides the back bar when we reach a confirmation component

This is a small visual thing which has bugged me since we introduced the save and return status pages which have the banner starting at the top of the page, making this look a little inconsistent.

**Why not just hide the entire backbar on a single condition?**
This would mean that the layout would jump in instances where the back button should be hidden (e.g. first card).

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
@@ -43,6 +43,7 @@ export interface PreviewStore extends Store.Store {
     upcomingCardIds?: Store.nodeId[],
   ) => Store.nodeId | undefined;
   canGoBack: (node: Store.node | null) => boolean;
+  getType: (node: Store.node | null) => TYPES | undefined;
   computePassport: () => Readonly<Store.passport>;
   record: (id: Store.nodeId, userData?: Store.userData) => void;
   resultData: (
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -187,13 +188,15 @@ export const previewStore: StateCreator<
...
     //      optimised/memoized, see related isFinalCard() comment below
-    const { flow, hasPaid, previousCard } = get();
-    return (
-      Boolean(node?.id) &&
-      flow[node!.id!]?.type !== TYPES.Confirmation &&
-      Boolean(previousCard(node)) &&
-      !hasPaid()
-    );
+    const { hasPaid, previousCard } = get();
+    return Boolean(node?.id) && Boolean(previousCard(node)) && !hasPaid();
```
Mike-Heneghan, line 192: _Non-blocking Question:_

Is it worth removing the check that the component is a `Confirmation` from the `canGoBack`? Part of me wonders if although the `back` is hidden in this new approach the `canGoBack` should maybe still be able to be a source of truth over whether it should be possible?

This is maybe silly and over defensive so please feel free to ignore.

Mike-Heneghan (approved):
Looks good to me 👍 

One trivial question about the change to `canGoBack` but certainly not a blocker.

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
DafyddLlyr: Yeah it's a fair question - my first attempt did just rely on `canGoBack()`.

However there are instances where we want the empty back bar in place for layout reasons, but `canGoBack()` would return a `false` - for example the first component in a flow. Without this the layout "jumps" a little which ends up being pretty noticable.

PR #2450, opened 2023-11-18
feat: Create an `analytics_summary` database view for analytics

DafyddLlyr:
## What does this PR do?
- Follow on from https://github.com/theopensystemslab/planx-new/pull/2430
- Should make the painful `JOIN` operations in the GUI we were making Silvia do redundant 🤞 
- Potentially it could mean that we could hide most other tables from Metabase, and just expose the summary tables
- Very much a first pass, open to suggestions here on improvements!


hasura.planx.uk/migrations/1700303840464_setup_analytics_summary_view/up.sql
```diff
@@ -0,0 +1,21 @@
+CREATE OR REPLACE VIEW public.analytics_summary AS 
+select 
+	a.id as analytics_id,
+	al.id as analytics_log_id,
+	f.slug as service_slug,
+	t.slug as team_slug,
+	a.type as analytics_type,
+	a.created_at as analytics_created_at,
+	a.user_agent as user_agent,
```
DafyddLlyr, line 9: I'm hoping we can keep this as a JSONB column with the changes in https://github.com/theopensystemslab/planx-new/pull/2451

If not, I can just pull out the key bits of data here like browser, platform etc

jessicamcinchak: Agree with initially keeping as JSONB! 

nit: `a.user_agent as user_agent` is redundant / unnecessary and adds a bit of noise? only need `as` if actually renaming, table reference won't be visible in final view (applies to a number of columns below too)

jessicamcinchak (commented):
Looks good to me, but I'll let Mike do final approval of this one as he's most familiar with pain points around current joins :+1:

hasura.planx.uk/migrations/1700303840464_setup_analytics_summary_view/up.sql
DafyddLlyr: Good catch 👍

jessicamcinchak:
Thanks for final review, Mike - I'm going to pop on this branch and remove the empty migration, then merge :heavy_check_mark:

Mike-Heneghan:
> Thanks for final review, Mike - I'll merge this ✔️
> 
> Are you happy to take care of making visible in Metabase and updating SQL so we can eventually "hide" raw tables if no longer necessary? (as you go, doesn't need to be an immediate swap!)

Once we've side stepped the name change issue I'd be happy to pick this up 👍

PR #2457, opened 2023-11-19
feat: add `hasPlanningData` property to `teams.settings`

jessicamcinchak:
As more councils are coming onboard, this change will let us skip [PRs for Trello tickets like this one](https://trello.com/c/JVlAfSp5/2701-hook-up-planning-data-constraints-for-st-albans) and instead just add the property `"hasPlanningData": true` to `teams.settings` in the prod Hasura console for this council. No changes to permissions or the staging data sync are necessary for this to work as expected.

We will **still** open PRs to add API GIS metadata when a council's granular Article 4 variables become available.

Also updated council onboarding step in README here: https://github.com/theopensystemslab/planx-team-onboarding/tree/main#after-running-this-script

Future scope: expose this property via the Team Settings page in the Editor or via admin endpoint in the short term (probably not worth it?)

### DIFF
--- a/editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
@@ -55,23 +55,11 @@ function Component(props: Props) {
   const wktPolygon: string | undefined =
     siteBoundary && stringify(siteBoundary);
 
-  // Configure which planx teams should query Digital Land (or continue to use custom GIS) and set URL params accordingly
-  //   In future, Digital Land will theoretically support any UK address and this list won't be necessary, but data collection still limited to select councils!
-  const digitalLandOrganisations: string[] = [
-    "opensystemslab", // for UK-wide testing, subject to data availability
-    "barnet",
-    "birmingham",
-    "buckinghamshire",
-    "canterbury",
-    "camden",
-    "doncaster",
-    "gloucester",
-    "lambeth",
-    "medway",
-    "newcastle",
-    "southwark",
-    "st-albans",
-  ];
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
```diff
@@ -55,23 +55,11 @@ function Component(props: Props) {
...
-    "canterbury",
-    "camden",
-    "doncaster",
-    "gloucester",
-    "lambeth",
-    "medway",
-    "newcastle",
-    "southwark",
-    "st-albans",
-  ];
```
jessicamcinchak, line 74: I'll update `teams.settings.hasPlanningData` for these councils on merge :heavy_check_mark: 

Please update a single record on the pizza to test: 
- You should see the "{Council} is not a supported Local Authority" error before updating `teams.settings`
- You should successfully query constraints as expected after updating

jessicamcinchak:
[Hook up Planning Data constraints for St Albans](https://trello.com/c/JVlAfSp5/2701-hook-up-planning-data-constraints-for-st-albans)

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
DafyddLlyr: I wonder if it's worth capturing this as a Hasura migration? The value of `hasPlanningData` will always be true across all environments, and would mean we wouldn't have to add the column to the sync script.

DafyddLlyr: Getting the expected results for `true`, `null` and `false` values 👍

DafyddLlyr (approved):
Thanks for picking this up!

PR #2460, opened 2023-11-20
chore: Drop p-memoize in favour of Apollo caching

DafyddLlyr:
## What the problem?
- `NotFound` errors on routes in staging (not replicable yet on local dev)
- See OSL Slack for context - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1700476380638669

## What's the solution?
- I'm not totally confident that this is the cause but as the query isn't firing it makes sense to try this
- At minimum we rationalise how we do caching / memoization, drop our bundle size, and rule a variable out of the equation.

### DIFF
--- a/editor.planx.uk/package.json
+++ b/editor.planx.uk/package.json
@@ -59,7 +59,6 @@
     "nanoid-good": "^3.1.0",
     "natsort": "^2.0.3",
     "navi": "^0.15.0",
-    "p-memoize": "^7.1.1",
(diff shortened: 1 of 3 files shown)


jessicamcinchak (approved):
Makes sense :+1:

DafyddLlyr:
Yeah not 100% sure this is the cause but as the query isn't firing it makes sense to try this. 

At minimum we rationalise how we do caching / memoization and drop our bundle size.

PR #2464, opened 2023-11-20
fix: Grant select permission to `submitted_at`

DafyddLlyr:
Fix for bug introduced here - https://github.com/theopensystemslab/planx-new/pull/2438/files

Caught by Airbrake here - https://planx.airbrake.io/projects/329753/groups/3675812447178036881


jessicamcinchak (approved):
Thank you!

PR #2468, opened 2023-11-21
feat: add toggle to optionally format Calculate output for automations

jessicamcinchak:
Solves what we're discussing in this thread: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1700494465669089

**Key changes:**
- Conditionally change the `output` data format depending on use for automations
- Simplify file structure - eg `Public.tsx` instead of `Public/index.tsx`
- Add tests

**How this works:**
By default, calculate writes a single value to the breadcrumbs - eg `"data": { "abPath": 2 }`. This is necessary to use multiple Calculate components in a row and be able to pass in number inputs (as best as I understand!).

This PR introduces an Editor-facing toggle to format the output as an array with a single string item which can then be successfully referenced to auto-answer Questions or Checklists - eg `"data": { "abPath": ["2"] }`. 

SetValue similarly always returns an array of a single string item which is how it's able to power automations. The Pay component can successfully read a value in either format.

**Test flow:**
https://2468.planx.pizza/testing/a-b-testing


jessicamcinchak:
[Update Calculate component output to support A/B testing automations](https://trello.com/c/1vqYwVUD/2710-update-calculate-component-output-to-support-a-b-testing-automations)

editor.planx.uk/src/@planx/components/Calculate/logic.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+    "data": {
+      "val": "2",
+      "text": "2"
+    },
+    "type": TYPES.Response,
+    "edges": [
+      "Group2Notice"
+    ]
+  }
+};
```
jessicamcinchak, line 185: Having a `logic.test.ts` isn't really an existing pattern in `@planx/components/`, but I think this feels like a more meaningful organisation than adding more noise to the preview store tests?

Mike-Heneghan (approved):
This looks great 🥳 

As this develops and gets more complex I'm wondering how this could be handled to differentiate between different multivariate tests and maybe even more complex stuff like nested tests. 

I think it's great this has the flexibility of what the passport variable is assigned so it could maybe be managed in a similar way to current passport variables. 

I'm really excited to see how this gets used and what we can learn 👀

PR #2469, opened 2023-11-21
Revert "chore: Stream S3 files directly to zip"

jessicamcinchak:
Reverts theopensystemslab/planx-new#2380 - building this pizza again to see if it introduced issue described here (**update: it did**): https://opendigitalplanning.slack.com/archives/C0241GWFG4B/p1700561792278499

Fixes issue where user-uploaded files appear not to be included in zip reported in #help-issues. On production, I can still see a `/tmp` folder &rarr; `{sessionId}_{s3FolderId}` folder &rarr; user uploaded files, but it sounds like for MS users the `/tmp` folder is being hidden altogether ! 

This one was a small tidy up without major performance impacts, so just going to revert & deploy fix for now & will add a Trello ticket to pick back up and solve properly :ok_hand:

### DIFF
--- a/api.planx.uk/modules/send/utils/exportZip.test.ts
+++ b/api.planx.uk/modules/send/utils/exportZip.test.ts
@@ -7,6 +7,7 @@ jest.mock("fs", () => ({
   existsSync: () => true,
   unlinkSync: () => undefined,
   createWriteStream: () => undefined,
+  writeFileSync: () => undefined,
   rmSync: () => undefined,
 }));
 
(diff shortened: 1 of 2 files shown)


Mike-Heneghan (approved):
Looks good to me 👍 

Some things I was surprised about but I don't think are anything to do with this piece of work. 

The `Overview.htm` and `RedactedOverview.htm` are name with `.htm` rather than `.html`

I couldn't see how the labels I added to the files could be associated with the uploaded files from what I could see in the zip.

jessicamcinchak:
Thanks @Mike-Heneghan ! 

- `htm` versus `html` is a very strange Uniform DMS requirement :see_no_evil: (but either should behave same when opened in your browser)
- Labels are only currently shown/associated to files in the BOPS payload (and forthcoming ODP Schema), this is another known limitation of OneApp XM

PR #2471, opened 2023-11-22
chore: remove react-feather icon library

ianjon3s:
## What does this PR do?

The app currently uses a mixture of the React Feather and MUI Icons libraries to serve icons.

Instances of React Feather replaced with MUI Icons and dependency removed.

Edit: also updates sizing and margins to increase visual and target size.

Before:
image

After:
image


Mike-Heneghan (approved):
Looks good to me 🥳

PR #2472, opened 2023-11-22
Revert "feat: Create an `analytics_summary` database view for analytics"

jessicamcinchak:
Reverts theopensystemslab/planx-new#2450

Original merge failed to deploy to staging & Pulumi action was hanging for ~20 minutes until timing out. `analytics_summary` was never created on dev. 

My best guess is that has to do with the renamed migration here? I think the quickest fix is to revert here & just open a fresh PR with the view creation again in a new migration ?


Mike-Heneghan (approved):
Sounds like a good plan! 

I'd incorrectly assumed that git magic would have resolved the naming change but as you say it seems like the most likely culprit.

PR #2475, opened 2023-11-22
chore: Unify colour refs in theme

ianjon3s:
## What does this PR do?

The theme.ts file currently uses a mixture of variables, palette variables and hardcoded colours, this PR simplifies it so that the palette variables are used throughout.


jessicamcinchak (approved):
Much tidier and all displaying as expected on /preview links for me :sparkles:

PR #2476, opened 2023-11-23
infra: update prod deploy Github Action PAT expiration date [skip pizza]

jessicamcinchak:
I created a new PAT and updated this repos Settings &rarr; Secrets &rarr; Actions &rarr; `PROD_DEPLOY_PAT`

I'll also set up a new Slack reminder for this new expiration date :alarm_clock: 

Will queue up a production deploy this afternoon to confirm all working as expected !

PR #2477, opened 2023-11-23
Production deploy

jessicamcinchak:
This one will specifically test that the new PAT is working as expected :crossed_fingers: #2476


Mike-Heneghan (approved):
Looks good to me 👍

PR #2478, opened 2023-11-23
fix: sanitise sharedb node data using `dompurify`

jessicamcinchak:
Testing notes in devops email

Changes:
- When creating an `_add` or `_update` operation against SharedDB, we would already "sanitize" values using custom regex to handle empty values. I've added an extra step here to additionally apply `DOMPurify.sanitize()` on the user data inputs as well

Todo:
- [x] Add unit tests
- [ ] Determine if there's anywhere else we need to similarly apply `DOMPurify.sanitize()` too (eg are these methods only catching user input and not interceptions between user input & postgres commands??) - see #2479 
- [ ] Test in Planx editor and determine if this is "over-purifying", eg do we need to allow safe HTML & SVG per [docs here ](https://github.com/cure53/DOMPurify#control-our-allow-lists-and-block-lists)


editor.planx.uk/src/@planx/graph/index.ts
```diff
@@ -46,7 +47,7 @@ const isExternalPortalNodeType = (id: string, graph: Graph): boolean =>
 
 const sanitize = (x: any) => {
   if ((x && typeof x === "string") || x instanceof String) {
-    return trim(x.replace(/[\u200B-\u200D\uFEFF↵]/g, ""));
+    return DOMPurify.sanitize(trim(x.replace(/[\u200B-\u200D\uFEFF↵]/g, "")));
```
jessicamcinchak, line 50: This pre-existing bit of regex deals with empty values like covered in this test - I don't think we want to lose this functionality !
```ts
test("ignores empty values", () => {
  const [graph, ops] = add({
    id: "test",
    type: 100,
    data: {
      info: "\u200B​",
      policyRef: "​\n",
      empty: "​",
      another: "↵",
      howMeasured: undefined,
      text: "efef",
      description: " ​",
    },
  })({});
  expect(graph).toEqual({
    _root: {
      edges: ["test"],
    },
    test: {
      type: 100,
      data: {
        text: "efef",
      },
    },
  });
  expect(ops).toEqual([
    { oi: { edges: ["test"] }, p: ["_root"] },
    { oi: { data: { text: "efef" }, type: 100 }, p: ["test"] },
  ]);
});
```

editor.planx.uk/src/@planx/graph/index.ts
```diff
@@ -44,12 +45,14 @@ const isSectionNodeType = (id: string, graph: Graph): boolean =>
...
 
-const sanitize = (x: any) => {
+export const sanitize = (x: any) => {
   if ((x && typeof x === "string") || x instanceof String) {
+    x = DOMPurify.sanitize(x as string);
     return trim(x.replace(/[\u200B-\u200D\uFEFF↵]/g, ""));
   } else if ((x && typeof x === "object") || x instanceof Object) {
     return Object.entries(x).reduce((acc, [k, v]) => {
       v = sanitize(v);
+      acc[k] = v;
```
jessicamcinchak, line 55: This block only `returns acc`, and `v = sanitize(v)` was indeed sanitizing, but not actually updating the `acc` which would be returned: 
- In cases of handling empty values, this didn't matter because we `delete acc[k]`
- But in cases of sanitising values we want to keep, we need to update the return value too

PR #2479, opened 2023-11-24
fix: sanitize sharedb node data using `dompurify` in sharedb.planx.uk

jessicamcinchak:
Complementary second attempt to editor-side change in #2478. Pointing to `main` to build a separate pizza for comparison testing. If works, can decide if we still want both sharedb & editor changes & cherry pick commits.

Not confident this will successfully build as I'm getting a `planx-new-sharedb-1 | Error: Cannot find module 'dompurify'` docker error locally, but pushing up so others can clone down this branch (please don't hesitate to push commits direct to it if you make fixes).


jessicamcinchak:
Replaced by #2484

PR #2484, opened 2023-11-27
feat: Sanitise operations on commit to ShareDB

DafyddLlyr:
## What does this PR do?
 - Uses [DOMPurify](https://github.com/cure53/DOMPurify) to sanitise input on "commit" to ShareDB

## Next steps
- Look at blocking flow updates via GraphQL API
- Try to work out a reliable testing methodology (in another PR). It looks like Playwright has a WebSocket class that should allow the intercept of an outgoing frame to attempt to inject unsafe content (Docs: https://playwright.dev/docs/api/class-websocket)

## Demo
https://github.com/theopensystemslab/planx-new/assets/20502206/6fa17f8e-f53f-42b6-94c2-314518a640da

### DIFF
--- a/sharedb.planx.uk/package.json
+++ b/sharedb.planx.uk/package.json
@@ -4,9 +4,11 @@
   "private": true,
   "dependencies": {
     "@teamwork/websocket-json-stream": "^2.0.0",
+    "dompurify": "^3.0.6",
+    "jsdom": "^23.0.0",
     "jsonwebtoken": "^8.5.1",
     "pg": "^8.11.3",
-    "sharedb": "^3.3.1",
+    "sharedb": "^4.1.1",
     "ws": "^8.14.2"
   },
   "scripts": {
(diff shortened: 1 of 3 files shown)


sharedb.planx.uk/package.json
```diff
@@ -4,9 +4,11 @@
   "private": true,
   "dependencies": {
     "@teamwork/websocket-json-stream": "^2.0.0",
+    "dompurify": "^3.0.6",
+    "jsdom": "^23.0.0",
     "jsonwebtoken": "^8.5.1",
     "pg": "^8.11.3",
-    "sharedb": "^3.3.1",
+    "sharedb": "^4.1.1",
```
DafyddLlyr, line 11: The only breaking changes in v4 is dropped support for Node 14.

Docs: https://github.com/share/sharedb/releases/tag/v4.0.0

sharedb.planx.uk/server.js
```diff
@@ -16,6 +18,10 @@ const sharedb = new ShareDB({
   }),
 });
 
+// Setup JSDOM and DOMPurify
```
DafyddLlyr, line 21: Docs: https://github.com/cure53/DOMPurify#running-dompurify-on-the-server

jessicamcinchak: I hadn't totally connected last week that sharedb qualifies as "on the server" and therefore was trying to use _without_ JSDOM, but this makes much more sense in hindsight :+1:

PR #2486, opened 2023-11-27
fix: help text pointer overlap

ianjon3s:
### What does this PR do?

The updated help text introduced a bug where the help text button container overlaps certain elements. This is caused by an absolute-positioned div that runs along the right-hand-side of the column.

By removing pointer-events from the column and reintroducing it on the button, no content is blocked.

Current example (bug present):
https://editor.planx.dev/testing/test-upload-and-label/preview?analytics=false

Fixed example:
https://2486.planx.pizza/testing/test-upload-and-label/preview?analytics=false


ianjon3s:
Thanks for quick review @jessicamcinchak !

PR #2454, opened 2023-11-18
feat: Analytics logs for input errors caught by `ErrorWrapper`

DafyddLlyr:

PR #2487, opened 2023-11-27
feat: track the answers to allow_list questions in analytics_logs

Mike-Heneghan:
## What:

- Passport variable tracking:
  - Begin tracking the `fn` or `val` of a `node` 
  - Store this data against a new column `node_fn` on the `analytics_logs`

- Answers to allow_list question tracking:
  - Defined a new array of questions which are part of the allow_list (i.e. useful for analytics but not sensitive and or PII)
  - On component transition check whether the `node` `fn` or `val` is in the allow list and if so update the log with the user's answer
  
## Why:

- Currently, node are generally identified in analytics by their `node_title` and `node_type`.
- This is useful information but there may be situations where it's usefully to additionally know their `fn` or `val` if they're assigned
- For example the `node_title` code change but we could easily continue to track via the passport variable which is unlikely to change
- Also, some questions in flowd could provide really useful insights to users but also not contain any PII or be invasive to track
- These questions are harmless to track and hence could be on the `allow_list` 

## Screen recoding

https://github.com/theopensystemslab/planx-new/assets/36415632/7994f4f7-0d14-49b5-8673-61d91adee213

## Note

- At the moment this only really works as a user goes forward in a flow. Do we need to track a user going backwards?
- There are analytics_logs records created going backwards so even if a user has completed an answer for `proposal.projectType` that wouldn't be stored on the analytics_log which was generated on hitting `Back`
- Would it make sense to track this when going backwards?

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
@@ -21,7 +21,7 @@ type AnalyticsLogDirection =
 
 const allowList = [
   "proposal.projectType",
-  "application.delcaration.connection",
+  "application.declaration.connection",
 ];
 
 export type HelpClickMetadata = Record<string, string>;
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
@@ -19,6 +19,11 @@ type AnalyticsLogDirection =
   | "reset"
   | "save";
 
+const allowList = [
+  "proposal.projectType",
+  "application.delcaration.connection",
+];
+
 export type HelpClickMetadata = Record<string, string>;
 export type SelectedUrlsMetadata = Record<"selectedUrls", string[]>;
 export type BackwardsNavigationInitiatorType = "change" | "back";
@@ -75,6 +80,7 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     previewEnvironment,
     flowId,
     flow,
+    previousCard,
   ] = useStore((state) => [
     state.currentCard,
     state.breadcrumbs,
@@ -84,8 +90,10 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     state.previewEnvironment,
     state.id,
     state.flow,
+    state.previousCard,
   ]);
   const node = currentCard();
+  const previousNodeId = previousCard(node);
   const isAnalyticsEnabled =
     new URL(window.location.href).searchParams.get("analytics") !== "false";
   const shouldTrackAnalytics =
@@ -171,6 +179,13 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
       updateLastLogWithNextLogCreatedAt(lastAnalyticsLogId, newLogCreatedAt);
     }
 
+    if (lastAnalyticsLogId && previousNodeId) {
+      await updateLastLogWithAllowListAnswers(
+        lastAnalyticsLogId,
+        previousNodeId,
+      );
+    }
+
     lastAnalyticsLogId = id;
   }
 
@@ -242,6 +257,34 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
         next_log_created_at: newLogCreatedAt,
       },
     });
+  }
+
+  async function updateLastLogWithAllowListAnswers(
+    lastAnalyticsLogId: number,
+    previousNodeId: string,
+  ) {
+    const allowListAnswers = getAllowListAnswers(previousNodeId);
+    if (allowListAnswers) {
+      await publicClient.mutate({
+        mutation: gql`
+          mutation UpdateAllowListAnswers(
+            $id: bigint!
+            $allow_list_answers: jsonb
+          ) {
+            update_analytics_logs_by_pk(
+              pk_columns: { id: $id }
+              _set: { allow_list_answers: $allow_list_answers }
+            ) {
+              id
+            }
+          }
+        `,
+        variables: {
+          id: lastAnalyticsLogId,
+          allow_list_answers: allowListAnswers,
+        },
+      });
+    }
   }
 
   async function trackHelpClick(metadata?: HelpClickMetadata) {
@@ -406,6 +449,19 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
       type: type,
     };
     return nodeMetadata;
+  }
+
+  // Relies on being called at a point where the user has answered the Qs on the node.
+  function getAllowListAnswers(nodeId: string) {
+    const { data } = flow[nodeId];
+    const nodeFn = data?.fn || data?.val || null;
+    if (nodeFn && allowList.includes(nodeFn)) {
+      const answerIds = breadcrumbs[nodeId].answers;
+      const answerValues = answerIds?.map((answerId) => {
+        return flow[answerId].data.val;
+      });
+      return answerValues;
+    }
   }
 
   /**
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
@@ -189,6 +189,7 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
           $node_type: Int
           $node_title: String
           $user_agent: jsonb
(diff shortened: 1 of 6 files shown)


editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -19,6 +19,11 @@ type AnalyticsLogDirection =
   | "reset"
   | "save";
 
+const allowList = [
+  "proposal.projectType",
+  "application.delcaration.connection",
```
jessicamcinchak, line 24: nit: typo here `delcaration` &rarr; `declaration`

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -405,6 +451,19 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     return nodeMetadata;
   }
 
+  // Relies on being called at a point where the user has answered the Qs on the node.
+  function getAllowListAnswers(nodeId: string) {
+    const { data } = flow[nodeId];
+    const nodeFn = data?.fn || data?.val || null;
```
jessicamcinchak, line 457: nit: is `|| null` necessary here if you're checking for existence of `nodeFn && ` on the next line?

hasura.planx.uk/migrations/1700845956731_alter_table_public_analytics_logs_add_column_allow_list_answers/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."analytics_logs" add column "allow_list_answers" jsonb
+ null default '[]'::jsonb;
```
jessicamcinchak, line 2: I wonder if there are any benefits/clarity of adding this column as [an array type](https://www.postgresql.org/docs/current/arrays.html) (eg `text[]`) rather than `jsonb`? 

An array feels more representative of the actual expected data structure to me, but I'd be curious if there's any notable Metabase pros/cons to working with one type over the other?

jessicamcinchak (commented):
Solution is looking good to me, but I am getting some unexpected results when going through a flow in the `analytics_logs` table on this pizza that I want to flag: 
Screenshot from 2023-11-29 08-44-03

What was happening ([db rows here](https://hasura.2487.planx.pizza/console/data/default/schema/public/tables/analytics_logs/browse?sort=created_at%3Basc)):
- The first row for `proposal.projectType` is right
- But then my answer is incorrectly recorded for two Set Values with un-allow-listed fns  of `proposal.droppedKerbOnly` & `proposal.treeWorksOnly` (is it possible to make mutation param check stricter to ensure it _only_ writes answers for allow-listed fns?)
- Then I hit a follow-up, more granular project type question "Select all the works to replace windows or doors" that correctly records answer `["alter.replace.windowsToWindows"]`
- There's also a number of duplicates (maybe writing on component renders or user interactions within the card rather that whole-card transition??) - I realise this may not be introduced here and a wider caveat of analytics tracking / happy for it to be thought about in a separate PR

As for the larger question in your description about currently only working on "forward" and whether we should track on "back":
- My high-level opinion about this would be that we should try to track allow-list answers as similarly as possible to other interactions like "help click" & "input errors" - so if those are tracked on back, then let's try to track answer changes on back too. Helps keep the "explainability" of analytics methodology more straightforward for external consumers in the long run!

hasura.planx.uk/migrations/1700845956731_alter_table_public_analytics_logs_add_column_allow_list_answers/up.sql
Mike-Heneghan: My understanding of this is that the `allow_list_answers` column is  `jsonb` but it defaults to `[]` if a value isn't provided. 

My thinking was that at the moment we expect this to be an array but it being `jsonb` gives us flexibility.

I was basing this on https://github.com/theopensystemslab/planx-new/pull/2454/files#diff-600c74a4ab8a37ce56adffbba43f1443538b77fcc91d4b9db7f344eb0066c167 although I've maybe misunderstood?

I'm not sure if this really answers your question 🤔

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
Mike-Heneghan: Fixed here: bb4e4fbb2f2378a3614a40b3ebb01c3b836c126a

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
Mike-Heneghan: Fixed here: 1a06af7fe335dc8d877d6c0680e3df5cfd4001d5

Mike-Heneghan:
> Solution is looking good to me, but I am getting some unexpected results when going through a flow in the `analytics_logs` table on this pizza that I want to flag: Screenshot from 2023-11-29 08-44-03
> 
> What was happening ([db rows here](https://hasura.2487.planx.pizza/console/data/default/schema/public/tables/analytics_logs/browse?sort=created_at%3Basc)):
> 
> * The first row for `proposal.projectType` is right
> * But then my answer is incorrectly recorded for two Set Values with un-allow-listed fns  of `proposal.droppedKerbOnly` & `proposal.treeWorksOnly` (is it possible to make mutation param check stricter to ensure it _only_ writes answers for allow-listed fns?)
> * Then I hit a follow-up, more granular project type question "Select all the works to replace windows or doors" that correctly records answer `["alter.replace.windowsToWindows"]`
> * There's also a number of duplicates (maybe writing on component renders or user interactions within the card rather that whole-card transition??) - I realise this may not be introduced here and a wider caveat of analytics tracking / happy for it to be thought about in a separate PR
> 
> As for the larger question in your description about currently only working on "forward" and whether we should track on "back":
> 
> * My high-level opinion about this would be that we should try to track allow-list answers as similarly as possible to other interactions like "help click" & "input errors" - so if those are tracked on back, then let's try to track answer changes on back too. Helps keep the "explainability" of analytics methodology more straightforward for external consumers in the long run!

I'm struggling to replicate the issue 🤔 Looking at the analytics it seems like you were using the `find-out-if-you-need-planning-permission` flow? Although that doesn't seem to have the questions you encountered. 

If the flow you tested with is available would you mind sharing a link to it?

hasura.planx.uk/migrations/1700845956731_alter_table_public_analytics_logs_add_column_allow_list_answers/up.sql
jessicamcinchak: Mimicing behavior or recently added input errors column works for me, thanks for pointing that out!

Mike-Heneghan:
> @Mike-Heneghan I was using Lambeth's FOI on the pizza (FOI flows outside of active council teams likely don't have maintained content) - I'm almost always the only Linux user in pizza logs 🐧 Screenshot from 2023-11-29 13-58-02
> 
> The project type checklist and set values are nested in the `permitteddevelopment` flow / external portal. The set values will always be hidden from the user journey, but auto-answered in the background based on my project selection - so logging at the same time is expected I suppose (rather than a duplication), but recording the answer _is not expected_ if the `fn` for that row does not match nor is allow-listed itself if that makes sense ?

Thanks @jessicamcinchak! That's really helpful, I imagine the issue is that I wasn't testing auto-answer questions which will completed really quickly and are more susceptible to this issue! I'll use the lambeth foi to have something a bit more representative to the real behaviour.

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -405,6 +456,19 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     return nodeMetadata;
   }
 
+  function getAllowListAnswers(nodeId: string) {
+    const { data } = flow[nodeId];
+    const nodeFn = data?.fn || data?.val;
+    if (nodeFn && allowList.includes(nodeFn)) {
+      const answerIds = breadcrumbs[nodeId]?.answers;
```
Mike-Heneghan, line 463: I think the reason this doesn't work going backwards is that the current approach has an assumption that we always want to update the analytics log for the previous node with the answer which is stored in the current breadcrumbs. 

When we move backwards through a flow the breadcrumbs are removed and the answers for the question go with them.

We could track what the saved answers were when a user loads a node they've already answered in the analytics logs. 

Although I'd be tempted not to and only track allow_list answers when a user explicitly hits continue. I think this might be more consistent with the broad approach to user's answers?

PR #2483, opened 2023-11-27
fix: apply `DOMPurify.santize` on `ReactMarkdownOrHtml`

jessicamcinchak:
Replaces #2478 by addressing issue at first point of user interaction in the client. Still requires #2479 to fix root issue.

PR #2488, opened 2023-11-28
fix: Restrict write access to `flows.data` column via GraphQL API

DafyddLlyr:
## What does this PR do?
 - Restricts write access to the `flows.data` column for all user roles
 - This means that this column cannot be updated via the GraphQL API (unless using the admin role)
 - The aim of this is to ensure that only sanitised data can be entered into this column

### Context
Follow on from https://github.com/theopensystemslab/planx-new/pull/2483 & https://github.com/theopensystemslab/planx-new/pull/2484

I was hoping for a slightly simpler solution of relying on Hasura's backend-only permissions but this is just per-table, not per-column.

We should probably extend this approach to similar columns such as `global_settings.footer_content` and `flows.settings`. It might be best to follow the pattern of Editor → API → Hasura for these operations where we require sanitation before writing to the DB. Let's discuss this at the dev call!


api.planx.uk/modules/flows/findReplace/service.ts
```diff
@@ -84,8 +84,10 @@ const findAndReplaceInFlow = async (
   }
 
   // if matches, proceed with mutation to update flow data
-  const { client: $client } = getClient();
-  const response = await $client.request<UpdateFlow>(
+
+  // XXX: This is using the API Hasura role
```
DafyddLlyr, line 88: Let me know if a comment is not good enough here!

Alternatives considered are - 
 - Attempt a mutation before `UpdateFlow` which we know user has permission for (e.g. set `flow_slug` = `flow_slug`). If this works we know user should be allowed to update `flow.data`
 - Manually query permission (e.g. check join of userId → team_members → teamId → flowId). The tricky thing about a manual check is that this logic will always have to be updated alongside our permission and roles model.
 
 For now, we're checking if the user is a platformAdmin via middleware and then executing this mutation. This is totally fine I think - I just wanted to leave a cautionary note behind.

jessicamcinchak: Agree with your decision/approach here :heavy_plus_sign: thanks for explaining alternatives

api.planx.uk/modules/flows/findReplace/controller.ts
```diff
@@ -17,7 +23,10 @@ export const findAndReplaceSchema = z.object({
   }),
   query: z.object({
     find: z.string(),
-    replace: z.string().optional(),
+    replace: z
+      .string()
+      .optional()
+      .transform((val) => val && DOMPurify.sanitize(val)),
```
jessicamcinchak, line 29: Thanks for catching this sanitation step too :+1:

jessicamcinchak (approved):

PR #2489, opened 2023-11-28
fix: get team from store, not route data, to display permissions icons in Editor

jessicamcinchak:
Fixes bug reported here: https://opendigitalplanning.slack.com/archives/C0241GWFG4B/p1701170825043769

Underlying permissions were correct, but edit-able teams were showing a view-only icon at the flow level. 

Steps to re-create issue:
- Login into editor as a non-platformAdmin
- At the teams-level, you see the correct icons
- Pick a team you can edit, you see the correct icons for the list of services
- At the graph view of an individual flow, you see the incorrect icon but have the correct permissions


DafyddLlyr (approved):
Perfect!

PR #2493, opened 2023-11-28
fix: remove zero padding for USRNs

jessicamcinchak:
Per feedback from Birmingham about not picking up a classified road as expected - has been tested by council and working as expected now :heavy_check_mark: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1701099275002579

Notes: 
- Original UPRN padding comes from here, but it was a faulty assumption that USRN should be same: https://github.com/theopensystemslab/planx-new/pull/694
- ODP schema only defines a `maxLength` for USRN & UPRN, so this will still validate fine!


Mike-Heneghan (approved):
Looks good to me 👍

PR #2495, opened 2023-11-28
chore: Stream S3 files directly to zip (take 2)

DafyddLlyr:


jessicamcinchak:
[Stream files directly from S3 to export ZIPs without nesting files](https://trello.com/c/WhC8fOY4/2706-stream-files-directly-from-s3-to-export-zips-without-nesting-files)

jessicamcinchak (approved):
Nice & simple - thanks for picking this one back up!

PR #2496, opened 2023-11-29
fix: add auth header to public image upload

Mike-Heneghan:
## What 

- Add auth header to the public file upload

## Why

- The public upload requires at least `teamEditor` permissions for uploading files

## Screen recording

Bug:

https://github.com/theopensystemslab/planx-new/assets/36415632/6224bd31-2b53-4906-b77e-b00841461268

Fix:

https://github.com/theopensystemslab/planx-new/assets/36415632/d704e7d5-4934-4077-b976-810e0594ecd0


DafyddLlyr (approved):
👌

PR #2501, opened 2023-11-30
chore: Remove obsolete Source Sans Pro ref from header

ianjon3s:
## What does this PR do?

The header files for PlanX contain obsolete references to Source Sans Pro, served from Google Fonts.

Removing these should save on potential unnecessary files being loaded in a user session.


jessicamcinchak (approved):
Looks like map fonts are all coming through/inheriting as expected direct from web component and no longer relying on these imports :+1: thanks for tidying this up!! 

Screenshot from 2023-11-30 14-06-12
Screenshot from 2023-11-30 14-05-51

PR #2503, opened 2023-11-30
refactor: Consolidate rich text into component + update styles

ianjon3s:
## What does this PR do?

Currently the styling for rich text sits in a standalone CSS file, this PR brings the styles into the `RichTextInput.tsx` component and makes small updates to tidy up styling in the editor and public interface.

#### Things to point out

- The CSS for rich text was also imported into the component `RichTextImage.tsx`, however I found no references to this styles in either the code or the output (example below), so I've not imported the new `RichContentContainer` styled div to this component.
- The styles for `.passport` remain applied as a CSS class, as there are programatic ways this is applied in the component that I were not able to unpick. Happy to revisit this if it is deemed worthy to refactor.

**Editor example:**
https://2503.planx.pizza/lambeth/content-example

**Public example:**
https://2503.planx.pizza/lambeth/content-example/preview?analytics=false


DafyddLlyr (approved):
Perfect - thanks for picking this one up 👍 

All looks and works as expected!

PR #2510, opened 2023-12-01
chore: tidy logic tests

jessicamcinchak:
Replaces #2299 

Changes:
- Splits up `preview.test.ts` by method name for more isolated testing (this isn't perfect because so many store methods call each other, but definitely an improvement in readability I think!)
- Takes a first pass through mock flow data and tidies up types (import from `TYPES`), replaces node ids with human readable ones, etc. Definitely more that could be consolidated here, but there's enough initial changes here for now

There's a few new & existing `.skip` and `.todo` annotated tests throughout which detail future expected behavior for autoAnswers, filters, getResult and more.

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/flags.test.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/flags.test.ts
@@ -94,33 +94,33 @@ describe("changing flag inside flag filter doesn't affect the filter's behaviour
             val: "MISSING_INFO",
             text: "Missing information",
           },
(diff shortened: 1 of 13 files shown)


jessicamcinchak:
[Try only recording an auto-answer when a node is actually reached, rather than in the future](https://trello.com/c/7GPqwZSa/2245-try-only-recording-an-auto-answer-when-a-node-is-actually-reached-rather-than-in-the-future)

DafyddLlyr (approved):
This is honestly great to see - should make a lot of this code easier to understand and update 🎉

PR #2529, opened 2023-12-03
feat: send ODP Schema payload to BOPS `/api/v2/planning_applications`

jessicamcinchak:
BOPS is ready to accept our ODP Schema payload via `/api/v2/planning_applications` ([Swagger docs here](https://southwark.bops-staging.services/api/docs/index.html?urls.primaryName=API%20V2%20Docs)). 

We will still send our existing payload via `/api/v1/planning_applications` for an initial testing phase (length to be determined).

**Changes:**
- Adds a new Planx API endpoint `/bops-v2/:localAuthority`
- Creates **two** Hasura scheduled events when "BOPS" is a selected destination in the Send component
  - Since BOPS will accept the ODP payload for the same application types and the same councils as the existing integration, we do _not_ need to go through the Editor hassle of adding another "destination" checkbox & updating existing flows. Instead we can simply just create two unique send events for the single destination. 
  - Alternatively, I anticipate we will have to manage Nexus as a new "destination" because it will support different application types and different councils than the existing "Uniform" integration
- Updates audit table to check if an application has been previously successfully submitted _for this version_ of the BOPS API

**Outstanding:**
- [ ] Test and confirm ODP schema validation errors bubble up as expected
- [ ] Confirm Slack notification will log for both (already includes `destination_url` which will have `v1` or `v2` respectively)
- [ ] Update Planx swagger docs
- [ ] Add BOPS `api/v2/ping` healthcheck endpoint to Uptime Robot

### DIFF
--- a/api.planx.uk/lib/hasura/metadata/index.ts
+++ b/api.planx.uk/lib/hasura/metadata/index.ts
@@ -11,6 +11,7 @@ interface ScheduledEvent {
 
 export interface CombinedResponse {
   bops?: ScheduledEventResponse;
+  bops_v2?: ScheduledEventResponse;
   uniform?: ScheduledEventResponse;
   email?: ScheduledEventResponse;
 }
--- a/api.planx.uk/modules/send/bops/bops.test.ts
+++ b/api.planx.uk/modules/send/bops/bops.test.ts
@@ -36,7 +36,7 @@ describe(`sending an application to BOPS`, () => {
       data: {
         bopsApplications: [],
       },
-      variables: { session_id: "123" },
+      variables: { session_id: "123", search_string: "%/api/v1/planning_applications" },
     });
 
     queryMock.mockQuery({
@@ -102,11 +102,104 @@ describe(`sending an application to BOPS`, () => {
           { response: { message: "Application created", id: "bops_app_id" } },
         ],
       },
-      variables: { session_id: "previously_submitted_app" },
+      variables: { session_id: "previously_submitted_app", search_string: "%/api/v1/planning_applications" },
     });
 
     await supertest(app)
       .post("/bops/southwark")
+      .set({ Authorization: process.env.HASURA_PLANX_API_KEY })
+      .send({ payload: { sessionId: "previously_submitted_app" } })
+      .expect(200)
+      .then((res) => {
+        expect(res.body).toEqual({
+          sessionId: "previously_submitted_app",
+          bopsId: "bops_app_id",
+          message: "Skipping send, already successfully submitted",
+        });
+      });
+  });
+});
+
+describe(`sending an application to BOPS v2`, () => {
+  beforeEach(() => {
+    queryMock.mockQuery({
+      name: "FindApplication",
+      data: {
+        bopsApplications: [],
+      },
+      variables: { session_id: "123", search_string: "%/api/v2/planning_applications" },
+    });
+
+    queryMock.mockQuery({
+      name: "CreateBopsApplication",
+      matchOnVariables: false,
+      data: {
+        insertBopsApplication: { id: 22 },
+      },
+    });
+  });
+
+  it.todo("throws an error if payload is invalid");
+
+  it.skip("proxies request and returns hasura id", async () => {
+    nock(`${submissionURL}/api/v2/planning_applications`).post("").reply(200, {
+      application: "0000123",
+    });
+
+    await supertest(app)
+      .post("/bops-v2/southwark")
+      .set({ Authorization: process.env.HASURA_PLANX_API_KEY })
+      .send({ payload: { sessionId: "123" } })
+      .expect(200)
+      .then((res) => {
+        expect(res.body).toEqual({
+          application: { id: 22, bopsResponse: { application: "0000123" } },
+        });
+      });
+  });
+
+  it("requires auth", async () => {
+    await supertest(app)
+      .post("/bops-v2/southwark")
+      .send({ payload: { sessionId: "123" } })
+      .expect(401);
+  });
+
+  it("throws an error if payload is missing", async () => {
+    await supertest(app)
+      .post("/bops-v2/southwark")
+      .set({ Authorization: process.env.HASURA_PLANX_API_KEY })
+      .send({ payload: null })
+      .expect(400)
+      .then((res) => {
+        expect(res.body.error).toMatch(/Missing application/);
+      });
+  });
+
+  it("throws an error if team is unsupported", async () => {
+    await supertest(app)
+      .post("/bops-v2/unsupported-team")
+      .set({ Authorization: process.env.HASURA_PLANX_API_KEY })
+      .send({ payload: { sessionId: "123" } })
+      .expect(400)
+      .then((res) => {
+        expect(res.body.error).toMatch(/not enabled for this local authority/);
+      });
+  });
+
+  it("does not re-send an application which has already been submitted", async () => {
+    queryMock.mockQuery({
+      name: "FindApplication",
+      data: {
+        bopsApplications: [
+          { response: { message: "Application created", id: "bops_app_id" } },
+        ],
+      },
+      variables: { session_id: "previously_submitted_app", search_string: "%/api/v2/planning_applications" },
+    });
+
+    await supertest(app)
+      .post("/bops-v2/southwark")
       .set({ Authorization: process.env.HASURA_PLANX_API_KEY })
       .send({ payload: { sessionId: "previously_submitted_app" } })
       .expect(200)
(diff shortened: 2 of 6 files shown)


api.planx.uk/modules/send/routes.ts
```diff
@@ -16,6 +16,7 @@ router.post(
   createSendEvents,
 );
 router.post("/bops/:localAuthority", useHasuraAuth, sendToBOPS);
+router.post("/bops-v2/:localAuthority", useHasuraAuth, sendToBOPSV2);
```
jessicamcinchak, line 19: Open to naming feedback here; also considered `/bops-odp` ? 

But ultimately, I'm hoping our overlap period for sending both is quite short & the new endpoint will simply be renamed soon to replace the existing `/bops` one. Since we only use our send routes internally, I figure renaming isn't a particular pain or inconvenience. 

Both endpoints write to the exisiting audit table `bops_applications` so no changes there.

DafyddLlyr: Naming convention here makes sense to me given it's - 
 - short term
 - can be easily updated as it's really just an internal ref

api.planx.uk/modules/send/bops/bops.ts
```diff
@@ -155,12 +289,17 @@ interface FindApplication {
  */
 async function checkBOPSAuditTable(
   sessionId: string,
+  version: string,
 ): Promise<Record<string, string>> {
+  const searchString = `%/api/${version}/planning_applications`;
```
DafyddLlyr, line 294: This is a good catch and quirk I hadn't anticipated!

I think once we're sending just one application type it would be nice to come back and remove this check also just to keep thing simple (even though it will still be valid code and work as expected).

api.planx.uk/modules/send/bops/bops.ts
```diff
@@ -155,12 +289,17 @@ interface FindApplication {
  */
 async function checkBOPSAuditTable(
   sessionId: string,
+  version: string,
```
DafyddLlyr, line 292:
```suggestion
  version: "v1" | "v2",
```

api.planx.uk/modules/send/routes.ts
DafyddLlyr: If we made `v2` a param it might be simpler to just re-use the existing function and pump in `"v2"` where needed but as this is just a short-lived thing I'm totally happy with copy/pasting the function tbh!

On the off chance that the original BOPS code will need to be updated in the next few weeks, we'll just need to remember to update both.

api.planx.uk/modules/send/bops/bops.test.ts
```diff
@@ -119,3 +125,102 @@ describe(`sending an application to BOPS`, () => {
...
+      matchOnVariables: false,
+      data: {
+        insertBopsApplication: { id: 22 },
+      },
+    });
+  });
+
+  it.todo("throws an error if payload is invalid");
+
+  it.skip("proxies request and returns hasura id", async () => {
```
DafyddLlyr, line 153: Where you hitting any issues with this test, or where you just waiting to see what the BOPS response to this should be?

api.planx.uk/lib/hasura/metadata/index.ts
```diff
@@ -11,6 +11,7 @@ interface ScheduledEvent {
 
 export interface CombinedResponse {
```
DafyddLlyr, line 12: Edge case - we might also need to think about updating logic for sending in Invite to Pay routes  - see `createPaymentSendEvents()` in `api.planx.uk/modules/pay/service/inviteToPay/createPaymentSendEvents.ts`

Ideally this would have been an exhaustive `switch` on `Destination` and this change would have flagged there - I'll make a note to tidy this up and cover this case when adding destinations once this PR has progressed and merged.

Currently an ITP submission would just go to the old BOPS endpoint. This very likely is not a problem realistically speaking.

If BOPS take a while to turn off the old v1 endpoint though we might encounter an ITP submission - I'd suggest we cover this eventuality there as it should be a very quick addition to cover this extra destination.

api.planx.uk/modules/send/bops/bops.test.ts
```diff
@@ -119,3 +125,102 @@ describe(`sending an application to BOPS`, () => {
...
+    queryMock.mockQuery({
+      name: "CreateBopsApplication",
+      matchOnVariables: false,
+      data: {
+        insertBopsApplication: { id: 22 },
+      },
+    });
+  });
+
+  it.todo("throws an error if payload is invalid");
```
DafyddLlyr, line 151: Let me know if you want another pair of eyes on this one, happy to jump on the branch and have a look.

api.planx.uk/modules/send/bops/bops.test.ts
jessicamcinchak: This test is timing out currently! It's posting an empty string/invalid payload, which I'd expect to quickly throw an error _on our side before_ getting to the POST to BOPS step - so I'm planning to investigate if a) I'm missing a mock somewhere for the `makeDigitalPlanningPayload` method in planx-core or b) errors are failing to bubble up correctly.

DafyddLlyr:
Manual testing feedback (in progress 🚀)
 - [x] Submitted LDCP to Bucks 
   - [x] Two BOPS events
   - [x] BOPS v1 succeeded
   - [ ] BOPS v2 succeeded ❌ 
     - No bubbled error logged by Airbrake
     - Session Id 1d81c0e9-e520-416e-bc61-64085172e906
     - Error - `http_exception / Connection failure`
     - Looking into it!
 - [ ] Submitted Prior Approval to Medway
   - [ ] Two BOPS events
   - [ ] BOPS v1 succeeded
   - [ ] BOPS v2 succeeded 
 - [ ] Submitted PP for Lambeth
   - [ ] Two BOPS events
   - [ ] BOPS v1 succeeded
   - [ ] BOPS v2 succeeded

DafyddLlyr:
@jessicamcinchak just a heads up that you won't be able to test PP on Medway as they're not yet setup for GovPay

jessicamcinchak:
@DafyddLlyr I'm doing resubmissions / disability exemptions to test non-GovPay councils !

DafyddLlyr:
> @DafyddLlyr I'm doing resubmissions / disability exemptions to test non-GovPay councils !

Ah brill - I must have missed that on the PP or PA flow I was testing - so used to doing it on LDCs!

api.planx.uk/modules/send/bops/bops.ts
jessicamcinchak: Yep fully agree :+1:

api.planx.uk/modules/pay/service/inviteToPay/createPaymentSendEvents.ts
```diff
@@ -66,6 +66,17 @@ const createPaymentSendEvents = async (
...
+      const isProduction = process.env.APP_ENVIRONMENT === "production";
+      if (!isProduction) {
+        const bopsV2Event = await createScheduledEvent({
+          webhook: `{{HASURA_PLANX_API_URL}}/bops-v2/${teamSlug}`,
+          schedule_at: new Date(now.getTime() + 30 * 1000),
+          payload: eventPayload,
+          comment: `bops_v2_submission_${payload.sessionId}`,
+        });
+        combinedResponse["bops_v2"] = bopsV2Event;
+      }
```
jessicamcinchak, line 79: I'm only creating the v2 events on non-production environments to match the current BOPS API functionality, this will eliminate noise of expected failures !

DafyddLlyr: Nice simple solution 👌

DafyddLlyr (approved):
Changes looking great ✅ Thanks again for super clear write up on PR - will no doubt be helpful in future.

> _Getting "BOPS not enabled for this local authority" error for both, will look into that separately_

This one is sort of odd - this should have been removed with this PR -https://github.com/theopensystemslab/planx-new/pull/2504/files#diff-f5517bb87dcf8ff9b1211115bf1f7797db89f6b570c7a8dd1ea93b797949d116

Not sure when that was merged relative to this PR last being rebased. It looks like the code from that PR is still on `main` though so it's possible I got something wrong on that branch with rebases 🤔

jessicamcinchak:
@DafyddLlyr that final error might just be me mis-remembering that Medway is set up with BOPS staging-only similar to Gloucester ?? Looking over that previous change, I'm reminded that Medway never had associated env vars!

DafyddLlyr:
@jessicamcinchak This looks like a silly GitHub mistake on my part! See https://github.com/theopensystemslab/planx-new/pull/2581

PR #2530, opened 2023-12-04
feat: Restrict certain users from production environments

DafyddLlyr:
## What does this PR do?
 - Adds an additional check on login, to ensure users can access production environment

## Why?
 As we sync the users table across environments, this means all users can access production. Generally this is what we want as "real" users live on production and this allows them to access test and staging environments - all good.

The catch is that occasionally we have a set of users who we do not wish to grant access to production - for example the user accounts associated with the recent pen test, or certain sets of users with "demo" access (until we have these roles set up). This is not a common occurrence or user type, but it feels like a sensible restriction to put in place restrictions here as a bit of a guard rail.

## TODO
- [x] Test sync script - working locally with commented out columns, will uncomment once column is on prod
- [x] Merge https://github.com/theopensystemslab/planx-core/pull/219 to `main`
- [x] Update commit refs across all projects to `planx-core` latest

### DIFF
--- a/api.planx.uk/modules/auth/service.ts
+++ b/api.planx.uk/modules/auth/service.ts
@@ -3,6 +3,7 @@ import { $api } from "../../client";
 import { User, Role } from "@opensystemslab/planx-core/types";
 
 export const buildJWT = async (email: string): Promise<string | undefined> => {
+  await checkUserCanAccessEnv(email, process.env.NODE_ENV);
   const user = await $api.user.getByEmail(email);
   if (!user) return;
 
@@ -58,4 +59,17 @@ const getAllowedRolesForUser = (user: User): Role[] => {
  */
 const getDefaultRoleForUser = (user: User): Role => {
   return user.isPlatformAdmin ? "platformAdmin" : "teamEditor";
+};
+
+/**
+ * A staging-only user cannot access production, but can access all other envs
+ */
+export const checkUserCanAccessEnv = async (
+  email: string,
+  env?: string,
+): Promise<boolean> => {
+  const isStagingOnlyUser = await $api.user.isStagingOnly(email);
+  const isProductionEnv = env === "production";
+  const userCanAccessEnv = !(isProductionEnv && isStagingOnlyUser);
+  return userCanAccessEnv;
 };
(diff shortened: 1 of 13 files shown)


api.planx.uk/modules/auth/service.ts
```diff
@@ -59,3 +60,16 @@ const getAllowedRolesForUser = (user: User): Role[] => {
...
 };
+
+/**
+ * A staging-only user cannot access production, but can access all other envs
+ */
+export const checkUserCanAccessEnv = async (
+  email: string,
+  env?: string,
+): Promise<boolean> => {
+  const isStagingOnlyUser = await $api.user.isStagingOnly(email);
```
DafyddLlyr, line 71: **Why is this it it's own function and not a field added to `user.getByEmail()`**

I appreciate that this does mean an additional query but the `user.isStagingOnly` property is only ever needed here once on login.

If we added this to the more general query, we'd be polluting our `User` type with a meaningless property that has no value outside of this context.

This is one of those cases where the full database model will not match the model we want in-application.

scripts/seed-database/write/users.sql
```diff
@@ -7,6 +7,8 @@ CREATE TEMPORARY TABLE sync_users (
   created_at timestamptz,
   updated_at timestamptz,
   is_platform_admin boolean
+  -- ,
+  -- is_staging_only boolean
```
jessicamcinchak, line 11: are these lines intended to be comment-out in sync script?

api.planx.uk/modules/auth/service.ts
jessicamcinchak: Makes sesne to me, thanks for explanation!

jessicamcinchak (approved):

scripts/seed-database/write/users.sql
DafyddLlyr: Yep - it's a bit of chicken and egg situation - we can't sync from the new column until it exists on prod.

Once this is across there'll be a quick PR uncommenting these lines (plus any potential sync script fixes as it's currently untestable).

DafyddLlyr: (And the catch here is that pizzas will fail unless it's commented out as they can't sync)

PR #2531, opened 2023-12-04
fix: Input focus styles for editor and public UI

ianjon3s:
# What does this PR do?

The recent changes applied to the editor introduced a small visual bug on the public-facing UI where a rich text input shifts 1px in each direction when focused. The PR introduces a default input focus style to the theme and extends it to be applied only to bordered inputs.

Live example (enter text and focus/unfocus to see layout shift):
https://editor.planx.uk/testing/rich-text-test/preview?analytics=false

Fixed example:
https://2531.planx.pizza/testing/rich-text-test/preview?analytics=false

### DIFF
--- a/editor.planx.uk/src/theme.ts
+++ b/editor.planx.uk/src/theme.ts
@@ -70,13 +70,19 @@ export const focusStyle = {
   outline: "3px solid transparent",
 };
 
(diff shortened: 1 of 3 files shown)


DafyddLlyr (approved):
Eagle eyes!

Fix is perfect thanks 🏅

PR #2532, opened 2023-12-04
fix: Assign missing A4 category

DafyddLlyr:
Just spotted this when testing - it looks like category for article 4s are not being returned by API so the UI is defaulting to `undefined`.

I don’t think there have really been any recent changes here that could have caused this - there might be another solution (or cause!) I'm missing here so this might be a naive fix...!

Waiting on Pizza to test changes 🚀 

image

Original request URL (undefined): 
https://api.2529.planx.pizza/gis/southwark?geom=POLYGON+%28%28-0.0807095109084621+51.50357969225416%2C+-0.08026678402163087+51.50357732850256%2C+-0.0803878763290122+51.50339599741571%2C+-0.0807095109084621+51.50357969225416%29%29&analytics=false

### DIFF
--- a/api.planx.uk/modules/gis/service/digitalLand.ts
+++ b/api.planx.uk/modules/gis/service/digitalLand.ts
@@ -230,6 +230,7 @@ async function go(
           value: true,
           text: baseSchema["article4"].pos,
           data: formattedResult[localCaz].data,
+          category: baseSchema["article4"].category,
         };
       }
     }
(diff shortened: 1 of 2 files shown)


DafyddLlyr:
On second thoughts - it looks like this will just give us a second "General Policy" list here?

jessicamcinchak:
Good spot - this is a weird one! 

I think it's a CAZ edge case: it looks like the "category" is correctly assigned to `article4.southwark.caz`, but then when that's "rolled up" to its' parent `article4` the category has been dropped? I haven't checked out the branch yet, but might be in an issue with the rollup method if this solution isn't working as expected yet?

CAZ are strangely/uniquely handled because they are a "child" granular passport variable of A4 heirarchy, but fetch from from a unique "dataset" different than A4.

DafyddLlyr:
Fix is working as I'd expect - the frontend is actually grouping these by category 

image

jessicamcinchak (approved):
Thank you!

PR #2485, opened 2023-11-27
fix: remove sessionId from URLs

Mike-Heneghan:
## What: 

- On saving an application remove the feature which stored the sessionId in the url
- When a user clicks a link to resume a session, read the sessionId but then remove it from the params

## Why:

- Exposing the sessionId can have security implications

PR #2533, opened 2023-12-04
Revert "fix: remove sessionId from URLs"

DafyddLlyr:
Reverts theopensystemslab/planx-new#2485

From Jumpsec - 

> Regarding the issue with a session value being contained within GET parameters, I believe this should be fine and I have removed it from the findings (due to an email address being required for further authorisation).

Reverting this restores the `allowResumeOnBrowserRefresh()` functionality which I didn't really have a decent / simple fix in mind for 👍 

Related change https://github.com/theopensystemslab/planx-new/pull/2494 is good to stay in place as this is still a valid alternate method of grabbing the `sessionId` value.

### DIFF
--- a/editor.planx.uk/src/pages/Preview/ResumePage.tsx
+++ b/editor.planx.uk/src/pages/Preview/ResumePage.tsx
@@ -16,7 +16,6 @@ import { ApplicationPath, SendEmailPayload } from "types";
 import Input from "ui/Input";
 import InputLabel from "ui/InputLabel";
 import InputRow from "ui/InputRow";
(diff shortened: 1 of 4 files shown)


jessicamcinchak (approved):
Really pleased this is safe to stay, was realising how much I rely on resuming on refresh or being able to quickly grab id from URL when testing submissions yesterday :slightly_smiling_face:

DafyddLlyr:
> Was realising how much I rely on resuming on refresh or being able to quickly grab id from URL when testing submissions yesterday

I know right - me too. Hit this so many times in just a few days, really pleased to revert it.

PR #2535, opened 2023-12-04
fix(style): ITP button underline

DafyddLlyr:
Another quick fix I've been meaning to get round to since a while - 

| Before | After |
|--------|--------|
| image |image | 

This also now more closely aligns to GDS style in that the underline / box shadow is retained on hover - https://design-system.service.gov.uk/components/button/


ianjon3s (approved):
Thanks for spotting and picking this up!

PR #2536, opened 2023-12-06
fix: Allow `target="_blank"` in anchors

DafyddLlyr:
Reason DOMPurify strips this - https://developer.chrome.com/docs/lighthouse/best-practices/external-anchors-use-rel-noopener/

> When you link to a page on another site using the target="_blank" attribute, you can expose your site to performance and security issues:
>
> The other page may run on the same process as your page. If the other page is running a lot of JavaScript, your page's performance may suffer.
>
> The other page can access your window object with the window.opener property. This may allow the other page to redirect your page to a malicious URL.
> Adding `rel="noopener"` or `rel="noreferrer"` to your `target="_blank"` links avoids these issues.

Discussion / solution found here: https://github.com/cure53/DOMPurify/issues/317

### DIFF
--- a/api.planx.uk/modules/flows/findReplace/controller.ts
+++ b/api.planx.uk/modules/flows/findReplace/controller.ts
@@ -26,7 +26,9 @@ export const findAndReplaceSchema = z.object({
     replace: z
       .string()
       .optional()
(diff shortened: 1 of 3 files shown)


jessicamcinchak (approved):

PR #2540, opened 2023-12-06
fix: drop pg function `get_flow_schema` in favor of simpler js data munging

jessicamcinchak:
Refactors the `/flows/:flowId/download-schema` endpoint to fetch & transform flow data directly in our API rather than via PostgreSQL function which would historically timeout for large flows.

This is an edge-casey utility endpoint exposed in the Editor debug console only. I've been using it a lot the last couple days to quickly return all the passport variables a given flow sets while fixing `proposal.projectType` validation issues / missing data dictionary values over here: https://github.com/theopensystemslab/digital-planning-data-schemas/pull/88

Lots of room to think about a more meaningful return data shape in the future as we move towards actual in-editor schema validation, but this simplification & db migration feel like a fresher starting point.

### DIFF
--- a/api.planx.uk/modules/flows/downloadSchema/controller.ts
+++ b/api.planx.uk/modules/flows/downloadSchema/controller.ts
@@ -6,8 +6,6 @@ import { ServerError } from "../../../errors";
 
 interface DownloadFlowSchemaResponse {
   message: string;
-  alteredNodes: Node[] | null;
-  description?: string;
 }
 
 export const downloadFlowSchema = z.object({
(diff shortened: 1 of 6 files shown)


api.planx.uk/modules/flows/downloadSchema/controller.ts
```diff
@@ -6,8 +6,6 @@ import { ServerError } from "../../../errors";
 
 interface DownloadFlowSchemaResponse {
```
DafyddLlyr, line 7: nit(docs): I think this change will mean that the YAML also needs to change.

DafyddLlyr (approved):
Super simple and meaningful change - CSV generation is rapid! ⚡

DafyddLlyr:
Also: we should use "data munging" more" 😂

PR #2544, opened 2023-12-07
feat: Add extra long text input (500 characters)

DafyddLlyr:
## What does this PR do?
- Adds a new input type, which allows 500 characters to be inputted 

image

### DIFF
--- a/editor.planx.uk/src/@planx/components/TextInput/TextInput.stories.tsx
+++ b/editor.planx.uk/src/@planx/components/TextInput/TextInput.stories.tsx
@@ -16,6 +16,7 @@ export default {
       options: [
         TextInputType.Short,
         TextInputType.Long,
+        TextInputType.ExtraLong,
         TextInputType.Email,
         TextInputType.Phone,
       ],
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/@planx/components/TextInput/TextInput.stories.tsx
```diff
@@ -33,11 +34,19 @@ export const EmptyFormShortText: StoryObj = {
...
   },
 };
 
+export const EmptyFormExtraLongText: StoryObj = {
+  args: {
+    title: "Describe your project",
+    description: "Be as descriptive as you can. You have 500 characters.",
+    type: TextInputType.ExtraLong,
+  },
+};
```
jessicamcinchak, line 48: Perfect example desription :+1:

editor.planx.uk/src/@planx/components/TextInput/model.ts
```diff
@@ -47,6 +51,9 @@ export const userDataSchema = (type?: TextInputType): SchemaOf<UserData> =>
         if (type === TextInputType.Long) {
           return Boolean(value && value.length < 250);
         }
+        if (type === TextInputType.ExtraLong) {
+          return Boolean(value && value.length < 500);
```
jessicamcinchak, line 55: nit: should we technically be using `<=` in these checks to match error message text?

PR #2545, opened 2023-12-07
test: Cover HTML anchor target attribute stipping

DafyddLlyr:
## What does this PR do?
- Adds test coverage for changes made in this PR - https://github.com/theopensystemslab/planx-new/pull/2536
- Very much imperfect - we'd need E2E tests for all scenarios to be covered, but my main aim here is to write a test that fails if DOMPurify is replaced / removed to the config updated in a breaking way


jessicamcinchak (approved):
Thanks for picking up this first piece!

PR #2548, opened 2023-12-08
chore: add staging file API key for Nexus integration

jessicamcinchak:
Adds a `FILE_API_KEY_NEXUS` variable for **staging-only** so that the Idox team can fully begin developing against the example ODP Schema payloads, including downloading the staging file links.

Local variables for docker pushed up to S3, please run `/scripts/pull-secrets.sh` before tesitng locally. Pizza will pull down on build.

Next steps:
- Merge to staging & confirm ODP example payload files (with `api.editor.planx.dev` links) can be downloaded successfully using this token in header
- Share token securely with Idox; noting we'll plan to rotate & issue prod key before "go live" early next year


DafyddLlyr (approved):
Tested via API docs - working as expected! ✅

PR #2549, opened 2023-12-08
chore: Update Hasura to v2.36.0

DafyddLlyr:
## What?
- Update Hasura from v2.8.4 → v.2.36.0
- Automated updates `tables.yaml` and `function.yaml`
  - I added/removed a permission (so no changes) which triggered an update to alphabetise these files
  - Without this step the next PR that touches the file gets very noisy
  - This also removed 5 instances of `backend_only: false` which is redundant as this is the default value

## Why?
- It's been a while! Last update was July 2022.
- There are a few features I'd love to pick up - inherited roles could be helpful (and we've previously discussed this), and I'm looking to implement [Postgres input validation](https://hasura.io/docs/latest/schema/postgres/input-validations/) to tackle recent sanitation questions

## Testing
- Testing locally seems fine ✅ 
- Testing on Pizza 🚀 
- Regression tests on branch ✅

### DIFF
--- a/hasura.planx.uk/Dockerfile
+++ b/hasura.planx.uk/Dockerfile
@@ -1,4 +1,4 @@
-FROM hasura/graphql-engine:v2.8.4.cli-migrations-v2
+FROM hasura/graphql-engine:v2.36.0.cli-migrations-v2
 WORKDIR /
 COPY metadata hasura-metadata
 COPY migrations hasura-migrations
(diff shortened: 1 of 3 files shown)


jessicamcinchak (approved):
All working as expected for me on pizza, thanks for taking care of this one !

Mike-Heneghan (approved):
LGTM

PR #2551, opened 2023-12-11
feat: Hasura input validation for tables storing user-submitted HTML

DafyddLlyr:
## What does this PR do?
 - Attempts to resolve the issue of sanitation of user-submitted HTML in a generalised method using [Hasura input validation](https://hasura.io/docs/latest/schema/postgres/input-validations/)
 - Adds a new API endpoint to handle input validation
 - Adds validation to `global_settings` and `flows` tables (this is not available on a per-column basis)
 - Adds very basic frontend feedback (a toast message)
 
Apologies for such a big PR... 🙏 

## How does this work?
Essentially, prior to opening a database transaction, Hasura makes a request to an endpoint which can run business logic to validate a payload.

**This endpoint does not modify the submitted payload, it merely accepts or rejects it. This is different to the previous approach of sanitising and accepting changes.**

Please see diagram below which demonstrates the data flow - 

```mermaid
sequenceDiagram
  participant PlanX as PlanX Editor
  participant Hasura as Hasura GraphQL API
  participant API as PlanX API
  participant Middleware as Validation Middleware
  participant Controller as Controller
  participant DB as PostgreSQL

  PlanX ->> Hasura: User-submitted content
  activate Hasura

  Hasura ->> API: Input validation request
  activate API

  activate API
  API ->> Middleware: Validate against Zod schema
  activate Middleware
  alt Valid
    Middleware ->> Controller: Transformed payload
    activate Controller
    Controller ->> Hasura: 200 OK
    deactivate Controller
    Hasura ->> DB: Write to Postgres
    activate DB
    DB ->> Hasura: Success
    deactivate DB
    Hasura ->> PlanX: 200 OK (Empty)
  else Invalid
    Middleware ->> Controller: Transformed payload
    deactivate Middleware
    activate Controller
    Controller ->> Hasura: 400 Bad Request
    deactivate Controller
  deactivate API
    Hasura ->> PlanX: 400 Bad Request (Invalid HTML content)
  end

  deactivate Hasura
```

### Why input validation? Why not Hasura actions?
This approach is fairly re-usable and generalisable so it can be used, re-used, and expanded upon. 

Our initial crack at this in https://github.com/theopensystemslab/planx-new/pull/2488 is pretty specific to implementation and has a few gotchas / edge cases / footguns around permissions which are covered by comments etc. If we needed to do this for multiple tables things could get quite messy.

This approach means there's a single bit of logic for user-submitted HTML which can be used by multiple tables.

The real catch here is that it's just validation - it doesn't sanitise the input and proceed. Realistically this is likely fine for unsafe HTML!

I had initially considered [Hasura Actions](https://hasura.io/docs/latest/actions/overview/) as a viable method as this could handle transforming data and then writing to the database but I ultimately decided against it as - 
 - It would require some custom code per-table
 - We don't actually need sanitised HTML in the DB, rejecting it is fine
 - It just seemed more complex all around

## Testing
https://github.com/theopensystemslab/planx-new/assets/20502206/cc544aec-dce5-4e3b-b440-270e691fded7
- Updates to `global_settings` works for clean content, is rejected for unsafe content
- Updates to `flows` works for clean content, is rejected for unsafe content (`flow.data` still updates via ShareDB)
- The same validation check happens via the GraphQL API directly

## Next steps...
 - A basic E2E test or two (another PR)
 - Revert previous implementation in https://github.com/theopensystemslab/planx-new/pull/2488 (another PR)
 - Merge https://github.com/theopensystemslab/planx-new/pull/2549 and rebase to remove diff noise in `tables.yaml`

image


api.planx.uk/modules/webhooks/routes.ts
```diff
@@ -53,6 +55,12 @@ router.post(
   sanitiseApplicationDataController,
 );
 
+router.post(
+  "/webhooks/hasura/validate-input/jsonb/clean-html",
```
DafyddLlyr, line 59: Open to thoughts on the URL path here - I figured having a column type parameter would handle us possibly expanding this in future, e.g - 

 - `/webhooks/hasura/validate-input/text/clean-html`
 - `/webhooks/hasura/validate-input/number/something`

api.planx.uk/modules/webhooks/service/validateInput/utils.test.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    isObjectValid(emptyObject, mockValidator);
+    isObjectValid(emptyArray, mockValidator);
+
+    expect(mockValidator).not.toHaveBeenCalled();
+  });
+});
+
+describe("isCleanHTML() helper function", () => {
+  const dirtyHTML = [
```
DafyddLlyr, line 84: Examples taken from https://github.com/cure53/DOMPurify?tab=readme-ov-file#some-purification-samples-please

api.planx.uk/modules/webhooks/service/validateInput/utils.ts
```diff
@@ -0,0 +1,45 @@
...
+ *   - true for valid data
+ *   - false for invalid data
+ */
+type Validator = (input: unknown) => boolean;
+
+/**
+ * Recursively iterate over an object, checking is validator callback condition is met
+ * Fails fast - will return false once first invalid value is encountered
+ */
+export const isObjectValid = (
```
DafyddLlyr, line 20: I was initially hoping to just use some lodash method to get all nested values from an object so I didn't have to write this (or tests for it!). I don't think there is a method for this however - a few requests on the lodash repo suggest using [flat](https://www.npmjs.com/package/flat) but a whole library for this seems a bit excessive.

If there's a simpler library method or approach for this that I'm just missing please let me know 👍

api.planx.uk/modules/webhooks/service/validateInput/schema.ts
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Schema which iterates over values of a JSONB column
+ * Checks using DOMPurify to ensure that user-submitted HTML is clean
+ * Fails fast - will reject on first instance of unclean HTML
+ */
+export const isCleanJSONBSchema: z.ZodType<
+  IsCleanJSONBSchema,
+  z.ZodTypeDef,
+  HasuraValidateInputRequestSchema
+> = hasuraValidateInputRequestSchema.transform((original) => {
```
DafyddLlyr, line 37: Using `.transform()` to join two schemas requires this pretty complex type.

Docs: https://zod.dev/?id=zodtype-with-zodeffects

api.planx.uk/modules/webhooks/service/validateInput/schema.ts
```diff
@@ -0,0 +1,50 @@
...
+  typeof hasuraValidateInputRequestSchema
+>;
+
+interface IsCleanJSONBTransformedRequest {
+  body: {
+    isClean: boolean;
+  };
+}
+
+type isCleanJSONBSchema = z.ZodType<
```
DafyddLlyr, line 28: Using .transform() to join two schemas requires this pretty gnarly type.

Docs: https://zod.dev/?id=zodtype-with-zodeffects

api.planx.uk/modules/webhooks/service/validateInput/schema.ts
```diff
@@ -0,0 +1,50 @@
...
+  IsCleanJSONBRequest,
+  z.ZodTypeDef,
+  HasuraValidateInputRequest
+>;
```
DafyddLlyr, line 32: Using .transform() to join two schemas requires this pretty gnarly type.

Docs: https://zod.dev/?id=zodtype-with-zodeffects

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -290,6 +290,15 @@
           - id
           - version
           - data
+        validate_input:
```
DafyddLlyr, line 293: These are repeated per table, per permission.

Now that we've updated Hasura it might be worth re-looking at [inherited roles](https://hasura.io/docs/latest/auth/authorization/inherited-roles/).

api.planx.uk/modules/webhooks/service/validateInput/utils.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    });
+  }
+
+  const cleanHTML = [
+    `<h1>Hello, World!</h1><p>This is simple HTML.</p>`,
+    `<ul><li>Item 1</li><li>Item 2</li><li>Item 3</li></ul>`,
+    `<p>This is <b>bold</b> text and <i>italic</i> text.</p>`,
+    `<div><p>This content is inside a div.</p></div>`,
+    `<h2>Subheading</h2><p>This is a paragraph under the subheading.</p>`,
+    `<div><h1>Main Title</h1><p>This is a <b>nested</b> paragraph with a <i>nested</i> element.</p></div>`,
```
jessicamcinchak, line 105: Based on what we know about Planx content / rich text editing - should we have a clean example with a valid `` tag too just to be safe?

api.planx.uk/modules/webhooks/service/validateInput/utils.ts
jessicamcinchak: This still feels readable to me, agree a whole library feels like overkill at this stage!

jessicamcinchak (approved):
This all looks good to me - really nice to see immediate use cases for these new features like input validation :tada:

api.planx.uk/modules/webhooks/service/validateInput/utils.test.ts
DafyddLlyr: This is a really good shout - test case added 👍

PR #2554, opened 2023-12-12
fix: Editor layout for small screens

ianjon3s:
# What does this PR do?

The current editor layout becomes unusable at smaller screen sizes due to columns and indentation. This PR introduces small changes to make it somewhat more usable. Also removes a rogue 25px hardcoded size for inputs that was only showing up at smaller breakpoints.

**Before (left) & after (right):**
image


DafyddLlyr (approved):
Nice - much better ✅

PR #2556, opened 2023-12-12
feat: add basic metadata to GOV.UK Pay requests

jessicamcinchak:
Short & simple as this is a bit of an exploratory task still at this point - we'll see if any councils notice & can extend the `metadata` type as necesary based on any future requests !

### DIFF
--- a/api.planx.uk/modules/pay/middleware.ts
+++ b/api.planx.uk/modules/pay/middleware.ts
@@ -128,7 +128,7 @@ export async function buildPaymentPayload(
     metadata: {
       source: "PlanX",
       flow:
-        req.query.returnURL.toString().split("?")?.[0]?.split("/").pop() ||
+        new URL(req.query.returnURL as string).pathname.split("/")[0] ||
         (req.query.returnURL as string),
       inviteToPay: true,
     },
--- a/api.planx.uk/modules/pay/middleware.ts
+++ b/api.planx.uk/modules/pay/middleware.ts
@@ -90,6 +90,11 @@ interface GovPayCreatePayment {
   reference: string;
   description: string;
   return_url: string;
+  metadata?: {
+    source: "PlanX";
+    flow: string;
+    inviteToPay: boolean;
+  };
 }
 
 export async function buildPaymentPayload(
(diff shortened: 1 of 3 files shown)


jessicamcinchak:
[Enrich GovPay metadata with service details](https://trello.com/c/JPmBeE0Z/2688-enrich-govpay-metadata-with-service-details)

api.planx.uk/modules/pay/middleware.ts
```diff
@@ -120,6 +125,13 @@ export async function buildPaymentPayload(
     reference: req.query.sessionId as string,
     description: "New application (nominated payee)",
     return_url: req.query.returnURL as string,
+    metadata: {
+      source: "PlanX",
+      flow:
+        req.query.returnURL.toString().split("?")?.[0]?.split("/").pop() ||
+        (req.query.returnURL as string),
```
jessicamcinchak, line 132: An example `returnURL` looks like this: https://planningservices.lambeth.gov.uk/apply-for-a-lawful-development-certificate?sessionId={uuid}&email={email}

So, this parsing: 1) Gets the URL without query params, then 2) gets the last segment (will work regardless of custom subdomain or not!)

DafyddLlyr: Doing this without a query is the way to go!

I think I'd be tempted to use the URL class (`url.pathname.split("/")[0]`) just for a bit more safety here.

editor.planx.uk/src/@planx/components/Pay/model.ts
```diff
@@ -63,6 +67,11 @@ export const createPayload = (
   reference,
   description: "New application",
   return_url: getReturnURL(reference),
+  metadata: {
```
DafyddLlyr, line 70: I think this might be redundant if we're just appending it in the API anyway - it might be simpler just to handle this once server-side.

DafyddLlyr (approved):
Nice and simple! A few small comments, no show stoppers 👌

editor.planx.uk/src/@planx/components/Pay/model.ts
jessicamcinchak: The change to `middleware.ts` actually only impacts invite-to-pay payment requests ! So this change to the frontend model is the _only_ place it's handled for all "regular" payments (confusingly so! there's more duplication and separation in the invite to pay work than I remembered - one to revisit someday!)

api.planx.uk/modules/pay/middleware.ts
jessicamcinchak: Good shout, updated !

editor.planx.uk/src/@planx/components/Pay/model.ts
DafyddLlyr: Ah good catch - thanks for clarifying that. Yep, one to tidy up at some point!

PR #2558, opened 2023-12-12
test: Add basic tests describing `isEqual()` function

DafyddLlyr:
Follow on from https://github.com/theopensystemslab/planx-new/pull/2557 as I was unsure how the API would handle an undefined API key.

Turns out it handles it just fine and an error is thrown 👍 

It will still match on `"" === ""` which isn't wrong but is very likely not going to be a valid check in the context we use `isEqual()` - I've added a guard for this. If you think this is redundant please let me know....!

This function already has coverage from a number other test suites (e.g. `api.planx.uk/modules/file/file.test.ts` and `api.planx.uk/modules/send/bops/bops.test.ts`) but a few unit tests just explaining how it handles these edge cases is a bit more reassuring I think!

### DIFF
--- a/api.planx.uk/modules/auth/middleware.ts
+++ b/api.planx.uk/modules/auth/middleware.ts
@@ -16,7 +16,10 @@ export const userContext = new AsyncLocalStorage<{ user: Express.User }>();
 /**
  * Validate that a provided string (e.g. API key) matches the expected value
  */
-const isEqual = (provided = "", expected: string): boolean => {
+export const isEqual = (provided = "", expected: string): boolean => {
+  // Reject test against falsey values - could indicate unset secret
+  if (!expected) return false;
+
   const hash = crypto.createHash("SHA512");
   return crypto.timingSafeEqual(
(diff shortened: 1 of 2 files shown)


api.planx.uk/modules/auth/middleware.ts
```diff
@@ -16,7 +16,9 @@ export const userContext = new AsyncLocalStorage<{ user: Express.User }>();
 /**
  * Validate that a provided string (e.g. API key) matches the expected value
  */
-const isEqual = (provided = "", expected: string): boolean => {
+export const isEqual = (provided?: string, expected?: string): boolean => {
+  if (!provided || !expected) throw Error("Values must be set for isEqual()");
```
DafyddLlyr, line 20: An error is thrown without this addition by `hash.copy().update()` which requires a string.

Setting this here just allows us a slightly nicer error message and we can pass through `string | undefined` as an argument (e.g. `req.headers.authorization` or `process.env.SOMETHING`).

jessicamcinchak (approved):

PR #2561, opened 2023-12-13
feat: fetch title boundary to autofill draw boundary

jessicamcinchak:
https://2561.planx.pizza/testing/title-boundaries/preview

**Key changes:**
- FindProperty now fetches `title-boundary` dataset in addition to `local-authority-district` & `region`
- PropertyInformation reference map displays static "blue line" title boundary. Visually hidden text is also updated
- DrawBoundary interactive map pre-prepopulates with the title boundary and its' calculate area. Visually hidden text is also updated
- All maps support multiple attributions, and display a second attribution for the title boundary if pre-populated
- Adds a passport variable `drawBoundary.action` to track the user action, with possible values: `Accepted the title boundary`, `Ammended the title boundary`, `Drew a custom boundary`, or `Uploaded a location plan`. 
  - Adds this passport variable to analytics `ALLOW_LIST`
  - We can similarly append it to the GeoJSON properties or elsewhere in submission payloads later

**Open questions for August/content team:**
- Do we want to update the default content for the `DrawBoundary` component to better reflect new default auto-populating behavior? Existing components would still need to be updated to reflect?
- Content on above test flow is already on production, but ideally needs to be rolled out to all services using DrawBoundary as automatically as possible - could also think about a content migration script here again to directly update latest published flows of any DrawBoundary node types?

 **What I'm not doing here:**
 - Any type of conditional rules for if/when to pre-populate the red line (we're keeping it generic as possible to start, and have growing good evidence that complex cases most likely to need ammending will be completed by professionals anways)
 - No changes to input fields or required props in the Editor components. Again, this means auto-populating behavior will always be "on" by default - it is not able to be toggled within the component
 
**Follow-up PRs:** 
- Adjust `planx-core` passport mapping for the digital planning application schema for `property.site.boundary` (original title boundary) & `proposal.site.boundary` (location plan boundary)


jessicamcinchak:
[Design Auto-red line journey](https://trello.com/c/DZtkdTOd/2585-design-auto-red-line-journey)

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -68,9 +70,10 @@ function Component(props: Props) {
   });
   options.append("dataset", "local-authority-district");
   options.append("dataset", "region"); // proxy for Greater London Authority (GLA) boundary
+  options.append("dataset", "title-boundary")
 
   // https://www.planning.data.gov.uk/docs#/Search%20entity
-  const root = `https://www.planning.data.gov.uk/entity.json?`;
+  const root = `https://www.planning.data.gov.uk/entity.geojson?`;
```
jessicamcinchak, line 76: By changing this request to return geojson, instead of json with a WKT `geometry` field, we can save ourselvs a WKT to GeoJSON conversion and store the title boundary as GeoJSON directly in the passport.

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -204,6 +211,10 @@ function Component(props: Props) {
           if (regions) {
             newPassportData["property.region"] = regions;
           }
+          if (titleBoundary) {
+            newPassportData["property.boundary.title"] = titleBoundary;
+            newPassportData["property.boundary.title.area"] = Number(area(titleBoundary as any).toFixed(4));
```
jessicamcinchak, line 216: todo: align rounding & units to `formatArea` helper method in map repo https://github.com/theopensystemslab/map/blob/45fa7284c7b0f3d15b2aadb69d64807d42749c77/src/components/my-map/utils.ts#L21

passport should store area of title boundary as both square metres & hectares similar to site boundary?

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -204,6 +213,14 @@ function Component(props: Props) {
           if (regions) {
             newPassportData["property.region"] = regions;
           }
+          if (titleBoundary) {
+            const areaSquareMetres =
+              Math.round(area(titleBoundary as FeatureCollection) * 100) / 100;
+            newPassportData["property.boundary.title"] = titleBoundary;
+            newPassportData["property.boundary.title.area"] = areaSquareMetres;
+            newPassportData["property.boundary.title.area.hectares"] =
+              areaSquareMetres / 10000;
```
jessicamcinchak, line 222: These passport variable names and area formatting are consistent with draw calculations handled here https://github.com/theopensystemslab/map/blob/45fa7284c7b0f3d15b2aadb69d64807d42749c77/src/components/my-map/utils.ts#L21

editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
```diff
@@ -12,36 +12,41 @@ import QuestionHeader from "@planx/components/shared/Preview/QuestionHeader";
...
 
 export type Props = PublicProps<DrawBoundary>;
 
-export type Boundary = GeometryObject | undefined;
+export type Boundary = Feature | undefined;
 
 // Buffer applied to the address point to clip this map extent
 //   and applied to the site boundary and written to the passport to later clip the map extent in overview documents
-const BUFFER_IN_METERS = 75;
+const BUFFER_IN_METERS = 100;
```
jessicamcinchak, line 29: This better accomodates large default boundaries - I think this is still a reasonable distance for our OS licensing concerns?

editor.planx.uk/src/@planx/components/DrawBoundary/model.ts
```diff
@@ -1,5 +1,12 @@
 import { MoreInformation, parseMoreInformation } from "../shared";
 
+export enum DrawBoundaryUserAction {
+  Accept = "Accepted the title boundary",
+  Ammend = "Ammended the title boundary",
+  Draw = "Drew a custom boundary",
+  Upload = "Uploaded a location plan",
+}
```
jessicamcinchak, line 8: Open to language for these!

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -5,8 +5,10 @@ import Typography from "@mui/material/Typography";
 import Card from "@planx/components/shared/Preview/Card";
 import QuestionHeader from "@planx/components/shared/Preview/QuestionHeader";
 import { PublicProps } from "@planx/components/ui";
-import { GeoJSONObject } from "@turf/helpers";
-import { useStore } from "pages/FlowEditor/lib/store";
+import area from "@turf/area";
```
DafyddLlyr, line 8: Feel like I'm maybe just missing something here, but presumably we can't just get the area from the `areaChangeHandler` if there's a pre-loaded GeoJSON? Is this just throwing an event on change of the boundary?

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
DafyddLlyr: nit: See comment above about needing an alternate way to calculate area - ideally it would be nice if this could be handled just once.

Currently it's the responsibility of the map component for `property.boundary.area` but PlanX for `property.boundary.title.area` which feels a bit inconsistent.

DafyddLlyr (approved):
A few small non-blocking questions 👍

editor.planx.uk/src/@planx/components/DrawBoundary/model.ts
```diff
@@ -1,5 +1,12 @@
 import { MoreInformation, parseMoreInformation } from "../shared";
 
+export enum DrawBoundaryUserAction {
+  Accept = "Accepted the title boundary",
+  Ammend = "Ammended the title boundary",
```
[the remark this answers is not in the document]
jessicamcinchak: Updated, thanks for spotting!

editor.planx.uk/src/@planx/components/PropertyInformation/Public.tsx
```diff
@@ -26,13 +27,19 @@ import type { PropertyInformation } from "./model";
...
-      state.overrideAnswer,
-    ]);
+  const [
+    address,
+    propertyType,
+    localAuthorityDistrict,
+    titleBoundary,
+    overrideAnswer,
+  ] = useStore((state) => [
+    state.computePassport().data?._address,
```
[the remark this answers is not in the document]
jessicamcinchak: Updated!

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
jessicamcinchak: Conversion from sq m to ha now added to utils as a helper function!

DafyddLlyr (approved):

PR #2567, opened 2023-12-14
chore: Don't error on unused vars in Editor

DafyddLlyr:
Been meaning to fix this for a while, current behaviour ends up being pretty painful when developing!


jessicamcinchak (approved):
:100:

PR #2568, opened 2023-12-14
feat(wip): Don't track analytics for auto-answered nodes

DafyddLlyr:


editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -126,6 +127,16 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
       const curLength = Object.keys(breadcrumbs).length;
       const prevLength = Object.keys(previousBreadcrumbs).length;
 
+      const lastBreadcrumbNodeId = Object.keys(breadcrumbs)[curLength - 1];
```
DafyddLlyr, line 130: Key order is guaranteed to match insertion order for (for string keys) beyond ES6 so this feels ok to me?

I think using an array or map would be a better idea for breadcrumbs and give us more safety / better DX in a few places longer term.

PR #2571, opened 2023-12-15
fix: decode organisation details from `access_token` for Uniform authentication

jessicamcinchak:
We had a failed Southwark submission this morning and found out that there was a new release of the Submissions API Tuesday that changes the response shape of our authentication endpoint. 

This change addresses that update, and we'll figure out a process for more pro-active communication and documentation with Idox going forward.

Regression tests failing against this branch but passing locally, going to pick up fix outside of this PR: https://github.com/theopensystemslab/planx-new/actions/runs/7222293774

### DIFF
--- a/api.planx.uk/modules/send/uniform/uniform.ts
+++ b/api.planx.uk/modules/send/uniform/uniform.ts
@@ -1,11 +1,12 @@
 import axios, { AxiosRequestConfig, isAxiosError } from "axios";
 import { NextFunction, Request, Response } from "express";
-import { Buffer } from "node:buffer";
 import FormData from "form-data";
 import fs from "fs";
-import { markSessionAsSubmitted } from "../../saveAndReturn/service/utils";
 import { gql } from "graphql-request";
+import jwt from "jsonwebtoken";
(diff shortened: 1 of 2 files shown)


DafyddLlyr (changes requested):
Just a heads up - as the behavior of the Uniform auth server has change we'll need to make a corresponding change to `e2e/tests/api-driven/src/invite-to-pay/mocks/server-mocks.yaml`

First thought would be to make a hardcoded JWT using something like https://jwt.io/ (changing the payload to be a minimal example match the JWT you have locally which returns the two required values only).

DafyddLlyr:
I think the string `eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJvcmdhbmlzYXRpb24tbmFtZSI6Ik1PQ0tFRCIsIm9yZ2FuaXNhdGlvbi1pZCI6Ik1PQ0tFRCJ9.p0DE8MUc9obE751XWOYPQWWtLXtq8-kJMPre4VuOBHg` will work.

image

DafyddLlyr (approved):

PR #2576, opened 2023-12-18
feat: add & populate column `analytics_logs.node_id`

jessicamcinchak:
Changes:
- Alters table `analytics_logs` to add new column `node_id`, populates value on record creation
- Updates view `analytics_summary` to add new column `node_id`

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
@@ -155,12 +155,15 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
       node?.type === TYPES.Content
         ? getContentTitle(node)
         : node?.data?.title ?? node?.data?.text ?? node?.data?.flagSet;
+    const nodeId = node?.id || "Unknown";
+
     // On component transition create the new analytics log
     const result = await insertNewAnalyticsLog(
       direction,
       analyticsId,
       metadata,
       nodeTitle,
+      nodeId,
     );
     const id = result?.data.insert_analytics_logs_one?.id;
     const newLogCreatedAt = result?.data.insert_analytics_logs_one?.created_at;
@@ -179,6 +182,7 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     analyticsId: number,
     metadata: NodeMetadata,
     nodeTitle: string,
+    nodeId: string,
   ) {
     const result = await publicClient.mutate({
       mutation: gql`
@@ -188,7 +192,7 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
           $metadata: jsonb
           $node_type: Int
           $node_title: String
-          $user_agent: jsonb
+          $node_id: String
         ) {
           insert_analytics_logs_one(
             object: {
@@ -198,6 +202,7 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
               metadata: $metadata
               node_type: $node_type
               node_title: $node_title
+              node_id: $node_id
             }
           ) {
             id
@@ -211,6 +216,7 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
         metadata: metadata,
         node_type: node?.type,
         node_title: nodeTitle,
+        node_id: nodeId,
       },
     });
     return result;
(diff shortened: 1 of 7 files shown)


jessicamcinchak:
[Add node_id column to analytics_logs and analytics_summary tables](https://trello.com/c/Ey6nXonQ/2720-add-nodeid-column-to-analyticslogs-and-analyticssummary-tables)

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -40,6 +40,7 @@
           - id
           - input_errors
           - metadata
+          - node_id
```
jessicamcinchak, line 43: lots of formatting noise here - but this is the key change - there's a new column `node_id` that the public user has permission to `insert` !

hasura.planx.uk/metadata/functions.yaml
DafyddLlyr, line 1: The two diffs here in `functions.yaml` and `tables.yaml` are the reverse of the ones made here https://github.com/theopensystemslab/planx-new/pull/2549 (which is just re-sorting keys alphabetically).

I think that this would indicate one of two things - 
 - Your local version of Hasura doesn't match `main` - `pnpm generate` locally should resolve this and update the image
 - Hasura CLI versions mismatched between `main` - `pnpm i` in `hasura.planx.uk` should resolve this
 
I should have highlighted this when the previous PR was merged sorry!

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -155,12 +155,15 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
       node?.type === TYPES.Content
         ? getContentTitle(node)
         : node?.data?.title ?? node?.data?.text ?? node?.data?.flagSet;
+    const nodeId = node?.id || "Unknown";
```
DafyddLlyr, line 158: As this is a nullable column, setting this to `node?.id || null` may make queries a little easier in Metabase?

jessicamcinchak: Yep updated !

hasura.planx.uk/metadata/functions.yaml
jessicamcinchak: Now fixed, thanks for spotting that ! I'd failed to destroy / recreate my docker images for Hasura since the new merge I think

DafyddLlyr (approved):
Perfect - thank you for picking this up 👍

hasura.planx.uk/migrations/1702899813258_update_view_analytics_summary_node_id/down.sql
Mike-Heneghan, line 1: _Suggestion:_

I think this down migration might fail as I think this would require columns to be dropped which I don't think can be done with a view? 

In an earlier PR I think I maybe had to delete the view then create it as you have here. 

Trying this locally I saw: 

```
hasura migrate apply --down 1 --admin-secret ANCHOVIES
ERRO skipping applying migrations on database '', encountered:                                                            
{
  "error": "query execution failed",
  "path": "$",
  "code": "postgres-error",
  "internal": {
    "arguments": [],
    "error": {
      "description": null,
      "exec_status": "FatalError",
      "hint": null,
      "message": "cannot drop columns from view",
      "status_code": "42P16"
    },
    "prepared": false,
    "statement": "CREATE OR REPLACE VIEW public.analytics_summary AS \nselect \n\ta.id as analytics_id,\n\tal.id as analytics_log_id,\n\tf.slug as service_slug,\n\tt.slug as team_slug,\n\ta.type as analytics_type,\n\ta.created_at as analytics_created_at,\n\tuser_agent,\n\treferrer,\n\tflow_direction,\n\tmetadata,\n\tal.user_exit as is_user_exit,\n\tnode_type,\n\tnode_title,\n\thas_clicked_help,\n\tinput_errors,\n\tCAST(EXTRACT(EPOCH FROM (al.next_log_created_at - al.created_at)) as numeric (10, 1)) as time_spent_on_node_seconds,\n\ta.ended_at as analytics_ended_at,\n\tCAST(EXTRACT(EPOCH FROM (a.ended_at - a.created_at))/60 as numeric (10, 1)) as time_spent_on_analytics_session_minutes\nfrom analytics a\n\tleft join analytics_logs al on a.id = al.analytics_id\n\tleft join flows f on a.flow_id = f.id\n\tleft join teams t on t.id = f.team_id;\n"
  }
} 
FATA[0000] applying migrations failed on database(s):  
```

Although I might have made a mistake / have a non representative local setup.

Mike-Heneghan (commented):
All working for me as expected on the pizza.

This is going to make debugging the logs so much easier! Thanks for doing this @jessicamcinchak

hasura.planx.uk/migrations/1702899813258_update_view_analytics_summary_node_id/down.sql
jessicamcinchak: Thanks for flagging this, sorry it was already merged - will definitely keep this in mind & adjust for future view migrations! Luckily, `down` migrations rarely, if ever, run in production. Small thing: please be careful not to expose local env secrets in code blocks like above!

Mike-Heneghan: D'oh! Thanks for flagging Jess, I'll definitely keep this in mind!

PR #2578, opened 2023-12-18
fix: `analytics_logs.node_type` should reference text component type rather than integer for new and historic records

jessicamcinchak:
Changes: 
- Alter the column type from integer to text (requires dropping & recreating dependent view)
- Do a one-time update of historic records
- Update analytics provider to insert text / human-readable `node_type` for all new `analytics_logs` created

Dependencies: 
- Once merged to production, I expect this will break a number of Metabase queries that currently reference the integer `node_type` in a condition - we'll have to update these independently !


jessicamcinchak:
[Expose human-readable component type in Metabase](https://trello.com/c/l0dDm8Bd/2721-expose-human-readable-component-type-in-metabase)

DafyddLlyr (approved):
Perfect! A much better solution that my original thought here 👌 

Tested and getting the expected results on Pizza 🎉 

image

image

hasura.planx.uk/migrations/1702916788570_alter_column_analytics_logs_node_type/up.sql
```diff
@@ -0,0 +1,70 @@
...
+    CASE 
+        WHEN node_type = '1' THEN 'Flow'
+        WHEN node_type = '3' THEN 'Result'
+        WHEN node_type = '7' THEN 'TaskList'
+        WHEN node_type = '8' THEN 'Notice'
+        WHEN node_type = '9' THEN 'FindProperty'
+        WHEN node_type = '10' THEN 'DrawBoundary'
+        WHEN node_type = '11' THEN 'PlanningConstraints'
+        WHEN node_type = '12' THEN 'PropertyInformation'
+        WHEN node_type = '100' THEN 'Statement'
```
DafyddLlyr, line 20: This has always felt like a bit of a tricky one - if we get feedback that this is oddly named we can very easily update to "Question" or something in future 👍

jessicamcinchak: Yep agree - that's probably a change we should consider making at the root type definition rather than just analytics! Hopefully ambiguous text is still a huge upgrade from integer for starters though!

Mike-Heneghan (approved):
Looks great 🥳

PR #2580, opened 2023-12-19
fix: rationalise analytics logs to remove duplication and inconsistencies

Mike-Heneghan:
## What: 

- Update the approach to gathering analytics
- Differentiate between `auto` answered questions and those the user explicitly sees in approach but also in logs.
- Capture `auto` question when breadcrumbs change
- Capture logs for nodes users see when rendered by the `Card` component 
- Ensure that when a user interacts with a visible component the meta info is recorded against the visible node

## Why:
- Previously when navigating between nodes if there were auto answered questions they may be logged but subsequently many duplicates of the visible node were added to the log. 
- This could result in logging info being inaccurate and not being representative of  the way the user navigates through a flow
- Now explicitly logging all auto answered questions but also differentiating them from those the user interacted with

## To do: 

- [x] Further testing
- [x] Rebase on main and resolve `node_type` conflict
- [x] Review logic and consider optimisations / improvements
- [x] Resolve issue with `input_errors` disappearing
- [x] Update as per feedback

## Testing Notes

Known Issue:

- I've noticed that nodes like `Planning Constraints` render twice and therefore are duplicated in the logs. I'm not sure if this is _wrong_ as such as the user did see this component twice i.e. once in loading state and the second with results although we don't differentiate at the moment. Personally, I'm curious is user wait to see the result or just hit 'continue' 😅

```json
{
  "id": 10,
  "node_id": "4gKrBdC9wk",
  "analytics_id": 2,
  "created_at": "2023-12-20T10:19:54.826965+00:00",
  "flow_direction": "forwards",
  "has_clicked_help": false,
  "input_errors": [],
  "metadata": {
    "isAutoAnswered": false
  },
  "next_log_created_at": "2023-12-20T10:20:13.669525+00:00",
  "node_title": "Planning constraints",
  "node_type": "PlanningConstraints",
  "user_exit": false
},
{
  "id": 9,
  "node_id": "4gKrBdC9wk",
  "analytics_id": 2,
  "created_at": "2023-12-20T10:19:51.592435+00:00",
  "flow_direction": "forwards",
  "has_clicked_help": false,
  "input_errors": [],
  "metadata": {
    "isAutoAnswered": false
  },
  "next_log_created_at": "2023-12-20T10:19:54.826965+00:00",
  "node_title": "Planning constraints",
  "node_type": "PlanningConstraints",
  "user_exit": false
},
``` 

- In testing this I noticed that I missed tracking a user hitting `change` on their "About property" so it won't be visible in the logs in the way other `change` clicks are. 

Screenshot 2023-12-20 at 11 02 18

Potentially useful:

- Adding `where: {metadata: {_contains: {isAutoAnswered: false}}}` to queries is useful for assessing what the user directly saw and comparing metadata like if the user hit "back" or comparing `created_at` of following nodes against `next_log_created_at`

```GraphQL
query MyQuery {
  analytics_logs(order_by: {created_at: desc}, where: {metadata:{_contains: {isAutoAnswered: false}}}) {
    id
    node_id
    analytics_id
    created_at
    flow_direction
    has_clicked_help
    input_errors
    metadata
    next_log_created_at
    node_title
    node_type
    user_exit
  }
}
```

Potential Gotcha:

- Something which I think is correct but unintuitive is that when you go `back` logs can show `forwards` these are because there'll be a `backwards` associated with the visible node but in the background questions can be auto answered and these are classed as going `forwards` which I makes sense as we're answering nodes in the future but was not immediately obvious to me. 

```json
{
  "id": 34,
  "node_id": "9MHz4F7MdL",
  "analytics_id": 2,
  "created_at": "2023-12-20T11:01:56.580244+00:00",
  "flow_direction": "forwards",
  "has_clicked_help": false,
  "input_errors": [],
  "metadata": {
    "isAutoAnswered": true
  },
  "next_log_created_at": null,
  "node_title": "What type of property is it?",
  "node_type": "Statement",
  "user_exit": false
},
{
  "id": 35,
  "node_id": "PyCAP4tqg0",
  "analytics_id": 2,
  "created_at": "2023-12-20T11:01:56.579727+00:00",
  "flow_direction": "forwards",
  "has_clicked_help": false,
  "input_errors": [],
  "metadata": {
    "isAutoAnswered": true
  },
  "next_log_created_at": null,
  "node_title": "What type of property is it?",
  "node_type": "Statement",
  "user_exit": false
},
{
  "id": 33,
  "node_id": "ybK0sfPqep",
  "analytics_id": 2,
  "created_at": "2023-12-20T11:01:56.567269+00:00",
  "flow_direction": "backwards",
  "has_clicked_help": false,
  "input_errors": [],
  "metadata": {
    "isAutoAnswered": false
  },
  "next_log_created_at": null,
  "node_title": "About the property",
  "node_type": "PropertyInformation",
  "user_exit": false
},
```

To do next: 

- [ ] Update the `type` associated with `back` and `change` navigation to use the string value. Also, update historical record and view to reflect this change: https://trello.com/c/vGxXLOBw/2738-update-nodetype-stored-in-analyticslogs-metadata?completedInviteSignup=1&filter=label:Analytics
- [ ] Track `change` via `overrideAnswer` https://trello.com/c/0iGhckG4/2739-track-change-on-overrideanswer
- [ ] Track `allow_list` answers https://trello.com/c/fauHLsFu/2569-track-answers-to-allow-listed-questions-for-guidance-services-non-save-return

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
@@ -3,8 +3,9 @@ import Button from "@mui/material/Button";
 import Container from "@mui/material/Container";
 import Fade from "@mui/material/Fade";
 import { styled, Theme, useTheme } from "@mui/material/styles";
+import { useAnalyticsTracking } from "pages/FlowEditor/lib/analyticsProvider";
 import { useStore } from "pages/FlowEditor/lib/store";
-import React from "react";
+import React, { useEffect } from "react";
 import { ApplicationPath } from "types";
 
 import SaveResumeButton from "./SaveResumeButton";
@@ -47,9 +48,20 @@ const Card: React.FC<Props> = ({
   ...props
 }) => {
   const theme = useTheme();
-  const path = useStore((state) => state.path);
+  const [path, currentCard] = useStore((state) => [
+    state.path,
+    state.currentCard,
+  ]);
   const showSaveResumeButton =
     path === ApplicationPath.SaveAndReturn && handleSubmit;
+  const { track } = useAnalyticsTracking();
+
+  useEffect(() => {
+    // The Card component is only rendered when there's content the user will
+    // see
+    const visibileNode = currentCard();
+    if (visibileNode?.id) track(visibileNode?.id);
+  }, []);
 
   return (
     <Fade
--- a/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
@@ -59,8 +59,8 @@ const Card: React.FC<Props> = ({
   useEffect(() => {
     // The Card component is only rendered when there's content the user will
     // see
-    const visibileNode = currentCard();
-    if (visibileNode?.id) track(visibileNode?.id);
+    const visibleNode = currentCard();
+    if (visibleNode?.id) track(visibleNode?.id);
   }, []);
 
   return (
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
@@ -33,9 +33,10 @@ type NodeMetadata = {
   flag?: Flag;
   title?: string;
   type?: TYPES;
+  isAutoAnswered?: boolean | null;
 };
 
-let lastAnalyticsLogId: number | undefined = undefined;
+let lastVisibleNodeAnalyticsLogId: number | undefined = undefined;
 
 const analyticsContext = createContext<{
   createAnalytics: (type: AnalyticsType) => Promise<void>;
@@ -100,19 +101,19 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
   const previousBreadcrumbs = usePrevious(breadcrumbs);
 
   const onPageExit = () => {
-    if (lastAnalyticsLogId && shouldTrackAnalytics) {
+    if (lastVisibleNodeAnalyticsLogId && shouldTrackAnalytics) {
       if (document.visibilityState === "hidden") {
         send(
           `${
             process.env.REACT_APP_API_URL
-          }/analytics/log-user-exit?analyticsLogId=${lastAnalyticsLogId.toString()}`,
+          }/analytics/log-user-exit?analyticsLogId=${lastVisibleNodeAnalyticsLogId.toString()}`,
         );
       }
       if (document.visibilityState === "visible") {
         send(
           `${
             process.env.REACT_APP_API_URL
-          }/analytics/log-user-resume?analyticsLogId=${lastAnalyticsLogId?.toString()}`,
+          }/analytics/log-user-resume?analyticsLogId=${lastVisibleNodeAnalyticsLogId?.toString()}`,
         );
       }
     }
@@ -166,9 +167,12 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
       : analyticsId;
     // Ensure essential log data is available
     if (nodeToTrack && logDirection && analyticsSession) {
-      const metadata = getNodeMetadata(nodeToTrack);
+      const metadata: NodeMetadata = getNodeMetadata(nodeToTrack);
       const nodeType = nodeToTrack?.type ? TYPES[nodeToTrack.type] : null;
       const nodeTitle = extractNodeTitle(nodeToTrack);
+      const isAutoAnswered = breadcrumbs[nodeId]?.auto || false;
+      metadata["isAutoAnswered"] = isAutoAnswered;
+
       // On component transition create the new analytics log
       const result = await insertNewAnalyticsLog(
         logDirection,
@@ -183,10 +187,17 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
         result?.data.insert_analytics_logs_one?.created_at;
       // On successful create of a new log update the previous log with the next_log_created_at
       // This allows us to estimate how long a user spend on a card
-      if (lastAnalyticsLogId && newLogCreatedAt) {
-        updateLastLogWithNextLogCreatedAt(lastAnalyticsLogId, newLogCreatedAt);
+      if (lastVisibleNodeAnalyticsLogId && newLogCreatedAt && !isAutoAnswered) {
+        // This should only be called when a new log is shown to the user!
+        updateLastLogWithNextLogCreatedAt(
+          lastVisibleNodeAnalyticsLogId,
+          newLogCreatedAt,
+        );
       }
-      lastAnalyticsLogId = id;
+      if (!breadcrumbs[nodeId]?.auto) {
+        // Only update if the node was actually shown to the user
+        lastVisibleNodeAnalyticsLogId = id;
+      }
     }
   }
 
@@ -237,7 +248,7 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
   }
 
   async function updateLastLogWithNextLogCreatedAt(
-    lastAnalyticsLogId: number,
+    lastVisibleNodeAnalyticsLogId: number,
     newLogCreatedAt: Date,
   ) {
     await publicClient.mutate({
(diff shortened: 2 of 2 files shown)


editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -140,43 +144,59 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
...
 
-    const metadata = getNodeMetadata(nodeToTrack);
+    if (!nodeToTrack || !logDirection || !analyticsSession) {
+      return;
+    }
+
+    const metadata: NodeMetadata = getNodeMetadata(nodeToTrack);
     const nodeType = nodeToTrack?.type ? TYPES[nodeToTrack.type] : null;
     const nodeTitle = extractNodeTitle(nodeToTrack);
+    const isAutoAnswered = breadcrumbs[nodeId]?.auto || false;
```
jessicamcinchak, line 170: nit: if you're defaulting to `false` when undefined here, it seems like we should perhaps be able to tighten this type then?

can current type definition `isAutoAnswered?: boolean | null;` actually be `isAutoAnswered: boolean;`  based on implementation here?

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -140,43 +144,59 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
...
       nodeTitle,
       nodeId,
     );
 
-    const id = result?.data.insert_analytics_logs_one?.id;
-    const newLogCreatedAt = result?.data.insert_analytics_logs_one?.created_at;
+    const { id, data } = result?.data.insert_analytics_logs_one || {};
+
+    if (!id || !data || !data.created_at) {
+      return;
```
jessicamcinchak, line 185: nit: is there a case to log an Airbrake error here like `Failed to insert analytics log`, or empty return is intended?

jessicamcinchak (commented):
Looking really good ! The extra context from `isAutoAnswered` is very helpful - I've attached the associated Trello card for that which we discussed earlier this week. 

Happy to approve once a few small questions/typos are addressed.

You note a few edge cases like x2 PlanningConstraints logs and not tracking Property Type "change" (which we actually call `overrideAnswer` in the store, different than `changeAnswer`) - neither of these feel critical to me, the logs already feel much more sensible as-is here.

I am noticing a couple other unexpected things that also aren't necessarily blockers or introduced here, but want to highlight as they should probably be picked up / sorted out in future changes:
- `analytics_logs.input_errors` is no longer populating when I'm hitting visible validation warnings in the frontend - do we know what's changed since this has been recently introduced? Maybe a bad rebase somewhere?
- as you also pointed out in Trello, "back" `analytics_logs.metadata` still includes an integer node type when that should also be updated to text for consistency. following the same approach of update gql insert mutation and do a one-time sql update of historic records should be straightforward i'd imagine!

jessicamcinchak:
[Add metadata to analytics_logs to indicate if a node was hidden from (aka auto-answered) or seen by the user](https://trello.com/c/DIjWixsq/2733-add-metadata-to-analyticslogs-to-indicate-if-a-node-was-hidden-from-aka-auto-answered-or-seen-by-the-user)

Mike-Heneghan:
> Looking really good ! The extra context from `isAutoAnswered` is very helpful - I've attached the associated Trello card for that which we discussed earlier this week.
> 
> Happy to approve once a few small questions/typos are addressed.
> 
> You note a few edge cases like x2 PlanningConstraints logs and not tracking Property Type "change" (which we actually call `overrideAnswer` in the store, different than `changeAnswer`) - neither of these feel critical to me, the logs already feel much more sensible as-is here.
> 
> I am noticing a couple other unexpected things that also aren't necessarily blockers or introduced here, but want to highlight as they should probably be picked up / sorted out in future changes:
> 
> * `analytics_logs.input_errors` is no longer populating when I'm hitting visible validation warnings in the frontend - do we know what's changed since this has been recently introduced? Maybe a bad rebase somewhere?
> * as you also pointed out in Trello, "back" `analytics_logs.metadata` still includes an integer node type when that should also be updated to text for consistency. following the same approach of update gql insert mutation and do a one-time sql update of historic records should be straightforward i'd imagine!

Thank you very much for this review @jessicamcinchak 🙌 

- Great spot with the `analytics_logs.input_errors`, I'm pretty sure I must have broken this in the last few commits as I'm pretty sure it was working on this branch previously. I'll have another look and resolve 👍 

- I'll update the code changes here as suggested and then as you mentioned have a list of `To do next` to cover the `type` change, tracking `overrideAnswer` and reintroducing the tracking of allow list answers which should fit into this approach pretty nicely 🤞 

- Something I also tried out but thought I'd ask about in the dev call is moving the `mutations` into their own file for readability and considering some other refactoring so that if we pause work on it for a while it'll still be understandable when it gets picked up again.

Mike-Heneghan:
> > Looking really good ! The extra context from `isAutoAnswered` is very helpful - I've attached the associated Trello card for that which we discussed earlier this week.
> > Happy to approve once a few small questions/typos are addressed.
> > You note a few edge cases like x2 PlanningConstraints logs and not tracking Property Type "change" (which we actually call `overrideAnswer` in the store, different than `changeAnswer`) - neither of these feel critical to me, the logs already feel much more sensible as-is here.
> > I am noticing a couple other unexpected things that also aren't necessarily blockers or introduced here, but want to highlight as they should probably be picked up / sorted out in future changes:
> > 
> > * `analytics_logs.input_errors` is no longer populating when I'm hitting visible validation warnings in the frontend - do we know what's changed since this has been recently introduced? Maybe a bad rebase somewhere?
> > * as you also pointed out in Trello, "back" `analytics_logs.metadata` still includes an integer node type when that should also be updated to text for consistency. following the same approach of update gql insert mutation and do a one-time sql update of historic records should be straightforward i'd imagine!
> 
> Thank you very much for this review @jessicamcinchak 🙌
> 
> * Great spot with the `analytics_logs.input_errors`, I'm pretty sure I must have broken this in the last few commits as I'm pretty sure it was working on this branch previously. I'll have another look and resolve 👍
> * I'll update the code changes here as suggested and then as you mentioned have a list of `To do next` to cover the `type` change, tracking `overrideAnswer` and reintroducing the tracking of allow list answers which should fit into this approach pretty nicely 🤞
> * Something I also tried out but thought I'd ask about in the dev call is moving the `mutations` into their own file for readability and considering some other refactoring so that if we pause work on it for a while it'll still be understandable when it gets picked up again.

I thought I was going mad for a second but I just realised something in my last commit fundamentally broke `lastVisibleNodeAnaltyicsLog` and I didn't spot it 🤦 

I'll review my last commit and get it sorted.

Mike-Heneghan:
[Rationalise analytics_logs](https://trello.com/c/UireDiIx/2736-rationalise-analyticslogs)

editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
```diff
@@ -47,9 +48,20 @@ const Card: React.FC<Props> = ({
...
+    state.currentCard,
+  ]);
   const showSaveResumeButton =
     path === ApplicationPath.SaveAndReturn && handleSubmit;
+  const { track } = useAnalyticsTracking();
+
+  useEffect(() => {
+    // The Card component is only rendered when there's content the user will
+    // see
+    const visibileNode = currentCard();
```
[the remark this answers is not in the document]
Mike-Heneghan: Fixed here: e93fa5292e9be45d586ffe0fb49246d81b074d2d

I need to add a spell checking extension as I make these mistakes way too often 😅

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
Mike-Heneghan: I've updated the logic around this here: c46b9aada1c6272f7a2a6ab39820241805680812

Although, I've left `isAutoAnswered?` as there's somewhere else `NodeMetadata` is used where I don't think it makes sense to look up if the question is auto answered.

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
Mike-Heneghan: I was just leaving it empty to act as a guard clause and reduce the need to nest subsequent code inside an `if` although this could be a handy place to catch errors. 

There isn't really an error handling in the `analyticsProvider` at the moment and as this is already quite a big PR I'd be tempted to leave it for now and maybe revisit error handling in it's own ticket?

jessicamcinchak: That's totally fair - happy for analytics error handling to be revisited separately / later more comprehensively.

jessicamcinchak (approved):
All working as expected for me now, thanks for changes here - this one feels like an especially nice investment in more valuable & clear `analytics_logs` going forward :tada:

PR #2581, opened 2023-12-19
chore: Drop `BOPS_SUBMISSION_URL_*` env vars in favour of DB values

DafyddLlyr:
Take two for this PR - https://github.com/theopensystemslab/planx-new/pull/2504

It looks like I merged this to a parent branch which was already merged to `main` so this never got picked up 🤦‍♂️ 

I'll rebase this now 👍 

## What does this PR do?
 - Drop `BOPS_SUBMISSION_URL_*` env vars
 - Drop `BOPS_API_ROOT_DOMAIN` env vars
 - Read these values from the DB instead

## Why?
 - Medway require their BOPS env added to PlanX and it now seems like a perfect time to tidy this up a little!

Depends on https://github.com/theopensystemslab/planx-core/pull/212

Before merging this I'd be grateful for another pair of eyes on db values now populated on prod and staging. This query is giving me the expected results - 

```gql
query GetBopsSubmissionURLs {
  teams(where: {integrations: {team_id: {_is_null: false}}}) {
    slug
    integrations {
      production_bops_submission_url
      staging_bops_submission_url
    }
  }
}
```
 
 Once merged I'll also remove these from the real `.env` files on S3 also 👌

PR #2587, opened 2023-12-20
fix: Syntax issue in `SetupTeamIntegrationE2E` mutation

DafyddLlyr:
## What does this PR do?
 - Fixes failing regression test due to bug introduced in https://github.com/theopensystemslab/planx-new/pull/2581
 
 ## What was the cause?
- Relying on `NODE_ENV` over `APP_ENVIRONMENT`

Failing regression test now passing locally on this branch - 

image

🚀 Regression tests running against this branch here - https://github.com/theopensystemslab/planx-new/actions/runs/7275585116

### DIFF
--- a/api.planx.uk/modules/send/bops/bops.ts
+++ b/api.planx.uk/modules/send/bops/bops.ts
@@ -50,7 +50,6 @@ const sendToBOPS = async (req: Request, res: Response, next: NextFunction) => {
     localAuthority,
     env,
   );
-  console.log({ bopsSubmissionURL });
   const isSupported = Boolean(bopsSubmissionURL);
   if (!isSupported) {
     return next(
@@ -186,7 +185,6 @@ const sendToBOPSV2 = async (
     localAuthority,
     env,
   );
-  console.log({ bopsSubmissionURL });
   const isSupported = Boolean(bopsSubmissionURL);
   if (!isSupported) {
     return next(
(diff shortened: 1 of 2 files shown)


api.planx.uk/modules/send/bops/bops.ts
```diff
@@ -44,11 +44,13 @@ const sendToBOPS = async (req: Request, res: Response, next: NextFunction) => {
   // a local or staging API instance should send to the BOPS staging endpoint
   // production should send to the BOPS production endpoint
   const localAuthority = req.params.localAuthority;
-  const env = process.env.NODE_ENV === "production" ? "production" : "staging";
+  const env =
+    process.env.APP_ENVIRONMENT === "production" ? "production" : "staging";
```
DafyddLlyr, line 48: I'd made the assumption that `NODE_ENV` would be a reliable indicator here, but of course we use a production build for E2E tests.

api.planx.uk/modules/send/bops/bops.ts
```diff
@@ -178,21 +180,14 @@ const sendToBOPSV2 = async (
...
   const bopsSubmissionURL = await $api.team.getBopsSubmissionURL(
     localAuthority,
     env,
   );
+  console.log({ bopsSubmissionURL });
```
jessicamcinchak, line 189: nit: is it intentional to keep this `console.log()` and the one above on line 53, or were these temporary debugging measures?

jessicamcinchak (approved):
Thanks for sorting this out!

api.planx.uk/modules/send/bops/bops.ts
DafyddLlyr: Ha good catch sorry! Yep will tidy this up.

PR #2593, opened 2023-12-21
chore: Ignore AWS healthcheck from API logs

DafyddLlyr:
## What does this PR do?
- Excludes AWS healthcheck logs
- Excludes local docker healthcheck logs

## Context
Been meaning to take a look at this small one for a while! 95%+ of the AWS logs are noise, just the healthchecker calling the API which makes troubleshooting locally or on the AWS console a more painful and slow experience.


jessicamcinchak (approved):
:100:

PR #2591, opened 2023-12-20
feat: Set `httpOnly` flag in cookie

DafyddLlyr:
## What does this PR do?
 - Adds the `httpOnly` flag to our auth cookie
 - Corresponding changes to not descructure and read the cookie in the frontend
 - Some refactoring of auth methods to make things a bit easier to follow

## How does auth work?
I'll move this diagram to a page in Notion as well, but this is a helpful diagram for making sense of the changes here - 

```mermaid
sequenceDiagram
  participant Frontend as PlanX Editor
  participant API as PlanX API
  participant GoogleOAuth as Google OAuth
  participant Hasura

  Frontend ->> API: Initiate Login
  activate API
  API ->> GoogleOAuth: Proxy OAuth request

  activate GoogleOAuth
  GoogleOAuth ->> API: Redirect to callback URL
  deactivate GoogleOAuth

  API ->> API: Generate JWT

  alt Staging/Production
    API ->> Frontend: Return JWT in cookie
  else Local Dev/Pizzas
    API ->> Frontend: Return JWT in search params
    deactivate API
    Frontend ->> Frontend: Generate cookie
  end
    
    Frontend ->> API: Make request to /user/me (with cookie)
    activate API
    API ->> Hasura: Get user details
    activate Hasura
    Hasura ->> API: Return user details
    deactivate Hasura
    API ->> Frontend: Return user details
    deactivate API
    Frontend ->> Frontend: Initialise userStore
```

## How do I test this?
 - I can log into the Pizza
 - I can repeatedly log in / log out of the Pizza
 - I can make a request to https://api.2591.planx.pizza/user/me when logged in and I get a response (request is made via cookie)
 - I can see that the `Access-Control-Allow-Origin` header is set to `https://2591.planx.pizza` when hitting `/user/me` as part of the auth process
 
(Once changes merged to staging)
- I can repeat the above steps on staging
- I can see the secure flag appended to my cookie
- I can still log in on Pizzas

PR #2595, opened 2023-12-21
fix: Detect presence of `httpOnly` cookie

DafyddLlyr:
## What the problem?
Following the deploy of #2591, the login / logout flow on staging is now broken. I believe that this is due to the fact that the browser cannot even detect the presence of the `httpOnly` cookie, so the `hasJWT()` function is always returning false on staging.

## What's the solution?
Add a second cookie without the `httpOnly` flag which can be used as a proxy to detect the presence of the `jwt` cookie.

Solution found here - https://stackoverflow.com/a/63877877


DafyddLlyr:
Closed in order to merge this into a more robust solution for https://github.com/theopensystemslab/planx-new/pull/2591

PR #2598, opened 2023-12-21
fix: Drop redundant sanitation for Find and Replace

DafyddLlyr:
## What does this PR do?
Removes sanitation step from Find and Replace API endpoint

## Why?
This is now redundant following https://github.com/theopensystemslab/planx-new/pull/2551 as this step is handled by Hasura in update of the flow table.

## Context
Introduced in https://github.com/theopensystemslab/planx-new/pull/2488 as an inital attempt at handling sanitation.


api.planx.uk/modules/flows/findReplace/findReplace.test.ts
```diff
@@ -264,280 +264,4 @@ describe("string replacement", () => {
         expect(res.body.error).toMatch(/Failed to find and replace/);
       });
   });
-});
-
-describe("HTML replacement", () => {
```
DafyddLlyr, line 269: Dropping this suite does not affect the percentage of test coverage, and I'll make sure it gets picked up in an E2E test that can genuinely test this.

jessicamcinchak (approved):

PR #2601, opened 2023-12-21
feat: track in analytics_logs instances where `changeAnswer` is called

Mike-Heneghan:
## What

- More comprehensively track instances where a user makes a `change` 
- This is associated with `changeAnswer` being called
- On `overrideAnswer` track "change" click but leave the target node metadata as `{}`
- Track the the id of the target node
- Refactor to use the existing `extractNodeTitle` function to more comprehensively extract the title. 

## Why 

- The metrics team have requested "change" clicks are tracked
- In the instance of `overrideAnswer` the logic is unlike `changeAnswer` and the basic requirement is tracking the "change" click not the target node. 
- When tracking target node data add the id as it can be helpful rather than relying on just the node title and type
- Also, in testing noticed that the node title was often undefined due to the limited way it was being extracted. Refactored to use an existing more comprehensive function. 

## Screen recording

https://github.com/theopensystemslab/planx-new/assets/36415632/455af116-1861-4e39-8f4d-c3c35f8ecaba

### DIFF
--- a/editor.planx.uk/src/@planx/components/PropertyInformation/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/PropertyInformation/Public.tsx
@@ -11,6 +11,7 @@ import { useFormik } from "formik";
 import { submitFeedback } from "lib/feedback";
 import { publicClient } from "lib/graphql";
 import find from "lodash/find";
+import { useAnalyticsTracking } from "pages/FlowEditor/lib/analyticsProvider";
 import { useStore } from "pages/FlowEditor/lib/store";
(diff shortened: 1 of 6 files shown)


Mike-Heneghan:
[Track `change` on `overrideAnswer`](https://trello.com/c/0iGhckG4/2739-track-change-on-overrideanswer)

Mike-Heneghan:
[Track `change` on `overrideAnswer`](https://trello.com/c/0iGhckG4/2739-track-change-on-overrideanswer)

jessicamcinchak (approved):
Looks good and working as expected for me!

PR #2602, opened 2023-12-22
feat: Add `team_themes` table

DafyddLlyr:
## What does this PR do?
 - Moves `team.theme` JSONB to a new `team_themes` table
 - Migrates data from column to table
 - Updates frontend queries and types

Relies on https://github.com/theopensystemslab/planx-core/pull/242

### DIFF
--- a/hasura.planx.uk/migrations/1703240593405_squashed/up.sql
+++ b/hasura.planx.uk/migrations/1703240593405_squashed/up.sql
@@ -1,0 +1,30 @@
+CREATE TABLE "public"."team_themes" (
+    "id" serial NOT NULL,
+    "team_id" integer NOT NULL,
+    "primary_colour" text NOT NULL DEFAULT '#0010A4',
+    "secondary_colour" text,
+    "logo" text,
+    "favicon" text,
+    PRIMARY KEY ("id"),
+    FOREIGN KEY ("team_id") REFERENCES "public"."teams"("id") ON UPDATE cascade ON DELETE cascade
+);
+
+ALTER TABLE
+    "public"."team_themes"
+ADD
+    CONSTRAINT "team_themes_team_id_key" UNIQUE ("team_id");
+
+COMMENT ON COLUMN "public"."team_themes"."primary_colour" IS E'Must be hex triplet (e.g. #112233)';
+
+COMMENT ON COLUMN "public"."team_themes"."secondary_colour" IS E'Must be hex triplet (e.g. #112233)';
+
+INSERT INTO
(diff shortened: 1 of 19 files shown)


hasura.planx.uk/migrations/1703240593405_squashed/up.sql
```diff
@@ -0,0 +1,30 @@
...
+INSERT INTO
+    team_themes (team_id, primary_colour, logo)
+SELECT
+    id AS team_id,
+    COALESCE(theme ->> 'primary', '#0010A4') AS primary_colour,
+    COALESCE(theme ->> 'logo', NULL) AS logo
+FROM
+    teams;
+
+ALTER TABLE "public"."teams" DROP COLUMN "theme" CASCADE;
```
DafyddLlyr, line 30: Migrating and then dropping the column will work on staging and prod, but on the initial migration on Pizza and locally it won't do anything as the data sync happens after Hasura migrations run.

I tested this locally without dropping the column. If preferred, I can split this into two steps / two PRs.

scripts/seed-database/write/teams.sql
```diff
@@ -3,6 +3,7 @@ CREATE TEMPORARY TABLE sync_teams (
   id integer,
   name text,
   slug text,
+  -- TODO: Drop this and fetch from team_themes
```
DafyddLlyr, line 6: TODO intentionally left here, we'll need to update sync scripts only once this new table is in prod.

hasura.planx.uk/migrations/1703240593405_squashed/up.sql
```diff
@@ -0,0 +1,30 @@
+CREATE TABLE "public"."team_themes" (
+    "id" serial NOT NULL,
+    "team_id" integer NOT NULL,
+    "primary_colour" text NOT NULL DEFAULT '#0010A4',
+    "secondary_colour" text,
```
DafyddLlyr, line 5: I'll check with @ianjon3s in the new year how we should handle defaults for this column.

jessicamcinchak (approved):

scripts/seed-database/write/teams.sql
Mike-Heneghan: Hey @DafyddLlyr 

I think this _might_ be related but I was having issues building `main` since coming back from the holidays. 

I was seeing: 

```
psql:write/teams.sql:18: ERROR:  invalid input syntax for type json
DETAIL:  Expected end of input, but found "-11".
```

Which was resolved by commenting out `theme jsonb,`

PR #2606, opened 2023-12-22
chore(editor): Minor update for all Editor dependencies

DafyddLlyr:


jessicamcinchak (approved):
Thanks for picking up, will wait to merge any dependabot "editor"-tagged PRs until this is in, as it should automatically close a few of those !

PR #2628, opened 2024-01-03
fix(e2e): Update `planx-core` to resolve failing test

DafyddLlyr:
✅ Regression tests passing against branch here - https://github.com/theopensystemslab/planx-new/actions/runs/7396684356


jessicamcinchak (approved):

PR #2629, opened 2024-01-03
feat(wip): do not auto-answer questions into the future by only calculating one `upcomingCardIds`

jessicamcinchak:
Getting expected behavior in simple cases, lots of testing still needed  / preview store unit tests need updated

Building a pizza to share with August & co for content testing !! 

Test flow with question, checklist & flag/filter auto-answer behavior:
- After: https://2629.planx.pizza/testing/auto-answer-next
- Before: https://editor.planx.uk/testing/auto-answer-next

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/preview/upcomingCardIds.test.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/preview/upcomingCardIds.test.ts
@@ -51,8 +51,8 @@ beforeEach(() => {
   setState({ flow });
 });
 
-test("The first _root edge is immediately queued up", () => {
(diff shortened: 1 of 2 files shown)


jessicamcinchak:
[Try only recording an auto-answer when a node is actually reached, rather than in the future](https://trello.com/c/7GPqwZSa/2245-try-only-recording-an-auto-answer-when-a-node-is-actually-reached-rather-than-in-the-future)

jessicamcinchak:
Closing this PR as this approach doesn't work with complex flows and it's quite outdated, see logic board for next idea: a component wrapper to handle visibility and further decouple navigation from visibility/automations. New work on a fresh branch soon.

PR #2630, opened 2024-01-03
fix: Unescape HTML entities for alternate DOMPurify check

DafyddLlyr:
## What's the problem?
Originally picked up here - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1703175603781519 (OSL Slack)

A permission error is thrown when attempting to copy some flows from the "Templates" team, despite the users in question having the requisite permission level.

## What's the cause?
The logged error is misleading, it's actually an [input validation](https://github.com/theopensystemslab/planx-new/pull/2551) error that's being thrown - the try/catch block just didn't account for this.

What's happening here is that the simple `length === length` check for DOMPurify is not accounting for HTML entities (see screenshot below).

image

There isn't a simple alternative way of checking that DOMPurify has modified content unfortunately. We can't check string equality directly as the order of HTML attributes in sanitised content can differ from the original (and there's not a method of preserving this). Please see comment from DOMPurify readme below on checking for removed attributes directly -

> After sanitizing your markup, you can also have a look at the property DOMPurify.removed and find out, what elements and attributes were thrown out. Please **do not use** this property for making any security critical decisions. This is just a little helper for curious minds.

Source: https://github.com/cure53/DOMPurify?tab=readme-ov-file#okay-makes-sense-lets-move-on

## What's the solution?
Check length, and then unescape known HTML entities which are converted, and re-check. I'll be honest this doesn't feel ideal but it's the best / simplest solution I've reached.

I'm throwing noisy errors here so that we can more easily spot any more issues going forward.

### DIFF
--- a/api.planx.uk/modules/webhooks/service/validateInput/utils.ts
+++ b/api.planx.uk/modules/webhooks/service/validateInput/utils.ts
@@ -40,9 +40,20 @@ export const isCleanHTML = (input: unknown): boolean => {
 
   const cleanHTML = DOMPurify.sanitize(input, { ADD_ATTR: ["target"] });
   // DOMPurify has not removed any attributes or values
-  const isClean = cleanHTML.length === unescapeHTML(input).length;
+  const isClean =
+    cleanHTML.length === input.length ||
+    unescapeHTML(cleanHTML).length === unescapeHTML(input).length;
+
+  if (!isClean) {
+    console.log("CLEAN HTML: ", cleanHTML);
+    console.log("INPUT HTML: ", input);
+    console.log("UNESCAPED HTML: ", unescapeHTML(input));
+  }
   return isClean;
 };
 
 const unescapeHTML = (input: string): string =>
-  input.replace(/&quot;/gi, '"').replace(/&apos;/gi, "'");
+  input
+    .replace(/&quot;/gi, '"')
+    .replace(/&apos;/gi, "'")
+    .replace(/&nbsp;/gi, " ");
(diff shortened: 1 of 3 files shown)


api.planx.uk/modules/webhooks/service/validateInput/utils.ts
```diff
@@ -39,7 +40,35 @@ export const isCleanHTML = (input: unknown): boolean => {
...
+
+/**
+ * Explicity log error when unsafe HTML is encountered
+ * This is very likely a content / sanitation error as opposed to a security issue
+ * Logging this should help us identify and resolve these
+ */
+const logUncleanHTMLError = (input: string, cleanHTML: string) => {
+  const userId = userContext.getStore()?.user.sub;
+
+  console.error({
```
DafyddLlyr, line 62: I think this should be fine, I'll check on staging. We might need to use the `logger.notify()` method Ben wrote if this doesn't get caught by Airbrake.

api.planx.uk/modules/webhooks/service/validateInput/utils.ts
```diff
@@ -39,7 +40,35 @@ export const isCleanHTML = (input: unknown): boolean => {
...
+    message: `Warning: Unclean HTML submitted!`,
+    userId,
+    input,
+    cleanHTML,
+  });
+};
+
+const unescapeHTML = (input: string): string =>
```
DafyddLlyr, line 70: I tested a load of HTML entities and these were the only one which were getting caught. 

Tiptap is converting the quotation marks, and DOMPurify appears to be converting all HTML entities apart from `&nbsp;`.

jessicamcinchak (approved):

PR #2634, opened 2024-01-05
feat: Set `httpOnly` flag in cookie, use cookie for auth

DafyddLlyr:
## Context
Merges PRs https://github.com/theopensystemslab/planx-new/pull/2591 and https://github.com/theopensystemslab/planx-new/pull/2595 which were previously reverted. These two PRs have had their commits cherry picked and have been previously reviewed.

Code from [`ecb270a`](https://github.com/theopensystemslab/planx-new/pull/2634/commits/ecb270a3e65990c098ea4aea043a7105a029133d) is new to this branch.

## What does this PR do?
- Sets `httpOnly` flag in cookie (see https://github.com/theopensystemslab/planx-new/pull/2591)
- Adds additional `auth` cookie to detect presence of unreadable `httpOnly` cookie (see https://github.com/theopensystemslab/planx-new/pull/2595)
- Pass JWT from `user/me` endpoint and retains in `UserStore` - this is required to pass `JWT` to `planx-core` as well as to pass into auth headers.

## Testing
- I can log in / log out locally
- I can log in / log out on Pizza
- I can make requests via Apollo (e.g. GetTeams), `planx-core` (e.g. get user permissions) and axios/fetch (e.g. file upload) on Pizza

## Other
 - I'm not totally convinced this will work as is with regards to CORS on staging - I'd like to merge and test this (and then fix forward).
 - There's room for simplification here - we could avoid the repetitive  process of building the auth headers if we used a custom Axios instance or followed an [interceptor](https://axios-http.com/docs/interceptors) pattern. I've not picked this up here as we use both `fetch` and `Axios` in places and there's enough change on this PR as is. I'll add an issue or ticket to look into this.

### DIFF
--- a/api.planx.uk/server.ts
+++ b/api.planx.uk/server.ts
@@ -38,19 +38,12 @@ useSwaggerDocs(app);
 
 app.set("trust proxy", 1);
 
-app.use((req, res, next) => {
-  res.header("Access-Control-Allow-Origin", req.headers.origin);
-  res.header(
-    "Access-Control-Allow-Headers",
-    "Origin, X-Requested-With, Content-Type, Accept",
-  );
-  next();
-});
-
 app.use(
   cors({
     credentials: true,
     methods: "*",
+    origin: process.env.EDITOR_URL_EXT,
+    allowedHeaders: "Origin, X-Requested-With, Content-Type, Accept",
   }),
 );
 
--- a/api.planx.uk/server.ts
+++ b/api.planx.uk/server.ts
@@ -43,7 +43,13 @@ app.use(
     credentials: true,
     methods: "*",
     origin: process.env.EDITOR_URL_EXT,
-    allowedHeaders: "Origin, X-Requested-With, Content-Type, Accept",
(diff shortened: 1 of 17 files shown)


api.planx.uk/server.ts
```diff
@@ -38,19 +38,18 @@ useSwaggerDocs(app);
...
-    "Origin, X-Requested-With, Content-Type, Accept",
-  );
-  next();
-});
-
 app.use(
   cors({
     credentials: true,
     methods: "*",
+    origin: process.env.EDITOR_URL_EXT,
```
DafyddLlyr, line 45: I suspect that we'll need to change this to be "pizzas + staging" if we're on the staging API, either here or specifically in the auth middleware somewhere. I'd like to merge to staging to test this before implementing anything just now just to double check.

jessicamcinchak (approved):
Testing notes all working as expected for me here! Merge to staging, test & fix-forward approach sounds good - shout when you're ready for extra eyes testing there :crossed_fingers:

PR #2635, opened 2024-01-08
fix: Use `he` library to handle HTML decoding

DafyddLlyr:
## What does this PR do?
- Addresses https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1704472261489119
- Uses the [`he`](https://github.com/mathiasbynens/he) library to handle HTML decoding in a more robust manner

## Testing
I'll also open this up to August to test before merging, but the change here should allow us to copy "permitted-development" flows from the OSL team without issues.


jessicamcinchak (approved):
Successfully copied pd flow in this pizza :+1:

PR #2638, opened 2024-01-08
fix: Setup `authMiddleware` for Apollo

DafyddLlyr:
**Take 1**
~~This seems like the simplest solution here - to use the cookie for requests via Apollo directly, instead of trying to read it (which isn't possible as it's now `httpOnly`!)~~

~~My original idea was to wait for the JWT to be returned before making the failing query, but it's actually a bit more complex as the client is already instantiated by this point.~~

**Take 2**
So it turns out Hasura won't accept either a cookie or an auth header - is has to be one or the other. This would mean a number of changes to how we handle tests, and authenticate via scripts, or the API etc, so seems a non-starter. For reference - this setting can be controlled via the [`header`](https://hasura.io/docs/latest/auth/authentication/jwt/#header) JWT configuration option.

I've gone back to my initial ideal of waiting for authentication first using a middleware pattern ([Apollo docs](https://www.apollographql.com/docs/react/networking/authentication/#header)).

### DIFF
--- a/api.planx.uk/modules/auth/controller.ts
+++ b/api.planx.uk/modules/auth/controller.ts
@@ -59,7 +59,8 @@ function setJWTCookie(returnTo: string, res: Response, req: Request) {
   res.cookie("jwt", req.user!.jwt, httpOnlyCookieOptions);
 
   // Set second cookie which can be read by browser to detect presence of the unreadable httpOnly cookie
-  res.cookie("auth", { loggedIn: true }, defaultCookieOptions);
+  const authCookie = btoa(JSON.stringify({ loggedIn: true }));
+  res.cookie("auth", authCookie, defaultCookieOptions);
 
   res.redirect(returnTo);
 }
--- a/editor.planx.uk/src/lib/graphql.ts
+++ b/editor.planx.uk/src/lib/graphql.ts
@@ -13,8 +13,6 @@ import { logger } from "airbrake";
 import { useStore } from "pages/FlowEditor/lib/store";
 import { toast } from "react-toastify";
 
-import { getCookie } from "./cookie";
-
 const toastId = "error_toast";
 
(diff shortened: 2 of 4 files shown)


jessicamcinchak (approved):

jessicamcinchak:
Tricky, thanks for working through these errors !

editor.planx.uk/src/lib/graphql.ts
```diff
@@ -131,11 +127,48 @@ const retryLink = new RetryLink({
...
+    headers: {
+      authorization: jwt ? `Bearer ${jwt}` : undefined,
+    },
+  };
+});
+
+/**
+ * Get the JWT from the store, and wait if not available
+ */
+const getJWT = async () => {
```
DafyddLlyr, line 147: I did spend a bit of time trying to make this a function in the `UserStore` but couldn't get the `subscribe()` on line 159 to work within this context.

api.planx.uk/modules/auth/controller.ts
```diff
@@ -59,7 +59,8 @@ function setJWTCookie(returnTo: string, res: Response, req: Request) {
   res.cookie("jwt", req.user!.jwt, httpOnlyCookieOptions);
 
   // Set second cookie which can be read by browser to detect presence of the unreadable httpOnly cookie
-  res.cookie("auth", { loggedIn: true }, defaultCookieOptions);
+  const authCookie = btoa(JSON.stringify({ loggedIn: true }));
+  res.cookie("auth", authCookie, defaultCookieOptions);
```
DafyddLlyr, line 63: This is just a tidy up / non-functional change - it just stops the cookie coming back as `[Object object]`. I'm just wrapping it in `btoa()` for a bit of obfuscation - the value it meaningless and never used anyway.

PR #2590, opened 2023-12-20
wip: track answers to allow-listed questions on `lowcal_sessions` for analytics

jessicamcinchak:

PR #2641, opened 2024-01-09
feat: setup a Cron trigger to update allow-list answers for analytics on submitted sessions

jessicamcinchak:
Replaces #2590 

**Changes:**
- Sets up a Hasura "Cron Triggers" event & webhook which checks for recently submitted sessions, parses their passport data for allow-list answers, and updates the new column `allow_list_answers`. 
  - This column is _not_ included in data sanitation so that it can be retained for analytics over all time
- Updates the existing view `submitted_sessions_summary` to include the new `allow_list_answers` field

**Why a Cron trigger instead of a plain database trigger like the original PR?**: 
- This approach will backfill all submitted but not-yet sanitised sessions on staging & production once merged, rather than creating a new trigger that would only run _after_ `submitted_at` is updated
- It's nicer to parse the passport and construct the new `allow_list_answers` JSONB data format using TS rather than PSQL
- We can use the existing data sanitation pattern for this which has: 
  - Good error handling & logging to #planx-notifications-internal
  - Scheduling & runtime visibility in the Hasura console "Events" tab
  - Easy to extend "operations" pattern, so if there's more requests for aggregate analytics on lowcal_sessions beyond tracking allow-list answers in the future, they could run as part of this same job
- Since Metabase caches data & doesn't immediately reflect updates anyways, there shouldn't be a huge delay for end-users here if we are running this operation on a schedule rather than immediately when `submitted_at` is updated

**Testing:**:
- Fill out a test flow (submitting to devops@ temporarily): https://2641.planx.pizza/testing/submissions-test/preview
- Open Hasura console to confirm that cron trigger event succeeded and that your lowcal_sessions record has values populated in `allow_list_answers` field (currently running every minute for quick feedback, will scale back before merging)


hasura.planx.uk/metadata/cron_triggers.yaml
```diff
@@ -1,3 +1,12 @@
+- name: analyze_sessions
+  webhook: '{{HASURA_PLANX_API_URL}}/webhooks/hasura/analyze-sessions'
+  schedule: '* * * * *'
```
jessicamcinchak, line 3: This is running every minute for easy testing on the pizza - what's our preferred schedule for staging/production? Nightly? More than once a day?

jessicamcinchak:
[Create new data structure that can track anonymised, aggregated analytics about save & return sessions](https://trello.com/c/S3A3dIf1/2568-create-new-data-structure-that-can-track-anonymised-aggregated-analytics-about-save-return-sessions)

hasura.planx.uk/metadata/cron_triggers.yaml
Mike-Heneghan: I think nightly to start with seems reasonable, as far as I'm aware there aren't any users / use cases where the lag between submission and processing would be an issue.

Mike-Heneghan (approved):
Looks great and working as expected for me 👍

hasura.planx.uk/metadata/cron_triggers.yaml
jessicamcinchak: Updated to run once daily at 3am!

PR #2643, opened 2024-01-09
wip: UI setup for team design settings

ianjon3s:
# What does this PR do?

Builds UI for editor team design settings, currently hidden behind a feature flag.

Example:
https://2643.planx.pizza/barnet/settings/design

Feature flag to toggle design settings:
`window.featureFlags.toggle("SHOW_TEAM_SETTINGS")`

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/index.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/index.tsx
@@ -8,16 +8,16 @@ import { styled } from "@mui/material/styles";
 import Tab from "@mui/material/Tab";
 import Tabs from "@mui/material/Tabs";
 import { HEADER_HEIGHT } from "components/Header";
-import React  from "react";
+import React from "react";
 import { Link, useNavigation } from "react-navi";
 
 interface SettingsProps {
-  currentTab: string,
+  currentTab: string;
   tabs: {
-    route: string,
-    name: string,
-    Component: React.FC,
-  }[]
+    route: string;
+    name: string;
+    Component: React.FC;
+  }[];
 }
 
 interface TabPanelProps {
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Settings/index.tsx
```diff
@@ -95,6 +95,7 @@ const Root = styled(Box)(({ theme }) => ({
   left: 0,
   right: 0,
   minHeight: `calc(100% - ${HEADER_HEIGHT}px)`,
+  zIndex: "2000",
```
DafyddLlyr, line 98: We don't do this consistently anywhere yet, but ideally we'd assign these at the theme level to avoid conflicts.

Might be a nice quick tidy up to pick up at some point?

Docs: 
- https://mui.com/material-ui/customization/z-index/
- https://mui.com/material-ui/customization/default-theme/?expand-path=$.zIndex

editor.planx.uk/src/ui/editor/EditorRow.tsx
```diff
@@ -27,7 +27,7 @@ const Root = styled(Box, {
   }),
 }));
 
-export default function InputRow({
+export default function EditorRow({
```
DafyddLlyr, line 30: 👍

DafyddLlyr (approved):
Looking fantastic and great to see Formik already getting wired up also 🙌

editor.planx.uk/src/pages/FlowEditor/components/Settings/index.tsx
ianjon3s: Ah yes, this is a leftover shim from the previous incarnation of the help text button, which was using fixed positioning and was being pushed above the setting panel. 

I'll remove this for now and look to update the z-index hierarchy at the theme level in a separate PR.

PR #2644, opened 2024-01-09
feat: create parallel BOPS send events for v1 & v2 on all environments

jessicamcinchak:
Per dev call update that v2 API is available on BOPS production.

Satisfying PR will be the next one when we can retire this parallel approach and only submit via v2 :slightly_smiling_face:


jessicamcinchak:
[[EPIC] Update BOPS submissions to use new Digital Planning Data payload](https://trello.com/c/L24kI8JK/2671-epic-update-bops-submissions-to-use-new-digital-planning-data-payload)

DafyddLlyr (approved):

PR #2645, opened 2024-01-09
Production deploy

jessicamcinchak:


DafyddLlyr:
> Anything specific to watch on production around cookie changes @DafyddLlyr ?

Nope - it should be exactly the same as staging which is working as intended. Thanks for checking!

If any users face issues we can prompt them to log out / log in, but I think they'll actually just all get redirected to `/login` anyway if they're missing the new `auth` cookie 👍

DafyddLlyr (approved):

PR #2654, opened 2024-01-10
fix: for backwards tracking events update analytics_logs.metadata to use the text component type rather than integer for new and historical records

Mike-Heneghan:
## What 

- Update the way `back` and `change` events are tracked to store the node type as text rather than an integer for all new tracking events
- Add a migration to update historical records

## Why 

- It can be difficult to understand the node type by their integer alone, using the text makes it more human readable

## Testing 

- I tried running the migrations via the Hasura console locally ensuring that the data updated in the correct way for both `up` and `down`
- Then ensured this worked using `hasura` cli to apply `up`/`down` migrations

### DIFF
--- a/hasura.planx.uk/migrations/1704887727847_sql_migration_update_analytics_logs_metadata_to_use_text_node_type/down.sql
+++ b/hasura.planx.uk/migrations/1704887727847_sql_migration_update_analytics_logs_metadata_to_use_text_node_type/down.sql
@@ -1,0 +1,80 @@
+UPDATE public.analytics_logs
+SET metadata = jsonb_set(
+    metadata,
+    '{back,type}',
+    CASE 
+        WHEN metadata->'back'->>'type' = 'Flow' THEN '1'
+        WHEN metadata->'back'->>'type' = 'Result' THEN '3'
+        WHEN metadata->'back'->>'type' = 'TaskList' THEN '7'
+        WHEN metadata->'back'->>'type' = 'Notice' THEN '8'
+        WHEN metadata->'back'->>'type' = 'FindProperty' THEN '9'
+        WHEN metadata->'back'->>'type' = 'DrawBoundary' THEN '10'
+        WHEN metadata->'back'->>'type' = 'PlanningConstraints' THEN '11'
+        WHEN metadata->'back'->>'type' = 'PropertyInformation' THEN '12'
+        WHEN metadata->'back'->>'type' = 'Statement' THEN '100'
+        WHEN metadata->'back'->>'type' = 'Checklist' THEN '105'
+        WHEN metadata->'back'->>'type' = 'TextInput' THEN '110'
+        WHEN metadata->'back'->>'type' = 'DateInput' THEN '120'
+        WHEN metadata->'back'->>'type' = 'AddressInput' THEN '130'
+        WHEN metadata->'back'->>'type' = 'ContactInput' THEN '135'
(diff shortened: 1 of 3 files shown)


hasura.planx.uk/migrations/1704887727847_sql_migration_update_analytics_logs_metadata_to_use_text_node_type/down.sql
Mike-Heneghan, line 1: _Comment_

I messed around for a while trying to update `back` and `change` in one query and or refactor to improve readability but I couldn't seem to get it to work. 

This is really verbose but I believe it does what we intend and it's a one off anyway.

hasura.planx.uk/migrations/1704887727847_sql_migration_update_analytics_logs_metadata_to_use_text_node_type/up.sql
Mike-Heneghan, line 1: Comment

I messed around for a while trying to update back and change in one query and or refactor to improve readability but I couldn't seem to get it to work.

This is really verbose but I believe it does what we intend and it's a one off anyway.

hasura.planx.uk/migrations/1704887727847_sql_migration_update_analytics_logs_metadata_to_use_text_node_type/up.sql
```diff
@@ -0,0 +1,80 @@
...
+        WHEN metadata->'back'->>'type' = '500' THEN '"Filter"'
+        WHEN metadata->'back'->>'type' = '600' THEN '"Review"'
+        WHEN metadata->'back'->>'type' = '650' THEN '"Send"'
+        WHEN metadata->'back'->>'type' = '700' THEN '"Calculate"'
+        WHEN metadata->'back'->>'type' = '725' THEN '"Confirmation"'
+        WHEN metadata->'back'->>'type' = '730' THEN '"NextSteps"'
+        ELSE metadata->'back'->>'type'
+    END::jsonb
+)
+WHERE metadata->'back'->>'type' IN ('1', '3', '7', '8', '9', '10', '11', '12', '100', '105', '110', '120', '130', '135', '140', '145', '150', '200', '250', '300', '310', '360', '380', '400', '500', '600', '650', '700', '725', '730');
```
jessicamcinchak, line 39: nit: is this `WHERE` clause actually necessary if 1) this is just running one-time and 2) we have the final catchall `ELSE metadata->'back'->>'type'` ?

jessicamcinchak (approved):
All working as expected for me on pizza :+1: totally happy for this to be two separate commands, finicky jsonb syntax isn't worth losing time to for a one-off migration !

One very small question/comment, but not at all a blocker to merging.

hasura.planx.uk/migrations/1704887727847_sql_migration_update_analytics_logs_metadata_to_use_text_node_type/up.sql
Mike-Heneghan: That's a really good question, I had issues with the SQL query failing if I didn't have this statement and I think it was related to expectations are around `jsonb_set`.

If I remember correctly I tried without any `WHERE` clause as well as a less verbose `WHERE metadata->'back'->>'type' IS NOT NULL;` but in both cases I think it was throwing errors. 

I think there must be a better way than what I've got here but as you say the jsonb syntax if finicky so I think I'll leave it as is 😂

PR #2656, opened 2024-01-12
fix: Exit parent function on `shouldSkipTracking()`

DafyddLlyr:
## What does this PR do?
- Fixes small logical bug introduced in https://github.com/theopensystemslab/planx-new/pull/2597
- `skipUpdateIfNotTracking()` does not correctly return from parent function
- I encountered an issue with this when the `ErrorWrapper` component was trying to write a log to the DB from within the Editor. You can test this by triggering an error state a "Send" node - https://editor.planx.dev/buckinghamshire/request-a-building-control-site-inspection/nodes/_root/nodes/9VzYz3damo/edit


Mike-Heneghan (approved):
Thanks for catching and fixing this @DafyddLlyr!

I also really appreciate the testing link 🙌 

I've tested this locally and your fix is working as expected for me 👍

PR #2646, opened 2024-01-09
feat: Access team secondary colour via MUI theme

DafyddLlyr:
## What does this PR do?
 - Update `team_themes` table to include `action_colour` and `link_colour` columns, and drops `secondary_colour` column
 - Pulls these new values into the MUI theme
 - Imports `Team` and associated types from `planx-core`
 - Depends upon https://github.com/theopensystemslab/planx-core/pull/259

## Not in this PR 
- No style changes are made
- `link_colour` value in the database is set to the current default
- `color="prompt"` is not used anywhere yet so this value is not pulled in and used
- @ianjon3s is planning to pick up these changes (thank you!)

If you wish to do so, you can test this locally by updating colours for a team in Hasura and then see them get picked up in `/preview` and `/unpublished` routes. You'll need to set `color="prompt"` on a component (e.g. button) to see the `action_colour` feeding through.

PR #2658, opened 2024-01-12
feat: Apply theme colours across app

ianjon3s:
# What does this pull request do?

Applies new theme colours created in https://github.com/theopensystemslab/planx-new/pull/2646 across the components and UI of the app.

Notes:
- The content component is correctly selecting the appropriate colours for contrast link text (default: link colour, plus black or white depending on contrast), however I was not able to get this to function in the notice component due to overriding specificity set in the theme. This is also a current issue on the live services. For simplicity I've set the link text to inherit paragraph text colour in this component, and will revisit this in a future PR.
- As we are now using both dark and light button background colours, I've adjusted the tonal offset to 0.1, ensuring that darkening a light background colour on hover does not reduce it's contrast with dark overlay by too much a margin.

### DIFF
--- a/editor.planx.uk/src/@planx/components/Content/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/Content/Public.tsx
@@ -20,7 +20,7 @@ const Content = styled(Box, {
       theme.palette.text.primary,
     ])?.toHexString() || theme.palette.text.primary,
   "& a": {
-    color: getContrastTextColor(color || "#fff", theme.palette.primary.main),
+    color: getContrastTextColor(color || "#fff", theme.palette.link.main),
   },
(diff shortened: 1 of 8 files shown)


DafyddLlyr (approved):

DafyddLlyr:
> * The content component is correctly selecting the appropriate colours for contrast link text (default: link colour, plus black or white depending on contrast), however I was not able to get this to function in the notice component due to overriding specificity set in the theme. This is also a current issue on the live services. For simplicity I've set the link text to inherit paragraph text colour in this component, and will revisit this in a future PR.

Thanks for the heads up here - it would be good to get to the bottom of this but no rush if it's currently an outstanding issue on production also.

PR #2659, opened 2024-01-12
fix: use email as default destination for new Send components and add warnings when selecting BOPS or Uniform

jessicamcinchak:
We've had a number of BOPS submission failures recently for services that shouldn't be attempting to send to BOPS in the first place because they're unsupported application types.

This makes two simple changes to the Send component: 
- Switches default "destination" from BOPS to email for all newly created Send components in the Editor
- Adds a simple window alert on selection of BOPS or Uniform if the flow slug isn't an obviously supported application type. This is just meant to be a quick deterent for now, and we'll revisit and think about more robust on-publish validation later this phase via [this Trello ticket](https://trello.com/c/LGIoRgym/2641-validate-that-flow-meets-schema-requirements-on-publish)


DafyddLlyr (approved):
Nice solution!

PR #2666, opened 2024-01-16
fix: Add custom domains to CORS allowlist

DafyddLlyr:
## What does this PR do?
- Fixes a CORS issue when requests go from custom domains to api.editor.planx.uk
- This was introduced here - https://github.com/theopensystemslab/planx-new/pull/2634/files#diff-a23b8e0cdc75fc858dc3a73983f67d257f78158596c37832202cf963efcf0a6aL42

## How is this resolved?
- Adding a new env var `CORS_ALLOWLIST`
- This is only populated on production where the API needs to be accessed from both editor.planx.uk and planningservices.COUNCIL.gov.uk
- Rather than adding a new place to update the custom domains list I'm reading from the existing variable in the infrastructure project
- In non-prod environments, we can just call back to the current value which is the EDITOR_URL_EXT env var

## To test
 - I can log in / log out locally ✅ 
 - I can make a non-auth API request locally (e.g upload file) ✅ 
 - I can log in / log out on Pizza ✅
 - I can make a non-auth API request on Pizza (e.g upload file) ✅
 - I can log in / log out on Staging
 - I can make a non-auth API request on Staging (e.g upload file)
 - I can log in / log out on Prod
 - I can make a non-auth API request on Prod (e.g upload file)
 - I can log in / log out on a custom domain
 - I can make a non-auth API request on a custom domain (e.g upload file)


docker-compose.yml
```diff
@@ -141,6 +141,7 @@ services:
       SLACK_WEBHOOK_URL: ${SLACK_WEBHOOK_URL}
       ORDNANCE_SURVEY_API_KEY: ${ORDNANCE_SURVEY_API_KEY}
       MINIO_PORT: ${MINIO_PORT}
+      CORS_ALLOWLIST: ${EDITOR_URL_EXT}
```
DafyddLlyr, line 144: As this is just a "derived" env var, we actually don't need to add it to the `.env` files and we can just populate it here.

If this is unclear / unhelpful I can add the the `.env` and `.env.example` files.

jessicamcinchak (approved):
Pizza working as expected for me, thanks for getting to the bottom of this one!

PR #2668, opened 2024-01-16
fix(css): Change `first-child` to `nth-child(1)`

DafyddLlyr:
## What does this PR do?
 - Fixes noisy React warning (in dev-mode console, and tests) warning about use of `:first-child` selector
 - Replaces with `:nth-child(1)` selector

image


Mike-Heneghan (approved):
Thanks for fixing this @DafyddLlyr!

It's working as expected for me 😌

PR #2669, opened 2024-01-16
feat: Wire up design settings (button colour)

DafyddLlyr:


ianjon3s:
If we include a contrast checker here is must be able to check whether the selected colour contrasts with white or black.

Currently when selecting a light colour (for example Camden's button colour of #ead95b), the text in the button is correctly selected as black, but the form cannot be saved due to validation.
image

I think we may be best placed to do without a contrast checker here.

Mike-Heneghan:
Trying this out on the pizza I noticed that the new form does update the DB but the change isn't seen in the flows. 

Is that part of the scope of this or is that work elsewhere / to be done later?

Mike-Heneghan (approved):
Looks great and working as expected for me on the pizza now 🥳

PR #2660, opened 2024-01-12
feat: Wire up design settings (theme colour & logo)

DafyddLlyr:
## What does this PR do?
 - Split up `DesignSettings` component into reusable components
 - Wire up theme and logo form. This PR is primarily concerned with updating values in DB and `primaryColour` in theme, but is implemented throughout the application consistently in https://github.com/theopensystemslab/planx-new/pull/2658
 - `SHOW_TEAM_SETTINGS` feature flag still required

## To test
 - I can update theme colour
 - I can update logo
 - Preview is "live" and works as expected
 - "Reset" button works as expected

image

PR #2670, opened 2024-01-16
feat: Wire up design settings (text link colour)

DafyddLlyr:
## What does this PR do?
 - Follows on from #2660 
 - Wires up text link colour form

image


ianjon3s:
Testing this on both the pizza link and locally, if a colour with adequate contrast is selected, the colour saves and is applied correctly in the app.

If a failing contrast colour is selected, the save button takes no effect, however there is no warning given about contrast ratio.

DafyddLlyr:
> If a failing contrast colour is selected, the save button takes no effect, however there is no warning given about contrast ratio.

Good catch thank you! Not sure how I missed this sorry 😅 Taking a look now...

ianjon3s (approved):
Yes, working for me now locally 🎉

PR #2683, opened 2024-01-19
feat: Freeze rich text popover

DafyddLlyr:
Please see https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1705661907092319 for context (OSL Slack)

This PR fixes the location of the rich text popover. A look through tiptap and tippy.js docs pointed at this solution which feels a little hacky but does the job!

https://github.com/theopensystemslab/planx-new/assets/20502206/6423b674-3619-4083-a6a0-c5683f1420f3


Mike-Heneghan (approved):
Working as expected for me on the pizza 👍

PR #2695, opened 2024-01-23
chore: Tidy up a few Editor types

DafyddLlyr:
Just picking a few of these up quickly - will aim to (very slowly!) chip away at a few of these.


jessicamcinchak (approved):
:100: :broom:

PR #2696, opened 2024-01-23
chore: account for title boundaries in document templates

jessicamcinchak:
See https://github.com/theopensystemslab/planx-core/pull/269

Not a blocker to merging public-facing aspects of title boundaries to production, but we do want to make sure any maps displayed on document templates in submission zips have proper attribution sooner than later.

To test:
- This small test flow sends to email devops@ https://2696.planx.pizza/testing/document-templates-test/preview - check the inbox for a recent existing submission or send your own
- In this example, each document should have correct attribution & clearly communiate that the user "Accepted the title boundary" - for visual maps this means a source line below the map & double OS licenses, and for .geojson this means a custom key `properties.planx_user_action`

### DIFF
--- a/editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
+++ b/editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
@@ -126,7 +126,7 @@ export default function PlotNewAddress(props: PlotNewAddressProps): FCReturn {
           drawType="Point"
           geojsonData={JSON.stringify(props?.boundary)}
           geojsonColor="#efefef"
-          geojsonBuffer="10"
+          geojsonBuffer={10}
           resetControlImage="trash"
           showScale
           showNorthArrow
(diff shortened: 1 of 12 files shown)


editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
```diff
@@ -126,7 +126,7 @@ export default function PlotNewAddress(props: PlotNewAddressProps): FCReturn {
           drawType="Point"
           geojsonData={JSON.stringify(props?.boundary)}
           geojsonColor="#efefef"
-          geojsonBuffer="10"
+          geojsonBuffer={10}
```
DafyddLlyr, line 129: Should have spotted this on the map PR sorry - but would this be a breaking change for the map component?

I'm very certain it's only us using this prop currently so it's not a blocker here right now, but maybe we should have gone to v0.8.0

I'd be happy with continuing as-is and we can maybe just let BOPS know tomorrow to be aware if they're relying on this. Linting should catch it anyway before it causes too many headaches!

DafyddLlyr (approved):

DafyddLlyr (commented):
Sorry - hit approve instead of comment. I think we're missing a bump for the API package.json also (not a blocker tbh!)

Once tests are passing on CI I'll manually test as-per instructions 👍

editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
DafyddLlyr: Thanks for the explanation - agreed that it's helpful yet headachey 😅

DafyddLlyr (approved):

PR #2697, opened 2024-01-24
fix: Add bg colour to favicon `ImgInput` [skip pizza]

DafyddLlyr:
Quick fix I noticed when watching @ianjon3s video!

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/DesignSettings/FaviconForm.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/DesignSettings/FaviconForm.tsx
@@ -62,6 +62,7 @@ export const FaviconForm: React.FC<FormProps> = ({
           <InputRowItem width={formik.values.favicon ? 90 : 50}>
             <ImgInput
               img={formik.values.favicon || undefined}
+              backgroundColor={formik.values.primaryColour}
               onChange={updateFavicon}
               acceptedFileTypes={{
                 "image/png": [".png"],


ianjon3s:
I'm not sure about this one, the background of the browser tab will depend on the browser in use (and will most commonly be a light shade of grey for most browsers used without customisation), so this could potentially be misleading. An ideal favicon would be set up to work against either dark or light.

editor.planx.uk/src/pages/FlowEditor/components/Settings/DesignSettings/FaviconForm.tsx
```diff
@@ -62,6 +62,7 @@ export const FaviconForm: React.FC<FormProps> = ({
           <InputRowItem width={formik.values.favicon ? 90 : 50}>
             <ImgInput
               img={formik.values.favicon || undefined}
+              backgroundColor={formik.values.primaryColour}
```
Mike-Heneghan, line 65: _nit:_

I'm being super pedantic here but am I right in saying the favicon won't actually be set in a place where the background will be in effect? As in it'll be on a background which they've set their browser to?

Mike-Heneghan: I'm just wondering whether the `primaryColour` might still not have enough contrast like we saw in the video with the OSL favicon?

DafyddLlyr:
Thanks for the feedback - all valid points. Leaving as-is because this would be most accurate for the user.

PR #2698, opened 2024-01-24
fix: Post correct payload to `/download-application` endpoint

DafyddLlyr:
## What's the problem?
- Issue raised by Stephen here - https://opendigitalplanning.slack.com/archives/C0241GWFG4B/p1706093472655149 (ODP Slack)
- Payload posted to `/download-application` endpoint does not match the definition set in the API (which expects a `data` key)
- This was introduced here and never caught / tested properly (sorry!) - https://github.com/theopensystemslab/planx-new/pull/2422

### To test 
- Start an application on a submission service on staging
- Save, then click "Download your application (.csv)"
- This will fail and return a 400 plus invalid CSV ❌ 
- Start an application on a submission service on Pizza
- Save, then click "Download your application (.csv)"
- This will return a 200 plus valid CSV ✅

### DIFF
--- a/editor.planx.uk/src/ui/public/FileDownload.tsx
+++ b/editor.planx.uk/src/ui/public/FileDownload.tsx
@@ -17,7 +17,11 @@ const Root = styled(Box)(({ theme }) => ({
 }));
 
 // Render a button which lets the applicant download their application data as a CSV
-export default function FileDownload(props: Props): FCReturn {
+export default function FileDownload({
+  data,
+  filename,
+  text,
+}: Props): FCReturn {
   const downloadCsv = (filename: string, content: string) => {
     const csv = "data:text/csv;charset=utf-8," + content;
     const data = encodeURI(csv);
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/ui/public/FileDownload.tsx
```diff
@@ -17,7 +17,11 @@ const Root = styled(Box)(({ theme }) => ({
 }));
 
 // Render a button which lets the applicant download their application data as a CSV
-export default function FileDownload(props: Props): FCReturn {
+export default function FileDownload({
+  data,
+  filename,
+  text,
+}: Props): FCReturn {
```
Mike-Heneghan, line 24: I really like how the props are extracted here!

Mike-Heneghan (approved):
Thanks for the thorough testing notes @DafyddLlyr!

I've replicated the error on staging and then successfully downloaded the application as a csv locally using your branch 👍

PR #2700, opened 2024-01-24
feat: Add required fields for question and checklist options

ianjon3s:
# What does this PR do?

Previously August reported inconsistencies with how we require input for user-added fields in components.

The current behaviour for questions and checklists is that if an option is added and left blank it is deleted when the component is saved. The PR introduces required fields on user-added options for questions and checklists.

This brings it in line with the following components, which all require input on user-added fields:
- Next steps
- Upload + label
- Task list
- Confirmation (next steps)


Mike-Heneghan (approved):
Working as expected for me on the pizza 👍

PR #2703, opened 2024-01-25
feat: Setup BOPS tokens

DafyddLlyr:
## What does this PR do?
 - Adds `team.staging_bops_secret` and `team.production_bops_secret` columns
 - Reads BOPS token from DB using the planx-core `team.getIntegrations()` helper function
 - Adds `ENCRYPTION_KEY` env var for encrypting / decrypting secrets
 - Adds script to generate secrets - currently just in the CLI to copy/paste to DB, but I could see the `encrypt()` function getting called in future if we ever allow team admins to set these values.
 - Relies on https://github.com/theopensystemslab/planx-core/pull/281

### To Do
 - [x] Add encrypted values to Pizza
 - [x] Test BOPS submissions on Pizza

### Next PR (post-prod merge)
 - [ ] Update sync script to pull new secrets
 - [ ] Add encrypted `staging_bops_secret` values to prod db, sync data to staging
 - [ ] Test BOPS submissions on staging


api.planx.uk/modules/send/bops/bops.ts
```diff
@@ -46,19 +46,12 @@ const sendToBOPS = async (req: Request, res: Response, next: NextFunction) => {
...
-  const bopsSubmissionURL = await $api.team.getBopsSubmissionURL(
-    localAuthority,
+
+  const { bopsSubmissionURL, bopsToken } = await $api.team.getIntegrations({
+    slug: localAuthority,
+    encryptionKey: process.env.ENCRYPTION_KEY!,
     env,
-  );
-  const isSupported = Boolean(bopsSubmissionURL);
-  if (!isSupported) {
```
DafyddLlyr, line 54: This check can be safely dropped as it's being done in `planx-core`

api.planx.uk/modules/send/bops/bops.ts
```diff
@@ -69,7 +62,7 @@ const sendToBOPS = async (req: Request, res: Response, next: NextFunction) => {
       adapter: "http",
       headers: {
         "Content-Type": "application/json",
-        Authorization: `Bearer ${process.env.BOPS_API_TOKEN}`,
+        Authorization: `Bearer ${bopsToken || process.env.BOPS_API_TOKEN}`,
```
DafyddLlyr, line 65: This fallback can be removed once BOPS deprecate the old key on staging and production. We can then add a check that this is set before submitting.

api.planx.uk/server.ts
```diff
@@ -1,4 +1,5 @@
 import "isomorphic-fetch";
+import "express-async-errors";
```
DafyddLlyr, line 2: This addresses an issue we've been having with errors thrown in `planx-core`. 

Express doesn't natively catch async errors (docs: https://expressjs.com/en/guide/error-handling.html) which is likely the reason a number aren't "bubbling up" correctly from `planx-core` correctly.

Without this added, the currently tests fail (unless I wrap another try/catch block around `team.getIntegrations()`) - I'd expect that we will now see an uptick of API errors now being caught by Airbrake 🤞

scripts/encrypt/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "name": "encrypt",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.ts",
+  "scripts": {
+    "encrypt": "ts-node index.ts"
+  },
+  "keywords": [],
+  "dependencies": {
+    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#ce539b0",
```
DafyddLlyr, line 11: I'd vote that we don't worry about keeping this up to date regularly. It's only using one exported function and we can update this more ad-hoc than the other projects without consequence.

DafyddLlyr:
[Use per-team tokens for BOPS submissions](https://trello.com/c/frSlxfqj/2743-use-per-team-tokens-for-bops-submissions)

api.planx.uk/modules/send/bops/bops.test.ts
```diff
@@ -259,9 +261,11 @@ describe(`sending an application to BOPS v2`, () => {
       .post("/bops-v2/unsupported-team")
       .set({ Authorization: process.env.HASURA_PLANX_API_KEY! })
       .send({ payload: { sessionId: "123" } })
-      .expect(400)
+      .expect(500)
       .then((res) => {
-        expect(res.body.error).toMatch(/not enabled for this local authority/);
```
jessicamcinchak, line 264: when we get around to reading Pay tokens from integrations - we'll need to remember that I think we do a bit of frontend error message string matching based on this old language to show an error on the Pay component - but shouldn't effect submission destinations as these just log to Hasura !

hasura.planx.uk/migrations/1706201208594_alter_table_public_team_integrations_add_column_staging_bops_secret/up.sql
```diff
@@ -0,0 +1,9 @@
+alter table "public"."team_integrations" add column "staging_bops_secret" text
+ null;
+
+comment on column "public"."team_integrations"."staging_bops_secret" is E'Secret required for BOPS submission in the format <API_KEY>:<INITIALIZATION_VECTOR>';
+
+alter table "public"."team_integrations" add column "production_bops_secret" text
+ null;
+
+comment on column "public"."team_integrations"."production_bops_secret" is E'Secret required for BOPS submission in the format <API_KEY>:<INITIALIZATION_VECTOR>';
```
jessicamcinchak, line 9: Helpful reminder :+1:

scripts/encrypt/index.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @param secret - The secret to be encrypted.
+ * @param encryptionKey - The encryption key - a 32-byte string
+ * @returns The encrypted secret and initialization vector in the format ${secret}:${iv}
+ * @example pnpm encode <secret> <encryptionKey> 
+ */
+function main() {
+  try {
+    if (process.argv.length < 4) {
+      console.error("Usage: pnpm encode <secret> <encryptionKey>");
```
jessicamcinchak, line 16: Is there any risk to logging out the `encryptionKey` / we generally avoid this with other env vars ? AWS logs are securely behind SSO I guess?

jessicamcinchak (approved):
Looks good & working as expected for me! 

Successfully submitted session `a5ad7705-6964-42c9-89ad-23de725ce363` to Lambeth BOPS staging :heavy_check_mark:

hasura.planx.uk/migrations/1706201208594_alter_table_public_team_integrations_add_column_staging_bops_secret/up.sql
DafyddLlyr: I actually have a small PR for this from the previous creation of `team_integrations` https://github.com/theopensystemslab/planx-team-onboarding/pull/13/files

I'll pick up the remaining updates shortly also 👍

scripts/encrypt/index.ts
DafyddLlyr: This is actually just a straight up string, no templating so there's no real risk here I think.

image

jessicamcinchak: aha of course - thanks for clarifying - I read as a template out of habit !

hasura.planx.uk/migrations/1706201208594_alter_table_public_team_integrations_add_column_staging_bops_secret/up.sql
DafyddLlyr: Next PR here - https://github.com/theopensystemslab/planx-team-onboarding/pull/14

PR #2706, opened 2024-01-26
chore: setup staging GOV_UK_PAY_TOKEN_GLOUCESTER

Mike-Heneghan:
## What:

- [x] Add `GOV_UK_PAY_TOKEN_GLOUCESTER` to pulumi staging stack
- [x] Update `.env.example` and `server.ts` accordingly. 
- [x] Locally pull `pizza-secrets`, add `GOV_UK_PAY_TOKEN_GLOUCESTER` to .env and push `pizza-secrets` to allow sandbox gov uk pay to work on pizzas
- [x] Add Medway and Gloucester to `teams` to allow gov pay uk to work on staging
- [x] Refactor `generateTeamSecrets` to only generate Medway and Gloucester secrets on staging to avoid errors on prod

## Why:

- Gloucester have shared their sandbox gov up pay api key
- We want this to be wired up on staging to allow testing
- When pairing with @DafyddLlyr realised that Medway and Gloucester need to be added to `teams.ts` to allow the key to work on staging: https://editor.planx.dev/medway/gov-pay-staging-test/preview?analytics=false
- This change would cause errors on prod as Medway and Gloucester don't have prod gov up pay api keys
- Refactor of `generateTeamSecrets` to handle the difference in staging/prod creds

## Testing:

- [x] Gov Pay works on pizza for Gloucester: https://2706.planx.pizza/gloucester/gov-pay-pizza-test/preview?analytics=false
- [ ] Gov Pay works on staging for Medway and Gloucester (post merge)
- [ ] Gov Pay work on prod for Buckinghamshire, Lambeth and Southwark but not Medway and Gloucester (post merge)


infrastructure/application/utils/generateTeamSecrets.ts
```diff
@@ -25,6 +25,7 @@ export const generateTeamSecrets = (
 ): awsx.ecs.KeyValuePair[] => {
   const secrets: awsx.ecs.KeyValuePair[] = [];
   teams.forEach((team) => {
+    if (env === "production" && team?.stagingOnly) return;
```
Mike-Heneghan, line 28: _Comment:_

I think this is a simple change and should work although I guess we can't test it until it's on staging/prod?

infrastructure/common/teams.ts
```diff
@@ -14,6 +14,7 @@ interface Team {
   name: string;
   // Must match "name" in the Digital Land dataset "local-authority-district" (https://www.planning.data.gov.uk/dataset/local-authority-district)
   uniformInstances?: string[];
+  stagingOnly?: boolean;
```
DafyddLlyr, line 17: nit: To be totally explicit here we should link this to GovPay, e.g. `govPayStagingOnly`

DafyddLlyr (changes requested):
One small suggestion!

infrastructure/common/teams.ts
Mike-Heneghan: I did consider this although I think as it's at a top level it would skip all secret generation on prod? Although we're only using it at the pay level it's not specific to gov pay? 

I'm happy to go with that though as I guess it does better match how we're using it right now?

Mike-Heneghan: I think I maybe misunderstood your suggestion, I'll refactor to put it inline with the `uniformInstances` pattern 👍

DafyddLlyr: Perfect!

DafyddLlyr (approved):

PR #2711, opened 2024-01-29
chore: Drop `BOPS_SUBMISSION_URL_<TEAM>` env vars from `.env` files

DafyddLlyr:
Tidy up I should have picked up sooner! These are now redundant as of https://github.com/theopensystemslab/planx-new/pull/2581 - these values are now read from the DB.


Mike-Heneghan (approved):
Looks good to me 👍

PR #2713, opened 2024-01-29
fix: don't link to locked or sanitised sessions in resume emails

jessicamcinchak:
There's currently `lowcal_sessions` that are neither submitted, locked, deleted, nor sanitised and created before our expiry timeframe (28 days) on production - which explains the buggy resume email I received (screenshot in attached Trello ticket). 

Eg try this query: 
```
query MyQuery {
  lowcal_sessions(where: {sanitised_at: {_is_null: true}, deleted_at: {_is_null: true}, submitted_at: {_is_null: true}, email: {_eq: "jessica@opensystemslab.io"}, locked_at: {_is_null: true}}, order_by: {created_at: asc}) {
    id
    created_at
    flow {
      slug
      team {
        slug
      }
    }
  }
}
```

Changes:
- Adds `locked_at` condition to resume template which should legitimately be there for invite to pay users, adds `sanitised_at` as a cautionary fallback
- Adds an extra check when prepping template content to skip expired sessions, updates tests accordingly
- Updates the Hasura webhook query to find sessions ready for sanitation that should retroactively handle any outstanding sessions on next/future runs now


jessicamcinchak:
[Ensure "Resume" email template only includes active sessions](https://trello.com/c/ca729ZZb/2746-ensure-resume-email-template-only-includes-active-sessions)

api.planx.uk/modules/webhooks/service/sanitiseApplicationData/operations.ts
```diff
@@ -68,8 +68,12 @@ export const getExpiredSessionIds = async (): Promise<string[]> => {
     query GetExpiredSessionIds($retentionPeriod: timestamptz) {
       lowcal_sessions(
         where: {
-          submitted_at: { _lt: $retentionPeriod }
           sanitised_at: { _is_null: true }
+          _or: [
+            { deleted_at: { _lt: $retentionPeriod } }
+            { submitted_at: { _lt: $retentionPeriod } }
+            { locked_at: { _lt: $retentionPeriod } }
+          ]
```
jessicamcinchak, line 76: Previously, I think this is would have failed to find _un-submitted_ sessions older than the retention period? Do these updated conditions seem correct to others as well?

DafyddLlyr: Yep - this makes sense - the previous condition is relying on `submitted_at` being set

DafyddLlyr (approved):
Thanks for getting to the bottom of this rather opaque one! Additional tests always a welcome site also thank you 👌

PR #2716, opened 2024-01-30
feat: scroll to feedback view on view change

Mike-Heneghan:
## What

Main Feedback:
- Add ref and useEffect to scroll down when it changes and isn't the "banner" or "thanks"

MoreInfo Feedback:
- Add ref and useEffect to scroll down when it changes is "input"

## Why

- When the currentFeedbackView changes the component can render partially off screen

## Risks

- This might have a negative impact on accessibility / usability?

## Screen recording

Before:

https://github.com/theopensystemslab/planx-new/assets/36415632/9ec4c454-b8c1-41e7-8bdd-adf20ae0e4bd

After:

https://github.com/theopensystemslab/planx-new/assets/36415632/bb9df931-e6c5-411f-9e38-eb9bd26cdbed

### DIFF
--- a/editor.planx.uk/src/components/Feedback/MoreInfoFeedback.tsx
+++ b/editor.planx.uk/src/components/Feedback/MoreInfoFeedback.tsx
@@ -9,7 +9,7 @@ import {
   getInternalFeedbackMetadata,
   insertFeedbackMutation,
 } from "lib/feedback";
-import React, { useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import FeedbackOption from "ui/public/FeedbackOption";
 
 import { FeedbackFormInput, UserFeedback } from ".";
@@ -36,6 +36,16 @@ const MoreInfoFeedbackComponent: React.FC = () => {
   const [currentFeedbackView, setCurrentFeedbackView] =
     useState<View>("yes/no");
   const [feedbackOption, setFeedbackOption] = useState<Sentiment | null>(null);
+  const feedbackComponentRef = useRef<HTMLDivElement | null>(null);
+
+  useEffect(() => {
+    if (currentFeedbackView === "input") {
+      feedbackComponentRef.current?.scrollIntoView({
+        behavior: "smooth",
+        block: "start",
+      });
+    }
+  }, [currentFeedbackView]);
 
   const handleFeedbackOptionClick = (event: Sentiment) => {
     switch (event) {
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/components/Feedback/index.tsx
```diff
@@ -364,7 +375,11 @@ const Feedback: React.FC = () => {
     }
   }
 
-  return <Feedback />;
+  return (
+    <div ref={feedbackComponentRef}>
```
DafyddLlyr, line 379: nit: Throughout the codebase we tend to use MUI components, could we change this `div` to a `Box` for consistency?

DafyddLlyr (approved):
Super nice improvement 👌 Make a big difference!

editor.planx.uk/src/components/Feedback/index.tsx
```diff
@@ -98,6 +98,28 @@ const Feedback: React.FC = () => {
...
   }, [breadcrumbs]);
 
+  const feedbackComponentRef = useRef<HTMLDivElement | null>(null);
+
+  const shouldScrollToView = () => {
+    switch (currentFeedbackView) {
+      case "banner":
+        return false;
+      case "thanks":
+        return false;
```
DafyddLlyr, line 108:
```suggestion
      case "banner":
      case "thanks":
        return false;
```

DafyddLlyr: nit: Sorry - small nitpick but a bit more readable / consistent with rest of codebase.

editor.planx.uk/src/components/Feedback/MoreInfoFeedback.tsx
```diff
@@ -36,6 +36,16 @@ const MoreInfoFeedbackComponent: React.FC = () => {
...
+  const feedbackComponentRef = useRef<HTMLDivElement | null>(null);
+
+  useEffect(() => {
+    if (currentFeedbackView === "input") {
+      feedbackComponentRef.current?.scrollIntoView({
+        behavior: "smooth",
+        block: "start",
+      });
+    }
+  }, [currentFeedbackView]);
```
DafyddLlyr, line 48: A custom hook might be a slightly cleaner way of reusing this logic across both components / files?

Something like - 

```ts
const useScrollIntoView = (currentFeedbackView) => {
  const feedbackComponentRef = useRef<HTMLDivElement | null>(null);

  const shouldScrollToView = () => **logic here**
  
  useEffect(() => {
    if (shouldScrollToView()) {
      feedbackComponentRef.current?.scrollIntoView({
        behavior: 'smooth',
        block: 'start',
      });
    }
  }, [currentFeedbackView]);

  return feedbackComponentRef;
};
```

Alternatively just a wrapper for the feedback components with this logic could work also. If this makes things more complex or there's something I'm missing here no sweat!

PR #2743, opened 2024-02-01
fix: Don't forward `showArrow` prop to MUI button

DafyddLlyr:
Quick follow on to https://github.com/theopensystemslab/planx-new/pull/2722


Mike-Heneghan (approved):
Awesome! Thanks again for picking this up @DafyddLlyr

PR #2748, opened 2024-02-02
fix: Update Pulumi encryption keys

DafyddLlyr:
I updated the key in the `.env` and tested on the Pizza, but never made the corresponding update on Pulumi for staging and prod envs (AWS).


jessicamcinchak (approved):
Thanks for quick fix!

PR #2750, opened 2024-02-02
feat: Add remove file option to modal

DafyddLlyr:
## What does this PR do?
 - Adds option to remove file when modal is open
 - Addresses feedback from recent UR

There's certainly scope here in these components to refactor and simplify some of the logic - there's a fair amount of prop drilling / shared state. 

As we're likely going to have to make changes here following the a11y audit I'm intentionally swerving this for now and we can always come back to pick this up!

image


DafyddLlyr:
[No way to delete files until after you have labelled them if you have uploaded the same one twice [upload and label component]](https://trello.com/c/MlHFMfBg/2680-no-way-to-delete-files-until-after-you-have-labelled-them-if-you-have-uploaded-the-same-one-twice-upload-and-label-component)

Mike-Heneghan (approved):
Looks good and working as expected for me on the pizza 👍

PR #2755, opened 2024-02-05
feat: Capture `FileList` to breadcrumbs

DafyddLlyr:


editor.planx.uk/src/pages/FlowEditor/lib/store/index.ts
```diff
@@ -21,12 +21,21 @@ export declare namespace Store {
   export type Store = Record<string | number | symbol, unknown>;
   export type nodeId = string;
   export type flow = Record<nodeId, node>;
+  /**
+   * State of node at time user visited it
+   * State will be stored to breadcrumbs, but not captured in passport
+   */
+  export type NodeState = {
+    type: TYPES,
```
DafyddLlyr, line 29: We could get this type by looking up breadcrumb ID in the flow, but it seemed like having it here would be a helpful shorthand?

PR #2756, opened 2024-02-05
chore: Import component types from `planx-core`

DafyddLlyr:
## What does this PR do?
 - Deletes `editor.planx.uk/src/@planx/components/types.ts`, replaces with imports from `plan-core`
 - Find & Replace to swap `TYPES.Statement` → `TYPES.Question`
 - Find & Replace to swap `TYPES.Response` → `TYPES.Answer`
 - Updates docs


Mike-Heneghan:
_nit_

I'm not sure if it's part of the scope of this feature although would we want to update the historical records for the `AnalyticsSummary` to also reflect `TYPES.Statement → TYPES.Question` && `TYPES.Response → TYPES.Answer`?

I think this might also require updates to the analytics cards if they're using either of these currently. I don't think they are at the moment but I'd need to double check.

Mike-Heneghan (approved):
This change looks good to me 👍 

It feels much better removing the duplicate file and pointing at the single source of truth. 

I ran through a guidance and a submission service on the pizza  and everything worked as expected for me. 

The only non-blocking implication of this change which I'm not sure if handled yet is the update of Analytics Data which has historical records where I think we might want to update the node types to be consistent with this change as per my comment.

hasura.planx.uk/migrations/1707239828082_update_node_type_names/down.sql
Mike-Heneghan, line 1: _nit_ 

I don't want to jump the gun but another place `node_type` is stored as a string is in the `metadata` when there are `back` or `change` events tracked as per: https://github.com/theopensystemslab/planx-new/pull/2654

DafyddLlyr: Good catch! Thanks for the link to the PR also, super helpful. I've now added migrations to cover this also.

Just as an insurance policy I'd be really grateful if you were able to test these migrations locally juuuust to make sure nothing is amiss. I've tested locally and getting the expected results 👍

hasura.planx.uk/migrations/1707239828082_update_node_type_names/down.sql
```diff
@@ -0,0 +1,8 @@
+UPDATE
+  public.analytics_logs
+SET
+  node_type = CASE
+    WHEN node_type 'Question' THEN 'Statement'
+    WHEN node_type 'Answer' THEN 'Response'
```
Mike-Heneghan, line 6:
```suggestion
    WHEN node_type = 'Question' THEN 'Statement'
    WHEN node_type = 'Answer' THEN 'Response
```

I was seeing an error with the down migration but I think it was just the `=` being missing on these lines. I found it surprisingly hard to spot the difference 😂

DafyddLlyr: Ah good catch, this was me being clumsy using `⌘ + D` I think! Fixed here [8b230c9](https://github.com/theopensystemslab/planx-new/pull/2756/commits/8b230c9b43a42818256691a6d8eeed085d66ce8d)

PR #2758, opened 2024-02-06
chore: update DrawBoundary upload key from `proposal.drawing.locationPlan` to `locationPlan`

jessicamcinchak:
This one isn't exposed as editor props, so it'll be live across services as soon as it's merged ! We'll coordinate merge with other file key changes.

Related to https://github.com/theopensystemslab/planx-core/pull/289 & https://github.com/theopensystemslab/planx-data-migrations/pull/7

BOPS logic is repeated in the editor's `Send` component still for frontend CSV generation - I'll take a look at properly refactoring using the planx-core public client today so we don't have to repeat the same find-and-replace across the v1 tagging functions.


jessicamcinchak:
[Update document types and add secondary schema for "files" when provided as data (eg Heritage Statements)](https://trello.com/c/kOBotqGJ/2755-update-document-types-and-add-secondary-schema-for-files-when-provided-as-data-eg-heritage-statements)

DafyddLlyr (approved):

PR #2759, opened 2024-02-06
chore: generate frontend CSVs using public planx-core client

jessicamcinchak:
We used to generate submission payloads client-side in the Send component, but now we do it via the API. But our frontend CSV downloads still relied on the submission methods and access to session data via the store, so we hadn't removed some of this code. By using an _authenticated_ `$public` CoreDomainClient, we can (finally) remove this duplication :scissors:

Spotted while updating file key names across BOPS & Uniform mocks - this deletion means we won't have to repeat renaming mocks & tests already updated here https://github.com/theopensystemslab/planx-core/pull/289

**Testing:**
- For any submission service (eg https://2759.planx.pizza/testing/csv-test), confirm you can download a populated CSV from the status page link after clicking "Save" at any point in a service & from the Confirmation page link


DafyddLlyr:
Excited for this one! 
image

DafyddLlyr (approved):
Perfect! Some satisfying deletions and type updates here 👍 

Testing - 
 - I can download a CSV partially through my journey from save & return page ✅ 
 - I can download CSV from confirmation page ✅ 
 - CSV formatted as expected ✅

PR #2762, opened 2024-02-06
feat: add feedback_summary db view

Mike-Heneghan:
## What

- Adds a new db view called `feedback_summary` which displays the `feedback` data enriched with data pulled out from the `user_data` column and enriched with data extracted from the associated `published_flow`

## Why

- To provide a more human readable summary of the `feedback` table enriched with `published_flow` data and with the pertinent `user_data` extracted.

## Testing

- On the [pizza](https://2762.planx.pizza/) leave feedback on services: 
- Review the [feedback_summary](https://hasura.2762.planx.pizza/console/data/default/schema/public/views/feedback_summary/browse) and check the data extracted: 
- Where possible I've tried to give fallback `keys` for accessing things like the `node_title` but there might be more opportunities to account for situations where the node data is formatted differently

### Capture Result Page Metadata 

- [Trello ticket](https://trello.com/c/Ojq2BzUj/2770-capture-result-metadata-in-internal-feedback-tool-view?completedInviteSignup=1)
- Feedback Fish adds the `resultData()` to the `componentMetadata` on the result page.
- Attempts were made to surface the result the user saw to capture similar data although this was deemed not to be feasible as it would required capturing the data at the point it was derived on the front end or replicating complex logic in the view. 
- The result data is a small subset of potential data so in discussion with @DafyddLlyr we opted to skip for now

### Capture "Report an Accuracy" type feedback

- [Trello ticket](https://trello.com/c/W8tbXnbX/2771-integrate-the-report-an-inaccuracy-type-feedback-into-new-feedback-tool)
- There are places where users have the opportunity to "Report an inaccuracy"  or "Is this information inaccurate? Tell us why"

Screenshot 2024-02-06 at 16 25 38

- This feedback is directly stored in the breadcrumbs

Screenshot 2024-02-06 at 16 33 50

- From discussion with @DafyddLlyr this is not a feature integrated with Feedback Fish so isn't required for feature parity.
- Integrating with the internal feedback feature could be a feature in the future

### DIFF
--- a/hasura.planx.uk/migrations/1707234695689_create_view_feedback_summary/up.sql
+++ b/hasura.planx.uk/migrations/1707234695689_create_view_feedback_summary/up.sql
@@ -20,10 +20,13 @@ SELECT
     published_flow_node.data ->> 'info' AS help_text,
     published_flow_node.data ->> 'policyRef' AS help_sources,
     published_flow_node.data ->> 'howMeasured' AS help_definition,
-    (fb.user_data -> 'passport' -> 'data' -> '_address' ->> 'single_line_address') AS street_address,
-    (fb.user_data -> 'passport' -> 'data' -> '_address' ->> 'uprn') AS UPRN,
+    COALESCE(
+        fb.user_data -> 'passport' -> 'data' -> '_address' ->> 'single_line_address',
+        fb.user_data -> 'passport' -> 'data' -> '_address' ->> 'title'
+    ) AS address,
+    (fb.user_data -> 'passport' -> 'data' -> '_address' ->> 'uprn') AS uprn,
     (fb.user_data -> 'passport' -> 'data' ->> 'proposal.projectType') AS project_type,
-    (fb.user_data -> 'passport' -> 'data' ->> 'property.constraints.planning') AS constraints,
+    (fb.user_data -> 'passport' -> 'data' ->> 'property.constraints.planning') AS intersecting_constraints,
     published_flow_node.data AS node_data
 FROM 
     feedback fb
--- a/hasura.planx.uk/migrations/1707234695689_create_view_feedback_summary/up.sql
+++ b/hasura.planx.uk/migrations/1707234695689_create_view_feedback_summary/up.sql
@@ -1,0 +1,46 @@
+CREATE OR REPLACE VIEW "public"."feedback_summary" AS
+SELECT 
+    fb.id AS feedback_id,
+    t.name AS team,
+    f.slug AS service_slug,
+    fb.created_at,
+    fb.node_id,
+    fb.device,
+    fb.user_context,
+    fb.user_comment,
+    fb.feedback_type,
+    fb.status,
+    fb.node_type,
+    COALESCE(
+        published_flow_node.data ->> 'title', 
+        published_flow_node.data ->> 'text', 
+        published_flow_node.data ->> 'flagSet'
+    ) AS node_title,
+    published_flow_node.data ->> 'description' AS node_text,
+    published_flow_node.data ->> 'info' AS help_text,
+    published_flow_node.data ->> 'policyRef' AS help_sources,
+    published_flow_node.data ->> 'howMeasured' AS help_definition,
+    (fb.user_data -> 'passport' -> 'data' -> '_address' ->> 'single_line_address') AS street_address,
+    (fb.user_data -> 'passport' -> 'data' -> '_address' ->> 'uprn') AS UPRN,
+    (fb.user_data -> 'passport' -> 'data' ->> 'proposal.projectType') AS project_type,
+    (fb.user_data -> 'passport' -> 'data' ->> 'property.constraints.planning') AS constraints,
+    published_flow_node.data AS node_data
+FROM 
+    feedback fb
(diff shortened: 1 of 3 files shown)


hasura.planx.uk/migrations/1707234695689_create_view_feedback_summary/up.sql
```diff
@@ -0,0 +1,46 @@
...
+    COALESCE(
+        published_flow_node.data ->> 'title', 
+        published_flow_node.data ->> 'text', 
+        published_flow_node.data ->> 'flagSet'
+    ) AS node_title,
+    published_flow_node.data ->> 'description' AS node_text,
+    published_flow_node.data ->> 'info' AS help_text,
+    published_flow_node.data ->> 'policyRef' AS help_sources,
+    published_flow_node.data ->> 'howMeasured' AS help_definition,
+    (fb.user_data -> 'passport' -> 'data' -> '_address' ->> 'single_line_address') AS street_address,
```
jessicamcinchak, line 23: nit: plain `address` is more consistent with how we call this other places. 

Perhaps also another good use case of `COALESCE` because this will differ depending if a user selects an OS address or plots their own new address:
```
COALESCE(
  fb.user_data -> 'passport' -> 'data' -> '_address' ->> 'single_line_address',
  fb.user_data -> 'passport' -> 'data' -> '_address' ->> 'title'
) as address,
```

hasura.planx.uk/migrations/1707234695689_create_view_feedback_summary/up.sql
```diff
@@ -0,0 +1,46 @@
...
+    (fb.user_data -> 'passport' -> 'data' -> '_address' ->> 'uprn') AS UPRN,
+    (fb.user_data -> 'passport' -> 'data' ->> 'proposal.projectType') AS project_type,
+    (fb.user_data -> 'passport' -> 'data' ->> 'property.constraints.planning') AS constraints,
```
jessicamcinchak, line 26: nit: can this be named to reflect that it only captures `intersecting_constraints` (which is fine & good!) because non-intersecting ones are stored in the passport separately under `_nots` for automations

hasura.planx.uk/migrations/1707234695689_create_view_feedback_summary/up.sql
```diff
@@ -0,0 +1,46 @@
...
+            (published_flows.data -> fb.node_id) -> 'data' AS data
+        FROM 
+            published_flows
+        WHERE 
+            published_flows.flow_id = fb.flow_id 
+            AND published_flows.created_at < fb.created_at
+        ORDER BY 
+            published_flows.created_at DESC
+        LIMIT 1
+    ) AS published_flow_node ON true;
```
jessicamcinchak, line 46: :+1:

hasura.planx.uk/migrations/1707234695689_create_view_feedback_summary/up.sql
```diff
@@ -0,0 +1,46 @@
...
+        published_flow_node.data ->> 'title', 
+        published_flow_node.data ->> 'text', 
+        published_flow_node.data ->> 'flagSet'
+    ) AS node_title,
+    published_flow_node.data ->> 'description' AS node_text,
+    published_flow_node.data ->> 'info' AS help_text,
+    published_flow_node.data ->> 'policyRef' AS help_sources,
+    published_flow_node.data ->> 'howMeasured' AS help_definition,
+    (fb.user_data -> 'passport' -> 'data' -> '_address' ->> 'single_line_address') AS street_address,
+    (fb.user_data -> 'passport' -> 'data' -> '_address' ->> 'uprn') AS UPRN,
```
jessicamcinchak, line 24: nit: without quotes, postgres evaluates this as lowercase `uprn` - can you update casing here for consistency/readability?

hasura.planx.uk/migrations/1707234695689_create_view_feedback_summary/up.sql
```diff
@@ -0,0 +1,46 @@
...
+    fb.user_context,
+    fb.user_comment,
+    fb.feedback_type,
+    fb.status,
+    fb.node_type,
+    COALESCE(
+        published_flow_node.data ->> 'title', 
+        published_flow_node.data ->> 'text', 
+        published_flow_node.data ->> 'flagSet'
+    ) AS node_title,
```
jessicamcinchak, line 18: smart fallbacks! :+1:

jessicamcinchak (approved):
All working as expected for me! 

A few small comments, no blockers to merging

hasura.planx.uk/migrations/1707234695689_create_view_feedback_summary/up.sql
Mike-Heneghan: I've updated for it to be `uprn` to be consistent with everything else, can I just double check this is what you were looking for rather than making it `"UPRN"`?

jessicamcinchak: yes exactly - lowercase is ideal default in my opinion, needing to quote when querying is a pain!

PR #2763, opened 2024-02-06
feat: Add `_requestedFiles` key to Passport

DafyddLlyr:
Prior art - https://github.com/theopensystemslab/planx-new/pull/2755

## What does this PR do?
 - Adds a new passport variable - `_requestedFile`
 - Keeps this updated each time we pass through an Upload or UploadAndLabel component

To do 
 - More test coverage
 - Capture files uploaded via DrawBoundary

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUpload/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUpload/Editor.tsx
@@ -62,7 +62,7 @@ function Component(props: any) {
           </InputRow>
           <InputRow>
             <Input
-              // required
+              required
               format="data"
               name="fn"
               value={formik.values.fn}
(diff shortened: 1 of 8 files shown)


editor.planx.uk/src/@planx/components/shared/utils.ts
```diff
@@ -41,7 +41,7 @@ export const makeData = <T>(
   props: any,
   value: T,
   overwriteKey?: string,
-): {} | { data: Record<string, T> } => {
+): Record<string, never> | { data: Record<string, T> } => {
```
DafyddLlyr, line 44: image

editor.planx.uk/src/@planx/components/FileUpload/Editor.tsx
```diff
@@ -62,7 +62,7 @@ function Component(props: any) {
           </InputRow>
           <InputRow>
             <Input
-              // required
+              required
```
jessicamcinchak, line 65: I agree this should be correct behavior - but I wonder if this will break any discretionary submission services like building control forms or something? Hopefully not / easy enough to fix forward if we find it does I guess? I don't necessarily think we should be putting as much proactive effort into find/replace etc outside of "standard" services?

jessicamcinchak (approved):

PR #2766, opened 2024-02-08
chore: remove feedback records which are beyond retention policy

Mike-Heneghan:
## What

- Delete `feedback` records which were created before the retention policy date
- Rename `development` -> `local` for Slack warnings

## Why

- `feedback` can contain personally identifiable information so needs to be sanitised or deleted
- Records older than 28 days are likely to have little value and wou;d already have been downloaded and processed
- The `development` name is easy to conflate with `staging` i.e. our `.dev` environment
- Renaming to `Local` explicits shows there's been a failure running sanitisation based on local environment

## Screenshot

Rename
Screenshot 2024-02-08 at 16 36 18

## Testing

- Added `feedback` records and manually changed the dates in the hasura console
- When cron job triggered the expected records were deleted successfully

### DIFF
--- a/api.planx.uk/helpers.ts
+++ b/api.planx.uk/helpers.ts
@@ -247,6 +247,10 @@ const getFormattedEnvironment = (): string => {
     const pizzaNumber = new URL(process.env.API_URL_EXT!).href.split(".")[1];
     environment += ` ${pizzaNumber}`;
   }
+  // For readability redefine development as local for slack warnings
+  if (environment === "development") {
+    environment = "local";
+  }
   return capitalize(environment);
 };
 
--- a/api.planx.uk/modules/webhooks/service/sanitiseApplicationData/operations.ts
+++ b/api.planx.uk/modules/webhooks/service/sanitiseApplicationData/operations.ts
@@ -32,6 +32,9 @@ export const getOperations = (): Operation[] => [
 
   // Queued up scheduled events (backup method to PG function/trigger)
   deleteHasuraScheduledEventsForSubmittedSessions,
+
+  // Feedback records
+  deleteFeedback,
 ];
 
 export const operationHandler = async (
@@ -291,3 +294,23 @@ export const deleteHasuraScheduledEventsForSubmittedSessions: Operation =
     const [_column_name, ...ids] = response?.result?.flat() || [];
     return ids;
   };
+
+export const deleteFeedback: Operation = async () => {
+  const mutation = gql`
+    mutation DeleteFeedback($retentionPeriod: timestamptz) {
+      feedback: delete_feedback(
+        where: { created_at: { _lt: $retentionPeriod } }
+      ) {
+        returning {
+          id
+        }
+      }
+    }
+  `;
+  const {
+    feedback: { returning: result },
+  } = await $api.client.request<{ feedback: Result }>(mutation, {
+    retentionPeriod: getRetentionPeriod(),
+  });
+  return result;
+};
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -248,6 +248,29 @@
           - user_context
           - user_data
       comment: Allow users who want to leave feedback on their experience to write to the table
+  select_permissions:
+    - role: api
+      permission:
+        columns:
+          - id
+          - team_id
+          - device
+          - user_data
+          - feedback_type
+          - node_id
+          - created_at
+          - flow_id
+          - node_type
+          - status
+          - user_comment
+          - user_context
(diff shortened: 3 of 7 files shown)


api.planx.uk/helpers.ts
```diff
@@ -247,6 +247,10 @@ const getFormattedEnvironment = (): string => {
     const pizzaNumber = new URL(process.env.API_URL_EXT!).href.split(".")[1];
     environment += ` ${pizzaNumber}`;
   }
+  // For readability redefine development as local for slack warnings
+  if (environment === "development") {
+    environment = "local";
```
Mike-Heneghan, line 252: _Context:_

I didn't expect to see a Slack message when my local test of running the sanitisation failed. I think the message was very useful but my first (incorrect) instinct was that I'd triggered something in a live environment. 

I was relieved when I saw "Development" although with it being the end of the day and being tired and I did wonder if it was referring to staging and if I'd misunderstood the `.env` setup for the webhook.

Realistically I don't think I'll make this mistake again and I'm happy to leave this as "Development" if there are any strong opinions.

api.planx.uk/modules/webhooks/service/sanitiseApplicationData/operations.ts
```diff
@@ -291,3 +294,23 @@ export const deleteHasuraScheduledEventsForSubmittedSessions: Operation =
...
+        returning {
+          id
+        }
+      }
+    }
+  `;
+  const {
+    feedback: { returning: result },
+  } = await $api.client.request<{ feedback: Result }>(mutation, {
+    retentionPeriod: getRetentionPeriod(),
```
Mike-Heneghan, line 313: _Comment_

I erroneously looked at the `getRetentionPeriod` from `pay`: editor.planx.uk/src/lib/pay.ts which is 28 days 🤦 

At the top of the file it has the period as 6 months. 

I'm wondering whether we stick to 6 months to be consistent with the rest of the sanitisation or swap it to 28 days as originally planned?

DafyddLlyr: Yep no problems here - the simpler the better 👍 

Let's make sure we communicate this to George / August / Silvia and update Notion to now reflect the changes made in this PR - https://www.notion.so/opensystemslab/Data-Retention-Policy-fcd47af2a8244888b544a1e9c92c88f8

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -248,6 +248,29 @@
           - user_context
           - user_data
       comment: Allow users who want to leave feedback on their experience to write to the table
+  select_permissions:
```
DafyddLlyr, line 251: nit/question: Are these select permissions needed for the mutation `DeleteFeedback` to work? It looks like we might need `id` for the `returning` clause but no more?

DafyddLlyr (approved):
One small comment / question, but happy for this to go ahead as-is.

I've added a checkbox to the Trello ticket to ensure we remember to update our [Data Retention Policy](https://www.notion.so/opensystemslab/Data-Retention-Policy-fcd47af2a8244888b544a1e9c92c88f8) on Notion. This could actually be done now, or after we strip out Feeback Fish.

api.planx.uk/modules/webhooks/service/sanitiseApplicationData/operations.ts
Mike-Heneghan: Great shout! I didn't think to look for a data policy on Notion. The other place is https://www.planx.uk/privacy and Alastair has advised I should give him a shout when the broader feature is ready and he'll update this. 

Another place would be the flow setting level privacy policy potentially.

hasura.planx.uk/metadata/tables.yaml
Mike-Heneghan: Ahahahaha I also wasn't sure about this 😅 

I added the full column permissions when I was having issues. I couldn't seem to get a straight answer from the docs and I saw that other sanitised tables had the full select permission. 

I can work this out by restricting access and testing it not that the feature is working.

Mike-Heneghan: Thanks again for picking up on this @DafyddLlyr I kind of wanted to do this testing but was worried I'd been taking too long on this piece of work. So it was a great excuse to try it out!

As discussed it seems the delete requires select permissions for any columns it references i.e. `created_at` for the `where` and `id` for the return value. 

I've updated the permissions to reflect this: b259adcb7545487aab16b61284acaf0a5e87ae5c

DafyddLlyr (approved):

PR #2768, opened 2024-02-08
fix: Handle additional breadcrumb info in `SummaryList` component

DafyddLlyr:
## What does this PR do?
 - Handles additional info in breadcrumbs for `FileUploadAndLabel` components added in #2763 when rendering `SummaryList` component
 - Updates tests


jessicamcinchak (approved):
Thanks for quick fix here!

PR #2770, opened 2024-02-09
feat: Move `requestedFiles()` to store

DafyddLlyr:
## What does this PR do?
 - Moves fetching, and returning default values, to a single shared store function
 - Previous attempt in https://github.com/theopensystemslab/planx-new/commit/f74f9da68edbce7406260051b8eb297fb21722ae led mysteriously failing tests - see comment on code explaining how this is now fixed

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
@@ -639,8 +639,13 @@ export const previewStore: StateCreator<
   },
 
   requestedFiles: () => {
+    // Importing PASSPORT_REQUESTED_FILES_KEY causes tests to fail - possible circular dependency issue?
+    // Repeating it here so find and replace still points to this hardcoded value
+    const PASSPORT_REQUESTED_FILES_KEY = "_requiredFiles";
+
     const { computePassport } = get();
-    const currentRequestedFiles = computePassport().data?._requestedFiles;
+    const currentRequestedFiles =
+      computePassport().data?.[PASSPORT_REQUESTED_FILES_KEY];
     const emptyFileList = { required: [], recommended: [], optional: [] };
 
     return currentRequestedFiles || emptyFileList;
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
@@ -78,7 +78,7 @@ export interface PreviewStore extends Store.Store {
   path: ApplicationPath;
   saveToEmail?: string;
   overrideAnswer: (fn: string) => void;
-  // requestedFiles: () => FileList;
+  requestedFiles: () => FileList;
 }
 
 export const previewStore: StateCreator<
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -638,15 +638,13 @@ export const previewStore: StateCreator<
...
 
-  // TODO: Fix this!
-  // requestedFiles: () => {
-  //   const { computePassport } = get();
-  //   const currentRequestedFiles =
-  //     computePassport().data?.[PASSPORT_REQUESTED_FILES_KEY];
-  //   const emptyFileList = { required: [], recommended: [], optional: [] };
+  requestedFiles: () => {
+    const { computePassport } = get();
+    const currentRequestedFiles = computePassport().data?._requestedFiles;
```
DafyddLlyr, line 643: Using `computePassport().data?.[PASSPORT_REQUESTED_FILES_KEY];` here leads to failing tests 🙃 

I guess it must be something to do with circular imports though there was no indication of that given via Jest output.

I'll added a comment / extra variable here so find and replace in future would still hopefully point at this hardcoded `_requiredFiles` value.

There might be an argument to place all `PASSPORT_*` consts in a single shared location if we hit this sort of curiosity again.

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -638,15 +638,18 @@ export const previewStore: StateCreator<
...
-  // TODO: Fix this!
-  // requestedFiles: () => {
-  //   const { computePassport } = get();
-  //   const currentRequestedFiles =
-  //     computePassport().data?.[PASSPORT_REQUESTED_FILES_KEY];
-  //   const emptyFileList = { required: [], recommended: [], optional: [] };
+  requestedFiles: () => {
+    // Importing PASSPORT_REQUESTED_FILES_KEY causes tests to fail - possible circular dependency issue?
+    // Repeating it here so find and replace still points to this hardcoded value
+    const PASSPORT_REQUESTED_FILES_KEY = "_requiredFiles";
```
DafyddLlyr, line 644: There might be an argument to move all `PASSPORT_*` consts to a single shared location if we come across a quirk like this again.

Mike-Heneghan:
Hey @DafyddLlyr, I see you've just marked this as draft but I'd already had a wee look and there was some behaviour which I didn't expect which I thought I'd run past you.

I noticed that the `_requestedFiles` get overwritten every time it's computed. 

I think that could be an issue if you were to use two components which have files associated. 

For example I uploaded a location plan in `DrawBoundary` and then used `FileUploadAndLabel` which resulted in the following:

Screenshot 2024-02-09 at 11 56 11

I _think_ what we'd want would be:

```
_requestedFiles:
  optional: ['utility']
  recommended: ['heritage']
  required: ['locationPlan', 'roof', 'site']
 ```

DafyddLlyr:
@Mike-Heneghan Yep sorry, moved this back to draft whilst I debugged! There's a test case for the behaviour you're describing which was failing.

Check out https://github.com/theopensystemslab/planx-new/pull/2770/commits/9cd4a754758e484dc37fe225f1cdc8b78dd6f116 🤦 

Going to make a coffee (clearly need it!) and then I'll come back and manually re-test once Pizza is up.

PR #2773, opened 2024-02-09
feat: Move add service button to top

DafyddLlyr:
No strong opinions here, just something to try out. A bit fed up of scrolling down so far to make a new service! 😅 

image


jessicamcinchak (approved):
here for this :heavy_plus_sign:

PR #2779, opened 2024-02-13
feat: Remove sandbox account

DafyddLlyr:
## What does this PR do?
 - Retires the PlanX "sandbox" infrastructure stack

## Steps taken
 - `pulumi destroy --stack sandbox`  on all layers (application, data, networking, certificates)
   - This removed most resources, but not all frustratingly. This means there will likely still be a small bill for the next 90 days.
 - `pulumi stack rm sandbox` on all layers
 - Close AWS account (this gets "suspended" for 90 days and is then closed)
 - Remove any DNS records on Cloudflare
 - Remove references to "sandbox" in codebase

## To test / verify
 - Pulumi dashboard no longer lists "sandbox" stack
 - AWS account shows as suspended when you attempt to log in
 - Unable to access Sandbox credentials via AWS SSO

image


Mike-Heneghan (approved):
Looking good to me 👍 

I had a wee look on Cloudflare and it seems like `planx.in` is still active but I thought that  was the one associated with the sandbox? 

I might be getting the wrong end of the stick and I think it's non-blocking regardless.

DafyddLlyr:
That's correct - the domain is still ours and registered in CloudFlare (I'll raise with Helen & Alastair), but all DNS records and Zero Trust rules have been removed.

PR #2781, opened 2024-02-13
fix: format Airbrake API errors for more readable Slack

jessicamcinchak:
Context https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1707753609576929

Before:
Screenshot from 2024-02-13 17-25-39

After:
Forgot that Airbrake won't log on pizzas, let's merge to staging & see !


DafyddLlyr (approved):
> Forgot that Airbrake won't log on pizzas, let's merge to staging & see !

This works for me!

PR #2783, opened 2024-02-13
fix: fetch & flatten the latest published version of external portal data when publishing the parent flow

jessicamcinchak:
**We're changing the rules of how we "publish" flows:** 
- Currently, "publishing" snapshots the draft content of the parent flow, and the draft content from each external portal (regardless of published status) into a single "published version" of that parent flow
- Now, the parent flow can only be published if each of its' external portals are also published
  - This means "publishing" now snapshots the draft of the parent flow, and the published content of each external portal
- This also introduces three unique routes for viewing your flow content in the Editor: 
  1. `/preview` (next &rarr; `/published`) - the latest published version of the parent flow
      - Icon disabled and displays NotFound if not published yet
  2. `/amber` (next &rarr; `/preview`) - the draft of the parent, plus the latest published version of each external portal
      - Reflects exactly what you'll get when you click publish next
      - Displays an error warning and NotFound if any external portals are unpublished (same error as when you click "Check for changes to publish")
  3. `/draft` - the draft of the parent, plus the draft of each external portal
      - Useful for testing automations and content changes that cross many different flows
      - Hidden unless your user role is `platformAdmin` 
      
**Key code changes:**
- [API & Editor] Updates `dataMerged` flattening logic (incl tests) in the API ONLY and removes the duplicate definition from the Editor
- [API] Adds a public GET endpoint `/flow/:flowId/flatten-data` (incl tests and Swagger doc) that can be called by the Editor views `/amber` & `/draft` instead of relying on the duplicate definition of `dataMerged`
- [API] Updates existing tests for `validateAndDiff` & `publish` to reflect new rules and error handling, and draw from consistent mock data
- [Editor] Aligns views to 'next' route names (`draft`, `preview`, `published`)
- [Editor] MVP simplified diff message of changes to publish and notification badges on publish button

**Testing notes:**
- publish succeeds when external portals are published
- publish throws a meaningful error when external portals are not published
- `/published` loads published flow and published external portal data
  -  `/preview` redirects to `/published`
- `/amber` loads draft flow and published external portal data
  - `/unpublished` redirects to `/draft`
- `/draft` loads draft flow and draft external portal data

**Future scope:**
- Redirect `/amber` &rarr; `/preview` after ~30 days (will set Slack channel reminder)
- Addressing blank screens when reaching portals in the editor sidebar
- Concept of when a flow is "live"


jessicamcinchak:
[Adjust "publish" function to read from published-version of external portals, rather than live version](https://trello.com/c/CzjnZsMu/2744-adjust-publish-function-to-read-from-published-version-of-external-portals-rather-than-live-version)

api.planx.uk/modules/flows/downloadSchema/controller.ts
```diff
@@ -20,7 +20,7 @@ export type DownloadFlowSchemaController = ValidatedRequestHandler<
 >;
 
 export const downloadFlowSchemaController: DownloadFlowSchemaController =
-  async (_res, res, next) => {
+  async (_req, res, next) => {
```
jessicamcinchak, line 23: this is an unrelated typo fix - no functional change here !

api.planx.uk/helpers.ts
```diff
@@ -4,15 +4,32 @@ import { Flow, Node } from "./types";
 import { ComponentType, FlowGraph } from "@opensystemslab/planx-core/types";
 import { $public, getClient } from "./client";
 
+export interface FlowData {
+  slug: string;
+  data: Flow["data"];
```
jessicamcinchak, line 9: This, like many other helper functions here, are still using outdated `Flow` & `Node` types from our API definitions rather than planx-core's `FlowGraph` & `Node`. I'll plan to pick up this type migration in a separate PR as it's going to introduce a fair bit of noise, and possibly require an update to the `Node` currently defined in planx-core.

editor.planx.uk/src/routes/views/publishedPreview.tsx
```diff
@@ -0,0 +1,63 @@
...
+  state.setFlowNameFromSlug(flowSlug);
+  state.setGlobalSettings(data.globalSettings[0]);
+  state.setFlowSettings(flow.settings);
+  state.setTeam(flow.team);
+
+  return (
+    <PublicLayout>
+      <SaveAndReturnLayout>
+        <View />
+      </SaveAndReturnLayout>
```
jessicamcinchak, line 42: I'm choosing to use the `<SaveAndReturnLayout>` for `/publish-preview` which means a lowcal session id will be generated and saved. 

`/unpublished` does _not_ currently use this layout which explains a number of the "cannot find session id" errors on production lately around CSV generation - do we want to make a decision to use _everywhere_?

DafyddLlyr (commented):
Just done a code-only review with a few small comments / nits, will test on Pizza now 👍

editor.planx.uk/src/routes/views/publishedPreview.tsx
DafyddLlyr: Thanks for getting to the bottom of this - for consistency I'd say we use it everywhere. I think this means we can hoist it higher up the component tree. Maybe it can just be merged into `PublicLayout`?

DafyddLlyr:
Had a good play around with this and all is working as expected on Pizza ✅ 

- [x] publish succeeds when external portals are published
- [x] publish throws a meaningful error when external portals are not published
- [x] `/preview loads` published flow and published external portal data
- [x] `/publish-preview` loads draft flow and published external portal data
- [x] `/unpublished` loads draft flow and draft external portal data

api.planx.uk/modules/flows/routes.ts
```diff
@@ -71,4 +74,10 @@ router.get(
   downloadFlowSchemaController,
 );
 
+router.get(
+  "/flows/:flowId/flatten-data",
+  validate(flattenFlowData),
+  flattenFlowDataController,
+);
```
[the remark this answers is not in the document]
jessicamcinchak: Right - this endpoint is public because `flows.data` & `published_flows.data` are publicly select-able in our database. Nothing to do with submission functionality here, that's handled by the frontend views & routing.

I've already been finding this endpoint to be a handy utility for copying flow data, generating mock flow data for tests etc.

api.planx.uk/modules/flows/flattenFlow/controller.ts
```diff
@@ -0,0 +1,48 @@
...
+import { getFlattenedFlowData } from "./service";
+
+type FlattenFlowDataResponse = FlowGraph;
+
+export const flattenFlowData = z.object({
+  params: z.object({
+    flowId: z.string(),
+  }),
+  query: z.object({
+    unpublished: z.string().optional(), // TODO make boolean
```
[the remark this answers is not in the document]
jessicamcinchak: This works, thanks for spotting!

editor.planx.uk/src/routes/views/publishedPreview.tsx
```diff
@@ -0,0 +1,63 @@
...
+  } catch (error) {
+    console.log(error);
+    if (error instanceof AxiosError) {
+      alert(
+        `Cannot open /publish-preview, navigate back to the graph to keep editing. \n\n${error.response?.data?.error}`,
+      );
+      throw new NotFoundError();
+    } else {
+      throw new NotFoundError();
+    }
```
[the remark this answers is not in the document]
jessicamcinchak: Yep, updated !

editor.planx.uk/src/routes/views/publishedPreview.tsx
jessicamcinchak: Slept on this one and having a few doubts that we should use `<SaveAndReturnLayout>` throughout rather than just clean up/improve our error handling for integrations that fail without a `sessionId` or `saveToEmail` on `/unpublished` & `/publish-preview` routes ? 

A couple considerations / possible side effects:
- Would this unnecessarily bloat `lowcal_sessions` records, making it more confusing to summarise analytics etc? Would we need to add an additional column to filter by editor route to get just the "real" `/preview` sessions? 
- Neither `/unpublished` or `/public-preview` load with `analytics=false` but will now have a valid `sessionId` in store and possibly create records? I didn't follow this thread too carefully through code, but feels a bit risky? 
- With the email login screen on non-all-published routes, we basically open ourselves up to the risk again that this change in publishing logic is meant to protect against - which is an application being submitted against work-in-progress content that wasn't intended for public use yet? Feels easier for an editor to confusingly share wrong link, etc?

I think I'd like to exclusively use `<SaveAndReturnLayout>` on `/preview` and revert original decision to also wrap `/publish-preview` here. Will pick up error handling in a separate followup PR if this sounds good?

DafyddLlyr: Thanks for thinking on this one - this seems sensible to me. I think the fact that we've narrowed it down to this is actually already super helpful when it comes to debugging etc, so let's keep as is for now 👍

DafyddLlyr (approved):
Changes look good thanks!

DafyddLlyr (approved):

DafyddLlyr (approved):
Reviewed routing changes ✅

PR #2787, opened 2024-02-14
chore: Sync govpay secrets

DafyddLlyr:
~~Relies on #2786 - please see this PR for context. Should be merged to `main` once the above PR is merged to `production`, and secrets populated.~~

Tested locally multiple times with `pnpm test-sync` ✅ 

Encrypted values added to prod db ✅

PR #2786, opened 2024-02-14
feat: Add GovPay columns to `team_integrations` table

DafyddLlyr:
## What does this PR do?
 - Follow on from https://github.com/theopensystemslab/planx-new/pull/2499 and https://github.com/theopensystemslab/planx-core/pull/212
 - Adds columns for GovPay secrets to `team_integrations` table
 - Does not read or use these secrets yet, this will be a follow on PR as the sequencing is a little tricky. Purposefully breaking this down into small atomic steps will hopefully make things a bit simpler.

## Next steps
- Populate columns with encrypted secrets (will be done once columns added to production)
- Update sync scripts to read new columns - https://github.com/theopensystemslab/planx-new/pull/2787
- Read and decrypt secrets using `planx-core`. PR here - https://github.com/theopensystemslab/planx-core/pull/303
- Remove env vars
- Work out how to inject secrets into e2e tests

PR #2790, opened 2024-02-14
feat: Read GovPay tokens from `team_integrations` table

DafyddLlyr:
## Context
 - Follow up from https://github.com/theopensystemslab/planx-new/pull/2786 & https://github.com/theopensystemslab/planx-new/pull/2787
 
 ## What does this PR do?
 - Reads `GovPay` tokens from `team_integrations` table instead of env vars
 - Does not yet remove GovPay env vars as I'd like to keep them there whilst testing - this means we could safely just revert this PR if we hit issues
 - Regressions tests pass on branch 
   - ✅ Locally 
   - ✅ CI - https://github.com/theopensystemslab/planx-new/actions/runs/7913688762

## Next steps
 - [x] Setup Camden's new GovPay token on db
 - [x]  Setup Gloucester's new GovPay token on db
 - [x] Test on Pizza
 - [ ] Test on Staging
 - [ ] Test on Prod
 - [ ] Remove env vars - Github secrets, Pulumi, `.env`, `.env.example`, `assert()` statements


e2e/tests/ui-driven/src/context.ts
```diff
@@ -327,3 +329,29 @@ async function deleteTeam(adminGQLClient: GraphQLClient, context: Context) {
     }
   }
 }
+
+async function setupGovPaySecret($admin: CoreDomainClient, context: Context) {
```
DafyddLlyr, line 333: A little bit inelegant maybe but this was the best approach I could find to set up this value in the DB!

DafyddLlyr:
As I'm on A/L next week, I'd rather not progress this just now. It can be reviewed and then I'll aim to merge once I'm back 👍

Mike-Heneghan (commented):
This looks good to me 👍 

I see there's some merge conflicts but I'm sure that'll be an easy fix. 

Am I right in saying it won't be possible to test this until it's on staging as we don't have gov pay uk sandbox setup on Pizzas and local environments?

DafyddLlyr:
> Am I right in saying it won't be possible to test this until it's on staging as we don't have gov pay uk sandbox setup on Pizzas and local environments?

Apologies for the slow reply here - this can be tested on the Pizza 👍 

The GovPay tokens are read and decrypted from the DB, and not read from the `.env` files here.

jessicamcinchak (approved):
Pizza teams with staging Pay enabled working as expected for me!

PR #2791, opened 2024-02-14
fix(types): Add better types to Airbrake logging

DafyddLlyr:
## What does this PR do?
- Improves types on `reportError()` to better match what Airbrake expects ([docs](https://github.com/airbrake/airbrake-js/tree/master/packages/browser#notice-annotations))
- Quick fix as this error did contain the expected context  - https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1707928622915259 (OSL Slack)
- A proper review of Airbrake logging should be picked up here with this ticket https://trello.com/c/TfmJiwID/2686-fix-airbrake-error-logging-on-api

### DIFF
--- a/api.planx.uk/modules/send/utils/helpers.ts
+++ b/api.planx.uk/modules/send/utils/helpers.ts
@@ -21,8 +21,7 @@ export async function logPaymentStatus({
 }): Promise<void> {
   if (!flowId || !sessionId) {
     reportError({
-      message:
-        "Could not log the payment status due to missing context value(s)",
+      error: "Could not log the payment status due to missing context value(s)",
       context: { sessionId, flowId, teamSlug },
     });
   } else {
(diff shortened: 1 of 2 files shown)


api.planx.uk/modules/send/utils/helpers.ts
```diff
@@ -38,21 +38,20 @@ export async function logPaymentStatus({
...
     }
   }
 }
 
 // tmp explicit error handling
-export function reportError(obj: object) {
+export function reportError(report: { error: any, context: object }) {
   if (airbrake) {
-    airbrake.notify(obj);
+    airbrake.notify(report);
```
DafyddLlyr, line 51: This just accepts an `any`, but the actual type is documented here - https://github.com/airbrake/airbrake-js/tree/master/packages/browser#notice-annotations

PR #2792, opened 2024-02-15
feat: Disable `/preview` for unpublished flows

DafyddLlyr:
## Context
Quick proof of concept to follow up on discussion point from yesterday about publishing flows. No strong feelings about when / how this proceeds - it might make sense to bundle a bunch of related "publish" changes together such as https://github.com/theopensystemslab/planx-new/pull/2783?

## What does this PR do?
 - Disable "preview" icon in Editor if flow is unpublished
 - Throws a `NotFound` error on `/preview` routes for unpublished flows (no fallback to `dataMerged()` / "unpublished" version of the flow)
   - If we go ahead with this it's probably worth improving this - not found for public, a meaningful message for logged in Editors maybe?
 - A few type tidy ups 🧹

### DIFF
--- a/e2e/tests/ui-driven/src/create-flow/create-flow.spec.ts
+++ b/e2e/tests/ui-driven/src/create-flow/create-flow.spec.ts
@@ -145,7 +145,45 @@ test.describe("Navigation", () => {
     await expect(nodes.getByText(noBranchNoticeText)).toBeVisible();
   });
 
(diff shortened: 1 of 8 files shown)


jessicamcinchak (approved):
Works as expected for me! 

Asked a UI coordination question in Slack very similar to what you've anticipated here - no strong feelings either way, but worth noting that I'm working on re-locating `dataMerged` to planx-core so would be nice to decide asap / avoid merge conflicts :ok_hand:

PR #2794, opened 2024-02-15
Revert "fix: format Airbrake API errors for more readable Slack"

jessicamcinchak:
Reverts theopensystemslab/planx-new#2781

Looks like this is actually a bit worse, so reverting! This should have been a "Failed to send to BOPS" error equivalent to the original PR "before" screenshot: 
Screenshot from 2024-02-15 11-43-31


DafyddLlyr (approved):
One thing I hit yesterday (sorry for not flagging this sooner) was the undocumented types which Airbrake expects - https://github.com/theopensystemslab/planx-new/pull/2791

Maybe an alternative to look at?

PR #2798, opened 2024-02-15
chore: Unwrap JSONB columns in `analytics_summary` view

DafyddLlyr:
image

## What does this PR do?
 - Unwraps `user_agent` column into `platform`, `os`, and `browser`
 - Unwraps metadata column into `change_metadata`, `back_metadata`, `help_metadata` and `selected_urls`

I'll cross check with @zz-hh-aa before we move this to production that this won't be a breaking change

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
@@ -44,6 +44,15 @@ type NodeMetadata = {
   isAutoAnswered?: boolean;
 };
 
+/**
+ * Describes the value held in analytics_logs.metadata
+ */
+type Metadata =
+  | Record<"change", NodeMetadata>
+  | Record<"back", NodeMetadata>
+  | SelectedUrlsMetadata
+  | HelpClickMetadata;
+
 let lastVisibleNodeAnalyticsLogId: number | undefined = undefined;
 
 const analyticsContext = createContext<{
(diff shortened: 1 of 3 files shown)


DafyddLlyr:
[Update `AnalyticsSummary` view to extract important data from metadata column](https://trello.com/c/V4sTDOoF/2760-update-analyticssummary-view-to-extract-important-data-from-metadata-column)

zz-hh-aa:
Had a quick look and don't think it'll break anything (but it will help with some yet-unanswered questions!).

DafyddLlyr:
@Mike-Heneghan Please feel free to take over / amend / fix / merge this branch next week so that this can progress 👍

hasura.planx.uk/migrations/1708016067772_run_sql_migration/up.sql
Mike-Heneghan, line 1: This is such a useful change 🥳

Mike-Heneghan (approved):
Thanks for picking this up @DafyddLlyr it's an awesome change!

I had one comment on the change to types which is non-blocking but I have a PR to extend it a little which I'd be keen to get your perspective on but there's no rush to add.

hasura.planx.uk/migrations/1708016067772_run_sql_migration/up.sql
Mike-Heneghan, line 1: _nit_

`run_sql_migrations` isn't a super meaningful name so it might be nice to call it something along the lines of `alter_view_analytics_summary_unfold_metadata`

PR #2805, opened 2024-02-19
Mh/metadata types update

Mike-Heneghan:
## What

- Expand on the update to types as outlined here: https://github.com/theopensystemslab/planx-new/pull/2798

## Why

- I'm not sure but I think this expands on the logic of implementing the `Metadata` type

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
@@ -28,28 +28,41 @@ const ALLOW_LIST = [
 ] as const;
 
 export type HelpClickMetadata = Record<string, string>;
-export type SelectedUrlsMetadata = Record<"selectedUrls", string[]>;
-export type BackwardsNavigationInitiatorType = "change" | "back";
+type SelectedUrlsMetadata = Record<"selectedUrls", string[]>;
+type BackwardsNavigationInitiatorType = "change" | "back";
 
-type NodeMetadata = {
+type ResultNodeMetadata = {
+  flag?: Flag;
   flagSet?: FlagSet;
   displayText?: {
     heading?: string;
     description?: string;
   };
-  flag?: Flag;
-  title?: string;
-  type?: string | null;
-  id?: string;
+};
+
+type AutoAnswerMetadata = {
   isAutoAnswered?: boolean;
 };
 
+type NodeMetadata = ResultNodeMetadata & AutoAnswerMetadata;
+
+type BackwardsTargetMetadata = {
+  id?: string;
+  type?: string | null;
+  title?: string;
+};
+
+type BackwardsNavigationMetadata =
+  | Record<"change", BackwardsTargetMetadata>
+  | Record<"back", BackwardsTargetMetadata>;
+
 /**
- * Describes the value held in analytics_logs.metadata
+ * Describes the possible values that can be written to analytics_logs.metadata
+ * by any one of the specific tracking event functions
  */
 type Metadata =
-  | Record<"change", NodeMetadata>
-  | Record<"back", NodeMetadata>
+  | NodeMetadata
+  | BackwardsNavigationMetadata
   | SelectedUrlsMetadata
   | HelpClickMetadata;
 
@@ -377,7 +390,7 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     if (shouldSkipTracking()) return;
 
     const targetNodeMetadata = nodeId ? getTargetNodeDataFromFlow(nodeId) : {};
-    const metadata: Metadata =
+    const metadata: BackwardsNavigationMetadata =
       initiator === "change"
         ? { change: targetNodeMetadata }
         : { back: targetNodeMetadata };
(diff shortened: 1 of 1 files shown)


hasura.planx.uk/migrations/1708016067772_run_sql_migration/up.sql
```diff
@@ -0,0 +1,49 @@
+DROP VIEW public.analytics_summary;
```
jessicamcinchak, line 1: I was surprised to see a migration here - but it looks to be the same migration from #2798, right? If you change your base branch to Daf's branch it may display a more meaningful comparison of changes, then if #2798 merges first it should automatically switch to `main` again & rebase ?

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -27,29 +27,50 @@ const ALLOW_LIST = [
...
+  | HelpClickMetadata;
+
 let lastVisibleNodeAnalyticsLogId: number | undefined = undefined;
 
 const analyticsContext = createContext<{
   createAnalytics: (type: AnalyticsType) => Promise<void>;
-  trackHelpClick: (metadata?: HelpClickMetadata) => Promise<void>;
-  trackNextStepsLinkClick: (metadata?: SelectedUrlsMetadata) => Promise<void>;
+  trackHelpClick: (metadata?: Metadata) => Promise<void>;
+  trackNextStepsLinkClick: (metadata?: Metadata) => Promise<void>;
```
jessicamcinchak, line 73: It seems that this change means the type is _less_ desscriptive/meaningful rather than _more_ - I'm not sure I understand? 

If `trackHelpClick` has type `HelpClickMetadata` then it already _is_ valid `Metadata` right? But with this update, I could incorrectly pass `trackHelpClick({ "back": { ...metadata } })` and I don't think it wouldn't get caught, right?

hasura.planx.uk/migrations/1708016067772_run_sql_migration/up.sql
Mike-Heneghan: I created my branch off of Daf's but when I created the PR I didn't notice I was pointing at `main` I'll get that sorted 👍

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
Mike-Heneghan: If I understand Daf's intention with adding `Metadata` I think it's nice to make clear the expected content that could be written to the `metadata` jsonb column which _could_ have anything written to it but we're setting our expectations which I really like and could see the value of. I was associating all functions which write to the column with `Metadata` so that it's clear it'll be one of the types it's composed of.

Although, as you say it does mean we're not enforcing the types on the functions which as you say is less descriptive. 

I've updated my PR to expand on the `Metadata` but the new type never gets used anywhere preserving the more explicit types on the functions 🤔 

Rather than go with this approach it would maybe be best to remove the `Metadata` type?

jessicamcinchak: In my mind this distinction is: 
- specific types like `HelpClickMetadata` should be used to restrict the arguments passed to methods like `trackHelpClick` 
- the broad union type `Metadata` should be used to define what the GraphQL mutation/query requests for this JSONB column expect

Mike-Heneghan: I definitely agree with that @jessicamcinchak although the bit I'm not sure on is whether we can actually assign the `Metadata` type to the GraphQL request, is that what you're suggesting? 

With `trackHelpClick` I think the function is associated with `HelpClickMetadata` but there's no where to add `Metadata` as I understand it although I think I might be missing something.

```
  async function trackHelpClick(metadata?: HelpClickMetadata) {
    if (shouldSkipTracking()) return;

    await publicClient.mutate({
      mutation: gql`
        mutation UpdateHasClickedHelp($id: bigint!, $metadata: jsonb = {}) {
          update_analytics_logs_by_pk(
            pk_columns: { id: $id }
            _set: { has_clicked_help: true }
            _append: { metadata: $metadata }
          ) {
            id
          }
        }
      `,
      variables: {
        id: lastVisibleNodeAnalyticsLogId,
        metadata,
      },
    });
  }
  ```
  
Or would you suggest we maintain the `Metadata` but just have it in the file not being called but serving as a marker?

I'm a little bamboozled I think but it would be good to fix this PR / close it as I fear I'm wasting your time with this.

File: editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx [hunk 3]
```diff
@@ -28,28 +28,41 @@ const ALLOW_LIST = [
...
+type AutoAnswerMetadata = {
   isAutoAnswered?: boolean;
 };
 
+type NodeMetadata = ResultNodeMetadata & AutoAnswerMetadata;
+
+type BackwardsTargetMetadata = {
+  id?: string;
+  type?: string | null;
+  title?: string;
```
Mike-Heneghan, line 52: These have to be optional as some change events we can't access the target `nodeId` so we only track `change: {}`

DafyddLlyr (approved):

PR #2810, opened 2024-02-23
Mh/more info feedback component testing

Mike-Heneghan:
## What: 

- Add tests for the `MoreInfoFeedback` component
- Update `FeedbackForm` to add a `data-testid`
- Add a default `aria-label` for inputs which don't have an explicit `input label` to meet accessibility requirements

## Why: 

- Increasing the test coverage

### DIFF
--- a/editor.planx.uk/src/components/Feedback/MoreInfoFeedback.test.tsx
+++ b/editor.planx.uk/src/components/Feedback/MoreInfoFeedback.test.tsx
@@ -1,0 +1,155 @@
+import "@testing-library/jest-dom/extend-expect";
+
+import { fireEvent, waitFor } from "@testing-library/react";
+// eslint-disable-next-line no-restricted-imports
+import userEvent from "@testing-library/user-event";
+import {
+  getInternalFeedbackMetadata,
+  insertFeedbackMutation,
+} from "lib/feedback";
+import React from "react";
+import { axe, setup } from "testUtils";
+
+import MoreInfoFeedbackComponent from "./MoreInfoFeedback";
+
+jest.mock("lib/feedback", () => {
+  return {
+    getInternalFeedbackMetadata: jest.fn(),
+    insertFeedbackMutation: jest.fn(),
+  };
+});
+
+const scrollIntoView = jest.fn();
+window.Element.prototype.scrollIntoView = scrollIntoView;
+
+describe("MoreInfoFeedbackComponent presentation and functionality", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/components/Feedback/MoreInfoFeedback.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+  test("Feedback form requires a comment before submitting", async () => {
+    const { getByTestId, getByText } = setup(<MoreInfoFeedbackComponent />);
+
+    fireEvent.click(getByText("Yes"));
+    await waitFor(() => {
+      expect(getByTestId("userCommentTextarea")).toBeInTheDocument();
+    });
+
+    expect(getByTestId("userCommentTextarea")).toHaveAttribute("required");
+  });
```
jessicamcinchak, line 109: This test highlights something interesting - while we can't test that the browser stops form submit in Jest, we'd normally test that an input error message becomes visible when trying to proceed without filling out required fields  - but it doesn't seem like we've accounted for that with any of these components and are relying on default browser behavior? 

I'd expect us to the use the same GOV.UK pattern of red left border and message that we use for all other public-facing inputs - is there a reason we aren't? I think it's very likely this will be picked up in a11y testing as inconsistent and insufficient.

Screenshot from 2024-02-23 15-43-21

Screenshot from 2024-02-23 15-36-07

jessicamcinchak (approved):
Tests look good here, thanks for picking this up! 

One comment about our error handling pattern on feedback input, which I'm happy to be picked up separately.

editor.planx.uk/src/components/Feedback/MoreInfoFeedback.test.tsx
Mike-Heneghan: At the time this feature was being built I think there were conversations about using the native `required` although to be fair that might have been scoped to the editor rather than the public view.

I did have a version with the gov uk style errors but they could never actually be seen by users as the `required` attribute stopped form submission which would be run to show these validation messages so that state could never be reached. 

As this was the case at the time I thought I'd reduce complexity but not showing any code which I thought a user could never see. 

It'll be really interesting to see what the a11y testing comes back with because on one hand not using `required` for a required field seems somewhat off? Although as you say it's inconsistent 🤔

Mike-Heneghan:
[Design & implement in-house feedback feature & retire Feedback Fish](https://trello.com/c/8pmX6wmZ/2545-design-implement-in-house-feedback-feature-retire-feedback-fish)

PR #2811, opened 2024-02-23
feat: simplified change logs on publish

jessicamcinchak:
Branched from #2783

Changes based on feedback from publishing conversations this week:
- "Drafts all the way down" route is too confusing to explain to average user, so only make it visible to `platformAdmins`
- Adds a simple badge to indicate there are unpublished changes (hidden if 0 or errors)
  - Ideally I think this would be text rather than count, but [MUI Badge](https://mui.com/material-ui/react-badge/) is a bit limiting here! 
  - In future, goal is to repeat badge on any portal nodes in graph that have changed
- Trying out a _very_ mvp version of a simplified publish change log message
  - Summarise updated & deleted nodes, with option to expand to full list (full list is exact same as default view now)
  - Link to any nested flows that you have `teamEditor` access to, or show as plain text otherwise
  - In future, also add expandable details for each nested flow like last published date, summary message, etc
  - Styles here are quite rough, but will be easy to revisit later once we have a new shape of "summary content" 
  

Screenshot from 2024-02-23 14-49-45

Screenshot from 2024-02-28 09-00-25

Screenshot from 2024-02-28 08-58-38


editor.planx.uk/src/pages/FlowEditor/components/PreviewBrowser.tsx
```diff
@@ -68,28 +69,71 @@ const DebugConsole = () => {
...
+          </li>
+        </ul>
+      )}
+      {changeSummary["portals"].length > 0 && (
+        <>
+          <Typography variant="body2">{`This includes recently published changes in the following nested services:`}</Typography>
+          <ul>
+            {changeSummary["portals"].map((portal, i) => (
+              <li key={i}>
+                <Typography variant="body2">{portal}</Typography>
```
DafyddLlyr, line 126: Could making these links be a quick UI win?

jessicamcinchak: Yep - but I'm only going to link them if `canUserEditTeam()` is true I think - as it's quite misleading to be linked to a portal you don't actually have permission to publish yourself? 

Same for change badges on portals - I don't think it feels as valuable to show that there's changes waiting to be published if you don't actually have permission to do the publishing? Versus only showing for the flow you've navigated direclty into feels a bit more straightforward at least in this first iteration?

editor.planx.uk/src/pages/FlowEditor/components/PreviewBrowser.tsx
```diff
@@ -225,45 +283,57 @@ const PreviewBrowser: React.FC<{
...
+          <Box
+            display="flex"
+            flexDirection="column"
+            alignItems="flex-end"
+            marginRight={1}
+          >
+            <Badge
+              sx={{ width: "100%" }}
+              badgeContent={alteredNodes && alteredNodes.length}
+              max={999}
```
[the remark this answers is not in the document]
jessicamcinchak: Helpful ideas here, will definitely make sure to pin this PR to Trello so we don't lose. 

But personally think this one is not-MVP! We know that there's also time/a ticket to revisit publish change diffs in phase 4, I think this change should stay as logically focused as possible with just a few visual tidy ups to help with presentation.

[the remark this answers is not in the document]
DafyddLlyr: Yep totally agree!

editor.planx.uk/src/pages/FlowEditor/components/PreviewBrowser.tsx
DafyddLlyr: Yep, let's keep it simple at first and see what's requested next / proves helpful.

editor.planx.uk/src/pages/FlowEditor/components/PreviewBrowser.tsx
```diff
@@ -68,7 +72,7 @@ const DebugConsole = () => {
   );
 };
 
-function PublishChangeItem(props: any) {
+function AlteredNodeItem(props: any) {
```
jessicamcinchak, line 75: I'm aware there's some lingering `any`'s around this file that I haven't cleaned up - one on my radar to come back to, want to keep up momentum with logic changes first !

DafyddLlyr (approved):

PR #2815, opened 2024-02-26
test: FeedbackPhaseBanner component functionality and accessibility

Mike-Heneghan:
## What

- Add tests for the FeedbackPhaseBanner component functionality and accessibility

## Why

- Improving test coverage

### DIFF
--- a/editor.planx.uk/src/components/Feedback/FeedbackPhaseBanner.test.tsx
+++ b/editor.planx.uk/src/components/Feedback/FeedbackPhaseBanner.test.tsx
@@ -1,6 +1,3 @@
-import "@testing-library/jest-dom/extend-expect";
-
-import { fireEvent } from "@testing-library/react";
 import React from "react";
 import { axe, setup } from "testUtils";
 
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/components/Feedback/FeedbackPhaseBanner.test.tsx
```diff
@@ -0,0 +1,67 @@
+import "@testing-library/jest-dom/extend-expect";
+
+import { fireEvent } from "@testing-library/react";
```
DafyddLlyr, line 3: This isn't something we're currently enforcing (maybe we should?) but `fireEvent` is rarely needed and we can rely on `user` for an experience which is closer to the real DOM. 

Instead of importing this we can can use the `user` returned from `setup()` like so - 

```ts
test("clicking on feedback link calls handleFeedbackClick", async () => {
  const { getByText, user } = setup(
    <FeedbackPhaseBanner
      handleFeedbackClick={handleFeedbackClick}
      handleReportAnIssueClick={handleReportAnIssueClick}
    />,
  );

  await user.click(getByText("feedback"));
  expect(handleFeedbackClick).toHaveBeenCalledTimes(1);
});
```

The notable change here is that this is now an `async` function - I think the test will just fail without the `await` before `user.click()`.

Docs: https://testing-library.com/docs/dom-testing-library/api-events/

DafyddLlyr (changes requested):
One small change required!

editor.planx.uk/src/components/Feedback/FeedbackPhaseBanner.test.tsx
Mike-Heneghan: Fixed here: 2035632e5c109b90ebf6e065192a143a7169d2c6

Mike-Heneghan:
[Design & implement in-house feedback feature & retire Feedback Fish](https://trello.com/c/8pmX6wmZ/2545-design-implement-in-house-feedback-feature-retire-feedback-fish)

PR #2831, opened 2024-02-28
feat: Styling for editor data inputs

ianjon3s:
# What does this PR do?

Requested by August and a general usability win:
The current formatting for data inputs differs from default inputs with only a slight background tint, meaning that data is often added to the component description in error.

This PR introduces a monospace font (CSS embed from Google Fonts), and a slightly different input style.

Before:
image

After:
image

Still to do (if possible):
- [x] Update font ref so that it is only loaded in for editor screens


DafyddLlyr (approved):
Looks great @ianjon3s 👏

PR #2835, opened 2024-02-29
fix: Read allow list values from breadcrumbs not flow

DafyddLlyr:
image

## What's the problem?
 - Allow list values not set as the explicit `fn` or `val` of a node are not correctly captured
 - This is an issue as some nodes set multiple values internally which we want to capture

## What's the solution?
 - Scans breadcrumbs for allow list values instead of node
 - This ensures we capture all allow list values - those set by Editors, and those set internally by nodes

## Testing
Currently this needs to be tested manually - see screenshot above.

I'm not exactly sure the best way of testing this as there's no existing framework. We could set this up for analytics in E2E tests but ideally we'd find a lighter weight way of testing this in the frontend. In order to move this forward I've slightly swerved this issue for now.

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
@@ -27,7 +27,7 @@ const ALLOW_LIST = [
   "drawBoundary.action",
   "user.role",
   "property.constraints.planning",
-] as const;
+];
 
 export type HelpClickMetadata = Record<string, string>;
 type SelectedUrlsMetadata = Record<"selectedUrls", string[]>;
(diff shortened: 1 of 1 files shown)


DafyddLlyr:
[Track `drawBoundary.action` passport variable in `analytics_logs` allow list, not only on submission services](https://trello.com/c/Y8Atst5y/2796-track-drawboundaryaction-passport-variable-in-analyticslogs-allow-list-not-only-on-submission-services)

jessicamcinchak (approved):
Working as expected for me now :heavy_check_mark: 

Thanks for outlining outstanding testing questions, happy to keep those in mind (e2e suite sounds like nice ultimate goal) but move forward this fix in meantime :+1:

DafyddLlyr:
Yep agreed - been thinking about it a bit just now and I think that actually the complexities and comprises required to handle it 100% in the frontend would be too much (plus there's a whole set of API side analytics we'd need to handle separately).
 
An E2E suite is probably the most robust and consistent way of describing and testing PlanX analytics 👍

PR #2839, opened 2024-02-29
fix: if same portal is nested many times at different levels, ensure latest version is always flattened and not old snapshot

jessicamcinchak:
Initially point to `main` to build fresh pizza for comparative testing, but will rebase this to point at `jess/fetch-published-portals-on-publish` before merging.

Context:
- The same portal can be referenced many times, but its' nodes only appear in the flattened flow data structure once 
- The same portal can be reference many times _at different levels_ 
  - For example, tthere's a parent flow that references two nested flows A & B. Flow B also references flow A. 
    - B can be published when A is at v1
    - The parent flow can be published when A is at v2 and B is at v1 (which snapshot A v1). 
    - In these cases, A v1 was captured in the flattened parent flow data, rather than forcing B to also consume the _latest_ published version of A (v2) which is a direct child of the parent flow

Technically, this means that A v2 could have a breaking change for portal B, but this isn't really different than content risks that already exist & will be tested/reviewed on the `/draft` route. August & I talked a lot about this behavior, can explain more if questions!

Needs updated complex unit test case if confirmed to work right on pizza now.

**Update**
- August confirmed working as expected on pizza now ! 
- Let's get this back over to original branch - and since subdomains are moving slow, I'm anticipating opening that pizza for PO testing next monday morning rather than staging - so I'll keep tidying/expanding complex test cases over there

### DIFF
--- a/api.planx.uk/helpers.ts
+++ b/api.planx.uk/helpers.ts
@@ -259,6 +259,7 @@ const dataMerged = async (
   }
 
   // for every external portal that has been merged, confirm its' latest version was merged. If not, overwrite older snapshot with newest version
+  //   ** this is a final/separate step because older snapshots can be nested in _already_ flattened data (eg not picked up as ComponentType.ExternalPortal above)
   if (!draftDataOnly) {
     for (const [nodeId, node] of Object.entries(ob).filter(
       ([_nodeId, node]) => node.data?.publishedFlowId,
--- a/api.planx.uk/helpers.ts
+++ b/api.planx.uk/helpers.ts
@@ -9,7 +9,7 @@ export interface FlowData {
   data: Flow["data"];
   team_id: number;
   team: { slug: string };
-  publishedFlows: { data: Flow["data"] }[] | [];
+  publishedFlows: { data: Flow["data"]; id: number }[] | [];
 }
 
 // Get a flow's data (unflattened, without external portal nodes)
@@ -29,6 +29,7 @@ const getFlowData = async (id: string): Promise<FlowData> => {
             order_by: { created_at: desc }
           ) {
             data
+            id
           }
         }
       }
@@ -125,6 +126,7 @@ interface PublishedFlows {
     publishedFlows: {
       // TODO: use FlowGraph from planx-core here
       data: Flow["data"];
+      id: number;
     }[];
   } | null;
 }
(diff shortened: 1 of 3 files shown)


api.planx.uk/helpers.ts
```diff
@@ -213,6 +258,20 @@ const dataMerged = async (
     else ob[nodeId] = node;
   }
 
+  // for every external portal that has been merged, confirm its' latest version was merged. If not, overwrite older snapshot with newest version
+  //   ** this is a final/separate step because older snapshots can be nested in _already_ flattened data (eg not picked up as ComponentType.ExternalPortal above)
```
jessicamcinchak, line 262: @DafyddLlyr does this comment help clarify what we discussed this morning at all?

editor.planx.uk/src/pages/FlowEditor/components/PreviewBrowser.tsx
```diff
@@ -196,19 +204,39 @@ function AlteredNodesSummaryContent(props: any) {
...
+                          {portal.summary}
+                        </Typography>
+                      )}
+                    </>
+                  }
+                />
+              </ListItem>
             ))}
-          </ul>
+          </List>
```
jessicamcinchak, line 239: This looks like this :
Screenshot from 2024-03-01 16-49-00

DafyddLlyr: This is such a nice outcome 🎉

api.planx.uk/helpers.ts
DafyddLlyr: Yep this is great - suuuper explicit and we'll be able to understand with zero context in 18 months' time 👍 

Thanks for the extra effort in documenting the changes in this PR.

editor.planx.uk/src/pages/FlowEditor/components/PreviewBrowser.tsx
```diff
@@ -196,19 +204,39 @@ function AlteredNodesSummaryContent(props: any) {
...
+                    ) : (
+                      <Typography variant="body2">{portal.text}</Typography>
+                    )
+                  }
+                  secondary={
+                    <>
+                      <Typography variant="body2" fontSize="small">
+                        {`Last published ${formatDistanceToNow(
+                          new Date(portal.publishedAt),
+                        )} ago by ${portal.publishedBy}`}
```
DafyddLlyr, line 227: This is currently a userId, and not firstname + lastname. As discussed, this is because the info is generated from a public query.

I think having just a number is probably more confusing for Editors? I'd say we may be best not including this just now as MVP, and if requested we can think of how we can get this info in a secure way?

DafyddLlyr (approved):

editor.planx.uk/src/pages/FlowEditor/components/PreviewBrowser.tsx
jessicamcinchak: Correct - `dataMerged` makes a public query to the `flows` & `published_flows` table which is where the flattened metadata is sourced from.

Once we're reading it in the Editor `PreviewBrowser` here, we will be able to make an authenticated call to lookup user details though ! It's on my final-tidy to-do list before opening for testing once I'm back to one branch :white_check_mark:

PR #2853, opened 2024-03-01
chore: update default copy for FindProperty plot a new address 

jessicamcinchak:
Per Doncaster feedback, see Trello ticket

New components only, not migrating existing content in this case

### DIFF
--- a/editor.planx.uk/src/@planx/components/FindProperty/model.ts
+++ b/editor.planx.uk/src/@planx/components/FindProperty/model.ts
@@ -12,12 +12,14 @@ export interface FindProperty extends MoreInformation {
 export const parseFindProperty = (
   data: Record<string, any> | undefined,
 ): FindProperty => ({
-  title: data?.title || "",
+  title: data?.title || DEFAULT_TITLE,
   description: data?.description || "",
   allowNewAddresses: data?.allowNewAddresses || false,
-  newAddressTitle: data?.newAddressTitle || "",
-  newAddressDescription: data?.newAddressDescription || "",
-  newAddressDescriptionLabel: data?.newAddressDescriptionLabel || "",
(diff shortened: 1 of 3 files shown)


jessicamcinchak:
[Update default copy of Find Property 'plot new addresses without a UPRN'](https://trello.com/c/bdDppPEN/2800-update-default-copy-of-find-property-plot-new-addresses-without-a-uprn)

editor.planx.uk/src/@planx/components/FindProperty/model.ts
```diff
@@ -54,5 +56,8 @@ export interface SiteAddress extends MinimumSiteAddress {
 }
 
 export const DEFAULT_TITLE = "Find the property" as const;
-export const DEFAULT_NEW_ADDRESS_TITLE = "Propose a new address" as const;
+export const DEFAULT_NEW_ADDRESS_TITLE =
+  "Click or tap at where the property is on the map and name it below" as const;
+export const DEFAULT_NEW_ADDRESS_DESCRIPTION =
+  "You will need to select a location and provide a name to continue" as const;
```
DafyddLlyr, line 62: These form values also appear to be referenced in `planx-new/editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx` which is actually redundant I think.

We should be able to mark these properties as no longer optional in the `FindProperty` interface and tidy this up a bit.

DafyddLlyr (approved):
Perfect, nice and simple!

editor.planx.uk/src/@planx/components/FindProperty/model.ts
jessicamcinchak: Yep good shout - cleaned up from index & Editor placeholders now !

PR #2855, opened 2024-03-03
chore: map user ids to names on publish change dialog for nested flows

jessicamcinchak:
Plus a few type / component tidy ups ! 

**Before**
Screenshot from 2024-03-01 16-49-00

**After**
Screenshot from 2024-03-03 19-03-38


editor.planx.uk/src/pages/FlowEditor/components/PreviewBrowser.tsx
```diff
@@ -286,13 +317,10 @@ const PreviewBrowser: React.FC<{
     "This flow is not published yet",
   );
   const [validationMessage, setValidationMessage] = useState<string>();
-  const [alteredNodes, setAlteredNodes] = useState<object[]>();
+  const [alteredNodes, setAlteredNodes] = useState<AlteredNode[]>();
```
DafyddLlyr, line 320: 👍

DafyddLlyr (approved):
Very nice 👌

PR #2856, opened 2024-03-04
feat: Setup `metabase_read_only` role

DafyddLlyr:
## What does this PR do?
- Sets up the `metabase_read_only` database role
- Ensures that Metabase can only read a small subset of tables and views, and has no write permissions
- Please see included doc file for more context and detail

## Testing
I've gone ahead and set up the user and permissions outlined here on staging Metabase so that we can test a little (I'll revoke the changes before merging this PR).

It works as expected - for example, I can read `sessions_summary`, but running SQL to select from `lowcal_sessions` gives a permissions error.

image

image

### DIFF
--- a/doc/how-to/how-to-grant-metabase-permissions.md
+++ b/doc/how-to/how-to-grant-metabase-permissions.md
@@ -24,6 +24,8 @@ The password for Staging and Production databases can be found the OSL 1Password
 
 The username and password for Metabase are not controlled via IaC - they are manually entered via the Metabase "Admin" dashboard (`Admin Setting > Databases > "staging" | "production" > Username / Password fields`).
 
+Please note - this is separate to the role used to read/write Metabase internal application data (such as dashboard and queries). This role is setup in IaC [here](https://github.com/theopensystemslab/planx-new/blob/main/infrastructure/application/index.ts#L100). For more information, please see [the Metabase docs](https://www.metabase.com/docs/latest/installation-and-operation/configuring-application-database).
+
 ### Locally & Pizzas
 If you wish to run Metabase locally using the "analytics" Docker profile (`pnpm analytics` from project root), you will need to manually run the above SQL on your local database with a password of your choice. Alternatively, you can use the root DB username/password.
 
--- a/doc/how-to/how-to-grant-metabase-permissions.md
+++ b/doc/how-to/how-to-grant-metabase-permissions.md
@@ -1,0 +1,52 @@
+# How to grant Metabase permissions
+
+## What is Metabase?
+[Metabase](https://www.metabase.com/) is an open source BI service which we self-host as part of PlanX. It allows teams to view and self-serve analytics dashboards related to their flows, applications, and users.
+
+Metabase is set up and running on both Staging and Production environments, but only the Production instance (with Production data) has dashboards maintained and curated for teams.
+
+## Context
+Metabase accesses our staging and production databases through the `metabase_read_only` role which has `SELECT` (read-only) access to a subset of tables and views. This ensures that sensitive user data cannot be inadvertently exposed via Metabase, and any new tables added to the PlanX database have to be explicitly exposed via this role.
+
+The permissions granted for the `metabase_read_only` role are applied via Hasura migrations. This ensures that we have a version-controlled history of this role, and it's access level is documented in code. This is not controlled via Pulumi (IaC) as this is not used for local development or test environments ("Pizzas"), which would necessitate a second method for these environments.
+
+## How is this role used?
+
+### Staging & Production
+The `metabase_read_only` role is granted to the `metabase_user` database user, which is manually set up on both staging and production with the following SQL -
+
+```sql
+CREATE USER metabase_user WITH PASSWORD `$PASSWORD`;
+GRANT metabase_read_only TO metabase_user;
+```
+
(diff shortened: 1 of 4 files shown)


doc/how-to/how-to-grant-metabase-permissions.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Staging & Production
+The `metabase_read_only` role is granted to the `metabase_user` database user, which is manually set up on both staging and production with the following SQL -
+
+```sql
+CREATE USER metabase_user WITH PASSWORD <$PASSWORD>;
+GRANT metabase_read_only TO metabase_user;
+```
+
+The password for Staging and Production databases can be found the OSL 1Password account.
```
DafyddLlyr, line 23: This is the solution I'm proposing for setting up the `metabase_user` user. Not perfect but hopefully workable?

Chances are it will exist on staging and production and we'll never really need to deal with it locally.

DafyddLlyr:
[Hide new tables from Metabase as a default](https://trello.com/c/oRqwiUe3/2786-hide-new-tables-from-metabase-as-a-default)

hasura.planx.uk/migrations/1709565833346_run_sql_migration/up.sql
```diff
@@ -0,0 +1,15 @@
+-- Create the role
+CREATE ROLE metabase_read_only;
+
+-- Grant usage on schema
+GRANT USAGE ON SCHEMA public TO metabase_read_only;
+
+-- Grant select permissions on public tables used by Metabase (useful for joins)
+GRANT SELECT ON public.flows TO metabase_read_only;
+GRANT SELECT ON public.published_flows TO metabase_read_only;
+GRANT SELECT ON public.teams TO metabase_read_only;
```
DafyddLlyr, line 10: This might actually be redundant - I think there might be enough information in the views? If it's simpler to remove this I'm all for it 👍

doc/how-to/how-to-grant-metabase-permissions.md
jessicamcinchak: This seems like a good thoughtful approach and will make for less-hassle rotation if necesary in future, but not entirely following based on this doc how this user relates to existing `metabase` user role provisioned/managed by Pulumi - eg https://github.com/theopensystemslab/planx-new/blob/main/infrastructure/application/index.ts#L100 

At minimum, do we need to think about more distinctive naming? At most, is there a way to consolidate responsibilities of each into a single user role?

DafyddLlyr: Yeah this caught me out initially also.

This section of code is generating a role and user to be used for the `MB_DB_CONNECTION_URI` env var, which is responsible for the schema and role of the internal metabase db, where dashboards etc are stored - not the PlanX schema where data is accessed via Hasura / ShareDB.

image

I'll add a code comment to Pulumi to make this explicit, as well as document this in the "how to" in this PR.

Do you think this would be enough?

jessicamcinchak: Yep happy with extra docs here !

hasura.planx.uk/migrations/1709565833346_run_sql_migration/up.sql
DafyddLlyr: I'll contact Ollie today, or manually check Metabase, to see if this is required 👍

jessicamcinchak (approved):
All looks good to me - thanks for extra docs!

hasura.planx.uk/migrations/1709565833346_run_sql_migration/up.sql
DafyddLlyr: Removed - these aren't currently used. See https://opensystemslab.slack.com/archives/C5Q59R3HB/p1709720214019159 (OSL Slack)

DafyddLlyr: Update - in order to move this forward I've granted access to the tables `analytics` and `analytics_logs`. They are safe to read, and currently used by SQL questions. The aim would be that once we migrate to GUI questions, we can rely on summary tables and tidy this up.

@zz-hh-aa Indicated that `lowcal_sessions` is used, which we do not want, so I've intentionally left this out to force a transition to `submission_services_summary`.

If this breaking change it too harsh, we can toggle the PG username/password used in Metabase until this is resolved.

jessicamcinchak (approved):
Newest changes look good - all for not exposing `lowcal_sessions` and just toggling at Metabase level until all migrated to new view :+1:

DafyddLlyr:
Perfect - thanks very much @jessicamcinchak for the speedy re-review here ✅

hasura.planx.uk/migrations/1709565833346_run_sql_migration/up.sql
zz-hh-aa: I think I've removed `lowcal_sessions` as a source from all the questions now anyway, and am only using summary tables + `analytics` or `analytics_logs`, so don't think anything should break! All good on my end.

DafyddLlyr: Ah that's perfect @zz-hh-aa thanks very much!

PR #2859, opened 2024-03-06
fix(a11y): Match ids and for values on DateInput

DafyddLlyr:


DafyddLlyr:
[Placeholder accessible name [A]](https://trello.com/c/djgd2XvE/2808-placeholder-accessible-name-a)

ianjon3s (approved):

PR #2860, opened 2024-03-06
fix(a11y): Replace generic `alt` text with `aria-describedby`

DafyddLlyr:
## What does this PR do?
 - Addresses issue `DAC_Non-Text_Content_02` in the Accessibility Report (page 15)
 - Removes the `alt` text in favour of `aria-describedby`
   - Example of this on MDN docs here - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role#description
   - The conditional logic around `howMeasured` ensures that the image is only displayed if there is an associated description, so it's already not possible to have an image without text.

image


DafyddLlyr:
[Non-text content 02 [A]](https://trello.com/c/fwZOTVtP/2803-non-text-content-02-a)

ianjon3s (approved):

PR #2861, opened 2024-03-06
fix(a11y): Remove redundant hidden text

DafyddLlyr:
## What does this PR do?
 - Addresses issue `DAC_Hidden_Text_01` in Accessability Report
 - Removes redundant `FormHelperText` component
   - Git blame shows this was introduced here https://github.com/theopensystemslab/planx-new/pull/1602

## Why is removing this the solution?
 - The description is already present and reachable by screen-readers
 - The markdown is actually wrong - rather than reference the id `DESCRIPTION_TEXT`, it's referencing the meaningless string `description-text` directly
 - 
image


DafyddLlyr:
[Hidden text [A]](https://trello.com/c/0WRo2YSH/2806-hidden-text-a)

ianjon3s (approved):

PR #2865, opened 2024-03-06
fix(a11y): Remove image alt from title/description of radio

ianjon3s:


ianjon3s:
[Non-text content 01 [A]](https://trello.com/c/svot2Qj2/2802-non-text-content-01-a)

DafyddLlyr (approved):

PR #2867, opened 2024-03-06
feat: show submitted applications for a flow in the flow settings 

Mike-Heneghan:


hasura.planx.uk/migrations/1709563913656_create_public_view_applications_summary/up.sql
```diff
@@ -0,0 +1,20 @@
...
+  sent_to_bops,
+  sent_to_uniform,
+  sent_to_email
+FROM submission_services_summary s
+LEFT JOIN (
+  SELECT DISTINCT ON (session_id) *
+  FROM payment_status
+  ORDER BY session_id, created_at DESC
+) ps ON ps.session_id = s.session_id
+WHERE submitted_at IS NOT NULL;
```
jessicamcinchak, line 20: A couple questions about this view: 
- I'd expect to see GOV UK ID, BOPS ID and Uniform IDs exposed for parity with #planx-notifications rather than a yes/no "sent_to". I think councils currently have a high need for a simple lookup from Planx ID <> integrations ID, rather than needing specific application details like the payment amount exposed here if that makes sense
- It feels very hard to semantically differentiate between `applications_summary` and `submission_services_summary` - is there a reason to not simply update `submission_services_summary` directly to include these additional columns?

editor.planx.uk/src/lib/applications.ts
```diff
@@ -0,0 +1,49 @@
...
+  teamSlug: string,
+) {
+  const { data } = await client.query({
+    query: gql`
+      query SubmittedApplications($service_slug: String!, $team_slug: String!) {
+        applications_summary(
+          where: {
+            service_slug: { _eq: $service_slug }
+            team_slug: { _eq: $team_slug }
+            submitted_at: { _is_null: false }
```
jessicamcinchak, line 27: nit: I think this third `where` clause is unnecessary as you're already doing this in the view definition https://github.com/theopensystemslab/planx-new/pull/2867/files#diff-887c1fcbaafca5d3d84a6324752b4d8edb1fba21b9f9525ede41d6782f394f36R27

editor.planx.uk/src/lib/applications.ts
```diff
@@ -0,0 +1,49 @@
...
+  user_invited_to_pay: boolean;
+  payment_status: string;
+  amount: number;
+  payment_date: Date | string;
+  sent_to_email: boolean;
+  sent_to_bops: boolean;
+  sent_to_uniform: boolean;
+};
+
+export async function fetchSubmittedApplications(
```
jessicamcinchak, line 16: nit: I personally think this function & query feel a bit out of place in `/lib` and should be co-located in the index.tsx component file?

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/ApplicationsTable.tsx
```diff
@@ -0,0 +1,99 @@
...
+  return formattedDate;
+}
+
+function formatBoolean(state: boolean) {
+  return state ? "Yes" : "No";
+}
+
+function formatAmount(amount: number) {
+  return "£" + (amount / 100).toFixed(2);
+}
```
jessicamcinchak, line 40: nit: there's some existing helper methods in the `Pay` component like `formattedPriceWithCurrencySymbol()` that might be able to be imported here instead for consistency?

editor.planx.uk/src/routes/flowSettings.tsx
```diff
@@ -58,32 +84,20 @@ const flowSettingsRoutes = compose(
...
         useStore.getState().setFlowSettings(settings);
 
+        function getTabs() {
+          const isUsingFeatureFlag = hasFeatureFlag(
+            "APPLICATION_SUBMISSION_VIEW",
+          );
+          return isUsingFeatureFlag
+            ? [...standardTabs, submissionsTab]
+            : standardTabs;
+        }
```
jessicamcinchak, line 94: Should the visibility of the `submissionsTab` also be scoped by `canUserEditTeam()`?

While the Slack channel is essentially public (but very minimal members), I think based on other conversations we've been having about future of editor that folks would expect this to only be visible to their own team members?

jessicamcinchak (commented):
Recognise this is still draft, but also know you're about to be away for a long weekend - so a few initial comments on first read through ! Nice to see test coverage and storybook additions here :+1:

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/ApplicationsTable.tsx
Mike-Heneghan: I was sure there must be but I hadn't spotted it, I'll get this updated to use the existing helper 👍

hasura.planx.uk/migrations/1709563913656_create_public_view_applications_summary/up.sql
Mike-Heneghan: Data Exposed: 
- That's a very good point regarding GOV PAY ID, BOPS ID and Uniform IDs.
- I hadn't properly parsed #planx-notifications so was just seeing what I thought were PlanX session ids 🤦 
- I'll get this updated 👍 
- Do we still want payment details other than the id, I think it was mentioned in the ticket but I got the wrong end of the stick?

editor.planx.uk/src/routes/flowSettings.tsx
jessicamcinchak: You're right, thanks for that reminder - since these tabs are within flow settings menu they are already hidden from anyone else (aka `teamViewer`) by default - that was my oversight! 

No need to do anything further here then in my opinion.

hasura.planx.uk/migrations/1709563913656_create_public_view_applications_summary/up.sql
Mike-Heneghan: View:

- I had originally used `submission_services_summary` but as that's exposed via Metabase and in my understanding would have the potential to be more public facing I thought I better hide the extra info like payment details
- I hadn't added it but I thought there might be a chance we'd add more explicit PII like email or name to this view depending on how CO would use it?
- If that's not the case and the data we'd have here wouldn't be sensitive then I think it could be added to the existing view 👍

editor.planx.uk/src/lib/applications.ts
Mike-Heneghan: That's fair enough, I fancies abstracting it out a bit like `model.ts` type files or `/lib/feedback` although this is very weakly held and I'm happy to update it 👍

editor.planx.uk/src/lib/applications.ts
Mike-Heneghan: Great spot! I had this before I decided to add the filter in the view. I'll get this updated 👍

hasura.planx.uk/migrations/1709563913656_create_public_view_applications_summary/up.sql
jessicamcinchak: I'd vote we aim for feature parity with the information in #planx-notifications which is not currently PII and would therefore let us use the same view. 

As I understand it this is meant to sort of be more a receipt of pay & send events, and it's the back office systems responsibility to display actual application _details_. If further requests come from POs, we could of course re-evaluate.

What I think the data we want is:
- session id 
- `payment_requests.status`, `govpay_payment_id`, `created_at` (amount not necessary)
- `bops_applications.bops_id` (maybe `destination_url` to differentiate between v1 & v2 recently)
- `uniform_applications.idox_submission_id` (`submission_reference` = session id so can be omitted)
- `email_applications.recipient` 
- submission timestamp for each applicable destination

That would all be considered non-PII as it's just how various identifiers link up. 

There's one consideration with my proposal though that I'm not sure your original data structure accounted for: there will be many `payment_requests` & `bops_applications` per each session and I think we want to show each of them, not just the latest. This might make a case for a separate view or just a note to `group by session_id` if able to use exisitng `submission_services_summary`?

Does that help at all?

Mike-Heneghan: This is super helpful thanks Jess!

I'll have a look into adding this to the view and I could maybe create arrays of `payment_requests` and `bops_applications` to maintain the expectation of one row in `submission_services_summary` per `session_id` 🤔

jessicamcinchak: Arrays make sense if easy enough to work with on the frontend - I'd imagine GraphQL should be good for this :+1: 

And just to be totally clear - any pay or send destination can technically be many events to one session id because of re-submissions, but payments & BOPS are just the two that happen most often/intentionally !

Mike-Heneghan: I've added a PR to update the `submission_services_summary` view to hopefully expose the data as discussed: https://github.com/theopensystemslab/planx-new/pull/2879

Mike-Heneghan:
Closing as this has been reworked as per feedback and split into smaller PRs.

PR #2868, opened 2024-03-06
fix(a11y): change links reference descriptive list titles whether dynamic or static

jessicamcinchak:
**Issue**
We have two component types - FindProperty & DrawBoundary - displayed on the review page where we can't rely on the node title or text to be a sensible row title in the descriptive list.

Previously, the row would display a static title but the visibly hidden change link still pointed at the node title - making it impossible to connect the change link to the correct row content.
 
Now, the change link consistently reflects the row titles. 

All nodes in "real" services should have titles (would be nice to enforce Editor validation consistently across component types here in future), so I'm also removing other static fallback titles for clarity as these would similarly fail a11y testing.

**Testing**
https://2868.planx.pizza/testing/review/unpublished
Screenshot from 2024-03-07 09-09-09

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
@@ -351,7 +351,7 @@ function FindProperty(props: ComponentProps) {
 function Checklist(props: ComponentProps) {
   return (
     <>
-      <dt>{props.node.data.text ?? "Checklist"}</dt>
+      <dt>{props.node.data.text}</dt>
       <dd>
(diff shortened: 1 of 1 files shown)


jessicamcinchak:
[Change links [A]](https://trello.com/c/FTyWyRTZ/2814-change-links-a)

DafyddLlyr (approved):
Looks good! Agreed that enforcing title is a useful step we should take.

We could even just enforce it via the Editor modal (add a `required` prop) for now which would stop any new content being written without a title, and revisit validation of existing flows when we take a closer look at that next phase.

PR #2873, opened 2024-03-08
feat: Add title to calculate component

DafyddLlyr:
## What does this PR do?
- Adds an optional "title" field to the `Calculate` component

image

image

### DIFF
--- a/editor.planx.uk/src/@planx/components/Calculate/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/Calculate/Editor.tsx
@@ -72,6 +72,13 @@ export default function Component(props: Props) {
     <form onSubmit={formik.handleSubmit} id="modal">
       <ModalSection>
         <ModalSectionContent title="Calculate" Icon={ICONS[TYPES.Calculate]}>
+          <Typography variant="body2" sx={{ mb: 2 }}>
+            This component does math! Write formulas using{" "}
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/Calculate/Editor.tsx
```diff
@@ -72,7 +72,16 @@ export default function Component(props: Props) {
...
+          <InputRow>
+            <Input
+              format="large"
+              placeholder="Title"
+              name="title"
+              value={formik.values.title}
+              onChange={formik.handleChange}
+            />
+          </InputRow>
+          <Typography variant="body2" sx={{mt: 2}}>
```
jessicamcinchak, line 84: nit: I think the order of Typography block _then_ title input would look a bit less choppy here!

jessicamcinchak (approved):
:zap: :100:

editor.planx.uk/src/@planx/components/Calculate/Editor.tsx
DafyddLlyr: Good shout, swapped in [e406e3c](https://github.com/theopensystemslab/planx-new/pull/2873/commits/e406e3cd3ff92e99563132dcadc276b4cb893bd5)

Screenshot in PR updated also ✅

PR #2878, opened 2024-03-12
feat(editor): Dynamic GOV.UK Pay Metadata in `Pay` component

DafyddLlyr:
## What does this PR do?
 - Adds new GOV.UK Pay Metadata section in Pay component
 - Adds form validation rules to match GOV.UK Pay requirements

## Context
Apologies - PR is much bigger than I'd like! A good chunk of it should just be tests though.

I'll pick up the API half of this in another PR 👍 

## TODO
- [ ] API - Pick up metadata from flow, format, and pass along to GovPay
- [x] Write testing notes on Trello

image

### DIFF
--- a/editor.planx.uk/src/@planx/components/Pay/Editor.test.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Editor.test.tsx
@@ -1,0 +1,216 @@
+import { User } from "@opensystemslab/planx-core/types";
+import { fireEvent, waitFor } from "@testing-library/react";
+import { FullStore, vanillaStore } from "pages/FlowEditor/lib/store";
+import React from "react";
+import { DndProvider } from "react-dnd";
+import { HTML5Backend } from "react-dnd-html5-backend";
+import { act } from "react-dom/test-utils";
+import { axe, setup } from "testUtils";
+
+import PayComponent from "./Editor";
+
+describe("Pay component - Editor Modal", () => {
+  it("renders", () => {
+    const { getByText } = setup(
+      <DndProvider backend={HTML5Backend}>
+        <PayComponent id="test" />
+      </DndProvider>,
+    );
+    expect(getByText("Payment")).toBeInTheDocument();
+  });
+
+  // Currently failing, Editor not a11y compliant
+  it.skip("should not have any accessibility violations upon initial load", async () => {
+    const { container } = setup(
+      <DndProvider backend={HTML5Backend}>
+        <PayComponent id="test" />
+      </DndProvider>,
+    );
+    const results = await axe(container);
+    expect(results).toHaveNoViolations();
+  });
+
+  describe("GOV.UK Pay Metadata section", () => {
+    // Set up mock state with platformAdmin user so all Editor features are enabled
+    const { getState, setState } = vanillaStore;
+    const mockUser: User = {
+      id: 123,
+      email: "b.baggins@shire.com",
+      isPlatformAdmin: true,
+      firstName: "Bilbo",
+      lastName: "Baggins",
+      teams: [],
+    };
+
+    let initialState: FullStore;
+
+    beforeAll(() => (initialState = getState()));
+    afterEach(() => act(() => setState(initialState)));
+
+    it("renders the section", () => {
+      const { getByText } = setup(
+        <DndProvider backend={HTML5Backend}>
+          <PayComponent id="test" />
+        </DndProvider>,
+      );
+      expect(getByText("GOV.UK Pay Metadata")).toBeInTheDocument();
+    });
+
+    it("lists the default values", () => {
+      const { getByDisplayValue } = setup(
+        <DndProvider backend={HTML5Backend}>
+          <PayComponent id="test" />
+        </DndProvider>,
+      );
+      expect(getByDisplayValue("flow")).toBeInTheDocument();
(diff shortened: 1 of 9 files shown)


editor.planx.uk/src/@planx/components/Pay/model.ts
```diff
@@ -1,9 +1,14 @@
 import { useStore } from "pages/FlowEditor/lib/store";
 import { ApplicationPath } from "types";
-import { boolean, object, string } from "yup";
+import { array, boolean, object, string } from "yup";
 
 import type { MoreInformation } from "../shared";
 
+export interface GovPayMetadata {
```
DafyddLlyr, line 7: Spent a bit of time trying to get the right "shape" for this. 

Settled on this as it makes form validation and population easy. Validating dynamic key values proved more tricky than it was worth. It just means there will be a step of formatting this for GovPay on the API side of things.

editor.planx.uk/src/@planx/components/Pay/Editor.tsx
```diff
@@ -1,225 +1,392 @@
...
+            <ModalSectionContent
+              title="GOV.UK Pay Metadata"
+              Icon={DataObjectIcon}
+            >
+              <Typography variant="subtitle2" sx={{ mb: 2 }}>
+                Include metadata alongside payments, such as VAT codes, cost
+                centers, or ledger codes. See{" "}
+                <Link href="#TODO">our guide</Link> for more details.
+              </Typography>
+              <ErrorWrapper
```
DafyddLlyr, line 237: There may be an argument to move this `ErrorWrapper` and the one on line 68 into the `ListManager` component.

I decided not to do this as - 
 - It's only being used here
 - Scope creep?
 - The PR has more than enough going on as-is

Just flagging as this is something we can revisit if we start repeating this pattern or refactoring Editor components.

editor.planx.uk/src/@planx/components/Pay/Editor.tsx
```diff
@@ -1,225 +1,392 @@
...
+          </ModalSection>
+          <ModalSection>
+            <ModalSectionContent
+              title="GOV.UK Pay Metadata"
+              Icon={DataObjectIcon}
+            >
+              <Typography variant="subtitle2" sx={{ mb: 2 }}>
+                Include metadata alongside payments, such as VAT codes, cost
+                centers, or ledger codes. See{" "}
+                <Link href="#TODO">our guide</Link> for more details.
```
DafyddLlyr, line 235: Intentional `TODO` - will write something up in Notion, likely after initial PO testing round.

jessicamcinchak: I wonder if we can just link directly to Gov Pay docs here and save ourselves a custom doc? Eg https://docs.payments.service.gov.uk/reporting/#add-more-information-to-a-payment-39-custom-metadata-39-or-39-reporting-columns-39

A bit more technical, but probably ideal for councils to be used to consulting Gov Pay directly anyways?

editor.planx.uk/src/@planx/components/Pay/Editor.tsx
```diff
@@ -281,7 +314,10 @@ const Component: React.FC<Props> = (props: Props) => {
                       setFieldValue("govPayMetadata", metadata);
                     }}
                     Editor={GovPayMetadataEditor}
-                    newValue={() => ({ key: "", value: "" })}
+                    newValue={() => {
+                      setTouched({});
+                      return { key: "", value: "" };
+                    }}
```
DafyddLlyr, line 320: This is a bit of an awkward workaround of this issue - https://github.com/jaredpalmer/formik/issues/1616

As pointed out by Silvia when testing (see comments here), if 
 - a field has an error
 - you delete it
 - and add a new field
 
The error persists, by array index presumably (?), on the new field.

This code change forces a reset of the touched status when adding a new field, and error display is now correctly tied to status also (an error will only display when a field is touched).

editor.planx.uk/src/@planx/components/Pay/Editor.tsx
DafyddLlyr: This is a really good shout tbh! Updated here [45a289b](https://github.com/theopensystemslab/planx-new/pull/2878/commits/45a289bb834a278f63e5038f162babb82e6987f5)

image

editor.planx.uk/src/@planx/components/Pay/Editor.test.tsx
```diff
@@ -0,0 +1,320 @@
...
+      <DndProvider backend={HTML5Backend}>
+        <PayComponent id="test" />
+      </DndProvider>,
+    );
+    const results = await axe(container);
+    expect(results).toHaveNoViolations();
+  });
+
+  describe("GOV.UK Pay Metadata section", () => {
+    jest.setTimeout(20000);
```
DafyddLlyr, line 34: Spent way too long trying to come up with something better than this!

As these are some of the only Editor tests I'm hitting up against a few things here - I think it's mainly the "update not wrapped in an act()" warning which we've butted up against on other occasions (https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning)

I'd like to come back and take another look at this, but in order to progress this PR I'd sidestepped the issue for now by just allowing the tests to run for longer.

PR #2881, opened 2024-03-12
refactor: Unify summary list tables

ianjon3s:
# What does this PR do?

The accessibility report identified an instance of an unstructured table (everything in <td>) on the Confirmation page of a service (p.46). The report suggested that we use <th> to add semantic structure to the table, however we can also achieve this using a Description List (`<dl>`), as we have done elsewhere.

Looking further into this I found three different uses of the same layout/pattern:

1. A Description List (`<dl>`) used for the Review page, as a styled component (SummaryList.tsx)
2. Another Description List with replicated styling from the above used for the Property Information page
3. An unstructured table with styling used for the Confirmation page (identified in the report)

Exporting the styling from the SummaryList component allows us to use this across each page as a Description List, which also addresses the accessibility issue.

### Preview

Navigate through the following service to see the three uses of the SummaryList component:

https://2881.planx.pizza/testing/summary-list/preview

### DIFF
--- a/api.planx.uk/modules/admin/session/html.test.ts
+++ b/api.planx.uk/modules/admin/session/html.test.ts
@@ -53,7 +53,7 @@ describe("HTML data admin endpoint", () => {
       .expect(200)
       .expect("content-type", "text/html; charset=utf-8")
       .then((res) =>
-        expect(res.text).toContain("<dt>Is this a test?</dt><dd>Yes</dd>"),
+        expect(res.text).toContain("<Box component="dt">Is this a test?</Box><Box component="dd">Yes</Box>"),
       );
   });
 });
--- a/api.planx.uk/modules/admin/session/html.test.ts
+++ b/api.planx.uk/modules/admin/session/html.test.ts
@@ -53,7 +53,7 @@ describe("HTML data admin endpoint", () => {
       .expect(200)
       .expect("content-type", "text/html; charset=utf-8")
       .then((res) =>
-        expect(res.text).toContain("<Box component="dt">Is this a test?</Box><Box component="dd">Yes</Box>"),
+        expect(res.text).toContain("<dt>Is this a test?</dt><dd>Yes</dd>"),
       );
   });
 });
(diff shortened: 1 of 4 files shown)


ianjon3s:
[Table structure [A]](https://trello.com/c/vS57J0WU/2816-table-structure-a)

editor.planx.uk/src/@planx/components/PropertyInformation/Public.tsx
```diff
@@ -248,12 +217,12 @@ function PropertyDetails(props: PropertyDetailsProps) {
...
   return (
-    <PropertyDetailsList component="dl">
+    <SummaryListTable showChangeButton={true}>
       {filteredData.map(({ heading, detail, fn }: PropertyDetail) => (
         <React.Fragment key={heading}>
-          <Box component="dt">{heading}</Box>
-          <Box component="dd">{detail}</Box>
-          <Box component="dd">
+          <dt>{heading}</dt>
+          <dd>{detail}</dd>
```
DafyddLlyr, line 224: nit: `<Box component="dt">` just compiles down to a `dt` so should be the same from an accessibility point of view.

Using a `Box` and setting a `component` prop is more consistent with the rest of the code base and would more easily allow us to customise this if needed (e.g. using the `sx` prop).

DafyddLlyr (approved):
One small comment, but all looking good!

editor.planx.uk/src/@planx/components/PropertyInformation/Public.tsx
ianjon3s: Thanks for the context on this, I've reverted and updated other instances of description lists to be consistent.

api.planx.uk/modules/admin/session/html.test.ts
```diff
@@ -53,7 +53,7 @@ describe("HTML data admin endpoint", () => {
       .expect(200)
       .expect("content-type", "text/html; charset=utf-8")
       .then((res) =>
-        expect(res.text).toContain("<dt>Is this a test?</dt><dd>Yes</dd>"),
+        expect(res.text).toContain("<Box component="dt">Is this a test?</Box><Box component="dd">Yes</Box>"),
```
DafyddLlyr, line 56: I think this test won't need to be updated, as the HTML on the build application will get compiled as a plain `dt`

ianjon3s: Ah good spot. Find and replace used too lavishly, I'll update.

DafyddLlyr (approved):

PR #2884, opened 2024-03-13
feat: add tab for submissions view to flowSettings

Mike-Heneghan:
## What

- Add a new placeholder component for a "Feature in development"
- Conditionally render this component as a flow settings tab based on the `SUBMISSION_VIEW` feature flag. 

## Why

- Adding development behind a feature flat allows the work to be shipped incrementally without being visible to user
- Setups the structure for adding a table for visualising application data

## Screenshots

Tab rendered with feature flag enabled:

Screenshot 2024-03-13 at 11 46 18

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/index.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/index.tsx
@@ -1,0 +1,24 @@
+import Box from "@mui/material/Box";
+import Typography from "@mui/material/Typography";
+import React from "react";
+import { FeaturePlaceholder } from "ui/editor/FeaturePlaceholder";
+
+const Submissions: React.FC = () => {
+  return (
+    <>
+      <Box>
+        <Typography variant="h2" component="h3" gutterBottom>
+          Submissions
+        </Typography>
+        <Typography variant="body1">
+          View data on the user submitted applications for this service.
+        </Typography>
+        <Box py={5}>
+          <FeaturePlaceholder title="Feature in development" />
+        </Box>
+      </Box>
+    </>
+  );
+};
+
+export default Submissions;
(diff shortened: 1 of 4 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/index.tsx
```diff
@@ -0,0 +1,24 @@
+import Box from "@mui/material/Box";
+import Typography from "@mui/material/Typography";
+import React from "react";
+import { FeaturePlaceholder } from "ui/editor/FeaturePlaceholder";
+
+const Submissions: React.FC = () => {
+  return (
+    <>
```
DafyddLlyr, line 8: nit: It looks like we have an additional level of fragments here which we don't need

DafyddLlyr (approved):
One crazy small nitpick sorry, but all good otherwise. Appreciate the small and simple PRs thanks 👍

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/index.tsx
Mike-Heneghan: That's interesting! I directly copy and pasted from an existing tab component and just updated the copy 😅 

I can test it out although I can't think of any reason why we'd need the fragment 🤷

Mike-Heneghan: It definitely seems like the fragment wrapper was unnecessary so I've removed it and from the associated components here: d67cec818356f245ffa8be7e3bc4f39a09c655cf

Thanks for spotting this @DafyddLlyr, I should have been more observant and nice to add a wee tidy up as part of doing this work 😊

DafyddLlyr: Nice outcome to tidy up the other components also 👍

Mike-Heneghan:
[Allow service owners and product owners to see that an application has been sent successfully](https://trello.com/c/Xf8bi4oN/2444-allow-service-owners-and-product-owners-to-see-that-an-application-has-been-sent-successfully)

PR #2888, opened 2024-03-14
chore: Remove GovPay env vars

DafyddLlyr:
## What does this PR do?
 - Removes GovPay env vars, as values are now being read from `team_integrations` table
 - Follow up to https://github.com/theopensystemslab/planx-new/pull/2790
 - Please note - `.env` files on S3 not yet updated - no rush to do this, as it could break open (non-rebased) Pizzas. I'll pick this up as a follow on task a few days after this PR merges

## To Do (before merging)
- [x] Test #2790 on staging
- [ ] Test #2790 on production


jessicamcinchak (approved):
Love these shrinking env files :scissors: so much friendlier for forks & non-OSL onboarding!

PR #2889, opened 2024-03-14
fix: reinstate allow_list_answers on submission_services_summary view

Mike-Heneghan:
## What

- Reinstated `allow_list_answers`
- Add `GRANT SELECT` permission to `submission_services_summary` as per https://github.com/theopensystemslab/planx-new/blob/main/doc/how-to/how-to-grant-metabase-permissions.md#views

## Why

- Accidentally removed the `allow_list_answers` from the view by basing updates on an old version of the view

### DIFF
--- a/hasura.planx.uk/migrations/1710431558909_alter_view_public_submission_services_summary_reinstate_allow_list_answers/up.sql
+++ b/hasura.planx.uk/migrations/1710431558909_alter_view_public_submission_services_summary_reinstate_allow_list_answers/up.sql
@@ -1,0 +1,119 @@
+drop view "public"."submission_services_summary";
+
+create or replace view "public"."submission_services_summary" as
+ with resumes_per_session as (
+    select 
+        session_id,
+        count(id) as number_times_resumed
+    from reconciliation_requests
+    group by session_id
+), bops_agg as (
+    select
+        session_id,
+        json_agg(
+            json_build_object(
+                'id', bops_id,
+                'submittedAt', created_at,
+                'destinationUrl', destination_url
+            ) order by created_at desc
+        ) as bops_applications
+    from bops_applications
+    group by session_id
+), email_agg as (
+    select
+        session_id,
+        json_agg(
+            json_build_object(
+                'id', id,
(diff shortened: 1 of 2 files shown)


Mike-Heneghan:
> I _think_ we need the additional step here in the `up.sql` to `GRANT SELECT ON public.submission_services_summary TO metabase_read_only;` as the view is being dropped.

I was going to run that manually but it's a great idea to just add it to the migrations 👍

Mike-Heneghan:
> > I _think_ we need the additional step here in the `up.sql` to `GRANT SELECT ON public.submission_services_summary TO metabase_read_only;` as the view is being dropped.
> 
> I was going to run that manually but it's a great idea to just add it to the migrations 👍

Now that I reread the docs I see that was what you were suggesting from the outset. In my limited imagination I assumed this would be a manual process via the Hasura SQL 😅 

I've added the `GRANT` after the view is created which is hopefully the right way round?

hasura.planx.uk/migrations/1710431558909_alter_view_public_submission_services_summary_reinstate_allow_list_answers/up.sql
```diff
@@ -0,0 +1,119 @@
...
+ left join payment_requests_agg pr on pr.session_id = ls.id
+ left join payment_status_agg ps on ps.session_id = ls.id
+ left join bops_agg ba on ba.session_id = ls.id::text
+ left join uniform_agg ua on ua.submission_reference = ls.id::text
+ left join email_agg ea on ea.session_id = ls.id
+where f.slug is not null
+ and t.slug is not null;
+
+--  After recreating the view grant Metabase access to it
+GRANT SELECT ON public.submission_services_summary TO metabase_read_only;
```
DafyddLlyr, line 119: Perfect! It's a bit of an annoying step to remember but I'm more confident in our ability to catch it at code review than to consistently do it manually on staging and prod deploys...! 😅

DafyddLlyr (approved):
Perfect - thanks!

PR #2892, opened 2024-03-15
fix: disable Locally Listed Buildings query from Planning Data until wider availability

jessicamcinchak:


jessicamcinchak:
[Remove locally listed buildings (temporarily) on PlanX](https://trello.com/c/FaVVO65d/2844-remove-locally-listed-buildings-temporarily-on-planx)

PR #2908, opened 2024-03-15
fix: CVE-2024-28849

DafyddLlyr:
- Resolves a number of dependabot security issues
- Bumps libraries using `follow-redirects` where it's already patched (e.g. Axios), otherwise falls back to a pnpm override


jessicamcinchak (approved):
Thanks for keeping up with these!

PR #2911, opened 2024-03-20
feat: Send metadata to GovPay

DafyddLlyr:


jessicamcinchak (approved):
Nice & simple solution :raised_hands:

PR #2912, opened 2024-03-20
feat: Update logic graph styling

ianjon3s:
# What does this PR do?

Visual changes to the logic graph, summarised as:

- Increased contrast to nodes, branches and add-node links
- Standardised padding across hangers and branches to increase visibility of structure
- Updated styling of portal nodes to show clear demarcation between node link (enter portal) and menu link (configure portal)
- Only questions and checklists without answers present in sticky-note styling
- Section nodes have dashed border to add subtle visual organisation to graph
- Assigned flag colours no longer apply to full background of an option - reduces contrast issue (made more acute by small size)
- GOV.UK focus style on nodes and add-node links
- Clearer green outline to mark traversed nodes

### Preview

Before:
image

After:
image

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Question.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Question.tsx
@@ -57,7 +57,7 @@ const Question: React.FC<Props> = React.memo((props) => {
     <>
       <Hanger hidden={isDragging} before={props.id} parent={parent} />
       <li
-        className={classNames("card", "decision", {
+        className={classNames("card", "decision", "type" + props.type, {
           isDragging,
           isClone: isClone(props.id),
           isNote: childNodes.length === 0,
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Question.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Question.tsx
@@ -57,13 +57,18 @@ const Question: React.FC<Props> = React.memo((props) => {
     <>
       <Hanger hidden={isDragging} before={props.id} parent={parent} />
       <li
-        className={classNames("card", "decision", "type" + props.type, {
-          isDragging,
-          isClone: isClone(props.id),
-          isNote: childNodes.length === 0,
-          wasVisited: props.wasVisited,
-          hasFailed: props.hasFailed,
-        })}
+        className={classNames(
+          "card",
+          "decision",
+          "type-" + TYPES[props.type as TYPES],
+          {
+            isDragging,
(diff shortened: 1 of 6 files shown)


DafyddLlyr:
@ianjon3s Destroyed and recreated the Vultr instance here this morning as I noticed it was failing.

Let us know when this is ready for review as I know you're keen to move it forward 👍

ianjon3s:
[Restyling the graph in the Editor](https://trello.com/c/DEeTU44m/2887-restyling-the-graph-in-the-editor)

ianjon3s:
Thanks @DafyddLlyr ! I've marked this as ready-to-review as I believe it is in good shape from a code perspective.

editor.planx.uk/src/pages/FlowEditor/floweditor.scss
DafyddLlyr, line 1: A nice refactor in future would be to replace this with styled components to match the rest of the codebase. Not a small task and no rush really, but it would be nice from a consistency point of view.

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Question.tsx
```diff
@@ -57,13 +57,18 @@ const Question: React.FC<Props> = React.memo((props) => {
...
-          isDragging,
-          isClone: isClone(props.id),
-          isNote: childNodes.length === 0,
-          wasVisited: props.wasVisited,
-          hasFailed: props.hasFailed,
-        })}
+        className={classNames(
+          "card",
+          "decision",
+          "type-" + TYPES[props.type as TYPES],
```
DafyddLlyr, line 63: Ha! I forgot about this neat little trick.

I _think_ this means we need to update the "how to add a component" guide to say we need to add a style for a new component also (not sure how this would handle defaults / new components right now).

DafyddLlyr (approved):

DafyddLlyr:
Looking great!

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Question.tsx
ianjon3s: In this case the style is only being used/applied as an enhancement for `type-Section`, so we wouldn't need a style unless it needed to be distinctive from the other components. As long as the type is set on a new component it'll pull through and be used if necessary (if I understand your point correctly).

editor.planx.uk/src/pages/FlowEditor/floweditor.scss
ianjon3s: Agreed, I'll see about doing this as a longer-term piece of work.

PR #2914, opened 2024-03-21
feat: Append, replace, removeOne, & removeAll operations for `SetValue`

DafyddLlyr:
## What does this PR do?
 - Changes SetValue component from having a single operation, to having an option to append, replace, removeOne or removeAll

## Testing
Basic test flow here - https://2914.planx.pizza/testing/set-value-playground

This (hopefully!) illustrates that the behaviour matches the desired outcomes discussion at our recent "Logic Camp" - full details here - https://miro.com/app/board/uXjVN9K07PQ=/?moveToWidget=3458764588027715730&cot=14

I feel like there's a decent level of unit testing here, as well as more psuedo-integration level tests calling store functions to mock a user's journey.

What I'm really interested in testing feedback on would be - 
- Unexpected, or breaking, changes in existing flows - possibly related to edge cases with current SetValue
- Is there something missing in how multiple SetValue components interact with each other?
- Anything that could break this!

## Next steps...
- Illustrative examples in Editor modar, similar to how Calculate shows examples to the user

image

### DIFF
--- a/editor.planx.uk/src/@planx/components/SetValue/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/SetValue/Editor.tsx
@@ -68,20 +68,18 @@ function SetValueComponent(props: Props) {
             />
           </InputRow>
         </ModalSectionContent>
-        {formik.values.operation !== "remove" && (
-          <ModalSectionContent title="Field value">
-            <InputRow>
-              <Input
-                required
-                format="data"
-                name="val"
-                value={formik.values.val}
-                placeholder="value"
-                onChange={formik.handleChange}
-              />
-            </InputRow>
-          </ModalSectionContent>
-        )}
+        <ModalSectionContent title="Field value">
+          <InputRow>
+            <Input
+              required
+              format="data"
+              name="val"
+              value={formik.values.val}
+              placeholder="value"
+              onChange={formik.handleChange}
+            />
+          </InputRow>
+        </ModalSectionContent>
         <ModalSectionContent title="Operation">
           <DescriptionText {...formik.values} />
           <Radio
--- a/editor.planx.uk/src/@planx/components/SetValue/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/SetValue/Editor.tsx
@@ -15,6 +15,30 @@ type Props = EditorProps<TYPES.SetValue, SetValue>;
 
 export default SetValueComponent;
 
+interface Option {
+  value: SetValue["operation"];
+  label: string;
+}
+
+const options: Option[] = [
+  {
+    value: "replace",
+    label: "Replace",
+  },
+  {
+    value: "append",
+    label: "Append",
+  },
+  {
+    value: "removeOne",
+    label: "Remove single value",
+  },
+  {
+    value: "removeAll",
+    label: "Remove all values",
+  },
+];
+
 const DescriptionText: React.FC<SetValue> = ({ fn, val, operation }) => {
   if (!fn || !val) return null;
 
@@ -33,11 +57,17 @@ const DescriptionText: React.FC<SetValue> = ({ fn, val, operation }) => {
           <strong>{val}</strong> appended to it
         </Typography>
       );
-    case "remove":
+    case "removeOne":
       return (
         <Typography mb={2}>
           Any existing value for <strong>{fn}</strong> set to{" "}
           <strong>{val}</strong> will be removed
+        </Typography>
+      );
+    case "removeAll":
+      return (
+        <Typography mb={2}>
+          All existing values for <strong>{fn}</strong> will be removed
         </Typography>
       );
   }
@@ -84,20 +114,7 @@ function SetValueComponent(props: Props) {
         <ModalSectionContent title="Operation">
           <DescriptionText {...formik.values} />
           <Radio
-            options={[
-              {
-                value: "replace",
-                label: "Replace",
-              },
-              {
-                value: "append",
-                label: "Append",
-              },
-              {
-                value: "remove",
-                label: "Remove",
-              },
-            ]}
+            options={options}
             value={formik.values.operation}
             onChange={(newOperation) => {
               formik.setFieldValue("operation", newOperation);
--- a/editor.planx.uk/src/@planx/components/SetValue/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/SetValue/Editor.tsx
@@ -36,7 +36,8 @@ const DescriptionText: React.FC<SetValue> = ({ fn, val, operation }) => {
     case "remove":
       return (
         <Typography mb={2}>
-          Any existing value for <strong>{fn}</strong> will be removed
+          Any existing value for <strong>{fn}</strong> set to{" "}
+          <strong>{val}</strong> will be removed
         </Typography>
       );
   }
--- a/editor.planx.uk/src/@planx/components/SetValue/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/SetValue/Editor.tsx
@@ -21,22 +21,22 @@ const DescriptionText: React.FC<SetValue> = ({ fn, val, operation }) => {
   switch (operation) {
     case "replace":
       return (
-        <Typography mt={2}>
-          any existing value for <strong>{fn}</strong> will be replaced by{" "}
+        <Typography mb={2}>
+          Any existing value for <strong>{fn}</strong> will be replaced by{" "}
           <strong>{val}</strong>
         </Typography>
       );
     case "append":
       return (
-        <Typography mt={2}>
-          any existing value for <strong>{fn}</strong> will have{" "}
+        <Typography mb={2}>
+          Any existing value for <strong>{fn}</strong> will have{" "}
           <strong>{val}</strong> appended to it
         </Typography>
       );
     case "remove":
       return (
-        <Typography mt={2}>
-          any existing value for <strong>{fn}</strong> will be removed
+        <Typography mb={2}>
+          Any existing value for <strong>{fn}</strong> will be removed
         </Typography>
       );
   }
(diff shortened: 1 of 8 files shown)


editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -249,14 +246,35 @@ export const previewStore: StateCreator<
...
+          }
+
+          if (operation === "append") {
+            const previousValue = acc.data?.[fn] || [];
+            const currentValue = responseData?.[fn] || [];
+            const combined = [...previousValue, ...currentValue];
+
+            passport.data![fn] = combined;
+          }
+        }
```
DafyddLlyr, line 275: I'm not massively keen on making `computePassport()` much longer or more complex, but hopefully this is fairly self contained?

If inclear, happy to move out to a helper function!

editor.planx.uk/src/@planx/components/SetValue/Editor.tsx
```diff
@@ -6,13 +7,41 @@ import ModalSection from "ui/editor/ModalSection";
...
+      return (
+        <Typography mb={2}>
+          Any existing value for <strong>{fn}</strong> will have{" "}
+          <strong>{val}</strong> appended to it
+        </Typography>
+      );
+    case "remove":
+      return (
+        <Typography mb={2}>
+          Any existing value for <strong>{fn}</strong> will be removed
```
jessicamcinchak, line 39: Can `remove` also support an `fn` _and_ `val`  ?

Eg see this example https://opensystemslab.slack.com/archives/C5Q59R3HB/p1701766927022339 - ideally this solution let's us remove need for project types like `not.alter.roofLight` etc !

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
jessicamcinchak: Rather than adding this logic to the Preview store, what if it lived directly in SetValue's `handleSubmit` ? Eg similar to Calculate and the use of `props.formatOutputForAutomations`? 

Interested in a future of smarter components, dumber stores!

jessicamcinchak (commented):
V exciting :tada: a couple questions after a quick first pass!

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
DafyddLlyr: Caught up about this one to discuss options. This isn't easily possible as it would require modifying past breadcrumbs to effectively handle the "remove" operation.

editor.planx.uk/src/@planx/components/SetValue/Editor.tsx
DafyddLlyr: Change now made 👍

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -249,14 +247,37 @@ export const previewStore: StateCreator<
...
+        const isSetValue = flow[id].type === TYPES.SetValue;
+
+        if (isSetValue) {
+          const { operation, fn } = flow[id]?.data as SetValue;
+          const previousValues = acc.data?.[fn] || [];
+          const currentValue = responseData?.[fn] || [];
+
+          if (operation === "remove") {
+            const removeCurrentValue = (val: string) => val !== currentValue[0];
+            const filtered = previousValues.filter(removeCurrentValue);
```
Mike-Heneghan, line 268: _Edge Case_:

I don't think it's very likely to happen but if a user wants to `append` or `remove` a passport `fn` which is previously not an array there's an error thrown:

Screenshot 2024-03-25 at 15 58 34

Screenshot 2024-03-25 at 16 00 32

Not very likely to happen but I thought I'd flag it.

editor.planx.uk/src/@planx/components/SetValue/Editor.tsx
Mike-Heneghan, line 1: _nit_:

I think the functionality for remove has changed but the editor text still has the "Any existing value..." message.

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -823,6 +828,45 @@ export const sortBreadcrumbs = (
...
+  const isObject =
+    typeof previousValues === "object" &&
+    !Array.isArray(previousValues) &&
+    previousValues !== null;
+  if (isObject) return passport;
+
+  previousValues = formatPreviousValues(previousValues);
+  const currentValue = responseData?.[fn] || [];
+
+  if (operation === "remove") {
```
Mike-Heneghan, line 853: Unfortunately I haven't been able to replicate the slightly odd behaviour I was seeing in a clear simple flow. Which potentially suggests it's too much of an edge case to worry about. 

On the pizza I've updated [FOIYNPP](https://2914.planx.pizza/lambeth/find-out-if-you-need-planning-permission) to add some extra `setValue` nodes and dummy questions to act as moments to run `computePassport()` and assess the values. 

What I think I'm seeing is that `findOut` isn't being removed from `application.type` and after the `append` later on it's appearing as the second item in the array. 

I feel like I must be missing something 🤷

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Node.tsx
```diff
@@ -145,6 +140,17 @@ const Node: React.FC<any> = (props) => {
   }
 };
 
+const getSetValueText = ({ operation, fn, val }: Store.node["data"]) => {
+  switch (operation) {
+    case "append":
+      return `Append ${val} to ${fn}`;
+    case "remove":
+      return `Remove ${fn}`;
```
Mike-Heneghan, line 148: _nit_: 

It would be handy to have the `Remove ${val} from `${fn}` here if you have the chance.

editor.planx.uk/src/@planx/components/SetValue/Editor.tsx
DafyddLlyr: See [4ea07af](https://github.com/theopensystemslab/planx-new/pull/2914/commits/4ea07af0cbeece87e967592b167b5db721c8d1fb)

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
DafyddLlyr: Very good catch! Please see [b598d7e](https://github.com/theopensystemslab/planx-new/pull/2914/commits/b598d7eb821c6777acd41249b4417f942b03ea67)

To follow - tests for non-arrays.

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Node.tsx
DafyddLlyr: Good catch! See [367fd74](https://github.com/theopensystemslab/planx-new/pull/2914/commits/367fd74a0a8c2425005596632e1184c490c7dac8)

editor.planx.uk/src/@planx/components/SetValue/Editor.tsx
jessicamcinchak: Revisiting this one and realising the `val` suggestion only makes sense when the `fn` is an array of schema-controlled `val` - we'll need to keep thinking about this for eg `NumberInputs` which set can any string. In that case, I think expected behavior would be to remove based on `fn` alone!

Similar for "replace" - currently it seems that string passport `fns` are converted to an array if a `SetValue` attempts to "Replace" them!

editor.planx.uk/src/@planx/components/SetValue/utils.test.ts
DafyddLlyr, line 1: This file should represent the inputs/outputs described in this "Logic Camp" diagram - 

image

Source: https://miro.com/app/board/uXjVN9K07PQ=/?moveToWidget=3458764578921902831&cot=14

editor.planx.uk/src/@planx/components/SetValue/Editor.tsx
DafyddLlyr: This has now been addressed 👍

The `SetValue` will always convert the output of a _modified_ value to an array - this means that it's outputs can be used for automations. If an operation cannot be completed, or is nonsensical (e.g. removing an item from a single string value) the output will not be modified.

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
DafyddLlyr: This has now been resolved!

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
DafyddLlyr: I've not been able to replicate this, but will set up a series of test flows to manual testing.

editor.planx.uk/src/@planx/components/SetValue/utils.ts
```diff
@@ -0,0 +1,89 @@
...
+  nodeData: SetValue;
+  previousValues: PreviousValues;
+  passport: Store.passport;
+}) => Store.passport;
+
+/**
+ * Handle modifying passport values when passing through a SetValue component
+ * Called by computePassport()
+ */
+export const handleSetValue: HandleSetValue = ({
```
DafyddLlyr, line 16: Following [previous conversations](https://github.com/theopensystemslab/planx-new/pull/2914#discussion_r1534124245), I've refactored this logic to be co-located with the component rather than being purely within the store.

The motivations here were two-fold - 
 - This is SetValue logic and doesn't relate to any other components - it just can't get called by this component directly
 - We actually want to use some of this code within the Editor modal to show the Editor examples of how this component works (upcoming PR).

jessicamcinchak (approved):
Sorry it's taken me a minute to get back to this one. Most recent changes all look good here - any final testing/coordination needed with Content team before merging? Let's aim to get this out in the wild this week!

PR #2917, opened 2024-03-22
test: add required inputs to teardown workflow

freemvmt:


freemvmt:
does teardown destroy DNS (not optimistic)

PR #2872, opened 2024-03-07
fix: update v1 bops mocks in e2e regression tests

jessicamcinchak:
Replaced and ultimately fixed by #2918, but I think we should still keep session mocks up-to-date & valid in meantime. 

Regressions tests passing against this branch after rebasing :heavy_check_mark:  https://github.com/theopensystemslab/planx-new/actions/runs/8432507577

PR #2918, opened 2024-03-25
fix: Failing regression test

DafyddLlyr:


e2e/tests/api-driven/src/globalHelpers.ts
```diff
@@ -10,6 +10,7 @@ export function createTeam(
       slug: "E2E",
       submissionEmail: TEST_EMAIL,
       homepage: "planx.uk",
+      referenceCode: "ABCD",
```
DafyddLlyr, line 13: This is a mandatory field to generate a valid payload. `planx-core` now makes this required with this change - https://github.com/theopensystemslab/planx-core/pull/337

e2e/tests/api-driven/src/invite-to-pay/mocks/flow.json
DafyddLlyr, line 1: I copy/pasted the most recent version of the flow here to resolve schema issues.

I needed to manually update the `Pay` component with the `govPayMetadata` field - I'll open a PR shortly to either properly handle defaults or to make a data migration.

jessicamcinchak: I'm a little confused by this - is it actually an issue of outdated _flow_ mocks or the mock session data referenced by the graphql query mocks (or both)? Eg I'm surprised to not see these updates repeated here  https://github.com/theopensystemslab/planx-new/pull/2872/files

jessicamcinchak (approved):
Thanks for looking into this one - one comment /question re: mock _flows_ versus _sessions_ here - but if CI is satisfied then happy for this fix to merge whenever!

e2e/tests/api-driven/src/invite-to-pay/mocks/flow.json
DafyddLlyr: The only errors were coming from outdated flow mocks, which is what I've updated here. I was also surprised by this, but I guess the variables in question are optional?

PR #2919, opened 2024-03-25
fix(temp): Pin `hasura-cli` version

DafyddLlyr:
Please see https://opensystemslab.slack.com/archives/C01E3AC0C03/p1711359461634859 for context.

Looks like there's a few issues with the latest releases - https://www.npmjs.com/package/hasura-cli?activeTab=versions

image

Suggesting this as a temp fix for now, and I'll check in on it next few days. If it lasts any longer I'll open an issue in the `hasura-cli` repo.


freemvmt (approved):
Makes sense! Probably a good idea for this to be pinned regardless I think :)

PR #2920, opened 2024-03-25
fix(wip): Broken address search

DafyddLlyr:
Rolling back change to Editor made here - https://github.com/theopensystemslab/planx-new/pull/2908

Pretty unsure if this is the fix, but it's worth merging to `main` to test, then we can at least rule this one out 👍


jessicamcinchak (approved):
Thanks for queuing up this test - pizza working as expected, happy to test again on staging after merge!

PR #2941, opened 2024-03-28
fix(a11y): `ResultReason` tabindex order

DafyddLlyr:
## Problem
image

## Solution
 - Restructure DOM so the order of elements is correct (move change button to come "after" accordion)
 - Use `useRef()` and `useLayoutEffect()` hooks to maintain consistent style - change button remains in same location

## Demo
https://github.com/theopensystemslab/planx-new/assets/20502206/0e18e26a-9bbd-4b1e-a2e7-e481bf74c740

🧪 Test flow: https://2941.planx.pizza/testing/result-tab-index-order/amber

## Also...
There were a few lines controlling the logic around `expanded` and `onChange` which are redundant - the MUI accordion has this as the default behaviour anyway, so I've removed this.


DafyddLlyr:
[Order of components [A]](https://trello.com/c/eQ6l2f6G/2810-order-of-components-a)

ianjon3s (approved):
This is a really strong solution. I had attempted to reorder the components as a quick solution and had identified the need to inherit the height to prevent the chevron shifting on expand. In the jQuery days this would be injected after the html render (not ideal), good to see the powers of React rendering used here.

PR #2942, opened 2024-03-28
chore: default analytics_logs allow_list_answers to null

Mike-Heneghan:
## What

- Default `analytics_logs` `allow_list_answers` to `null`
- Update existing empty arrays to `null`

## Why

- In Metabase the empty areas were found to be challenging to filter out

## Note

- Initially intended to allow do this for `metadata` and `input_errors` as per the ticket although due to the mutation using `_append` and the nature of those tracking events it wasn't practical to do.


Mike-Heneghan:
[Update analytics_logs JSONB default values to `null`](https://trello.com/c/zKgDcQAK/2860-update-analyticslogs-jsonb-default-values-to-null)

DafyddLlyr (approved):

PR #2945, opened 2024-03-28
refactor: store analytics_logs allow_list_answers from arrays of objects to object

Mike-Heneghan:
## What

- Store `allow_list_answers` as an object
- Update `getAnswers` to filter out `null` `answerValues` to avoid storing `null` values: https://metabase.editor.planx.uk/question/353-null-allow-list-values
- Update historical records to move form array of object to an object:
  - The SQL for handling an array greater than 1 was very challenging
  - Realised that although it's theoretically possible to have `allow_list_answers` > 1 none of our ALLOW_LIST actually gets set by the same node: https://metabase.editor.planx.uk/question/354-number-of-analytics-logs-with-allow-list-answers-greater-than-1
  - This allows us to use the more naive approach here:  5d2cec68f179e635b5b194d6817b579616270eaf
- Update `analytics_summary` to handle the change in data structure of `allow_list_answers`
- Remove `node_fn` from tracking 
  
## Why

- It was found in Metabase / SQL that the `allow_list_answers` as an array is harder to work with
- Useful to align the data structure to that of `lowcal_sessions`
- Currently, we are occasionally storing `null` values from some edge case nodes

## Follow Up

- Although the changes in this PR should stop `null` being stored in `allow_list_answers` it doesn't cleanup historical data which will be done in a follow up PR.

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
@@ -208,7 +208,6 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     const metadata: NodeMetadata = getNodeMetadata(nodeToTrack, nodeId);
     const nodeType = nodeToTrack?.type ? TYPES[nodeToTrack.type] : null;
     const nodeTitle = extractNodeTitle(nodeToTrack);
-    const nodeFn = nodeToTrack?.data?.fn || nodeToTrack?.data?.val;
 
     const result = await insertNewAnalyticsLog(
       logDirection,
@@ -217,7 +216,6 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
       nodeType,
       nodeTitle,
       nodeId,
-      nodeFn,
     );
 
     const { id, created_at: newLogCreatedAt } =
@@ -250,7 +248,6 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     nodeType: string | null,
     nodeTitle: string,
     nodeId: string | null,
-    nodeFn: string | null,
   ) {
     const result = await publicClient.mutate({
       mutation: gql`
@@ -261,7 +258,6 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
           $node_type: String
           $node_title: String
           $node_id: String
-          $node_fn: String
         ) {
           insert_analytics_logs_one(
             object: {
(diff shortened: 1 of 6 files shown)


Mike-Heneghan:
[Align `allow_list_answers` data in `analytics_logs` to the structure in the `lowcal_sessions`.](https://trello.com/c/kbQ5lwFK/2878-align-allowlistanswers-data-in-analyticslogs-to-the-structure-in-the-lowcalsessions)

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -497,32 +493,38 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
...
-  function getAnswers(nodeId: string) {
+  function getAnswers(
+    nodeId: string,
+  ): Partial<Record<AllowListKey, any>> | undefined {
     const { data } = flow[nodeId];
-    const nodeFn: string = data?.fn || data?.val;
-    if (!nodeFn || !ALLOW_LIST.includes(nodeFn)) return [];
+    const nodeFn: string | undefined = data?.fn || data?.val;
+
+    if (!nodeFn || !ALLOW_LIST.includes(nodeFn as AllowListKey)) return;
```
DafyddLlyr, line 516:
```suggestion
       const isAllowListKey = (nodeFn: string): nodeFn is AllowListKey => {
      return (ALLOW_LIST as readonly string[]).includes(nodeFn);
    };

    if (!nodeFn || !isAllowListKey(nodeFn)) return;
```

If we add a type guard here instead of just `.includes()` we'll get type narrowing and can drop the casting to `AllowListKey` on line 526.

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
```diff
@@ -531,13 +533,21 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
...
+    breadcrumb: Store.userData,
+  ): Partial<Record<AllowListKey, any>> | undefined {
     const dataSetByNode = breadcrumb.data;
-    if (!dataSetByNode) return [];
+    if (!dataSetByNode) return;
+
+    const filteredEntries = Object.entries(dataSetByNode)
+      .filter(
+        ([key, value]) =>
+          ALLOW_LIST.includes(key as AllowListKey) && Boolean(value),
```
DafyddLlyr, line 545: See previous comment on using a type guard (`is` function) - we can probably use one here to avoid casting twice.

DafyddLlyr (approved):
Small comment on some possible TypeScript tidy ups but looking good!

editor.planx.uk/src/pages/FlowEditor/lib/analyticsProvider.tsx
Mike-Heneghan: Should also be sorted by: 9bccce1bbfd6b0271de9d7796f7b3bed4791a779

DafyddLlyr (approved):

PR #2946, opened 2024-03-28
fix(a11y): Split `ConstraintsList` into multiple `ul` groupings

DafyddLlyr:
## Problem
image

## Solution
- Make each `groupedConstraint` it's own `List` (`ul`)
- Move title out of the list. As it's a `h3` and in the correct order in the DOM I don't think we need aria-controls here - it should already be semantically meaningful.
- Zero visual changes, just structural / hierarchical changes

## Next steps...
The solution suggested by the report is as follows - 

> Ensure an appropriate list structure is implemented for each section introduced by a heading.
> As these components function similarly to accordions, we advise following the ARIA Authoring Practices Guide (APG) accessible accordion example found on their website.

I'll open another PR to look at how the accordions are set up and if we need additional aria-controls (e.g. aria-expanded)

Update: PR here - https://github.com/theopensystemslab/planx-new/pull/2966

### DIFF
--- a/editor.planx.uk/src/@planx/components/PlanningConstraints/List.tsx
+++ b/editor.planx.uk/src/@planx/components/PlanningConstraints/List.tsx
@@ -64,35 +64,36 @@ export default function ConstraintsList({
 
   return (
     <Box mb={3}>
-      <List dense disablePadding>
-        {Object.keys(groupedConstraints).map(
-          (category: string, index: number) => (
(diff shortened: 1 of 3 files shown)


DafyddLlyr:
[List structure [A]](https://trello.com/c/44V5wcN8/2805-list-structure-a)

ianjon3s (approved):

PR #2961, opened 2024-04-02
chore: Tidy up unused Editor dependencies

DafyddLlyr:
Noticed that https://github.com/theopensystemslab/planx-new/pull/2956 was attempting to update `marked` which isn't imported anywhere in the Editor.

I used `npx depcheck` to identify unused imports (https://www.npmjs.com/package/depcheck) and then manually checked if any of these are imported/used anywhere.


editor.planx.uk/package.json
```diff
@@ -10,7 +10,6 @@
     "@emotion/styled": "^11.11.0",
     "@mui/icons-material": "^5.15.2",
     "@mui/material": "^5.15.2",
-    "@mui/styles": "^5.15.2",
```
jessicamcinchak, line 13: this feels like a big one / nice milestone on our MUI migration to styled components :100:

DafyddLlyr: 💯

PR #2968, opened 2024-04-02
fix(a11y): Add visually hidden context to "Restart Application" button

DafyddLlyr:
image
image

## What does this PR do?
- Adds a visually hidden description to the "Restart application" button

## Context
I'd dispute this is really totally necessary as our current markdown matches [the suggestion on MDN perfectly](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) (`<button>` with `aria-label` wrapping an `<svg>` with `aria-hidden=true`), and clicking the button opens a dialog with more details. However, it's an easy fix to just add an additional description element here.


DafyddLlyr:
[Non-text label [Usability]](https://trello.com/c/Z7YmY10I/2836-non-text-label-usability)

jessicamcinchak (approved):

PR #2969, opened 2024-04-02
fix(a11y): collapse map attributions and wrap SummaryList columns on small screens

jessicamcinchak:
**From page 61:**
> Content is lost when resizing the webpage to a smaller viewing dimension. This instance can 
be found with the site boundary section of the definition list. The map feature has been 
resized however the tooltip provided is covering the user from identifying the boundary of 
the location. This issue affects low vision users who are likely to require a smaller window to 
progress through the journey as content is affected. 
Additionally, due to the change links being covered on the side of the window, part of the 
focus highlight is obscured. 

**Changes:**
- Openlayers map attributions are collapsed on screens less than 500px
- Review page `SummaryList` 3-column layout is collapsed to single column on screens less than 500px ("small" MUI breakpoint), consistent with GOV.UK component behavior

Screenshot from 2024-04-02 13-19-28

### DIFF
--- a/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
@@ -211,6 +211,7 @@ export default function Component(props: Props) {
                 osProxyEndpoint={`${process.env.REACT_APP_API_URL}/proxy/ordnance-survey`}
                 osCopyright={`Basemap subject to Crown copyright and database rights ${new Date().getFullYear()} OS (0)100024857`}
                 drawGeojsonDataCopyright={`<a href="https://www.planning.data.gov.uk/dataset/title-boundary" target="_blank" style="color:#0010A4;">Title boundary</a> subject to Crown copyright and database rights ${new Date().getFullYear()} OS (0)100026316`}
+                collapseAttributions={self.innerWidth < 500 ? true : undefined}
               />
             </MapContainer>
             <MapFooter>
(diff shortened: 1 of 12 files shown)


jessicamcinchak:
[Reflow map attribution [AA]](https://trello.com/c/rW2d45bD/2821-reflow-map-attribution-aa)

editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
```diff
@@ -211,6 +211,7 @@ export default function Component(props: Props) {
                 osProxyEndpoint={`${process.env.REACT_APP_API_URL}/proxy/ordnance-survey`}
                 osCopyright={`Basemap subject to Crown copyright and database rights ${new Date().getFullYear()} OS (0)100024857`}
                 drawGeojsonDataCopyright={`<a href="https://www.planning.data.gov.uk/dataset/title-boundary" target="_blank" style="color:#0010A4;">Title boundary</a> subject to Crown copyright and database rights ${new Date().getFullYear()} OS (0)100026316`}
+                collapseAttributions={self.innerWidth < 500 ? true : undefined}
```
DafyddLlyr, line 214:
```suggestion
                collapseAttributions={self.innerWidth < 500}
```

Would this work?

jessicamcinchak: It's a weird Lit "boolean" thing where the prop explitly needs to evaluate to `true` or `undefined` - `false` will be read as truthy (_existance_ of prop) by Lit !

DafyddLlyr: Aaaah that makes sense - we've hit this before. Thanks for the explanation 👍

PR #2970, opened 2024-04-02
feat: Move back button out of main component, ensure single main component only

DafyddLlyr:
## What does this PR do?
 - Move back component out of `#main` element (report page 91 - [Trello ticket](https://trello.com/c/C0e3PdIu/2835-back-button-positioning-usability))
 - Ensure we only use a single `#main` element (report page 82 - [Trello ticket](https://trello.com/c/ksGgKnK7/2830-multiple-main-roles-usability?search_id=23e5b034-13a3-424f-9c07-fd0dbf48fbb6))
 - Use `<main>` element instead of `<div role="main">` - not in report just something I spotted. From MDN docs - 
 
> Prefer using the `<main>` element over declaring role="main", unless there are [legacy browser support concerns](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main#browser_compatibility).

### DIFF
--- a/editor.planx.uk/src/pages/layout/PublicLayout.tsx
+++ b/editor.planx.uk/src/pages/layout/PublicLayout.tsx
@@ -52,10 +52,10 @@ const PublicFooter: React.FC = () => {
 
   const globalFooterItems = globalSettings?.footerContent
     ? Object.entries(globalSettings?.footerContent).map(([slug, item]) => ({
-        title: item.heading,
-        content: item.content,
-        href: makeHref(slug),
-      }))
+      title: item.heading,
+      content: item.content,
+      href: makeHref(slug),
+    }))
     : [];
 
   const footerItems = [...flowSettingsContent, ...globalFooterItems].filter(
@@ -96,7 +96,7 @@ const PublicLayout: React.FC<PropsWithChildren> = ({ children }) => {
     <StyledEngineProvider injectFirst>
       <ThemeProvider theme={teamMUITheme}>
         <Header />
-        <MainContainer id="main-content">
+        <MainContainer>
           <ErrorBoundary FallbackComponent={ErrorFallback}>
             {children}
           </ErrorBoundary>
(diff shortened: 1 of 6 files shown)


DafyddLlyr:
[Back button positioning [Usability]](https://trello.com/c/C0e3PdIu/2835-back-button-positioning-usability)

DafyddLlyr:
[Multiple main roles [Usability]](https://trello.com/c/ksGgKnK7/2830-multiple-main-roles-usability)

editor.planx.uk/src/pages/layout/PublicLayout.tsx
```diff
@@ -96,7 +96,7 @@ const PublicLayout: React.FC<PropsWithChildren> = ({ children }) => {
     <StyledEngineProvider injectFirst>
       <ThemeProvider theme={teamMUITheme}>
         <Header />
-        <MainContainer id="main-content">
```
Mike-Heneghan, line 99: _Question:_

I think the `Skip To Main Content` might have been broken by this change and the target for the content maybe needs to be updated: https://github.com/theopensystemslab/planx-new/blob/67d6f182634dc5c57804e9971ce290abe3525abb/editor.planx.uk/src/components/Header.tsx#L348

On dev `skip to main content` goes to the `back` button but on the pizza I think it doesn't go anywhere when it's selected. 

https://github.com/theopensystemslab/planx-new/assets/36415632/6d97fa40-a245-4113-b8df-f878029b3097

DafyddLlyr: Thanks for spotting this and testing on the Pizza - this wasn't the result I was getting was getting locally. I think that as `Node` just yields to a child it wasn't handling the new properties correctly.

This is fixed by making an explicit main component in [0f01730](https://github.com/theopensystemslab/planx-new/pull/2970/commits/0f01730b33ba909b525e74f9d7e0bc31d74ac2f0) 👍

Mike-Heneghan (approved):
Looks great and working as expected for me 🥳

PR #2972, opened 2024-04-02
chore: Fix eslint rule "no-case-declarations"

DafyddLlyr:
Fixes https://eslint.org/docs/latest/rules/no-case-declarations by just wrapping the cases in `{}` as suggested.

Also removes a few unused imports 🗑️


Mike-Heneghan (approved):
Looks good 👍

PR #2973, opened 2024-04-03
fix(a11y): `FileUploadAndLabel` error messages

DafyddLlyr:
## What does this PR do?
 - Displays an error per-unlabeled file in both modal and public component

## Context
From accessibility report, page 39 - 

> An error caused by not applying any tags to a document are not easily identifiable through use of the error text. This instance can be found when not including tags to uploaded documents. The error that appears has been placed above the done button, without a status message, and does not relate to the input to which the error has been caused by. This issue affects screen reader users as the error message does not immediately explain what is wrong, not does it highlight the input relating to the issue.

The "alert" role is added in this PR - https://github.com/theopensystemslab/planx-new/pull/2974

## Demo
**Per-file error messages in modal**
image

**Per-file error message in Public component**
image

**Validation for whole list still in place**
image


Mike-Heneghan (approved):
Looks great, seems like a really useful change 🥳

DafyddLlyr:
[Error identification for Upload&Label [A]](https://trello.com/c/4rqZjL5Q/2813-error-identification-for-uploadlabel-a)

PR #2974, opened 2024-04-03
fix(a11y): Use `role="alert"` as default value for `ErrorWrapper`

DafyddLlyr:
## What does this PR do?
 - Adds an optional `role` property to the `ErrorWrapper` component, defaulting to `"alert"`
 - (I believe) `ErrorWrapper` is only currently used for input feedback errors, but I've allowed "status" to be a valid role prop if there are exceptions in future

## Why is this necessary?
 - Recommended several times throughout accessibility report (pages [40](https://trello.com/c/4rqZjL5Q/2813-error-identification-for-uploadlabel-a) & [53](https://trello.com/c/THkjn5rm/2818-status-messages-aa))
 - MDN docs show that "alert" is correct for form feedback - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role
 
 > The alert role is used to communicate an important and usually time-sensitive message to the user. When this role is added to an element, the browser will send out an accessible alert event to assistive technology products which can then notify the user.
>
> The alert role should only be used for information that requires the user's immediate attention, for example:
>
> - An invalid value was entered into a form field
> - The user's login session is about to expire
> - The connection to the server was lost so local changes will not be saved

### DIFF
--- a/editor.planx.uk/src/@planx/components/DateInput/Public.test.tsx
+++ b/editor.planx.uk/src/@planx/components/DateInput/Public.test.tsx
@@ -179,7 +179,7 @@ it("should not have any accessibility violations whilst in the error state", asy
   // Main ErrorWrapper does display, and is in error state
   await waitFor(() => expect(mainErrorMessage).not.toBeEmptyDOMElement());
   const [mainErrorWrapper, ..._rest] = screen.getAllByTestId("error-wrapper");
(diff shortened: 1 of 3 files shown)


Mike-Heneghan (approved):
Looks good to me 👍

PR #2977, opened 2024-04-04
chore: Drop `SUBMISSION_VIEW` feature flag

DafyddLlyr:
## What does this PR do?
 - Drop feature flag to create Pizza for UAT
 - Some type tidy ups
 
In this instance we’re better off removing the feature flag and encouraging PO testing on production as - 
- there’s real submission data there to test against
 - it’s a pretty unobtrusive feature - if it needs changes we can easily do this without effecting anything else


DafyddLlyr:
[Allow servicer owners and product owners to to see than an application has been sent successfully](https://trello.com/c/7e94HTns/2871-allow-servicer-owners-and-product-owners-to-to-see-than-an-application-has-been-sent-successfully)

Mike-Heneghan (approved):
Looks great!

PR #2980, opened 2024-04-04
fix: allow larger basemap buffer for large sites

jessicamcinchak:
Now determines the basemap clip buffer based on the area of the site, meaning we don't unnecessarily expose more vector tiles to regular householder applicants.

Ticket only has one example address (5 Pancras Square, N1C 4AG) which is now fully in view if viewing the default title boundary: 
Screenshot from 2024-04-04 12-47-49


jessicamcinchak:
[Allow users to zoom out and move the location plan further](https://trello.com/c/667gGc0d/2799-allow-users-to-zoom-out-and-move-the-location-plan-further)

DafyddLlyr (approved):
Perfect 👍

PR #2982, opened 2024-04-04
fix(a11y): enforce positive numbers by default on NumberInput

jessicamcinchak:
**From page 84:** 
> An input accepts negative numbers where it shouldn’t be possible.
> 
> **Solution:**
> We suggest implementing an error message similarly to how other errors have been
implemented for the negative values.

**Changes:** 
- `NumberInput` should enforce positive numbers by default & show a specific error message (will flag this to George & August)
- Adds an Editor toggle to "Allow negative numbers", which is "off" by default

Screenshot from 2024-04-04 13-58-33

Screenshot from 2024-04-04 13-58-48


jessicamcinchak:
[Negative numbers [Usability]](https://trello.com/c/pgjA90gK/2831-negative-numbers-usability)

DafyddLlyr (approved):
Perfect 👌

PR #2983, opened 2024-04-04
chore: Remove close tab option

DafyddLlyr:
## What does this PR do?
- Removes "Close tab" option 
- Replaces with a message ("You may now close this tab") and makes "Start a new application" the primary action

## Why?
- Reported as a bug, does not work in Firefox and Edge
- It turns out `window.close()` is not allowed by these browsers - https://caniuse.com/mdn-api_window_close

image


DafyddLlyr:
[Fix "Close tab" button on Save & Return confirmation page](https://trello.com/c/qhiKlCZL/2868-fix-close-tab-button-on-save-return-confirmation-page)

ianjon3s (approved):
Seems a solid solution. Closing a browser/tab from within browser content is (I would estimate) a very niche interaction in any case.

PR #2988, opened 2024-04-04
feat: track feedback on help text usefulness

Mike-Heneghan:
## What

- Track when a user provides feedback on the helpfulness of help text
- Update the `analytics_summary` to expose this new metadata in it's own column

## Why

- Although we record feedback on help text it's necessary for user to explicitly leave a comment and it's not accessible in the `analytics_logs`
- Tracking "yes" / "no" clicks via analytics give us a higher level view of whether users are finding help text useful with a lower barrier

### DIFF
--- a/editor.planx.uk/src/components/Feedback/MoreInfoFeedback.tsx
+++ b/editor.planx.uk/src/components/Feedback/MoreInfoFeedback.tsx
@@ -8,6 +8,7 @@ import {
   getInternalFeedbackMetadata,
   insertFeedbackMutation,
 } from "lib/feedback";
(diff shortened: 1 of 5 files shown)


Mike-Heneghan:
[Add tracking of "Yes" and "No" events to `analytics_logs`](https://trello.com/c/eY8uDhXJ/2798-add-tracking-of-yes-and-no-events-to-analyticslogs)

DafyddLlyr (approved):
Great to get this across the line 🎉 

Also super super nice to have a discrete PR as a reference point for how to add analytics events in future 👏

PR #2993, opened 2024-04-05
feat: Replace `Select` with `Autocomplete` in `FileUploadAndLabel` component

DafyddLlyr:
## What does this PR do?
- Replaces the MUI Select component with a MUI Autocomplete component within the `FileUploadAndLabel` modal
- This is done in order to resolve the accessibility report issue "Customised Combobox" (Page 35)

## Why do we need to do this?
After spending a good while amending the markup of the `Select` I realised it wasn't possible to meet the requirements laid out the in report. These are the two fundamental barriers I hit -
 - MUI was hardcoding `role="option"` on the children of the Select - we require `role="group"`
 - A single `ul` element is forced as the first child element

Using the Autocomplete allows us access to grouping, which allows us to inject some custom HTML elements in the order and location we require. I've been aiming to match the following example of a grouped listbox shared in the report - https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/

### Potential issues
~~As this autocomplete in a HTML `Input` under the hood, users can natively type here. I've disabled this to match the current user experience however this may be problematic in itself.~~

Update: At the dev call on 10/04 we agreed to reinstate typing (see [5769135](https://github.com/theopensystemslab/planx-new/pull/2993/commits/5769135a4242c04647168f0e968301764fdcf13e)). This is seen as an additional helpful feature, and resolves the accessibility concerns raised above.

### Outstanding visual regressions
There are a few small visual disparities between the two versions which I've not yet tackled.
- [x] Styling for sticky group headers (do we even want them to be sticky?)
- [ ] Transition in
- [ ] Popper positioning
- [x] Focus styles
- [x] Element is larger (more padding) than previous iteration

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.test.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.test.tsx
@@ -1,5 +1,6 @@
 import {
   act,
+  findByRole,
   fireEvent,
   screen,
   waitFor,
@@ -342,7 +343,15 @@ describe("Modal trigger", () => {
 describe("Adding tags and syncing state", () => {
   test("Can continue when all required file types are uploaded and tagged", async () => {
     const handleSubmit = jest.fn();
-    const { user } = setup(
+    const {
+      getAllByRole,
+      getAllByTestId,
+      getByRole,
+      getByTestId,
+      getByText,
+      queryByRole,
+      user,
+    } = setup(
       <FileUploadAndLabelComponent
         title="Test title"
         handleSubmit={handleSubmit}
@@ -351,7 +360,7 @@ describe("Adding tags and syncing state", () => {
     );
 
     // No file requirements have been satisfied yet
-    let incompleteIcons = screen.getAllByTestId("incomplete-icon");
+    let incompleteIcons = getAllByTestId("incomplete-icon");
     expect(incompleteIcons).toHaveLength(2);
 
     // Upload one file
@@ -363,7 +372,7 @@ describe("Adding tags and syncing state", () => {
     });
 
     const file1 = new File(["test1"], "test1.png", { type: "image/png" });
-    const input = screen.getByTestId("upload-input");
+    const input = getByTestId("upload-input");
     await user.upload(input, [file1]);
 
     // Modal opened automatically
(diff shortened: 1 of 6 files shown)


editor.planx.uk/src/@planx/components/FileUploadAndLabel/SelectMultiple.tsx
```diff
@@ -40,194 +45,176 @@ const ListHeader = styled(Box)(({ theme }) => ({
...
-        What does this file show?
-      </StyledInputLabel>
-      <StyledSelect
-        native={false}
-        key={`select-${uploadedFile.id}`}
-        id={`select-multiple-file-tags-${uploadedFile.id}`}
-        variant="standard"
-        multiple
-        value={tags}
+      <Autocomplete
```
jessicamcinchak, line 144: I know this is still WIP - but just thought I'd flag prior MUI `<Autocomplete />` a11y PRs for prop reference!
- https://github.com/theopensystemslab/planx-new/pull/865 - this one highlights some obscure ones like `aria-atomic`, `handleHomeEndKeys` etc that it doesn't look like we're supporting here yet
 - https://github.com/theopensystemslab/planx-new/pull/887 - really detailed notes about expected behavior per device in this one (hopefully MUI component has improved enough in last ~2 years that we'll be able to pass using it this time!)

DafyddLlyr: This is super helpful thanks - will take a detailed look through.

DafyddLlyr:
[Customised combobox [A]](https://trello.com/c/2qftxyhY/2812-customised-combobox-a)

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.test.tsx
DafyddLlyr, line 1: Another nice bonus of this approach is that testing this component is much simpler! With the `Select` we had a fairly hard time targeting the dropdown but this is now much simpler.

editor.planx.uk/src/@planx/components/FileUploadAndLabel/SelectMultiple.tsx
```diff
@@ -1,233 +1,318 @@
...
   };
 
-  useEffect(() => {
-    updateFileListWithTags(previousTags, tags);
-  }, [tags]);
+  /**
+   * Custom Listbox component
+   * Used to wrap options within the autocomplete and append a custom element above the option list
+   */
+  const ListboxComponent = forwardRef<typeof Box, PropsWithChildren>(
```
DafyddLlyr, line 240: I spent a fair bit of time trying to take this component out of the scope of `SelectMultiple` and pass in the `closePopper()` function as a prop without much luck.

editor.planx.uk/src/@planx/components/FileUploadAndLabel/SelectMultiple.tsx
```diff
@@ -1,233 +1,318 @@
...
+    [initialTags],
+  );
+
+  /**
+   * Update FileList (from main Public.tsx component) when values are added or removed from Autocomplete
+   */
+  const handleChange = (
+    _event: React.SyntheticEvent,
+    value: Option[],
+    reason: AutocompleteChangeReason,
```
DafyddLlyr, line 204: Another nice DX bonus - the `Autocomplete` has a much nicer API for handling changes as it expects items to be added/removed/changed.

DafyddLlyr:
This Pizza has now been handed across to DAC for re-testing.

ianjon3s (approved):
I think we should merge this one in and pick up any subsequent feedback from DAC as separate PRs.

DafyddLlyr:
Thanks @ianjon3s ! Merged, and I'll open a prod deploy shortly

PR #2995, opened 2024-04-08
chore: add file api keys for Barnet's PowerBI project

jessicamcinchak:


jessicamcinchak:
[Automatically generate application forms/reports from ODP Schema JSON file using Microsoft SharePoint & PowerBI](https://trello.com/c/FtJ2svyh/2861-automatically-generate-application-forms-reports-from-odp-schema-json-file-using-microsoft-sharepoint-powerbi)

DafyddLlyr (approved):
In future it might be nice to consider how we could store these values in the DB, similar to how `team_integrations` are handled. Totally agree that this is the best solution for now though!

PR #2999, opened 2024-04-09
feat: add AWS S3 bucket as "Send" destination for staging envs only

jessicamcinchak:
Editor-side changes for the Barnet prototype described in linked Trello ticket. API changes to follow in a separate PR

Screenshot from 2024-04-09 14-03-32

### DIFF
--- a/editor.planx.uk/src/@planx/components/Send/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/Send/Editor.tsx
@@ -57,6 +57,14 @@ const SendComponent: React.FC<Props> = (props) => {
     },
   ];
 
+  // Show S3 option on staging only
+  if (process.env.REACT_APP_ENV !== "production") {
+    options.push({
+      value: Destination.S3,
+      label: "Upload to AWS S3 bucket",
+    });
+  }
+
   const changeCheckbox = (value: Destination) => (_checked: any) => {
     let newCheckedValues: Destination[];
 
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/Send/Editor.tsx
```diff
@@ -57,6 +57,14 @@ const SendComponent: React.FC<Props> = (props) => {
     },
   ];
 
+  // Show S3 option on staging only
+  if (process.env.REACT_APP_ENV !== "production") {
+    options.push({
+      value: Destination.S3,
+      label: "Upload to AWS S3 bucket",
+    });
+  }
```
jessicamcinchak, line 66: Rather than introduce a feature-flag here which we'll then have to communicate to Editors how to use - I figure we can just _only_ display this option on non-prod environments at first.

editor.planx.uk/src/@planx/components/Send/Editor.tsx
```diff
@@ -76,7 +84,8 @@ const SendComponent: React.FC<Props> = (props) => {
 
     // Show warnings on selection of BOPS or Uniform for likely unsupported services
     //   Don't actually restrict selection because flowSlug matching is imperfect for some valid test cases
-    const flowSlug = window.location.pathname?.split("/")?.[1];
+    const teamSlug = window.location.pathname?.split("/")?.[1];
+    const flowSlug = window.location.pathname?.split("/")?.[2];
```
jessicamcinchak, line 88: The `flowSlug` index was actually wrong here, but only impacted the in-editor warning messages - now fixed throughout!

jessicamcinchak:
[Automatically generate application forms/reports from ODP Schema JSON file using Microsoft SharePoint & PowerBI](https://trello.com/c/FtJ2svyh/2861-automatically-generate-application-forms-reports-from-odp-schema-json-file-using-microsoft-sharepoint-powerbi)

editor.planx.uk/src/@planx/components/Send/Editor.tsx
DafyddLlyr: Yeah this works well I think 👍

DafyddLlyr (approved):

PR #3000, opened 2024-04-09
chore: Tidy up hardcoded MUI classnames

DafyddLlyr:
Small tidy up / refactor I hit when working on https://github.com/theopensystemslab/planx-new/pull/2993

MUI offers classnames as an import so we can avoid hardcoding these 👍


ianjon3s (approved):
Looks good, I'll use these changes for ref when adding state classes to components.

PR #3009, opened 2024-04-11
chore(e2e): Update e2e tests to match GovPay journey

DafyddLlyr:
Most recent regression test are failing - https://github.com/theopensystemslab/planx-new/actions/runs/8642231150 ❌ 

Tests passing on this branch - https://github.com/theopensystemslab/planx-new/actions/runs/8643303652 ✅ 

It looks like this is due to GovPay changing their payment journeys - card details need to be re-entered.


jessicamcinchak (approved):

PR #3012, opened 2024-04-11
fix: show help text if defined for Content components

jessicamcinchak:
Bug report via August. You can define help text in the editor for Content nodes, but it was never displayed on the frontend - presumably because this component doesn't use the `QuestionHeader />` . 

This now displays the "More information" link below the rich text content block if defined like this: 
Screenshot from 2024-04-11 17-44-58

### DIFF
--- a/editor.planx.uk/src/@planx/components/Content/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/Content/Public.tsx
@@ -1,12 +1,20 @@
 import { mostReadable } from "@ctrl/tinycolor";
+import HelpIcon from "@mui/icons-material/Help";
 import Box from "@mui/material/Box";
 import { styled } from "@mui/material/styles";
+import Typography from "@mui/material/Typography";
 import type { Content } from "@planx/components/Content/model";
 import Card from "@planx/components/shared/Preview/Card";
 import { PublicProps } from "@planx/components/ui";
+import { useAnalyticsTracking } from "pages/FlowEditor/lib/analytics/provider";
 import React from "react";
 import { getContrastTextColor } from "styleUtils";
+import { emptyContent } from "ui/editor/RichTextInput";
 import ReactMarkdownOrHtml from "ui/shared/ReactMarkdownOrHtml";
+
+import MoreInfo from "../shared/Preview/MoreInfo";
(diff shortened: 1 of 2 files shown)


jessicamcinchak:
[Content component has no 'More information' button / does not display help text](https://trello.com/c/YFtvmSvx/2891-content-component-has-no-more-information-button-does-not-display-help-text)

editor.planx.uk/src/@planx/components/Content/Public.tsx
```diff
@@ -32,19 +40,81 @@ Content.defaultProps = {
...
           openLinksOnNewTab
           manuallyIncrementHeaders
         />
       </Content>
+      {!!(info || policyRef || howMeasured) && (
+        <Typography variant="subtitle1" component="div">
+          <HelpButton
+            variant="help"
+            title={`More information`}
+            aria-label={`See more information about this content`}
```
jessicamcinchak, line 79: This whole block is borrowed straight from our existing `<QuestionHeader />`.

Ideally this `aria-label` would smartly parse out the h1 from the rich text rather than say "this content" but I think this should be sufficient for now?

DafyddLlyr: Agreed - I think this is fine for now, and possibly fine full stop?

There's only one context for "this content", and not multiple options (like our `aria-label`s on change buttons for example".

DafyddLlyr (approved):

editor.planx.uk/src/@planx/components/Content/Public.tsx
ianjon3s: One for possible future discussion — should the content component also have a title, separated from the rich text input? It would mitigate the "every page must have a H1" issue we've seen highlighted elsewhere, although a review of the current usage of this component would be needed.

jessicamcinchak: We currently have a rich text helper method `incrementHeaderElements()` which automatically adjusts header heirarchy to help combat that "every page must have an H1" issue - it's not perfect, but does help catch a number of users-errors!

Agree that another possible way forward here would be to require more structured input for Content nodes - but then they basically become Notices? Let's keep an eye on it :+1:

PR #3013, opened 2024-04-12
fix: skip Send event creation if not a published route

jessicamcinchak:
Should cut down on these `cannot find session` errors on Send endpoints! 
Screenshot from 2024-04-12 10-45-33

**Testing:**
- On a `/draft` or `/amber` route, the Send component shows the following error message - you can click through to proceed to a Confirmation page but _no_ events will be created in Hasura
Screenshot from 2024-04-12 10-40-04
- On a `/published` route, the Send component queues up the applicable events and only shows a brief loading indicator just as before


DafyddLlyr (approved):
Super nice solution 👌

PR #3017, opened 2024-04-15
feat: basic platform admin panel for checking onboarding status & available integrations

jessicamcinchak:
**Why / current challenges / user stories:** 
- As more & more councils are onboarding or preparing to go live with services, we have a fair amount of back & forth in Slack about what's done versus still outstanding when it comes to their integrations. As a Planx product owner, I want to be able to check these details myself in one place and not rely on a dev to sign into the Hasura console/database to check ad-hoc. 
- For help-issues related to send to email, devs currently have to sign into the Hasura console/database to check if a submission email is configured. As anyone on the Planx team, I want to be able to see _if_ a submission email is configured for a certain environment, and which email it is so I can respond to the council person to help them troubleshoot.

**What I'm prototyping here:**
- Creates a database view `teams_summary` which collates info related to integrations, settings, and theme info for each team. It does _not_ expose any API keys, but it does expose the email addresses a council has configured for GOV Notify autoreplies and send to email. Currently it's only selectable by Platform Admins. 
- Adds a page "Admin Panel", which similar to Global Settings, is only visible to & accessible by Platform Admin users
- Displays a simple list of all Planx teams, with ability to expand to see their currently configured stuff (lots of room for visual improvement here I'm sure, but should do the job for now!) 

Screenshot from 2024-04-15 09-31-04

Screenshot from 2024-04-15 09-16-18

### DIFF
--- a/editor.planx.uk/src/pages/PlatformAdminPanel.tsx
+++ b/editor.planx.uk/src/pages/PlatformAdminPanel.tsx
@@ -16,6 +16,7 @@ import Caret from "ui/icons/Caret";
 const StyledTeamAccordion = styled(Accordion, {
   shouldForwardProp: (prop) => prop !== "primaryColour",
 })<{ primaryColour?: string }>(({ theme, primaryColour }) => ({
+  borderTop: "none", // TODO figure out how to remove top border (box shadow?) when collapsed
   borderLeft: `10px solid ${primaryColour}`,
   backgroundColor: theme.palette.background.paper,
   width: "100%",
@@ -36,7 +37,7 @@ function Component() {
       <Typography variant="h1">Platform Admin Panel</Typography>
       <Typography variant="body1" mb={3}>
         {`This is an overview of each team's integrations and settings for the `}
-        <strong>{process.env.NODE_ENV}</strong>
+        <strong>{process.env.REACT_APP_ENV}</strong>
         {` environment`}
       </Typography>
       {adminPanelData?.map((team) => <TeamData key={team.id} data={team} />)}
@@ -54,7 +55,7 @@ const Configured: React.FC = () => <Done color="success" fontSize="small" />;
 
 const TeamData: React.FC<TeamData> = ({ data }) => {
   return (
-    <StyledTeamAccordion primaryColour={data.primaryColour}>
+    <StyledTeamAccordion primaryColour={data.primaryColour} elevation={0}>
       <AccordionSummary
         id={`${data.name}-header`}
         aria-controls={`${data.name}-panel`}
--- a/editor.planx.uk/src/pages/PlatformAdminPanel.tsx
+++ b/editor.planx.uk/src/pages/PlatformAdminPanel.tsx
@@ -1,5 +1,8 @@
 import Close from "@mui/icons-material/Close";
 import Done from "@mui/icons-material/Done";
+import Accordion from "@mui/material/Accordion";
+import AccordionDetails from "@mui/material/AccordionDetails";
+import AccordionSummary from "@mui/material/AccordionSummary";
 import Box from "@mui/material/Box";
 import Grid from "@mui/material/Grid";
 import { styled } from "@mui/material/styles";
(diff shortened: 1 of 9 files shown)


editor.planx.uk/src/pages/PlatformAdminPanel.tsx
```diff
@@ -0,0 +1,167 @@
...
+                  {data.planningDataEnabled ? (
+                    <Configured />
+                  ) : (
+                    <NotConfigured />
+                  )}
+                </Box>
+              </>
+              <>
+                <Box component="dt">{"Article 4s"}</Box>
+                <Box component="dd">{"?"}</Box>
```
jessicamcinchak, line 117: Followup PR: figure out how to check if granular A4s are being returning by our Planx GIS API yet ! 

This is perhaps the trickiest one to currently measure (do we mean the spreadsheet, flow, API response etc) - but I've included this "?" placeholder as a reminder to manually check.

editor.planx.uk/src/routes/authenticated.tsx
```diff
@@ -68,6 +69,48 @@ const editorRoutes = compose(
       });
     }),
 
+    "/admin-panel": map(async (req) => {
+      const isAuthorised = useStore.getState().user?.isPlatformAdmin;
+      if (!isAuthorised)
+        throw new NotFoundError(
+          `User does not have access to ${req.originalUrl}`,
+        );
```
jessicamcinchak, line 77: Followup PR: the `/admin-panel` page is successfully reachable from the index, but throws this error if you hard-refresh form the page itself. This also appears to be a current bug with `/global-settings`, so not introduced here. Since these routes are only accessible by platform admins to begin with, there shouldn't be much harm in merging & pushing an overall fix as a followup.

hasura.planx.uk/migrations/1713084872473_create_view_teams_summary/up.sql
```diff
@@ -0,0 +1,26 @@
...
+    ti.has_planning_data as planning_data_enabled,
+    '@todo' as article_4s_enabled,    
+    t.notify_personalisation as govnotify_personalisation,
+    CASE 
+        WHEN coalesce(ti.production_govpay_secret, ti.staging_govpay_secret) is not null 
+        THEN true
+        ELSE false
+    END as govpay_enabled,
+    t.submission_email as send_to_email_address,
+    coalesce(ti.production_bops_submission_url, ti.staging_bops_submission_url) as bops_submission_url,
```
jessicamcinchak, line 17: Since we only sync `staging_` variables down to non-production environments, `coalesce()` is an easy way for making sure we display the right "done" per environment. 

Please note that if you want to check if a certain team has, for example, both a production & staging Gov Pay key configured - you'll have to check _both_ `editor.planx.uk/admin-panel` & `editor.planx.dev/admin-panel`. 

Having each environment ONLY reflect their own keys is the best way to ensure completeness in my opinion (eg because `teams.submission_email` for staging is actually only available on staging)

editor.planx.uk/src/pages/PlatformAdminPanel.tsx
```diff
@@ -0,0 +1,167 @@
...
+          </Grid>
+          <Grid item xs={4}>
+            <SummaryListTable dense={true}>
+              <>
+                <Box component="dt">{"GOV.UK Notify"}</Box>
+                <Box component="dd">
+                  {data.govnotifyPersonalisation?.helpEmail || (
+                    <NotConfigured />
+                  )}
+                </Box>
```
jessicamcinchak, line 139: Rather than display the whole JSONB personalisation here, I think the email should be enough of a cue if a council is still on the default `example@council.gov` or has updated it to their own? 

Updating still a manual database step & Gov Notify admin panel step, but a good candidate to move all personalisation fields related to this to team settings soon!

editor.planx.uk/src/pages/PlatformAdminPanel.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+function Component() {
+  const adminPanelData = useStore((state) => state.adminPanelData);
+
+  return (
+    <Box p={3}>
+      <Typography variant="h1">Platform Admin Panel</Typography>
+      <Typography variant="body1" mb={3}>
+        {`This is an overview of each team's integrations and settings for the `}
+        <strong>{process.env.NODE_ENV}</strong>
```
jessicamcinchak, line 39: Question: any ideas of a better/more accurate way to display the environment here? We have a nice `getFormattedEnvironment()` helper in the API, but I'm looking for an alternative using frontend/React env vars only?

Eg pizza is currently displaying "production"
Screenshot from 2024-04-15 10-07-43

editor.planx.uk/src/pages/PlatformAdminPanel.tsx
DafyddLlyr: Yep for sure - this is a good solution 👍

DafyddLlyr (approved):
This is fantastic 👍 

Great solution, and nice to be able to easily access and view this data.

One teeny tiny style nitpick - there's a border-top which only appears when the accordion is open.

editor.planx.uk/src/pages/PlatformAdminPanel.tsx
jessicamcinchak: Yep I was thinking of a similar endpoint variation - agree API is the thing we care about checking most here compared to flow content.

Will plan to pick this one back up when I pick up other constraints work in the coming weeks!

PR #3019, opened 2024-04-16
chore: Add `decrypt()` util to scripts

DafyddLlyr:
Seems sensible to have this in place! I've just been toggling `encypt()` with `decrypt()` when needed but it's a bit of a faff.

The order of the arguments has also been changed as this is easier for writing or reading multiple values (the key stays constant, and only the secret needs to change).


jessicamcinchak (approved):
Handy!

PR #3028, opened 2024-04-17
feat: Editor Menu

ianjon3s:
https://github.com/theopensystemslab/planx-new/assets/20502206/088c779c-cc9f-422e-bc7f-53ac11327c96

## What does this PR do?
- Adds `EditorMenu` (behind a feature flag)
- Handles routes, layouts and views
- Displays current flow settings components in new locations
  - Do we really want this, or just the "useable" components?

## Next steps
 - Handle "View edit history button"
 - Drop feature flag, drop "Flow Settings" from top right menu


editor.planx.uk/src/components/TestEnvironmentBanner.tsx
```diff
@@ -17,37 +17,33 @@ const TestEnvironmentWarning = styled(Box)(({ theme }) => ({
 }));
 
 const TestEnvironmentBanner: React.FC = () => {
-  const [showWarning, setShowWarning] = useState(true);
+  const [isTestEnvBannerVisible, hideTestEnvBanner] = useStore(state => [state.isTestEnvBannerVisible, state.hideTestEnvBanner]);
```
DafyddLlyr, line 20: This is now controlled at a store level as it was being re-rendered (and therefore going back to visible) on each route change.

I think this is the same issue we're seeing with everything re-rendering when we got to `/nodes` routes. See comment below on this 👍

editor.planx.uk/src/pages/FlowEditor/components/EditorMenu.tsx
```diff
@@ -0,0 +1,123 @@
...
+}));
+
+function EditorMenu() {
+  const { navigate } = useNavigation();
+  const { lastChunk } = useCurrentRoute();
+  const rootPath = rootFlowPath();
+
+  const isActive = (route: string) => lastChunk.url.pathname.endsWith(route);
+
+  const routes = [
```
DafyddLlyr, line 78: @ianjon3s This pattern of splitting up "content" and "code" is a good one to be aware of. It's now a bit simpler to add or remove a menu item, or control the visual appearance of all of them consistently 👍

editor.planx.uk/src/routes/flow.tsx
```diff
@@ -174,98 +176,83 @@ const nodeRoutes = mount({
...
+      }
     }),
 
-    "/nodes": nodeRoutes,
+    "/nodes": compose(
+      withView((req) => {
+        const [flow, ...breadcrumbs] = req.params.flow.split(",");
+        return (
+          <>
+            <FlowEditor key={flow} flow={flow} breadcrumbs={breadcrumbs} />
```
DafyddLlyr, line 210: This duplication of `FlowEditor` is worth being aware of. It matches the current setup where this is attached the view and duplicated that way (just a `<span/>` is rendered in the current setup.

It's the cause of our re-rendering when we go to `/node` routes - react-navi is re-rendering all other views, even those unaffected for some reason.

I repeated previous attempts with memoization here to attempt to solve this with the same results as before. 

The `Header` is also re-rendering when children change, so this may be a wider (or higher level) issue. Happy to carry on looking into it without holding up this PR.

editor.planx.uk/src/routes/flow.tsx
```diff
@@ -174,98 +176,83 @@ const nodeRoutes = mount({
...
+    "/data": compose(
+      withView(SettingsContainer),
+
+      route(async (req) => ({
+        title: makeTitle([req.params.team, req.params.flow, "data"].join("/")),
+        view: DataManagerSettings
+      })),
+    ),
+
+    "/submissions-log": compose(
```
DafyddLlyr, line 248: I've not hidden these routes behind a feature flag as there's nothing sensitive or different here, and they've not very discoverable.

editor.planx.uk/src/ui/images/CustomIcons.tsx
DafyddLlyr, line 1: nit: There's a folder for icons already. This should probably be `editor.planx.uk/src/ui/icons/Editor.tsx`

image

ianjon3s: Good catch, I've updated.

DafyddLlyr:
[Reshaping Editor interface (navigation menu)](https://trello.com/c/DWafqthl/2889-reshaping-editor-interface-navigation-menu)

jessicamcinchak (approved):
Functionally this works as expected for me! Thanks especially for tweaking some component heirarchies and starting to make a dent in the unexpected re-renders. 

Based on level of surprise/confusion this week to new editor styles, merging behind a feature flag definitely feels right for now. 

For me, the left menu icons are visually competing a bit with the icons/buttons at the top of the right sidebar now, not sure it feels intuitive yet where I find each type of thing (eg submissions log on left but analytics on right?). One to easily revisit in a followup PR & keep adjusting behind feature flag tho :slightly_smiling_face:

DafyddLlyr:
Thanks for the review! Agreed that these changes are probably best released as part of a wider refresh with a bit of communication around it 👍

PR #3037, opened 2024-04-18
fix: Ensure main components have full-width

DafyddLlyr:
Follow up to #3029  - noticed a few other pages with the same issue!


ianjon3s (approved):
Thanks for picking these up @DafyddLlyr , I'll keep an eye out for any more.

PR #3039, opened 2024-04-19
fix: Prevent editor sidebar from shrinking

ianjon3s:
# What does this PR do?

Quick one: prevent the editor sidebar/preview pane from shrinking if no content is present.

**Before:**
image

**After:**
image


DafyddLlyr (approved):
Thanks for spotting this! This could likely be fallout from my quick fix here - https://github.com/theopensystemslab/planx-new/pull/3025

There's a commit somewhere in https://github.com/theopensystemslab/planx-new/pull/3023 which broke this in the first place.

PR #3040, opened 2024-04-22
feat: add Microsoft variables to `team_integrations`

jessicamcinchak:
**Changes:** 
- Adds `*_file_api_key` variables per environment (will retire existing env var `FILE_API_KEY_BARNET` in a future PR)
- Adds `power_automate_webhook_url` 
  - Currently assuming that councils are going to have a _single_ environment for this
  - Still exploring authorisation options with Kev, so only storing/handling URL to begin with

**Follow-ups / future PRs:**
- [x] Update types in planx-core https://github.com/theopensystemslab/planx-core/pull/358
- [ ] Add staging columns to data sync script https://github.com/theopensystemslab/planx-new/pull/3042
- [ ] Update API
- [ ] Populate variables for Barnet and test

PR #3042, opened 2024-04-23
chore: add new staging `team_integrations` columns to data sync script

jessicamcinchak:
A bit of tricky sequencing with this one - Hasura should be healthy, but expect pizza db to be empty with following error because these columns aren't merged to production yet! 
Screenshot from 2024-04-23 08-56-01

Can be safely merged later after #3040 !

### DIFF
--- a/scripts/seed-database/container.sh
+++ b/scripts/seed-database/container.sh
@@ -35,7 +35,7 @@ done
 
 # Copy subset of team_integrations columns
 # Do not copy production values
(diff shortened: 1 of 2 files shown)


DafyddLlyr:
Very open to ideas here - it's currently pretty awkward to do handle this multi step process of adding db columns, and then correctly synching them down. I think your approach of just adding db columns in one PR, resolving quickly with queued up sync script PR, and _then_ actually using those columns is the best approach so far!

DafyddLlyr (approved):

PR #3047, opened 2024-04-23
chore: add Power Automate Webhook notification step to `/upload-submission` process 

jessicamcinchak:
**Changes:** 
- Bumps planx-core throughout to capture type changes to `TeamIntegrations` - now reads & decrypts both the Power Automate URL and API Key fields
- Sets up API endpoint which generates schema payload, uploads to S3, and now additionally sends an authorised notification to a Power Automate webhook

**Testing:**
- From this pizza, we can test that the send event is successfully created & executed - but these tests will fail on Kev's side because the user uploaded files can't be downloaded (eg pizzas upload to a different bucket)
- Kev's files API keys only work against staging & production `user-data-` buckets, so this needs to merge to `main` to do a true "end to end test" of the whole Planx submission &rarr; Power Automate flow

**Future PRs (so we can unblock initial testing on `main` in the meantime!):**
- Audit table (not necessary for very early testing yet)
- Finalise sanitation strategy (see comment below)


api.planx.uk/modules/send/s3/index.ts
```diff
@@ -38,16 +46,35 @@ export async function sendToS3(
     const { fileUrl } = await uploadPrivateFile(
       exportData,
       `${sessionId}.json`,
-      "barnet-prototype",
```
jessicamcinchak, line 41: I've dropped the static folder name here which means we'll default to using our standard `nanoid()`-generated random folder name. I think this ultimately has a couple benefits:
- A bit more secure/un-guessable 
- Existing S3 sanitation job will "just work" for submission JSON files the same as user-uploaded ones

DafyddLlyr: Just a heads up here (I know the PR is still draft sorry!) - I think currently the file will be "orphaned" in S3 which we don't want. The S3 sanitation operations loops over expired sessions, and then gets all associated file URLs from the session data by scanning the passport and this file won't be listed there.

One (hopefully fairly simple) solution which still allows us to use a nanoid would be to keep the generated URL in the db somewhere - presumably the submission audit table for this send destination - and then we can match this to the sessionId currently being sanitised and delete it.

A few options here other than the above - happy to talk through if easier 👍

jessicamcinchak: Thanks for clarifying this! An audit table would indeed be an easy solution to this case, but understanding this highlights another loophole for me: files uploaded on any environment from `/draft` or `/preview` testing links are also then being currently orphaned/omitted from sanitation because they won't have an associated `lowcal_session` record, right? 

I think we'll need to make a sanitation change sooner than later that looks at S3 _date_ rather than our own session to fix all edge cases? Being careful to omit Planx service image uploads, but those are already in a separate bucket!

editor.planx.uk/src/@planx/components/Send/Editor.tsx
```diff
@@ -90,15 +90,7 @@ const SendComponent: React.FC<Props> = (props) => {
...
       const flowSlug = window.location.pathname?.split("/")?.[2];
-      if (
-        value === Destination.BOPS &&
-        newCheckedValues.includes(value) &&
-        ![
-          "apply-for-a-lawful-development-certificate",
-          "apply-for-prior-approval",
-          "apply-for-planning-permission",
-        ].includes(flowSlug)
-      ) {
```
jessicamcinchak, line 101: Small unrelated change here: based on recent BOPS demos at Show & Tells, they can accept any ODP Schema application type & officers can self-setup in BOPS to receive it - so we don't need to warn about this !

jessicamcinchak:
[Automatically generate application forms/reports from ODP Schema JSON file using AWS S3, MS Power Automate & PowerBI](https://trello.com/c/FtJ2svyh/2861-automatically-generate-application-forms-reports-from-odp-schema-json-file-using-aws-s3-ms-power-automate-powerbi)

api.planx.uk/modules/send/s3/index.ts
DafyddLlyr: Yep, I think a "catch-all" sweep based on age is a good solution (I think it's actually possible to make files "expire" by date when uploading them).

Currently both public and private files are in the same bucket which makes this harder (there are some legacy public files in another bucket).

A good solution may be - 
 - public files, in their own bucket, separated by flow
   - easy cleanup when a service is deleted, easy to identify file associated with a flow
   
 - private files, in their own bucket, grouped by session (even if there's no `lowcal_session` record there's a session generated in state)
   - These could then all expire based on submission date + retention period, or after a set time interval.

editor.planx.uk/src/@planx/components/Send/Editor.tsx
DafyddLlyr: Nice to drop some of these hardcoded service names 👍

editor.planx.uk/src/@planx/components/Send/Editor.tsx
```diff
@@ -118,10 +110,10 @@ const SendComponent: React.FC<Props> = (props) => {
       if (
         value === Destination.S3 &&
         newCheckedValues.includes(value) &&
-        teamSlug !== "barnet"
+        !["barnet", "lambeth"].includes(teamSlug)
       ) {
         alert(
-          "AWS S3 uploads are currently being prototyped with Barnet only. Please do not select this option for other councils yet.",
+          "AWS S3 uploads are currently being prototyped with Barnet and Lambeth only. Please do not select this option for other councils yet.",
```
DafyddLlyr, line 116: 👍

DafyddLlyr (approved):
Looks great!

api.planx.uk/modules/send/s3/index.ts
jessicamcinchak: Sounds like a plan - I'll add a Trello ticket for this to new requests so we pick it up next sprint / this phase :+1:

PR #3049, opened 2024-04-23
Production deploy

jessicamcinchak:
Gloucester subdomain appears to be up so would like to get final pieces across the line. 

Any objections to merging Editor "tabs" along with it? Or shall I push a quick edit to feature flag behind `UNDO` first?


DafyddLlyr (commented):
Maybe we can just replace the string "History" which is displays with the `<FeaturePlaceholder title="Feature in development" />` component used elsewhere? (Or hide behind flag!)

jessicamcinchak:
@DafyddLlyr good shout - see #3050 !

PR #3052, opened 2024-04-24
feat: display formatted edit history of a flow

jessicamcinchak:
This remains behind the "UNDO" feature flag until full functionality in place, but getting this onto `main` sooner than later should make it easier for Ian to continue working in parallel on styling while I hook up API functionality to icon button. 

**Changes:**
- Populates side panel with formatted change history of last 15 edits
- If using the "UNDO" feature flag, sets default tab to `History` rather than `PreviewBrowser` - otherwise our known rendering bug makes checking/watching consecutive edits very annoying

Run `window.featureFlags.toggle("UNDO")` for tesitng please ! 

Screenshot from 2024-04-24 13-43-07

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/index.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/index.tsx
@@ -2,12 +2,13 @@ import "./components/Settings";
 import "./floweditor.scss";
 
 import { gql, useSubscription } from "@apollo/client";
+import UndoOutlined from "@mui/icons-material/UndoOutlined";
 import Box from "@mui/material/Box";
 import Typography from "@mui/material/Typography";
 import { formatOps } from "@planx/graph";
 import { OT } from "@planx/graph/types";
-import { format } from "date-fns";
-import { hasFeatureFlag } from "lib/featureFlags";
+import DelayedLoadingIndicator from "components/DelayedLoadingIndicator";
+import { formatDistanceToNow } from "date-fns";
 import React, { useRef } from "react";
 
 import { rootFlowPath } from "../../routes/utils";
@@ -17,6 +18,7 @@ import { useStore } from "./lib/store";
 import useScrollControlsAndRememberPosition from "./lib/useScrollControlsAndRememberPosition";
 
 interface Operation {
+  id: string;
   createdAt: string;
   actor?: {
     firstName: string;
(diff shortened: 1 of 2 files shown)


jessicamcinchak:
E2E CI is failing on a pnpm install that is not specific to these changes. I think we're going to see this one pop up widely & will need to address separately 
Screenshot from 2024-04-24 13-49-14

editor.planx.uk/src/pages/FlowEditor/index.tsx
```diff
@@ -99,14 +104,88 @@ export const LastEdited = () => {
...
-            <Typography variant="body2" fontSize="small" component="li" key={i}>
-              {op}
-            </Typography>
-          ))}
-        </Typography>
+    </Box>
+  );
+};
+
+export const EditHistory = () => {
```
DafyddLlyr, line 111: This should probably move to its own file now 👍

editor.planx.uk/src/pages/FlowEditor/index.tsx
```diff
@@ -25,29 +27,42 @@ interface Operation {
   data: Array<OT.Op>;
 }
 
-export const LastEdited = () => {
-  const [flowId, flow] = useStore((state) => [state.id, state.flow]);
+const formatLastEditDate = (date: string): string => {
```
DafyddLlyr, line 30: I wondered about using `formatDistanceToNow()` for the initial version (as it's what's used in the last published text).

The reason I decided against it was that it gets stale - it's correct each time it's rendered, but if there's no operations for a while and time passes, it's incorrect.

We could certainly update them by calling `refetch()` on a timeout (30 seconds?). I think also just re-rendering here without a refetch would achieve the same - I'd guess the Apollo cache may make both of these equal but would need to try it.

I'm not sure what the right solution / pattern is here - maybe a good question for @ianjon3s!

Not a blocking issue, but it is one we'll need to address sooner or later I think.

editor.planx.uk/src/pages/FlowEditor/index.tsx
```diff
@@ -99,14 +104,88 @@ export const LastEdited = () => {
...
-          ))}
-        </Typography>
+    </Box>
+  );
+};
+
+export const EditHistory = () => {
+  const [flowId, flow] = useStore((state) => [state.id, state.flow]);
+
+  const { data, loading, error } = useSubscription<{ operations: Operation[] }>(
```
DafyddLlyr, line 114: See above, `useSubscription()` also returns a `refetch()` function (if we want to use it!)

DafyddLlyr (approved):
Looking great! No blocking issues, but it's probably worth moving `EditHistory` to it's own file at this stage I think 👍

editor.planx.uk/src/pages/FlowEditor/index.tsx
jessicamcinchak: Yes I agree - I've tried this twice - creating an `EditHistory.tsx` at the same level of `PreviewBrowser.tsx` but consistently getting an opaque loading error on the whole page when trying to import the component from there. Maybe our render bug again?

Don't want to lose more time to that right now before actual functionality is in place, this sholud be easy enough to come back to & tidy.

jessicamcinchak:
[Add an 'undo' function in editor](https://trello.com/c/nmvj7Ca9/2442-add-an-undo-function-in-editor)

editor.planx.uk/src/pages/FlowEditor/index.tsx
DafyddLlyr: Rough implementation of this here - https://github.com/theopensystemslab/planx-new/pull/3059

Likely an issue which can be resolved by UI alone 👍

PR #3068, opened 2024-04-29
fix: Autocomplete reset/re-render on keyboard events

DafyddLlyr:
## What does this PR do?
 - Stops drop down list "resetting" to the top when an option is selected via keyboard

## Context
Please see https://github.com/theopensystemslab/planx-new/pull/2993#discussion_r1559815505

Originally I had to nest `ListboxComponent ` this within `SelectMultiple` in order to access functions related to the state of the popper. 

What I _think_ is happening is that selecting an item with they keyboard was causing a re-render of the popper and effectively resetting the view to the top of the list.

I've actually removed these here as currently clicking "Done" counts as a clickaway as it's outside of the scope of the popper (`ref` and `props` are applied to the `role="listbox"` child element within `ListboxComponent`). This actually in effect matches the current behaviour -
 - there's no custom action being triggered by "Done" (just a close)
 - it's already not keyboard navigable as a button

### Before
After selecting an item, the focus shifts back to the first item
https://github.com/theopensystemslab/planx-new/assets/20502206/9f960cbf-6ff0-4007-a0fb-8597fe68f8ef

### After
After selecting an item, the focus remains in place
https://github.com/theopensystemslab/planx-new/assets/20502206/6e890a66-cf08-4b43-96c4-b6bc0e78a03c

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/SelectMultiple.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/SelectMultiple.tsx
@@ -19,6 +19,7 @@ import Typography from "@mui/material/Typography";
 import capitalize from "lodash/capitalize";
 import React, { forwardRef, PropsWithChildren, useMemo, useState } from "react";
 import { borderedFocusStyle } from "theme";
+import Checkbox from "ui/shared/Checkbox";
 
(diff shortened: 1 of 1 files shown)


ianjon3s (approved):
This is working perfectly as far as I can see

PR #2976, opened 2024-04-03
fix(a11y): Input grouping, fieldset and legend

ianjon3s:
# What does this PR do?

- Adds `<fieldset>` and `<legend>` elements for input groupings
- Applies to Checklist and Question components, and Yes/No feedback buttons in help text feedback.

## Context

Identified in the accessibility report pages 21–26:
`Checkboxes relating to one another have not been grouped programmatically.`
`Ensure that the legend is the first element instantiated within the fieldset to make sure that
no sporadic results occur for users of screen reading assistive technologies.`
`Buttons relating to a question have not been grouped within a fieldset.`

## Demo

https://2976.planx.pizza/testing/input-grouping/published

PR #3072, opened 2024-04-30
fix: `<fieldset>` element overflows

DafyddLlyr:
I think this is the route of the problem causing the recent layout quirks we've been encountering.

Regression originally introduced here - https://github.com/theopensystemslab/planx-new/pull/2976

It turns out `<fieldset>` elements have [`min-inline-size: min-content`](https://developer.mozilla.org/en-US/docs/Web/CSS/min-inline-size) applied by default ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#styling_with_css)). This means that child content was overflowing and affecting layouts in a number of ways.

**Broken layout on staging**
https://editor.planx.dev/opensystemslab/listed-building-consent-files

https://github.com/theopensystemslab/planx-new/assets/20502206/d49727a9-3c9c-443e-ab41-83729c015f40

**Fixed layout on pizza**
https://3072.editor.planx.pizza/opensystemslab/listed-building-consent-files

### DIFF
--- a/editor.planx.uk/src/@planx/components/Checklist/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/Checklist/Public.tsx
@@ -146,7 +146,6 @@ const ChecklistComponent: React.FC<Props> = ({
             container
             spacing={layout === ChecklistLayout.Images ? 2 : 0}
             component="fieldset"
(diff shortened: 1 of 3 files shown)


jessicamcinchak (approved):

PR #3073, opened 2024-04-30
fix: Replace remaining `/amber` references

DafyddLlyr:
Follow up to https://github.com/theopensystemslab/planx-new/pull/3069, resolves issues outlined here https://opensystemslab.slack.com/archives/C5Q59R3HB/p1714474156311369 (OSL Slack)


jessicamcinchak (approved):
Thanks for catching this one!!

PR #3098, opened 2024-05-02
chore: Bump `planx-core`

DafyddLlyr:


jessicamcinchak (approved):
This pizza working as expected - can do everything that was broken yesterday ! Seems to pinpoint that our bug was cache related, not this version bump :white_check_mark:

PR #3099, opened 2024-05-02
fix(wip): Disable scrollable backdrop

DafyddLlyr:


DafyddLlyr:
Manual testing on iOS showed that this approach does not work - the background is still scrollable.

PR #3103, opened 2024-05-03
chore: Regenerate lockfiles

DafyddLlyr:


scripts/encrypt/package.json
```diff
@@ -9,7 +9,7 @@
   },
   "keywords": [],
   "dependencies": {
-    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#296592c",
+    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#722e1c7",
```
DafyddLlyr, line 12: Seems sensible to bump this at the same time!

jessicamcinchak: Honestly might not be a bad idea to regenerate lockfiles in planx-core too at some point!

jessicamcinchak (approved):

PR #3107, opened 2024-05-03
fix: Conditionally set test env header

DafyddLlyr:
Regression introduced here - https://github.com/theopensystemslab/planx-new/pull/3028

Variable hardcoded to `true` for testing and never spotted sorry 🤦‍♂️ 

Original conditional logic now restored, see here for reference - https://github.com/theopensystemslab/planx-new/pull/3028/files#diff-9bc006127599ede1eeba7b86caa661cc41ccd88a25bbccb5fbce32f9aad0ba49L22


jessicamcinchak (approved):

jessicamcinchak (approved):
Thanks for such quick fix, and apologies for not spotting in original review !

DafyddLlyr:
> Thanks for such quick fix, and apologies for not spotting in original review !

Honestly zero apologies needed - combination of a feature flagged PR and bigger fish to fry today!

PR #3109, opened 2024-05-03
feat: Setup query and placeholder components for "Team members" page

DafyddLlyr:
## What does this PR do?
 - Makes new page `/:team/members`
 - Sets up query to get team members, and to group by role
 - No styling, no feature flags!
 - Ready for handover to @ianjon3s!

Would be good to have a handover here to talk about how routes, queries, and components are lining up here 👍 

## Questions
 - "Inactive" users have an email address of `null` - should they be grouped by this, or are they fine within their previous roles?
 - We can add sorting to the query if we want to, that should then carry over into the `groupBy()`

### DIFF
--- a/editor.planx.uk/src/components/Header.tsx
+++ b/editor.planx.uk/src/components/Header.tsx
@@ -539,9 +539,16 @@ const EditorToolbar: React.FC<{
 
             {/* Only show team settings link if inside a team route  */}
             {isTeamSettingsVisible && (
-              <MenuItem onClick={() => navigate(`${rootTeamPath()}/settings`)}>
-                Team Settings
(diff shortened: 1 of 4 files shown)


jessicamcinchak (approved):
This is great ! I think users with `null` email make sense to group by their previous roles, then would simply be greyed out or similar in the UI to signal "inactive". Sorting by first name within each role at the query role feels like an easy win :slightly_smiling_face:

DafyddLlyr:
> then would simply be greyed out or similar in the UI to signal "inactive".

My thoughts exactly!

Added sorting by first name and hidden (commented out) menu item until UI built so this can merge 👍

PR #3113, opened 2024-05-07
wip: store every selected value in passport, regardless of granularity

jessicamcinchak:
See thread https://opensystemslab.slack.com/archives/C5Q59R3HB/p1715070535334409

Previously, if you have a checklist with options: `new`, `new.agriculture`, `new.agriculture.glasshouse` and you select options 2 & 3 - your passport would only store 3 `new.agriculture.glasshouse` because it is the _most granular_ and any less granular future automations for `new` or `new.agriculture` could be handled based on this value alone.

On this pizza, we'll store _every_ selection you make regardless of relative granularity or duplication with pre-existing passport values. Just spinning this up as a demo to wrap our minds around scope of "problem" / possible routes to solving (code or schema heirarchy changes)

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
@@ -222,17 +222,16 @@ export const previewStore: StateCreator<
           if (passportValue.length > 0) {
             const existingValue = acc.data?.[key] ?? [];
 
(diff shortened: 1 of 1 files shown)


jessicamcinchak:
Closing this one - linked here in case we want to revisit with new logic: https://trello.com/c/dboRkua3/3112-automation-filtering-logic-followups

PR #3117, opened 2024-05-07
feat: build pizzas on alpine

freemvmt:
As per title - use Alpine OS instead of Ubuntu 22.04 as the base OS for our Vultr pizzas. This seems to provide some substantial time savings, according to some tests I ran (see below). In particular, the work of updating the server, ensuring docker and aws-cli are installed, and actually doing the `docker compose` command to build the container stack, all of which are encapsulated in the `Create commands` step, is reduced from an average of 10:42 to 8:05, a saving of `157s`!

I didn't compare the `Update commands` step across the different operating systems, but there may also be gains there (I think unlikely that there would be losses).

Note that Vultr now provides an OS for the new Ubuntu LTS release (24.04, Noble Numbat). This is only just released as of this PR. Once it's more mature (e.g. at least 24.04.1 is released), it may in turn provide benefits compared to Alpine (or not). We can test that theory at a later date.

## Alpine

image

## Ubuntu 22.04

image

### DIFF
--- a/scripts/pullrequest/create.sh
+++ b/scripts/pullrequest/create.sh
@@ -10,15 +10,6 @@ echo "root:$SSH_PASSWORD" | chpasswd
 # https://www.digitalocean.com/community/tutorials/how-to-add-swap-space-on-ubuntu-22-04
 swapon --show
 
-# install docker
-apt-get install apt-transport-https ca-certificates curl gnupg lsb-release -y
-curl -fsSL https://download.docker.com/linux/ubuntu/gpg | gpg --batch --yes --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
-echo \
-  "deb [arch=amd64 signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu \
-  $(lsb_release -cs) stable" | tee /etc/apt/sources.list.d/docker.list > /dev/null
-apt-get update -y
-apt-get install docker-ce docker-ce-cli containerd.io docker-compose-plugin -y
-
 # set env for this shell
 set -o allexport
 source .env.pizza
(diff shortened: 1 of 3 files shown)


scripts/pullrequest/create.sh
```diff
@@ -10,23 +10,14 @@ echo "root:$SSH_PASSWORD" | chpasswd
...
-
 # set env for this shell
 set -o allexport
 source .env.pizza
 DOCKER_BUILDKIT=1
 set +o allexport
 
 # start services
-docker compose \
+time docker compose \
```
jessicamcinchak, line 20: :eyes: :eyes:  ~3 minutes faster ?! looking promising :crossed_fingers: :checkered_flag:

Screenshot from 2024-05-08 10-18-20

scripts/pullrequest/create.sh
```diff
@@ -10,15 +10,6 @@ echo "root:$SSH_PASSWORD" | chpasswd
...
 
-# install docker
-apt-get install apt-transport-https ca-certificates curl gnupg lsb-release -y
-curl -fsSL https://download.docker.com/linux/ubuntu/gpg | gpg --batch --yes --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
-echo \
-  "deb [arch=amd64 signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu \
-  $(lsb_release -cs) stable" | tee /etc/apt/sources.list.d/docker.list > /dev/null
-apt-get update -y
-apt-get install docker-ce docker-ce-cli containerd.io docker-compose-plugin -y
-
```
jessicamcinchak, line 21: satisfying :scissors:

jessicamcinchak (approved):

PR #3022, opened 2024-04-16
fix(a11y): continue button should always be enabled on FindProperty & DrawBoundary

jessicamcinchak:
WIP

Ideas / questions / todos:
- Error handling for postcode input & address autocomplete
  - Can we conditionally render address autocomplete still or should we _always_ show a disabled select until valid postcode?
- How to error handle supplemental data fetch that FindPropery does? Wrap continue button in error if loading spinner still visible?
- ErrorWrapper for maps (Plot new address & draw boundary)
- ErrorWrapper for location plan upload (re-use FileUpload validation schema?)

PR #3131, opened 2024-05-09
fix(a11y): DrawBoundary link to switch between map and upload should always be enabled

jessicamcinchak:
**Previously:**
- We disabled the "Upload a location plan instead" and "Draw the boundary on a map instead" links to switch pages if the data on the respective page was fulfilled - meaning the user had to erase the boundary or remove their upload to switch pages back

**Now:**
- The links to switch pages are always enabled regardless of data
- Data is retained when switching pages
- `handleSubmit()` only submits the data from the page which the use "Continues" onwards from - ensuring the application payload has GeoJSON _or_ an uploaded location plan, never both

Screenshot from 2024-05-09 12-44-51

**Note:**
- Enabling the link & updating the corresponding `visuallyHidden` text should fix the A-level issue described on page 17 (enabling by default also resolves a AA contrast issue)
- The overall "Continue" button on this component is still conditionally enabled (similar to FindProperty). As I understand, this is a usability-level issue only and we can address separately in a follow-up PR (see work started in #3022)

### DIFF
--- a/editor.planx.uk/src/@planx/components/DrawBoundary/Public/Public.test.tsx
+++ b/editor.planx.uk/src/@planx/components/DrawBoundary/Public/Public.test.tsx
@@ -370,3 +370,86 @@ test("appends to existing '_requestedFiles' value", async () => {
   expect(recommended).toEqual(["elevations.existing"]);
   expect(optional).toHaveLength(0);
 });
+
+test("submits data based on the page you continue onwards from", async () => {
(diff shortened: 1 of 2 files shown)


jessicamcinchak:
[Sensory characteristics [A]](https://trello.com/c/sJmebZD4/2804-sensory-characteristics-a)

ianjon3s (approved):
Works exactly as expected.

PR #3133, opened 2024-05-09
fix(a11y): Feedback link as <a> element

ianjon3s:
# What does this PR do?

Fixes issue highlighted by DAC report p.88 (DAC_Buttons_Styled_As_Links_Usability_01) in that the feedback link is a button with the appearance of a link.

To fix this I've added a hash link (href="#") and used `preventDefault` to remove link functionality. This works similar to how GOV.UK use it here: https://design-system.service.gov.uk/components/phase-banner/

Also, the layout has been tidied up for smaller screens.


ianjon3s:
[Buttons styled as links [Usability]](https://trello.com/c/8zT76ZBT/2833-buttons-styled-as-links-usability)

PR #3135, opened 2024-05-10
chore: set & track `findProperty.action` in analytics allow lists

jessicamcinchak:
- Adds column `find_property_action` to both `analytics_summary` & `submission_services_summary` views
  - Will be populated with one of two values: "Selected an existing address" or "Proposed a new address" 
- Will help us understand how users are proceeding through the service and evidence roadmap requests related to improving search capability & map interactions around proposing new addresses
Screenshot from 2024-05-10 13-39-26

### DIFF
--- a/editor.planx.uk/src/@planx/components/FindProperty/model.ts
+++ b/editor.planx.uk/src/@planx/components/FindProperty/model.ts
@@ -1,5 +1,12 @@
 import { MoreInformation, parseMoreInformation } from "../shared";
 
+export enum FindPropertyUserAction {
+  Existing = "Selected an existing address",
+  New = "Proposed a new address",
+}
+
+export const PASSPORT_COMPONENT_ACTION_KEY = "findProperty.action";
+
 export interface FindProperty extends MoreInformation {
   title: string;
   description: string;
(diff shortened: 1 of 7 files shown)


jessicamcinchak:
[Track how users proceed through FindProperty - by selecting existing address or plotting a new one](https://trello.com/c/fnz1Fbkb/2897-track-how-users-proceed-through-findproperty-by-selecting-existing-address-or-plotting-a-new-one)

editor.planx.uk/src/@planx/components/FindProperty/model.ts
```diff
@@ -1,5 +1,12 @@
 import { MoreInformation, parseMoreInformation } from "../shared";
 
+export enum FindPropertyUserAction {
+  Existing = "Selected an existing address",
+  New = "Proposed a new address",
+}
+
+export const PASSPORT_COMPONENT_ACTION_KEY = "findProperty.action";
```
jessicamcinchak, line 8: Creating a new passport variable, rather than using `_address.source` has two main benefits to keep this quick & simple:
- The values can be human readable in the source, rather than just "os" & "proposed"
- We don't have to refactor allow list handling to work with deconstructed, nested passport variables like `_address`

DafyddLlyr (approved):
Perfect - nice and simple 👍

PR #3136, opened 2024-05-10
feat: Inline loading indicator for address lookup

ianjon3s:
## What does this PR do?

To both build on the work in progress here: https://github.com/theopensystemslab/planx-new/pull/3022,
and also as an improvement to the current setup — updates the progress indicator to be inline with the continue button.

**Changes:**
- Progress indicator given a wrapper that is omnipresent (absolute positioning on desktop), eliminates layout shift on both desktop and small screens
- `inline` prop added to `DelayedLoadingIndicator`, making it easier to use either as a full screen takeover or as part of page content
- Text updated to "Fetching address data" to be more descriptive to users

**Preview:**
image

**Test:**
https://3136.planx.pizza/testing/address-lookup/published

### DIFF
--- a/editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
@@ -1,6 +1,7 @@
 import { gql } from "@apollo/client";
 import Box from "@mui/material/Box";
 import Link from "@mui/material/Link";
+import { styled } from "@mui/material/styles";
 import Typography from "@mui/material/Typography";
 import Card from "@planx/components/shared/Preview/Card";
 import QuestionHeader from "@planx/components/shared/Preview/QuestionHeader";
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -33,6 +34,18 @@ export const FETCH_BLPU_CODES = gql`
...
+  marginBottom: theme.spacing(-2.5),
+  minHeight: theme.spacing(3),
+  pointerEvents: "none",
+  [theme.breakpoints.up("md")]: {
+    position: "absolute",
+    bottom: theme.spacing(1),
+    margin: 0,
+    paddingRight: theme.spacing(14),
+  },
+}));
```
jessicamcinchak, line 47: nit: this works perfectly on full screen routes, but if I use FindProperty from the editor side panel the loading indicator is overlapping the disabled Continue button

jessicamcinchak (approved):
Nice! Definitely less disruptive & jumpy :+1:

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
ianjon3s: Good catch, I've updated to be aligned left and padded to leave space for the continue button, a more universal solution.

jessicamcinchak: Working perfectly, let's merge :rocket:

PR #3137, opened 2024-05-10
fix(a11y): always enable continue button on FileUpload and DrawBoundary upload page

jessicamcinchak:
We currently have three components that don't enable the "Continue" button by default which is flagged as a bad pattern in the "usability" feedback section of our Accessbility report: 
1. FindProperty (two pages - select & map)
2. DrawBoundary (two pages - map & upload)
3. FileUpload

These changes address enabling the continue button for all "upload" instances. I'll continue to address the map & select instances in a followup PR (see #3022).

**Testing:**
- DrawBoundary 
  - When you switch to the upload page the "continue" button should always be enabled now
  - If you click it before uploading, you should see an error wrapper around the upload preventing you from proceeding
  - If you upload a file, you can proceed successfully
- FileUpload
  -  "Continue" should be enabled by defualt
  - If you click before uploading, or while a large file is still loading, you should see an error wrapper preventing you from proceeding
  - If you upload a file, you can proceed successfully

### DIFF
--- a/editor.planx.uk/src/@planx/components/DrawBoundary/Public/Public.test.tsx
+++ b/editor.planx.uk/src/@planx/components/DrawBoundary/Public/Public.test.tsx
@@ -139,7 +139,13 @@ test("shows the file upload option by default and requires user data to continue
   // Navigate to upload a file screen
   await user.click(screen.getByTestId("upload-file-button"));
   expect(screen.getByText("Upload a file")).toBeInTheDocument();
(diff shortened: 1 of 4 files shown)


jessicamcinchak:
[Disabled buttons and unexpected input functionality [Usability]](https://trello.com/c/elmTB1Pq/2828-disabled-buttons-and-unexpected-input-functionality-usability)

ianjon3s (approved):
Both upload instances working great, with appropriate error messaging in place 👍

PR #3141, opened 2024-05-13
fix(a11y): always enable continue button on FindProperty map page

jessicamcinchak:
Breaking these change up by "page" as this one is quite tricky! 

**Changes:**
- Continue is always enabled on FindProperty's "my site does not have an address page" 
- Previously only the input had an error wrapper, now the map does as well
- Both errors can be triggered via `validateAndSubmit` when clicking continue

Screenshot from 2024-05-13 15-48-56

### DIFF
--- a/editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
+++ b/editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
@@ -11,10 +11,10 @@ import {
   MapContainer,
   MapFooter,
 } from "@planx/components/shared/Preview/MapContainer";
-import { GeoJSONObject } from "@turf/helpers";
 import { useStore } from "pages/FlowEditor/lib/store";
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import InputLabel from "ui/public/InputLabel";
+import ErrorWrapper from "ui/shared/ErrorWrapper";
 import Input from "ui/shared/Input";
 
 import { DEFAULT_NEW_ADDRESS_LABEL, SiteAddress } from "../model";
@@ -23,10 +23,15 @@ interface PlotNewAddressProps {
   setAddress: React.Dispatch<React.SetStateAction<SiteAddress | undefined>>;
   setPage: React.Dispatch<React.SetStateAction<"os-address" | "new-address">>;
   initialProposedAddress?: SiteAddress;
-  boundary?: GeoJSONObject | undefined;
   id?: string;
   description?: string;
   descriptionLabel?: string;
+  mapValidationError?: string;
+  setMapValidationError: React.Dispatch<
+    React.SetStateAction<string | undefined>
+  >;
+  showSiteDescriptionError: boolean;
+  setShowSiteDescriptionError: React.Dispatch<React.SetStateAction<boolean>>;
 }
 
 type Coordinates = {
@@ -55,8 +60,6 @@ export default function PlotNewAddress(props: PlotNewAddressProps): FCReturn {
   const [siteDescription, setSiteDescription] = useState<string | null>(
     props.initialProposedAddress?.title ?? null,
   );
-  const [showSiteDescriptionError, setShowSiteDescriptionError] =
-    useState<boolean>(false);
 
   const [environment, boundaryBBox] = useStore((state) => [
     state.previewEnvironment,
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -89,21 +90,6 @@ function Component(props: Props) {
...
-      props.allowNewAddresses && teamSettings?.boundary
-        ? teamSettings.boundary
-        : null,
-    fetcher,
-    {
-      shouldRetryOnError: true,
-      errorRetryInterval: 500,
-      errorRetryCount: 1,
-    },
-  );
```
jessicamcinchak, line 105: We don't actually need to make this fetch, we're getting the `boundaryBBox` directly from the Store in the map in the child component !

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -124,11 +110,62 @@ function Component(props: Props) {
     }
   }, [data, address]);
 
-  useEffect(() => {
-    if (geojson) setBoundary(geojson);
-  }, [geojson]);
+  const validateAndSubmit = () => {
```
jessicamcinchak, line 113: Refactoring to `validateAndSubmit` & `getBody` are consistent with DrawBoundary, which should make managing these components feel a bit more aligned in the future

editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
```diff
@@ -111,54 +125,65 @@ export default function PlotNewAddress(props: PlotNewAddressProps): FCReturn {
...
+              coordinates &&
+              JSON.stringify({
+                type: "Feature",
+                geometry: {
+                  type: "Point",
+                  coordinates: [coordinates?.longitude, coordinates?.latitude],
+                },
+                properties: {},
+              })
+            }
```
jessicamcinchak, line 153: Our current implementation  was actually failing to properly display the new address point when going "back" - now it's fixed to set `drawGeojsonData` here. 

We don't show a visual border of the team boundary, so we also no longer need to set `geojsonData`; `clipGeojsonData` is sufficient !

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -143,11 +180,20 @@ function Component(props: Props) {
...
+            mapValidationError={mapValidationError}
+            setMapValidationError={setMapValidationError}
+            showSiteDescriptionError={showSiteDescriptionError}
+            setShowSiteDescriptionError={setShowSiteDescriptionError}
           />
+          {Boolean(address) && isValidating && (
+            <DelayedLoadingIndicator
+              msDelayBeforeVisible={50}
+              text="Fetching data..."
+            />
```
jessicamcinchak, line 195: _Either_ page needs to wait to fetch supplmentary address data, but we were incorrectly only showing the loading indicator on the select-existing-address page before. Now we'll show it on both pages (this impacts this draft PR #3136).

Currently, the overall `Card` will disable the continue button when `isValidating === true` - I'll address this last.

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -124,11 +110,62 @@ function Component(props: Props) {
...
+  return (
+    <Card
+      handleSubmit={validateAndSubmit}
+      isValid={
+        page === "new-address" && !isValidating
+          ? true
+          : Boolean(address) && !isValidating
+      }
+    >
+      {getBody()}
```
DafyddLlyr, line 164: nit: This was a bit unexpected - it should really be a `React.FC`, e.g. `<Body/>` (or some other name).

I know this isn't introduced here so it can be ignored or fixed forward as you feel best - same applies to DrawBoundary

DafyddLlyr (approved):
One very small comment, working as expected on Pizza. Thanks for detailed PR description and breaking this one up 👍

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
jessicamcinchak: Thanks for noting this - will fix forward in this case to not over-complicate dependent PR right now !

PR #3143, opened 2024-05-13
fix(a11y): always enable continue button on FindProperty select page

jessicamcinchak:
Follow-on from #3141 - please review that first as fewer open questions than here !

### DIFF
--- a/editor.planx.uk/src/@planx/components/FindProperty/Public/Autocomplete.tsx
+++ b/editor.planx.uk/src/@planx/components/FindProperty/Public/Autocomplete.tsx
@@ -10,6 +10,7 @@ import find from "lodash/find";
 import { parse, toNormalised } from "postcode";
 import React, { useEffect, useState } from "react";
 import InputLabel from "ui/public/InputLabel";
+import ErrorWrapper from "ui/shared/ErrorWrapper";
 import Input from "ui/shared/Input";
 
 import type { SiteAddress } from "../model";
@@ -25,6 +26,12 @@ interface PickOSAddressProps {
   initialSelectedAddress?: Option;
   id?: string;
   description?: string;
+  showPostcodeError: boolean;
+  setShowPostcodeError: React.Dispatch<React.SetStateAction<boolean>>;
+  addressAutocompleteError?: string;
+  setAddressAutocompleteError: React.Dispatch<
+    React.SetStateAction<string | undefined>
+  >;
 }
 
 const AutocompleteWrapper = styled(Box)(({ theme }) => ({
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/@planx/components/FindProperty/Public/Autocomplete.tsx
```diff
@@ -184,16 +191,21 @@ export default function PickOSAddress(props: PickOSAddressProps): FCReturn {
...
+          <address-autocomplete
+            id="address-autocomplete"
+            data-testid="address-autocomplete-web-component"
+            postcode={sanitizedPostcode}
+            initialAddress={selectedOption?.title || ""}
+            osProxyEndpoint={`${process.env.REACT_APP_API_URL}/proxy/ordnance-survey`}
+            arrowStyle="light"
+            labelStyle="static"
+          />
+        </ErrorWrapper>
```
jessicamcinchak, line 208: Wrapping the web component in our error wrapper isn't perfect because it means the error message & select label will be out of normal order, with the error message _above_ the label rather than _below_.

We can: 
1. Ignore this for now - it'll never be visible at the same time as the postcode error message, so hard to notice
2. Implement an error wrapper directly in our web component code (and create/pass new `error` prop?)
    - Looking through all the alpha gov examples of this component, none of them have error handling :melting_face: https://alphagov.github.io/accessible-autocomplete/examples/
3. Other suggestions?

Screenshot from 2024-05-13 20-36-07

editor.planx.uk/src/@planx/components/FindProperty/Public/index.tsx
```diff
@@ -111,7 +116,16 @@ function Component(props: Props) {
   }, [data, address]);
 
   const validateAndSubmit = () => {
-    // TODO `if (isValidating)` on either page, wrap Continue button in error mesage?
+    if (isValidating) {
+      setDataFetchError("Please wait for data fetching to complete");
+    }
```
jessicamcinchak, line 121: After the user has selected an address on either page, we need to make a quick request to Planning Data to fetch supplmental data about it that we can't get from Ordnance Survey - and we want to prevent the user from "continuing" before this fetch completes. 

We used to simply set `isValid={!isValidating}` on the `Card`, but now we want to keep the continue button always enabled.

A couple questions here: 
- What's the best language for explaining this? 
- Is wrapping the "Continue" button in the error right considering the proposed design in #3136 ? If not the button, what else could we wrap? 
  - [TODO] Prevent "Continue" button from expanding to full-screen width when in error state (see screenshot below) 
  
Screenshot from 2024-05-13 21-02-50

jessicamcinchak: Open questions from dev call: 
- Should we move the loading indicator _into_ the button in these instances rather than wrap it in an error?

PR #3144, opened 2024-05-14
feat: create `submission_services_log` view

jessicamcinchak:
Creates a new view of submission & pay events - now exposing successful _and failed_ events - more similar to what we log in `#planx-notifications`.  

Here's a snippet of what this looks like running against an environment with submissions:
Screenshot from 2024-05-14 14-31-46

Next steps:
- Query via frontend and update visual display in submissions log tab
- Remove any views or outdated data structures associated with current implementation

### DIFF
--- a/hasura.planx.uk/migrations/1715689232003_create view submission services log /up.sql
+++ b/hasura.planx.uk/migrations/1715689232003_create view submission services log /up.sql
@@ -1,0 +1,45 @@
+create or replace view public.submission_services_log as 
+with payments as (
+    select 
+      session_id,
+      payment_id::text as event_id,
+      'Pay' as event_type,
+      initcap(status) as status,
+      '{}'::jsonb as response,
+      created_at
+    from payment_status
+    where status != 'created'
+      and created_at >= '2024-01-01'
+), submissions as (
+    select 
+      (seil.request -> 'payload' -> 'payload' ->> 'sessionId')::uuid as session_id,
+      se.id as event_id,
+      case 
+        when se.webhook_conf::text like '%bops%' then 'Submit to BOPS'
+        when se.webhook_conf::text like '%uniform%' then 'Submit to Uniform'
+        when se.webhook_conf::text like '%email-submission%' then 'Send to email'
+        when se.webhook_conf::text like '%upload-submission%' then 'Upload to AWS S3'
+        else se.webhook_conf::text
+      end as event_type,
+      case 
+        when seil.status = 200 then 'Success'
+        else format('Failed (%s)', seil.status)
+      end as status,
+      seil.response::jsonb,
+      seil.created_at
+    from hdb_catalog.hdb_scheduled_events se
+      left join hdb_catalog.hdb_scheduled_event_invocation_logs seil on seil.event_id = se.id
+    where se.webhook_conf::text not like '%email/%'
+      and seil.created_at >= '2024-01-01'
+), all_events as (
+    select * from payments 
+    union all
+    select * from submissions
+)
+SELECT
+  ls.flow_id,
+  ae.*
+FROM all_events ae
+  left join public.lowcal_sessions ls on ls.id = ae.session_id
+WHERE ls.flow_id is not null
+order by ae.created_at desc;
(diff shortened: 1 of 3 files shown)


hasura.planx.uk/migrations/1715689232003_create view submission services log /up.sql
```diff
@@ -0,0 +1,45 @@
+create or replace view public.submission_services_log as 
+with payments as (
+    select 
+      session_id,
+      payment_id::text as event_id,
+      'Pay' as event_type,
```
jessicamcinchak, line 6: in the future we may want to expand this to more clearly distinguish 'Pay' & 'Invite to Pay' - but we know the latter is rarely used currently and all payment attempts will be captured here regardless of method.

hasura.planx.uk/migrations/1715689232003_create view submission services log /up.sql
```diff
@@ -0,0 +1,45 @@
...
+      initcap(status) as status,
+      '{}'::jsonb as response,
+      created_at
+    from payment_status
+    where status != 'created'
+      and created_at >= '2024-01-01'
```
jessicamcinchak, line 12: Is there any reason to query & display submissions prior to January 2024? This filter gives us a bit more consistency in event formatting & we could always look up submissions outside this timeframe on request.

hasura.planx.uk/migrations/1715689232003_create view submission services log /up.sql
```diff
@@ -0,0 +1,45 @@
...
+      end as event_type,
+      case 
+        when seil.status = 200 then 'Success'
+        else format('Failed (%s)', seil.status)
+      end as status,
+      seil.response::jsonb,
+      seil.created_at
+    from hdb_catalog.hdb_scheduled_events se
+      left join hdb_catalog.hdb_scheduled_event_invocation_logs seil on seil.event_id = se.id
+    where se.webhook_conf::text not like '%email/%'
```
jessicamcinchak, line 32: I'm filtering out email events (reminders, confirmations, saves etc) - they'd be easy to add in the future if we want, but let's keep straightforward for now :+1:

hasura.planx.uk/migrations/1715689232003_create view submission services log /up.sql
```diff
@@ -0,0 +1,45 @@
...
+    select * from payments 
+    union all
+    select * from submissions
+)
+SELECT
+  ls.flow_id,
+  ae.*
+FROM all_events ae
+  left join public.lowcal_sessions ls on ls.id = ae.session_id
+WHERE ls.flow_id is not null
```
jessicamcinchak, line 44: `flow_id` is null in older cases where we used to create send events on `/draft` or `/preview` routes - this filters those out & there's been a fix in place to not create events on those routes at all anymore.

jessicamcinchak:
[Viewer of submissions in the Editor - v. 2.0](https://trello.com/c/gAmYcGIi/2892-viewer-of-submissions-in-the-editor-v-20)

hasura.planx.uk/migrations/1715689232003_create view submission services log /up.sql
DafyddLlyr: Totally on board with keeping this filter in place - makes a whole lot of sense.

hasura.planx.uk/migrations/1715689232003_create view submission services log /up.sql
DafyddLlyr: Alternatively, could we remove those records which should not have been created?

DafyddLlyr (approved):
Looks like a solid improvement 👍

hasura.planx.uk/migrations/1715689232003_create view submission services log /up.sql
jessicamcinchak: Event logs will simply be cleaned up by scheduled data sanitation, right? Personally would like to avoid a manual removal of records - they were arguably created "correctly" at the time, so I don't see an issue with keeping the audit history of them!

PR #3146, opened 2024-05-14
feat: Styled team viewer table

ianjon3s:


editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx
```diff
@@ -1,19 +1,158 @@
...
+}));
 
 export type TeamMember = Omit<User, "teams" | "isPlatformAdmin"> & {
   role: Role;
 };
 
+const roleLabels: Record<string, string> = {
+  platformAdmin: "Admin",
+  teamEditor: "Editor",
+};
```
jessicamcinchak, line 39: It's technically possible for the query to return `teamViewer` and I think we also want to account for that here - even though it's quite rare!

editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx
```diff
@@ -1,19 +1,158 @@
...
-      {JSON.stringify(teamMembersByRole, null, 4)}
-    </Box>
+    <Container maxWidth="contentWrap">
+      <Box py={7}>
+        <EditorRow>
+          <Typography variant="h2" component="h3" gutterBottom>
+            Team editors
+          </Typography>
+          <Typography variant="body1">
+            Editors have access to edit your flows and services.
```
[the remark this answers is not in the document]
jessicamcinchak: I think it's only possible to navigate to `/members` directly and we actually need to still add a link to this page from the Team Settings menu or elsewhere! 

The entire Team Settings menu is currently scoped to `teamEditors`-only though if that's where we decide to include it :heavy_check_mark:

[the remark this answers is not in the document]
jessicamcinchak: nit: `flows` & `services` are essentially synonymns, can we start standardising our editor language around one or the other? The overall list within a team uses "My services"?

editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx
```diff
@@ -1,19 +1,158 @@
...
+            Admins have editor access across all teams.
+          </Typography>
+          {renderMemberTable(platformAdmins)}
+        </EditorRow>
+        <EditorRow>
+          <Typography variant="h2" component="h3" gutterBottom>
+            Archived team editors
+          </Typography>
+          <Typography variant="body1">
+            Past members who no longer have access to edit your team.
```
jessicamcinchak, line 151: nit: maybe "Past team members who no longer have access to the Editor, but may be part of the edit history of your services" provides a little more context?

editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx
```diff
@@ -1,19 +1,158 @@
...
+        <EditorRow>
+          <Typography variant="h2" component="h3" gutterBottom>
+            Admins
+          </Typography>
+          <Typography variant="body1">
+            Admins have editor access across all teams.
+          </Typography>
+          {renderMemberTable(platformAdmins)}
+        </EditorRow>
+        <EditorRow>
```
jessicamcinchak, line 146: nit: we could be clever about only showing sections that apply to this team by doing something like this:
```
{archivedMembers.length > 0 && (
   <EditorRow>
     .... 
   </EditorRow>
)}

**Review by jessicamcinchak — COMMENTED** [review]
Looking so good !

File: editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx (hunk 1 above)
Comment (ianjon3s) [reply editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx:39]: Added, thanks!

File: editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx (hunk 3 above)
Comment (ianjon3s) [reply editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx:151]: Makes a lot more sense written this way, with continuity of referring to "services" 👍

File: editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx (hunk 4 above)
Comment (ianjon3s) [reply editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx:146]: Great suggestion, will add

File: editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx [hunk 5]
```diff
@@ -1,19 +1,162 @@
...
+    <Container maxWidth="contentWrap">
+      <Box py={7}>
+        <EditorRow>
+          <Typography variant="h2" component="h3" gutterBottom>
+            Team editors
+          </Typography>
+          <Typography variant="body1">
+            Editors have access to edit your services.
+          </Typography>
+          {renderMemberTable(otherRoles)}
```
DafyddLlyr, line 136: nit: Having a function here is a little bit of an anti-pattern I think. 

For readability, I'd suggest making a React component. Something like this - 

```tsx
<MemberTable members={otherRoles}/>
```

This could live outside of the scope of `TeamMembers` and we'd end up with easier to test components as we can just pass props in. Happy to talk through this refactor if a call is the best way 👍

editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx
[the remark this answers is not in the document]
DafyddLlyr: This is a great point.

Currently there's a commented-out bit of code in the menu which can be found here - https://github.com/theopensystemslab/planx-new/blob/ian/team-viewer-table/editor.planx.uk/src/components/Header.tsx#L549

Once we're ready to proceed with this we can comment it back in 👍 

In terms of who can access the page, only those with permissions could see the link currently. However, any savvy logged in user could navigate to `not-my-team/members` and view this page.

I'll open a small PR just now to add a guard for this in the routes page.

DafyddLlyr (approved):
Looking great - love it! Can easily see an "add member" feature being added here soon 😉 

I think we should comment in the link in the dropdown menu before we merge this!

editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx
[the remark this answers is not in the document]
DafyddLlyr: PR here - https://github.com/theopensystemslab/planx-new/pull/3151

editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx
```diff
@@ -1,19 +1,162 @@
...
+const StyledAvatar = styled(Avatar)(({ theme }) => ({
+  background: theme.palette.background.dark,
+  color: theme.palette.common.white,
+  fontSize: "1em",
+  fontWeight: "600",
+  marginRight: theme.spacing(1),
+}));
+
+const StyledTableRow = styled(TableRow)(({ theme }) => ({
+  "&:nth-child(even)": {
```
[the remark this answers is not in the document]
ianjon3s: Good spot, I'll try to remember this one for future

jessicamcinchak (approved):

editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx
DafyddLlyr, line 1: Perfect! This diff really nicely illustrates the idea that a React functional component (`React.FC`) is just a function, and arguments become React props.

You could also "unnest" this component and take it outside of the scope of `TeamMembers` - it's maybe worth doing as there's no need for it to be nested within `TeamMembers`. I think a few of the helper functions would also need to move with this. This shows a nice compositional model where we can make many small independent components and compose them together 👍 

Not a totally necessary step here, happy to approve as-is.

DafyddLlyr (approved):

DafyddLlyr (approved):
Thanks very much for picking up that change - happy for this to merge and go to PO testing once we get it to production 👍

PR #3149, opened 2024-05-15
feat(a11y): Remove disabled button from `Question` component

DafyddLlyr:
## What does this PR do?
- Enables the "Continue" button by default on all `Question` components
- Adds an error message if no option is selected
  - Copy taken from error message of text input

image

Test flow here with all variants of the question component - https://3149.planx.pizza/testing/question-error-test

### DIFF
--- a/editor.planx.uk/src/@planx/components/Question/Public/Question.tsx
+++ b/editor.planx.uk/src/@planx/components/Question/Public/Question.tsx
@@ -16,7 +16,7 @@ import React from "react";
 import FormWrapper from "ui/public/FormWrapper";
 import FullWidthWrapper from "ui/public/FullWidthWrapper";
 import ErrorWrapper from "ui/shared/ErrorWrapper";
-import { number, object, string } from "yup";
+import { mixed, object, string } from "yup";
 
 export interface IQuestion {
   id?: string;
@@ -65,7 +65,9 @@ const Question: React.FC<IQuestion> = (props) => {
     validationSchema: object({
       selected: object({
         id: string().required("Select your answer before continuing"),
-        a: number().required(),
+        a: mixed().required().test(value => 
+          typeof value === "number" || 
+          typeof value === "string")
       })
     }),
   });
--- a/editor.planx.uk/src/@planx/components/Question/Public/Question.tsx
+++ b/editor.planx.uk/src/@planx/components/Question/Public/Question.tsx
@@ -103,7 +103,7 @@ const Question: React.FC<IQuestion> = (props) => {
           >
             {props.text}
           </FormLabel>
-          <ErrorWrapper id={`${props.id}-error`} error={formik.errors.selected?.id}>
+          <ErrorWrapper id={props.id} error={formik.errors.selected?.id}>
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/@planx/components/Question/Public/Question.tsx
```diff
@@ -57,15 +59,17 @@ const Question: React.FC<IQuestion> = (props) => {
...
-    validate: () => {
-      // do nothing
-    },
+    onSubmit: (values) => props.handleSubmit({ answers: [values.selected.id] }),
+    validateOnBlur: false,
+    validateOnChange: false,
+    validationSchema: object({
+      selected: object({
+        id: string().required("Select your answer before continuing"),
+        a: mixed().required().test(value =>
```
jessicamcinchak, line 68: `selected.a` can be `undefined` in `initialValues` - is it actually required here or optional? 

Also nice to see the `mixed()` syntax in action :muscle:

editor.planx.uk/src/@planx/components/Question/Public/Question.tsx
```diff
@@ -100,68 +103,70 @@ const Question: React.FC<IQuestion> = (props) => {
...
-          <RadioGroup
-            aria-labelledby={`radio-buttons-group-label-${props.id}`}
-            name={`radio-buttons-group-${props.id}`}
-            value={formik.values.selected.id}
-          >
-            <Grid
-              container
-              spacing={layout === QuestionLayout.Basic ? 0 : 2}
-              alignItems="stretch"
+          <ErrorWrapper id={`${props.id}-error`} error={formik.errors.selected?.id}>
```
jessicamcinchak, line 106: nit: the `ErrorWrapper` component already prepends `error-message-` to whatever `id` is set here, so no need to explicitly use `-error` suffix as you have

jessicamcinchak (approved):
Thank you for picking this one up!! 

Two small comments/questions before merging, but looks nearly good to go & working as expected on pizza !

jessicamcinchak:
[Disabled buttons and unexpected input functionality [Usability]](https://trello.com/c/elmTB1Pq/2828-disabled-buttons-and-unexpected-input-functionality-usability)

editor.planx.uk/src/@planx/components/Question/Public/Question.tsx
DafyddLlyr: Good question! I think what I have is correct, as this is a required value _at time of submission_. 

When the component loads the value is not set as it doesn't exist (so it's optional in the props). When navigating "back" it will be set already. The value for `selected.a` is set alongside `selected.id` here - https://github.com/theopensystemslab/planx-new/blob/61937317759cd41518d52a397a204584eb2e5a53/editor.planx.uk/src/%40planx/components/Question/Public/Question.tsx#L115-L116

jessicamcinchak: That distinction makes sense - thanks for clarifying!

PR #3150, opened 2024-05-15
a11y: Result accordions non-expanding option

ianjon3s:
## What does this PR do?

Address issue where result summaries that do not have expandable content (help text) are still given expand functionality and associated aria labels.

To prevent this the accordion component is only used if there is associated content to expand.

Current:
https://editor.planx.dev/testing/result/published

Updated:
https://3150.planx.pizza/testing/result/published?analytics=false


DafyddLlyr (approved):
Very good catch! 👏 Nice simple solution also

PR #3152, opened 2024-05-15
chore: Replace `team.name` for `team.slug` in `feedback_summary` view

DafyddLlyr:
Having `team_slug` means we can match on variables in existing dashboards without an additional join in Metabase.


hasura.planx.uk/migrations/1715784133713_run_sql_migration/up.sql
```diff
@@ -0,0 +1,49 @@
+CREATE OR REPLACE VIEW "public"."feedback_summary" AS
+SELECT 
+    fb.id AS feedback_id,
+    t.slug AS team,
```
DafyddLlyr, line 4: This line is the only change here.

jessicamcinchak (approved):

PR #3153, opened 2024-05-15
feat: submissions feed reflects events rather than successful audit entries only

jessicamcinchak:
Currently, the submissions log only displays "successful" events, and it shows a lot of empty space for event types that may not apply to your service. 

Now, it reflects events applicable to this service only, including failed ones & retries, like a filtered view of Slack's `#planx-notifications` channel.

Screenshot from 2024-05-17 12-34-08

**Testing links:**
- LDC with payment & "dual processing" https://3153.planx.pizza/lambeth/apply-for-a-lawful-development-certificate/settings/submissions (1 application in log)
- Discretionary service with send to email only https://3153.planx.pizza/testing/quick-test/settings/submissions (3 unique applications in log)


editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
```diff
@@ -0,0 +1,170 @@
...
+        }}
+      >
+        {submission.status === "Success"
+          ? JSON.stringify(JSON.parse(submission.response?.data?.body), null, 2)
+          : JSON.stringify(
+              JSON.parse(submission.response?.data?.message),
+              null,
+              2,
+            )}
+      </Box>
```
jessicamcinchak, line 165: This is doing a basic if/else right now and grouping all variations of "Send", but we can easily customise responses to each unique event type if desired to emphasise third-party IDs/reference numbers etc. 

For now, I'm just pretty-printing the raw response which is what we most commonly use/share with partners when troubleshooting failures.

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/index.tsx
```diff
@@ -1,40 +1,87 @@
...
+
+export interface Submission {
+  sessionId: string;
+  eventId: string;
+  eventType:
+    | "Pay"
+    | "Submit to BOPS"
+    | "Submit to Uniform"
+    | "Send to email"
+    | "Upload to AWS S3";
```
jessicamcinchak, line 19: nit: I feel like the tense of these isn't quite right - should they all be past tense? What's a word that works for _either_ a successful or failed payment event (eg why I didn't go with "Paid" to begin with!)?

Open to suggestions!

jessicamcinchak:
[Viewer of submissions in the Editor - v. 2.0](https://trello.com/c/gAmYcGIi/2892-viewer-of-submissions-in-the-editor-v-20)

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
```diff
@@ -0,0 +1,170 @@
...
+  );
+};
+
+const FormattedResponse: React.FC<Submission> = (submission) => {
+  if (submission.eventType === "Pay") {
+    return (
+      <Box sx={{ margin: 1 }}>
+        <Typography variant="body2">
+          {`GOV.UK Pay reference: ${submission.eventId}`}
+        </Typography>
```
jessicamcinchak, line 144: "Pay" events currently have a null `response` in the view, but I think I can update the view to improve this a bit with a join to the `payment_status_enum` table, then we could have a more consistently formatted code block in the dropdown like : 
```
{
  status: "payment completed"
  reference: "gov-pay-id"
}
```

This should also give us that extra context of "failed how" (eg cancelled by user, rejected payment method, etc) that we currently get in Slack.

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
```diff
@@ -0,0 +1,170 @@
...
+        </TableCell>
+        <TableCell>{submission.sessionId}</TableCell>
+        <TableCell>
+          <IconButton
+            aria-label="expand row"
+            size="small"
+            onClick={() => setOpen(!open)}
+          >
+            {open ? <KeyboardArrowUp /> : <KeyboardArrowDown />}
+          </IconButton>
```
DafyddLlyr, line 124: I haven't actually spotted the issue yet in code, but the behaviour here is not quite right.

When I open the accordion for a Uniform submission, this button disappears and can't be closed.


https://github.com/theopensystemslab/planx-new/assets/20502206/0a7af909-07a1-40ea-baee-6582ccd5dd90

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/index.tsx
DafyddLlyr: Yeah this is tricky right! I wonder if "Submission to..." and "Payment" are any better?

"Paid" seems to imply success, but a "Payment" could be successful or failed? See comment in review about success and failure states which relate to this also.

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
```diff
@@ -0,0 +1,170 @@
...
+            </TableCell>
+            <TableCell>
+              <strong>Session ID</strong>
+            </TableCell>
+            <TableCell></TableCell>
+          </TableRow>
+        </TableHead>
+        <TableBody>
+          {submissions.map((submission) => (
+            <CollapsibleRow {...submission} />
```
DafyddLlyr, line 74: There's a few ways to make each even instance of this shaded I think.

If native CSS in a styled component won't work with something like `&:nth-of-type(even)` in a styled component (I would think is possible), then we could achieve this with JS.

```tsx
<TableBody>
  {submissions.map((submission, i ) => (
    <CollapsibleRow {...submission} isEven={(i + 1) % 2} />
  ))}
</TableBody>
```

And then we can use isEven is our `sx` prop.

DafyddLlyr (commented):
A few comments on code, a few comments below!

I think we probably want to make our success / failure states more visually obvious. I really like the icons denoting the event types but I feel the green/red distinction is not quite visually strong enough.

A nice usability improvement here would be a copy button next to the session id - super nice though that the click interaction is just on the caret so I can easily copy/paste as-is. I wasn't expecting this from the initial screenshot.

I'm also seeing a slight visual issue with some of the table row styles with a doubled-up bottom border - 

image

Similar to @ianjon3s' comments on Slack, grouping by sessionId might be nice but Hasura doesn't natively support this in queries which I think would mean getting back to the view, or "post-processing" in TS when we have the data - not ideal? Grouping by sessionId also breaks the chronological order of events - also not easily understandable probably? I think I'd favour keeping this as-is for now and awaiting feedback before jumping into changes here - this is a whole lot better than the current submission view and Slack imho 👍

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
ianjon3s: The property needed here is `white-space: pre-wrap` which took a bit of digging. I've added to an incoming PR for style update

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
ianjon3s: This would be achievable in CSS using `:nth-of-type(4n+3)`, which is applying to every 4th row (every other summary row) starting at row three (the second summary row).

However as we're using a shaded background for the expanding element, I think we'd be best to leave the summary rows as white background.

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
jessicamcinchak: Updated in latest db migration !

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/index.tsx
jessicamcinchak: No one else mentioned this, so keeping as-is for now :+1:

api.planx.uk/modules/send/email/index.ts
```diff
@@ -69,7 +69,9 @@ export async function sendToEmail(
     );
 
     return res.status(200).send({
-      message: `Successfully sent "Submit" email`,
+      message: `Successfully sent to email`,
+      inbox: sendToEmail,
+      govuk_notify_template: "Submit",
```
jessicamcinchak, line 74: Slightly tweaking API responses here so they offer a bit more context in the "response" row (will apply to future events only) :heavy_check_mark:

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
```diff
@@ -0,0 +1,185 @@
...
+}));
+
+// Style the table container like an event feed with bottom-anchored scroll
+const Feed = styled(TableContainer)(() => ({
+  maxHeight: "60vh",
+  overflow: "auto",
+  display: "flex",
+  flexDirection: "column-reverse",
+  readingOrder: "flex-visual",
+}));
```
jessicamcinchak, line 45: This post was helpful for figuring out the "feed" style, and has some good accessibility notes to revisit one day when we're ready to get the Editor certified https://kittygiraudel.com/2024/02/26/css-only-bottom-anchored-scrolling-area/

ianjon3s (approved):
This is a huge leap forward as a v2! I vote we get it on production and iterate further based on feedback

PR #3155, opened 2024-05-16
feat: Submissions log styles update

ianjon3s:
## What does this PR do?

Builds on the submissions log work:
https://github.com/theopensystemslab/planx-new/pull/3153

Summary of changes:
- Move border to collapsable element to prevent double border
- Prevent overflow by setting explicit column widths and wrapping code
- Use separate icons for success/error, so that it not communicated by colour only
- Wrap cell inner contents in a div with `flex` to allow cells to expand vertically

**Preview:**
https://3155.planx.pizza/testing/test-submissions/settings/submissions

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
@@ -1,3 +1,4 @@
+import CheckCircleIcon from "@mui/icons-material/CheckCircle";
 import KeyboardArrowDown from "@mui/icons-material/KeyboardArrowDown";
 import KeyboardArrowUp from "@mui/icons-material/KeyboardArrowUp";
 import Payment from "@mui/icons-material/Payment";
@@ -54,7 +55,7 @@ const EventsLog: React.FC<GetSubmissionsResponse> = ({
 
   return (
     <TableContainer>
-      <Table>
+      <Table sx={{ tableLayout: "fixed" }}>
         <TableHead>
           <TableRow>
             <TableCell sx={{ width: 300 }}>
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
```diff
@@ -85,30 +87,32 @@ const CollapsibleRow: React.FC<Submission> = (submission) => {
...
+              ) : (
+                <CancelIcon color="error" />
+              )}
+            </Avatar>
+
+            {submission.eventType === "Pay" ? <Payment /> : <Send />}
+            <Typography variant="body2" ml={0.5}>
+              {submission.eventType}
+            </Typography>
+          </Box>
```
jessicamcinchak, line 115: nit: I'm not sure I'm sold on the double icon, it feels quite awkward to me. If success/failure isn't actually clear enough by red/green color alone would it be better to just add a column to the table where we print the actual word "Success" or "Failded (XXX)"? Would have additional benefit of providing additional context of status code on fail? 

Screenshot from 2024-05-16 13-54-59

jessicamcinchak: Eg columns: "Event", "Status", "Date", "Session ID" ?

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
```diff
@@ -153,6 +166,7 @@ const FormattedResponse: React.FC<Submission> = (submission) => {
           margin: 1,
           maxWidth: "contentWrap",
           overflowWrap: "break-word",
+          whiteSpace: "pre-wrap",
```
jessicamcinchak, line 169: :eyes: :bulb: thank you!!

jessicamcinchak (approved):
Huge improvement, thank you!! 

One nit picky suggestion for communicating success/failure, but not a blocker !

PR #3156, opened 2024-05-16
feat: Capture `node_data` directly alongside feeback

DafyddLlyr:
## What does this PR do?
 - Adds new `node_data` column to `feedback` table
 - Writes to the `node_data` column when feedback is logged
 - Rebuilds `feedback_summary` view to read from this column

## Context
The current `feedback_summary` view in pretty slow running, likely due to the filtered join per-row to the `published_flows` table. This solution stores the node data directly alongside the feedback, thus mitigating this issue (hopefully!).

Local testing is good - significantly faster to query the view now.

## Next steps...
Once the following changes are live on production, I'm proposing to run the following SQL to populate the new column. I'm opting to do this manually instead of as a Hasura migration as it will be slightly easier to manage and monitor, and it's pretty low-risk (we're just populating a new column).
 
```sql
UPDATE feedback
SET node_data = (
    SELECT 
        (published_flows.data -> feedback.node_id) -> 'data'
    FROM 
        published_flows
    WHERE 
        published_flows.flow_id = feedback.flow_id 
        AND published_flows.created_at < feedback.created_at
    ORDER BY 
        published_flows.created_at DESC
    LIMIT 1
);
```
I've tested this locally with a copy of feedback data, however the sync script just copies a single published flow so it only populates a few records.

This matches the current join on the flow, as seen here - https://github.com/theopensystemslab/planx-new/blob/4eeea8f6f460c3997a6f62e729e735c32d5833b4/hasura.planx.uk/migrations/1715784133713_run_sql_migration/up.sql#L37-L49

### DIFF
--- a/hasura.planx.uk/migrations/1715868696352_run_sql_migration/up.sql
+++ b/hasura.planx.uk/migrations/1715868696352_run_sql_migration/up.sql
@@ -1,0 +1,40 @@
+DROP VIEW "public"."feedback_summary";
+
+CREATE OR REPLACE VIEW "public"."feedback_summary" AS
+SELECT 
+    fb.id AS feedback_id,
+    t.slug AS team,
+    f.slug AS service_slug,
+    fb.created_at,
+    fb.node_id,
+    fb.device,
+    fb.user_context,
+    fb.user_comment,
+    fb.feedback_type,
+    fb.status,
+    fb.node_type,
+    fb.node_data,
+    COALESCE(
+        fb.node_data ->> 'title', 
+        fb.node_data ->> 'text', 
+        fb.node_data ->> 'flagSet'
+    ) AS node_title,
+    fb.node_data ->> 'description' AS node_text,
+    fb.node_data ->> 'info' AS help_text,
+    fb.node_data ->> 'policyRef' AS help_sources,
+    fb.node_data ->> 'howMeasured' AS help_definition,
+    COALESCE(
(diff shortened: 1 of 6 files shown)


hasura.planx.uk/migrations/1715868696352_run_sql_migration/up.sql
```diff
@@ -0,0 +1,40 @@
...
+    ) AS address,
+    (fb.user_data -> 'passport' -> 'data' -> '_address' ->> 'uprn') AS uprn,
+    (fb.user_data -> 'passport' -> 'data' ->> 'proposal.projectType') AS project_type,
+    (fb.user_data -> 'passport' -> 'data' ->> 'property.constraints.planning') AS intersecting_constraints
+FROM 
+    feedback fb
+LEFT JOIN 
+    flows f ON f.id = fb.flow_id
+LEFT JOIN 
+    teams t ON t.id = fb.team_id;
```
jessicamcinchak, line 38: :100:

jessicamcinchak (approved):
Code looks good & plan sounds solid! Agree since it's a new column, it's an easy to manage manual migration for existing records (with comfort we can remove/retry if any surprises) :+1:

DafyddLlyr:
Update: SQL executed on production, getting expected results and `feedback_summary` view now significantly faster and loading in Metabase without issue  🎉

PR #3158, opened 2024-05-17
feat: Setup routing, queries, and permissions for `/:flow/feeback`

DafyddLlyr:
## What does this PR do?
 - Sets up permissions to allow logged in users to access data from `feedback_summary` view
 - Sets up routing and placeholder component to display this in the Editor, following the pattern established in https://github.com/theopensystemslab/planx-new/pull/3109

image

## Next steps
- Display data in table
- Get more detailed info to a user - either CSV download per-row, or established accordion pattern maybe? I've included a function to fetch this data and we can have a think about this one.

## Questions
My intention here is to display a small subset of the data - are these minimal columns the most relevant ones?

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/EditorMenu.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/EditorMenu.tsx
@@ -1,4 +1,5 @@
 import FactCheckIcon from "@mui/icons-material/FactCheck";
+import RateReviewIcon from "@mui/icons-material/RateReview";
 import TuneIcon from "@mui/icons-material/Tune";
 import Box from "@mui/material/Box";
 import IconButton from "@mui/material/IconButton";
@@ -68,10 +69,10 @@ const MenuButton = styled(IconButton, {
 
 function EditorMenu() {
   const { navigate } = useNavigation();
-  const { lastChunk } = useCurrentRoute();
+  const { url } = useCurrentRoute();
   const rootPath = rootFlowPath();
 
-  const isActive = (route: string) => lastChunk.url.pathname.endsWith(route);
+  const isActive = (route: string) => url.pathname.endsWith(route);
   const handleClick = (route: string) =>
     !isActive(route) && navigate(rootPath + route);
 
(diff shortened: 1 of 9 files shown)


editor.planx.uk/src/pages/FlowEditor/components/EditorMenu.tsx
```diff
@@ -68,10 +69,10 @@ const MenuButton = styled(IconButton, {
 
 function EditorMenu() {
   const { navigate } = useNavigation();
-  const { lastChunk } = useCurrentRoute();
+  const { url } = useCurrentRoute();
   const rootPath = rootFlowPath();
 
-  const isActive = (route: string) => lastChunk.url.pathname.endsWith(route);
+  const isActive = (route: string) => url.pathname.endsWith(route);
```
DafyddLlyr, line 75: This change was required as the `/feedback` route (in its own file) ended with a `/`

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -268,6 +268,76 @@
...
+          - user_context
+          - created_at
+        filter:
+          team:
+            members:
+              _and:
+                - user_id:
+                    _eq: x-hasura-user-id
+                - role:
+                    _eq: teamEditor
```
DafyddLlyr, line 339: Permission model copied from `team_themes` table.

hasura.planx.uk/migrations/1715929881942_run_sql_migration/up.sql
```diff
@@ -0,0 +1,40 @@
+DROP VIEW "public"."feedback_summary";
+
+CREATE OR REPLACE VIEW "public"."feedback_summary" AS
+SELECT 
+    fb.id AS feedback_id,
+    t.slug AS team_slug,
```
DafyddLlyr, line 6: Apologies for the churn in this view recently, I've renamed `team` to the more explicit and correct `team.slug` so that the relationship can just be called `team` in Hasura.

jessicamcinchak (approved):
Looks good to me - initiial set of data fields seems sensible until we hear otherwise !

PR #3159, opened 2024-05-17
feat: add quick note to contextualise length of edit history

jessicamcinchak:
"Undo" is live on prod this morning which is exciting because all flows are loading with an _existing_ edit history (whereas pizzas and staging do _not_ sync edit history). 

When the earliest timeline point is not "Created flow", it's not always obvious why the timeline ends. This adds a quick contextual note in cases where edit history for this service is longer than the 15 operations we show! 

Screenshot from 2024-05-17 10-30-37


editor.planx.uk/src/pages/FlowEditor/components/EditHistory.tsx
```diff
@@ -110,142 +111,155 @@ const EditHistory = () => {
   return (
     <Box>
       {loading && !data ? (
-        <DelayedLoadingIndicator />
+        <DelayedLoadingIndicator
+          msDelayBeforeVisible={0}
+          text="Fetching edit history..."
```
DafyddLlyr, line 116: 👍

DafyddLlyr (approved):

PR #3160, opened 2024-05-17
chore: Feedback sync script (example)

DafyddLlyr:
Not really advocating for this to be merged, but I wanted some sort of history here and it may prove to be helpful for others.

I added a `feedback.sql` so I could get a copy of production data locally.

Commenting out an item in the list seemed a nicer pattern than a flag per table or some other optional inputs to the script. A bit quick and dirty but worked well for me.

### DIFF
--- a/scripts/seed-database/container.sh
+++ b/scripts/seed-database/container.sh
@@ -18,7 +18,19 @@ mkdir -p /tmp
 # Create sync.sql file for all our comnands which will be executed in a single transaction
 touch '/tmp/sync.sql'
 
-tables=(flows users teams flow_document_templates team_members team_themes)
+tables=(
+  # Mandatory tables
+  flows 
+  users 
+  teams 
+  flow_document_templates 
+  team_members 
+  team_themes
+  # Optional tables
+  # Please comment in if working on a feature and you require example data locally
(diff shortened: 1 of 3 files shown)


jessicamcinchak (approved):
This is a nice example to have! 

Was definitely thinking about this recently with undo (syncing `operations`) and submissions (syncing events tables in `hdb_catalog` schema) - and instead just lost a bit of time to creating them the old fashioned way :+1:

scripts/seed-database/container.sh
```diff
@@ -18,7 +18,19 @@ mkdir -p /tmp
...
+  # Mandatory tables
+  flows 
+  users 
+  teams 
+  flow_document_templates 
+  team_members 
+  team_themes
+  # Optional tables
+  # Please comment in if working on a feature and you require example data locally
+  # You will need to manually grant select permissions to the github_actions on production, and update main.sql
```
jessicamcinchak, line 31: :+1:

DafyddLlyr:
The operation example is actually a good one and a second use for this type of things. I'll actually go ahead and merge - if it feels excessive or noisy in future we can move to docs or an alternate script or something in future.

PR #3164, opened 2024-05-17
fix: only show active users in "Team editors" 

jessicamcinchak:
Spotted while testing here https://editor.planx.dev/newcastle/members


ianjon3s (approved):
Good spot!

PR #3165, opened 2024-05-17
fix: Position of address loading with save and return

ianjon3s:
## What does this PR do?

The updated loading indicator for address lookup is positioned `absolute` with a unit reference to `bottom`. This fails on when save & return is enabled as the button is no longer the last element in the container.

This PR adjusts so that the indicator is positioned with reference to the `top` of the container.


jessicamcinchak (approved):
Thanks for catching!

PR #3167, opened 2024-05-17
fix: Remove input checkbox elements and replace with spans

DafyddLlyr:
Quick proof of concept, we can now replace ✅ and ❎ with CSS, svgs, other elements to get a PlanX looking "checkbox".

Ian: Updated with CSS spans, matching the structure of the ARIA authoring guide exactly (no wrapper element on the label)


DafyddLlyr:
✅ I approve all of @ianjon3s changes here but can't make it official as I opened the PR

ianjon3s (approved):
Looks good to me, also tested well on mobile

PR #3170, opened 2024-05-20
feat(backend): Toggle flow online or offline

DafyddLlyr:
## What does this PR do?
- Adds `flows.status` column, which can be one of "online" or "offline"
- Adds `flow_status_history` audit table which could be used to check the status of a flow at a given point in time, and could be used to filter analytics if we wished to do so
- Populates table to indicate that all flows are currently "online"
- Adds trigger `tack_flow_status_history()` which has the following behaviour -
  - Called on `UPDATE` to `flow.status` or `INSERT` to `flows`
  - Closes "open" `flow_status_history` record, and opens new one with toggled status
  
## Testing
- Hasura introspection tests for table permissions
- Gherkin integration tests for trigger behaviour 

## Next steps...
- Simple toggle in Editor to control status
- Restrict public access to flow based on status
- Slack message on toggle (?)

Regression tests passing here - https://github.com/theopensystemslab/planx-new/actions/runs/9162033526

### DIFF
--- a/e2e/tests/api-driven/src/flowStatusHistory/flowStatusHistory.feature
+++ b/e2e/tests/api-driven/src/flowStatusHistory/flowStatusHistory.feature
@@ -1,0 +1,14 @@
+Feature: Flow status history
+
+  @regression @flow-status-history
+  Scenario: Creating a flow
+    When a new flow is added
+    Then the status of the flow is online by default
+    And a flow_status_history record is created
+
+  @regression @flow-status-history
+  Scenario: Updating a flow 
+    When a new flow is added
+    And the flow status is changed to offline
+    Then the open flow_status_history record is updated
+    And a new flow_status_history record is created
--- a/e2e/tests/api-driven/src/flowStatusHistory/flowStatusHistory.feature
+++ b/e2e/tests/api-driven/src/flowStatusHistory/flowStatusHistory.feature
@@ -2,13 +2,13 @@ Feature: Flow status history
 
   @regression @flow-status-history
   Scenario: Creating a flow
-    When a new flow is added
-    Then the status of the flow is online by default
+    Given a flow exists
+    Then the status of the flow is offline by default
     And a flow_status_history record is created
 
   @regression @flow-status-history
   Scenario: Updating a flow 
-    When a new flow is added
-    And the flow status is changed to offline
+    Given a flow exists
+    When the flow status is changed to online
     Then the open flow_status_history record is updated
     And a new flow_status_history record is created
--- a/e2e/tests/api-driven/src/flowStatusHistory/helpers.ts
+++ b/e2e/tests/api-driven/src/flowStatusHistory/helpers.ts
@@ -20,16 +20,16 @@ export const cleanup = async () => {
 interface GetFlowStatus {
   flows: {
     status: FlowStatus;
-  }[];
+  };
 }
 
 export const getFlowStatus = async (flowId: string) => {
   const {
-    flows: [{ status }],
+    flows: { status },
   } = await $admin.client.request<GetFlowStatus>(
     gql`
       query GetFlowStatus($flowId: uuid!) {
-        flows(where: { id: { _eq: $flowId } }) {
+        flows: flows_by_pk(id: $flowId) {
           status
         }
       }
--- a/e2e/tests/api-driven/src/flowStatusHistory/helpers.ts
+++ b/e2e/tests/api-driven/src/flowStatusHistory/helpers.ts
@@ -1,0 +1,72 @@
+import { FlowStatus } from "@opensystemslab/planx-core/types";
+import { $admin } from "../client";
+import { createTeam } from "../globalHelpers";
+import gql from "graphql-tag";
+
+export const setup = async () => {
+  const teamId = await createTeam();
+
+  const world = { teamId };
+
+  return world;
+};
+
+export const cleanup = async () => {
+  await $admin.user._destroyAll();
+  await $admin.flow._destroyAll();
+  await $admin.team._destroyAll();
+};
+
+interface GetFlowStatus {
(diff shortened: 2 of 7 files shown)


DafyddLlyr:
[Toggle on/off to take a service live](https://trello.com/c/QzdRsmRf/2857-toggle-on-off-to-take-a-service-live)

e2e/tests/api-driven/package.json
```diff
@@ -6,7 +6,7 @@
   },
   "dependencies": {
     "@cucumber/cucumber": "^9.3.0",
-    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#2732b6e",
+    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#58a0647",
```
DafyddLlyr, line 9: I'll update this cross-project once I've merged https://github.com/theopensystemslab/planx-core/pull/388 to `main`

e2e/tests/api-driven/src/flowStatusHistory/flowStatusHistory.feature
```diff
@@ -0,0 +1,14 @@
+Feature: Flow status history
+
+  @regression @flow-status-history
+  Scenario: Creating a flow
+    When a new flow is added
+    Then the status of the flow is online by default
+    And a flow_status_history record is created
```
jessicamcinchak, line 7: Not sure I agree with this assumption! 

I'd personally think that _existing_ flows should be "online by default", but newly added ones should be "offline by default" as that will match our user's mental models of "preparing to go live" ?

e2e/tests/api-driven/src/flowStatusHistory/flowStatusHistory.feature
```diff
@@ -0,0 +1,14 @@
...
+
+  @regression @flow-status-history
+  Scenario: Updating a flow 
+    When a new flow is added
```
jessicamcinchak, line 11: nit: should this be `When an existing flow is updated` ?

e2e/tests/api-driven/src/flowStatusHistory/helpers.ts
```diff
@@ -0,0 +1,72 @@
...
+  }[];
+}
+
+export const getFlowStatus = async (flowId: string) => {
+  const {
+    flows: [{ status }],
+  } = await $admin.client.request<GetFlowStatus>(
+    gql`
+      query GetFlowStatus($flowId: uuid!) {
+        flows(where: { id: { _eq: $flowId } }) {
```
jessicamcinchak, line 32: nit: is `flows_by_pk` slightly more performant & descriptive here?

e2e/tests/api-driven/src/flowStatusHistory/steps.ts
```diff
@@ -0,0 +1,110 @@
...
+
+Then("the status of the flow is online by default", async function () {
+  const status = await getFlowStatus(this.flowId);
+
+  assert.equal(
+    status,
+    "online",
+    `Flow status is ${status} - it should be "online"`,
+  );
+});
```
jessicamcinchak, line 37: See above regarding "default" assumptions - I think this should be flipped!

hasura.planx.uk/tests/flow_status_history.test.js
```diff
@@ -0,0 +1,78 @@
...
+
+    test("cannot query flow_status_history", () => {
+      expect(i.queries).not.toContain("flow_status_history");
+    });
+
+    test("cannot create, update, or delete flows or their associated operations", () => {
+      expect(i).toHaveNoMutationsFor("flow_status_history");
+    });
+  });
+});
```
jessicamcinchak, line 78: :+1:

jessicamcinchak (commented):
Looking good & thorough (appreciate all the testing here), but a few questions around domain assumptions here! 

Plus one more point I'm thinking about for "Next steps..."
- Only write analytics records when URL does not have param `?analytics=false` _AND_ when flow status is "online" ! 
  - If we can just skip writing the tracking events altogether when flows are "offline", that should save us having to add joins to `flow_status_history` to the analytics views? 
  - This would similarly assume that _existing flows_ are online by default & new ones are offline I think?

e2e/tests/api-driven/src/flowStatusHistory/flowStatusHistory.feature
DafyddLlyr: This is a good shout - agree! Changes made in [8dc5368](https://github.com/theopensystemslab/planx-new/pull/3170/commits/8dc5368a9deb87e6e3735eabd01a46c714c09327)

e2e/tests/api-driven/src/flowStatusHistory/flowStatusHistory.feature
DafyddLlyr: Fixed - this was just a shortcut to re-use a step but I've replaced both with `Given()`

e2e/tests/api-driven/src/flowStatusHistory/helpers.ts
DafyddLlyr: Good catch thanks - fixed in [7c43879](https://github.com/theopensystemslab/planx-new/pull/3170/commits/7c43879c50d3639f7a2eb0e70764eba128c9fd4a)

DafyddLlyr:
> Only write analytics records when URL does not have param ?analytics=false AND when flow status is "online" 

This is an elegant and clever solution - very much on board! Much easier than a tricky join and actually matches the model we want and users would expect (analytics are collected for "online" services only).

e2e/tests/api-driven/package.json
DafyddLlyr: Done ✅

PR #3175, opened 2024-05-21
chore: Remove unused `/themes` folder

DafyddLlyr:
Looks like legacy code which is no longer used - none of the exported objects are used elsewhere in the codebase, we now use `themes.ts` for this functionality.


jessicamcinchak (approved):
:broom:

PR #3176, opened 2024-05-21
chore: Rename `QuestionHeader` to `CardHeader`

DafyddLlyr:


ianjon3s (approved):
Useful change and works as expected

PR #3181, opened 2024-05-21
Revert "chore: bump planx-core"

jessicamcinchak:
Reverts theopensystemslab/planx-new#3180 (I'll fix-forward in #planx-core next) 

Regression tests passing against this branch :white_check_mark: which fixes our failure last night: https://github.com/theopensystemslab/planx-new/actions/runs/9186773259

**Context:**
A Planx tester didn't wait for constraints to load (x3 apps for a demo) and submitted to BOPS. The payloads were valid with empty constraints, and we got a successful response from BOPS, but the apps didn't display in BOPS. Likely because they have an internal dependency on expected constraints? I was tagged in the help-issue request instead of the BOPS team, and should have been initially clear which side of the fence this bug was on. 

I thought this simple payload mapping change would be fine & allow me to resubmit the apps with minimal constraints that reflect "user answers", but it doesn't actually validate as expected (eg https://api.editor.planx.dev/admin/session/74068ec3-2e9c-408a-ac6f-7832d1eb011f/digital-planning-application?skipValidation=true). Bad testing on my part here and will need to look more closely at the PlanningDesignations type definition in the schema.

**Next steps:**
- [x] Manually search constraints & add `_constraints` key to to lowcal session `data`. Update session, confirm it validates and populates constraints, re-submit to Lambeth BOPS staging (x3) https://opendigitalplanning.slack.com/archives/C0241GWFG4B/p1716310913515129?thread_ts=1715963778.547149&cid=C0241GWFG4B
- [ ] Revert this change on planx-new (to hopefully also fix failing regression tests?)
- [ ] Fix-forward in planx-core and add testing. Bump planx-new again

### DIFF
--- a/api.planx.uk/package.json
+++ b/api.planx.uk/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@airbrake/node": "^2.1.8",
-    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#1b95557",
+    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#2732b6e",
     "@types/isomorphic-fetch": "^0.0.36",
     "adm-zip": "^0.5.10",
     "aws-sdk": "^2.1467.0",
--- a/api.planx.uk/pnpm-lock.yaml
+++ b/api.planx.uk/pnpm-lock.yaml
@@ -12,8 +12,8 @@ dependencies:
     specifier: ^2.1.8
     version: 2.1.8
   '@opensystemslab/planx-core':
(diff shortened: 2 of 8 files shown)


jessicamcinchak:
Was able to successfully validate & resubmit here by adding `_constraints` to the session data manually (luckily all three tests used same addess). Closing this PR as we don't actually need to revert, we shouldn't hit this edge case in the meantime until we can fix forward.

Previously, when we sent empty constraints (which is valid per the schema) we got a successful response from BOPS but the applications weren't actually showing up for officers - lesson learned that this maybe shouldn't have been our help-issue to begin with !

DafyddLlyr (approved):
Thanks for the quick fix!

PR #3183, opened 2024-05-21
feat(frontend): Toggle flow online or offline

DafyddLlyr:


ianjon3s (approved):
This works great as a solution using current form styles. I'll be standardising the form styles imminently so will factor this in. I'll also think about how we can show the online/offline status on the 'select a service' screen.

editor.planx.uk/src/pages/FlowEditor/lib/store/settings.ts
```diff
@@ -33,6 +37,19 @@ export const settingsStore: StateCreator<
 
   setFlowSettings: (flowSettings) => set({ flowSettings }),
 
+  flowStatus: undefined,
```
jessicamcinchak, line 40: Should this be `"online"` by default rather than `undefined` ? 

Noticing that all existing flows on pizza are showing up as "offline" by default, which is opposite of intended behavior right? Perhaps just a data sync issue?

editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings.tsx
```diff
@@ -200,25 +221,82 @@ const ServiceSettings: React.FC = () => {
...
+                }
+              />
+            }
+          />
+          <Typography variant="body1">
+            Toggle your service between "offline" and "online".
+          </Typography>
+          <Typography variant="body1">
+            A service must be online to be accessed by the public, and to enable
+            analytics gathering.
```
jessicamcinchak, line 268: nit: I wonder if it's worth adding one more sentence here like "Offline services can still be edited and published as normal." to encourage good maintenance?

editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings.tsx
```diff
@@ -200,25 +221,82 @@ const ServiceSettings: React.FC = () => {
...
+            }}
+            control={
+              <Switch
+                name="service.status"
+                color="primary"
+                checked={statusForm.values.status == "online"}
+                onChange={() =>
+                  statusForm.setFieldValue(
+                    "status",
+                    statusForm.values.status == "online" ? "offline" : "online",
```
jessicamcinchak, line 257: nit: is there a reason to not use `===` equality checks here and above on L253?

jessicamcinchak (approved):

editor.planx.uk/src/pages/FlowEditor/lib/store/settings.ts
jessicamcinchak: Thanks for explanation here - totally agree `undefined` is correct store value now that data sync/migration is fixed in background :+1:

PR #3191, opened 2024-05-24
feat: show which planning constraints are checked in the editor (non-interactive)

jessicamcinchak:
Planning Constraints is one of our most complex Editor components, but actually exposes very few configuration options to an editor. This both limits flexibilty, and causes a fair number of async questions & help-issue reports around how it actually works. 

**Changes:**
- Adds a link to our Notion service spec page about "how constraints work" directly to the editor modal
- Disables the main `data field` input as valid payloads are actually very reliant on `property.constraints.planning`
- Adds a table with selectable-rows showing which individual constraints are queried and how each is sourced
  - Currently fully disabled & for information-only, but can obviously be select-able in future - allowing multiple constraints components to be used throughout a service, each checking a subset of datasets on the branches relevant

Screenshot from 2024-05-31 09-52-17

**Next steps:**
- [This PR] Design feedback - it's a dense, crowded table! 
- [Future PR] Wire up rows to be selectable, and set correct params on GIS API request based on row selections
  - Adjust public display to account for fewer constraints - split lists structure wouldn't make sense if only 1 thing is checked for example
  - Adjust `handleSubmit` & `makeData` to account for _multiple_ constraints components


editor.planx.uk/src/@planx/components/PlanningConstraints/model.ts
```diff
@@ -23,3 +23,135 @@ export type IntersectingConstraints = Record<string, string[]>;
...
+    source: "Planning Data",
+    datasets: ["tree", "tree-preservation-order", "tree-preservation-zone"],
+  },
+  {
+    name: "Classified Road",
+    key: "road.classified",
+    source: "Ordnance Survey",
+    datasets: ["OS Mastermap Highways"],
+  },
+];
```
jessicamcinchak, line 157: My thinking here is that once this is "hooked up", we'd exclusively maintain this list here & remove https://github.com/theopensystemslab/planx-new/blob/main/api.planx.uk/modules/gis/service/local_authorities/metadata/base.ts

editor.planx.uk/src/@planx/components/PlanningConstraints/Editor.tsx
```diff
@@ -33,34 +43,122 @@ function PlanningConstraintsComponent(props: Props) {
...
+              />
+            </InputRow>
             <InputRow>
               <Input
                 format="data"
                 name="fn"
                 placeholder={formik.values.fn}
                 value={formik.values.fn}
-                onChange={formik.handleChange}
+                disabled
```
jessicamcinchak, line 80: I think it's helpful to still show (rather than hide) this input so you know how this component will be recorded in your passport, but disabling it helps us ensure valid payloads later. If we're okay with this pattern, it's definitely one that would also be relevant & repeatable in FindProperty & DrawBoundary.

jessicamcinchak:
[Give 'Planning Constraints' component a 'hide from user' toggle](https://trello.com/c/Vjb3omgj/2845-give-planning-constraints-component-a-hide-from-user-toggle)

editor.planx.uk/src/@planx/components/PlanningConstraints/Editor.tsx
```diff
@@ -33,34 +43,117 @@ function PlanningConstraintsComponent(props: Props) {
...
+                            </InputRow>
+                            <InputRow>
+                              <Input
+                                format="data"
+                                name="val"
+                                value={dataset.val}
+                                disabled
+                              />
+                            </InputRow>
+                          </InputGroup>
```
jessicamcinchak, line 134: I'm now using disabled `InputRows` within each `TableRow`/`TableCell` here, as I think this has a more similar feel to Question & Checklist "Options" ? I think a table, rather than list, is still right here as it will facilitate the "(De)Select all" action across rows. 

These styles can absolutely be further tweaked before we wire-up & actually allow interaction!  

Here's my original draft using plain `Typography`, which I think looked more crowded? 
Screenshot from 2024-05-24 10-08-15

editor.planx.uk/src/@planx/components/PlanningConstraints/model.ts
jessicamcinchak: As I'm working on #3216 - I'm realising I need the list of Planning Data `datasets` more easily accessible via the Editor to make a correct external map link when more than one dataset is checked for a single constraint - hence opening this for review now!

editor.planx.uk/src/@planx/components/PlanningConstraints/Editor.tsx
```diff
@@ -1,15 +1,25 @@
+import Box from "@mui/material/Box";
+import Checkbox from "@mui/material/Checkbox";
```
DafyddLlyr, line 2: design: We currently use a number of different components for controlling boolean values in the Editor - 
 - MUI switches (in design settings)
 - PlanX checkboxes (in `Send` component)
 - MUI checkboxes here

This is a fix forward / design question which @ianjon3s may want his eyes on rather than a recommendation to make this change in the current PR 👍

DafyddLlyr (approved):
Looking good! A few comments and design ideas, all fix-forwardable 👍

editor.planx.uk/src/@planx/components/PlanningConstraints/model.ts
```diff
@@ -23,3 +23,135 @@ export type IntersectingConstraints = Record<string, string[]>;
...
+
+interface Dataset {
+  text: string;
+  val: string;
+  source: "Planning Data" | "Ordnance Survey";
+  datasets: string[];
+  entity?: number;
+}
+
+export const availableDatasets: Dataset[] = [
```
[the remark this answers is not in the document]
jessicamcinchak: This makes sense, but `as const` seems to break the `Dataset` type from automatically propagating through correctely when mapped/iterated over - will come back to this one !

PR #3192, opened 2024-05-24
fix: Add new `flow.status` column to sync script

DafyddLlyr:
This PR adds the `flows.status` column to our sync script which was added in #3170 

Tested locally with `pnpm test-sync` command ✅

### DIFF
--- a/scripts/seed-database/write/flows.sql
+++ b/scripts/seed-database/write/flows.sql
@@ -10,7 +10,8 @@ CREATE TEMPORARY TABLE sync_flows (
   updated_at timestamptz,
   settings jsonb,
   copied_from uuid,
-  analytics_link text
+  analytics_link text,
+  status text
 );
 
 \copy sync_flows FROM '/tmp/flows.csv' WITH (FORMAT csv, DELIMITER ';');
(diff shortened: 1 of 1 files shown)


RODO94:
@DafyddLlyr how significant are the check errors for Vultr? And how do I access the pizza?

DafyddLlyr:
@RODO94 This PR is currently failing for the reasons outlined here - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1716548233407109 (OSL Slack)

As this is holding you up from getting data locally you have a few options here - 
 - You could checkout this branch locally, sync data, then carry on working on your branch
 - You could `git cherry-pick` this commit into your branch, sync data, and carry on. When we merge / rebase your branch, this cherry picked commit will be removed by git as it will already be on the target branch 🍒

jessicamcinchak:
@DafyddLlyr I've removed instance 3192 from vultr & fix is on `main`, so if you rebase to `main` here & push vultr should build fresh & succeed here for you on next go :crossed_fingers:

RODO94 (approved):
Tested locally and it runs fine, I can now see the database data.

PR #3193, opened 2024-05-24
test: try building vultr with ubuntu

jessicamcinchak:
Latest Alpine release (3.20) seems to have dropped `aws-cli` and we don't have option via Vultr to specify 3.19, so let's just try going back to Ubuntu builds temporarily until next Alpine release or fix 

See https://www.reddit.com/r/AlpineLinux/comments/1cyd3ws/am_i_tripping_or_was_awscli_removed_from_the/


freemvmt (approved):
nice ✨

jessicamcinchak:
Replaced by #3194

PR #3195, opened 2024-05-24
fix: GIS API should return geometry

jessicamcinchak:
Quick adjustment to start returning the WKT `geometry` & `point` fields from the raw Planning Data "entity" response in our own formatted response for any intersecting entities. 

We used to filter these out for simplicity, but now we want to start prototyping displaying maps of intersecting constraints on the frontend so we need the geometry! (We already have a WKT library in the editor to handle the GeoJSON conversion).


DafyddLlyr (approved):
Looks like we could also remove `omitGeometry()` if this was it's only use?

PR #3197, opened 2024-05-27
feat: List component - basic model, types & UI

DafyddLlyr:

PR #3198, opened 2024-05-27
feat: Basic CRUD for `userData` structure, setup `ListContext`

DafyddLlyr:
## What does this PR do?
- Follows on from https://github.com/theopensystemslab/planx-new/pull/3197
- Sets up `ListContext` to wrap components and share state and logic
- List items can be added, edited and removed
- Tests to describe this behaviour

https://github.com/theopensystemslab/planx-new/assets/20502206/5cf251de-6cff-458c-9651-22eabe40a1d9

## What's not in this PR?
- No form state / form validation - any values typed into inputs is not persisted
- Logic to stop new items being added when there's still an active/incomplete item
- UI polish, e.g. scroll to new item, scrol on delete or edit

## Context / Design
I feel like one of the main pain points with the `FileUploadAndLabel` component was having a group of interdependent components which relied on a shared state in the "root" component. This wasn't a problem initially, but as things grew it got harder and harder to manage, and there's a lot of prop-drilling and derived state there. I've tried to anticipate this a little by setting up a `React.Context` to wrap these components. This also means co-location of logic, and separation from style/structure which is nice.

Again, a lot of unfinished work and outstanding `TODO`s in this PR!

### DIFF
--- a/editor.planx.uk/src/@planx/components/List/Public/Context.tsx
+++ b/editor.planx.uk/src/@planx/components/List/Public/Context.tsx
@@ -1,6 +1,6 @@
-import React, { createContext, ReactNode,useContext, useState } from "react";
+import React, { createContext, ReactNode, useContext, useState } from "react";
 
-import { generateNewItem,Schema, UserData } from "../model";
+import { generateNewItem, Schema, UserData } from "../model";
 
 interface ListContextValue {
   schema: Schema;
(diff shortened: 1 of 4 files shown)


jessicamcinchak (approved):
Went through this one fairly quickly, appreciate these are iterative & feature-flagged. 

Thanks for super descriptive tests, looking good !

PR #3199, opened 2024-05-27
fix: ensure all maps are referencing same container

jessicamcinchak:
Spotted while grabbing screenshots & updating service spec pages for geo features. 

`PropertyInformation` component was using a redundant local definition rather than importing the shared `MapContainer` styled component. This was also leading to an oversized map on the PropertyInformation page, now it's a smaller reference map as expected.


DafyddLlyr (approved):
Nice spot!

PR #3203, opened 2024-05-28
feat: Adding a "Name" Column to "Flows" table

RODO94:
### Hasura Console
- Added *name* to **flows** table using the Hasura Console to generate the "up" and "down" sql migrations
- Altered the permissions on the *name* column inline with permissions allowed for the *slug* column

### Up File
- Added a comment to *name* column
- Added script to populate records where ``name = NULL`` with a formatted version of the slug

### Down File
- Added ``DROP COLUMN``

### Flows.sql Script
- Added *name* column


DafyddLlyr (commented):
Looking good! 🎉  

All as expected but one suggestion on how we can handle the transformation of "my-flow-slug" to "My flow slug".

hasura.planx.uk/migrations/1716902907392_alter_table_public_flows_add_column_name/up.sql
```diff
@@ -0,0 +1,9 @@
+alter table "public"."flows" add column "name" text
+ null;
+ comment on column "public"."flows"."name" is 'The name of the flow, entered by the user and used to generate the "slug"';
+
+UPDATE flows
+SET name = replace(initcap(replace(slug,'-','xax')),'xax',' ')
```
DafyddLlyr, line 6: Ah I figured `initcap()` would just handle the first letter, not the first letter of each word. This makes it trickier for sure!

Using `"xax"` is clever and certainly works as we don't have this substring in any flow slugs (tested with `select * from flows where slug ilike '%xax%';`

However it's a little bit unintuitive and took me a while to work out what was going on here 😅 It also relies on strings which users _could_ modify.

Is there another approach here that might be more robust? It should be possible to use substrings (e.g. upper case just first letter, concatenate with remaining lowercase string).

https://pgplayground.com/ can be handy as a quick playground for functions like this 👍

RODO94: Aha, so I should be explicitly altering the first letter as it'll always output the same result no matter what the user adds (unless they use dashes), and it's clearer to read?

Something like this:

```sql
SET name = replace(concat(upper(substring(slug,1,1)),right(slug,-1)),'-',' ')
```

And using the above ``UPPER`` method is easier to read than my previous method, so even changing out the ``xax`` string to use something like ``_`` wouldn't make it more robust, as shown below. 

```sql
SET name = replace(initcap(replace(slug,'-','_')),'_',' ')
```

I've been using the Hasura SQL console to try different methods since it required less setup... any concerns with that?

DafyddLlyr: No - totally legit and safe to use your local Hasura console for this also! 👌 

I'd favour the first option as despite being fairly complex syntax, the intention feels pretty clear.

However, the second approach would also work well and I have no objections! 👍 

As seen below, users cannot have special characters in the existing slug, so using the special character `_` as a replacement string is safe - 

image
image

And however unlikely it is that they use the string `xax`, it's certainly possible, so it's best not to use this - 

image
image

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -277,6 +277,7 @@ export const editorStore: StateCreator<
             where: { team: { id: { _eq: $teamId } } }
           ) {
             id
+
```
DafyddLlyr, line 280: nit: This looks like a whitespace change which has slipped through

scripts/seed-database/write/flows.sql
```diff
@@ -51,6 +54,7 @@ SET
   copied_from = EXCLUDED.copied_from,
   analytics_link = NULL,
   status = EXCLUDED.status;
+  name = EXCLUDED.name;
```
DafyddLlyr, line 57:
```suggestion
  status = EXCLUDED.status,
  name = EXCLUDED.name;
```

DafyddLlyr: A typo I've made myself many times when updating these scripts 😅 

You can actually test this locally by running `pnpm test-sync` which is a nice way of testing it.

hasura.planx.uk/migrations/1716902907392_alter_table_public_flows_add_column_name/up.sql
```diff
@@ -0,0 +1,9 @@
+alter table "public"."flows" add column "name" text
+ null;
+ comment on column "public"."flows"."name" is 'The name of the flow, entered by the user and used to generate the "slug"';
+
+UPDATE flows
+SET name = replace(concat(upper(substring(slug,1,1)),right(slug,-1)),'-',' ')
```
DafyddLlyr, line 6: 👍

DafyddLlyr (changes requested):
One minor typo in the sync script, but we're good to go after that! 😄

DafyddLlyr (approved):
Perfect! 🏅

PR #3206, opened 2024-05-28
fix(a11y): Non-text contrast (3:1)

DafyddLlyr:
image

image

Colour are derived from current colours, darkened to hit 3:1 ratio.

### DIFF
--- a/editor.planx.uk/src/theme.ts
+++ b/editor.planx.uk/src/theme.ts
@@ -57,13 +57,13 @@ const DEFAULT_PALETTE: Partial<PaletteOptions> = {
   action: {
     selected: "#F8F8F8",
     focus: "#FFDD00",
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/theme.ts
```diff
@@ -57,13 +57,13 @@ const DEFAULT_PALETTE: Partial<PaletteOptions> = {
   action: {
     selected: "#F8F8F8",
     focus: "#FFDD00",
-    disabled: "#B4B4B4",
+    disabled: "#949494",
```
DafyddLlyr, line 60: This meets the 3:1 requirement - https://webaim.org/resources/contrastchecker/?fcolor=949494&bcolor=FFFFFF

jessicamcinchak (approved):

PR #3207, opened 2024-05-28
fix(a11y): Remove placeholders in `DateInput`

DafyddLlyr:
This is the only public facing placeholder in PlanX - all others are in the Editor. To meet the requirements set out in the report, we'd need to increase font size to 18pt which is a fair bit bigger than the current size.

Removing placeholder text also meets the [GDS recommendation](https://design-system.service.gov.uk/components/text-input/), as well as the pattern currently used in [their Date Input components](https://design-system.service.gov.uk/components/date-input/) - 

> ### Avoid placeholder text
> Do not use placeholder text in place of a label, or for hints or examples, as:
>
> - it vanishes when the user starts typing, which can cause problems for users with memory conditions or when reviewing answers
> - not all screen readers read it out
> - its browser default styles often do not meet [minimum contrast requirements](https://www.w3.org/TR/WCAG22/#contrast-minimum)

image
image

### DIFF
--- a/editor.planx.uk/src/@planx/components/DateInput/Editor.test.tsx
+++ b/editor.planx.uk/src/@planx/components/DateInput/Editor.test.tsx
@@ -29,9 +29,9 @@ describe("DateInputComponent - Editor Modal", () => {
       </DndProvider>,
     );
 
-    const [minDay, maxDay] = screen.getAllByPlaceholderText("DD");
-    const [minMonth, maxMonth] = screen.getAllByPlaceholderText("MM");
-    const [minYear, maxYear] = screen.getAllByPlaceholderText("YYYY");
+    const [minDay, maxDay] = screen.getAllByLabelText("Day");
+    const [minMonth, maxMonth] = screen.getAllByLabelText("Month");
+    const [minYear, maxYear] = screen.getAllByLabelText("Year");
 
     expect(screen.queryByText(minError)).toBeNull();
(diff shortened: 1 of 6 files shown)


editor.planx.uk/src/@planx/components/DateInput/Editor.test.tsx
DafyddLlyr, line 1: I had some of these tests timeout on a CI run, then pass on multiple retries. 

Locally they run well within the expected timeout though not exactly fast. I'm not sure if this is an issue with the test or the CI runner? If this proves to be flaky I'll bump the timeout.

image

editor.planx.uk/src/ui/shared/DateInput.tsx
```diff
@@ -95,12 +101,15 @@ export default function DateInput(props: Props): FCReturn {
...
+          <Label
+            variant="body1"
+            htmlFor={`${props.id}-year`}
+            component={"label"}
+          >
             Year
           </Label>
           <Input
             value={year || ""}
-            placeholder="YYYY"
```
jessicamcinchak, line 103: Very happy to remove placeholders here, but should we also flag to content team that any DateInput components should now have descriptions like "For example, 23 4 2005" ? 

Noticed all the GDS examples have this and I wonder if it's how to get a better "usability" score in the absense of official placeholders? https://design-system.service.gov.uk/components/date-input/

jessicamcinchak (approved):

PR #3208, opened 2024-05-29
feat: Styled feedback table

ianjon3s:
## What does this PR do?

Adds a styled feedback table using the 'feed' style used for the submissions log.

- Data is separated into a summary row and further feedback on expand.
- Comments are truncated to 50 characters within the table, full comment in the expanded content. The first (bottom) example shows this.
- Help text is shown only for helpful/unhelpful help text feedback.
- User context is shown as the first item for "issue" type feedback, which corresponds to "What were you doing?" in the feedback form.

Example:
https://3208.planx.pizza/testing/find-out-if-you-need-planning-permission/feedback

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
@@ -95,8 +95,8 @@ const getDetailedFeedback = async (feedbackId: number) => {
     detailedFeedback.helpDefinition || ""
   } ${detailedFeedback.helpSources || ""}`.trim();
   const truncatedHelpText =
-    combinedHelpText.length > 80
-      ? `${combinedHelpText.slice(0, 80)}...`
+    combinedHelpText.length > 65
+      ? `${combinedHelpText.slice(0, 65)}...`
       : combinedHelpText;
 
   return {
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
@@ -91,9 +91,14 @@ const getDetailedFeedback = async (feedbackId: number) => {
     variables: { feedbackId },
   });
 
-  const combinedHelpText = `${detailedFeedback.helpText || ""} ${
-    detailedFeedback.helpDefinition || ""
-  } ${detailedFeedback.helpSources || ""}`.trim();
+  const combinedHelpText = [
+    detailedFeedback.helpText,
+    detailedFeedback.helpDefinition,
+    detailedFeedback.helpSources,
+  ]
+    .filter(Boolean)
+    .join(" ")
+    .trim();
   const truncatedHelpText =
     combinedHelpText.length > 65
       ? `${combinedHelpText.slice(0, 65)}...`
@@ -233,10 +238,10 @@ const CollapsibleRow: React.FC<CollapsibleRowProps> = (item) => {
 
   const labelMap: Record<string, string> = {
     userComment: item.type === "issue" ? "What went wrong?" : "User comment",
-    address: "Project address",
+    address: "Property address",
     projectType: "Project type",
     where: "Where",
-    browserPlatform: "Browser / platform",
+    browserPlatform: "Browser / device",
     combinedHelp: "Help text (more information)",
     userContext: "What were you doing?",
   };
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
@@ -83,7 +83,11 @@ const getDetailedFeedback = async (feedbackId: number) => {
     query: GET_FEEDBACK_BY_ID_QUERY,
     variables: { feedbackId },
   });
-  return detailedFeedback;
+  return {
+    ...detailedFeedback,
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
```diff
@@ -46,20 +82,168 @@ const getDetailedFeedback = async (feedbackId: number) => {
...
+
+  return (
+    <Container maxWidth="contentWrap">
+      <Box py={4}>
+        <EditorRow>
+          <Typography variant="h2" component="h3" gutterBottom>
+            Feedback log
+          </Typography>
+          <Typography variant="body1">
+            User-sumbitted feedback gathered for this service.
```
jessicamcinchak, line 136: nit: I know this is still draft, but spotted a typo "sumbitted" -> "submitted" while reviewing pizza ! 

Also, nitpick, is "Feedback from users about this service" perhaps more clear?

ianjon3s: Yes, clearer and simpler language

editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
```diff
@@ -46,20 +83,188 @@ const getDetailedFeedback = async (feedbackId: number) => {
...
+            }}
+          >
+            <DetailedFeedback>
+              <SummaryListTable sx={{ margin: "0" }}>
+                {detailedFeedback &&
+                  filteredFeedbackItems.map(
+                    (key, index) =>
+                      detailedFeedback[key] !== undefined && (
+                        <React.Fragment key={index}>
+                          <Box component="dt">{key}</Box>
```
DafyddLlyr, line 256: To display a more human-friendly label here we could create a map / object to handle this - 

```ts
const labelMap = {
    "userComment": "User comment",
    "address": "Address",
    "projectType": "Project type",
    ...etc
}
```

Then we could reference it here like this - 

```ts
<Box component="dt">{labelMap[key]}</Box>
```

ianjon3s: Thanks Daf, I've added this

editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
```diff
@@ -46,20 +90,216 @@ const getDetailedFeedback = async (feedbackId: number) => {
     query: GET_FEEDBACK_BY_ID_QUERY,
     variables: { feedbackId },
   });
-  console.log(detailedFeedback);
+
+  const combinedHelpText = `${detailedFeedback.helpText || ""} ${
+    detailedFeedback.helpDefinition || ""
+  } ${detailedFeedback.helpSources || ""}`.trim();
```
jessicamcinchak, line 96: nit: maybe this is a bit more readable for concatenating strings & filtering out any null ones? 
```
[detailedFeedback.helpText, detailedFeedback.helpDefinition, detailedFeedback.helpSources].filter(Boolean).join(" ").trim()
```

editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
```diff
@@ -46,20 +90,216 @@ const getDetailedFeedback = async (feedbackId: number) => {
...
+  const truncatedHelpText =
+    combinedHelpText.length > 65
+      ? `${combinedHelpText.slice(0, 65)}...`
+      : combinedHelpText;
```
jessicamcinchak, line 100: :+1: clever! I always forget the css syntax to do this, good reminder it can be clearly handled this way too

editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
```diff
@@ -46,20 +90,216 @@ const getDetailedFeedback = async (feedbackId: number) => {
...
+      case "unhelpful":
+        return ["combinedHelp", ...item.displayFeedbackItems];
+      default:
+        return item.displayFeedbackItems;
+    }
+  })();
+
+  const labelMap: Record<string, string> = {
+    userComment: item.type === "issue" ? "What went wrong?" : "User comment",
+    address: "Project address",
```
jessicamcinchak, line 236: nit: I'm not sure we use "Project address" anywhere else - is "Property address" or "Site address" perhaps more consistent with service content ?

jessicamcinchak (approved):
Changes look great here! A few final minor comments, but no stoppers - very happy for this to merge! 

Perhaps in a follow-up PR we can add a simple "Feedback" tab alongside "Submissions" & similar under "Flow settings"?

editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
ianjon3s: I'll change to "Property address" as this matches the question "Enter the postcode of the property" where the data is gathered

editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
ianjon3s: Thanks, I wasn't aware of the simplicity of `.filter(Boolean)`

PR #3210, opened 2024-05-29
fix: Incorrect status in `flow_status_history`

DafyddLlyr:
Apologies - just spotted this issue now when testing. This must have been missed when changing the model to offline by default here - https://github.com/theopensystemslab/planx-new/pull/3170/commits/81d7c7be64566b770ca5abb98b1b429e00a01a6f

I've actually just updated the _existing_ migration file here, with proposal to be to update the trigger directly on staging and production as a more immediate fix.

✅ Regression tests passing here - https://github.com/theopensystemslab/planx-new/actions/runs/9289391705


jessicamcinchak (approved):

DafyddLlyr:
Updated function on staging and prod dbs ✅ ✅

PR #3214, opened 2024-05-30
Production deploy

jessicamcinchak:
So our nightly database sync doesn't fail on the new `flows.name` column :slightly_smiling_face:


freemvmt (approved):
Had a quick scan, nothing overly contentious I don't think? so let's rolll ✨

PR #3201, opened 2024-05-28
feat: display a map for intersecting Planning Constraints

jessicamcinchak:
Branched off style updates in #3179 

A visual map of how a given constraint relates to your site boundary or address point is the first step of providing context necessary to understand if you should "override" or not.

PR #3216, opened 2024-05-30
feat: allows users to override inaccurate planning constraints

jessicamcinchak:
Begins to incorporate design feedback from call, replaces #3201 (links to a filtered, centered Planning Data map instead of rendering our own _per_ constraint).

Lots of new questions! So many links & actions to contend with inside of accordions.

**Outstanding tasks here if someone else picks up before 27 June:** 
- [ ] Finalise UI
  - [x] Show links to centred, filtered Planning Data maps for _any_ constraint
  - [x] Move Planning Data into "More information" per constraint (similar to Upload & Label help text pattern?)
  - [x] Show override/change links _per_ entity for _intersecting_ constraints only (because changing a negative constraint to positive would require user to give details about specific entity for valid ODP Schema payload, and all adjoining constraint examples are based on "false positives")
  - [ ] Show input box like "Tell us why" if changing? or Confirm button?
 - [ ] Wireup UI to store override value `onSubmit`
   - Keep existing `_constraints` passport var to reflect "original"/"auditable" Planning Data response (this does not power future automations)
   - If you change _all_ of the entities within a single positive constraint category, that value needs to be removed from `property.constraints.planning` and added to `_nots: { property.constraints.planning... }` (unless granular A4)
   - If you only change one of many entities within a single positive constraint category, `property.constraints.planning` & `_nots` should stay the same (unless granular A4)
   - Add a new internal passport variable like `_overrides: { property.constraints.planning... } ` to record the specific Planning Data entity IDs that were overriden
   - Test that auto-answers work as expected
- [ ] Add override data to ODP Schema payloads & update passport-mappings to populate
  - We don't currently share override info in the ODP Schema (for `property.type` or `property.constraints.planning`)
- [ ] Open for PO testing (can happen in parallel to final payload step)
- [ ] [Nice to have] Analytics tracking on override/change?
- [ ] [Nice to have] Slack log or feedback loop to council GIS teams?


jessicamcinchak:
Closing this one in favor of #3432. Will transfer checklist to Trello ticket instead :heavy_check_mark:

PR #3219, opened 2024-05-31
draft: Add name column changes to Editor files

RODO94:
## What does this PR do?

This Pull Request (PR) integrates changes to the data model regarding *flows* where we have added a ``name`` column to the database. This is a non-slug version of a flow name which should be displayed to users across the editor pages.

## Info on Changes

For queries and mutations looking to **edit a flow** or **retrieve a flow**, I have added ``name`` as a record to retrieve.

For validation logic, I have changed the ``Rename`` mutation to use ``name`` instead of ``slug`` for comparison check.

```diff
-  if(newName && slugify(newName) !== flow.slug)
+  if(newName && newName !== flow.name)
```

This will make it easier for users to make capitalisation changes

>This means that if you add **This is planning for england**  but you want to change it to **This is planning for England**, you won't receive an error for duplicate ``slug``


editor.planx.uk/src/pages/FlowEditor/lib/store/shared.ts
```diff
@@ -88,8 +90,8 @@ export const sharedStore: StateCreator<
     removeSessionIdSearchParam();
   },
 
-  setFlow({ id, flow, flowSlug }) {
-    this.setFlowNameFromSlug(flowSlug);
+  setFlow({ id, flow, flowSlug, flowName }) {
+    this.setFlowName(flowName);
     set({ id, flow, flowSlug });
```
jessicamcinchak, line 95: A couple questions: 
- Above on L37, you have `setFlowName: (flowSlug: string) => void;` - should the expected argument here actually be `flowName` based on how you're calling it later on L94? 
- If the expected argument is `flowName`, not `flowSlug`, could the call here be simplified to this ?
```
setFlow({ id, flow, flowSlug, flowName }) {
  set({ id, flow, flowSlug, flowName });
  get().initNavigationStore();
}
```

editor.planx.uk/src/routes/views/draft.tsx
```diff
@@ -32,12 +32,14 @@ export const draftView = async (req: NaviRequest) => {
   const flowData = await fetchDraftFlattenedFlowData(flow.id);
 
   const state = useStore.getState();
-  state.setFlow({ id: flow.id, flow: flowData, flowSlug });
-  state.setFlowNameFromSlug(flowSlug);
+  state.setFlow({ id: flow.id, flow: flowData, flowSlug, flowName: flow.name });
+  state.setFlowName(flow.name);
```
jessicamcinchak, line 36: similar to above comment, do we actually need both of these calls, or can it be simplified to just the first `setFlow({ ... })` one since that includes `flowName` now?

editor.planx.uk/src/routes/views/draft.tsx
```diff
@@ -32,12 +32,14 @@ export const draftView = async (req: NaviRequest) => {
...
   state.setGlobalSettings(data.globalSettings[0]);
   state.setFlowSettings(flow.settings);
   state.setTeam(flow.team);
 
+  console.log(flow.name);
```
jessicamcinchak, line 41: nit: please remove console.logs before merging to `main`

editor.planx.uk/src/routes/views/preview.tsx
```diff
@@ -28,8 +28,8 @@ export const previewView = async (req: NaviRequest) => {
   const flowData = await fetchFlattenedFlowData(flow.id);
 
   const state = useStore.getState();
-  state.setFlow({ id: flow.id, flow: flowData, flowSlug });
-  state.setFlowNameFromSlug(flowSlug);
+  state.setFlow({ id: flow.id, flow: flowData, flowSlug, flowName: flow.name });
+  state.setFlowName(flow.name);
```
jessicamcinchak, line 32: same question as above again!

editor.planx.uk/src/routes/views/standalone.tsx
```diff
@@ -24,14 +24,14 @@ const standaloneView = async (req: NaviRequest) => {
   const teamSlug =
     req.params.team || (await getTeamFromDomain(window.location.hostname));
   const data = await fetchDataForStandaloneView(flowSlug, teamSlug);
-
+  console.log(data);
```
jessicamcinchak, line 27: nit: oops another stray console.log

editor.planx.uk/src/routes/views/standalone.tsx
```diff
@@ -24,14 +24,14 @@ const standaloneView = async (req: NaviRequest) => {
...
   const {
     flows: [{ team, settings: flowSettings }],
     globalSettings,
   } = data;
 
   const state = useStore.getState();
-  state.setFlowNameFromSlug(flowSlug);
+  state.setFlowName(flowSlug);
```
jessicamcinchak, line 34: similar to above, can we clarify/confirm we're actually still expecting to pass `flowSlug` as the argument to `setFlowName()` now that we're _starting from_ name and generating slug?

RODO94: I was just trying to get a place to test this workflow, where would I be able to see and interact with the Standalone view?

editor.planx.uk/src/routes/views/draft.tsx
RODO94: Yeah, so removing ``setFlowName`` in favour of just using ``setFlow`` and pulling the ``name`` type from there?

editor.planx.uk/src/routes/views/standalone.tsx
jessicamcinchak: Right standalone views are a little tricky to get to - the best example is on the "Invite to Pay" journey.

Try these steps locally: 
- Set up a simple local test flow in a council that has GOV.UK Pay configured (eg Lambeth)
- This flow should have these components at minimum: FindProperty, Checklist that sets data field "proposal.projectType" with at least one option, SetValue with "fee = 1", Pay referencing "fee" data field and "Allow users to invite someone else to pay" option toggled on, Send (select 'email' option even though it won't be configured locally & will fail), Confirmation - 
- Publish your flow
- From the `/published` link of the public form, go trhough the test service and when you reach Pay, click "Invite to pay" then fill out the form using your own contact info so you get the invitation email
- Check your inbox, there should be an invitation to pay with a "magic link" - open it 
- That's a "standalone" page ! "Standalone" because it's a disconnected journey from the main service

There's a lot of new stuff in those steps, so please don't hesitate to shout if you get stuck at any point & want to screenshare!

RODO94: Ah! Okay cool, I'll have a look.

Can I ask why this standalone page is ``state.SetFlowName()`` should it not just read the flowName from the global store instead of setting it?

jessicamcinchak: So because this is a short separate journey only accessed by someone invited to _pay_ for the application (different than the person who has filled out the service), the original _flow_ itself never actually gets initiated or created when this "standalone" route loads, therefore flow attributes won't be set in global store! It's just a discrete, short set of static questions not controlled via the Editor, so we repeat a couple of these small details like setting flow & team names on the fly.

RODO94:
[Add "name" property to a flow](https://trello.com/c/G32D9unb/2598-add-name-property-to-a-flow)

editor.planx.uk/src/routes/team.tsx
```diff
@@ -76,4 +96,17 @@ const routes = compose(
   }),
 );
 
+const fetchDraftFlattenedFlowData = async (
```
RODO94, line 99: @jessicamcinchak here's the thing I added and mentioned in our Slack chat*


* *in the Slack chat, we discussed issues with switching from flowName and flowSlug state changes to condensing into a setFlow state change, but the change led to issues with reading ("_root" and "edges") which got resolved by fetching the flowData*

PR #3220, opened 2024-05-31
feat: List component - submission and back navigation

DafyddLlyr:
## What does this PR do?
 - Submits validated form values via `handleSubmit()`
   - This adds the values to the passport and breadcrumbs
   - I don't think that what's in this PR is necessarily the right or final shape of this data, but this just serves to complete the wiring up of this component
 - Pre-populate form values on "back" navigation
 - Replaces brittle and repetitive test selectors with `data-testid` values (see https://github.com/theopensystemslab/planx-new/pull/3198#discussion_r1617316748)


jessicamcinchak (approved):
Thanks for initial wire-up - this all looks good !

PR #3222, opened 2024-05-31
feat: List component - First iteration of a residential units schema

DafyddLlyr:
## What does this PR do?
 - Adds a first pass of a possible "Residential units" schema
 - Adds the concept (only enforced via UI) of a unique constraint

image

## Context
This is based on the following description - 

>For every unit we need to know:
> - Change type (proposed / existing)
> - Tenure type (single choice of 5 options)
> - Unit type (single choice of 6 options)
> - Number of bedrooms (number)
> - Optional: total number of identical units of this type (a way for the user to add multiples of identical units/clones without having to report them individually)

Source: https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1709126029217489 (OSL Slack)

## Next steps..
This is very much a rough pass. I think a sensible next step would be to pass this along to the content team on a Pizza to at least have a look at to try and evaluate the next steps. 

I think it's realistic, with the right setup, for the content team to be authoring and committing changes to these schemas as we move forward (even though configuration via the Editor modal is the way to go eventually!).

### DIFF
--- a/editor.planx.uk/src/@planx/components/List/Public/Fields.tsx
+++ b/editor.planx.uk/src/@planx/components/List/Public/Fields.tsx
@@ -27,7 +27,10 @@ export const TextFieldInput: React.FC<Props<TextField>> = ({
   const { formik, activeIndex } = useListContext();
 
   return (
-    <InputLabel label={data.title} htmlFor={id}>
+    <InputLabel
+      label={required === false ? data.title + " (optional)" : data.title}
+      htmlFor={id}
+    >
       <Input
         type={((type) => {
           if (type === "email") return "email";
@@ -71,7 +74,10 @@ export const NumberFieldInput: React.FC<Props<NumberField>> = ({
   const { formik, activeIndex } = useListContext();
 
   return (
-    <InputLabel label={data.title} htmlFor={id}>
+    <InputLabel
+      label={required === false ? data.title + " (optional)" : data.title}
+      htmlFor={id}
+    >
       <Box sx={{ display: "flex", alignItems: "baseline" }}>
         <Input
           required={required}
@@ -105,6 +111,7 @@ export const NumberFieldInput: React.FC<Props<NumberField>> = ({
 export const RadioFieldInput: React.FC<Props<QuestionField>> = ({
   id,
   data,
+  required,
 }) => {
   const { formik, activeIndex } = useListContext();
 
@@ -120,7 +127,7 @@ export const RadioFieldInput: React.FC<Props<QuestionField>> = ({
           },
         })}
       >
-        {data.title}
+        {required === false ? data.title + " (optional)" : data.title}
       </FormLabel>
       <ErrorWrapper
         id={`${id}-error`}
@@ -161,7 +168,10 @@ export const SelectFieldInput: React.FC<Props<QuestionField>> = (props) => {
   };
 
   return (
-    <InputLabel label={data.title} id={`select-label-${id}`}>
+    <InputLabel
+      label={required === false ? data.title + " (optional)" : data.title}
+      id={`select-label-${id}`}
+    >
       <ErrorWrapper
         id={`${id}-error`}
         error={getIn(formik.errors, `userData[${activeIndex}][${data.fn}]`)}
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/@planx/components/List/model.ts
```diff
@@ -42,6 +42,7 @@ export type NumberField = {
 export type QuestionField = {
   type: "question";
   required?: boolean;
+  unique?: boolean;
```
DafyddLlyr, line 45: The value of a real example in spotting oversights!

This reminded me that one of the restrictions we discussed would be that certain fields could be unique - meaning only one item in the list could contain this value.

editor.planx.uk/src/@planx/components/List/Public/Fields.tsx
```diff
@@ -145,12 +146,19 @@ export const RadioFieldInput: React.FC<Props<QuestionField>> = ({
...
-export const SelectFieldInput: React.FC<Props<QuestionField>> = ({
-  id,
-  data,
-  required,
-}) => {
+export const SelectFieldInput: React.FC<Props<QuestionField>> = (props) => {
   const { formik, activeIndex } = useListContext();
+  const { id, data, required } = props;
+
+  const isDisabled = (option: Option) => {
```
DafyddLlyr, line 153: This is just visual feedback, it would be nice to also incorporate this into the schema, most likely with a yup test ([docs](https://github.com/jquense/yup?tab=readme-ov-file#schematestoptions-object-schema)).

editor.planx.uk/src/@planx/components/List/model.ts
jessicamcinchak: This is a super good spot & a tricky one to consider! A few open questions: 
- I like the visual disabled feedback above, and it made me wonder if when there is an item in the schema set as "unique", should we also think of future validation that the "max" can't exceed the number of items in the unique field - otherwise you could add an item but the select could be entirely disabled? (Wouldn't necessarily make sense if "unique" ever has to be applied outside of `QuestionField` tho?)
- Can _many_ fields within the same schema have a unique constraint? Again, interesting if we ever want to try to automatically set "max" (calculate possible permutations?!)

editor.planx.uk/src/@planx/components/List/schemas/ResidentialUnits.ts
```diff
@@ -0,0 +1,73 @@
...
+        title: "How many bedrooms are there?",
+        fn: "numberBedrooms",
+        allowNegatives: false,
+      },
+    },
+    {
+      type: "number",
+      required: false,
+      data: {
+        title: "How many identical units of this type are there?",
```
jessicamcinchak, line 66: nit: we know from `AddressInput` & `ContactInput` that optional inputs should concat ` (optional)` to their titles for accessibility - can obviously hardcode that now, but may be worthwhile to automatically append based on `required: false` in future?

jessicamcinchak (approved):
Solid first pass ! Happy to get this across to content team for feedback & iteration early next week :+1:

editor.planx.uk/src/@planx/components/List/model.ts
DafyddLlyr: Yep! Hit both these exact questions also 💡 

I think right now...
 - We have no form validation for creation of `Schema` models, so can't enforce max === unique length natively in TS, we have to trust Editor's judgements to set max correctly. Eventually we'd enforce this via the Editor modal form that makes the schema. However, this actually isn't a real issue right now as if you have 5 items in a `unique` question and add a 6th list item, you can't save or proceed - so it's sort of self-limiting at least?
 - Yep - super interesting, and we start hitting permutation questions of a unique constraint on the _combination_ of question A and question B. I think this is actually possible and not super hard (again, we can calculate a max dynamically - maybe a better approach than the above "trust Editors" approach). I think this is relatively easily enforced via the UI, trickier to generate a dynamic yup schema which does this. 
 
Personal opinion is that either or both of these are sensible potential next steps, but maybe not day 1 requirements (likely a consideration for the content team to decide).

editor.planx.uk/src/@planx/components/List/schemas/ResidentialUnits.ts
DafyddLlyr: Very good spot 👍 Added in [442427f](https://github.com/theopensystemslab/planx-new/pull/3222/commits/442427f179d1a4fea1eb16b51b87422950b11316)

image

PR #3239, opened 2024-06-03
chore(deps): bump axios from 1.6.8 to 1.7.2 in all subdirectories

jessicamcinchak:
Dependabot only picked up the API subdirectory, but I think we want to keep this consistent throughout all subdirectories that require it.

Once merged, should automatically close #3235


jessicamcinchak:
Unclear why e2e are failing (cleared GA cache etc) - choosing to ignore this minor version update for now !

PR #3241, opened 2024-06-04
fix: list of flows should display and be sorted by most recent operations, not `flows.updated_at`

jessicamcinchak:
**Changes:**
August mentioned that it's really disorienting that all the flows within a team now show the same "Edited 5 days ago" timestamp - this is because this message was referencing `flows.updated_at` which changed when `flows.status` merged and is updated, whereas we really think about "edits" to a flow coming from the associated `operations`. 

Now updated to display based on `operations`, sharing the same formatting as in-graph view of "Last edited..." too. This means that both the list & graph timestamps will now match (previously they didn't :see_no_evil:), and that the list timestamps now ignore bulk flow migrations or non-data changes like renaming, toggling status, etc. 

See thread: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1717571853477949

**Testing:**
Click into a flow, edit the graph data, click back out to the list of services within the team and confirm it's at the top of the list with a correct "Last edited {at} by {user}" message. 

Note that real `operations` are _not_ synced from production to other environments (staging, local, pizza), so you won't see a first and last name associated with the timestamp on the pizza until new edits are made. Only production has a full history!

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
@@ -278,11 +278,15 @@ export const editorStore: StateCreator<
           ) {
             id
             slug
-            updated_at
-            operations(limit: 1, order_by: { id: desc }) {
+            updatedAt: updated_at
+            operations (
+              limit: 1,
+              order_by: { created_at: desc }
+            ) {
+              createdAt: created_at
               actor {
-                first_name
-                last_name
+                firstName: first_name
+                lastName: last_name
               }
             }
           }
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
@@ -272,17 +272,11 @@ export const editorStore: StateCreator<
     const { data } = await client.query({
       query: gql`
         query GetFlow($teamId: Int!) {
-          flows(
-            order_by: { updated_at: desc }
-            where: { team: { id: { _eq: $teamId } } }
-          ) {
+          flows(where: { team: { id: { _eq: $teamId } } }) {
             id
             slug
             updatedAt: updated_at
-            operations (
-              limit: 1,
-              order_by: { created_at: desc }
-            ) {
+            operations(limit: 1, order_by: { created_at: desc }) {
               createdAt: created_at
               actor {
                 firstName: first_name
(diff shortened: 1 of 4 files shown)


editor.planx.uk/src/pages/FlowEditor/utils.ts
```diff
@@ -17,4 +17,7 @@ export const formatLastEditMessage = (
 
   const name = `${actor.firstName} ${actor.lastName}`;
   return `Last edited ${formatLastEditDate(date)} by ${name}`;
-};
\ No newline at end of file
+};
+
+export const formatLastPublishMessage = (date: string, user: string): string =>
+  `Last published ${formatLastEditDate(date)} ago by ${user}`;
```
jessicamcinchak, line 23: This is just a small tidy up so that all of the helper methods that rely on `formatDistanceToNow()` are co-located in the same file

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -278,11 +278,15 @@ export const editorStore: StateCreator<
           ) {
             id
             slug
-            updated_at
-            operations(limit: 1, order_by: { id: desc }) {
+            updatedAt: updated_at
```
jessicamcinchak, line 281: Note that `GetFlow` is still ordered by `flows.updated_at`, and can't actually be ordered easily at the GraphQL level by `flows.operations.created_at` - so this is going to appear a bit strange on production! 

Do we want to keep the `updated_at` ordering, switch to alphabetical flow order, or something else (eg "online" flows at top, "offline" flows at bottom)?

editor.planx.uk/src/pages/Team.tsx
```diff
@@ -286,17 +279,28 @@ const Team: React.FC = () => {
...
       .then((res: { flows: any[] }) => {
-        setFlows(res.flows);
+        // Copy the array and sort by most recently edited desc using last associated operation.createdAt, not flow.updatedAt
+        const sortedFlows = res.flows
+          .slice()
+          .sort((a, b) =>
+            b.operations[0]["createdAt"].localeCompare(
+              a.operations[0]["createdAt"],
+            ),
+          );
```
jessicamcinchak, line 295: Previously, we sorted directly via the GraphQL query like `flows (order_by: { updated_at: desc }) {...}` - but we want to sort via operations, not flows, so now we're doing a custom sort here !

RODO94: nit: possibly good use for ToSorted() to copy and sort array with one method

RODO94 (approved):
Lovely code! Added "nit" comment

PR #3249, opened 2024-06-05
feat: add flattened passport structure for List items that is compatible with Calculate

jessicamcinchak:
The basic passport implementation was just returning the list of items as an array. 

This adds an additional "flattened" structure, which will write each individual list item as it's own passport field with a path of only text segments, no numbers, so that fields will be compatible with the Calculate component later if necessary.


jessicamcinchak:

PR #3250, opened 2024-06-06
chore: split residential unit List component schemas into six options

jessicamcinchak:
Schemas based on this flow: https://editor.planx.uk/opensystemslab/residential-units

See thread here: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1717676834944659?thread_ts=1717416055.509189&cid=C5Q59R3HB

Also adjusts summary totals calculated `onSubmit` to include both count of list items _and_ sum of all units.

Eventually, we'll want to calculate totals by tenure and unit type - but we should consider capturing the `val`, rather than `text` in the passport first.

### DIFF
--- a/editor.planx.uk/src/@planx/components/List/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/List/Public/index.tsx
@@ -19,7 +19,6 @@ import CardHeader from "../../shared/Preview/CardHeader";
 import type { Field, List } from "../model";
 import { ListProvider, useListContext } from "./Context";
 import {
-  ChecklistFieldInput,
   NumberFieldInput,
   RadioFieldInput,
   SelectFieldInput,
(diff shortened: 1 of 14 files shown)


editor.planx.uk/src/@planx/components/List/Public/index.tsx
```diff
@@ -181,7 +181,7 @@ const Root = () => {
               sx={{ width: "100%" }}
               data-testid="list-add-button"
             >
-              + Add a new {schema.type.toLowerCase()} description
+              + Add another {schema.type.toLowerCase()} description
```
jessicamcinchak, line 184: One of the GLA schemas has `type: "New built residentail unit"`, so "another" prevents "new new"

jessicamcinchak:

PR #3253, opened 2024-06-07
feat: basic summary of List component on Review page

jessicamcinchak:
**Changes:**
- Adds basic `List` component representation to existing `SummaryList` Review page 
  - Makes "title" & "fn" required in Editor to ensure necessary info is consistently available
- Displays one row _per_ List component
  - If that List's schema includes a field that sets `identicalUnits`, then summary is "X units total"
    - Else default to how many items/responses were provided, eg "X items total"
  - "Change" link takes you back to the List component where you can edit your responses one-by-one

Screenshot from 2024-06-07 13-10-32

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
@@ -338,12 +338,15 @@ interface ComponentProps {
 }
 
 function List(props: ComponentProps) {
+  // `...total.units` is only set when the schema includes a field with fn = `identicalUnits`
   const totalUnits = props.passport.data?.[`${props.node.data.fn}.total.units`];
+  // `...total.listItems` is always set for any schema
   const totalListItems =
     props.passport.data?.[`${props.node.data.fn}.total.listItems`];
+
   const summary = totalUnits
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
```diff
@@ -337,6 +337,22 @@ interface ComponentProps {
...
+  const totalUnits = props.passport.data?.[`${props.node.data.fn}.total.units`];
+  const totalListItems =
+    props.passport.data?.[`${props.node.data.fn}.total.listItems`];
+  const summary = totalUnits
+    ? `${totalUnits} units total`
+    : `${totalListItems} items total`;
+
+  return (
+    <>
+      <Box component="dt">{props.node.data.title}</Box>
```
jessicamcinchak, line 350: this could also be schema "type", but I think node title is most consistent with other components and gives Editors more control here since schemas will be hard-coded for initial launch ?

ianjon3s: Agreed, I think this is the best approach, until proven otherwise

ianjon3s (approved):

PR #3259, opened 2024-06-07
refactor: Standardise error summary

ianjon3s:
## What does this PR do?

- Introduces a standardised error summary with three 'styles'; error, warning and info.
- Updates replicated error summaries
- Introduces error summary for feedback log with no entries
- Storybook story with each permutation 

**Preview:**
https://storybook.3259.planx.pizza/?path=/docs/design-system-molecules-errorsummary--docs

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
@@ -25,6 +25,7 @@ import { client } from "lib/graphql";
 import React, { useState } from "react";
 import { Feedback } from "routes/feedback";
 import EditorRow from "ui/editor/EditorRow";
+import ErrorSummary from "ui/shared/ErrorSummary";
 import ReactMarkdownOrHtml from "ui/shared/ReactMarkdownOrHtml";
 
 interface Props {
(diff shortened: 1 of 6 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
```diff
@@ -164,35 +165,43 @@ export const FeedbackPage: React.FC<Props> = ({ feedback }) => {
...
-                  />
-                ))}
-              </TableBody>
-            </Table>
-          </Feed>
+          {feedback.length === 0 ? (
+            <ErrorSummary
+              format="info"
+              heading="No feedback found for this service"
+              message="If you're looking for feedback before January 1, 2024, please contact a PlanX developer."
```
jessicamcinchak, line 172: nit: in the case of feedback, I think it should read "If you're looking for feedback from more than six months ago, please contact a PlanX developer" - because feedback availability is subject to our data sanitation policy (6 month retention) ! Submission event receipts are always retained (no personal data), and that view is actually filtered from January 2024 forwards for simplicity.

editor.planx.uk/src/ui/shared/ErrorSummary.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+  },
+} satisfies Story;
+
+export const Info = {
+  args: {
+    heading: "Application info",
+    message: "Summary of information message.",
+    format: "info",
+  },
+} satisfies Story;
```
jessicamcinchak, line 36: :100:

editor.planx.uk/src/ui/shared/ErrorSummary.tsx
```diff
@@ -0,0 +1,36 @@
...
+import React from "react";
+
+interface Props {
+  heading?: string;
+  message?: string;
+  format?: "error" | "warning" | "info";
+}
+
+const Root = styled(Box, {
+  shouldForwardProp: (prop) => !["format"].includes(prop.toString()),
```
jessicamcinchak, line 13: the syntax of `shouldForwardProp` is still never quite intuitive to me, but I think this can simply be ??
```
  shouldForwardProp: (prop) => prop !== "format",
```

and then on L14 just below, might need to adjust to `<Props["format"]>` rather than current `<Props>`, as we're just forwarding along the single property?

jessicamcinchak (commented):
Super appreciated update, thanks for picking this one up! 

Beyond these files, there's a couple more that come to mind that I think might good candidates for similar update/refactor to share this new standard component: 
- `.../editor.planx.uk/src/@planx/components/shared/Preview/ErrorSummaryContainer.tsx`
  - Which is then imported/used a couple of places like `ErrorFallback`, `PlanningConstraints` & `PropertyInformation` graph warnings

editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackPage.tsx
ianjon3s: Ah yes, I coped this over from the submission log and had made a note to clarify. Thanks for this, will update.

editor.planx.uk/src/ui/shared/ErrorSummary.tsx
ianjon3s: Thanks, this works without adjusting line 14 👍

ianjon3s:
Thanks for the review @jessicamcinchak , I will follow up with a further PR to update `ErrorSummaryContainer.tsx` with this standard styling

jessicamcinchak (approved):

PR #3261, opened 2024-06-10
feat: Add Proposed Advertisements to List component and schema

augustlindemer:
Added Proposed Advertisements schema and added it to the list component.


editor.planx.uk/src/@planx/components/List/schemas/Adverts.ts
```diff
@@ -0,0 +1,40 @@
...
+    {
+      type: "number",
+      data: {
+        title: "How many other advertisements are you applying for?",
+        fn: "other",
+        allowNegatives: false,
+      },
+    },
+  ],
+  min: 0,
```
jessicamcinchak, line 39: nit: when `min: 0`, the component actually confusing loads like this - so we probably really want `min: 1` ?

Screenshot from 2024-06-10 20-17-04

if we want it to be possible to "continue" without submitting any items, maybe there's a better way we can control that in the Editor modal using a simple toggle rather than in the schema via `min` ?

editor.planx.uk/src/@planx/components/List/schemas/Adverts.ts
```diff
@@ -0,0 +1,40 @@
+import { Schema } from "@planx/components/List/model";
+
+export const ProposedAdvertisements: Schema = {
+  type: "Proposed advertisements",
```
jessicamcinchak, line 4: nit: this piece of text will be read into the title of each item & the "+ Add another ..." button - so it's actually best if it's singular rather than plural !

jessicamcinchak (commented):
Two small things, but looks great ! Thanks for just running with this :checkered_flag:

editor.planx.uk/src/@planx/components/List/schemas/Adverts.ts
augustlindemer: Thanks, I misunderstood what was being set by min and max. Looks like min: 0 isn't even possible, as it throws this when you try to add any item: 
image

PR #3262, opened 2024-06-10
feat: Add Proposed Advertisements to List component and schema

augustlindemer:


jessicamcinchak (approved):
All working exactly as expected now ! Learned a lot of new words in this one :slightly_smiling_face: :gb:

PR #3263, opened 2024-06-10
fix: List should record `val`, not `text`, in passport and calculate total units by development type when applicable

jessicamcinchak:
**Why:**
- For residential units, we need to be able to breakdown the total sum of identical units by their `development` type 
  - MathJS/Calculate can't do list iteration & grouping, so we're going to handle this `onSubmit` for any schemas that set a field with `fn = development` for now
- In order to accomplish this, we also need the passport to record the `val` of the selected option if set, rather than its `text`
  - The `val` will also be useful later when populating the ODP Schema, as it fits our current pattern of looking up a description _from value_ rather than the other way around
  - The inactive list view should still always display the human-readable `text`, even if a `val` is recorded in the passport

**Changes**
- Updates existing schemas to set `val` where expected
- Updates question & select inputs to record `val || text` (val is optional & text is required)
- Adds additional summary statistic for total units by development type when applicable
- Updates tests with "generic units" schema example to help illustrate expected stats


jessicamcinchak:

PR #3264, opened 2024-06-11
feat: Add new GLA schemas for residential units and existing and proposed uses

augustlindemer:
Changed GLA schemas from new/rebuild/retained/lost to gained/lost structure. Added existing and proposed uses schema.

Need to check subsequently as to new summary/count total requirements or if they can be created manually.


jessicamcinchak (approved):
These work as expected for me!

PR #3265, opened 2024-06-11
feat: when List schema sets `max: 1`, don't display add another button or index in card header

jessicamcinchak:
**Changes:** 
- Re-organises all "test" schemas and their mocks for unit tests into a subfolder
- Hides the "+ Add another" & card header index if the schema sets `max: 1` ; always displays when max > 1 or undefined, or if there is no user data (eg max is 1, but first item is "cancelled" before any input, then "removed").

See new "Proposed advertisements" schema as example use case & feedback from August: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1718050782075389?thread_ts=1717416055.509189&cid=C5Q59R3HB

Screenshot from 2024-06-11 11-59-42


jessicamcinchak:

PR #3267, opened 2024-06-11
feat: Add new GLA schemas

augustlindemer:
Includes the following schemas:

- GLA gained & lost structure
- GLA existing and proposed uses 
- GLA building details
- GLA communal spaces
- GLA open spaces
- GLA protected spaces

### DIFF
--- a/editor.planx.uk/src/@planx/components/List/schemas/GLA/CommunalSpace.ts
+++ b/editor.planx.uk/src/@planx/components/List/schemas/GLA/CommunalSpace.ts
@@ -1,0 +1,36 @@
+import { Schema } from "@planx/components/List/model";
+
+export const CommunalSpaceGLA: Schema = {
+  type: "Unit of communal space",
+  fields: [
+    {
+      type: "question",
+      data: {
+        title: "Is this unit of communal space lost or gained?",
+        fn: "development",
+        options: [
+          { id: "gained", data: { text: "Gained", val: "gained" } },
+          { id: "lost", data: { text: "Lost", val: "lost" } },
+        ],
+      },
+    },
+    {
+      type: "number",
+      data: {
+        title: "What is the Gross Internal Floor Area (GIA) of this unit?",
+        units: "m²",
+        fn: "area",
+        allowNegatives: false,
+      },
+    },
+    {
+      type: "number",
+      data: {
+        title: "How many identical units does the description above apply to?",
+        fn: "identicalUnits",
+        allowNegatives: false,
+      },
+    },
+  ],
+  min: 1,
+} as const;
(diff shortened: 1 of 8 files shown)


augustlindemer:
Strange bug: Existing and proposed uses do not allow me to enter any numbers into the GIA entry fields. Can't see any errors with the schema.

editor.planx.uk/src/@planx/components/List/schemas/GLA/CommunalSpace.ts
```diff
@@ -0,0 +1,36 @@
...
+    {
+      type: "question",
+      data: {
+        title: "Is this unit of communal space lost or gained?",
+        fn: "development",
+        options: [
+          { id: "gained", data: { text: "Gained", val: "gained" } },
+          { id: "lost", data: { text: "Lost", val: "lost" } },
+        ],
+      },
```
jessicamcinchak, line 15: my first go at summing identical units by development type was only accounting for the previous three options, but I'll update it to work the same when `gained` & `lost` are options too :+1:

editor.planx.uk/src/@planx/components/List/schemas/GLA/ExistingAndProposedUses.ts
```diff
@@ -0,0 +1,56 @@
...
+          { id: "unknown", data: { text: "Unknown", val: "unknown" } },
+        ],
+      },
+    },
+    {
+      type: "number",
+      data: {
+        title: "What is the existing gross internal floor area?",
+        units: "m²",
+        fn: "area.existing",
```
jessicamcinchak, line 32: I suspect these `.` separated `fn` are the source of your bug! 

For now, I think we need to write all `fn` & `val` _without_ separators - but I'll work on finding a fix to support this / try to wrap my head around why Daf may not have done it from the start ! 

Feel free to note "ideal" `fn` in comment at line ends like this so we don't forget: 
```
fn: "areaExisting", // area.existing
```

editor.planx.uk/src/@planx/components/List/schemas/GLA/CommunalSpace.ts
jessicamcinchak: also nit picky thing: you use `loss` in later schemas, but `lost` here as "development" options - are they actually different in legislation or can we standardise?

augustlindemer: Thanks for spotting. loss/gain is what the data hub has in their technical schema across categories (including resi units), so I updated the lost/gained to loss/gain.

jessicamcinchak (approved):
Looks great !! Thanks for getting all these in :tada:

PR #3268, opened 2024-06-11
fix: make flow name required and update all contexts for passing e2e tests

jessicamcinchak:
These changes work locally for me, hopefully will pass on CI too :crossed_fingers:


RODO94 (approved):
Changes look good to me, thank you for picking this up and spotting the fixes!

PR #3273, opened 2024-06-13
feat: Updating Email Templates with Flow Name

RODO94:
# What does this PR do?

This PR follows on from other PRs which have added the ``name`` column to the ``flows`` table and created functionality for users to add a flow name to their team's flows. 

The final part of this work is to update email templates so the ``Session Name`` is the user generated ``name`` rather than the augmented ``flowSlug``


RODO94:
Changes have been made to ``planx-core`` to enable the use of Name in the email templates.

[planx-core PR](https://github.com/theopensystemslab/planx-core/pull/412)

RODO94:
@theopensystemslab/planx would anyone have advice on how best to test the email templates in the pizza? Any particular flows I could run through?

jessicamcinchak (approved):
Emails all populating correctly for me :email: :heavy_check_mark: 

Lots of files to touch here considering test mocks, but all appears to be working as expected, nicely done!

PR #3275, opened 2024-06-13
fix: preserve editor redirects when logging in

jessicamcinchak:
At first, I suspected this might be at the react-navi level, but our editor routes look consistent with docs [here](https://frontarm.com/navi/en/reference/matchers/#examples-1) & [here](https://www.tabnine.com/code/javascript/classes/navi/NaviRequest).

When we handle the request to `/auth/google`, we were previously setting the `req.session!.returnTo` URL as the referrer which comes through _without_ the `?redirectTo=` param so this explicitly preserves it.

Doesn't seem to break anything in local testing, but wasn't fixing the issue yet either? Opening a pizza to see if any different.

To test, try to open this editor url directly before you're logged in: https://3275.planx.pizza/opensystemslab
- Prior behavior was to login and land on `/`, without being redirected
- Ideal behavior it to login and then be properly redirected to intended path

### DIFF
--- a/api.planx.uk/modules/auth/middleware.ts
+++ b/api.planx.uk/modules/auth/middleware.ts
@@ -111,7 +111,13 @@ export const useJWT = expressjwt({
 });
 
 export const useGoogleAuth: RequestHandler = (req, res, next) => {
-  req.session!.returnTo = req.get("Referrer");
+  // Ensure returnTo maintains `?redirectTo=` param if provided
+  let returnToURL = req.get("Referrer");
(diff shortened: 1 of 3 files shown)


jessicamcinchak:
[Can we fix the redirect when clicking on a pizza link?](https://trello.com/c/Ftu0Vbbt/2922-can-we-fix-the-redirect-when-clicking-on-a-pizza-link)

DafyddLlyr:
Closed in favour of https://github.com/theopensystemslab/planx-new/pull/3302

PR #3277, opened 2024-06-14
refactor: Remove last edited banner from editor

ianjon3s:
## What does this PR do?

The 'last edited' banner seems redundant now we have full edit history available. Removing it also maximises graph canvas area.

**Example:**
https://3277.planx.pizza/buckinghamshire/upload-completion-of-works-notice


jessicamcinchak (approved):
Very much agree with this simplification ! :scissors:

PR #3282, opened 2024-06-18
fix: Allow dot notation of `fn` values in List component schemas

DafyddLlyr:
## What does this PR do?
 - Resolves bug described here - https://github.com/theopensystemslab/planx-new/pull/3267#discussion_r1635090905
 - Updates tests to include `fn` values with dot notation syntax
 - Updates schemas which had commented out `fn` values


editor.planx.uk/src/@planx/components/List/Public/Fields.tsx
```diff
@@ -49,7 +49,7 @@ export const TextFieldInput: React.FC<Props<TextField>> = ({
         rows={
           data.type && ["long", "extraLong"].includes(data.type) ? 5 : undefined
         }
-        name={`userData[${activeIndex}][${data.fn}]`}
+        name={`userData[${activeIndex}]['${data.fn}']`}
```
DafyddLlyr, line 52: This syntax is a bit weird / complex, but it's what Formik requires! Docs: https://formik.org/docs/guides/arrays#avoid-nesting

ianjon3s (approved):
Values have been consistently replaced with `.`

PR #3284, opened 2024-06-18
chore: Make `flow.name` column required

DafyddLlyr:
## What does this PR do?
- Following on from @RODO94's work to add the `flow.name` column, this has now been updated to be a required field in the DB, which matches the associated types and expected behaviour
- Updates a number of tests and functions which required a value for `flow.name`

## Next steps (?)
The reason there's so much "whack-a-mole" with some of these tests is that `flow.name` and `flow.team_id` are now really the compound primary key / unique constraints, not `flow.slug` and `flow.team_id`. A more robust solution may be to make this change in reality in the database, and to correspondingly update mocks to just have a `flow.name` value.

Regression tests passing here - https://github.com/theopensystemslab/planx-new/actions/runs/9565438594 ✅

### DIFF
--- a/sharedb.planx.uk/sharedb-postgresql.js
+++ b/sharedb.planx.uk/sharedb-postgresql.js
@@ -81,8 +81,8 @@ PostgresDB.prototype.commit = function (
 
         client.query("BEGIN", (err) => {
           client.query(
-            "INSERT INTO flows (id, slug) VALUES ($1, $2) ON CONFLICT DO NOTHING",
-            [id, id],
+            "INSERT INTO flows (id, slug, name) VALUES ($1, $2, $3) ON CONFLICT DO NOTHING",
+            [id, id, id],
             (err, _res) => {
               if (err) {
                 rollback(client, done);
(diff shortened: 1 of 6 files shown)


RODO94 (approved):
All looks good from me

sharedb.planx.uk/sharedb-postgresql.js
```diff
@@ -81,8 +81,8 @@ PostgresDB.prototype.commit = function (
 
         client.query("BEGIN", (err) => {
           client.query(
-            "INSERT INTO flows (id, slug) VALUES ($1, $2) ON CONFLICT DO NOTHING",
-            [id, id],
+            "INSERT INTO flows (id, slug, name) VALUES ($1, $2, $3) ON CONFLICT DO NOTHING",
+            [id, id, id],
```
DafyddLlyr, line 85: This was an interesting one - without this I could create a flow, but not update it via ShareDB without an error.

I actually think this like of SQL will never get truly executed (it will _always_ hit the `ON CONFLICT` clause). This is because the flow is created via Hasura prior to any ShareDB interaction takes place. Postgres must just be "clever enough" to know that this query is incorrect without a `name` value, despite it never actually being executed. This is why `id, id, id` is ok as these values are never read.

PR #3286, opened 2024-06-18
fix: Trim URLs to avoid sanitation errors

DafyddLlyr:
## What's the problem?
 - URLs with whitespace at the end trigger sanitation errors. This means that these flows cannot be copied or moved to other teams.
 - Please see https://opensystemslab.slack.com/archives/C5Q59R3HB/p1718717945966979 (OSL Slack) for context
 
 ## What's the solution
 - Calling `.trim()` on the draft URL means this cannot happen, even on copy/paste of a link


ianjon3s (approved):
Nice solution, quick diagnosis and execution!

PR #3303, opened 2024-06-20
feat: Updated editor teams and team selection pages

ianjon3s:
## What does this PR do?

Introduces style and minor structural changes across Teams and Team pages.

Summary of changes:

- Background changed to light
- Council branding added to teams list
- Teams list separated into separate lists based on permissions, teams which a user has editing permissions to will appear first
- Services styled similar to editor portals

In addition to this a `<Dashboard>` component has been created to standardise layout and spacing across editor pages (applied to only these two pages for now). This also contains layout styles that make it ready to drop in the editor in-page navigation elements.

### DIFF
--- a/editor.planx.uk/src/pages/Team.tsx
+++ b/editor.planx.uk/src/pages/Team.tsx
@@ -16,6 +16,7 @@ import Typography from "@mui/material/Typography";
 import React, { useCallback, useEffect, useState } from "react";
 import { Link, useNavigation } from "react-navi";
 import { FONT_WEIGHT_SEMI_BOLD } from "theme";
+import { borderedFocusStyle } from "theme";
 import Dashboard from "ui/editor/Dashboard";
 import { slugify } from "utils";
 
@@ -59,6 +60,9 @@ const DashboardLink = styled(Link)(({ theme }) => ({
   padding: theme.spacing(2),
   margin: 0,
   width: "100%",
+  "&:focus-within": {
+    ...borderedFocusStyle,
+  },
 }));
 
 const StyledSimpleMenu = styled(SimpleMenu)(({ theme }) => ({
@@ -70,21 +74,6 @@ const LinkSubText = styled(Box)(({ theme }) => ({
   color: theme.palette.grey[400],
   fontWeight: "normal",
   paddingTop: "0.5em",
-}));
-
-const HiddenText = styled(Box)(() => ({
-  color: "transparent",
-  position: "absolute",
-  height: "100%",
-  width: "100%",
-  top: 0,
-  left: 0,
-  display: "flex",
-  "&::selection": {
-    background: "hotpink",
-    color: "black",
-    padding: "1em",
-  },
 }));
 
 const Confirm = ({
@@ -205,7 +194,6 @@ const FlowItem: React.FC<FlowItemProps> = ({
           <Typography variant="h4" component="h2">
             {flow.name}
           </Typography>
-          <HiddenText aria-hidden="true">{flow.slug}</HiddenText>
           <LinkSubText>
             {formatLastEditMessage(
               flow.operations[0].createdAt,
--- a/editor.planx.uk/src/pages/Team.tsx
+++ b/editor.planx.uk/src/pages/Team.tsx
@@ -1,10 +1,11 @@
 import { gql } from "@apollo/client";
-import Add from "@mui/icons-material/Add";
+import AddCircleOutlineIcon from "@mui/icons-material/AddCircleOutline";
 import Edit from "@mui/icons-material/Edit";
 import Visibility from "@mui/icons-material/Visibility";
 import Box from "@mui/material/Box";
 import Button from "@mui/material/Button";
 import ButtonBase from "@mui/material/ButtonBase";
+import Container from "@mui/material/Container";
 import Dialog from "@mui/material/Dialog";
 import DialogActions from "@mui/material/DialogActions";
 import DialogContent from "@mui/material/DialogContent";
@@ -12,10 +13,10 @@ import DialogContentText from "@mui/material/DialogContentText";
 import DialogTitle from "@mui/material/DialogTitle";
 import { styled } from "@mui/material/styles";
 import Typography from "@mui/material/Typography";
-import orderBy from "lodash/orderBy";
 import React, { useCallback, useEffect, useState } from "react";
 import { Link, useNavigation } from "react-navi";
 import { FONT_WEIGHT_SEMI_BOLD } from "theme";
+import Dashboard from "ui/editor/Dashboard";
 import { slugify } from "utils";
 
 import { client } from "../lib/graphql";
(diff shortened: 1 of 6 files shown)


DafyddLlyr:
@ianjon3s Don't sweat too much about failing E2E tests here, I'll jump on and take a look once the PR is ready. It's very likely something very simple to do with the content or location of links changing in the DOM. I'll take a look through and maybe you, me and @RODO94 can debug and work through the fixes together 👍

ianjon3s:
@DafyddLlyr I think it may be as simple as me changing a `<h2>` to `<h3>` and not updating tests, I've pushed up a fix for this.

editor.planx.uk/src/routes/authenticated.tsx
```diff
@@ -31,6 +31,11 @@ const editorRoutes = compose(
               name
               slug
             }
+            teamThemes: teams_summary {
+              slug
+              primaryColour: primary_colour
+              logo
+            }
```
DafyddLlyr, line 38: Rather than adding an additional query here, a better pattern (and actually I think the intended one) would be to use the existing relationships to find the corresponding theme details for the relevant team directly in this query (as opposed to having to match them up later using the `.find()` iterator).

Try this out in Hasura - it should be both faster/more efficient and allow us to simplify our code - 

```gql

query GetTeams {
  teams(order_by: {name: asc}) {
    id
    name
    slug
    theme {
      primaryColour: primary_colour
      logo
    }
  }
}
```

editor.planx.uk/src/pages/Teams.tsx
```diff
@@ -1,71 +1,102 @@
...
 
-const StyledLink = styled(Link)(() => ({
-  textDecoration: "none",
-}));
+const Teams: React.FC<Props> = ({ teams, teamTheme }) => {
+  const canUserEditTeam = useStore.getState().canUserEditTeam;
 
-const Teams: React.FC<Props> = ({ teams }) => {
+  const editableTeams = teams.filter((team) => canUserEditTeam(team.slug));
+  const viewOnlyTeams = teams.filter((team) => !canUserEditTeam(team.slug));
```
DafyddLlyr, line 57: nit: You could also do this in a single iteration with a `.reduce()` or `.forEach()`

**reduce()**
```ts
const { editableTeams, viewOnlyTeams } = teams.reduce(
  (acc, team) => {
    canUserEditTeam(team.slug)
      ? acc.editableTeams.push(team)
      : acc.viewOnlyTeams.push(team)
    
    return acc;
  },
  { editableTeams: [], viewOnlyTeams: [] }
);
```

**forEach()**
```ts
const editableTeams = [];
const viewOnlyTeams = [];

teams.forEach((team) => (
  canUserEditTeam(team.slug)
    ? editableTeams.push(team)
    : viewOnlyTeams.push(team)
));
```

editor.planx.uk/src/pages/Teams.tsx
```diff
@@ -1,71 +1,102 @@
...
+
+  const renderTeams = (teamsToRender: Array<Team>) =>
+    teamsToRender.map(({ name, slug }) => {
+      const theme = teamTheme.find((t) => t.slug === slug);
+      const primaryColour = theme ? theme.primaryColour : "#000";
```
DafyddLlyr, line 62: `team_theme.primary_colour` is a required field (with a default value) in the database so we should be fine to not have a default here. 

It's possible that this is required to get around some TypeScript issues, but this should be resolved by improving the query and dropping the `find()` (see previous comment).

editor.planx.uk/src/ui/editor/Dashboard.tsx
```diff
@@ -0,0 +1,21 @@
...
+import React, { PropsWithChildren } from "react";
+
+const Root = styled(Box)(({ theme }) => ({
+  backgroundColor: theme.palette.background.default,
+  color: theme.palette.text.primary,
+  display: "flex",
+  flexDirection: "row",
+  width: "100%",
+  minHeight: `calc(100vh - ${HEADER_HEIGHT}px)`,
+  "& > .MuiContainer-root": {
```
DafyddLlyr, line 13: I think we should be able to import this classname from MUI 👍

editor.planx.uk/src/pages/Team.tsx
```diff
@@ -191,17 +201,18 @@ const FlowItem: React.FC<FlowItemProps> = ({
...
       )}
       <DashboardListItem>
-        <Box pr={4}>
-          <DashboardLink href={`./${flow.slug}`} prefetch={false}>
+        <DashboardLink href={`./${flow.slug}`} prefetch={false}>
+          <Typography variant="h4" component="h2">
             {flow.name}
-          </DashboardLink>
+          </Typography>
+          <HiddenText aria-hidden="true">{flow.slug}</HiddenText>
```
DafyddLlyr, line 208: I'm not too sure how much this is really needed? I know it's helpful as a pseudo-search until we have this feature but we can still `CTRL+F` on flow names?

DafyddLlyr (changes requested):
A few small comments on, visually looking good though.

This isn't introduced here, but the missing focus styles on the `DashboardListItem`s are a bit more noticeable now. This can be addressed now or later (this is fairly inconsistent across the Editor tbh).


https://github.com/theopensystemslab/planx-new/assets/20502206/33e47318-93a0-4658-8b4d-a9786aa75df9

DafyddLlyr:
(Also good work on solving the E2E test issues 🙌)

editor.planx.uk/src/pages/Teams.tsx
ianjon3s: Thanks for this, I've gone with `forEach` as it reads better as a piece of logic.

editor.planx.uk/src/pages/Teams.tsx
ianjon3s: I'm now calling this as `team.theme.primaryColour` without a default value

editor.planx.uk/src/ui/editor/Dashboard.tsx
ianjon3s: Good spot, applied

editor.planx.uk/src/pages/Team.tsx
ianjon3s: Agreed, I've pulled it out. I added as I was getting frustrated when using `CTRL+F` and typing in a slug, but it's just a case of waiting for new features to be in place.

DafyddLlyr (approved):
Looking good 👍 

Before merging this, let's get a prod deploy done first so we can batch a few Editor changes together. There's a PR here to approve which will let us clear the decks a little first - https://github.com/theopensystemslab/planx-new/pull/3310

ianjon3s:
Thanks for the detailed review and suggestions @DafyddLlyr. I've also added focus outline style to the team cards on initial screen.

DafyddLlyr:
Thanks @ianjon3s, prod deploy underway so please feel free to merge this 👍

PR #3304, opened 2024-06-20
feat: `ChecklistFieldInput` for List component

DafyddLlyr:
## What does this PR do?
- Introduces a `ChecklistField` for the List component
  - As with our other fields, this requires making the validation schema shared
- Updates existing schemas to use checklists instead of a series of questions
- Updates tests
- Update `flatten()` logic to handle checklist responses

image

### DIFF
--- a/editor.planx.uk/src/@planx/components/List/Public/Context.tsx
+++ b/editor.planx.uk/src/@planx/components/List/Public/Context.tsx
@@ -171,7 +171,7 @@ export const ListProvider: React.FC<ListProviderProps> = (props) => {
       const defaultPassportData = makeData(props, values.userData)?.["data"];
 
       // flattenedPassportData makes individual list items compatible with Calculate components
-      const flattenedPassportData = flatten(defaultPassportData);
+      const flattenedPassportData = flatten(defaultPassportData, { depth: 2 });
 
       // basic example of general summary stats we can add onSubmit:
       //   1. count of items/responses
--- a/editor.planx.uk/src/@planx/components/List/utils.tsx
+++ b/editor.planx.uk/src/@planx/components/List/utils.tsx
@@ -1,24 +1,40 @@
-import { QuestionField, Schema, UserResponse } from "./model";
+import React from "react";
+
+import { Field, UserResponse } from "./model";
 
 /**
- * In the case of "question" fields, ensure the displayed value reflects option "text", rather than "val" as recorded in passport
+ * In the case of "question" and "checklist" fields, ensure the displayed value reflects option "text", rather than "val" as recorded in passport
  * @param value - the `val` or `text` of an Option defined in the schema's fields
- * @param schema - the Schema object
- * @returns string - the `text` for the given value `val`, or the original value
+ * @param field - the Field object
+ * @returns string | React.JSX.Element - the `text` for the given value `val`, or the original value
  */
-export function formatSchemaDisplayValue(value: string | string[], schema: Schema) {
-  const questionFields = schema.fields.filter(
-    (field) => field.type === "question",
-  ) as QuestionField[];
-  const matchingField = questionFields?.find((field) =>
-    field.data.options.some((option) => option.data.val === value),
-  );
-  const matchingOption = matchingField?.data.options.find(
-    (option) => option.data.val === value,
-  );
-
-  // If we found a "val" match, return its text, else just return the value as passed in
-  return matchingOption?.data?.text || value;
+export function formatSchemaDisplayValue(
+  value: string | string[],
+  field: Field,
+) {
+  switch (field.type) {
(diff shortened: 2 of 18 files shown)


editor.planx.uk/src/@planx/components/List/utils.tsx
```diff
@@ -1,24 +1,40 @@
...
-  );
-  const matchingOption = matchingField?.data.options.find(
-    (option) => option.data.val === value,
-  );
-
-  // If we found a "val" match, return its text, else just return the value as passed in
-  return matchingOption?.data?.text || value;
+export function formatSchemaDisplayValue(
+  value: string | string[],
+  field: Field,
```
DafyddLlyr, line 13: Passing in the `Field` as opposed to the `Schema` allows us to simplify how we find the matching fields.

editor.planx.uk/src/@planx/components/List/utils.tsx
```diff
@@ -74,14 +95,19 @@ export function sumIdenticalUnitsByDevelopmentType(
...
 /**
  * Flattens nested object so we can output passport variables like `{listFn}.{itemIndexAsText}.{fieldFn}`
  *   Adapted from https://gist.github.com/penguinboy/762197
  */
 export function flatten<T extends Record<string, any>>(
-  object: T,
-  path: string | null = null,
-  separator = ".",
+    object: T,
+    { depth = Infinity, path = null, separator = ".", }: FlattenOptions = {}
```
DafyddLlyr, line 110: Without this change (adding `depth` argument), this will output keys in for format `${componentFn}.${number}.${fieldFn}.{$number}` via the existing recursive approach, For example - 

```json
{
  "residentialUnits.one.compliance.one": "m42",
  "residentialUnits.one.compliance.two": "m42a",
  "residentialUnits.one.compliance.three": "m42b",
}
```

I think the right approach isto output an array of responses, as we do with other checklists.This results in output formatted like this (see tests also) - 

```json
{
  "residentialUnits.one.compliance": ["m42", "m42a", "m42b"]
}
```

ianjon3s (approved):
Checklist outputs with correct a11y markup and functions correctly on the front-end, with the exception of the bug addressed in https://github.com/theopensystemslab/planx-new/pull/3316

editor.planx.uk/src/@planx/components/List/Public/Context.tsx
```diff
@@ -168,7 +168,7 @@ export const ListProvider: React.FC<ListProviderProps> = (props) => {
       const defaultPassportData = makeData(props, values.userData)?.["data"];
 
       // flattenedPassportData makes individual list items compatible with Calculate components
-      const flattenedPassportData = flatten(defaultPassportData);
+      const flattenedPassportData = flatten(defaultPassportData, { depth: 2 });
```
jessicamcinchak, line 171: :+1:

editor.planx.uk/src/@planx/components/List/utils.tsx
```diff
@@ -1,24 +1,40 @@
...
+    case "checklist": {
+      const matchingOptions = field.data.options.filter((option) =>
+        (value as string[]).includes(option.id),
+      );
+      return (
+        <ul>
+          {matchingOptions.map((option) => (
+            <li key={option.id}>{option.data.text}</li>
+          ))}
+        </ul>
```
jessicamcinchak, line 28: nit: can we slightly adjust padding/margin here so bullet points are aligned to right text rather than indented? 
Screenshot from 2024-06-27 12-43-26

I think that would be more consistent with existing SummaryList/Review page spacing

editor.planx.uk/src/@planx/components/List/utils.tsx
```diff
@@ -32,7 +48,11 @@ export function sumIdenticalUnits(
   passportData: Record<string, UserResponse[]>,
 ): number {
   let sum = 0;
-  passportData[`${fn}`].map((item) => (sum += parseInt(item?.identicalUnits)));
+  passportData[`${fn}`].map((item) => {
+    if (!Array.isArray(item?.identicalUnits)) {
+      sum += parseInt(item?.identicalUnits);
+    }
+  });
```
jessicamcinchak, line 55: :+1:

jessicamcinchak (approved):
Works exactly as I'd expect it to & records a list of my selected values in the passport !

Really nice List extension/addition :slightly_smiling_face:

editor.planx.uk/src/@planx/components/List/utils.tsx
DafyddLlyr: Great catch, updated styles to match 👍

jessicamcinchak (approved):

PR #3306, opened 2024-06-20
fix: Image upload stretch and design settings

ianjon3s:
## What does this PR do?

Quick one, addresses small issues with image upload and design settings layout.

### Image upload

Previously the uploaded image was being stretched which makes it hard to decipher the image if it is not square:

image

### Design settings

Having the preview left-aligned means you can't preview what is being applied when the colour picker is open, centering helps with this:

image


DafyddLlyr (approved):
Nice!

PR #3305, opened 2024-06-20
feat: Team Settings Editor Form UI Changes

RODO94:
## What is this PR?

Following on from previous PR: #3289 -- which added a new ``team_settings`` table to the database, this PR makes the relevant UI and Editor changes to interact with the new table. In addition, a new settings form is made so users can manage their own team settings.

- [x] Remove "commented out" validation on TeamSettings route view

PR #3308, opened 2024-06-20
chore: Make `SettingsForm` shared and generic

DafyddLlyr:
## What does this PR do?
 - Moves `SettingsForm` to a shared location so that it can be used outside the context of design settings
 - Updates props to become generic, meaning we could pass in a form other than `TeamTheme`
 - No functional changes

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/shared/SettingsForm.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/shared/SettingsForm.tsx
@@ -1,0 +1,65 @@
+import Box from "@mui/material/Box";
+import Button from "@mui/material/Button";
+import Typography from "@mui/material/Typography";
+import { FormikProps } from "formik";
+import React from "react";
+import EditorRow from "ui/editor/EditorRow";
+import InputGroup from "ui/editor/InputGroup";
+import InputLegend from "ui/editor/InputLegend";
+import ErrorWrapper from "ui/shared/ErrorWrapper";
+
+type SettingsFormProps<TFormikValues> = {
+  legend: string;
+  description: React.ReactElement;
+  input: React.ReactElement;
+  formik: FormikProps<TFormikValues>;
+  preview?: React.ReactElement;
+};
+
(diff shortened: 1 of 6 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Settings/shared/SettingsForm.tsx
```diff
@@ -0,0 +1,65 @@
...
+import Button from "@mui/material/Button";
+import Typography from "@mui/material/Typography";
+import { FormikProps } from "formik";
+import React from "react";
+import EditorRow from "ui/editor/EditorRow";
+import InputGroup from "ui/editor/InputGroup";
+import InputLegend from "ui/editor/InputLegend";
+import ErrorWrapper from "ui/shared/ErrorWrapper";
+
+type SettingsFormProps<TFormikValues> = {
```
DafyddLlyr, line 11: Here's the core difference in this PR.

The previous props took a hardcoded `TeamTheme`, we're now using a [generic type](https://www.typescriptlang.org/docs/handbook/2/generics.html) here.

editor.planx.uk/src/pages/FlowEditor/components/Settings/shared/SettingsForm.tsx
```diff
@@ -0,0 +1,65 @@
...
+  legend: string;
+  description: React.ReactElement;
+  input: React.ReactElement;
+  formik: FormikProps<TFormikValues>;
+  preview?: React.ReactElement;
+};
+
+export const SettingsForm = <TFormikValues,>({
```
DafyddLlyr, line 19: Apart from changes to types and an update of the `ErrorWrapper` id, this component is copy/pasted from `/DesignSettings/index.tsx`

RODO94 (approved):
Looks good to me, I can apply it to the Team Settings PR #3305 and suggest changes if any arise

DafyddLlyr:
@RODO94 Once CI is done I'll merge to `main` and you should be able to rebase to pick up these changes 👍

PR #3311, opened 2024-06-24
fix: Update "cancel" functionality for List component

DafyddLlyr:


RODO94:
@DafyddLlyr How do I test the Page component proxies? Is there space to set maximum number of listed items?

DafyddLlyr:
> @DafyddLlyr How do I test the Page component proxies? Is there space to set maximum number of listed items?

Great question! There's currently a schema called "Adverts" which will allow you to test this out.

RODO94:
> > @DafyddLlyr How do I test the Page component proxies? Is there space to set maximum number of listed items?
> 
> Great question! There's currently a schema called "Adverts" which will allow you to test this out.

Aha! And you do ``max=1`` in the Data Field input?

RODO94 (approved):
Tests well with the situations mentioned, looks good to me

DafyddLlyr:
Thanks for the review @RODO94. Going to also await feedback from @ianjon3s here before merging as the functionality is slightly different to the approach he originally outlined.

ianjon3s (approved):
This is working great @DafyddLlyr, with cancel being an action to revert the user behaviour of "I wish to edit a saved list item" or "I wish to add a new item to the list". Ideally we'd not have cancel available on initial load as the user hasn't taken any actions that can be cancelled, but I'm happy to pick that up in a separate PR.

PR #3313, opened 2024-06-25
chore: Run `pnpm lint:fix` in Editor

DafyddLlyr:
Noticed a few of these as warnings when running the Editor, a few things that Husky seems to have missed.


RODO94 (approved):
I'm happy to bow down to the lint gospel here... all looks good, no accidental typos or issues with the new formatting

PR #3314, opened 2024-06-25
chore: Remove ``has_planning_data`` from team_settings table

RODO94:
## What does this PR do?

Following #3305, it was found that the ``has_planning_data`` column of the "team_settings" table is redundant and should be removed.

This PR creates the necessary migrations and table permission changes to remove the column


DafyddLlyr (approved):
I think we also want to remove `isPlanningDataCollected` from `GeneralSettings` which was the placeholder for this. Happy for this to be handled in this PR or elsewhere 👍

RODO94:
> I think we also want to remove `isPlanningDataCollected` from `GeneralSettings` which was the placeholder for this. Happy for this to be handled in this PR or elsewhere 👍

@DafyddLlyr in this part of the ``index.tsx``?

 ```ts
 export interface GeneralSettings{...}
 ```

DafyddLlyr (approved):

RODO94:

PR #3316, opened 2024-06-25
fix: Prevent event bubbling on click of checkbox

DafyddLlyr:
## What's the problem?
I hit some strange checkbox behaviour when working on https://github.com/theopensystemslab/planx-new/pull/3304 which sidetracked me somewhat! See video below for demo - 

https://github.com/theopensystemslab/planx-new/assets/20502206/47b90a82-ebf6-477d-9a63-184fb7879cd6

What's happening is - 
 - Clicking labels behaves as expected
 - Clicking checkbox fires event (`onChange()` function) twice, resulting in nondeterministic updates when combined with async state updates (via Formik)

We can check this out by using a `console.log()` inside our `onChange()` function and seeing that it fires multiple times (event outside the scope of React's strict mode). Another good example of this, which is now fixed in this PR, is that selecting checkboxes in the Editor's send component triggers an alert twice on staging, but only once on this PR.

## What's the solution?
Use `event.preventDefault()` to stop the event bubbling up further. Docs: https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault

## To test...
Let's make sure this hasn't broken anything!

 - I can check public facing checklists via...
   - Click on checkbox
   - Click on label
   - Keyboard navigation

 - I can check Editor facing checklists (e.g. Send components) via...
   - Click on checkbox
   - Click on label
   - Keyboard navigation
   - And events only fire once (single alert popup)

### DIFF
--- a/editor.planx.uk/src/@planx/components/Checklist/Public.test.tsx
+++ b/editor.planx.uk/src/@planx/components/Checklist/Public.test.tsx
@@ -173,6 +173,7 @@ describe("Checklist Component - Grouped Layout", () => {
       answers: ["S1_Option1", "S3_Option1"],
     });
   });
+
   it("should not have any accessibility violations", async () => {
     const { container } = setup(
       <Checklist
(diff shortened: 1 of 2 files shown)


ianjon3s (approved):
Checklists functioning correctly on public interface (Checklist component) and editor (Send component)

PR #3319, opened 2024-06-25
feat: fetching team_settings data from database and populating forms

RODO94:
## What does this PR do?

Continuing work on the ``team_settings`` implementation, this PR pulls in data from the database with a graphQL query and populates the corresponding form inputs.

This PR also uses the update mutation, similar to `team_themes` to update the form fields and database. 

This PR aligns with a planx-core change where the requests and types for a Team were altered to incorporate the new ``team_settings`` table.

[Reference to the planx-core changes](https://github.com/theopensystemslab/planx-core/pull/424)

> Important to note that this PR has not deleted any functions, methods, or types related to previous versions of `notifyPersonalisation` or the original ``settings`` column of _"teams"_ table. Thinking is to do this clear out after changes have been made and tested using the new ``team_settings`` table

### DIFF
--- a/editor.planx.uk/src/components/Header.test.tsx
+++ b/editor.planx.uk/src/components/Header.test.tsx
@@ -85,7 +85,7 @@ describe("Header Component - Editor Route", () => {
           id: 123,
           email: "test@example.com",
         },
-      }),
+      })
     );
 
     jest.spyOn(ReactNavi, "useCurrentRoute").mockImplementation(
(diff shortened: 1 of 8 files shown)


editor.planx.uk/src/components/Header.test.tsx
DafyddLlyr, line 1: The changes in this file (removed commas) don't match our linting style. Husky should be picking this up and fixing it as part of a pre-commit hook but this clearly isn't working right atm for your setup it seems. We can also look at checking this at the CI stage if we need to.

You can fix this file with `pnpm lint:fix`

RODO94:
PR Closed as overlapping changes became too tangled, changes have been moved to fresh branch:

https://github.com/theopensystemslab/planx-new/pull/3366

PR #3321, opened 2024-06-26
test: Fix skipped List component test

DafyddLlyr:
## What's the problem?
 - There was a skipped test introduced in #3263 
 - This was failing when un-skipped, as react-testing-library was unable to locate DOM elements to interact with

## What's the solution?
Variables need be be reinstated when a new list item is added, otherwise they hold onto their old references and are invalid.

### DIFF
--- a/editor.planx.uk/src/@planx/components/List/schemas/mocks/GenericUnits.ts
+++ b/editor.planx.uk/src/@planx/components/List/schemas/mocks/GenericUnits.ts
@@ -89,7 +89,7 @@ export const mockUnitsPayload = {
     "proposal.units.residential.three.identicalUnits": 2,
     "proposal.units.residential.total.listItems": 3,
     "proposal.units.residential.total.units": 5,
-    "proposal.units.residential.total.units.newBuid": 3,
-    "proposal.units.residential.total.units.changeOfUseTo": 2,
+    "proposal.units.residential.total.units.development.newBuild": 3,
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/@planx/components/List/schemas/mocks/GenericUnits.ts
```diff
@@ -89,7 +89,7 @@ export const mockUnitsPayload = {
     "proposal.units.residential.three.identicalUnits": 2,
     "proposal.units.residential.total.listItems": 3,
     "proposal.units.residential.total.units": 5,
-    "proposal.units.residential.total.units.newBuid": 3,
-    "proposal.units.residential.total.units.changeOfUseTo": 2,
+    "proposal.units.residential.total.units.development.newBuild": 3,
+    "proposal.units.residential.total.units.development.changeOfUseTo": 2,
```
DafyddLlyr, line 93: This now matches the other test example here - 

https://github.com/theopensystemslab/planx-new/blob/be6a176da39bff13cef8164b9b4468f75e0417b3/editor.planx.uk/src/%40planx/components/List/utils.test.ts#L73-L74

`development` is added to the path here - 

https://github.com/theopensystemslab/planx-new/blob/be6a176da39bff13cef8164b9b4468f75e0417b3/editor.planx.uk/src/%40planx/components/List/utils.ts#L68-L72

RODO94 (approved):
Changes look good to me and relates well to the PR description.

PR #3322, opened 2024-06-26
fix: List component radio remove extra margin

ianjon3s:
## What does this PR do?

Quick one: adds a negative margin to `</RadioGroup>` in the List component to negate the extra margin caused by using this inline (radio button rows have margin-bottom to ensure layout spacing).

**Before vs after:**

image


DafyddLlyr (approved):
Super nice and easy, thanks for picking this one up! 

Been bugging me but hadn't yet turned my attention to it 🙌

PR #3330, opened 2024-06-27
fix: Trim all form values on node creation

DafyddLlyr:
Raised as an issue here - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1719495568931289 (OSL Slack)

Currently, this works on node update via `sanitize()` within the `_update()` function which handled operations. `sanitise()` is not set up in quite the same way for add node, so I'm handling this at the `parseFormData()` level.


jessicamcinchak (approved):
Good catch, tricky one - this works for me !

PR #3333, opened 2024-06-28
fix: Make `val` optional in SetValue component

DafyddLlyr:
## What does this PR do?
 - Makes `val` an optional field in the SetValue component if the operation is `removeAll`
 - Updates types
 - Conditionally hides form field

image


DafyddLlyr:
[Should 'remove all' in the Set Value component still make you add a specific value?](https://trello.com/c/0ZzHg6v6/2936-should-remove-all-in-the-set-value-component-still-make-you-add-a-specific-value)

jessicamcinchak (approved):
Perfect :+1:

PR #3334, opened 2024-06-28
feat: Repointing data fetches from ``Teams`` table to ``team_settings`` table

RODO94:
## What does this PR do?

This PR follows work done on the ``team_settings`` table implementation. This PR focuses on work to change references from the ``settings`` and ``notifyPersonalisation`` columns to the new ``team_settings`` table which replace these columns.

Changes have been made to the ``planx-core`` repo as well to coincide with these changes, contained here: https://github.com/theopensystemslab/planx-core/pull/424


editor.planx.uk/package.json
```diff
@@ -12,8 +12,8 @@
     "@mui/lab": "5.0.0-alpha.170",
     "@mui/material": "^5.15.2",
     "@mui/utils": "^5.15.2",
-    "@opensystemslab/map": "^0.8.3",
-    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#b975cf9",
+    "@opensystemslab/map": "^0.8.2",
```
DafyddLlyr: This looks like a mistake when rebasing - we want to maintain the more recent 0.8.3 (matching `main`, and updated here - https://github.com/theopensystemslab/planx-new/pull/3337)

RODO94: Sorted! Cheers

editor.planx.uk/src/routes/views/draft.tsx
```diff
@@ -69,12 +69,18 @@ const fetchSettingsForDraftView = async (
                 favicon
               }
               name
-              settings
+              teamSettings: team_settings {
```
DafyddLlyr, line 72:
```suggestion
              settings: team_settings {
```

nit: I think we maintain this as just `settings`

This means we can just use `team.settings` instead of the more verbose and less meaningful `team.teamSettings`.

editor.planx.uk/src/ui/public/ExternalPlanningSiteDialog.tsx
```diff
@@ -53,7 +53,7 @@ const getTitleAndContent = (
 
 interface Props {
   purpose: DialogPurpose;
-  teamSettings?: TeamSettings;
+  teamSettings?: Partial<TeamSettings>;
```
DafyddLlyr, line 56: nit: It actually looks like this prop is never used (please check this!) - this means we should remove this 👍 The settings are fetched via the store rather than being passed in as props.

DafyddLlyr (changes requested):
A few small comment / suggestions, all very minor 👍

editor.planx.uk/src/ui/public/ExternalPlanningSiteDialog.tsx
```diff
@@ -53,7 +53,6 @@ const getTitleAndContent = (
 
 interface Props {
   purpose: DialogPurpose;
-  teamSettings?: Partial<TeamSettings>;
```
DafyddLlyr: It looks like there's a TS issue related to this change causing the built to fail - 
 - "Find all references" on the component and manually check
 - Run `pnpm check` to see if there's any errors
 
To be fair this second one should be running on a pre-commit hook (using Husky) which should block breaking commits. Husky is a bit awkward in mono-repos so I'll take a look at this again to see if we can improve this.

editor.planx.uk/src/routes/views/draft.tsx
```diff
@@ -69,7 +69,7 @@ const fetchSettingsForDraftView = async (
                 favicon
               }
               name
-              teamSettings: team_settings {
+              settings: team_settings {
```
DafyddLlyr, line 72: I'd expect these changes to also need to be reflected in the store? Take a look at where `team.teamSettings` is being referred to (`setTeam()` I think) 👍

DafyddLlyr (changes requested):

editor.planx.uk/src/routes/views/draft.tsx
DafyddLlyr: Ah @RODO94 sorry - just realised this is typed as `Team.teamSettings` in planx-core and I missed it there when reviewing. Please ignore https://github.com/theopensystemslab/planx-new/pull/3334#discussion_r1661227147 for now in the name of expediency and we can take another look at this later.

RODO94: Yeah, maybe for another PR?

DafyddLlyr (approved):

RODO94:

PR #3366, opened 2024-07-03
feat: Fetch `team_settings` data for Editor forms and adding Update fn

RODO94:
## What does this PR do?

_This is a cleaner branch from the work undertaken on this previous branch: #3319_ 

Continuing work on the ``team_settings`` implementation, this PR pulls in data from the database with a graphQL query and populates the corresponding form inputs.

This PR also uses the update mutation, similar to `team_themes` to update the form fields and database. 

This PR aligns with a planx-core change where the requests and types for a Team were altered to incorporate the new ``team_settings`` table.

[Reference to the planx-core changes](https://github.com/theopensystemslab/planx-core/pull/424)

> Important to note that this PR has not deleted any functions, methods, or types related to previous versions of `notifyPersonalisation` or the original ``settings`` column of _"teams"_ table. Thinking is to do this clear out after changes have been made and tested using the new ``team_settings`` table

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/BoundaryForm.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/BoundaryForm.tsx
@@ -33,6 +33,7 @@ export default function BoundaryForm({ formikConfig, onSuccess }: FormProps) {
     <SettingsForm
       formik={formik}
       legend="Boundary"
+      isErrors={false}
       description={
         <>
           <p>
@@ -53,16 +54,21 @@ export default function BoundaryForm({ formikConfig, onSuccess }: FormProps) {
         </>
       }
       input={
-        <InputLabel label="Boundary URL" htmlFor="boundaryUrl">
-          <Input
-            name="boundary"
-            value={formik.values.boundaryUrl}
-            onChange={(ev: ChangeEvent<HTMLInputElement>) => {
-              formik.setFieldValue("boundaryUrl", ev.target.value);
-            }}
-            id="boundaryUrl"
-          />
-        </InputLabel>
+        <ErrorWrapper
+          error={Object.values(formik.errors).join(", ")}
+          id="settings-error"
+        >
+          <InputLabel label="Boundary URL" htmlFor="boundaryUrl">
+            <Input
+              name="boundary"
+              value={formik.values.boundaryUrl}
+              onChange={(ev: ChangeEvent<HTMLInputElement>) => {
+                formik.setFieldValue("boundaryUrl", ev.target.value);
+              }}
+              id="boundaryUrl"
+            />
+          </InputLabel>
+        </ErrorWrapper>
       }
     />
   );
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/BoundaryForm.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/BoundaryForm.tsx
@@ -1,4 +1,5 @@
 import { useFormik } from "formik";
+import { useStore } from "pages/FlowEditor/lib/store";
 import React, { ChangeEvent } from "react";
 import InputLabel from "ui/editor/InputLabel";
 import Input from "ui/shared/Input";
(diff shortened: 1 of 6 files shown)


editor.planx.uk/src/routes/teamSettings.tsx
```diff
@@ -47,11 +47,11 @@ const teamSettingsRoutes = compose(
...
-              // {
-              //   name: "General",
-              //   route: "general",
-              //   Component: GeneralSettings,
-              // },
+              {
+                name: "General",
+                route: "general",
+                Component: GeneralSettings,
+              },
```
DafyddLlyr, line 54: nit: I think I'd reorder this array to be General → Design → Team

editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/ContactForm.tsx
```diff
@@ -20,9 +21,16 @@ export default function ContactForm({ formikConfig, onSuccess }: FormProps) {
...
     ...formikConfig,
     validationSchema: formSchema,
-    onSubmit(values, { resetForm }) {
-      onSuccess();
-      resetForm({ values });
+    onSubmit: async (values, { resetForm }) => {
+      const isSuccess = await useStore.getState().updateTeamSettings({
+        helpEmail: values.helpEmail,
+        helpOpeningHours: values.helpOpeningHours,
+        helpPhone: values.helpPhone,
```
DafyddLlyr, line 28: `homepage` is missing here which means that changes to the URL are not persisted.

editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/BoundaryForm.tsx
```diff
@@ -9,9 +10,14 @@ import { FormProps } from ".";
 export default function BoundaryForm({ formikConfig, onSuccess }: FormProps) {
   const formik = useFormik({
     ...formikConfig,
-    onSubmit(values, { resetForm }) {
-      onSuccess();
-      resetForm({ values });
+    onSubmit: async (values, { resetForm }) => {
```
DafyddLlyr, line 13: We're still missing some basic validation on this field. I appreciate there's still work to be done on the boundary field, but if we're going to unhide the menu and allow Editors to access it we should either put this in place first, or hide this field.

image

DafyddLlyr (changes requested):
Looking good - a lot simpler and cleaner following those naming updates 👍 

There's a few outstanding issues to address, please see comments. 

Also, I'd take a look at co-locating errors alongside their inputs. This will lead to a better user experience, and is required for accessibility reasons. This can be addressed here or in a follow up PR - whatever works best for you.

image

editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/BoundaryForm.tsx
RODO94: I'll do the error handling in another PR, hopefully make it a lot cleaner. Current ones were just to test my understanding of formik and yup

editor.planx.uk/src/routes/teamSettings.tsx
RODO94: @DafyddLlyr do you think it is worth doing these types of changes as a final PR when error handling is all sorted?

RODO94: Thought about it again, and tit doesn't really make it difference, so I've added it to this PR

editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/BoundaryForm.tsx
RODO94: I have added basic validation now though

DafyddLlyr (approved):
Perfect! Looking great 😄 ✅ 

Agree that validation in the next PR will be the best way to go here 👍

RODO94:
> Perfect! Looking great 😄 ✅
> 
> Agree that validation in the next PR will be the best way to go here 👍

Maybe I should move that last commit to a new PR then? @DafyddLlyr

DafyddLlyr:
Sorry for not being clear, I think this should merge as-is, and we can tackle per-field validation next in a follow up PR.

RODO94:

RODO94:

PR #3367, opened 2024-07-03
chore: show number units on inactive List if defined

jessicamcinchak:
A small thing spotted during demo this morning!

Since many schemas are using a Number input to collect area, we should ensure the `m2` units is consistently displayed on _both_ the active & inactive List when it's defined. 

Screenshot from 2024-07-03 15-43-27

### DIFF
--- a/editor.planx.uk/src/@planx/components/List/utils.tsx
+++ b/editor.planx.uk/src/@planx/components/List/utils.tsx
@@ -1,13 +1,14 @@
+import { styled } from "@mui/material/styles";
 import React from "react";
+import { optional } from "zod";
 
 import { Field, UserResponse } from "./model";
-import { styled } from "@mui/material/styles";
 
 const List = styled("ul")(() => ({
   listStylePosition: "inside",
   padding: 0,
   margin: 0,
-}))
+}));
 
 /**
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/@planx/components/List/utils.tsx
```diff
@@ -21,6 +22,7 @@ export function formatSchemaDisplayValue(
 ) {
   switch (field.type) {
     case "number":
+      return field.data.units ? `${value} ${field.data.units}` : value;
```
jessicamcinchak, line 25: This is my only change ! 

Rest are formatting only, I guess prettier has changed with recent dependabot updates? Sorry for extra noise.

DafyddLlyr (approved):
Very nice spot! 🦅 👀

editor.planx.uk/src/@planx/components/List/utils.tsx
DafyddLlyr: Ah yep that makes sense - I'll open a `pnpn lint:fix` PR to fix this in a oner and try to remember this next time there's a prettier update 👍

DafyddLlyr: PR here - https://github.com/theopensystemslab/planx-new/pull/3372

PR #3368, opened 2024-07-03
feat: `team_settings` form error validation

RODO94:
# What does this PR do?

Following previous work done on `team_settings` form in the Editor: https://github.com/theopensystemslab/planx-new/pull/3366

This PR will introduce the Error Handling while making changes to shared `SettingsForm` to handle error messaging on an input level rather than form level.

The only change to an _Input Component_ will be to the ``<ColourPicker>`` which is used in the ``Theme`` settings forms. The prop ``errorMessage`` has been added to the ``<ColourPicker>`` component to enable this input level error messaging

### DIFF
--- a/editor.planx.uk/src/ui/editor/ColorPicker.tsx
+++ b/editor.planx.uk/src/ui/editor/ColorPicker.tsx
@@ -2,13 +2,16 @@ import Box, { BoxProps } from "@mui/material/Box";
 import ButtonBase, { ButtonBaseProps } from "@mui/material/ButtonBase";
 import { styled } from "@mui/material/styles";
 import Typography from "@mui/material/Typography";
+import { ErrorMessage } from "formik";
 import React, { useState } from "react";
 import { ChromePicker, ColorChangeHandler } from "react-color";
+import ErrorWrapper from "ui/shared/ErrorWrapper";
 
 export interface Props {
   label?: string;
   inline?: boolean;
   color?: string;
+  errorMessage?: string | undefined;
   onChange?: (newColor: string) => void;
 }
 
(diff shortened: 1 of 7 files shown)


RODO94:
## The ``errorMessage`` Prop in Colour Picker

The prop ``errorMessage`` in the `<ColourPicker> `Component has been added as optional due to it being used in the Editor in components like ``Notice``.

RODO94:
## Error Message Guidance

The tone of voice guiding the error messaging follows information here: [How to write good planning service content](https://opensystemslab.notion.site/How-to-write-good-planning-service-content-30313a55fc5b4499846e32e88b1cf76a#3341fee8d57c42e998f32c4c83d9b75e)

RODO94:

editor.planx.uk/src/ui/editor/ColorPicker.tsx
```diff
@@ -2,13 +2,16 @@ import Box, { BoxProps } from "@mui/material/Box";
...
+import { ErrorMessage } from "formik";
 import React, { useState } from "react";
 import { ChromePicker, ColorChangeHandler } from "react-color";
+import ErrorWrapper from "ui/shared/ErrorWrapper";
 
 export interface Props {
   label?: string;
   inline?: boolean;
   color?: string;
+  errorMessage?: string | undefined;
```
DafyddLlyr, line 14:
```suggestion
  errorMessage?: string;
```

nit: This is the same thing but a bit more terse and matches our code style. The `?` following the property name indicates that this property can be `undefined`.

editor.planx.uk/src/ui/editor/ColorPicker.tsx
```diff
@@ -93,28 +96,30 @@ export default function ColorPicker(props: Props): FCReturn {
...
-          />
-          <ChromePicker
-            color={props.color}
-            disableAlpha={true}
-            onChange={handleChange}
-          />
-        </Popover>
-      ) : null}
-    </Root>
+    <ErrorWrapper error={props.errorMessage || undefined} id="settings-error">
```
DafyddLlyr, line 99: We should use a unique id here, for example `colour-picker`

DafyddLlyr (approved):
Happy to merge following a few (very very minor!) fixes.

PR #3369, opened 2024-07-04
feat: Add error handling for failed file uploads

DafyddLlyr:
## What does this PR do?
 - Follow up to https://github.com/theopensystemslab/planx-core/pull/434 - this PR now catches these errors client side and allows the user the opportunity to resolve them
 - Displays error if a file upload fails
 - Displays error if user tries to proceed with files with a status of `"error"`

## Future improvements
 - It could be nice to add a "retry" button on failed uploads, but this will be covering a edge case. To our knowledge, a user submitting files which have failed to upload has only happened twice.

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
@@ -148,6 +148,7 @@ function Component(props: Props) {
             break;
           }
           case "allRequiredFilesUploaded":
+          case "errorStatus":
             setFileListError(err?.message);
             break;
         }
@@ -188,10 +189,6 @@ function Component(props: Props) {
   return (
     <Card
       handleSubmit={props.hideDropZone ? props.handleSubmit : validateAndSubmit}
-      isValid={
-        props.hideDropZone ||
-        slots.every((slot) => slot.url && slot.status === "success")
-      }
     >
       <FullWidthWrapper>
         <CardHeader {...props} />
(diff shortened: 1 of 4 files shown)


editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
```diff
@@ -188,10 +189,6 @@ function Component(props: Props) {
   return (
     <Card
       handleSubmit={props.hideDropZone ? props.handleSubmit : validateAndSubmit}
-      isValid={
-        props.hideDropZone ||
-        slots.every((slot) => slot.url && slot.status === "success")
-      }
```
DafyddLlyr, line 194: @ianjon3s Having a condition here on the `isValid` prop means that the "Continue" button is disabled, which goes against our general a11y principles. 

Did we ever discuss this one, or is removing this here the right way to go?

editor.planx.uk/src/@planx/components/shared/PrivateFileUpload/UploadedFileCard.tsx
```diff
@@ -91,74 +92,89 @@ export const UploadedFileCard: React.FC<Props> = ({
...
-            variant="body1"
-            pb="0.25em"
-            sx={{ overflowWrap: "break-word", wordBreak: "break-all" }}
+    <ErrorWrapper
+      error={
+        status === "error"
+          ? "Upload failed, please remove file and try again"
+          : undefined
+      }
+    >
```
DafyddLlyr, line 104: Adding this wrapper is the change here, the diff below is purely a whitespace change due to the extra indentation but git hasn't quite caught this.

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public.tsx
DafyddLlyr: I've actually removed it in this PR - added to checklist and marked as done!

jessicamcinchak (approved):
Errors working as expected for me!

PR #3373, opened 2024-07-04
feat: add ODP Schema validation example for FileTypes to publish checks

jessicamcinchak:
Relies on https://github.com/theopensystemslab/planx-core/pull/436

**Changes:**
- Adds a new validation check for "File types" which compares the data fields set by `FileUpload` and/or `FileUploadAndLabel` nodes to the ODP Schema's `FileType` enum values
  - Skips validation check if the `FileUploadAndLabel` component is used in info-only mode with a hidden dropzone
- Adds a new status "Warn" which will _not_ disable the "Publish" button (eg discretionary services may legitmately want to collect bespoke, "unsupported" file types!)
- Adds/updates API tests

**What this doesn't do yet / future design scope:**
- Link to exact node(s) with unsupported variables
- Link to ODP Schema definitions (could even be Issue template?)

Screenshot from 2024-07-05 10-03-07


jessicamcinchak:
[Validate that flow meets schema requirements on "Publish"](https://trello.com/c/LGIoRgym/2641-validate-that-flow-meets-schema-requirements-on-publish)

api.planx.uk/modules/flows/validate/service.ts
```diff
@@ -55,16 +56,20 @@ const validateAndDiffFlow = async (
...
+  const warningChecks = validationChecks.filter((v) => v.status === "Warn");
+  const passingChecks = validationChecks.filter((v) => v.status === "Pass");
   const notApplicableChecks = validationChecks.filter(
     (v) => v.status === "Not applicable",
   );
-  const sortedValidationChecks = applicableChecks.concat(notApplicableChecks);
+  const sortedValidationChecks = failingChecks
+    .concat(warningChecks)
+    .concat(passingChecks)
+    .concat(notApplicableChecks);
```
jessicamcinchak, line 72: Anyone have a more clever suggestion for how to sort in a custom order by "status" ? This is explicit and easy-to-read, but a bit lengthy and I feel like there's probably some cool one-liner?

api.planx.uk/modules/flows/validate/service.ts
```diff
@@ -206,4 +211,63 @@ const inviteToPayEnabled = (flowGraph: FlowGraph): boolean => {
...
+  // Exclude Upload & Label nodes used in "info-only" mode with a hidden dropzone
+  const uploadAndLabelNodes = Object.entries(flowGraph).filter(
+    (entry): entry is [string, Node] =>
+      isComponentType(entry, ComponentType.FileUploadAndLabel) &&
+      entry[1].data?.hideDropZone !== true,
+  );
+  const uploadAndLabelFileTypes = uploadAndLabelNodes
+    .map(([_nodeId, node]) => node.data?.fileTypes)
+    .flat();
+  return uploadAndLabelFileTypes?.map((file: any) => file?.fn as string);
```
jessicamcinchak, line 270: `any` is less than ideal here, but file is an editor-only type definition right now and not a particularly simple one to repeat because of the `conditions` !

api.planx.uk/modules/flows/validate/service.ts
DafyddLlyr: Yeah I can think of a few ways!

`Map.groupBy` is pretty new but seems to have decent coverage https://caniuse.com/?search=map.groupby

I think it would then just be `Map.groupBy(passingChecks, (check) => check.status))`.

If we hit browser compatibility issues here we could import `groupBy` from lodash, or use a `reduce()`.

Docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/groupBy

DafyddLlyr: Ah just realised we're in the API here on Node 18 - I don't think `Map.groupBy()` would be available.

Lodash or `reduce()` would be the way to go here 👍

api.planx.uk/modules/flows/validate/service.ts
```diff
@@ -206,4 +212,69 @@ const inviteToPayEnabled = (flowGraph: FlowGraph): boolean => {
...
+    title: "File types",
+    status: "Pass",
+    message:
+      "Files collected via FileUpload or UploadAndLabel are all supported by the ODP Schema",
+  };
+};
+
+const getFileUploadNodeFns = (flowGraph: FlowGraph): string[] => {
+  const fileUploadNodes = Object.entries(flowGraph).filter(
+    (entry): entry is [string, Node] =>
```
DafyddLlyr, line 261: So, as of TS 5.5 I think we can lose the `entry is [string, Node]` here and it _should_ be inferred from `isComponentType()`.

Docs: https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#inferred-type-predicates

api.planx.uk/modules/flows/validate/service.ts
DafyddLlyr: Yeah this is a fair call, no issues, and thanks for the explanation.

It would be very very meaningful to put some work into a typed Flow data structure (been on the wish list for a while). I'd be keen to have a start at this at some point!

api.planx.uk/modules/flows/validate/validate.test.ts
DafyddLlyr, line 1: Always appreciate seeing meaningful and well structured tests! 👍

api.planx.uk/modules/flows/validate/service.ts
```diff
@@ -23,7 +25,7 @@ type AlteredNode = {
 
 type ValidationResponse = {
   title: string;
-  status: "Pass" | "Fail" | "Not applicable";
+  status: "Pass" | "Fail" | "Warn" | "Not applicable";
```
DafyddLlyr, line 28: I was expecting to see this change also picked up in `docs.yaml`, but it looks like we've missed adding in the `validationChecks` property.

Happy for this to be picked up here, or in a follow up PR 👍

DafyddLlyr (approved):
All looking very good, a few small comments / suggestions here and there 👍 

> Link to exact node(s) with unsupported variables

Nice idea! Yep let's pick this up once we have this function ready for search.

> Link to ODP Schema definitions (could even be Issue template?)

Another great idea 💡

api.planx.uk/modules/flows/validate/service.ts
jessicamcinchak: Yep, you're correct this is just properly inferred from `isComponentType()` - removed unnecessary verbose-ness!

api.planx.uk/modules/flows/validate/service.ts
jessicamcinchak: Thanks for spotting this, Swagger docs now updated :white_check_mark: 

Screenshot from 2024-07-08 12-22-07

PR #3375, opened 2024-07-05
feat: `team_settings` column rename from `boundary_json` to `boundary_bbox`

RODO94:
## What does this PR do?

This PR will set up work to add a outer boundary json to the database for a local authority using the data in the `boundary_url` column

Following this work, when a user updates the `boundary_url` column, it will trigger a fetch to get the GeoJSON and then pass it through **turf.js** to turn it into a ``boundary_bbox`` JSON. This initial migration will enable easier follow on work.

### DIFF
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -157,7 +157,7 @@
       definition:
         enable_manual: false
         insert:
-          columns: "*"
+          columns: '*'
       retry_conf:
         interval_sec: 30
         num_retries: 1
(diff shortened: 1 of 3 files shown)


RODO94:
@DafyddLlyr @jessicamcinchak I seem to be pulling in a few quotation mark changes on the .yml file, is this a prettier issue?

DafyddLlyr (approved):
Once this is merged I'll line up a production deploy so that the sync scripts can also be easily updated to point at the new column 👍

DafyddLlyr:
@RODO94 I don't think it's prettier in this instance - maybe a different version of the Hasura CLI or something like that? Ideally we wouldn't commit this change, but it doesn't make a functional difference if either `'` or `"` is used.

PR #3378, opened 2024-07-05
feat: Team Settings Form Fetching Boundary GeoJSON using Boundary URL

RODO94:
## What does this PR do? 

This PR uses the ``boundary_url`` provided by users to generate the boundary GeoJSON from https://www.planning.data.gov.uk/ API - using the entity number is the guide. 

The returned GeoJSON is then passed through turf.js to output a boundary box geoJSON, a simplified geoJSON object, that is stored in the ``team_settings`` table

### DIFF
--- a/editor.planx.uk/package.json
+++ b/editor.planx.uk/package.json
@@ -36,6 +36,8 @@
     "@tiptap/react": "^2.4.0",
     "@tiptap/suggestion": "^2.0.3",
     "@turf/area": "^7.0.0",
+    "@turf/bbox": "^7.0.0",
+    "@turf/bbox-polygon": "^7.0.0",
     "@turf/buffer": "^7.0.0",
     "@turf/helpers": "^7.0.0",
     "array-move": "^4.0.0",
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/BoundaryForm.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/BoundaryForm.tsx
@@ -1,4 +1,7 @@
+import { bbox, bboxPolygon, feature } from "@turf/turf";
+import axios from "axios";
 import { useFormik } from "formik";
+import { GeoJsonObject } from "geojson";
 import { useStore } from "pages/FlowEditor/lib/store";
 import React, { ChangeEvent } from "react";
 import InputLabel from "ui/editor/InputLabel";
@@ -9,10 +12,21 @@ import { SettingsForm } from "../shared/SettingsForm";
 import { FormProps } from ".";
 
 export default function BoundaryForm({ formikConfig, onSuccess }: FormProps) {
+  const jsonRegEx = /\.json$/;
+  const geoRegEx = /\.geojson$/;
+
   const formSchema = Yup.object().shape({
     boundaryUrl: Yup.string()
+      .matches(
+        jsonRegEx,
+        "Enter a boundary URL in the correct format, https://www.planning.data.gov.uk/entity/123456",
+      )
+      .matches(
+        geoRegEx,
+        "Enter a boundary URL in the correct format, https://www.planning.data.gov.uk/entity/123456",
+      )
       .url(
-        "Enter a boundary URL in the correct format, https://www.planning.data.gov.uk/",
+        "Enter a boundary URL in the correct format, https://www.planning.data.gov.uk/entity/123456",
       )
       .required("Enter a boundary URL"),
   });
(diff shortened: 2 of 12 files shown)


RODO94:
@DafyddLlyr @jessicamcinchak  One for when I'm back, but the Pizza DB doesn't have any

editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/BoundaryForm.tsx
```diff
@@ -21,12 +29,23 @@ export default function BoundaryForm({ formikConfig, onSuccess }: FormProps) {
...
+          boundaryBbox: bboxFeature,
+        });
+        if (isUpdateSuccess) {
+          onSuccess();
+          resetForm({ values });
+        }
+      } catch (error) {
+        formik.errors.boundaryUrl =
+          "We are unable to retrieve your boundary, check your boundary URL and try again";
+        console.error(error);
```
DafyddLlyr, line 48: This looks good!

We might want slightly more fine-grained checking of the error type (i.e. using `isAxiosError()`) or a bit more control over the URL we send the request to.

hasura.planx.uk/metadata/tables.yaml
DafyddLlyr, line 1: I think that this column will need to be accessed by the `public` role also in order to correctly center the map.

package.json
```diff
@@ -13,5 +13,8 @@
     "tests": "./scripts/start-containers-for-tests.sh",
     "analytics": "docker compose -f ./docker-compose.yml -f ./docker-compose.local.yml --profile mock-services --profile analytics  up -d --quiet-pull",
     "logs": "docker compose logs --tail 30 -f"
+  },
+  "dependencies": {
+    "@types/geojson": "^7946.0.14"
```
DafyddLlyr, line 18: This looks like an `npm i` at the wrong level

editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/BoundaryForm.tsx
```diff
@@ -9,10 +11,16 @@ import { SettingsForm } from "../shared/SettingsForm";
 import { FormProps } from ".";
 
 export default function BoundaryForm({ formikConfig, onSuccess }: FormProps) {
+  const suffixRegEx = /\d{1,10}$/;
```
DafyddLlyr, line 14: This is a good approach 👍 

We might actually want to restrict the domain as well (e.g. we don't want to allow users to make a request to `https://something-dodgy.com/123456`). I'll check with @ianjon3s but I think an an input prefix here is likely the best way forward.

DafyddLlyr (changes requested):
A few small comments!

I'll pick up the changes and add them as commits to this branch, but still wanted to follow through the feedback process so you can see the changes made here when you're back 👍

package.json
DafyddLlyr: Fixed by [c410ab5](https://github.com/theopensystemslab/planx-new/pull/3378/commits/c410ab5a1fd45e150c86baf527063a0dcb869e72)

editor.planx.uk/package.json
```diff
@@ -38,6 +38,7 @@
     "@turf/area": "^7.0.0",
     "@turf/buffer": "^6.5.0",
     "@turf/helpers": "^6.5.0",
+    "@turf/turf": "^7.0.0",
```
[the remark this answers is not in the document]
DafyddLlyr: Fixed in [c4f1e3d](https://github.com/theopensystemslab/planx-new/pull/3378/commits/c4f1e3d053263e6f9bbf840104559113fac687b4)

editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/BoundaryForm.tsx
```diff
@@ -21,12 +29,23 @@ export default function BoundaryForm({ formikConfig, onSuccess }: FormProps) {
...
-      const isSuccess = await useStore.getState().updateTeamSettings({
-        boundaryUrl: values.boundaryUrl,
-      });
-      if (isSuccess) {
-        onSuccess();
-        resetForm({ values });
+      try {
+        const { data } = await axios.get(`${values.boundaryUrl}.geojson`);
+        const bboxPoly = bboxPolygon(bbox(data));
+        const bboxFeature = feature(bboxPoly.geometry);
```
DafyddLlyr, line 35: I'm not quite sure if this is the simplest way to achieve this. If we add types and maybe wrap this in a descriptive helper function it may help clear this up.

DafyddLlyr (changes requested):

editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/BoundaryForm.tsx
DafyddLlyr: Addressed here - [74e7089](https://github.com/theopensystemslab/planx-new/pull/3378/commits/74e708982d93980efd3fc5141d7f669a74bc90d0)

The confusion may have come from the fact that there's a `bbox` property on the object, which sort of looks wrong if you're not expecting it. This is actually fine and is an optional part of the GeoJSON spec ([docs](https://datatracker.ietf.org/doc/html/rfc7946#autoid-23)).

image

editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/BoundaryForm.tsx
DafyddLlyr: Fixed here - [21cfa53](https://github.com/theopensystemslab/planx-new/pull/3378/commits/21cfa5347c5f13f87945626a6a6b7928ca36a2b8)

For now I've not added an prefix - I think just copy/pasting the full URL is likely an easier user interaction than grabbing the entity ID.

editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/BoundaryForm.tsx
DafyddLlyr: Addressed here - [7c1d9fb](https://github.com/theopensystemslab/planx-new/pull/3378/commits/7c1d9fb8ba8a738e661c2733ba49d70fb3af0589)

I've also used the helper method `setFieldError()` as there may be a side effect of directly modifying the `errors` object ([docs](https://formik.org/docs/api/formik#setfielderror-field-string-errormsg-string--void)).

jessicamcinchak (approved):
Changes work for me here !

DafyddLlyr (approved):

PR #3382, opened 2024-07-08
chore: Bump `planx-core`

DafyddLlyr:
This should fix failing regression tests.

Currently running against this branch here - https://github.com/theopensystemslab/planx-new/actions/runs/9837286637

Update: Yep, this resolved it. The above regression tests are now passing ✅


jessicamcinchak (approved):
Thanks for getting to the bottom of this one - apologies just merged validation checks PR which also bumped planx-core, therefore causing conflicts here! Should be an easy rebase :crossed_fingers:

PR #3383, opened 2024-07-08
chore: Update all turf packages to v7

DafyddLlyr:
Small issue I noticed when picking up#3378

Alongside updating `@turf/area` ([PR here](https://github.com/theopensystemslab/planx-new/pull/3348)), we're best off keeping all turf packages on the same version.


editor.planx.uk/package.json
```diff
@@ -115,6 +115,7 @@
     "@testing-library/user-event": "^14.4.3",
     "@types/dompurify": "^3.0.5",
     "@types/draft-js": "^0.11.12",
+    "@types/geojson": "^7946.0.14",
```
DafyddLlyr, line 118: This is now required for v7, please see https://github.com/Turfjs/turf/blob/master/CHANGELOG.md#700

jessicamcinchak (approved):

PR #3384, opened 2024-07-08
feat: Add material details to list component and schema

augustlindemer:
Added material detail questions to list component and schema


augustlindemer:
This will result in changes to material questions data structure (previously split between property. and proposal.), now appended as .type, .existing, and .proposed to (presumably) proposal.materials as the list component data field.

DafyddLlyr (approved):

augustlindemer:
perfect @DafyddLlyr!

PR #3386, opened 2024-07-08
feat: validate ODP Schema project types on publish

jessicamcinchak:
Builds on #3373

**Changes:**
- Adds "Project types" validation to publish checks
- Refactors API `modules/flows/validate` file structure a bit: 
  - `service.ts` was getting quite long, so now we have a `service/` subdirectory with an `index.ts` file & topic-specific files for each type of validation check 
  - I think this should be a bit easier to manage and extend going forward !


api.planx.uk/modules/flows/validate/service/projectTypes.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  // Get all project types supported by current release of ODP Schema & compare
+  const validProjectTypes = getValidSchemaValues("ProjectType");
+  const invalidProjectVals: string[] = [];
+  projectTypeVals.forEach((val) => {
+    if (!validProjectTypes?.includes(val)) {
+      invalidProjectVals.push(val);
+    }
+  });
```
DafyddLlyr, line 27: nit: A `filter()` might be a bit more semantically meaningful here. I think we do use `difference()` from lodash for similar checks elsewhere.

DafyddLlyr (approved):
Changes to `/service` folder look great - much neater organisation and feels a bit simpler to expand and grow these checks going forward.

PR #3392, opened 2024-07-09
test: Remove redundant test following `teamEditor` permission update

DafyddLlyr:
Follow up to https://github.com/theopensystemslab/planx-new/pull/3390

🚀 Regression tests running against this branch here - https://github.com/theopensystemslab/planx-new/actions/runs/9853501058


jessicamcinchak (approved):
Thanks for catching & fixing this !

PR #3391, opened 2024-07-09
feat: Read from new `team_settings` table

DafyddLlyr:
## What does this PR do?
 - Reads from new `team_settings` table instead of `team.settings` column throughout the application
 - Import `Team` type from planx-core in the API (and many associate type fixes)
 - Relies on https://github.com/theopensystemslab/planx-core/pull/444
 
 ## Next steps...
 - Write to new table in `team.create()` (planx-core)
 - Drop `team.settings` column

PR #3396, opened 2024-07-09
fix: Use `teamSettings` instead of `settings` in query

DafyddLlyr:
Quick fix for https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1720531966270509

This PR ensures that the queries for `/draft`, `/preview`, and `/published` request `teamSettings` - this matches the current type, as expected by the store function `setTeam()`.

A more comprehensive fix, which avoid the nested `team.teamSettings` syntax and fixed types is incoming here - https://github.com/theopensystemslab/planx-new/pull/3391


jessicamcinchak (approved):
Thank you!

PR #3397, opened 2024-07-10
feat: add an audit table for applications submitted via S3 and Power Automate

jessicamcinchak:
Quick final step in helping Barnet go live with S3/Power Automate submissions soon. 

Adds a basic audit table for tracking application data.

### DIFF
--- a/hasura.planx.uk/migrations/1720597665798_create_table_s3_applications/up.sql
+++ b/hasura.planx.uk/migrations/1720597665798_create_table_s3_applications/up.sql
@@ -1,0 +1,11 @@
+CREATE TABLE "public"."s3_applications" (
+    "id" serial NOT NULL, 
+    "session_id" text NOT NULL,
+    "team_slug" text NOT NULL,
+    "webhook_request" JSONB NOT NULL,
+    "webhook_response" JSONB NOT NULL,
+    "created_at" timestamptz NOT NULL DEFAULT now(),
+    PRIMARY KEY ("id")
+);
+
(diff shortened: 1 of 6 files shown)


jessicamcinchak:
[Support Barnet in going live with Report a Breach using Power Automate](https://trello.com/c/fbJfqvTY/2947-support-barnet-in-going-live-with-report-a-breach-using-power-automate)

hasura.planx.uk/migrations/1720597665798_create_table_s3_applications/up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE "public"."s3_applications" (
```
DafyddLlyr, line 1: My understanding is that these applications will be automatically picked up by the `submission_services_log` view as that's reading Hasura metadata tables directly right?

However, I think that the `submission_services_summary` view will require an update to also read from this new table.

DafyddLlyr (commented):
One small question / comment!

hasura.planx.uk/migrations/1720597665798_create_table_s3_applications/up.sql
jessicamcinchak: Good question - you're right: 
- `submission_services_log` is automatically picking up these records because it reads from Hasura events
- `submission_services_summary` will need to be manually updated (then Metabase too?) to join to this new audit table so analytics can be correctly tracked

I'd prefer to pick this up in a follow-up PR !

DafyddLlyr: Yep, makes total sense! ✅

DafyddLlyr (approved):

PR #3398, opened 2024-07-10
fix: Reduce React re-renders when checking `currentCard()`

DafyddLlyr:
## What's the problem?
 - Navigating through a sufficiently large flow can crash PlanX
  - This generally happens when navigating backwards, and can be triggered consistently when navigating back to or from a section component (though not exclusively)
  - Chrome freezer / runs out of memory
  - Raised as an issue here (PlanX Slack) - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1720449876847459
  - Consistenly recreateable on this flow when reaching the first section marker - https://editor.planx.dev/buckinghamshire/apply-for-planning-permission/published

## What's the cause?
When running the application in dev mode, the following two errors generally coincide with the above occurrence - 

image

image

The stack traces of these errors point to the following lines of code respectively - 

https://github.com/theopensystemslab/planx-new/blob/20ce9d60c30d44092274e67620c45240bd6fc594/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts#L512

https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/src/@planx/components/Section/Public.tsx#L31

After a lot of trial and error debugging these sections it essentially appears that the problem is caused by `upcomingCardIds()` (called when hitting "back") clashing with the rendering of the `AnalyticsProvider` (which passed the updated `node` down to it's descendants).

## What's the solution?
- Refactor `currentCard()`
  - This is a function with side effects - it updates state in the Zustand store by calling `upcomingCards()` and is ultimately responsible for displaying a card to the user
  - This should only be called once, when the continue button or back button is hit (via `record()`)
  - Generally `currentCard()` is called as a shorthand to get the current node Id, so I've made a new property in the store to hold this variable
  - This has the (very helpful) side effect of reducing the number of re-renders as we're no longer frequently calling `upcomingCardIds()`, just referring to the stored `currentCard()` from the last call.
 - Remove `node` from the `AnalyticsProvider`, and just store in Zustand
   - This resolves the errors we've been seeing, as well as making sense from a hierarchy POV

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
@@ -48,7 +48,7 @@ const Card: React.FC<Props> = ({
   ...props
 }) => {
   const theme = useTheme();
-  const [path, currentCard] = useStore((state) => [
+  const [path, visibleNode] = useStore((state) => [
     state.path,
     state.currentCard,
   ]);
@@ -57,9 +57,7 @@ const Card: React.FC<Props> = ({
   const { track } = useAnalyticsTracking();
 
   useEffect(() => {
-    // The Card component is only rendered when there's content the user will
-    // see
-    const visibleNode = currentCard();
+    // The Card component is only rendered when there's content the user will see
     if (visibleNode?.id) track(visibleNode?.id);
   }, []);
 
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
@@ -41,6 +41,7 @@ export interface PreviewStore extends Store.Store {
   ) => Array<string>;
   currentCard: ({ id: Store.nodeId } & Store.node) | null;
   setCurrentCard: () => void;
+  getCurrentCard: () => ({ id: Store.nodeId } & Store.node) | null;
   hasPaid: () => boolean;
   previousCard: (
     node: Store.node | null,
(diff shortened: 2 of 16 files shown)


editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
```diff
@@ -48,7 +48,7 @@ const Card: React.FC<Props> = ({
   ...props
 }) => {
   const theme = useTheme();
-  const [path, currentCard] = useStore((state) => [
+  const [path, visibleNode] = useStore((state) => [
```
DafyddLlyr, line 51: Throughout this PR I've tired to maintain existing variable names when deconstructing state from the store. I think this minimises the number of changes, and hopefully maintains context within the component. For example, sometimes the `currentCard` is referred to as the `node`, `visibleNode` or `currentCard` across differing components.

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -379,6 +380,7 @@ export const previewStore: StateCreator<
 
   resumeSession(session: Session) {
     set({ ...session });
+    get().setCurrentCard();
```
DafyddLlyr, line 383: When resuming, we handle the initial state setup here after getting a `ReconciliationResponse`

Please see https://github.com/theopensystemslab/planx-new/pull/1701 for context.

jessicamcinchak (approved):
All working as expected for me & pizza feels snappy :clap: 

Really valuable refactor here even if unfortunately not root of August's issue yet !!

PR #3399, opened 2024-07-10
fix: React warnings on `PlanningConstraints` component

DafyddLlyr:
## What does this PR do?
- Adds missing `key` prop
- Ensures we use correct semantic HTML hierarchy by not wrapping HTML in paragraph element

image

image

image


jessicamcinchak (approved):
Thanks for picking this up!

PR #3336, opened 2024-06-28
feat: In-page editor navigation menus

ianjon3s:
## What does this PR do?

Adds a new component `EditorNavMenu` that is used at the root, team and flow levels to link to all settings and panel screens.

The menu has an optional `compact` prop that minimises the menu to icons only and introduces tooltips in place of text labels, this is used in the flow editor screen and settings pages.

Routes are initially defined at the component level.

Still to do:

- [x] Update route behaviour so that they function correctly at all levels
- [x] Ensure `isActive` class is passed correctly to active items
- [x] Remove references to settings pages from the account (dropdown) menu 
  - PR - https://github.com/theopensystemslab/planx-new/pull/3408
- [x] Remove Settings component
  - PR - https://github.com/theopensystemslab/planx-new/pull/3409
- [x] Hide menu items based on permissions (i.e. global settings)
   - PR - https://github.com/theopensystemslab/planx-new/pull/3418
- [x] Handle standard wrapper/containers
- [x] Fix Flow layout

PR #3407, opened 2024-07-11
feat: `Permission` component

DafyddLlyr:
## What does this PR do?
 - Adds a `Permission` component which abstracts away the need to fetch and check a user in multiple places
 - Over time we can build this up to either map to a permission model, or just a series of static properties like the two added here
 - Useful to have in place for https://github.com/theopensystemslab/planx-new/pull/3336 where this was a bit of a blocker


ianjon3s (approved):
I'm current unable to test locally and switch between being PlatformAdmin and not, but this is functioning correctly on the pizza link for PlatformAdmin level.

PR #3411, opened 2024-07-11
feat: Hook up granular A4 GIS API variables for Gloucester

augustlindemer:
Hooked up granular A4 GIS API variables for Gloucester


api.planx.uk/modules/gis/service/local_authorities/metadata/gloucester.ts
```diff
@@ -0,0 +1,25 @@
...
+import { LocalAuthorityMetadata } from "../../digitalLand";
+
+const planningConstraints: LocalAuthorityMetadata["planningConstraints"] = {
+  article4: {
+    // Planx granular values link to Digital Land entity.reference
+    records: {
+      "article4.gloucester.SouthgateStreetConservationArea":
+        "Southgate Street Conservation Area",
+      "article4.gloucester.StMichaelsSquare":
+        "St Michael’s Square Article 4 direction",
```
DafyddLlyr, line 20:
```suggestion
      "article4.gloucester.southgateStreetConservationArea":
        "Southgate Street Conservation Area",
      "article4.gloucester.stMichaelsSquare":
        "St Michael’s Square Article 4 direction",
```

I think that these should be in camelCase not PascalCase right?

DafyddLlyr (changes requested):
One question / issue!

Also pushed a commit to handle linting which should allow tests and CI to pass 👍

jessicamcinchak (approved):

PR #3408, opened 2024-07-11
chore: Remove additional options from dropdown menu

DafyddLlyr:
## What does this PR do?
- Remove items from the drop down menu which now belong in the `EditorNavMenu`

PR #3412, opened 2024-07-12
fix: Correctly populate flow settings and status

DafyddLlyr:
## What's the problem?
When creating a new flow, the status shows as "online" if accessed via the Editor Menu. If we check the database, this is set to "offline" (and all tests for triggers etc are passing) - it's just a frontend visual bug.

## What's the cause?
Somewhere in the Editor Menu work a regression has slipped in which changed the way data was fetched. When accessing it via the dropdown menu, there's no issues. When accessing via the Editor Menu, the correct data is not fetched.

This is what's causing E2E tests to fail in https://github.com/theopensystemslab/planx-new/pull/3408 & https://github.com/theopensystemslab/planx-new/pull/3409

## What's the solution?
 - Only fetch flow settings when navigating to flow/settings
 - Make this an async request
 - Make query names unique (made debugging this more tricky as we had multiple `GetFlow` queries)

### DIFF
--- a/editor.planx.uk/src/routes/views/flowEditor.tsx
+++ b/editor.planx.uk/src/routes/views/flowEditor.tsx
@@ -6,17 +6,14 @@ import { View } from "react-navi";
 
 import { client } from "../../lib/graphql";
 import { useStore } from "../../pages/FlowEditor/lib/store";
-import type { FlowSettings } from "../../types";
 
 interface FlowMetadata {
-  flowSettings: FlowSettings;
   flowAnalyticsLink: string;
   isFlowPublished: boolean;
 }
 
 interface GetFlowMetadata {
   flows: {
-    flowSettings: FlowSettings;
     flowAnalyticsLink: string;
     publishedFlowsAggregate: {
       aggregate: {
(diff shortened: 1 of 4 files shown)


editor.planx.uk/src/routes/views/flowEditor.tsx
```diff
@@ -72,9 +67,9 @@ const getFlowMetadata = async (
  */
 export const flowEditorView = async (req: NaviRequest) => {
   const [flow] = req.params.flow.split(",");
-  const { flowSettings, flowAnalyticsLink, isFlowPublished } =
+  const { flowAnalyticsLink, isFlowPublished } =
     await getFlowMetadata(flow, req.params.team);
-  useStore.setState({ flowSettings, flowAnalyticsLink, isFlowPublished });
+  useStore.setState({ flowAnalyticsLink, isFlowPublished });
```
DafyddLlyr, line 72: `flowSettings` is not needed here - it's only actually used and displayed when we navigate to the settings page.

ianjon3s (approved):
Settings and online/offline status appear to be populating correctly across flows

PR #3417, opened 2024-07-12
fix: Correct destructuring in queries for `teamSettings`

DafyddLlyr:
Fixes for a few errors which popped up post production-deply.

### DIFF
--- a/api.planx.uk/modules/saveAndReturn/service/utils.ts
+++ b/api.planx.uk/modules/saveAndReturn/service/utils.ts
@@ -97,7 +97,16 @@ const validateSingleSessionRequest = async (
             team {
               name
               slug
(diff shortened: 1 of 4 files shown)


api.planx.uk/modules/saveAndReturn/service/utils.ts
```diff
@@ -97,7 +97,16 @@ const validateSingleSessionRequest = async (
             team {
               name
               slug
-              settings: team_settings
+              settings: team_settings {
```
DafyddLlyr, line 100: Columns need to be specified for relationships. This response now matches the `TeamSettings` type.

jessicamcinchak (approved):

PR #3418, opened 2024-07-12
feat: Make `EditorNavMenu` items permission based

DafyddLlyr:
## What does this PR do?
- Only displays menu items which a user can access
- Please see https://opensystemslab.slack.com/archives/C01E3AC0C03/p1720786056596119 (OSL Slack) for context

Ideally it would be nice to encode this in our routing system once and use it to also limit access, as opposed to keeping it in the menu and having to add wrappers to views. But this feels like a quick and pragmatic option for now 👍


ianjon3s (approved):
Working as expected locally

PR #3421, opened 2024-07-12
chore: refactor `formatRawProjectTypes` (take 2)

jessicamcinchak:
See https://github.com/theopensystemslab/planx-core/pull/445 for context!

Regression tests timing out on this branch - no errors, not sure why yet ! https://github.com/theopensystemslab/planx-new/actions/runs/9911311940 (likely related to team_settings - let's merge to `main` & fix forward!)

Will drop `project_types` table in a followup PR.

### DIFF
--- a/editor.planx.uk/src/pages/Pay/MakePayment.tsx
+++ b/editor.planx.uk/src/pages/Pay/MakePayment.tsx
@@ -171,16 +171,7 @@ export default function MakePayment({
     );
 
   const PaymentDetails = () => {
-    const [projectType, setProjectType] = useState<string | undefined>();
-
-    useEffect(() => {
-      const fetchProjectType = () => {
-        const projectType = formatRawProjectTypes(rawProjectTypes);
-        setProjectType(projectType);
-      };
-      fetchProjectType();
-    }, []);
-
+    const projectType = formatRawProjectTypes(rawProjectTypes);
     const data = [
       { term: "Application type", details: flowName },
       {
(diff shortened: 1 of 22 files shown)


editor.planx.uk/src/pages/Pay/MakePayment.tsx
```diff
@@ -170,14 +171,11 @@ export default function MakePayment({
...
-    const $public = useStore((state) => state.$public);
     const [projectType, setProjectType] = useState<string | undefined>();
 
     useEffect(() => {
-      const fetchProjectType = async () => {
-        const projectType = await $public().formatRawProjectTypes(
-          rawProjectTypes,
-        );
+      const fetchProjectType = () => {
+        const projectType = formatRawProjectTypes(rawProjectTypes);
```
DafyddLlyr, line 178: nit: We have to use this slightly awkward syntax when calling an async function inside a `useEffect()` which required us to define the function in scope.

Now it's no longer async I think we can just simplify this by dropping the `useEffect()` and associated `useState()` - this isn't a reactive value, just a formatted prop.

```ts
const projectType = formatRawProjectTypes(rawProjectTypes);

// or if the above is a little slow/intensive (I don't really think so?)

const projectType = useMemo(formatRawProjectTypes(rawProjectTypes), [rawProjectTypes])
```

DafyddLlyr (approved):
One suggested simplification / tidy up 👍

DafyddLlyr (approved):

editor.planx.uk/src/pages/Pay/MakePayment.tsx
jessicamcinchak: Good catch ! Updated here `24ba79f` (no indication it needs a `useMemo`, but we can keep an eye on that in prod)

PR #3429, opened 2024-07-16
chore: drop table `project_types`

jessicamcinchak:
Final step to #3421 

There were never introspection tests for the permissions of this table, so changes here only include a new migration & `tables.yaml` changes :heavy_check_mark:


jessicamcinchak:
[[Research] Can we have configurable templates for GovNotify?](https://trello.com/c/dKhIEOIV/2920-research-can-we-have-configurable-templates-for-govnotify)

DafyddLlyr (approved):

PR #3432, opened 2024-07-16
feat: add Planning Data entity & filtered map view links to constraints

jessicamcinchak:
A minor first step at working towards "overriding" constraints by providing the user more context about each constraint. 

Replaces #3216

**Changes:**
- For a constraint that does apply, each entity listed will link to it's entity page on Planning Data (much more reliable than "document-url" field previously used as "source" link)
  - In the case of roads, which are not sourced from Planning Data and therefore can't link there, show the plain text list item as before
- For a constraint that does apply, if it does NOT have a `name` property on Planning Data (eg flood zones), ensure we have smarter fallback text
- For _any_ constraint, show a filtered, centred link to the Planning Data map
  - The `dataset=` query param reflects multiple datasets where applicable (eg WHS)
  - In the case of roads, show a sentence describing how we checked and display the USRN
- Add a subheader "How is it defined" to visually separate metadata
- Improve types (remove explicit definitions where it's already correctly inferred etc)

**Testing:**
- [Addresses here now](https://docs.google.com/spreadsheets/d/1fHy952Ey-yOI5os_W9PxNR8l3qrAguXqWo_Dv-qrTEI/edit?usp=sharing) :slightly_smiling_face: 
- Please especially double check properties where flood zones & classified roads apply!
  
**Before:**
Screenshot from 2024-07-17 10-29-03
Screenshot from 2024-07-17 10-30-34

**After:**
Screenshot from 2024-07-17 10-29-40
Screenshot from 2024-07-17 10-30-19

### DIFF
--- a/editor.planx.uk/src/@planx/components/PlanningConstraints/List.tsx
+++ b/editor.planx.uk/src/@planx/components/PlanningConstraints/List.tsx
@@ -22,6 +22,7 @@ import Caret from "ui/icons/Caret";
 import ReactMarkdownOrHtml from "ui/shared/ReactMarkdownOrHtml";
 
 import { SiteAddress } from "../FindProperty/model";
+import { availableDatasets } from "./model";
 
 const CATEGORY_COLORS: Record<string, string> = {
   "General policy": "#99C1DE",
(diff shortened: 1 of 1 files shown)


DafyddLlyr (approved):
Some solid improvements!

Tested with flood zone and classified roads and getting the expected results 👍

PR #3435, opened 2024-07-17
feat: Initial frontend search proof of concept with Fuse.js

DafyddLlyr:
## What does this PR do?
- Handles search in the frontend using [Fuse.js](https://www.fusejs.io/)
- Creates `useSearch()` hook as a facade for this library
  - This means we could search elsewhere fairly easily by re-using this hook (e.g. teams, flows)
  - We could also more easily swap out the search functionality in future (maybe with our own API or another library) 
- Displays list of external portals to contextualise search results ([see Figma designs here](https://www.figma.com/design/bnUUrsVRG6qPwDkTmVKACI/PlanX-Design?node-id=15558-12317&t=SjQkPvPMMYyjxwVa-4))

### Why search in the frontend?
 - We have all flow data already in the frontend and don't need to fetch this from the db, add parentIds etc (on each debounced search request)
 - Most flows aren't actually _that_ big, so handling this in the frontend isn't as intensive a task as I would have thought (I have some SQL queries for avg. number of nodes, flow size in kb etc I can share).
 - Initial testing indicates it's pretty performant
 - If we start hitting performance issues, the Fuse.js code could still run server side so there's not much "lost" work as such

### Why Fuse.js?
 - Docs and API are comprehensive - covers all our needs and then some
 - Well established and stable
 - Plenty of other options out there if we find it's either not performant or inaccurate

## Next steps...
I'll pick up the below as follow up PRs - 
- [ ] UI tests
- [x] Generate path for node
  - https://github.com/theopensystemslab/planx-core/pull/463
- [ ] Generate URL from path
  - https://github.com/theopensystemslab/planx-new/pull/3460
- [ ] Visual display of nodes within portals
- [ ] Handle data values for complex nodes (FileUploadAndLabel, List, etc)
    - https://github.com/theopensystemslab/planx-new/pull/3448
- [ ] Search all node data, not just `data.fn` and `data.val`

https://github.com/user-attachments/assets/1fe2b1bf-6309-4b3c-a0d8-74744f079862

### DIFF
--- a/editor.planx.uk/src/hooks/useSearch.ts
+++ b/editor.planx.uk/src/hooks/useSearch.ts
@@ -1,0 +1,48 @@
+import Fuse, { IFuseOptions } from "fuse.js";
+import { useEffect, useMemo, useState } from "react";
+
+const DEFAULT_OPTIONS: Required<SearchOptions> = {
+  limit: 20,
+};
+
+interface SearchOptions {
+  limit?: number;
+}
+
+interface UseSearchProps<T extends Record<string, unknown>> {
+  list: T[];
+  keys: string[];
+  options?: SearchOptions;
+}
+
+export const useSearch = <T extends Record<string, unknown>>({
+  list,
+  keys,
+  options,
+}: UseSearchProps<T>) => {
+  const [pattern, setPattern] = useState("");
+  const [results, setResults] = useState<T[]>([]);
+
+  const fuseOptions: IFuseOptions<T> = useMemo(
+    () => ({
+      threshold: 0.3,
+      useExtendedSearch: true,
+      keys,
+    }),
+    [keys],
+  );
+
+  const fuse = useMemo(
+    () => new Fuse<T>(list, fuseOptions),
+    [list, fuseOptions],
+  );
+
+  useEffect(() => {
+    const fuseResults = fuse.search(pattern, {
+      limit: options?.limit || DEFAULT_OPTIONS.limit,
+    });
+    setResults(fuseResults.map((result) => result.item));
+  }, [pattern]);
+
+  return { results, search: setPattern };
+};
--- a/editor.planx.uk/src/hooks/useSearch.ts
+++ b/editor.planx.uk/src/hooks/useSearch.ts
@@ -6,16 +6,26 @@ interface UseSearchProps<T extends object> {
   keys: string[];
 }
 
+export interface SearchResult<T extends object> {
+  item: T;
+  key: string;
+  matchIndices?: [number, number][];
+}
+
+export type SearchResults<T extends object> = SearchResult<T>[];
+
 export const useSearch = <T extends object>({
   list,
   keys,
 }: UseSearchProps<T>) => {
   const [pattern, setPattern] = useState("");
-  const [results, setResults] = useState<T[]>([]);
+  const [results, setResults] = useState<SearchResults<T>>([]);
 
   const fuseOptions: IFuseOptions<T> = useMemo(
     () => ({
       useExtendedSearch: true,
+      includeMatches: true,
+      minMatchCharLength: 3,
       keys,
     }),
     [keys],
@@ -28,7 +38,15 @@ export const useSearch = <T extends object>({
 
   useEffect(() => {
     const fuseResults = fuse.search(pattern);
-    setResults(fuseResults.map((result) => result.item));
+    setResults(
+      fuseResults.map((result) => ({
+        item: result.item,
+        key: result.matches?.[0].key || "",
+        // We only display the first match
+        matchIndices:
+          (result.matches?.[0].indices as [number, number][]) || undefined,
+      })),
+    );
   }, [pattern]);
 
   return { results, search: setPattern };
--- a/editor.planx.uk/src/hooks/useSearch.ts
+++ b/editor.planx.uk/src/hooks/useSearch.ts
@@ -1,31 +1,20 @@
 import Fuse, { IFuseOptions } from "fuse.js";
 import { useEffect, useMemo, useState } from "react";
 
-const DEFAULT_OPTIONS: Required<SearchOptions> = {
-  limit: 20,
-};
-
-interface SearchOptions {
-  limit?: number;
-}
-
-interface UseSearchProps<T extends Record<string, unknown>> {
+interface UseSearchProps<T extends object> {
   list: T[];
   keys: string[];
-  options?: SearchOptions;
 }
 
-export const useSearch = <T extends Record<string, unknown>>({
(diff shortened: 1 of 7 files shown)


DafyddLlyr:
@RODO94 @jessicamcinchak - Still plenty of TODO comments, and outstanding tasks + some planx-core / flow navigation questions for the next dev call, but I'm opening this up for review just now to get feedback on the code that's here. 

Once I've resolved a few questions on https://github.com/theopensystemslab/planx-core/pull/462 (which this PR relies on) we can progress this (still feature flagged) to `main` if we're happy.

editor.planx.uk/src/hooks/useSearch.ts
```diff
@@ -0,0 +1,53 @@
...
+export const useSearch = <T extends object>({
+  list,
+  keys,
+}: UseSearchProps<T>) => {
+  const [pattern, setPattern] = useState("");
+  const [results, setResults] = useState<SearchResults<T>>([]);
+
+  const fuseOptions: IFuseOptions<T> = useMemo(
+    () => ({
+      useExtendedSearch: true,
```
RODO94, line 26: To clarify these Fuse.js configs, I've read the docs so just looking to confirm...

- useExtendedSearch
- - _This is used to allow tooling for exact searching, and statements, and other operators?_
- includeMatches
- - _Gives you the location of the matching text?_

Would we want to sort the results by the Fuse score? Don't know if we've discussed this already... looks like Fuse has that scoring built in if we did want to.

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search.tsx
```diff
@@ -1,132 +1,283 @@
...
+          variant="h3"
+          mb={1}
+          display={"block"}
+        >
+          Search this flow and internal portals
+        </Typography>
+        <Input
+          name="search"
+          value={formik.values.input}
+          onChange={(e) => {
```
RODO94, line 259: What's the reason for switching from a handleChange to using a handleSubmit() within an onChange? 

_Is this to ensure the form is submitted with all form data for the search on each change to the input field?_

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -461,4 +471,20 @@ export const editorStore: StateCreator<
     const inverseOps: OT.Op[] = type.invert(ops);
     send(inverseOps);
   },
+
+  orderedFlow: undefined,
+
+  setOrderedFlow: () => {
```
RODO94, line 477: Were we not ordering / sorting the flow already? 

I tried to follow the sortFlow() fn in planx-core to understand the logic, but got a bit caught up in it all.

RODO94 (commented):
Hiya! Great to spend time looking through the code here, and my comments are just clarifying how it all works. I can cast a more critical eye once it's clear to me, really appreciate the early review request.

editor.planx.uk/src/hooks/useSearch.ts
DafyddLlyr: That's correct 👍 

- `useExtendedSearch` gives us a few additional options for searching, e.g. starts with, ends with, exact etc
- `includeMatches` is required to give us indices to use for highlighting the results in bold

Fuse.js is sorting the results by score already -`shouldSort` is `true by default - ([docs](https://www.fusejs.io/api/options.html#shouldsort))).

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
DafyddLlyr: No, the flow is not stored in a sorted / ordered state.

It's an object with keys ordered alphabetically, and the order of traversal is derived from the linked sequence of edges beginning at `_root`. I have a few questions about how this is handled so we'll cover this a little later at the dev call 👍 

Storing the flow sorted would avoid this step being necessary, but would be more complex when it comes up updating and caching the flow.

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search.tsx
DafyddLlyr: That's correct - it's just to make it semantically clear what's happening here. When the input value changes, we're submitting the form.

This was originally called `handleChange` in an earlier commit as I hadn't set up formik and this wasn't a HTML `form` element, and it was just a function being called when the input changed.

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search.tsx
```diff
@@ -1,132 +1,283 @@
...
-        <SearchResultCard key={result.nodeId} {...result} />
+    <>
+      {text.split("").map((char, index) => (
+        <Typography
+          fontWeight={isHighlighted(index) ? FONT_WEIGHT_BOLD : "regular"}
+          component="span"
+          p={0}
+          key={`headline-character-${index}`}
+          fontFamily={
+            variant === "data" ? '"Source Code Pro", monospace' : "inherit"
```
jessicamcinchak, line 87: nit / future thing: it'd be handy to have our "data" font set in MUI's Theme and accessible via `theme.text.code` or something similar ! Feels like a nice style pattern we're repeating more & more lately

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search.tsx
```diff
@@ -1,132 +1,283 @@
...
+    // TODO: Generate display key from key
+    let displayKey = "Data";
+    const headline = get(item, key).toString() || "";
+
+    // For Answer nodes, update display values to match the parent question
+    if (item.type === ComponentType.Answer) {
+      const parentNode = useStore.getState().flow[item.parentId];
+      Icon = ICONS[ComponentType.Question];
+      title = parentNode!.data.text!;
+      displayKey = "Answer (data)";
```
jessicamcinchak, line 120: nit: in the Editor modals, we refer to them as "Options" not "Answers" - might be nice to keep user-facing language similar here!

jessicamcinchak (approved):
Very cool prototype - happy for this to merge whenever since feature-flagged !

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search.tsx
DafyddLlyr: Very very good shout! Picked this up here, and will add to this PR once merged https://github.com/theopensystemslab/planx-new/pull/3474

PR #3427, opened 2024-07-15
fix: temporary redirect for Canterbury FOI service

jessicamcinchak:
See thread https://opensystemslab.slack.com/archives/C5Q59R3HB/p1721054184936579

Follows same pattern as #3147 so that the advertised `/preview` link redirects to the "offline" `/published` link.

PR #3437, opened 2024-07-18
fix: react-navi error for Canterbury redirect

DafyddLlyr:
Fixes small typo from https://github.com/theopensystemslab/planx-new/pull/3427 which is causing a react-navi warning.

image


jessicamcinchak (approved):
Thanks for catching!

PR #3439, opened 2024-07-18
chore: fix audit entries and setup Slack notifications for production S3 + Power Automate submissions

jessicamcinchak:
**Two key changes here:** 
- Fixes a previous error where the S3 send events were failing because of a circular async/await error (Kev was still getting webhook notification fine, but we were failing to populate our audit table). 
- Adds a Slack notification on new row insert to `s3_applications` audit table following existing pattern so we can keep a basic pulse on production integration activity

Notes on testing:
- The Slack notification is only called on prod, so the "correct" response in the Hasura event triggers tab here is "Skipping because not production" 
- As for the audit table entry, please note that the _all_ of our Planx environments are sending to the same Power Automate webhook URL on Barnet's side (eg Kev does _not_ have a staging versus production distinction in Power Automate).  
  - While the notification message will say which Planx env the application is coming from, I'm not entirely sure how Kev is handling these (eg if he's saving all tests to same S3 folder as prod apps or filtering/ignoring them etc?) - so trying to not test excessively, or at least flag to him when I am !

### DIFF
--- a/api.planx.uk/modules/send/s3/index.ts
+++ b/api.planx.uk/modules/send/s3/index.ts
@@ -1,6 +1,7 @@
-import axios from "axios";
+import axios, { AxiosRequestConfig } from "axios";
 import type { NextFunction, Request, Response } from "express";
 import { gql } from "graphql-request";
+
 import { $api } from "../../../client";
 import { Passport } from "../../../types";
 import { uploadPrivateFile } from "../../file/service/uploadFile";
@@ -60,8 +61,8 @@ const sendToS3 = async (req: Request, res: Response, next: NextFunction) => {
       `${sessionId}.json`,
     );
 
-    // Send a notification with the file URL to the Power Automate webook
-    const webhookRequest = {
+    // Send a notification with the file URL to the Power Automate webhook
+    const webhookRequest: AxiosRequestConfig = {
       method: "POST",
       url: powerAutomateWebhookURL,
       adapter: "http",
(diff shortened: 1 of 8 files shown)


api.planx.uk/modules/send/s3/index.ts
```diff
@@ -125,10 +122,11 @@ const sendToS3 = async (req: Request, res: Response, next: NextFunction) => {
         );
       });
 
-    return res.status(200).send({
+    res.status(200).send({
```
DafyddLlyr, line 125: I guess the `return` here was the original issue - the audit table step wasn't being properly awaited?

Can't quite follow what the exact issue was but it could be caused by mixing `await` and `then()`, which can lead to some issues. It might be worth just using async/await throughout and adding back the `return` here. 

Not sure if this is totally necessary but it might make things a little easier to follow and reason with.

DafyddLlyr (approved):
One small question - looks good!

Haven't manually tested so that Kev doesn't get a ping unnecessarily 👍

api.planx.uk/modules/send/s3/index.ts
```diff
@@ -109,14 +107,13 @@ const sendToS3 = async (req: Request, res: Response, next: NextFunction) => {
             session_id: sessionId,
             team_slug: localAuthority,
             webhook_request: webhookRequest,
-            webhook_response: res,
+            webhook_response: res.headers,
```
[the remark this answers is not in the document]
jessicamcinchak: Totally agree - thanks for spotting this - intention is to store whole `res`, this was a small remnant from debugging. Now updated!

api.planx.uk/modules/send/s3/index.ts
jessicamcinchak: Correct! Everything working as expected without this `return` now.

Agree the mix of async/await and an Axios `.then()` are a bit confusing to reason about, but this now exactly matches the pattern we have for auditing BOPS etc (perhaps worth revisiting in whole sometime for consistency!): https://github.com/theopensystemslab/planx-new/blob/main/api.planx.uk/modules/send/bops/bops.ts#L134

PR #3441, opened 2024-07-19
feat: Add New Team for Platform Admins 

RODO94:
## What does this PR do?

This PR introduces a new button on the **Teams** page of the **Editor** which allows a PlatformAdmin to add a new Team to the database.

> By adding a new Team to the database, only ``name`` and ``slug`` are user generated, other fields in the **teams**, **team_themes**, **team_settings**, and **team_integrations** tables all rely on default values for the user to add later

Validation has also been added to ensure all team names are unique in the database, with an alert being added to indicate to the user that the team has not been added, as shown below.

Validation has been set up to use the ``slug`` rather than the ``name`` to catch capitalisation similarities.

>Capitalisation similarities would be like _Newcastle-upon-tyne_ being the same as _Newcastle-Upon-Tyne_ 

image

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/team.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/team.ts
@@ -10,16 +10,6 @@ import type { StateCreator } from "zustand";
 
 import { SharedStore } from "./shared";
 
-interface NewTeam {
-  name: string;
-  slug: string;
-  domain?: string;
-  reference?: string;
-  submissionEmail?: string;
-  settings?: Partial<TeamSettings>;
-  theme?: Partial<TeamTheme>;
-}
-
 export interface TeamStore {
   teamId: number;
   teamIntegrations: TeamIntegrations;
@@ -35,7 +25,7 @@ export interface TeamStore {
   fetchCurrentTeam: () => Promise<Team>;
   updateTeamTheme: (theme: Partial<TeamTheme>) => Promise<boolean>;
   updateTeamSettings: (teamSettings: Partial<TeamSettings>) => Promise<boolean>;
-  create: (newTeam: NewTeam) => Promise<number>;
+  create: (newTeam: { name: string; slug: string }) => Promise<number>;
 }
 
 export const teamStore: StateCreator<
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/team.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/team.ts
@@ -10,6 +10,16 @@ import type { StateCreator } from "zustand";
 
 import { SharedStore } from "./shared";
 
+interface NewTeam {
+  name: string;
+  slug: string;
+  domain?: string;
+  reference?: string;
+  submissionEmail?: string;
+  settings?: Partial<TeamSettings>;
+  theme?: Partial<TeamTheme>;
+}
+
 export interface TeamStore {
   teamId: number;
   teamIntegrations: TeamIntegrations;
@@ -25,6 +35,7 @@ export interface TeamStore {
   fetchCurrentTeam: () => Promise<Team>;
   updateTeamTheme: (theme: Partial<TeamTheme>) => Promise<boolean>;
   updateTeamSettings: (teamSettings: Partial<TeamSettings>) => Promise<boolean>;
+  create: (newTeam: NewTeam) => Promise<number>;
 }
 
 export const teamStore: StateCreator<
@@ -64,6 +75,13 @@ export const teamStore: StateCreator<
     slug: get().teamSlug,
     theme: get().teamTheme,
   }),
+
+  create: async (newTeam) => {
+    const { $client } = get();
+    console.log(newTeam);
+    const isSuccess = await $client.team.create(newTeam);
+    return isSuccess;
+  },
 
   initTeamStore: async (slug) => {
     const { data } = await client.query({
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/team.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/team.ts
@@ -78,7 +78,6 @@ export const teamStore: StateCreator<
 
   create: async (newTeam) => {
     const { $client } = get();
-    console.log(newTeam);
     const isSuccess = await $client.team.create(newTeam);
     return isSuccess;
   },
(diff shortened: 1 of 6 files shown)


editor.planx.uk/src/pages/Teams.tsx
```diff
@@ -72,9 +77,51 @@ const Teams: React.FC<Props> = ({ teams, teamTheme }) => {
...
+          Select a team
+        </Typography>
+        {isPlatformAdmin ? (
+          <AddButton
+            onClick={async () => {
+              const newTeamName = prompt("Team name");
+
+              if (newTeamName) {
+                const newSlug = slugify(newTeamName);
+                const teamNameDuplicate = teams.find(
```
RODO94, line 100: If there's a slicker way to validate the name, I'd love to get feedback on this

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -1612,6 +1612,26 @@
     - name: team
       using:
         foreign_key_constraint_on: team_id
+  insert_permissions:
```
RODO94, line 1615: All fields have been altered to accept all, but I do feel an urge that it could only need to be ``id`` and ``team_id``. Let me know if the door is too wide here...

RODO94:
Currently updating Hasura tests to match permissions changes...

editor.planx.uk/src/pages/Teams.tsx
DafyddLlyr: I think this is perfect for now! If we had a huge number of teams, or pagination, or users which couldn't view all teams, we'd need something a bit more sophisticated but this works great 👍 

One small nitpick / comment would be that the variable is called `teamNameDuplicate` but it's actually checking the team slug. 

We could also use a guard clause here to avoid conditional nesting, which makes things a little easier to read.

```ts
const isDuplicate = teams.find((team) => team.slug === newSlug);
if (isDuplicate) {
  alert(
    `A team called "${newSlug}" already exists. Enter a unique team name to continue.`
  );
}

// make team and redirect!

```

editor.planx.uk/src/pages/FlowEditor/lib/store/team.ts
```diff
@@ -65,6 +66,12 @@ export const teamStore: StateCreator<
     theme: get().teamTheme,
   }),
 
+  create: async (newTeam) => {
```
DafyddLlyr, line 69: We should call this `createTeam()` 👍 

Even though `create` is a useful and clear name in the context of the `TeamStore`, we actually merge all stores into one (checkout `editor.planx.uk/src/pages/FlowEditor/lib/store/index.ts`). 

In isolation in another component `useStore.getState().create()` doesn't have much context.

editor.planx.uk/src/pages/Teams.tsx
```diff
@@ -72,9 +77,49 @@ const Teams: React.FC<Props> = ({ teams, teamTheme }) => {
...
+                  (team) => team.slug === newSlug,
+                );
+                if (teamNameDuplicate === undefined) {
+                  await useStore
+                    .getState()
```
DafyddLlyr, line 105: nit: We could simplify this by extracting `create()` (or `createTeam()` 😉) at the top of the component.

```ts
const [canUserEditTeam, createTeam] = useStore((state) => [state.canUserEditTeam, state.createTeam])
```

hasura.planx.uk/metadata/tables.yaml
DafyddLlyr: I think for now it's fine to keep this as `INSERT` only for `platformAdmins` - it's a pretty limited user group and we'll always be able to adjust.

editor.planx.uk/src/pages/Teams.tsx
```diff
@@ -72,9 +77,49 @@ const Teams: React.FC<Props> = ({ teams, teamTheme }) => {
...
+                    .then(() => navigation.navigate(`/${newSlug}`));
+                } else {
+                  alert(
+                    `A team with the name "${teamNameDuplicate.name}" already exists. Enter a unique team name to continue.`,
+                  );
+                }
+              }
+            }}
+          >
+            Add a new Team
```
DafyddLlyr, line 119:
```suggestion
            Add a new team
```

DafyddLlyr (commented):
A few small comments and suggestions.

I'm also getting an error when I - 
 - create a team
 - create a flow for this team
 - step back and go to settings
 
Maybe there's a dependency somewhere on the data fetched from the /teams page?

https://github.com/user-attachments/assets/7446f2fe-1aa6-4261-8d6e-c57ced54b4a7

editor.planx.uk/src/pages/Teams.tsx
```diff
@@ -72,9 +77,49 @@ const Teams: React.FC<Props> = ({ teams, teamTheme }) => {
...
+          flexDirection: "row",
+          justifyContent: "space-between",
+          alignItems: "center",
+          marginBottom: "40px",
+        }}
+      >
+        <Typography variant="h2" component="h1">
+          Select a team
+        </Typography>
+        {isPlatformAdmin ? (
```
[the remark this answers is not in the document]
RODO94: This is a super cool component

RODO94:
> A few small comments and suggestions.
> 
> I'm also getting an error when I -
> 
> * create a team
> * create a flow for this team
> * step back and go to settings
> 
> Maybe there's a dependency somewhere on the data fetched from the /teams page?
> 
>  Screen.Recording.2024-07-19.at.17.23.14.mov

@DafyddLlyr I couldn't recreate this and it's working for me. The lack of Homepage shouldn't break anything... would you be able to try again?

editor.planx.uk/src/pages/Teams.tsx
RODO94: Switched the variable name to include Slug rather than Name, makes more sense for another dev. I have a few other points below on the other parts of the code you mentioned. 

Since they are adding a ``teamName`` I think this should be used in the Alert message rather than the ``newSlug``. Could create a wee bit of confusion with someone thinking "I didn't add that" which would be reduced if we used the name. 

I kept away from using ``isDuplicate`` since the ``find`` method will return an ``object{} || undefined`` rather than a bool. 

Guard clause is a good idea and implemented now! 

Thanks for the feedback

RODO94:
[Ability to "Add team" in Editor](https://trello.com/c/wl3v7OIe/2952-ability-to-add-team-in-editor)

DafyddLlyr (approved):
> @DafyddLlyr I couldn't recreate this and it's working for me. The lack of Homepage shouldn't break anything... would you be able to try again?

Yep can't re-create! Very glad I recorded that so I know it wasn't all in my head 😅 

Happy to merge as-is, but a few very small tidy ups suggested 👍

PR #3442, opened 2024-07-19
Production deploy

jessicamcinchak:
So that we'll have auditing & notifications in place if Barnet launches as expected on Monday, July 22.

### DIFF
--- a/api.planx.uk/modules/webhooks/service/sendNotification/index.ts
+++ b/api.planx.uk/modules/webhooks/service/sendNotification/index.ts
@@ -5,6 +5,7 @@ import {
   EmailEventData,
   EventData,
   EventType,
+  S3EventData,
   UniformEventData,
 } from "./types";
 import { $api } from "../../../../client";
@@ -41,6 +42,11 @@ const getMessageForEventType = (data: EventData, type: EventType) => {
     const { request, session_id, team_slug } = data as EmailEventData;
     return `New email submission "${request.personalisation.serviceName}" *${session_id}* [${team_slug}]`;
   }
+
+  if (type === "s3-submission") {
+    const { session_id, team_slug } = data as S3EventData;
+    return `New S3 + Power Automate submission *${session_id}* [${team_slug}]`;
+  }
 };
 
 const getSessionIdFromEvent = (data: EventData, type: EventType) =>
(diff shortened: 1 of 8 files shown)


api.planx.uk/modules/webhooks/service/sendNotification/index.ts
```diff
@@ -41,13 +42,19 @@ const getMessageForEventType = (data: EventData, type: EventType) => {
...
+    return `New S3 + Power Automate submission *${session_id}* [${team_slug}]`;
+  }
 };
 
 const getSessionIdFromEvent = (data: EventData, type: EventType) =>
   ({
     "bops-submission": (data as BOPSEventData).session_id,
     "uniform-submission": (data as UniformEventData).payload?.sessionId,
     "email-submission": (data as EmailEventData).session_id,
+    "s3-submission": (data as S3EventData).session_id,
```
RODO94, line 57: @jessicamcinchak is ``s3-submissions`` a reference to S3 doc storage? 

Also, are we now setting up a new notification system because Barnet are submitting applications in a new way?

DafyddLlyr (approved):

api.planx.uk/modules/webhooks/service/sendNotification/index.ts
jessicamcinchak: Yes ! This is in reference to API `modules/send/s3` which works as follows:
- Creates a validated or non-validated ODP Schema JSON file payload depending on application type
- Uploads it as a private file to our AWS S3 user data bucket
- Pings a webhook endpoint provided by the council with a notification message including the file link that the application is ready to download and process (in Barnet's case, the webhook and processing is configured via Power Automate, but in theory it could be more generic than that).

It's essentially a more automated option than "Send to email" for bulk-processing submissions. 

The notifications are already in place for our other "Send destinations" (they're council agnostic), and I'm adding them for this destination now too that the first council using this integration is ready to go live soon (we don't notify on staging, so hasn't been any reason to setup until now!)

RODO94: @jessicamcinchak So are they using power automate to extract data from the file in S3?

PR #2870, opened 2024-03-07
fix(a11y): require title for all @planx components

jessicamcinchak:
Follow on from comment here: https://github.com/theopensystemslab/planx-new/pull/2868#pullrequestreview-1921800369

PR #3446, opened 2024-07-22
feat: Add new `MapAndLabel` component to @planx components

RODO94:
## What does this PR do?

This PR creates the framework for a new Editor "@planx" component. 

To create this PR, I have followed the steps laid out here: https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/docs/adding-a-new-component.md

*_Replacing `SetValue` with `MapAndLabel`_

With this PR, there has also been an update to the planx-core ref in the package.json

### Still To-Do In Other PRs

- [ ]  Refine MapAndLabel modal component
- [ ]  Define shape of data outputted from MapAndLabel component
- [ ]  Create tests for new component

### DIFF
--- a/e2e/tests/api-driven/package.json
+++ b/e2e/tests/api-driven/package.json
@@ -8,7 +8,7 @@
   "dependencies": {
     "@cucumber/cucumber": "^9.3.0",
     "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#fc14234",
-    "axios": "^1.6.8",
+    "axios": "^1.7.2",
     "dotenv": "^16.3.1",
     "dotenv-expand": "^10.0.0",
     "graphql": "^16.8.1",
--- a/e2e/tests/api-driven/package.json
+++ b/e2e/tests/api-driven/package.json
@@ -7,8 +7,8 @@
   "packageManager": "pnpm@8.6.6",
   "dependencies": {
     "@cucumber/cucumber": "^9.3.0",
-    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#6c2cc59",
-    "axios": "^1.7.2",
+    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#fc14234",
+    "axios": "^1.6.8",
     "dotenv": "^16.3.1",
     "dotenv-expand": "^10.0.0",
     "graphql": "^16.8.1",
--- a/e2e/tests/ui-driven/package.json
+++ b/e2e/tests/ui-driven/package.json
@@ -8,8 +8,8 @@
     "postinstall": "./install-dependencies.sh"
   },
   "dependencies": {
-    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#6c2cc59",
-    "axios": "^1.7.2",
+    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#fc14234",
+    "axios": "^1.6.8",
     "dotenv": "^16.3.1",
     "eslint": "^8.56.0",
     "graphql": "^16.9.0",
--- a/e2e/tests/ui-driven/package.json
+++ b/e2e/tests/ui-driven/package.json
@@ -9,7 +9,7 @@
   },
   "dependencies": {
     "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#fc14234",
-    "axios": "^1.6.8",
+    "axios": "^1.7.2",
     "dotenv": "^16.3.1",
     "eslint": "^8.56.0",
     "graphql": "^16.9.0",
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
@@ -41,6 +41,7 @@ function MapAndLabelComponent(props: Props) {
                 placeholder={"Title"}
                 value={formik.values.title}
                 onChange={formik.handleChange}
+                required
               />
             </InputRow>
             <InputRow>
@@ -57,7 +58,7 @@ function MapAndLabelComponent(props: Props) {
                 name="fn"
                 placeholder={"Data Field"}
                 value={formik.values.fn}
-                disabled
+                required
               />
             </InputRow>
           </InputGroup>
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
@@ -1,8 +1,8 @@
 import { ComponentType as TYPES } from "@opensystemslab/planx-core/types";
+import { EditorProps } from "@planx/components/ui";
 import { useFormik } from "formik";
 import React from "react";
 
-import { EditorProps } from "../ui";
 import { MapAndLabel, parseContent } from "./model";
 
 type Props = EditorProps<TYPES.MapAndLabel, MapAndLabel>;
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
@@ -1,8 +1,8 @@
 import { ComponentType as TYPES } from "@opensystemslab/planx-core/types";
-import { EditorProps } from "@planx/components/ui";
 import { useFormik } from "formik";
 import React from "react";
 
+import { EditorProps } from "../ui";
 import { MapAndLabel, parseContent } from "./model";
 
 type Props = EditorProps<TYPES.MapAndLabel, MapAndLabel>;
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
@@ -1,0 +1,28 @@
+import { ComponentType as TYPES } from "@opensystemslab/planx-core/types";
+import { useFormik } from "formik";
+import React from "react";
+
+import { EditorProps } from "../ui";
+import { MapAndLabel, parseContent } from "./model";
+
+type Props = EditorProps<TYPES.MapAndLabel, MapAndLabel>;
+
+export default MapAndLabelComponent;
+
+function MapAndLabelComponent(props: Props) {
+  const formik = useFormik({
+    initialValues: parseContent(props.node?.data),
+    onSubmit: (newValues) => {
+      props.handleSubmit?.({
+        type: TYPES.MapAndLabel,
+        data: newValues,
+      });
+    },
+  });
+
+  return (
+    <form onSubmit={formik.handleSubmit} id="modal">
+      //...
+    </form>
+  );
+}
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
@@ -1,8 +1,14 @@
 import { ComponentType as TYPES } from "@opensystemslab/planx-core/types";
 import { useFormik } from "formik";
 import React from "react";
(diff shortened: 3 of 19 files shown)


editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
```diff
@@ -0,0 +1,28 @@
+import { ComponentType as TYPES } from "@opensystemslab/planx-core/types";
+import { useFormik } from "formik";
+import React from "react";
+
+import { EditorProps } from "../ui";
```
RODO94, line 5: Tried to get this working with the path "@planx/..." but it kept throwing an error so I changed to a relative path

RODO94:
@DafyddLlyr and @jessicamcinchak would be great to get a sense check on these changes. There's a lot more detail and work to be done, but I hope I've at least set it all up in the right places.

editor.planx.uk/src/@planx/components/MapAndLabel/model.ts
```diff
@@ -0,0 +1,12 @@
+import { MoreInformation, parseMoreInformation } from "../shared";
+
+export interface MapAndLabel extends MoreInformation {
+  fn: string;
```
jessicamcinchak, line 4: At minimum, every component also has a `title` & `description` available to configure in the Editor - I'd expect to see those fields here !

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Node.tsx
```diff
@@ -66,6 +66,8 @@ const Node: React.FC<any> = (props) => {
       return <Question {...allProps} text="Find property" />;
     case TYPES.List:
       return <Question {...allProps} text={node?.data?.title ?? "List"} />;
+    case TYPES.MapAndLabel:
+      return <Question {...allProps} text="Map and Label" />;
```
jessicamcinchak, line 70: nit: once you add a `title` field in the Editor, then we could follow the pattern `text={node?.data?.title ?? "Map and label"} />` - typically makes nodes much easier to navigate in the graph if using custom title rather than generic component name !

editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
```diff
@@ -62,6 +62,7 @@ const NodeTypeSelect: React.FC<{
         <option value={TYPES.AddressInput}>Address Input</option>
         <option value={TYPES.ContactInput}>Contact Input</option>
         <option value={TYPES.List}>List</option>
+        <option value={TYPES.MapAndLabel}>Map and Label</option>
```
jessicamcinchak, line 65: nit: I'd comment-out this line in this initial PR - basically like a pseudo feature flag so no one can select/use this component type yet ! Or create a new "TREES" feature flag and wrap it :slightly_smiling_face: :deciduous_tree:

editor.planx.uk/src/pages/FlowEditor/components/forms/index.ts
```diff
@@ -53,6 +54,7 @@ const components: {
   "internal-portal": InternalPortal,
   "file-upload-and-label": FileUploadAndLabel,
   list: List,
+  "map-and-label": MapAndLabelComponent,
```
jessicamcinchak, line 57: nit: don't think we need `Component` prefix here?

editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
```diff
@@ -0,0 +1,28 @@
...
+      props.handleSubmit?.({
+        type: TYPES.MapAndLabel,
+        data: newValues,
+      });
+    },
+  });
+
+  return (
+    <form onSubmit={formik.handleSubmit} id="modal">
+      //...
```
jessicamcinchak, line 25: nit: I'd expect to see a little bit more initially outlined here - see below comment re: title & description. Default props also already include internal notes! 

Maybe more like this:
```
      <ModalSection>
        <ModalSectionContent
          title="Map and label"
          Icon={ICONS[TYPES.MapAndLabel]}
        >
          <InputGroup>
            <InputRow>
              <Input
                format="large"
                name="title"
                placeholder={formik.values.title}
                value={formik.values.title}
              />
            </InputRow>
            <InputRow>
              <RichTextInput
                name="description"
                placeholder="Description"
                value={formik.values.description}
              />
            </InputRow>
            <InputRow>
              <Input
                format="data"
                name="fn"
                placeholder={formik.values.fn}
                value={formik.values.fn}
                disabled
              />
            </InputRow>
          </InputGroup>
        </ModalSectionContent>
      </ModalSection>
      <InternalNotes
        name="notes"
        value={formik.values.notes}
      />
    </form>
```

editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
```diff
@@ -0,0 +1,20 @@
...
+type Props = PublicProps<MapAndLabel>;
+
+function MapAndLabelComponent(props: Props) {
+  return (
+    <Card handleSubmit={props.handleSubmit} isValid>
+      <CardHeader
+        info={props.info}
+        policyRef={props.policyRef}
+        howMeasured={props.howMeasured}
+      />
```
jessicamcinchak, line 17: nit: similar to other comments, let's get `title` & `description` props in here to start too !

jessicamcinchak (commented):
Recognise this is still draft, but just a few initial spots ! Definitely on the right track, always nice to see the component library growing :books:

e2e/tests/api-driven/package.json
```diff
@@ -7,8 +7,8 @@
   "packageManager": "pnpm@8.6.6",
   "dependencies": {
     "@cucumber/cucumber": "^9.3.0",
-    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#6c2cc59",
-    "axios": "^1.7.2",
+    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#fc14234",
+    "axios": "^1.6.8",
```
jessicamcinchak, line 11: This looks like a rebasing bug - we want to keep `axios` at v1.7.2 please as merged this morning because that fixed e2e issues !

e2e/tests/ui-driven/package.json
```diff
@@ -8,8 +8,8 @@
     "postinstall": "./install-dependencies.sh"
   },
   "dependencies": {
-    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#6c2cc59",
-    "axios": "^1.7.2",
+    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#fc14234",
+    "axios": "^1.6.8",
```
jessicamcinchak, line 12: Same as above - this shouldn't be downgrading

editor.planx.uk/src/@planx/components/MapAndLabel/model.ts
```diff
@@ -0,0 +1,16 @@
...
+  title?:string;
+  description?:string;
```
jessicamcinchak, line 6: I'd expect `title` to be required, while `description` can be optional ! Requiring titles means that components like the `SummaryList` can be guaranteed to have accurate row headers for accessibility - see context here https://github.com/theopensystemslab/planx-new/pull/2870

Also a nitpick: I personally find it much easier to read when properties are defined in the order that we think of the editor form - eg title first, description, then fn.

editor.planx.uk/src/@planx/components/MapAndLabel/model.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export const parseContent = (
+  data: Record<string, any> | undefined,
+): MapAndLabel => ({
+  fn: data?.fn || "",
+  ...parseMoreInformation(data),
+  title:"",
+  description:""
+});
```
jessicamcinchak, line 16: Similar to above: 
- title & description should reference their respective form inputs - eg `title: data?.title || ""` & `description: data?.description` (no fallback necessary if optional field)
- order according to editor form for readability, and keep any spread (`...`) values last !

editor.planx.uk/src/pages/Preview/Node.tsx
```diff
@@ -172,6 +173,9 @@ const Node: React.FC<any> = (props: Props) => {
     case TYPES.List:
       return <List {...allProps} />;
 
+    case TYPES.MapAndLabel:
+      return <MapAndLabelComponent />;
```
jessicamcinchak, line 177: Looks like this is missing `{...props}` ! Passing these in should allow us to access/display title in graph etc

jessicamcinchak (commented):
Thanks for changes so far! A couple more spots on another review - especially please watch package versions on rebase!

editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
```diff
@@ -62,6 +62,9 @@ const NodeTypeSelect: React.FC<{
         <option value={TYPES.AddressInput}>Address Input</option>
         <option value={TYPES.ContactInput}>Contact Input</option>
         <option value={TYPES.List}>List</option>
+        {hasFeatureFlag("TREES") && (
+          <option value={TYPES.MapAndLabel}>Map and Label</option>
```
RODO94, line 66: @jessicamcinchak do you think this should be under **Inputs** or **Location?**

RODO94: My thinking with **Inputs** was that the user needs to input data we need...

RODO94:
> Thanks for changes so far! A couple more spots on another review - especially please watch package versions on rebase!

Yeah, sorry to miss those, I always forget the e2e package.json. 

Should be all good now, and I've updated the Node file which was pulling in the Editor component instead of the Public one.

jessicamcinchak (commented):
Most recent changes look good, but I can't open flows on this pizza? Can you?

Would like to be able to do one final in-editor manual test before approving! Maybe Vultr needs a destroy/re-create ?

editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
```diff
@@ -0,0 +1,80 @@
...
+        >
+          <InputGroup>
+            <InputRow>
+              <Input
+                format="large"
+                name="title"
+                placeholder={"Title"}
+                value={formik.values.title}
+                onChange={formik.handleChange}
+              />
```
jessicamcinchak, line 44: nit: one final thing I noticed while manually testing in the Editor - I can currently create a new component without entering any input, which doesn't match the types you've defined in `model.ts` (eg I'd expect `title` & `fn` to be required and throw an error if no value!)

Simply adding the `required` prop on the `Input` should do the trick! (Will just throw a native browser error if empty then, not full Gov UK styled error - but that's fine for the Editor for now)

editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
```diff
@@ -0,0 +1,80 @@
...
+            </InputRow>
+            <InputRow>
+              <Input
+                format="data"
+                name="fn"
+                placeholder={"Data Field"}
+                value={formik.values.fn}
+                disabled
```
jessicamcinchak, line 60: Why is this one `disabled` ?

jessicamcinchak (approved):
Two minor last things - but overall working for me! 

Since feature flagged, feels safe to merge whenever you're ready !

editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
RODO94: Lingering mistake from me turning things off and on! I'll correct before merging

editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
RODO94: Great! I'll update before merging

PR #3449, opened 2024-07-23
feat: update passport based on inaccurate constraints

jessicamcinchak:
**Changes:**
- Adds function `handleOverrides()` and associated tests to ensure that passport data (specifically `property.constraint.planning` & `_nots["property.constraints.planning"]` which power automations) correctly account for any constraints marked as inaccurate
- Preserves previously marked overrides when coming "back" to the component
- Shows listed building grade in the list of applicable entities - as this is both useful info and makes it easier to double check associated granular automations

API tests are passing for me fine locally on this branch, I suspect another stale lock file?? Will sort that soon.

**Testing:**
- Hopefully the `utils.test.ts` file helps explain a lot of the complexity and various scenarios - do shout with any example addresses you find that behave differently
- Try this flow for manual testing which has automated questions setup: https://3449.planx.pizza/opensystemslab/digital-land-api-search-test
- & see notes in #planx-design thread here https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1721219483725279


editor.planx.uk/src/@planx/components/PlanningConstraints/utils.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    expect(result["intersectingConstraints"][DEFAULT_FN]).toContain("listed");
+    expect(result["intersectingConstraints"][DEFAULT_FN]).toContain("listed.grade.II");
+  });
+});
+
+describe("handleOverrides with non-granular constraint categories", () => {
+  test("when all entities in a constraint category are inaccurate, the whole category no longer applies", () => {
+    const overrides = {
+      "road.classified": {
+        "entities": ['undefined'], // @todo roads don't have entity ids, record the property.OBJECTID via OS?
```
jessicamcinchak, line 55: This test case highlights how `road.classified` is a still a bit of a strange edge case:
- Because it's returned via OS and not Planning Data, it doesn't match the "entity" type we compare throughout this function (see the `mockConstraints` data as an example)
- But because of our UPRN matching method, we'll only ever return ONE applicable classified road, so the `handleOverride` always fits the first scenario of "every entity is inaccurate, therefore whole category is" and still works as expected :heavy_check_mark: 
- What I need to come back and fix is the `_overrides` audit record, which is currently incorrectly recording 'undefined' !

editor.planx.uk/src/@planx/components/PlanningConstraints/utils.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    const overrides = undefined;
+    const result = handleOverrides(DEFAULT_FN, mockConstraints, overrides, defaultIntersectingConstraints, defaultNots);
+
+    expect(result).toEqual({
+      intersectingConstraints: defaultIntersectingConstraints,
+      nots: defaultNots,
+    });
+  });
+
+  test.skip("when a constraint was reported as inaccurate, but is not part of the latest GIS API response, it's skipped", () => {
```
jessicamcinchak, line 16: This is a super edge casey thing that I _suspect_ can technically happen (because the component fetches fresh data on each load, but we also keep your previous overrides when coming back), but should be very unlikely. 

Mocked this test to not forget about it, but the `handleOverrides` function isn't handling this yet

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
```diff
@@ -125,12 +126,12 @@ function Component(props: Props) {
   );
 
   // XXX handle both/either Digital Land response and custom GIS hookup responses; merge roads for a unified list of constraints
-  const constraints: GISResponse["constraints"] | {} = {
+  const constraints: GISResponse["constraints"] | Record<string, any> = {
```
jessicamcinchak, line 129: There's a few types that are a bit messy here. 

GIS types are mostly managed in planx-core, and I'll plan to open a followup PR to try to tidy some of this - we're notably missing a Planning Data "entity" type which we've avoided in the past because of non-standard fields per dataset - but I think we should be able to get a decent union type in place to eliminate some of these newly introduced `any`s

jessicamcinchak:
[Ability to override planning constraints [Solution to GIS layers pulling constraints (and planning histories) that apply to adjoining properties]](https://trello.com/c/S6zRlO42/2637-ability-to-override-planning-constraints-solution-to-gis-layers-pulling-constraints-and-planning-histories-that-apply-to-adjoini)

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
```diff
@@ -71,9 +77,10 @@ function Component(props: Props) {
 
   // Check if this team should query Planning Data (or continue to use custom GIS) and set URL params accordingly
   //   In future, Planning Data will theoretically support any UK address and this db setting won't be necessary, but data collection still limited to select councils!
-  const hasPlanningData = useStore(
-    (state) => state.teamIntegrations?.hasPlanningData,
-  );
+  const hasPlanningData = true;
+  // useStore(
+  //   (state) => state.teamIntegrations?.hasPlanningData,
+  // );
```
jessicamcinchak, line 83: :warning: this is a temporary hack so that we can fetch planning constraints in the editor sidebar & view automations in the graph view side-by-side for easier testing :warning: 

Currently, `state.teamIntegrations` is undefined in the editor view - adding an `updateTeamIntegrations` method to the `TeamStore` & planx-core looks like the best route to permanently fix this.

editor.planx.uk/src/@planx/components/PlanningConstraints/utils.ts
```diff
@@ -0,0 +1,69 @@
+import { Constraint } from "@opensystemslab/planx-core/types";
+import { InaccurateConstraints } from "./Public";
+import { IntersectingConstraints } from "./model";
+
+export function handleOverrides(
```
DafyddLlyr, line 5: This is a pretty dense function I had a bit of a hard time following (clearly time for another coffee! ☕). Unit tests are helpful and clear however 👍 

There's maybe a few minor things that could be done here, like splitting up into smaller functions or guard clauses etc that might clear things up, but I'm not sure how much overall difference this would make.

I wonder if a significant change could be just to update the framing slightly. Rather than filtering from intersecting and pushing to nots, I wonder if it might be clearer if we were constructing a new object of intersecting constraints? For example - 
 - For each constraint key and value...
 - If not inaccurate, push to new constraint list
 - If inaccurate, push to nots
 
This might not be a viable strategy, and is maybe just totally personal preference, but thought it was worth sharing 👍

DafyddLlyr (approved):
A few comments / suggestions below.

Working as expected on Pizza - both in terms of UI and passport population 👍

editor.planx.uk/src/@planx/components/PlanningConstraints/List.tsx
```diff
@@ -343,6 +355,7 @@ export function formatEntityName(
   metadata?: Metadata,
 ): string {
   return (
+    (entity["listed-building-grade"] && `${entity.name} - Grade ${entity["listed-building-grade"]}`) ||
```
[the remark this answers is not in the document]
jessicamcinchak: Very fair - now updated !

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
jessicamcinchak: Thanks for `initTeamStore()` pointer - that's a much simpler solution than I originally anticipated ! 

Fixed here https://github.com/theopensystemslab/planx-new/pull/3458 (so we can get to `main` sooner/independently and I'll rebase this one)

editor.planx.uk/src/@planx/components/PlanningConstraints/utils.ts
jessicamcinchak: Totally agree this is a dense one - re-organised a bit and I think it feels marginally better ? Very much counting on the tests to do the hard work here :sweat_smile: 

I'm using an approach of iterating through each inaccuracy, rather than each constraint, for a couple of reasons:
- The average case probably won't have inaccuracies, so this will exit immediately
- For cases that do have inaccuracies, the hard part of checking is reconciling the entities within the overall constraint categories and updating parent (eg `designated`) or child (eg `listed.grade.II`) keys correctly - I think this is a bit more straightforward to think about when starting from a single entity ID that we know is inaccurate, rather than starting from the constraint category that has zero to many entities which we then have to check if they are inaccurate or not - but that may very well just be my preference for thinking about this lately !

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
```diff
@@ -42,19 +43,24 @@ export type InaccurateConstraints =
...
-    (useStore(
-      (state) => state.computePassport().data?._address,
-    ) as SiteAddress) || {};
+
+  // PlanningConstraints must come after at least a FindProperty in the graph
   const showGraphError = !x || !y || !longitude || !latitude;
 
-  const teamSlug = useStore((state) => state.teamSlug);
+  // Even though this component will fetch fresh GIS data when coming "back",
+  //   still prepopulate any previously marked inaccurateConstraints
```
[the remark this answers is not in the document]
jessicamcinchak: I think we could theoretically populate this component when coming "back" using something like `cachedBreadcrumbs?.[currentCardId]?.["data"]?.["_constraints"]` which is basically the full collated GIS API response.

But in my opinion, while the load is a bit slow, it's in our best interest to always check for the latest data! Feels like a much easier mental model to communicate to users, and a safer guarantee that we're giving the best suggestion we can.

Currently, we do disable `revalidateOnFocus` and SWR has other similar options (https://swr.vercel.app/docs/revalidation#disable-automatic-revalidations) - but again I'm not sure it's totally worth it unless we hear otherwise in feedback ! If we go the full immutable route, which does make for a totally smooth, refresh-free "back" on quick try, we're still breaking that mental model above.

Hopeful that a forthcoming update to the Planning Data API to omit geometry fields when fetching will marginally speed things up, and could also revisit roads query just to ensure we aren't missing an optimisation there (but last I remember it's just a pretty big slow string match).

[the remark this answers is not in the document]
DafyddLlyr: Yep this makes sense, I think the planning data change should have a decent impact here 👍

editor.planx.uk/src/@planx/components/PlanningConstraints/utils.ts
DafyddLlyr: Thanks for taking another look, agree this is better! And yep - the tests are fantastic here which is massively helpful 😄

DafyddLlyr (approved):

PR #3450, opened 2024-07-23
feat: Not applicable constraints styling

ianjon3s:
## What does this PR do?

- Updates constraints accordion to use a secondary button to report non-applicable constraints
- Creates a theme-level chip style named `notApplicableTag`
- Updates styling of non-applicable constraints to use `notApplicableTag`, rendered conditionally for data reported non-applicable constraints and user-reported non-applicable constraints

Example:
https://3450.planx.pizza/lambeth/check-constraints-on-a-property

A good property with multiple constraints:
SW2 1AH
FLAT A, 102, BRIXTON HILL

Feature flag to enable: `window.featureFlags.toggle("OVERRIDE_CONSTRAINTS")`

### DIFF
--- a/editor.planx.uk/src/@planx/components/PlanningConstraints/List.tsx
+++ b/editor.planx.uk/src/@planx/components/PlanningConstraints/List.tsx
@@ -2,8 +2,10 @@ import Accordion from "@mui/material/Accordion";
 import AccordionDetails from "@mui/material/AccordionDetails";
 import AccordionSummary from "@mui/material/AccordionSummary";
 import Box, { BoxProps } from "@mui/material/Box";
+import Button from "@mui/material/Button";
 import Link from "@mui/material/Link";
 import List from "@mui/material/List";
+import Chip from "@mui/material/Chip";
 import ListItem from "@mui/material/ListItem";
 import ListSubheader from "@mui/material/ListSubheader";
 import { styled } from "@mui/material/styles";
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/PlanningConstraints/List.tsx
```diff
@@ -198,14 +200,31 @@ function ConstraintListItem({ children, ...props }: ConstraintListItemProps) {
...
+            >
+              {children}
+            </Typography>
+            {allEntitiesInaccurate && (
+              <>
+                <Chip
+                  label="Not applicable"
+                  variant="notApplicableTag"
+                  size="small"
+                  sx={{ mr: 0.75, ":not(#negative-constraints-list &)": { display: "none !important" } }}
```
DafyddLlyr, line 217: Here's how I'd have this with props and not CSS - 

- Add an additional prop to `ConstraintsListProps` (something like `variant: "positive" | "negative"` maybe?)
- Populate this x2 in `Public.tsx`
- Drill the same prop down into `ConstraintListItem`
- Use this prop to conditionally render this chip

jessicamcinchak: FYI the existing boolean prop `value` on `ConstraintsListItemProps` already indicates applicable / not applicable - I don't think you should have to add a new one here !

DafyddLlyr: Oh nice, I missed this!

ianjon3s: Thanks both, I've got this working based on the `value` prop 👍

jessicamcinchak (approved):
Looks good, thanks again for quick turnaround here !

PR #3221, opened 2024-05-31
[DEMO][api] add microsoft strategy to auth module

freemvmt:
Letting folks sign into the editor with their Microsoft account is shaping up nicely - the code works on local!

To see this first replace the Microsoft client credentials in `api.planx.uk/modules/auth/strategy/microsoft.ts` (see [Slack](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1717191433602679)), then do `pnpm run up` in project root, `pnpm start` in editor, and click 'Login with Microsoft'.

A lot of the effort was actually trawling the Azure docs to understand how the application that you have to set up with them needs to be configured, and then trying to get the `passportjs` strategy to play nicely with that. This was especially true because the 'official' strategy I tried to implement turned out to be deprecated (without replacement!) by Microsoft. Steps for setting up the application as required are documented [here](https://www.notion.so/opensystemslab/Allow-editors-to-log-in-with-their-Microsoft-ID-fb42bac468f44aabbde7909a75075393?pvs=4#8c25c33288fc47a7ac6aade606a959bf).

Rather than syndicate it all here I'll just point anyone interested in more detail to my [notes in Notion](https://www.notion.so/opensystemslab/Allow-editors-to-log-in-with-their-Microsoft-ID-fb42bac468f44aabbde7909a75075393?pvs=4#8d58cf718c544cfdbcfe8ce2147f92c8), including various hints re security hardenings observed etc.

Development continuing on #3453.

PR #3453, opened 2024-07-25
[api] add Microsoft strategy to auth module (single sign-on)

freemvmt:
🎫 Ticket: https://trello.com/c/owvV9UZh

## Summary

This PR adds the ability to sign into PlanX using a Microsoft account (personal or organisational), which most users (working in local government) are more likely to have to hand as compared to a Google account.

We make use of the existing passportjs implementation, adding a new `microsoft-oidc` strategy which leverages [openid-client](https://www.passportjs.org/packages/openid-client/), which we consider reliable thanks to [certification](https://github.com/panva/node-openid-client?tab=readme-ov-file#certification) by [OpenID themselves](https://openid.net/certification/#OPENID-RP-P).

Since this package has some asynchronous behaviour for which the best code pattern involved using top-level awaits, this project induced some bonuses in `api.planx.uk`, namely adopting ESM throughout (#3464), which in turn required a migration to Vitest (#3555). We also took the opportunity to bump passportjs to latest (#3502). 

See also #3221, which I used as a development/demo pizza, and has a much more granular commit structure.

## TODO

- [x] Resolve various TODOs and polish
- [x] Resolve issue with Jest (see below)
- [ ] Add [front channel logout](https://learn.microsoft.com/en-us/entra/identity-platform/v2-protocols-oidc#how-to-set-up-single-sign-out-using-front-channel-logout-feature) to OIDC setup
- [ ] Further security hardening (e.g. [validating ID tokens](https://openid.net/specs/openid-connect-implicit-1_0.html#IDTokenValidation))
- [ ] Add unit and integration tests
- [ ] Document changes in README and as a new ADR (including Azure/Entra setup)

We will address anything not included here in further PRs.

### Jest/ESM issue ✅ 

The stubborn issue here is that Jest (or `ts-jest`, the transformer we're using to transpile `.ts` files into `.js` files that Jest can read) will not treat files as ES modules (despite following [docs](https://kulshekhar.github.io/ts-jest/docs/guides/esm-support/) to the letter), and hence [throws errors like the following](https://github.com/theopensystemslab/planx-new/actions/runs/10409160785/job/28828151209?pr=3453):

```
 FAIL  ./server.test.js
  ● Test suite failed to run

    server.ts:126:18 - error TS1378: Top-level 'await' expressions are only allowed when the 'module' option is set to 'es2022', 'esnext', 'system', 'node16', 'nodenext', or 'preserve', and the 'target' option is set to 'es2017' or higher.

    126 const passport = await getPassport();
                         ~~~~~
```

I examine some of the background for this in #3528, ~onto which this PR is chained~.

Below is a non-exhaustive list of possible approaches to this problem, which I've marked complete if they've been fully explored and not provided a solution:

- [x] Use an appropriate `ts-jest` [preset](https://kulshekhar.github.io/ts-jest/docs/getting-started/presets/) instead of manual config
- [x] Try [latest Jest v30 alpha](https://github.com/jestjs/jest/releases)  - for example, `alpha5` [includes a fix](https://github.com/jestjs/jest/pull/14739) which purports to 'allow Node16/NodeNext/Bundler moduleResolution in project's tsconfig', which seems relevant
- [x] Change config file to `jest.config.js`
- [x] Take seriously the idea (suggested by above error) that Jest is failing to read the tsconfig at all
- [x] Switch off above [diagnostic message](https://github.com/microsoft/TypeScript/blob/5fd6a6fd8d48be283528b73f0727d4816d97d65b/src/compiler/diagnosticMessages.json#L1223), as @jessicamcinchak did to resolve a different problem in #3507 - doing so [reveals](https://github.com/theopensystemslab/planx-new/actions/runs/10426634170/job/28879939976?pr=3453) `ReferenceError: await is not defined`, which only adds to our case that modules are not being treated as ESM
- [x] Implement `verbatimModuleSyntax` (didn't resolve issue, but was edifying - see #3528)
- [ ] Replace `ts-jest` with `@swc/jest` or `babel-jest` (failing examples in `api-add-microsoft-sso-swc` and `api-add-microsoft-sso-babel` respectively) - could include using multiple transformers, e.g. babel for `.js`, ts-jest for `.ts`
- [ ] Try using `.mjs`/`.mts` file endings (not a desirable outcome)
- [ ] Just let Jest consider whole codebase as commonJS
- [ ] Replace top-level await with IIFE (would feel like a retreat, but may be necessary...)

Worth noting that even Jest advises that [ESM support](https://jestjs.io/docs/ecmascript-modules) is experimental, and indeed [their issue tracking this](https://github.com/jestjs/jest/issues/9430) is very much still open. This could suggest we should lean to some of the latter, more climb down-y options.

🏁 In the end we **abandoned Jest in favour of Vitest**, which is much more ESM-friendly! See #3555 :)


api.planx.uk/modules/auth/strategy/microsoft-oidc.ts
```diff
@@ -0,0 +1,69 @@
...
+    },
+    async (req: any, tokenSet: TokenSet, done: any): Promise<void> => {
+      // TODO: use state to pass the redirectTo query param through the auth flow
+      const state = tokenSet.state;
+
+      const claims = tokenSet.claims();
+      const email = claims.email;
+      const returned_nonce = claims.nonce;
+      // we grab login_hint to provide to the logout endpoint later (as per OpenID spec)
+      const login_hint = claims.login_hint;
```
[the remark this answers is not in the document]
freemvmt: Actually based on discussion on Slack we aren't going to do this account-wide logout just because someone logs out of PlanX - although I will implement the 'front-end logout channel' so that the opposite is true (i.e. when someone logs out of their MS account, they should also be logged out of PlanX).

api.planx.uk/modules/auth/passport.ts
```diff
@@ -0,0 +1,38 @@
...
+  customPassport.use(
+    "microsoft-oidc",
+    getMicrosoftOidcStrategy(microsoftOidcClient),
+  );
+
+  // do other aspects of passport setup which can be handled here
+  customPassport.use("google", googleStrategy);
+  customPassport.serializeUser((user: any, done) => {
+    done(null, user);
+  });
```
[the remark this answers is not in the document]
freemvmt: great tip thanks!

freemvmt:
Rebased this PR onto #3528 - full explanation given in that PR description.

freemvmt:
Rebased PR onto #main now that it includes #3555, which makes the #3528 redundant for the purpose of fixing the Jest issue. Force pushed simplified commits.

api.planx.uk/server.ts
```diff
@@ -122,19 +121,16 @@ app.use(
...
-
-passport.serializeUser(function (user, cb) {
-  cb(null, user);
-});
-
-passport.deserializeUser(function (obj: Express.User, cb) {
-  cb(null, obj);
-});
+// equip passport with auth strategies early on, so we can pass it to route handlers
+const passport = await getPassport();
```
freemvmt, line 125: the fabled top-level await!

freemvmt (commented):
Some notes

api.planx.uk/modules/auth/strategy/microsoft-oidc.ts
```diff
@@ -0,0 +1,69 @@
...
+        response_mode: "form_post",
+      },
+      // need the request in the verify callback to validate the returned nonce
+      passReqToCallback: true,
+    },
+    verifyCallback,
+  );
+};
+
+const verifyCallback: StrategyVerifyCallbackReq<Express.User> = async (
```
freemvmt, line 43: Does this seem correct? curious to get a read from someone with more TS nous 🧠

api.planx.uk/server.ts
```diff
@@ -200,4 +197,12 @@ declare global {
       };
     }
   }
+
+  namespace Http {
```
freemvmt, line 201: In particular is this a hacky solution? I'm not sure if it's a like for like case compared with the `Express` namespace for which we are already using (abusing?) this technique :)

freemvmt (commented):
And a query requiring some typescript tekkers

api.planx.uk/modules/auth/strategy/microsoft-oidc.ts
DafyddLlyr: This looks right to me 👍

api.planx.uk/server.ts
DafyddLlyr: I think this is necessary here and probably the right way to go 👍

DafyddLlyr:
Merging to `main` - will pick up Pulumi vars in a follow up PR shortly, and outstanding items from the checklist in description can follow along also.

DafyddLlyr (approved):

PR #3454, opened 2024-07-25
feat: Add Props to MapAndLabel Component

RODO94:
## What does this PR do?

This PR looks to extend the new ``MapAndLabel`` component by adding a ``my-map`` component to the Public file and introduce the passing of props from the Editor Modal version to the Public version to control how the map is formatted.

Two props have been tested

- drawColour
- drawType

drawColour is also used to dictate ``drawFillColour`` and ``drawPointColour`` which are built in props for the ``my-map`` component.

Following this PR, I will start to add more data interactivity with what is happening on the graph and how it updates other elements of the component as shown in the ``FindProperty`` component as an example then continue to integrate work being undertaken on the map repo.

### DIFF
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
@@ -1,7 +1,10 @@
 import ColorLensIcon from "@mui/icons-material/ColorLens";
+import FormControl from "@mui/material/FormControl";
+import RadioGroup from "@mui/material/RadioGroup";
+import { getContrastRatio, useTheme } from "@mui/material/styles";
 import { ComponentType as TYPES } from "@opensystemslab/planx-core/types";
 import { useFormik } from "formik";
-import React from "react";
+import React, { useState } from "react";
 import ColorPicker from "ui/editor/ColorPicker";
 import InputGroup from "ui/editor/InputGroup";
 import ModalSection from "ui/editor/ModalSection";
@@ -11,7 +14,7 @@ import Input from "ui/shared/Input";
 import InputRow from "ui/shared/InputRow";
 import InputRowItem from "ui/shared/InputRowItem";
 
-import { MapContainer } from "../shared/Preview/MapContainer";
+import BasicRadio from "../shared/Radio/BasicRadio";
 import { EditorProps, ICONS, InternalNotes, MoreInformation } from "../ui";
 import { MapAndLabel, parseContent } from "./model";
 
@@ -20,9 +23,22 @@ type Props = EditorProps<TYPES.MapAndLabel, MapAndLabel>;
 export default MapAndLabelComponent;
 
 function MapAndLabelComponent(props: Props) {
+  const theme = useTheme();
+
   const formik = useFormik({
     initialValues: parseContent(props.node?.data),
+    validate: ({ lineColour }) => {
+      const isContrastThresholdMet =
+        getContrastRatio("#FFF", lineColour) > theme.palette.contrastThreshold;
+      if (!isContrastThresholdMet) {
+        return {
+          lineColour:
+            "Theme colour does not meet accessibility contrast requirements (3:1)",
+        };
+      }
+    },
     onSubmit: (newValues) => {
+      console.log({ onSubmit: props });
       props.handleSubmit?.({
         type: TYPES.MapAndLabel,
         data: newValues,
@@ -62,34 +78,49 @@ function MapAndLabelComponent(props: Props) {
                 name="fn"
                 placeholder={"Data Field"}
                 value={formik.values.fn}
+                onChange={formik.handleChange}
                 required
               />
             </InputRow>
           </InputGroup>
         </ModalSectionContent>
-      </ModalSection>
-      <ModalSection>
-        <ModalSectionContent title="Formatting" Icon={ColorLensIcon}>
+        <ModalSectionContent title="Map Formatting">
           <InputGroup>
             <InputRow>
               <InputRowItem>
-                <ColorPicker label="Line Colour" />
-              </InputRowItem>
-            </InputRow>
-            <InputRow>
-              <InputRowItem>
-                <ColorPicker label="Fill Colour" />
+                <ColorPicker
+                  label="Line Colour"
+                  color={formik.values.lineColour}
+                  onChange={(color) => {
+                    formik.setFieldValue("lineColour", color);
+                  }}
+                  errorMessage={formik.errors.lineColour}
+                />
               </InputRowItem>
             </InputRow>
           </InputGroup>
-          <MapContainer environment="standalone">
-            <my-map
-              drawMode
-              drawPointer="crosshair"
-              zoom={20}
-              showCentreMarker
-            />
-          </MapContainer>
+        </ModalSectionContent>
+        <ModalSectionContent title="Map Drawing Type">
+          <FormControl component="fieldset">
+            <RadioGroup defaultValue="default" value={formik.values.drawType}>
+              {[
+                { id: "Polygon", title: "Polygon" },
+                { id: "Point", title: "Point" },
+              ].map((type) => (
+                <BasicRadio
+                  key={type.id}
+                  id={type.id}
+                  title={type.title}
+                  variant="compact"
+                  value={type.id}
+                  onChange={(e: React.SyntheticEvent<Element, Event>) => {
+                    const target = e?.target as HTMLInputElement;
+                    formik.setFieldValue("drawType", target.value);
+                  }}
+                />
+              ))}
+            </RadioGroup>
+          </FormControl>
         </ModalSectionContent>
       </ModalSection>
       <MoreInformation
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
@@ -80,11 +80,11 @@ function MapAndLabelComponent(props: Props) {
                 <ColorPicker
                   label="Drawing Colour"
                   inline={false}
-                  color={formik.values.drawColour}
+                  color={formik.values.drawColor}
                   onChange={(color) => {
-                    formik.setFieldValue("drawColour", color);
+                    formik.setFieldValue("drawColor", color);
                   }}
-                  errorMessage={formik.errors.drawColour}
+                  errorMessage={formik.errors.drawColor}
                 />
               </InputRowItem>
             </InputRow>
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
@@ -33,7 +33,7 @@ function MapAndLabelComponent(props: Props) {
       if (!isContrastThresholdMet) {
         return {
           drawColour:
-            "Theme colour does not meet accessibility contrast requirements (3:1)",
+            "Drawing colour does not meet accessibility contrast requirements (3:1)",
         };
       }
     },
@@ -88,7 +88,7 @@ function MapAndLabelComponent(props: Props) {
             <InputRow>
               <InputRowItem>
                 <ColorPicker
-                  label="Line Colour"
+                  label="Drawing Colour"
                   color={formik.values.drawColour}
                   onChange={(color) => {
                     formik.setFieldValue("drawColour", color);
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
```diff
@@ -17,6 +18,19 @@ function MapAndLabelComponent(props: Props) {
         policyRef={props.policyRef}
         howMeasured={props.howMeasured}
       />
+      <MapContainer environment="standalone">
+        <my-map
+          drawMode
+          drawPointer="crosshair"
+          zoom={20}
+          // adding 1A to the hexcode string adds 10% opacity to the colour
+          drawFillColor={props.drawColour + "1A"}
```
RODO94, line 27: Here, I tried to use the ``lighten`` method from material UI Core but it kept returning an rgb string, ``rbg(10, 10, 10)`` which kept throwing an error. I could then pass it through a rgbToHex fn but it seemed more complicated, and maybe the note creates the clarity on what's happening?

RODO94: Following this, I done more research and found ``alpha()`` which returned an `rgba(10, 10, 10) `string which is being accepted by the `<my-map>` component, I'll move forward with this approach

editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
```diff
@@ -16,8 +23,20 @@ type Props = EditorProps<TYPES.MapAndLabel, MapAndLabel>;
...
+    validate: ({ drawColour }) => {
+      const isContrastThresholdMet =
+        getContrastRatio("#FFF", drawColour) > theme.palette.contrastThreshold;
+      if (!isContrastThresholdMet) {
+        return {
+          drawColour:
+            "Drawing colour does not meet accessibility contrast requirements (3:1)",
+        };
+      }
+    },
```
jessicamcinchak, line 39: This is a cool formik validation proof-of-concept & nice re-use of this from team themes. 

Just for complete clarity here though:
- The draw color is being tested against white, not the vector tiles basemap which is a range of grays/beige etc (the basemap itself we know doesn't pass contrast requirements!) so this might not always hold up? Especially thinking of yellows/oranges etc
- We'd also want to re-think the baseline color once we introduce satellite view which will probably favor lighter draw colors on typically dark green background imagery ! 
- Interactive maps are exempt from a11y audits though so it's not a thing we need to get exactly right or necessarily have in place at MVP stage!

editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
```diff
@@ -58,11 +77,50 @@ function MapAndLabelComponent(props: Props) {
                 name="fn"
                 placeholder={"Data Field"}
                 value={formik.values.fn}
+                onChange={formik.handleChange}
                 required
               />
             </InputRow>
           </InputGroup>
         </ModalSectionContent>
+        <ModalSectionContent title="Map Formatting">
```
jessicamcinchak, line 86: nit: I think most pages & headers throughout Planx favor only capitalising the letter of the first word, eg `Map formatting` (very big nit pick sorry)

editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
```diff
@@ -58,11 +77,50 @@ function MapAndLabelComponent(props: Props) {
...
+          <InputGroup>
+            <InputRow>
+              <InputRowItem>
+                <ColorPicker
+                  label="Drawing Colour"
+                  color={formik.values.drawColour}
+                  onChange={(color) => {
+                    formik.setFieldValue("drawColour", color);
```
jessicamcinchak, line 94: The mix of `ou` & `o` here feels easy to trip over in the future - we should probably pick one way for consistency?

editor.planx.uk/src/@planx/components/MapAndLabel/model.ts
```diff
@@ -2,15 +2,19 @@ import { MoreInformation, parseMoreInformation } from "../shared";
 
 export interface MapAndLabel extends MoreInformation {
   fn: string;
-  title?:string;
-  description?:string;
+  title: string;
+  description?: string;
+  drawColour: string;
+  drawType?: "Polygon" | "Point";
```
jessicamcinchak, line 8: Is there a reason `drawType` is optional but `drawColour` is required here? Can they both be required with smart defaults (which looks like how you've set them up below!)?

editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
```diff
@@ -17,6 +19,18 @@ function MapAndLabelComponent(props: Props) {
...
         howMeasured={props.howMeasured}
       />
+      <MapContainer environment="standalone">
+        <my-map
+          drawMode
+          drawPointer="crosshair"
+          zoom={20}
+          drawFillColor={alpha(props.drawColour, 0.1)}
+          drawColor={props.drawColour}
+          drawPointColor={props.drawColour}
```
jessicamcinchak, line 29: Really like how you're using a single draw color input to cross-populate all necesary style props here - also feels like good inspiration to think about deprecating some of those redundant map props & simplifying at the source if we can :+1:

editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
```diff
@@ -17,6 +19,18 @@ function MapAndLabelComponent(props: Props) {
...
+          showCentreMarker
+          drawType={props.drawType}
+        />
```
jessicamcinchak, line 32: nit: this map is currently using a default center point in central London, but we probably want it to use the team boundary so it's a local orientation when you land here! See FindProperty's map as an example of this

I personally would also omit `showCentreMarker` as it feels a bit confusing when plotting a point!

editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
```diff
@@ -58,11 +77,50 @@ function MapAndLabelComponent(props: Props) {
...
+                  }}
+                  errorMessage={formik.errors.drawColour}
+                />
+              </InputRowItem>
+            </InputRow>
+          </InputGroup>
+        </ModalSectionContent>
+        <ModalSectionContent title="Map Drawing Type">
```
jessicamcinchak, line 102: nit: I think we want `Map drawing type` to be associated as an input label to the radio group rather than a modal section content header?

jessicamcinchak (commented):
Very cool to configure a map via the editor ! :pushpin: :art: 

Functionality all working as expected, left a few questions & comments & suggestions throughout code!

editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
RODO94: Feels like the American spelling is most used in the map component so we should move forward with that?

jessicamcinchak: Yep I think aligning editor to map makes sense here, and we have majority `color` in planx-new too because dependencies like MUI etc !

editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
RODO94: @jessicamcinchak so it reads in what team the flow is in and uses the ``boundaryBbox`` from the corresponding team to clip the GeoJSON?

editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
RODO94: For this, it looks like user facing text uses ``Colour`` and variable names and text in the code base uses ``color`` so I'll follow that pattern and switch my variable names to ``color`` but keep display text as ``colour``

editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
jessicamcinchak: Yes exactly - it clips the extent of the underlying vector tiles in the basemap so they extend only a certain distance beyond the team's boundary geojson, and centers the viewport of the map based on the boundary geojson :+1:

RODO94: Cool, from the way the sidebar preview updates and when the team is set, that the team boundary will only be used when you only open the flow in a new draft, preview, or published tab? @jessicamcinchak

RODO94: Just trying to see if I should ``fetchCurrentTeam`` so the sidebar preview can have the boundary, or if it's the standard pattern to only have that specific feature in the new tab previews/drafts etc...

RODO94: @jessicamcinchak would be worth testing this new functionality, make sure I've got it working correctly with the orientation after its clipped, or if there's more I should do. I didn't end up in the middle of the local authority boundary when I tested Bucks, but this may be me having more optimistic expectations

jessicamcinchak: This works for me as expected - there's a known issue that's repeatedly come up in #help-issues related to the center being off that we need to look further into separately, but I'm generally in the council I expect and can't pan the map outside of it, which is the clip we want! https://trello.com/c/TAY09Mpw/2980-diagnose-and-fix-map-center-zoom-issue-when-displaying-team-boundaries

editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
```diff
@@ -17,6 +39,22 @@ function MapAndLabelComponent(props: Props) {
...
+          zoom={16}
+          drawFillColor={alpha(props.drawColor, 0.1)}
+          drawColor={props.drawColor}
+          drawPointColor={props.drawColor}
+          drawType={props.drawType}
+          clipGeojsonData={
+            teamSettings !== undefined
+              ? JSON.stringify(teamSettings.boundaryBBox)
+              : JSON.stringify(defaultBboxJSON)
+          }
```
jessicamcinchak, line 55: I think we could make this check a little tighter - eg right now it looks like it's possible for `teamSettings` to be definied, but `teamSettings.boundaryBBox` to not be, and for the default to never be hit? 

Maybe this would result in more intended fallback behavior here: 
```
teamSettings?.boundaryBBox 
  ? JSON.stringify(teamSettings.boundaryBBox) 
  : JSON.stringify(defaultBboxJSON)
```

editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
```diff
@@ -1,13 +1,35 @@
...
+          [1.9134116, 49.528423],
+          [1.9134116, 61.331151],
+          [1.9134116, 61.331151],
+          [-10.76418, 61.331151],
+          [-10.76418, 49.528423],
+        ],
+      ],
+    },
+    properties: {},
+  };
```
jessicamcinchak, line 31: nit: it'd be helpful to have a code comment that explains what this extent is and where you got it (eg full UK? or only England?)

We can also just pass `undefined` to the `clipGeojsonData` prop & the map repo will handle it accordingly ! https://github.com/theopensystemslab/map/blob/main/src/components/my-map/index.ts#L213-L218

editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
jessicamcinchak: Or simply `clipGeojsonData={teamSettings?.boundaryBBox && JSON.stringify(teamSettings.boundaryBbox)}` if we forego this default per comment above? 

(Feels like ideal scenario would be handling default at `teamSettings` database level, not within this component.)

jessicamcinchak (approved):

editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
RODO94: Aw yeah, that's great! Added the default as a fallback because it was still throwing an error when I tried to return undefined, but now I see adding the ``teamSettings?`` resolves this... 

Great feedback, thanks!

editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
RODO94: Removed this now

RODO94:

PR #3455, opened 2024-07-25
style: Tighten editor interface layout

ianjon3s:
## What does this PR do?

Ahead of producing more detailed Figma components the editor screens I wanted to update the editor general interface so that it closer matches existing designs.

Summary of changes:
- The current editor uses the same header parent styles as the public interface, causing it to have a header min-height that is unnecessarily tall. By introducing a dedicated container for each environment we can allow each to have an optimised height
- Slight reduction of breadcrumb type size to allow for longer council and service names on narrower screens (breadcrumbs are close to wrapping on smaller laptop sizes)
- Introduction of new `body3` type style that can be used for the smaller text sizes used in the editor (equates to around 14px, smaller than we'd generally use for the public interface but very useful for editor components)
- Using the `body3` style for the editor nav menu, the width can be reduced slightly

### DIFF
--- a/editor.planx.uk/src/components/EditorNavMenu.tsx
+++ b/editor.planx.uk/src/components/EditorNavMenu.tsx
@@ -24,8 +24,8 @@ interface Route {
   accessibleBy: Role[];
 }
 
-const MENU_WIDTH_COMPACT = "52px";
-const MENU_WIDTH_FULL = "178px";
+const MENU_WIDTH_COMPACT = "51px";
+const MENU_WIDTH_FULL = "164px";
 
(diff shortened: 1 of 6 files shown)


RODO94:
@ianjon3s I think your first bullet point ends with a cliffhanger, but I do like the new header min-height. It looks sleek.

RODO94 (approved):
Compared with the pizza and my local branch, it's lovely to have the extra space and everything looks good to me. Nothing breaking or looking strange

ianjon3s:
Thanks for the review @RODO94 ! I've updated the first sentence for posterity

PR #3445, opened 2024-07-19
feat: Flow Name Duplicate Validation

RODO94:
## What does this PR do?

Following work done on Team Name in #3441 to validate duplicate team names, the same logic has been carried into Flow Name. 

Greater detail on the thinking behind the code is given in the previously mentioned PR, #3441.

PR #3457, opened 2024-07-25
feat: Add duplicate name logic to Rename flow onClick

RODO94:
## What does this PR do?

This PR builds on previous work completed in #3445 to add duplicate name validation when you add a new flow.

What that previous PR didn't pick up was adding the validation to the renaming process.

In order to allow a comparison between the new name and the current list of names, I had to alter the props being passed to the ``<FlowItem>`` component which contains the ``<StyledSimpleMenu>``. 

The alteration was just to add ``flows`` into the component.

### DIFF
--- a/editor.planx.uk/src/pages/Team.tsx
+++ b/editor.planx.uk/src/pages/Team.tsx
@@ -128,6 +128,7 @@ export function AddButton({
 
 interface FlowItemProps {
   flow: any;
+  flows: any;
   teamId: number;
   teamSlug: string;
   refreshFlows: () => void;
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/pages/Team.tsx
```diff
@@ -128,13 +128,15 @@ export function AddButton({
 
 interface FlowItemProps {
   flow: any;
+  flows: any;
```
RODO94, line 131: Use of the ``any`` type followed current practices in the ``Team`` component, may be worth a refactor in future to follow TS best practices

jessicamcinchak (approved):
Works as expected for me ! Appreciate you double-checking post-merge and finding this edge case :raised_hands:

PR #3459, opened 2024-07-25
feat: Reduced colour fill print css

ianjon3s:
## What does this PR do?

Quick one — there has been a lot of talk about users printing from PlanX recently.

If someone does print from the public interface we can save them a lot of printer ink with a few lines of CSS. This PR uses the `@media print` media query to remove the largest blocks of solid colour from the header and footer.

**Before:**
image

**After:**
image


DafyddLlyr (approved):
Nice! Very simple and effective.

PR #3464, opened 2024-07-26
[api] adopt ESM throughout

freemvmt:
This PR does the following:

- in tsconfig, bumps `lib` and `target` to `esnext`
- in tsconfig, bumps `module` and `moduleResolution` to `nodenext`
- adds `type: module` in `package.json`, as well as `exports` and `engines` fields for clarity
- adds `.js` file endings to all imports for ESM compliance
- makes significant changes to the Jest config in order to play nicely with ESM
- similarly adapts the test scripts (e.g. `pnpm test`) to use ESM (with `--experimental-vm-modules` flag)
- replaces `ts-node-dev` (which is [incompatible with ESM](https://typestrong.org/ts-node/docs/recipes/watching-and-restarting)) with `tsx`
- replaces any instances of `require(...)` with top-line `import * as x from ...` statements

The motivation here is to enable top-level await statements, which is required for the Microsoft SSO work (see #3453).

Will be best reviewed commit-by-commit. 17125bf3 is small, handling the initial `tsconfig.json` changes to force adoption of ESM throughout, while 89c2ee68 is *huge* but fairly straightforward.

5cae4845 is perhaps the meatiest, and also required the changes in [planx-core#465](https://github.com/theopensystemslab/planx-core/pull/465) to get tests running. Note that [planx-core#464](https://github.com/theopensystemslab/planx-core/pull/464) was also prompted by this work, enabling the fixes to imports in b29fab9a.

**Resources**

- https://jestjs.io/docs/ecmascript-modules
- https://kulshekhar.github.io/ts-jest/docs/guides/esm-support/
- https://dev.to/mangadev/set-up-a-backend-nodejs-typescript-jest-using-es-modules-1530
- https://github.com/kulshekhar/ts-jest/issues/1057#issuecomment-1068342692
- https://tsx.is/typescript
- https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

**Still to investigate** (possibly after this PR is merged)

- [x] Check the codebase can now handle a top-level await, as I intend to use in `api-add-microsoft-sso-strategy-2`?
- [ ] If so, is there a good reason why I can't directly replace `require(...)` statements with `await import(...)` where they occur (what I did in 3f221a6, but had to further adapt in 6aa01c86)? And if not is ts-jest still just configured incorrectly/being buggy, are the files being interpreted as cjs still, or is it something else?
- [x] ~Should I be able to import files as `.ts`, rather than `.js`? As I understand, this would [only be possible](https://www.typescriptlang.org/tsconfig/#allowImportingTsExtensions) if we weren't usins TS to 'emit' JS. One option to explore [could be to use](https://github.com/microsoft/TypeScript/pull/51669) `moduleResolution: "bundler"` instead of "nodenext".~

Note that although some changes may survive type checking, and even build, this doesn't necessarily imply that the tests will check out. The code needs to be solid from various angles:

- Type checking, as well as building the actual application, rely directly on `tsc` (which reads `tsconfig`)
- Tests (Jest) relies on `ts-jest` to transform files for execution (which reads `tsconfig.test`, an extension of `tsconfig`)
- The dev environment (`pnpm dev`) relies on `tsx` (which also reads the `tsconfig`)


api.planx.uk/modules/auth/service.ts
```diff
@@ -1,4 +1,4 @@
-import { sign } from "jsonwebtoken";
+import jsonwebtoken from "jsonwebtoken";
```
DafyddLlyr, line 1: You're not doing anything wrong here or missing anything, this is just a CJS library without named exports.

See https://github.com/auth0/node-jsonwebtoken/pull/884

nit: `import jwt from "jsonwebtoken";` is a bit more terse and still meaningful?

api.planx.uk/modules/file/service/uploadFile.ts
```diff
@@ -1,6 +1,6 @@
 import S3 from "aws-sdk/clients/s3.js";
 import { customAlphabet } from "nanoid";
-import { getType } from "mime";
+import mime from "mime";
```
DafyddLlyr, line 3: We could probably keep the current functionality by upgrading to v4 which drops CJS https://github.com/broofa/mime/blob/v4.0.0/CHANGELOG.md

api.planx.uk/helpers.ts
```diff
@@ -1,5 +1,5 @@
 import { gql } from "graphql-request";
-import { capitalize } from "lodash";
+import lodash from "lodash";
```
DafyddLlyr, line 2: There's a few options here to avoid us having to import all of lodash - 

- Cherry pick import via named path, i.e. `import capitalize from "lodash/capitalize";` ([docs](https://lodash.com/))
- Switch to [lodash-es](https://www.npmjs.com/package/lodash-es)
- Install named functions as packages, i.e. add `lodash/capitalize` to `package.json`

I think `lodash-es` might be the best/simplest option? Happy for this to be a follow up PR or to be picked up here - whatever works best for you.

api.planx.uk/package.json
```diff
@@ -3,9 +3,10 @@
   "license": "MPL-2.0",
   "private": true,
   "packageManager": "pnpm@8.6.6",
+  "type": "module",
   "dependencies": {
     "@airbrake/node": "^2.1.8",
-    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#6b2fd26",
+    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#dd24b0a",
```
DafyddLlyr, line 9: This may now be an outdated commit hash - worth checking!

DafyddLlyr (approved):
Thanks for bearing with me whilst I got through this one! A bit more complexity than initially anticipated, but certainly a worthwhile change to have made for a whole host of reasons.

Looking great and working as expected. A few comments which can be fixed here or as follow up PRs - very much your call.

api.planx.uk/modules/file/service/uploadFile.ts
freemvmt: Interestingly bumping `mime` doesn't actually enable us to revert this change (see issue #[295](https://github.com/broofa/mime/issues/295)), but I went ahead and did it anyway :)

api.planx.uk/helpers.ts
freemvmt: Tried out `lodash-es` but I'm sure it was implemented very well, and it turned out to be easier to go with your first suggestion!

api.planx.uk/modules/file/service/uploadFile.ts
freemvmt: Bumping to v4 actually threw up a different error, which I think is related to the package* rather than anything we're doing, so I used the workaround referenced [here](https://github.com/broofa/mime/issues/300#issuecomment-1909792619).

* not sure on specifics, but basically they've tried to drop support for commonJS and go pure ESM, but I think they haven't set it up quite right (e.g. could be something as simple as `allowJs` missing from the [tsconfig](https://github.com/broofa/mime/blob/main/tsconfig.json)), resulting in ye olde 'Cannot use import statement outside a module' error

api.planx.uk/package.json
freemvmt: Bumped this to latest and squashed into same commit as before :)

freemvmt:
> Thanks for bearing with me whilst I got through this one! A bit more complexity than initially anticipated, but certainly a worthwhile change to have made for a whole host of reasons.
> 
> Looking great and working as expected. A few comments which can be fixed here or as follow up PRs - very much your call.

Thanks! Definitely picked up a lot of nuances about TypeScript etc while moving through this - have tried to list any particularly helpful resources in the PR description for posterity.

I've addressed all your comments and rebased onto main, so I think this should be good to go? Maybe just have a quick look at my changes (commit e5ff9fa4) ✨

jessicamcinchak (approved):
Latest commit looks good to me! Lots of good stuff here, thanks for working through it all & carefully noting so many docs/resources :tada: 

Let's get a prod deploy through first this morning (queued up here #3503), then this can hang out on staging for a few days into early next week just in case anything comes up :slightly_smiling_face:

PR #3465, opened 2024-07-26
style: External portals within search results

ianjon3s:
## What does this PR do?

- Reduces visual weight of external portals within search results
- Uses a `<a>` tag for the links, allowing for opening in new tabs (right/ctrl click as native behaviour)

**Existing (left) vs updated (right):**
image

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search.tsx
@@ -3,7 +3,7 @@ import Container from "@mui/material/Container";
 import List from "@mui/material/List";
 import ListItem from "@mui/material/ListItem";
 import ListItemButton from "@mui/material/ListItemButton";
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search.tsx
```diff
@@ -191,34 +187,24 @@ const ExternalPortalList: React.FC = () => {
...
-      </List>
+      </PortalList>
     </Box>
   );
 };
 
 interface SearchNodes {
   input: string;
-  facets: ["data.fn", "data.val"],
+  facets: ["data.fn", "data.val"];
```
ianjon3s, line 207: @DafyddLlyr there also appears to have been some auto-linting in play here

DafyddLlyr (approved):
Nice! If it's ok with you we'll hold off merging this to `dp/fuse-js` just now until that PR has been approved and merged, then we can repoint this to `main` in a couple of days 👍

PR #3466, opened 2024-07-29
feat: setup a Slack notification when flow status is updated

jessicamcinchak:
**Changes:**
- Adds a new Hasura event trigger whenever the `flows.status` column is updated
- This event trigger calls the existing `/webhooks/hasura/send-slack-notification` endpoint with the query param `type="flow-status"`
- Refactors a few helper methods within this API module to be less submission-event-type specific
- Will only send to Slack for updates on the production environment, staging & pizzas will be skipped

**Todo:**
- Debug zod invalid union type error
- Fetch additional data based on parsed flowId to display message with `${team_slug}/${flow_slug}`

**Open questions:**
- Do we want to omit notification if flow is within certain teams - eg Templates, Testing, Open Systems Lab ? 
- Is `#planx-notifications` channel okay or too noisy already? 
  - If too noisy already, could we actually filter what's logged there now that we have in-editor submissions log? (eg only log statutory services like LDCs so we don't miss key errors but omit building control, etc?) - or preferred to just spin up a new internal channel??


jessicamcinchak:
[Can we set up notifications for admins, to know when a new service goes 'online'?](https://trello.com/c/pznE5Sqk/2940-can-we-set-up-notifications-for-admins-to-know-when-a-new-service-goes-online)

api.planx.uk/modules/webhooks/service/sendNotification/schema.ts
```diff
@@ -75,9 +75,26 @@ export const s3SubmissionSchema = z.object({
...
+          id: z.string(),
+          status: z.string(),
+        }),
+      }),
+    }),
+  }),
+  query: z.object({
+    type: z.literal("flow-status"),
+  }),
+});
```
DafyddLlyr, line 92: Just taken a closer look at the code here to see what's causing the Zod union error on the Pizza.

There's nothing wrong with the code - pretty certain of it! 

Running locally (and in API tests on CI) this works as expected - I think it's actually a Vultr build issue maybe? The error message returned to Hasura doesn't include this schema as a member of the union. I'll recreate the pizza just now to see if this resolves it.

DafyddLlyr: Recreating Vultr gives the same error. Next thought - do we have some sort of env var issue here - is it pointing at the staging Hasura instance

DafyddLlyr: Nope, just checked out the Pizza using `docker inspect` and it's set to the expected value `HASURA_PLANX_API_URL=http://api:7002` 🤔

jessicamcinchak: Thanks for the second set of eyes here - it's a weird one !! 

Are you actually getting a succeeding `200` event locally? I have passing API tests, but failing local event similar to pizza.

Another thought - could it be a bad migration or error in the yaml?? I "updated" the event definition in the console at least once which seemed to correctly update `tables.yaml`? I'll try a docker destroy/recreate locally on this branch to trigger the migration again just to confirm

DafyddLlyr: Yeah I was getting past the validation middleware issue locally, and hitting the "staging application, skipping post to Slack" message which I was expecting to see.

There is an issue with the YAML docs here - there's an additional layer of `body` wrapping in the payload in the docs (issue not introduced by you!), and the endpoint is actually kebab-case in reality, not camelCase as in the docs. Was going to push up a small PR to address both of these points but happy to jump on a quick call to take a look here if helpful.

DafyddLlyr: Please see API doc fixes at https://github.com/theopensystemslab/planx-new/pull/3485

api.planx.uk/modules/webhooks/docs.yaml
```diff
@@ -113,12 +113,34 @@ components:
                           type: string
       required:
         - body
+      FlowStatusChange:
+        type: object
+        properties:
+          body:
+            type: object
```
DafyddLlyr, line 120: This body wrapper isn't required - changes to other events in the Swagger docs already on staging.

DafyddLlyr (approved):
Code reads as correct, tests passing, and working locally. It looks like a cache / vultr build issue. 

I'm super happy to merge to staging to test this there 👍

PR #3469, opened 2024-07-29
feat: Add event listener trackers to Map and Label

RODO94:
## What does this PR do?

This PR begins to add event listeners and map dispatches to the public component of ``MapAndLabel`` which can then lay the foundations for Work to Trees and drawing shapes on a map and tracking them with List components. 

- [x] Track Geometry and Area
- [x] Add light validation 
- [x] Generate a dummy card with Geo data

### DIFF
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
@@ -4,6 +4,7 @@ import { Feature } from "geojson";
 import { useStore } from "pages/FlowEditor/lib/store";
 import React, { useEffect, useState } from "react";
 import { FONT_WEIGHT_SEMI_BOLD } from "theme";
+import ErrorWrapper from "ui/shared/ErrorWrapper";
 
 import Card from "../shared/Preview/Card";
 import CardHeader from "../shared/Preview/CardHeader";
@@ -13,13 +14,14 @@ import { MapAndLabel } from "./model";
 
 type Props = PublicProps<MapAndLabel>;
 
-export type Boundary = Feature | undefined;
+type Boundary = Feature | undefined;
 
 function MapAndLabelComponent(props: Props) {
   const teamSettings = useStore.getState().teamSettings;
   const [boundary, setBoundary] = useState<Boundary>();
   const [area, setArea] = useState<number | undefined>(0);
   const [objectArray, setObjectArray] = useState<any>([]);
+  const [mapValidationError, setMapValidationError] = useState<string>();
 
   let count = 0;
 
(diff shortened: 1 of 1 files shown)


RODO94:

editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
```diff
@@ -1,17 +1,62 @@
...
 import { PublicProps } from "../ui";
 import { MapAndLabel } from "./model";
 
 type Props = PublicProps<MapAndLabel>;
 
+export type Boundary = Feature | undefined;
+
 function MapAndLabelComponent(props: Props) {
   const teamSettings = useStore.getState().teamSettings;
+  const [boundary, setBoundary] = useState<Boundary>();
```
jessicamcinchak, line 20: nit: for simplicity, I wouldn't export a type `Boundary` (unless there's a use case outside of this file??), I'd do this all in the same line like so: 
```
const [boundary, setBoundary] = useState<Feature | undefined>()
```

editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
```diff
@@ -1,17 +1,62 @@
...
+  const [area, setArea] = useState<number | undefined>(0);
+  const [objectArray, setObjectArray] = useState<any>([]);
```
jessicamcinchak, line 22: I'm not sure I understand the purpose of `objectArray` here (variable name is quite vague & it's never rendered)? Is boundary sufficient alone?

jessicamcinchak (commented):
Overall looking very good and working as expected ! 

Your PR description mentions "light validation" which I'd expect to see implemented as an `<ErrorWrapper>` around the `<MapContainer>` with a validation rule (for example, at least one feature must be drawn to "Continue" - see DrawBoundary for existing example of this). Does that sound like the right track or are you thinking about validation differently here?

editor.planx.uk/src/@planx/components/MapAndLabel/Public.tsx
RODO94: It's a way for me to track the event listener dispatching events and collecting info into an array. Most likely negated in a PR or two when multiple items get added to the map and we can use the geoJSON features, but helpful for now if I wanted to test adding a list of items to the Public comp... is it too rough to add in the PR? @jessicamcinchak

jessicamcinchak: Thanks for explanation - happy for it to merge here now better understanding it's work-in-progress :slightly_smiling_face:

jessicamcinchak (approved):

PR #3475, opened 2024-07-30
refactor: Make ``label`` conditional on ``<ColorPicker>``

RODO94:
## What does this PR do?

This PR updates the ``ColorPicker`` component to conditionally add a `label` if provided it as a prop, and don't add any label if the `label` prop doesn't exist.

The previous behaviour was to add a label stating ``"background colour"`` if no `label` prop was provided. 

To ensure continuous behaviour, I have updated the two components using ``<ColorPicker>`` without the ``label`` prop and I have updated the `MapAndLabel` component to use the ``<ColorPicker>`` without the `label` prop, which was the catalyst for this change.

### DIFF
--- a/editor.planx.uk/src/@planx/components/Content/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/Content/Editor.tsx
@@ -41,6 +41,7 @@ const ContentComponent: React.FC<Props> = (props) => {
           </InputRow>
           <ColorPicker
             inline
+            label="Background colour"
             color={formik.values.color}
             onChange={(color) => {
               formik.setFieldValue("color", color);
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Editor.tsx
@@ -13,6 +13,7 @@ import InputLabel from "ui/public/InputLabel";
 import Input from "ui/shared/Input";
 import InputRow from "ui/shared/InputRow";
 import InputRowItem from "ui/shared/InputRowItem";
+import InputRowLabel from "ui/shared/InputRowLabel";
 
 import BasicRadio from "../shared/Radio/BasicRadio";
 import { EditorProps, ICONS, InternalNotes, MoreInformation } from "../ui";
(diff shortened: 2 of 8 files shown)


DafyddLlyr (commented):
Just want to check a few things here - what's the motivation for this change? Are we aiming for a different label layout in the `MapAndLabel` component?

I think ideally we'd want consistency of styles/layout throughout the Editor as well as props which accurately describe how a component should get used. In reality this component will always need a label to be meaningful, so should the prop actually be required, but maybe support alternate layouts (if this is a requirement)?

ianjon3s (approved):
Looks good to me, all existing instances of `<ColorPicker>` have also been addressed.

RODO94:
@DafyddLlyr it was to have a layout where I could have a label above like I put here: https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1722245161220269

Definitely understand the point of keeping the label prop required and adding an ``align`` prop to change the arrangements (for example). Could also be worth discussing how input comps should be structured with the label, this particular one is using the ``RadioGroup`` and the ``ColourPicker`` so the labels would have a different type I think...

RODO94:
Following discussions in our dev call, we decided to make ``label`` a **required** prop while using another prop to alter orientation. After a bit of investigation, I decided to utilise the current ``inline`` prop which is used in the ``<Root>`` part of the component.

Thinking was that if you added inline for that, it means that's the orientation you would like to use. Carrying this logic through, I use a couple ternary expressions to alter `<Root>` styling:

```diff
const Root = styled(Box, {
  shouldForwardProp: (prop) => prop !== "inline",
})<RootProps>(({ inline, theme }) => ({
  padding: 0,
  position: "relative",
  display: "flex",
++  alignItems: inline ? "center" : "flex-start",
++  flexDirection: inline ? "row" : "column",
  ...(inline && {
    padding: theme.spacing(2, 0),
    "& .popover": {
      top: "calc(100% - 4px)",
    },
  }),
}));

```

With this change, it means that each ``<ColorPicker>`` will require an explicit ``inline`` prop to have the label and input in a row orientation:

```diff
<ColorPicker
        color={formik.values.actionColour}
++      inline
        onChange={(color) => formik.setFieldValue("actionColour", color)}
        label="Button colour"
        errorMessage={formik.errors.actionColour}
/>
```

The new ``label`` for column orientation has also received a styling change to bring it in line with ``<InputLabel>`` 

```diff
        {props.inline ? (
          <Typography mr={2} variant="body2" component="label">
            {props.label}:{" "}
          </Typography>
        ) : (
          <Typography 
++          sx={{ pb: 1 }} 
++          variant="body1" 
          component="label" >
            {props.label}
          </Typography>
        )}
```

ianjon3s (approved):
Looks good to me, and the code notation in the comments is top stuff.

PR #3479, opened 2024-07-31
fix: Sections navigation bar padding

ianjon3s:
## What does this PR do?

Fixes a regression that has removed the vertical padding from the sections navigation bar.

Before:
image

After:
image


DafyddLlyr (approved):

DafyddLlyr:
Good catch @ianjon3s 🙌

PR #3483, opened 2024-07-31
[WIP] feat: add share button for flow links

RODO94:
## What does this PR do?

This PR comes from the Trello ticker: https://trello.com/c/k8cQojk7/2958-share-service-link-in-editor

The intention is to add a **share** link or **copy** link button to the FlowEditor page for users to easily see and share the right links for their flows.

From the Trello ticket:

> _As an Editor
I want to be able to copy/paste my service URL (with custom subdomain)
so that I can share the link_

I worked through the Figma designs given here: [Figma mock-ups](https://www.figma.com/design/bnUUrsVRG6qPwDkTmVKACI/PlanX-Design?node-id=15777-6403&t=Z9BAN3CGnRhbQZNB-4)

There has been a range of components added, created, and altered. From the new styled buttons, `CopyButton` and `ToggleButton`. 

The reuse of the `ImageWrapper` component used previously in Design Settings but customised to new sizes, I thought keeping the name consistent would point at similar use but different settings, since neither are exported from their origin.

A `LinkComponent` was made (naming may change to be more explicit) to harmonise styling across all the different links in the `Dialog` box, this includes the ability to add either an icon or image or neither to the Title of a link. 

The ``ToggleButton`` component was used to move the sidebar toggle from the top header to the Sidebar, in line with the other buttons for publishing and sharing links. This has been positioned ``absolute`` with a position ``relative`` set around the box containing the buttons so it can hang over the left edge of the siderbar.

Feedback requests:

_Feel free to give me feedback on anything, but these are the particular areas I found many options in._

1. Structure and use of ``<LinkComponent``
2. The way I have used MUI components and styled them
3. Closeness to the initial brief

To Do's 

- [x] Remove icon buttons from top of sidebar and replace with buttons
- [x] Create new dialog for seeing links
- [x] Make Published link opaque if the flow has not been published
- [x] Add previous buttons for `refresh` and `console`
- [ ] Create tests for new component

### DIFF
--- a/editor.planx.uk/src/components/Header.tsx
+++ b/editor.planx.uk/src/components/Header.tsx
@@ -585,7 +585,8 @@ const Header: React.FC = () => {
 
   useEffect(() => {
     const findPathname =
-      pathnameArray.pop() === "preview" || pathnameArray.pop() === "draft";
+      pathnameArray.slice(-1)[0] === "preview" ||
+      pathnameArray.slice(-1)[0] === "draft";
 
     setIsOutsideEditor(Boolean(findPathname));
   }, pathnameArray);
--- a/editor.planx.uk/src/components/Header.tsx
+++ b/editor.planx.uk/src/components/Header.tsx
@@ -584,9 +584,9 @@ const Header: React.FC = () => {
   const pathnameArray = window.location.pathname.split("/");
 
   useEffect(() => {
-    const findPathname = pathnameArray.find(
-      (pathname) => pathname === "preview" || pathname === "draft",
-    );
+    const findPathname =
+      pathnameArray.pop() === "preview" || pathnameArray.pop() === "draft";
+
     setIsOutsideEditor(Boolean(findPathname));
   }, pathnameArray);
 
--- a/editor.planx.uk/src/components/Header.tsx
+++ b/editor.planx.uk/src/components/Header.tsx
@@ -24,7 +24,7 @@ import { clearLocalFlow } from "lib/local";
 import { capitalize } from "lodash";
 import { Route } from "navi";
 import { useAnalyticsTracking } from "pages/FlowEditor/lib/analytics/provider";
-import React, { RefObject, useRef, useState } from "react";
+import React, { RefObject, useEffect, useRef, useState } from "react";
 import {
   Link as ReactNaviLink,
   useCurrentRoute,
@@ -577,8 +577,19 @@ const Toolbar: React.FC<ToolbarProps> = ({ headerRef }) => {
 };
 
 const Header: React.FC = () => {
+  const [isOutsideEditor, setIsOutsideEditor] = useState<boolean>(false);
   const headerRef = useRef<HTMLDivElement>(null);
   const theme = useStore((state) => state.teamTheme);
+  const environment = useStore((state) => state.previewEnvironment);
+  const pathnameArray = window.location.pathname.split("/");
+
+  useEffect(() => {
+    const findPathname = pathnameArray.find(
+      (pathname) => pathname === "preview" || pathname === "draft",
+    );
+    setIsOutsideEditor(Boolean(findPathname));
+  }, pathnameArray);
+
   return (
     <Root
       position="static"
@@ -586,7 +597,10 @@ const Header: React.FC = () => {
       color="transparent"
       ref={headerRef}
       sx={{
-        backgroundColor: theme?.primaryColour || "#2c2c2c",
+        backgroundColor:
+          isOutsideEditor || environment === "standalone"
+            ? theme.primaryColour
+            : "#2c2c2c",
         "@media print": { backgroundColor: "white", color: "black" },
       }}
     >
--- a/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/LinkDialog.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/LinkDialog.tsx
@@ -16,6 +16,7 @@ import { SvgIconProps } from "@mui/material/SvgIcon";
 import Tooltip from "@mui/material/Tooltip";
 import Typography from "@mui/material/Typography";
 import gql from "graphql-tag";
+import { useStore } from "pages/FlowEditor/lib/store";
 import React, { useEffect, useState } from "react";
 
 import { client } from "../../../../lib/graphql";
@@ -193,7 +194,10 @@ const LinkContainer = (props: LinkProps) => {
 };
 
 export default function LinkDialog(props: DialogBaseProps) {
-  const [flowStatus, setFlowStatus] = useState<string | undefined>();
+  const [setFlowStatus, flowStatus] = useStore((state) => [
+    state.setFlowStatus,
+    state.flowStatus,
+  ]);
   // Retrieving flow status to determine which links to show in View Links
   useEffect(() => {
     const fetchFlowStatus = async () => {
--- a/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/LinkDialog.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/LinkDialog.tsx
@@ -16,11 +16,10 @@ import { SvgIconProps } from "@mui/material/SvgIcon";
 import Tooltip from "@mui/material/Tooltip";
 import Typography from "@mui/material/Typography";
 import gql from "graphql-tag";
-import { client } from "lib/graphql";
-import { useStore } from "pages/FlowEditor/lib/store";
-import { prop, props } from "ramda";
-import React, { FC, ReactNode, useEffect, useState } from "react";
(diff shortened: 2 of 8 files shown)


RODO94:
[Share service link in Editor](https://trello.com/c/k8cQojk7/2958-share-service-link-in-editor)

RODO94:
Using Buckinghamshire as a team is good for testing the Subdomain flow, I updated the Hasura with a close to real dummy data.

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
```diff
@@ -152,6 +191,88 @@ const DebugConsole = () => {
...
+      gap={"8px"}
+      mb={1}
+    >
+      <Box
+        display={"flex"}
+        flexDirection={"row"}
+        alignItems={"center"}
+        gap={"7px"}
+      >
+        {typeof props.titleIcon === "string" ? (
```
DafyddLlyr, line 221: This took me a minute to parse and to understand that `titleIcon` is either a URL to a logo, `undefined` or an SVG icon.

It may be a better design to make the props for this component a discriminated union of two types - one for a logo, and one for an icon. This would mean that as opposed to multiple optional properties without any relationship with each other, we can have a stricter type which better represents how this component can get used.

Here's an example - 

```ts
// Shared values which are always present
interface BaseProps { 
  title: string;
  // ...
}

// Props always required if we're representing a logo
interface LogoProps extends BaseProps {
  type: "logo";
  primaryColour: string;
  logoURL: string;
}

// Props always required if we're representing an icon
interface IconProps extends BaseProps {
  type: "icon";
  icon: SvgIconProps;
}

// Our union type
type Props =  LogoProps | IconProps;


// Examples
const validLogo: Props = {
  title: "abc",
  type: "logo",
  primaryColour: "#FFF",
  logoURL: "abc"
}

const validIcon: Props = {
  title: "abc",
  type: "icon",
  icon: <SomeIcon/>
}

const invalidType: Props = {
  title: "abc",
  type: "logo",
  icon: <SomeIcon/>
}
```

DafyddLlyr: This may seem a little complex initially, but it's a super useful pattern to learn - we're trading a little bit of type complexity upfront for a lot of type safety, a better developer experience, and clearer code (imho!) 👌

DafyddLlyr (commented):
Thanks for sharing and asking some some initial feedback on Slack. I've had a pass through this PR focusing more on the structure and code, as opposed to the design fidelity.

There's a few specific comments on lines of code below, but I have two more general comments. 

**1. Aim for one feature per-PR (or even better, many PRs to one feature)** 

This PR had two main things happening - the refactor and change of the "toggle sidebar" function, and the addition of the dialog with a link to services. 

Despite both sharing a single set of Figma designs they're both pretty atomic and separate and should be treated as such. I'd recommend taking the changes made to "toggle" out of this PR, and we can pick this up separately.

**2. Testing**

There's a fair amount of business logic at play here in the modal - 
 - Show/hide certain links to those of different permission sets
 - Construct various URLs based on domain, slug, preview, draft etc
 - Disable certain links if service is not published
 - Conditionally show logo or icon
 
It would be nice to have test coverage for this - it can be part of this PR, or come before or after, whatever works best for you. This is a good opportunity to get stuck into `react-testing-library` a little. Let me know if you need a hand with any setup here.

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
```diff
@@ -248,78 +385,92 @@ const Sidebar: React.FC<{
 
   // useStore.getState().getTeam().slug undefined here, use window instead
   const teamSlug = window.location.pathname.split("/")[1];
+  const flowSlug = window.location.pathname.split("/")[2];
```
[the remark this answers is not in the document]
RODO94: @DafyddLlyr  in order to get this working with the store, I need to add `team_theme` to the `initTeamStore` query. While doing that it complicates the `<Header>` comp styling which relies on ``theme.primaryColour`` being undefined until you hit the Published link.

```ts
const Header: React.FC = () => {
  const headerRef = useRef<HTMLDivElement>(null);
  const theme = useStore((state) => state.teamTheme);
  return (
    <Root
      position="static"
      elevation={0}
      color="transparent"
      ref={headerRef}
      sx={{
        backgroundColor: theme?.primaryColour || "#2c2c2c",
        "@media print": { backgroundColor: "white", color: "black" },
      }}
    >
      <Toolbar headerRef={headerRef}></Toolbar>
    </Root>
  );
};
```

Any thoughts on moving forward with this?

editor.planx.uk/src/components/Header.tsx
```diff
@@ -584,9 +584,9 @@ const Header: React.FC = () => {
   const pathnameArray = window.location.pathname.split("/");
 
   useEffect(() => {
-    const findPathname = pathnameArray.find(
-      (pathname) => pathname === "preview" || pathname === "draft",
-    );
+    const findPathname =
+      pathnameArray.pop() === "preview" || pathnameArray.pop() === "draft";
```
DafyddLlyr, line 588: `.pop()` is a mutating method - `pathname` won't remain consistent for both these checks.

image

image

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
```diff
@@ -152,6 +191,88 @@ const DebugConsole = () => {
...
 
+const LinkComponent = (props: {
+  primaryColour?: string;
+  titleIcon?: string | SvgIconProps;
+  title: string;
+  link: string;
+  description?: string;
+  isPublished?: boolean;
+}) => {
+  const [copyMessage, setCopyMessage] = useState<"copy" | "copied">("copy");
```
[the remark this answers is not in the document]
RODO94: What would this give me, over the current ``navigator.clipboard.writeText(props.link)``?

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
```diff
@@ -72,15 +72,27 @@ const Header = styled("header")(({ theme }) => ({
...
+  background: theme.palette.common.white,
+  border: `1px solid ${theme.palette.border.main}`,
+  boxShadow: "none",
+  color: theme.palette.common.black,
+  width: "33%",
+  display: "flex",
+  flexDirection: "row",
+  gap: "8px",
+  borderRadius: "5px",
+  padding: "8px",
```
jessicamcinchak, line 95: nit: Feels a bit surprising & brittle to me to see properties like `display`, `flexDirection`, and `gap` defined directly on a `Button` - I think I'd expect to see a wrapper `Box`/`div` with these properties that then accepts buttons as children ??

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/LinkDialog.tsx
```diff
@@ -0,0 +1,302 @@
...
+        <InactivePublishedLink
+          title="With subdomain"
+          message={"There is not a domain configured for this team"}
+        />
+      )}
+      {props.isPublished && props.status === "online" ? (
+        <ActivePublishedLink title="PlanX link" link={props.link} />
+      ) : (
+        <InactivePublishedLink title="PlanX link" message={props.link} />
+      )}
```
jessicamcinchak, line 148: I find this ternary very difficult to read, and similar to another comment would much prefer more verbose `if/else` or `switch` syntax here to describe the expected conditional rendering rules. 

Also a great example where tests would help describe expected behavior here !

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/LinkDialog.tsx
```diff
@@ -0,0 +1,302 @@
...
+      </LinkBox>
+      {props.type === "published" ? (
+        <PublishedLinkContent
+          status={props.status}
+          subdomain={props.subdomain}
+          link={props.link}
+          isPublished={props.isPublished}
+        />
+      ) : (
+        <Link pl={"31px"} href={props.link}>
```
jessicamcinchak, line 188: nit: noticing lots of places throughout these file changes where `31px` is hard-coded - it'd be much more ideal to try to use an existing global `theme.spacing` value here I think ?

jessicamcinchak (commented):
Took a quick first pass at this! Looking like a really solid start, but finding some of the styling and conditional syntax used throughout to be a bit confusing - let's talk through any  initial feedback points / questions if they aren't clear!  

Best tip I can offer is "do less" / try to reference more from our already existing MUI global theme ! Personally I think our MUI theme & existing style patterns should be your source of truth here, and Figma as a visual guide only (we almost never use Figmas to pixel exactness).

PR #3485, opened 2024-08-01
chore: Fix Slack notification API docs

DafyddLlyr:
## What does this PR do?
- Fixes two issues in `/send-slack-notification` endpoint Swagger docs, annotated below
- Please see https://github.com/theopensystemslab/planx-new/pull/3466#discussion_r1698777844 for context

### DIFF
--- a/api.planx.uk/modules/webhooks/docs.yaml
+++ b/api.planx.uk/modules/webhooks/docs.yaml
@@ -17,26 +17,23 @@ components:
     BopsSubmission:
       type: object
       properties:
-        body:
+        event:
           type: object
           properties:
-            event:
+            data:
               type: object
               properties:
-                data:
+                new:
                   type: object
                   properties:
-                    new:
-                      type: object
-                      properties:
-                        session_id:
-                          type: string
-                        bops_id:
-                          type: string
-                        destination_url:
-                          type: string
+                    session_id:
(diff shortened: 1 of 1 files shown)


api.planx.uk/modules/webhooks/docs.yaml
```diff
@@ -253,7 +244,7 @@ components:
                 type: string
                 description: A message returned to the client
 paths:
-  /webhooks/hasura/sendSlackNotification:
+  /webhooks/hasura/send-slack-notification:
```
DafyddLlyr, line 247: This now matches the route - https://github.com/theopensystemslab/planx-new/blob/782f982894bdfd1d5853ef5c5e22991214743320/api.planx.uk/modules/webhooks/routes.ts#L39-L43

api.planx.uk/modules/webhooks/docs.yaml
```diff
@@ -66,53 +63,47 @@ components:
     EmailSubmission:
       type: object
       properties:
-        body:
+        event:
```
DafyddLlyr, line 66: Previously, these were incorrectly wrapped in `body` which means the example payload in the API docs are incorrect.

image

jessicamcinchak (approved):

PR #3489, opened 2024-08-01
[editor] serve microsoft-identity file

freemvmt:
We want to demonstrate to Microsoft/Azure that we own `editor.planx.uk`, for the purpose of verifying it as a 'publisher domain', which will then be displayed to users attempting to sign on via a Microsoft account, for added trust.

NB. This can be reverted once the verification is complete.

### DIFF
--- a/editor.planx.uk/public/.well-known/microsoft-identity-association.json
+++ b/editor.planx.uk/public/.well-known/microsoft-identity-association.json
@@ -1,0 +1,7 @@
+{
+  "associatedApplications": [
+    {
+      "applicationId": "4469ee56-9275-47f3-b29d-e0df0991099a"
+    }
+  ]
+}


DafyddLlyr (changes requested):
I think that the following changes would be required for this to work - 
 - Remove the `<link>` from `index.html`
 - Nest `microsoft-identity-association.json` at `/public/.well-known`
 
Not 100% about the double slash notation MS are using - but we can merge and test on staging 👍

freemvmt:
@DafyddLlyr like so?

I meant to nest the file, but I thought the `<link ... />` tag was required to serve the file? Cool if not :)

freemvmt:
> I meant to nest the file, but I thought the `<link ... />` tag was required to serve the file? Cool if not :)

Ignore - I see for example that `robots.txt` doesn't need referencing anywhere!

DafyddLlyr (approved):

PR #3492, opened 2024-08-05
refactor: Team component folder structure

jamdelion:
I've split this into a separate PR from the ticket work I'm doing so that this can easily be discarded if it's rocking the boat too much / destroying a convention in the codebase!

But...I've basically massively shortened the length of `TeamMembers.tsx` by splitting out the two components that were defined in there into `MembersTable` and `TeamMembers`, and also moving type definitions and styling declarations into their own files. 

In my option, this makes the component files a lot cleaner, easier to read and easier to work with! I just hate scrolling past loads of styling to find out what a component does 😅 And I thought I'd kind of start with this before I dig into making logic changes/doing ticket work. 

What do you all think? Is this too much of a departure from the longer component file convention that you seem to have going on (at least in `FlowEditor` components)?


DafyddLlyr (approved):
Makes total sense to pick up this change before adding more code and complexity here.

Certainly not wedded to keeping styles and component in a single file - it does get harder to manage as more is added.

PR #3497, opened 2024-08-06
feat: Make external portals open in new tab when clicked

RODO94:
## What does this PR do? 

This PR adds functionality to open external portals in new tabs when they are clicked in the flow editor. 

The work follows this Trello ticket: https://trello.com/c/MJ5VSADN/2960-could-external-portals-open-in-a-new-browser-tab-rather-than-in-window

I did think if it were better to enable right clicking to select what you'd like depending on your brower (new tab, new window etc...) but decided I'd follow the ticket explicitly first and see if more is required.

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Portal.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Portal.tsx
@@ -95,7 +95,13 @@ const ExternalPortal: React.FC<any> = (props) => {
         className={classNames("card", "portal", { isDragging })}
         onContextMenu={handleContext}
       >
-        <Link href={`/${href}`} prefetch={false} ref={drag}>
+        <Link
+          href={`/${href}`}
+          target="_blank"
+          rel="noopener noreferrer"
+          prefetch={false}
+          ref={drag}
+        >
           <span>{href}</span>
         </Link>
         <Link href={editHref} prefetch={false} className="portalMenu">


DafyddLlyr (commented):
Also left this on Trello for full visibility - bit of a design issue here I've just realised.

The default behaviour when right-clicking (opening the context menu) is disabled when clicking on nodes as this is our hidden copy/paste shortcut. 

If we enable right-click on a node, we’d need an alternative copy/paste option handled somehow.

Currently I can click on a node and open in-place, or `⌘ + click` to open in a new tab - both options are available. If we just make nodes open in a new tab, the user loses the option to open in-place and is forced to open in new tab.

The PR does exactly as the ticket asks, but on reflection maybe this is not the right approach?

RODO94:
@DafyddLlyr Great feedback! We're definitely on the same wavelength with this, I was tossing and turning over whether it is a simpler flow to ``⌘ + click`` or click to open in a new tab and navigate back to the previous tab if you wanted to stay on the same tab. 

Are you suggesting we look rather into a feature to right click and open in a new tab?

DafyddLlyr:
> Are you suggesting we look rather into a feature to right click and open in a new tab?

I actually think the right thing to do here is nothing (for now!), until we design and implement an alternative approach to copy/paste which is currently hijacking the right-click interaction. We can discuss with the wider group on Thursday at planning 👍 

Any thoughts on this @ianjon3s?

ianjon3s:
@RODO94 and I briefly discussed this and are thoughts are — as external portals cannot be cloned, if there is a simple way to allow right-click on only external portals (removing the right-click hijack for only them), then we'll proceed with that.

If not, leave as-is for now in favour of moving towards a context menu (or alternative to right-click to clone) for all components.

DafyddLlyr:
Ah that's a very good thought 🙌 

There is already error handling for pasting external portals - so it makes sense to not allow them to be copied. If you just remove the `onContextMenu` prop from external portals this should solve it 👍

RODO94:
Converting to Draft as there's still a bit of investigation and code to be added prior to any further review

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Portal.tsx
```diff
@@ -40,11 +36,6 @@ const ExternalPortal: React.FC<any> = (props) => {
           "/",
         );
         setHref(href);
-        addExternalPortal({
-          id: props.data.flowId,
-          name: data.flows_by_pk.name,
-          href,
-        });
```
DafyddLlyr, line 47: This function is required for search and will need to remain. It shouldn't be affecting the right-click functionality. For reference this was added here (via `git blame`) - https://github.com/theopensystemslab/planx-new/pull/3435

DafyddLlyr (changes requested):
One small thing which has been missed!

jessicamcinchak:
@RODO94 happy to do a final review of this one once that final comment from Daf is accounted for so it can be merged before he's back ! Will check back later today :rocket:

jessicamcinchak (approved):
Thanks for final adjustments here, works exactly as expected for me :heavy_check_mark:

PR #3498, opened 2024-08-08
feat: send a Slack notification when flow status is updated

jessicamcinchak:
Take 2


jessicamcinchak:
Had spotted a legitimate syntax error in the Swagger docs and migration this time around which I thought would clear this up - but alas still same zod union error - closing for an alternate approach, stay tuned !

PR #3500, opened 2024-08-08
feat: Submit new team editor data - happy path

jamdelion:
⚠️ This PR is quite big enough (and it is feature flagged) so hopefully a good place to leave it - note some future plans listed below.

### What does it do?
This is the 'happy path' of the feature ['Add user in editor](https://trello.com/c/piMcGsJa/2973-add-user-in-editor)' - somebody fills in the form correctly and submits it; a user is created and added to the team.

I started from a modal that opened on button click but did nothing but close when submitted.

### ✅  **Done:**
- Implement Formik to handle form validation and submission
- Get basic 'createUser' and 'addUserToTeam' working in two separate mutations
- Optimistically update the table with the values from the form
- Test that a row is added to the table when the form is filled in correctly

### ✅  **Done after code reviews:**
- Use store for fetching teamId
- Refactor `teamMembers` to useStore instead of passing members as a prop
- Combine mutations into one 
- Refetch users query as well as optimistic update using the store

### ⛔  **NOT included in this PR**
There are improvements outstanding that I will tackle in a future PR:
- Re-look at ‘it closes the modal’ test
- Try to use at ‘user.type’ instead of ‘fireEvent.change’
- Error handling
- Show toast message on success
- Ensure only platformAdmins can see this feature
- Move mutation to planx-core

### Demo
add-team-editors-modal


editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
```diff
@@ -42,58 +45,70 @@ export const MembersTable = ({
...
-            </TableRow>
-          )}
-        </TableBody>
-      </Table>
-    </TableContainer>
+          </TableHead>
+          <TableBody
+            data-testid={`members-table${showAddMemberButton && "-add-editor"}`}
+          >
+            {tableData.map((member) => (
```
jamdelion, line 67: This looks like a lot of changes but it's literally just adding a testId to the members table if it's the one with the 'add editor' button 😅

editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
```diff
@@ -42,58 +45,70 @@ export const MembersTable = ({
...
+          </TableBody>
+        </Table>
+      </TableContainer>
+      {showModal && (
+        <AddNewEditorModal
+          showModal={showModal}
+          setShowModal={setShowModal}
+          setTableData={setTableData}
+          tableData={tableData}
+        />
```
jamdelion, line 110: and moving the modal to `MembersTable.tsx` from `TeamMembers.tsx`

editor.planx.uk/src/pages/FlowEditor/components/Team/components/lib/optimisticallyUpdateTable.tsx
```diff
@@ -0,0 +1,18 @@
...
+import React from "react";
+
+import { AddNewEditorFormValues, TeamMember } from "../../types";
+
+export const optimisticallyUpdateTable = (
+  values: AddNewEditorFormValues,
+  setTableData: React.Dispatch<React.SetStateAction<TeamMember[]>>,
+  tableData: TeamMember[],
+) => {
+  const randomNumber = getRandomNumberInRange(100, 200); // each table row needs a unique numerical key so randomise for now
```
jamdelion, line 11: Seems a bit hacky, and also not scalable if a council has more than 100 users in this team Editors table... Would welcome thoughts on improving!

editor.planx.uk/src/pages/FlowEditor/components/Team/tests/TeamMembers.addNewEditor.test.tsx
```diff
@@ -1,33 +1,85 @@
...
+  }),
+}));
+jest.mock("pages/FlowEditor/components/Team/queries/addUserToTeam.tsx", () => ({
+  addUserToTeam: jest.fn().mockResolvedValue({
+    team_id: 1,
+    role: "teamEditor",
+    user_id: 1,
+    __typename: "team_members",
+  }),
+}));
```
jamdelion, line 36: I know this is a lot of mocks! I'm looking forward to discussing testing in general, but this is more how I've been used to doing testing in the past. 

Ideally I'd like to find a way to extract these away so the test file is cleaner. ✨

editor.planx.uk/src/pages/FlowEditor/components/Team/tests/TeamMembers.addNewEditor.test.tsx
```diff
@@ -1,33 +1,85 @@
...
+    const createUserButton = await screen.findByTestId(
+      "modal-create-user-button",
+    );
+    user.click(createUserButton);
   });
-  it("opens the modal and displays the input fields", async () => {
-    expect(await screen.findByTestId("modal-create-user-button")).toBeVisible();
-    expect(await screen.findByLabelText("First name")).toBeVisible();
+
+  it("adds the new user row to the Team Editors table", async () => {
```
jamdelion, line 76: I wanted an `it` block for 'it closes the modal` but had to give up - it was repeatedly throwing false positives (passing even though I don't believe the modal was closed). Would welcome any thoughts on whether I should continue trying to get a test in for that.

editor.planx.uk/src/pages/FlowEditor/components/Team/tests/helpers/userEntersInput.tsx
```diff
@@ -0,0 +1,13 @@
...
+export const userEntersInput = async (
+  labelText: string,
+  inputString: string,
+  container: HTMLElement,
+) => {
+  const inputField = await within(container).findByLabelText(labelText);
+
+  fireEvent.change(inputField, {
+    target: { value: inputString },
+  });
```
jamdelion, line 12: Originally I had `user.type(inputField, inputString)` here (based on feedback from @DafyddLlyr in the last PR I opened) but I just couldn't get it working with this. 🤷 Probably something to dig into again soon - maybe in the upcoming testing discussion.

DafyddLlyr (commented):
A few thoughts and comments (mainly in response to your questions!)

Overall impressions are great! Code is well organised and structured, and the level of tests is great to see.

**Do I need to use a store?**
Yes - I think you shouldn't have to call `fetchCurrentTeam()` - the team we're currently in should be already be in the store. For example this would mean calling `useStore(state => state.teamId)` to get the ID.
 
**Is it possible to combine the mutations into one?**
Yep! Something like this would work, tested on GraphiQL playground on https://hasura.editor.planx.dev/ - 

```gql
mutation MyMutation {
  insert_users(
    objects: {
      email: "daf@daf.com"
      first_name: "daf"
      last_name: "daf"
      teams: { data: { role: teamEditor, team_id: 20 } }
    }
  ) {
    returning {
      id
    }
  }
}

```

Docs: https://hasura.io/docs/latest/mutations/postgres/insert/#one-to-one--one-to-many-relationships

**what role does the user need to be created with?**
There are currently two roles which a user can have (defined by the `user_roles` enum table - https://hasura.editor.planx.dev/console/data/default/schema/public/tables/user_roles/browse)

Right now, only users with `teamEditor` role are displayed (and actively used). This means we should always just be setting the role to `teamEditor` for this feature 👍 

In future we might expand the use of `teamViewer` roles, and allow these to be added, but this isn't a requirement for now.

**Does the user need to be a platform admin? (isPlatformAdmin)**
Added users should not be platformAdmins. This role essentially maps to OSL staff, and any new users assigned to a `teamEditor` role would not be platformAdmins. Just checked permissions and currently only platformAdmins can add users - we likely want to change this to allow `teamEditors` to add new users (to their team only).

I had assumed this was in place already sorry, hence why there was a comment about not updating permissions was on the ticket. We can catch up to talk this through 👍 

image

**Should I be hooking more into planx-core functions**
I think that might be a slightly cleaner approach, where we just remove the query from here and move to planx-core. This might also make mocking a little easier? If you search for `jest.mock("@opensystemslab/planx-core"` you'll get some examples of this - mostly from the API I think but it should be a useable approach.

I'll add a meeting next week to dive into `planx-core` with both you and @RODO94.

**Optimistic updates** 
Currently we're getting the list of team members in the route (`editor.planx.uk/src/routes/teamMembers.tsx`) and just passing as a prop into the `TeamMembers` table. A better approach here might be to - 
 - Keep the fetch request in the route
 - Store the list of team members in the store instead of passing as a prop
 - The "optimistic update" would then be to push the created team member to this array in the store instead of refetching all team members
 
As always, happy to talk all of this through if clearer or we can look to pair on a few of these options 👍

DafyddLlyr (changes requested):
Apologies for taking so long to get back to you here! A few things to look at - happy to talk it through if helpful 👍 

Also reading through these comments I noticed we never picked up and solved the `user` / `setup()` test helpers - @RODO94 also had a similar approach on https://github.com/theopensystemslab/planx-new/pull/3525

Today's pretty busy with calls so I've added a half hour slot Tuesday for us to dive into this one.

editor.planx.uk/src/pages/FlowEditor/components/Team/components/lib/filterTeamMembers.tsx
```diff
@@ -0,0 +1,13 @@
+import { TeamMember } from "../../types";
+
+export const hasEmailPresent = (member: TeamMember) => !!member.email;
+
+export const filterByEmailPresent = (members: TeamMember[]): TeamMember[] => {
```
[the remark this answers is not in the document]
jamdelion: You're totally right, that is so much better, thanks! 😁

editor.planx.uk/src/pages/FlowEditor/components/Team/queries/createAndAddUserToTeam.tsx
```diff
@@ -0,0 +1,40 @@
...
+      ) {
+        insert_users_one(
+          object: {
+            email: $email
+            first_name: $firstName
+            last_name: $lastName
+            teams: { data: { role: teamEditor, team_id: $teamId } }
+          }
+        ) {
+          id
```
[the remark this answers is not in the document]
jamdelion: Great suggestion, thanks! 🌟

editor.planx.uk/src/pages/FlowEditor/components/Team/tests/TeamMembers.addNewEditor.test.tsx
jamdelion: I managed to fix this, I think I was just missing the right combination of `queryBy` and `waitFor` but it turns out that react testing library has a [good article on testing for disappearance](https://testing-library.com/docs/guide-disappearance). This is in the follow-on PR: https://github.com/theopensystemslab/planx-new/pull/3543

editor.planx.uk/src/pages/FlowEditor/lib/store/team.ts
```diff
@@ -130,22 +135,20 @@ export const teamStore: StateCreator<
...
-    const isSuccess = await $client.team.updateTeamSettings(
-      teamId,
-      teamSettings,
-    );
-    return isSuccess;
+    return await $client.team.updateTeamSettings(teamId, teamSettings);
+  },
+
+  setTeamMembers: async (teamMembers: TeamMember[]) => {
+    set((state) => ({ teamMembers: [...state.teamMembers, ...teamMembers] }));
```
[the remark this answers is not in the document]
jamdelion: Nice catch! I fixed this then hit an interesting bug where it seemed like the database query to `GetUsersForTeam` wasn't being run when the user went back on the `/members` route. (e.g. Add member, see new member in the table, click on Design then click back to Team members and the new user had disappeared). 

🐰 I went down a whole rabbit hole about query caching and found this great article: https://www.apollographql.com/blog/demystifying-cache-normalization.

But the solution I've ended up with now involves refetching when the mutation completes, which maybe makes the store solution a bit redundant? I suppose it's still useful for what would otherwise be a loading state between mutating and querying, but I'd be interested to hear your thoughts!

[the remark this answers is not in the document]
DafyddLlyr: This is actually a great solution - we don't really leverage the caches in Apollo enough. All working as expected. Agreed - I think this does now make the store approach redundant, it might be worth tidying this up? 

I think if we were making this query just via planx-core, this approach wouldn't work as easily. Something I really want to take a proper look at (soon!) is the number of different ways the app is fetching and managing data - there's room for simplification and standardisation here.

DafyddLlyr (approved):
Looking great 👍

PR #3501, opened 2024-08-08
feat: setup MapAndLabel with React Context and schema input fields per map feature

jessicamcinchak:
**A few key changes:**
- Rebases to the latest `@opensystemslab/map` pre-release so we can `drawMany` 
- Adds the concept of a "Schema" to the Editor modal, with an initial option for "Trees" (very similar to `List`)
- Displays a form reflecting the schema fields for each feature plotted on the map; "resetting" the map also removes the forms
  - Each form is rendered in a MUI Vertical Tabs panel ([docs](https://mui.com/material-ui/react-tabs/#vertical-tabs)) - still a bit of polishing needed to match latest Figmas! 
- Refactors the main component to use React Context ([docs](https://react.dev/learn/passing-data-deeply-with-context))
  - This had big benefits for reducing validation complexity and prop drilling with the `List` component, and we'll have similar layers of complexity to manage here
  - To start, I copied _a lot_ directly from the List - this gets us to a play-with-able thing quick, but we'll definitely want to reduce duplication and extract some logic (especially around "Fields" I think) going forward - but this exercise should help us better understand which logic bits these components will share and what will need to be unique ! 
  
Screenshot from 2024-08-11 22-25-17

This component is feature-flagged and there's lots of known gaps/bugs/follow-ups to spin out of this, including and not limited to: 
- Form validation & error handling
- `handleSubmit` - add form inputs to GeoJSON properties?
- Polish up designs to match Figma; can we highlight map features based on active tab?? (set `color` property on `drawGeojsonData` ??)
- Debug shifty map labels on modify


editor.planx.uk/src/@planx/components/List/Public/index.tsx
```diff
@@ -25,7 +25,7 @@ import { ListProvider, useListContext } from "./Context";
 
 export type Props = PublicProps<List>;
 
-const ListCard = styled(Box)(({ theme }) => ({
+export const ListCard = styled(Box)(({ theme }) => ({
```
DafyddLlyr, line 28: nit: Maybe this should be in a shared location (shared/Schema/styles.tsx?) instead of importing directly from the List component. Without this, we could make a change here without really realising it has impacts elsewhere.

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+    const geojsonChangeHandler = ({ detail: geojson }: any) => {
+      if (geojson["EPSG:3857"]?.features) {
+        setFeatures(geojson["EPSG:3857"].features);
+      } else {
+        // if the user clicks 'reset' on the map, geojson will be empty object, so set features to undefined
+        setFeatures(undefined);
+      }
+    };
+
+    const map: any = document.getElementById("map-and-label-map");
```
DafyddLlyr, line 127:
```suggestion
    const map: HTMLDivElement | null = document.getElementById("map-and-label-map");
```

DafyddLlyr: And then maybe throw an exception if no map found in order to type narrow within this function?

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+        title={title}
+        description={description}
+        info={info}
+        policyRef={policyRef}
+        howMeasured={howMeasured}
+      />
+      <FullWidthWrapper>
+        <MapContainer environment="standalone">
+          {/* @ts-ignore */}
+          <my-map
```
DafyddLlyr, line 148: nit/follow-up: We need FindProperty again to display this in-Editor, we should have a GraphError component here 👍

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+  features,
+}) => {
+  const { schema, activeIndex, formik } = useMapAndLabelContext();
+  const [activeTab, setActiveTab] = useState<string>(
+    features[features.length - 1].properties?.label || "",
+  );
+
+  const handleChange = (_event: React.SyntheticEvent, newValue: string) => {
+    setActiveTab(newValue);
+  };
```
DafyddLlyr, line 39: Is `activeTab` actually going to be synonymous with `activeIndex` - do we need to manage both, or just find a way of mapping one onto the other?

DafyddLlyr (approved):
A few small comment / questions / fix-forwards. Solid groundwork here! 👍

editor.planx.uk/src/@planx/components/List/Public/index.tsx
jessicamcinchak: Thanks for catching this - no longer exported at all! This was a remnant from the first non-tab prototype where we just showed a `ListCard` for each feature on the map, but now we've got the tabs

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
jessicamcinchak: Yep now implemented !

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
jessicamcinchak: I think/hope it's just a matter of mapping one onto the other - but TBD a bit ! Trying to find right balance of not having to introduce too many fancy map highlight/zoom-to functionalities right off the bat

PR #3502, opened 2024-08-08
[chore] bump passport to 0.7.0

freemvmt:
🔧   [Ticket in Trello](https://trello.com/c/BKP8MVAK).

We've wanted to bump the [passport](https://www.passportjs.org/) version in use for auth routes for some time, because `0.6.0` fixes a 'moderate' [security vulnerability](https://nvd.nist.gov/vuln/detail/CVE-2022-25896) (as [flagged by dependabot](https://github.com/theopensystemslab/planx-new/security/dependabot/1) 2 years ago).

But this version throws up a bug related to our use of `cookie-session`, which `passport` does not commit to being compatible with (see [here](https://github.com/LinkedInLearning/node-authentication-2881188/issues/2)) (it only claims to support `express-session`). The issue is that passport used to implement the `req.session.regenerate` and `req.session.save` methods, which are called by `cookie-session`, but no longer does.

So in this PR we jump to `0.7.0` ([changelog](https://github.com/jaredhanson/passport/blob/master/CHANGELOG.md)), and implement a fix for this bug by stubbing out dummy methods on `req.session`. Credit to [this comment in passport#904](https://github.com/jaredhanson/passport/issues/904#issuecomment-1307558283) for the fix.

Supertest does not make the `request` objection available for testing - only the `response`. Therefore in order to test things about the request we are actually constructing, I added a new module with a dedicated test route. Very open to other ideas on how to test these dummy methods if that doesn't seem best practice :)

### DIFF
--- a/api.planx.uk/modules/test/routes.ts
+++ b/api.planx.uk/modules/test/routes.ts
@@ -1,0 +1,9 @@
+import { Router } from "express";
+import { testSessionMethods } from "./controller";
+
+const router = Router();
+
+// in order to test the session setup, we need a dedicated test route
+router.get("/test-session", testSessionMethods);
+
+export default router;
--- a/api.planx.uk/session.ts
+++ b/api.planx.uk/session.ts
@@ -1,0 +1,12 @@
+import { RequestHandler } from "http-proxy-middleware";
+
+// this middleware registers dummy methods on req.session to resolve passport/cookie-session incompatibility
+export const registerSessionStubs: RequestHandler = (req, _res, next): void => {
+  if (req.session && !req.session.regenerate) {
+    req.session.regenerate = (cb: (err?: Error) => void) => cb();
+  }
+  if (req.session && !req.session.save) {
+    req.session.save = (cb: (err?: Error) => void) => cb();
+  }
+  next();
+};
--- a/api.planx.uk/modules/test/controller.ts
+++ b/api.planx.uk/modules/test/controller.ts
@@ -1,0 +1,7 @@
+import { RequestHandler } from "express";
+
+export const testSessionMethods: RequestHandler = (req, res) => {
+  const hasRegenerate = typeof req.session?.regenerate === "function";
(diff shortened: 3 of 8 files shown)


api.planx.uk/modules/test/routes.ts
```diff
@@ -0,0 +1,9 @@
+import { Router } from "express";
+import { testSessionMethods } from "./controller";
+
+const router = Router();
+
+// in order to test the session setup, we need a dedicated test route
+router.get("/test-session", testSessionMethods);
+
+export default router;
```
jessicamcinchak, line 9: Seems like a solid light-touch solution to the `supertest` limitation - I like this approach !

api.planx.uk/session.ts
```diff
@@ -0,0 +1,12 @@
+import { RequestHandler } from "http-proxy-middleware";
+
+// this middleware registers dummy methods on req.session to resolve passport/cookie-session incompatibility
+export const registerSessionStubs: RequestHandler = (req, _res, next): void => {
+  if (req.session && !req.session.regenerate) {
+    req.session.regenerate = (cb: (err?: any) => void) => cb();
+  }
+  if (req.session && !req.session.save) {
+    req.session.save = (cb: (err?: any) => void) => cb();
```
jessicamcinchak, line 9: nit: I think these `any` might be able to use built-in `Error` interface? (Haven't tested locally, `any` isn't a blocker to merging if not!)

jessicamcinchak (approved):
Looks good! Very satisfying to address this longstanding dependabot alert :ok_hand: 

Can login & logout of pizza as expected :white_check_mark:

api.planx.uk/session.ts
freemvmt: aye this is a nice little improvement, will get it in!

jessicamcinchak:
[Upgrade the `passport` dependency](https://trello.com/c/BKP8MVAK/2138-upgrade-the-passport-dependency)

PR #3504, opened 2024-08-09
feat: try adding a MapField to the List component

jessicamcinchak:
The code needs to be refactored - it's still rewritten as a MapAndLabel which is a List component dupe! 

But you'll get the idea 

Screenshot from 2024-08-09 16-32-02


jessicamcinchak:
Closing in favor of #3505 - much tidier !

PR #3505, opened 2024-08-11
feat: add a "map" field to the List component

jessicamcinchak:
After catching up about "Map & Label" design with Ian & Rory last Friday - there's still basically one big outstanding question: 
- Is the user's entry point a map _and then_ they annotate each feature they've plotted (eg #3501 & vertical tab design in Figma) 
- OR is the most accessible entry point a list/table _and then_ they spatially locate what they've already described

This PR prototypes the latter option - which I actually don't think requires a "new" component as our original instinct at all, rather we can add a new "map" field to the existing List component. 

Screenshot from 2024-08-11 13-32-13

Still outstanding:
- Fix formik & validation schema to record geojson features
- Inactive card for "map" field needs a bit of special handling - we want to render a static reference map similar to existing `SummaryList`
- Decide how to structure data `onSubmit` if a "map" field is present - do we want to collate into a singular GeoJSON `FeatureCollection` where other "fields" are reflected as `properties` of each individual feature?

### DIFF
--- a/editor.planx.uk/src/@planx/components/List/Public/Fields.tsx
+++ b/editor.planx.uk/src/@planx/components/List/Public/Fields.tsx
@@ -5,9 +5,12 @@ import Grid from "@mui/material/Grid";
 import MenuItem from "@mui/material/MenuItem";
 import RadioGroup from "@mui/material/RadioGroup";
 import { visuallyHidden } from "@mui/utils";
+import { MapContainer } from "@planx/components/shared/Preview/MapContainer";
 import { getIn } from "formik";
+import { Feature } from "geojson";
 import { get } from "lodash";
-import React from "react";
+import { useStore } from "pages/FlowEditor/lib/store";
+import React, { useEffect, useState } from "react";
 import SelectInput from "ui/editor/SelectInput";
 import InputLabel from "ui/public/InputLabel";
 import ChecklistItem from "ui/shared/ChecklistItem";
@@ -19,6 +22,7 @@ import { DESCRIPTION_TEXT, ERROR_MESSAGE } from "../../shared/constants";
 import BasicRadio from "../../shared/Radio/BasicRadio";
 import type {
   ChecklistField,
+  MapField,
   NumberField,
   QuestionField,
   TextField,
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/@planx/components/List/Public/Fields.tsx
```diff
@@ -252,3 +256,76 @@ export const DateFieldInput: React.FC<Props<DateField>> = ({
...
+        id={id}
+      >
+        <MapContainer environment="standalone">
+          {/* @ts-ignore */}
+          <my-map
+            id={id}
+            ariaLabelOlFixedOverlay={`An interactive map for plotting and describing ${schema.type.toLocaleLowerCase()}`}
+            height={400}
+            drawMode
+            drawMany={drawMany}
```
jessicamcinchak, line 312: note that `drawMany` won't actually do anything here yet until next map release !

editor.planx.uk/src/@planx/components/List/model.ts
```diff
@@ -65,11 +66,28 @@ export type DateField = {
   data: DateInput & { fn: string };
 };
 
+export type MapField = {
+  type: "map";
+  data: {
+    title: string;
+    fn: "features";
```
jessicamcinchak, line 73: I _think_ in cases of the `"map"` field it will be helpful to have a static/predictable `fn` rather than user-configurabled free text `string` type if we want to do special  `handleSubmit` geojson formatting - any objections?

editor.planx.uk/src/@planx/components/List/model.ts
```diff
@@ -81,7 +99,7 @@ export interface Schema {
   max?: number;
 }
 
-export type UserResponse = Record<Field["data"]["fn"], string | string[]>;
+export type UserResponse = Record<Field["data"]["fn"], string | any[]>; // string | string[] | Feature[]
```
jessicamcinchak, line 102: `Feature[]` threw some type errors that'd be nice to come back, debug & tighten later ! Just defaulting to `any[]` for this first go

editor.planx.uk/src/@planx/components/List/Public/index.tsx
```diff
@@ -65,6 +66,8 @@ const InputField: React.FC<Field> = (props) => {
       return <ChecklistFieldInput id={inputFieldId} {...props} />;
     case "date":
       return <DateFieldInput id={inputFieldId} {...props} />;
+    case "map":
+      return <MapFieldInput id={inputFieldId} {...props} />;
```
DafyddLlyr, line 70: It would be nice to also add some tests in `/List/Public/index.test.tsx`, similar to this - https://github.com/theopensystemslab/planx-new/pull/3519/commits/960a28531d208c1c9d10db266ec7cb6002a1531e

I'm not _quite_ sure how map interactions are going to work in react-testing-library. If it's not simple I'd be happy to pass this by for now, but if it's a case off adding a `MapField` to the test Zoo schema and just clicking a point within a div then it's worth it I think!

DafyddLlyr (commented):
I'm getting an issue on the Pizza with the map instantiating. No errors in console associated with this - tried in a few different teams and in Editor and /preview

image

jessicamcinchak:
@DafyddLlyr re your rendering bug - in order to test a "map" field, your List component needs to come after a FindPropert (so map knows where to center!) - need to introduce "graph error" for this similar to [here](https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/src/%40planx/components/PlanningConstraints/Public.tsx#L64-L65), sorry invisible requirement still at this MVP stage !

editor.planx.uk/src/@planx/components/List/Public/index.tsx
jessicamcinchak: Added this as a follow-up to the Trello checklist for a couple reasons:
- Gotta re-fresh my memory how it works to mix React & native web components using react-testing-library
- We aren't handling key test-able states here yet like "Inactive" card which makes it hard to simply add to existing test schema
- Based on FindProperty order dependency, we'll also have to expand mock test data a bit with some fake breadcrumbs/passport as if FindProp has already been seen ! 

All totally do-able, but don't want to hold this up longer at this stage while still feature flagged if that's okay

editor.planx.uk/src/@planx/components/List/model.ts
jessicamcinchak: Switched to generic `"string;"` for now then! 

Traditionally, I think it's come in very handy to have static "fn" for geo components like FindProp, Draw, Constraints etc - but we can always update this later based on feedback / better understanding of real scenarios (eg if there ever needs to be > 1 map field).

editor.planx.uk/src/@planx/components/List/Public/index.tsx
DafyddLlyr: Yep, on board with this 👌

DafyddLlyr (approved):

PR #3507, opened 2024-08-12
fix: refactor uses of `__dirname` now that API is using `esnext`

jessicamcinchak:
Should fix failing e2e regression tests on `main` :crossed_fingers: https://opensystemslab.slack.com/archives/C04EEBR1E0P/p1723443286868999

Regression tests passing against this branch :white_check_mark: https://github.com/theopensystemslab/planx-new/actions/runs/10358321445

**Two part bug:** 
1.  `__dirname is not defined` error was being thrown when trying to generate zip folders for submissions (hence Uniform-related e2e failure) - we can now access this by using `fileURLToPath` as outlined in [this post](https://iamwebwiz.medium.com/how-to-fix-dirname-is-not-defined-in-es-module-scope-34d94a86694d)
2. BUT that solution throws the following error on `pnpm test` : `The 'import.meta' meta-property is only allowed when the '--module' option is 'es2020', 'es2022', 'esnext', 'system', 'node16', or 'nodenext'` - which is a bit strange because these options do indeed match our API's `tsconfig` 
    - So then I spent some time working through various suggestions outlined in [this thread](https://github.com/kulshekhar/ts-jest/issues/3888) & ended up using the final solution to add an extra dev dependency `ts-jest-mock-import-meta`
    
**Manual testing notes:**
- For a valid session on staging, we can't successfully generate a zip and it throws the `__dirname` error: https://api.editor.planx.dev/admin/session/be215bcf-1acf-4d76-893c-3c319a00966a/zip (connect to Warp to access, open this GET endpoint in your browser, expect it to trigger a download but instead throws error)
- For the same valid session copied to this pizza, we can successfully generate a zip: https://api.3507.planx.pizza/admin/session/be215bcf-1acf-4d76-893c-3c319a00966a/zip (if you get token error, login to this pizza editor first then retry so you have a valid JWT in your browser - should see download start!)

### DIFF
--- a/api.planx.uk/modules/send/utils/exportZip.ts
+++ b/api.planx.uk/modules/send/utils/exportZip.ts
@@ -1,21 +1,22 @@
-import os from "os";
-import path from "path";
-import { $api } from "../../../client/index.js";
-import { stringify } from "csv-stringify";
-import fs from "fs";
-import str from "string-to-stream";
-import AdmZip from "adm-zip";
-import { getFileFromS3 } from "../../file/service/getFile.js";
 import {
-  hasRequiredDataForTemplate,
-  generateMapHTML,
   generateApplicationHTML,
   generateDocxTemplateStream,
+  generateMapHTML,
+  hasRequiredDataForTemplate,
+  Passport,
 } from "@opensystemslab/planx-core";
-import { Passport } from "@opensystemslab/planx-core";
-import type { Passport as IPassport } from "../../../types.js";
-import type { Stream } from "node:stream";
 import type { PlanXExportData } from "@opensystemslab/planx-core/types";
+import AdmZip from "adm-zip";
+import { stringify } from "csv-stringify";
+import fs from "fs";
+import type { Stream } from "node:stream";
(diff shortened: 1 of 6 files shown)


api.planx.uk/modules/send/utils/exportZip.ts
```diff
@@ -1,21 +1,22 @@
...
+import { stringify } from "csv-stringify";
+import fs from "fs";
+import type { Stream } from "node:stream";
+import os from "os";
+import path from "path";
+import str from "string-to-stream";
+import { fileURLToPath } from "url";
+import { $api } from "../../../client/index.js";
+import type { Passport as IPassport } from "../../../types.js";
+import { getFileFromS3 } from "../../file/service/getFile.js";
```
jessicamcinchak, line 19: this is mostly noise from running "organise imports" in this file, only one new import here which is `fileURLToPath`

RODO94 (approved):
Ran through the manual testing listed and it works as expected. 

It all looks good to me

PR #3508, opened 2024-08-13
feat: add analytics tracking to "Did you use FOI service" question in LDCs

jessicamcinchak:
Context: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1723126009340879

**Testing:** 
- Answer the question in this sample flow & click "Continue" https://3508.planx.pizza/testing/used-foi-analytics-test/published
- Confirm your "allow listed" answer is recorded in the `analytics_summary` view https://hasura.3508.planx.pizza/console/data/default/schema/public/views/analytics_summary/browse (one row _per_ component/card, so answer is only expected to be present on corresponding "Question" type)

PR #3510, opened 2024-08-13
chore: add analytics tracking to capture `PlanningConstraints` & `PropertyInformation` "overrides"

jessicamcinchak:
Adds new passport variables that we'll track via our analytics `ALLOW_LIST`:
- `propertyInformation.action` with possible values "Accepted the property type" or "Changed the property type"
- `planningConstraints.action` with possible values "Accepted all planning constraints" or "Reported at least one inaccurate planning constraint"
- `_overrides` which is only defined _if_ the user has changed the property type _or_ reported inaccurate planning constraints and will include the specific details about _what_ was changed/overrode
  - This data format can also be what we reference in the submissions payload to share override info with officers & back-office systems

Todo: 
- [x] Update `analytics_summary` & `submissions_summary` views ! Waiting on #3508 to not have competing migrations open for the same thing

Testing: 
- Change your property type & override a constraint & confirm your passport is setting `_overrides` and the `.action` variables above as you'd expect !

### DIFF
--- a/api.planx.uk/modules/gis/service/classifiedRoads.ts
+++ b/api.planx.uk/modules/gis/service/classifiedRoads.ts
@@ -128,6 +128,7 @@ export const classifiedRoadsSearch = async (
             text: `is on a Classified Road`,
             data: features.map((feature: any) => ({
               name: `${feature.properties["RoadName1"]} - ${feature.properties["RoadClassification"]}`,
+              entity: feature.properties["GmlID"], // match Planning Data "entity" identifier for convenience when reporting inaccurate constraints
               properties: feature.properties,
             })),
             category: "General policy",
--- a/editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
@@ -50,6 +50,7 @@ function Component(props: Props) {
     siteBoundary,
     { x, y, longitude, latitude, usrn },
     hasPlanningData,
+    priorOverrides,
   ] = useStore((state) => [
     state.currentCard?.id,
     state.cachedBreadcrumbs,
@@ -57,6 +58,7 @@ function Component(props: Props) {
     state.computePassport().data?.["property.boundary.site"],
     (state.computePassport().data?.["_address"] as SiteAddress) || {},
     state.teamIntegrations?.hasPlanningData,
+    state.computePassport().data?.["_overrides"],
   ]);
 
   // PlanningConstraints must come after at least a FindProperty in the graph
@@ -66,7 +68,7 @@ function Component(props: Props) {
   //   still prepopulate any previously marked inaccurateConstraints
   const initialInaccurateConstraints =
     currentCardId &&
-    cachedBreadcrumbs?.[currentCardId]?.["data"]?.["_overrides"];
+    cachedBreadcrumbs?.[currentCardId]?.["data"]?.["_overrides"]?.[props.fn];
   const [inaccurateConstraints, setInaccurateConstraints] =
     useState<InaccurateConstraints>(initialInaccurateConstraints);
 
@@ -176,7 +178,13 @@ function Component(props: Props) {
               if (data) _constraints.push(data as GISResponse["constraints"]);
             }
 
-            const _overrides = inaccurateConstraints;
+            const hasInaccurateConstraints = inaccurateConstraints && Object.keys(inaccurateConstraints).length > 0;
+            const _overrides = hasInaccurateConstraints ? { ...priorOverrides, [props.fn]: inaccurateConstraints } : undefined;
+
+            // `planningConstraints.action` is for analytics
+            const userAction = hasInaccurateConstraints
+              ? "Reported at least one inaccurate planning constraint" 
+              : "Accepted all planning constraints";
 
             // `[props.fn]` & `_nots[props.fn]` are responsible for future service automations
             const _nots: IntersectingConstraints = {};
@@ -206,6 +214,7 @@ function Component(props: Props) {
             const passportData = {
               _constraints,
               _overrides,
+              "planningConstraints.action": userAction,
               _nots: notsAfterOverrides,
               ...intersectingConstraintsAfterOverrides,
             };
--- a/editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
@@ -216,7 +216,7 @@ function Component(props: Props) {
               _overrides,
               "planningConstraints.action": userAction,
               _nots: notsAfterOverrides,
-              ...intersectingConstraintsAfterOverrides,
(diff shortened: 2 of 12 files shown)


jessicamcinchak:
[Ability to override planning constraints [Solution to GIS layers pulling constraints (and planning histories) that apply to adjoining properties]](https://trello.com/c/S6zRlO42/2637-ability-to-override-planning-constraints-solution-to-gis-layers-pulling-constraints-and-planning-histories-that-apply-to-adjoini)

api.planx.uk/modules/gis/service/classifiedRoads.ts
```diff
@@ -128,6 +128,7 @@ export const classifiedRoadsSearch = async (
             text: `is on a Classified Road`,
             data: features.map((feature: any) => ({
               name: `${feature.properties["RoadName1"]} - ${feature.properties["RoadClassification"]}`,
+              entity: feature.properties["GmlID"], // match Planning Data "entity" identifier for convenience when reporting inaccurate constraints
```
jessicamcinchak, line 131: This is a small change to address an "undefined" entity bug noted in the PlanningConstraints `utils.test.ts` file - this ensures `_overrides` will always have investigate-able values!

editor.planx.uk/src/@planx/components/PropertyInformation/Public.tsx
```diff
@@ -45,7 +45,17 @@ function Component(props: PublicProps<PropertyInformation>) {
       titleBoundary={passport.data?.["property.boundary.title"]}
       blpuCodes={blpuCodes}
       overrideAnswer={overrideAnswer}
-      handleSubmit={props.handleSubmit}
+      handleSubmit={() => {
+        // If the user changed their property type, they'll already have a previous PropertyInformation breadcrumb that set `_overrides`
+        const hasOverrodeAnswer = passport.data?.["_overrides"]?.["property.type"];
+        const passportData = {
+          "propertyInformation.action": hasOverrodeAnswer ? "Changed the property type" : "Accepted the property type",
+        };
```
jessicamcinchak, line 53: There's a bit of a tricky edge case possible here - "Accepted the property type" isn't _really_ semantically correct if `showPropertyTypeOverride` wasn't enabled to begin with - but I think a binary makes most sense for analytics clarity here rather than introducing a third possible track like "Proceeded through Property Information" ? 

Any objections?

editor.planx.uk/src/pages/FlowEditor/lib/analytics/provider.tsx
```diff
@@ -40,8 +40,11 @@ export const ALLOW_LIST = [
   "application.declaration.connection",
   "drawBoundary.action",
   "findProperty.action",
+  "_overrides",
+  "planningConstraints.action",
   "property.constraints.planning",
   "property.type",
+  "propertyInformation.action",
```
jessicamcinchak, line 47: With this change we now have basic analytics coverage to understand how users proceed through our ALL our geospatial components (based on component type, not restricted to whether used in a submission or guidance flow!) :heavy_check_mark:

editor.planx.uk/src/@planx/components/PropertyInformation/Public.tsx
DafyddLlyr: Thanks for flagging this - agree with your decision here. 

I think a binary choice here is clearer and easier to understand and manage than something that's maybe semantically perfect 👍

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
```diff
@@ -176,7 +178,13 @@ function Component(props: Props) {
               if (data) _constraints.push(data as GISResponse["constraints"]);
             }
 
-            const _overrides = inaccurateConstraints;
+            const hasInaccurateConstraints = inaccurateConstraints && Object.keys(inaccurateConstraints).length > 0;
+            const _overrides = hasInaccurateConstraints ? { ...priorOverrides, [props.fn]: inaccurateConstraints } : undefined;
```
DafyddLlyr, line 182: An observation / pattern I didn't really notice last time round when `inaccurateConstraints` were added to this component.

This component has transitioned from being informational only, to now capturing data much like other components. This data is generally captured and validated using formik (unlike the `useState()` hook here for `inaccurateConstraints`).

This particular function is essentially handling the "back" behaviour usually managed by `getPreviouslySubmittedData()`.

If / when we come back to handle using formik for the `OverrideEntitiesModal` instances, it likely makes sense that this state would live at this root component level to match the behaviour of other components as opposed to individual instances within each modal as I think I'd previously suggested.

DafyddLlyr (approved):
All looking good!

Tested manually and hit an issue I previously missed - if you remove all constraints you passports ends up with `[]` as the value for `property.constraints.planning`. As this is a truthy value it could cause some issues - more accurately this should be `undefined` (which I think will then get it stripped out by `computePassport()` and then match a journey of a user who had no constraints).

image

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
jessicamcinchak: Thanks for noting this here so we have record when coming back; will leave unresolved so stays expanded :+1:

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
```diff
@@ -216,7 +216,7 @@ function Component(props: Props) {
               _overrides,
               "planningConstraints.action": userAction,
               _nots: notsAfterOverrides,
-              ...intersectingConstraintsAfterOverrides,
+              ...(intersectingConstraintsAfterOverrides[props.fn]?.length === 0 ? undefined : intersectingConstraintsAfterOverrides),
```
DafyddLlyr, line 219: Nice!

DafyddLlyr (approved):

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
jessicamcinchak: Good catch with this one !

jessicamcinchak: Sorry @DafyddLlyr - actually second guessing this now - can you please test on this flow https://3510.planx.pizza/opensystemslab/digital-land-api-search-test/preview

When undefined instead of empty array, then any future questions asking about `property.constraints.planning` are NOT going to be automated because the passport doesn't have the key at all / thinks it's seeing it for the first time!

jessicamcinchak: Okay full circle :cyclone: I think this is actually consistent with current production behavior for sites with no constraints & good to go. Agree that at minimum, `undefined` is the most easy-to-reason-about behavior and semantically correct for this journey! 

Going to merge this since `handleOverrides` has quite solid test coverage, but then carve out some more time in a new PR to really go through edge cases and write more tests :+1:

PR #3518, opened 2024-08-15
fix: update ResidentialUnits list schemas based on feedback

augustlindemer:
Add 'type' to title of all ResidentialUnits list schemas to better signify that the description can apply to multiple units.

Reword 'identical units' of all ResidentialUnits list schemas question to 'how many fit the description above' wording.

Remove development type question from ResidentialUnits GLA New and Rebuilt list schema.


jessicamcinchak (approved):
Looks good !! 

Reminder there's still a slightly annoying quirk that after this is merged, you'll need to fully remove / re-add any existing List components that use these schemas - won't pick up changes "automatically" yet

augustlindemer:
> Looks good !!
> 
> Reminder there's still a slightly annoying quirk that after this is merged, you'll need to fully remove / re-add any existing List components that use these schemas - won't pick up changes "automatically" yet

Thanks! I think it's enough to edit & re-select the schema within the component, so we can keep title and help texts etc.. Also I realised I missed something in the Proposed schema, so running checks again.

jessicamcinchak (approved):

PR #3519, opened 2024-08-15
feat(works-to-trees): Add `DateInputField` to `List` component

DafyddLlyr:
## What does this PR do?
Allows dates to be added to list component schemas

## Why
Required for works to trees feature, useful for other list schemas.

image

### DIFF
--- a/editor.planx.uk/src/@planx/components/List/Public/Fields.tsx
+++ b/editor.planx.uk/src/@planx/components/List/Public/Fields.tsx
@@ -5,12 +5,14 @@ import Grid from "@mui/material/Grid";
 import MenuItem from "@mui/material/MenuItem";
 import RadioGroup from "@mui/material/RadioGroup";
 import { visuallyHidden } from "@mui/utils";
+import { paddedDate } from "@planx/components/DateInput/model";
 import { getIn } from "formik";
 import { get } from "lodash";
 import React from "react";
 import SelectInput from "ui/editor/SelectInput";
 import InputLabel from "ui/public/InputLabel";
 import ChecklistItem from "ui/shared/ChecklistItem";
+import DateInput from "ui/shared/DateInput";
 import ErrorWrapper from "ui/shared/ErrorWrapper";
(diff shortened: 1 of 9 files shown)


DafyddLlyr:

editor.planx.uk/src/@planx/components/List/Public/Fields.tsx
```diff
@@ -225,3 +228,26 @@ export const ChecklistFieldInput: React.FC<Props<ChecklistField>> = (props) => {
     </InputLabel>
   );
 };
+
+export const DateFieldInput: React.FC<Props<DateField>> = ({
+  id,
+  data,
+}) => {
+  const { formik, activeIndex } = useListContext();
```
jessicamcinchak, line 236: As we think next about how to better share "schemas" between components that may use them (`List` here, `MapAndLabel` next) - I'm very interested if this `useListContext()` could be extracted out into a `useSchemaContext` or similar ?? Then the "Schema" becomes a Context & formik aware component that can be referenced in many components?

Was anticipating "fields" would be something I could externally reference/"just pull into" initial `MapAndLabel` setup - but this bit of context very much collided / lead me to initially just duplicating field definitions with a minor `useMapAndLabelContext()` difference. 

Definitely not a thing to address in this PR, just starting to think bigger direction!

jessicamcinchak (approved):
Thanks for picking this one up, works perfect !

editor.planx.uk/src/@planx/components/List/Public/Fields.tsx
DafyddLlyr: Totally - this is the exact issue I hit when I was trying to rush out `Page` before show and tell 😅 

There's really two layers of responsibility here - something which handles fields, and something which handles lists (or the wider component). So errors relating to min, max, etc would be in the lists "layer" and the field validation logic etc can we handled elsewhere.

Planning on taking a look at this next pretty much as I know it's a bit of a blocker atm 👍

PR #3482, opened 2024-07-31
chore: add `IDOX_NEXUS_CLIENT` for test environment to Planx staging

jessicamcinchak:
API tests should be good to go here once rebased to #3507 :white_check_mark: 

**How to test:**
- Go to any of these three services which are configured to "Send" to Idox Nexus & published:
  - LDC: https://3482.planx.pizza/lambeth/apply-for-a-lawful-development-certificate/published
  - PP: https://3482.planx.pizza/buckinghamshire/apply-for-planning-permission/published
  - PA: https://3482.planx.pizza/buckinghamshire/apply-for-prior-approval/published
- Fill out the form as an applicant and do a test submisison
  - Choose "householder" related project paths (mainly relevant to PP & PA)
  - Upload test files
  - Make mock payments if prompted ([Gov UK mock details here](https://docs.payments.service.gov.uk/testing_govuk_pay/#mock-card-numbers))
- Once you've hit the Confirmation page, there should now be a "one-off event" queued up in the Hasura console that will submit to `/idox/{localAuthority}`
  - Check the "Submissions log" of the service you used directly in the editor to see if the event succeeded or errored
    - On success, confirm a couple details in the response body:
      - `zipAttached` is `true`
      - `application.submission_reference` matches the PlanX session ID
      - `application.entity` is the prefix of the ODP Schema application type data field (eg `"ldc"`, `"pp"` or `"pa"`)
      - `application.module` is `"dcplanx"` 
      - `application.description` is `"Staging submission from PlanX"`
      - `application.organisationId` will randomly be one of four values `210051`, `210052`, `210053` or `210054` (NOT representative of "real" council staging environments yet!)
      - `application.processor_type` is `"PLANX_QUEUE"` which triggers an internal Idox task to download the user-uploaded `"files"` from the JSON payload (we don't have visibility into this task yet though!)
  - There should also be a new record in the `uniform_applications` audit table (check via Hasura console)
  - If you want to inspect exactly what the submitted zip contained, you can use this admin endpoint `api.3482.planx.pizza/admin/session/{your-session-id}/zip?onlyDigitalPlanningJSON=true`
- :lady_beetle: Raise errors here or on Slack & let's debug together ! 

---
- ** After you've done at least one full frontend journey through a service and confirmed the payload is valid, it's possible to _resubmit_ that session ID via the `/create-send-events/:sessionId` webhook endpoint (nicer than directly using `/idox/:localAuthority` because it'll create a Hasura event & inspectable Submissions log entry as above)
[...]

PR #3524, opened 2024-08-16
Production deploy

jessicamcinchak:
Waiting for #3523 (then will merge #3482 post prod deploy)

Are regression tests okay on `main` to proceed here ?


DafyddLlyr (approved):
Not yet resolved E2E tests, but as they're passing locally I'd be happy to move forward here 👍

PR #3525, opened 2024-08-16
feat: Shareable Public Link in Service Settings

RODO94:
## What does this PR do?

Follows work completed here: https://github.com/theopensystemslab/planx-new/pull/3483

And addresses the Trello ticket: https://trello.com/c/k8cQojk7/2958-share-service-link-in-editor

I have added a new section to the **Status** card of the **Service Settings** page. The new section displays a _Public Link_ with the intent to expose a Team's `subdomain` or `published` link (in both cases the query param `?analytics=false` as been excluded from the link).

To achieve this, I added three new components `<PublicLink>` ,`<TitledLink>` and `<CopyButton>` to simplify the code and contain certain parts of the logic within contextual components. 

Main piece of logic in these components is the handling of what link to show when, and when to show it active or inactive.

```ts
const PublicLink: React.FC<{
  isFlowPublished: boolean;
  status: FlowStatus;
  subdomain: string;
  publishedLink: string;
}> = ({ isFlowPublished, status, subdomain, publishedLink }) => {

  const isFlowPublic = isFlowPublished && status === "online"; 
  // this variable helps contain boolean logic for active / inactive
  
  const hasSubdomain = Boolean(subdomain); 
  // this variable makes subdomain strictly true / false instead of truthy / falsey

  switch (true) {
    case isFlowPublic && hasSubdomain:
      return <TitledLink isActive={true} link={subdomain} />;
    case isFlowPublic && !hasSubdomain:
      return <TitledLink isActive={true} link={publishedLink} />;
    case !isFlowPublic && hasSubdomain:
      return <TitledLink isActive={false} link={subdomain} />;
    case !isFlowPublic && !hasSubdomain:
      return <TitledLink isActive={false} link={publishedLink} />;
  }
  // decided to contain the subdomain and publishedLink within this component instead of...
  // adding it to the main component and just passing a "link" prop down to this component...
  // since it contains the logic in one place instead of two
};
```

---

I also created to new variables in the main `<ServiceSettings>` component - `publishedLink` and `subdomainLink` to control the creation of each link. 

_The `subdomainLink` is built differently to the `publishedLink` as the `publishedLink` was taken from the main index which controls the FlowEditor, I'd be interested to know the differences in how they are built, and if I should be harmonising these. The way I built `subdomainLink` felt more explicit for me with exposed ``flowSlug`` in the url. I've highlighted them in the snippet below_

```ts
   const publishedLink = `${window.location.origin}${rootFlowPath( false)}/published`; 
  // above is taken from flowEditor/index.tsx line 43


  const subdomainLink = teamDomain
    ? `${removeLinkEndSlash(teamDomain)}/${flowSlug}` // removelinkEndSlash is a fn I used to remove any trailing slashes
    : teamDomain;
    
    // the use of the ternary ensures that if no teamDomain exists, subdomainLink will be falsey
```
---

For displaying help text to the user, I have created a function managing a `switch` statement, adjusting help based on if the flow is Online and/or Published. Simplest approach seemed to let the function control the string that feeds in to the standard `<p>` being used in the section of the card rather than the function return a react component for each case.


editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings.tsx
```diff
@@ -24,6 +28,78 @@ import InputRowItem from "ui/shared/InputRowItem";
...
+        <CopyButton isActive={isActive} link={link} />
+      </Typography>
+      {isActive ? (
+        <Link variant="body2" href={link}>
+          {link}
+        </Link>
+      ) : (
+        <Typography style={{ color: "GrayText" }} variant="body2">
+          {link}
+        </Typography>
```
jessicamcinchak, line 76: These links aren't taking me where I'd expect them to! 

If I set a fake test subdomain, it's trying to _internally_ route me, rather than treating it as an external route - eg 
Screenshot from 2024-08-16 14-01-41
is pointing to `https://3525.planx.pizza/newcastle/apply-for-planning-permission/planningservices.test.gov.uk/apply-for-planning-permission`

Also two minor things:
- I'd expect this to open in a new tab by default?
- Can we use default `Link` styles here (eg underline) rather than `variant="body2"` ?

jessicamcinchak (commented):
Really appreciate this refactor - it's so much easier to read, the styling and code patterns feel much more consistent, and it achieves the original user story really cleanly ! 

I did experience a couple unexpected navigation things when actually clicking the links though! Plus a few other comments throughout

editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings.tsx
jessicamcinchak: Let's also chat about how to add test coverage to catch this if helpful ! Still think this is an important feature that would benefit from a few tests :+1:

RODO94: Ah! yeah - if you don't have a protocol prefix `https://...` it'll just add it on to the current URL.  Would a better way to route it be including ``href={ "//" + link }`` or should it be something similar to the trailing end "/" where it's managed when adding a new `subdomain`?

Added ``target={ _blank ) rel...`` to the Link now. Good catch!

For the last point, you mean for an inactive link, we should add the underline? Or switch the component to a ``<Link>``?

editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings.tsx
```diff
@@ -279,7 +390,16 @@ const ServiceSettings: React.FC = () => {
               enable analytics gathering.
             </p>
             <p>Offline services can still be edited and published as normal.</p>
+            <p>{publicLinkHelpText()}</p>
```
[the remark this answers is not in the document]
RODO94: Yeah that could be worthwhile, I was thinking this. Kept it within the current text block cause it seemed to be on a similar vein to the others, but probably easier to be close to the thing it's describing.

editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings.tsx
```diff
@@ -24,6 +28,78 @@ import InputRowItem from "ui/shared/InputRowItem";
 import type { FlowSettings } from "../../../../types";
 import { useStore } from "../../lib/store";
 
+const CopyButton = (props: { link: string; isActive: boolean }) => {
+  const [copyMessage, setCopyMessage] = useState<"copy" | "copied">("copy");
+  return (
+    <Tooltip title={copyMessage}>
+      <Button
+        disabled={!props.isActive}
+        component={"button"}
```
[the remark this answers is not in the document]
RODO94: Sorry for making you repeat the feedback from the previous PR!

editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings.tsx
jessicamcinchak: Personally I'd add the `https://` direct via a string template - tacking onto the comment above, could be: 
```const subdomainLink = teamDomain && `https://${teamDomain}/${flowSlug}`;``` ? 

A disabled `<Link />` would make semantic sense to me (haven't double checked MUI API docs if this is actually possible) or just underlined to match default PlanX link styles!

Comment (RODO94) [reply editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings.tsx:76]: Sick, so we assume teamDomain never has "https://" and I'll always add it.

Yeah I couldn't see a ``disabled`` prop in the `<Link>` [API](https://mui.com/material-ui/api/link/), so I'll add the underline

**Comment by jessicamcinchak** [comment]
When I see a "container unhealthy" or "in use elsewhere" error on Vultr - I typically destroy then re-run failed action only ! Can stem from many commits close together I think and overlapping "update" jobs colliding.
Screenshot from 2024-08-16 16-06-27

**Comment by jessicamcinchak** [comment]
I'd also personally still like to see a bit of test coverage here ahead of a final review please !! With your latest refactor, it should be much more straightforward to write tests that confirm links are formatted as expected, correct text is displaying etc.

**Comment by RODO94** [comment]
[Share service link in Editor](https://trello.com/c/k8cQojk7/2958-share-service-link-in-editor)

File: editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings.tsx [hunk 4]
```diff
@@ -280,6 +421,14 @@ const ServiceSettings: React.FC = () => {
             </p>
             <p>Offline services can still be edited and published as normal.</p>
           </SettingsDescription>
+
+          <PublicLink
+            isFlowPublished={isFlowPublished}
+            status={statusForm.values.status}
```
DafyddLlyr, line 427: This might be the required fix here?

```suggestion
            status={flowStatus}
```

RODO94: @DafyddLlyr I may have missed something, what is it fixing? I think this is the change we chatted about on Slack?

RODO94:
Converted to Draft as I work on test coverage

editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings.tsx
DafyddLlyr: Yep - this is it - I linked to this comment in the Slack chat, just wanted a note here also linked to the specific line.

image

Apologies for any confusion, please mark as resolved if sorted! 👍

jessicamcinchak (approved):
We talked through this one this afternoon - all final comments / suggestions / questions resolved there - looks good to merge to me !!

Thanks again for wading through all the initial design uncertainty with this task and seeing through this new solution - comprehensive test coverage and all :muscle:

PR #3526, opened 2024-08-16
refactor: Create `useSchema` hook to make validation logic and `InputField` components shareable

DafyddLlyr:
## Context
Whilst building the `List` component, the use of the `Schema` type and associated logic was quite closely linked to the implementation and behaviour of the list component itself. As more use cases have come forward for the concept of a `Schema`, it's time to pick this apart a little and split out logic which related to inputs, fields, validation and that which applies only to the `List` component such as error handling, multiple responses and payload generation.

I appreciate there's a fair amount of noise / churn in this PR sorry. I did spend a while trying to make into a few smaller PRs but it feels more cohesive and understandable as a single refactor in my mind. The commits are fairly clean and should allow you to navigate the steps I've taken here.

## What does this PR do?
 - Create a `useSchema` hook which generates a `FormikConfig<SchemaData>` - this can then be used within components which which to embed a schema. Just returning a partial `FormikConfig<T>` achieves a few things -
   - Allows custom `onSubmit` logic (e.g. flattening a list, merging GeoJSON Features to a a FeatureCollection)
   - `SchemaData` can be expanded if the component requires it
   - Clearly outlines the relationships and responsibilities - the form values are controlled by the parent component, `useSchema` is a helper method
 - Remove dependency on `ListContext` from `InputFields` and individual `${Type}Field` components. Instead we just require a `FromikProps<SchemaData>` to be passed in as a prop
 - Moves the above files from `/components/List` to `/components/shared/Schema`
 - Renames types - see [e93c3e1](https://github.com/theopensystemslab/planx-new/pull/3526/commits/e93c3e19863e741d957d43d6f3643926a5849fa3)

**Responsibilities**
| Schema | ListProvider |
|--------|--------|
| Generate validation schema and formik config | Generate and update formik instance |
| Generate initial values | Provide initial values (e.g. previous values) |
| Map schema to create inputs | Display inputs |
| Manage schema specific errors in inputs | Manage component specific errors |
| | Handle payload generation and submission | 

## Next steps...
 - Split up tests in `/List/Public/index.test.tsx` - many of these no longer require the `List` component to be tested
 - Improved types for `SchemaData` - worth a discussion in another PR - see comments below
 - Implement and use these changes in the `Page` and / or `MapAndLabel` component


editor.planx.uk/src/@planx/components/shared/Schema/Fields.tsx
```diff
@@ -17,45 +26,49 @@ import ErrorWrapper from "ui/shared/ErrorWrapper";
...
+type Props<T extends Field> = {
+  formik: FormikProps<SchemaData>
+  activeIndex: number;
+} & T;
 
-type Props<T> = T & { id: string };
+/**
+ * Helper function to get shared props derived from `Field` and `props.formik`
+ */
+const getFieldProps = <T extends Field>(props: Props<T>) => ({
```
DafyddLlyr, line 40: This function allows the components below to have a cleaner prop syntax and ensures we handle all of these in a standardised way.

editor.planx.uk/src/@planx/components/shared/Schema/Fields.tsx
```diff
@@ -178,11 +187,10 @@ export const SelectFieldInput: React.FC<Props<QuestionField>> = (props) => {
...
 export const ChecklistFieldInput: React.FC<Props<ChecklistField>> = (props) => {
-  const { formik, activeIndex } = useListContext();
-  const {
-    id,
-    data: { options, title, fn },
-  } = props;
+  const { data: { title, options }, formik } = props;
+  const { id, errorMessage, name, value } = getFieldProps(props);
+
+  if (!Array.isArray(value)) throw Error("'value' prop for ChecklistFieldInput must be of type string[]");
```
DafyddLlyr, line 193: Error handling is required here for type narrowing as `SchemaResponse` has no relationship to a `Field. It's not currently able to know that `value` will be a `string[]` here as it's derived from a `ChecklistFiled`.

We also see this in lines of code like this that require a manual type narrowing - 

https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/src/@planx/components/List/utils.tsx#L61-L63

This would be more meaningful / maintainable if the logic was something like - 

```
if (!field.type === "Checklist") {
  sum += parseInt(item?.identicalUnits);
}
```

As the range of fields grows (and the options in `SchemaResponse) this feels like it will be an import consideration.

There's a few solutions for this, I'll open a follow up PR shortly.

editor.planx.uk/src/@planx/components/shared/Schema/SchemaFields.tsx
```diff
@@ -0,0 +1,57 @@
...
+      if (props.data.options.length === 2) {
+        return <RadioFieldInput {...props} />;
+      }
+      return <SelectFieldInput {...props} />;
+    case "checklist":
+      return <ChecklistFieldInput {...props} />;
+    case "date":
+      return <DateFieldInput {...props} />;
+    default:
+      return null;
```
DafyddLlyr, line 32: To do (follow up PR): Add an exhaustive check on a few of these switch statements, or take a look at https://typescript-eslint.io/rules/switch-exhaustiveness-check/ as an option.

editor.planx.uk/src/@planx/components/shared/Schema/model.ts
```diff
@@ -0,0 +1,146 @@
...
+  min: number;
+  max?: number;
+}
+
+export type SchemaResponse = Record<Field["data"]["fn"], string | string[]>;
+
+/**
+ * Output data from a form using the useSchema hook
+ */
+export type SchemaData = { schemaData: SchemaResponse[] };
```
DafyddLlyr, line 87: Why change these type names? `UserData` is used elsewhere in a few components and it means "the data generated by the user in this component". By renaming this we indicate that it's intrinsically linked to the data generated by just the schema (and may not represent the entire user output of a component). Thoughts welcome - let me know if this isn't a constructive change.

jamdelion: Seems fine to me, I like that it flags that the component is a lot more generic.

editor.planx.uk/src/@planx/components/shared/Schema/Fields.tsx
```diff
@@ -17,45 +26,49 @@ import ErrorWrapper from "ui/shared/ErrorWrapper";
...
-  TextField,
-} from "../model";
-import { useListContext } from "./Context";
+import { DESCRIPTION_TEXT, ERROR_MESSAGE } from "../constants";
+import BasicRadio from "../Radio/BasicRadio";
+
+type Props<T extends Field> = {
+  formik: FormikProps<SchemaData>
+  activeIndex: number;
+} & T;
```
jamdelion, line 35: Feel free to ignore as this is just my preference on structure, but I would like to see types and helpers in a separate file to leave component files cleaner :)

editor.planx.uk/src/@planx/components/shared/Schema/Fields.tsx
```diff
@@ -67,34 +80,30 @@ export const TextFieldInput: React.FC<Props<TextField>> = ({ id, data }) => {
   );
 };
 
-export const NumberFieldInput: React.FC<Props<NumberField>> = ({
-  id,
-  data,
-}) => {
-  const { formik, activeIndex } = useListContext();
+export const NumberFieldInput: React.FC<Props<NumberField>> = (props) => {
```
jamdelion, line 83: also feel free to ignore: It would probably also be my preference to have all the different inputField components in separate files - I have a hatred of long files (I try to keep to max 150 lines). 🤷

jamdelion (approved):
The changes seem sensible enough to me!

I know you said that it made more sense to keep as one bigger PR but I think I probably would have preferred incremental smaller ones. No biggie though, it was easy enough to follow your train of thought by stepping through the commits 🚂 👍

editor.planx.uk/src/@planx/components/shared/Schema/hook.tsx
```diff
@@ -0,0 +1,49 @@
...
+  schema,
+  previousValues,
+}) => {
+  const validationSchema = generateValidationSchema(schema);
+  const initialValues = generateInitialValues(schema);
+
+  const getInitialValues = () => {
+    if (previousValues) return previousValues;
+
+    return schema.min ? [initialValues] : [];
```
jessicamcinchak, line 36: quick question: can't quite work out/remember logically why this return depends on `schema.min` ? If `initialValues` are defined, don't we want to return them regardless of the `min` set by the schema?

editor.planx.uk/src/@planx/components/shared/Schema/model.ts
jessicamcinchak: Agree `UserData` needed updating here, but are there any hesitations of `SchemaData` being a little too interchangable with payload jargon (eg how we commonly talk about _application output_)? 

Is `SchemaUserData` a happy medium, eg `UserData` suffix keeps mental link to _component output_ while still indicating it's about "Fields"? (Very nit picky, feel free to dismiss!)

editor.planx.uk/src/@planx/components/shared/Schema/Fields.tsx
jessicamcinchak: Helpful!

jessicamcinchak (approved):
Thanks for this ! Very excited to rebase `MapAndLabel` on top of sharable `Schema` :raised_hands:

editor.planx.uk/src/@planx/components/shared/Schema/Fields.tsx
DafyddLlyr: This is a really good shout - and I so nearly did it before opening the PR!

Implemented in [d0bdced](https://github.com/theopensystemslab/planx-new/pull/3526/commits/d0bdced2dbf1bc0a269024ae97a767db7527743f)

editor.planx.uk/src/@planx/components/shared/Schema/model.ts
DafyddLlyr: This works well for me! Fixed in [896e395](https://github.com/theopensystemslab/planx-new/pull/3526/commits/896e3954212d80ecd3dfc445ab9f8531e505e10f)

editor.planx.uk/src/@planx/components/shared/Schema/hook.tsx
DafyddLlyr: Yeah this is a good question! This was a UX decision we made when working on the List component.

If `!schema.min` we don't "pre-load" an empty item into the component as it's valid to continue without adding an item. If we didn't do this check, a user would need to cancel the item and then hit continue to get past the component. Likewise, if `schema.min` is truthy, we don't want the user to have to hit "Add new item" when this is something they would always have to do.

jessicamcinchak: Thanks for reminder :bulb:

PR #3527, opened 2024-08-16
fix: ensure Idox Nexus audit entries don't rely on XML generation and format submission log event name

jessicamcinchak:
Two quick bugs:
- Fixes XML generation bug related to testing Idox Nexus - see thread https://opensystemslab.slack.com/archives/C01E3AC0C03/p1723729231085529
- Ensures event name is nicely formatted in the editor Submissions Log (currently overflowing)

### DIFF
--- a/api.planx.uk/modules/send/idox/nexus.ts
+++ b/api.planx.uk/modules/send/idox/nexus.ts
@@ -373,8 +373,6 @@ const createUniformApplicationAuditRecord = async ({
   localAuthority: string;
   submissionDetails: UniformSubmissionResponse;
 }): Promise<UniformApplication> => {
-  const xml = await $api.export.oneAppPayload(payload?.sessionId);
-
   const application: Record<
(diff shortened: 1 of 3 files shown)


api.planx.uk/modules/send/idox/nexus.ts
```diff
@@ -413,7 +411,7 @@ const createUniformApplicationAuditRecord = async ({
       destination: localAuthority,
       response: submissionDetails,
       payload,
-      xml,
+      xml: "ODP Schema",
```
jessicamcinchak, line 414: In the future we could think about a new column to snapshot the schema payload or an entirely separate audit table - but for this stage of testing we can simply make a static note here and then do any schema troubleshooting using our admin endpoitns (eg `admin/session/:sessionId/digital-planning-application` where PlanX session ID = Idox submission reference)

DafyddLlyr (approved):

PR #3528, opened 2024-08-16
[api] implement verbatim module syntax in tsconfig

freemvmt:
This PR ~is currently~ was serving a purpose (as a rebase platform) in troubleshooting errors with Jest in #3453 (I explain my findings below), but it may also be worth merging at some point on its own merit.

The majority of the actual `import ...` fixes were done automatically using the [`consistent-type-imports` eslint rule](https://typescript-eslint.io/rules/consistent-type-imports/). I note that the typescript-eslint folks [don't recommend](https://typescript-eslint.io/rules/consistent-type-imports/#comparison-with-importsnotusedasvalues--verbatimmodulesyntax) using `verbatimModuleSyntax` and this rule together, but I'm not sure I'm convinced. I think any conflict here is outweighed by the value of being able to autofix lints? But open to being wrong!

If we'd rather bin the new eslint rule, it has largely served its purpose in automating the initial round of fixes, and TypeScript should flag any new failure to adhere to `verbatimModuleSyntax` in-IDE anyway, so it's not a significant loss.

## Exploration

See the _No transformations between module systems_ section [here](https://github.com/microsoft/TypeScript/issues/51479) for an important corollary of imposing [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax) (which was introduced in TS v5 and implies `isolatedModules`, `importsNotUsedAsValues` and `preserveValueImports`, the latter two of which are thereafter deprecated):

> ESM syntax cannot be used in files that will emit CommonJS syntax.

More than any other switch I can throw, this change seems to cause TypeScript to aggressively enforce that *every module* is ESM-compliant, or else it will throw an error.

This is exactly why I have to add the `ts-node` options to `tsconfig.json`, and adapt the `tsconfig.test.json` which the `jest.config.ts` reads. As per [this comment](https://github.com/kulshekhar/ts-jest/issues/4081#issuecomment-1503684089), both tools will error out if they abide by the `"verbatimModuleSyntax": true` option in the root tsconfig.

That is, if we comment out `"verbatimModuleSyntax": false` in the `ts-node` specific compiler options in `tsconfig.json`, we get the following instead:

```
Error: Jest: Failed to parse the TypeScript config file /home/dan/code/planx-new/api.planx.uk/jest.config.ts
  TSError: ⨯ Unable to compile TypeScript:
jest.config.ts:49:1 - error TS1286: ESM syntax is not allowed in a CommonJS module when 'verbatimModuleSyntax' is enabled.

49 export default config;
```

The thing to understand here is that Jest uses `ts-node` to transpile the `jest.config.ts` file into JS in order to parse it, and it [instructs ts-node to consider all files as commonJS](https://github.com/jestjs/jest/blob/0d96de975783ee78192c1037862ea3089f52dc25/packages/jest-config/src/readConfigFileAndSetRootDir.ts#L146), in a way which is difficult to overwrite from tsconfig.

This is not too much of a problem - `ts-node` is only used for this specific purpose, and we can probably stand one file being interpreted as cjs for the sole purpose of initiating tests. After the config is read in, Jest takes over, and the first thing is does it try to transform all files into JS using `ts-jest` as the transpiler.

If however, we comment out `tsconfig: ...` in `jest.config.ts`, we get something like the following:

```
 FAIL  ./server.test.js
  ● Test suite failed to run

    server.ts:3:8 - error TS1286: ESM syntax is not allowed in a CommonJS module when 'verbatimModuleSyntax' is enabled.
```

Now _this_ is a helpful error, and illustrates the real problem at hand. It tells us that `ts-jest` is genuinely not interpreting `server.ts` as an ESM module, despite the `"type": "module"` in `package.json`, the `useEsm` and `extensionsToTreatAsEsm` options in the jest config, etc!

### Conclusion

This demonstrates that perhaps the [Jest errors](https://github.com/theopensystemslab/planx-new/actions/runs/10409160785/job/28828151209?pr=3453) are more genuine/instructive than I imagined. Despite all the config in the world, jest/ts-jest refuse to recognise our `api.planx.uk` codebase as being composed of ESM modules. Perhaps the solution is simply to not enforce ESM for tests? 🤔

### DIFF
--- a/api.planx.uk/.eslintrc
+++ b/api.planx.uk/.eslintrc
@@ -18,6 +18,7 @@
       }
     ],
     "@typescript-eslint/no-non-null-assertion": "off",
+    "@typescript-eslint/consistent-type-imports": "error",
     "no-nested-ternary": "error",
     "@vitest/expect-expect": [
       "error",
--- a/api.planx.uk/docs/index.ts
+++ b/api.planx.uk/docs/index.ts
@@ -1,4 +1,4 @@
-import { Express } from "express";
+import type { Express } from "express";
 import swaggerJSDoc from "swagger-jsdoc";
 import swaggerUi from "swagger-ui-express";
 
--- a/api.planx.uk/helpers.ts
+++ b/api.planx.uk/helpers.ts
@@ -1,7 +1,8 @@
 import { gql } from "graphql-request";
 import capitalize from "lodash/capitalize.js";
-import { Flow, Node } from "./types.js";
-import { ComponentType, FlowGraph } from "@opensystemslab/planx-core/types";
+import type { Flow, Node } from "./types.js";
+import type { FlowGraph } from "@opensystemslab/planx-core/types";
+import { ComponentType } from "@opensystemslab/planx-core/types";
 import { $public, getClient } from "./client/index.js";
 
 export interface FlowData {
--- a/api.planx.uk/lib/hasura/metadata/index.test.ts
+++ b/api.planx.uk/lib/hasura/metadata/index.test.ts
@@ -1,10 +1,11 @@
-import { createScheduledEvent, RequiredScheduledEventArgs } from "./index.js";
+import type { RequiredScheduledEventArgs } from "./index.js";
+import { createScheduledEvent } from "./index.js";
 import axios from "axios";
 import type { Mocked } from "vitest";
 
 describe("Creation of scheduled event", () => {
   vi.mock("axios", async (importOriginal) => {
-    const actualAxios = await importOriginal<typeof import("axios")>();
+    const actualAxios = await importOriginal<typeof axios>();
     return {
       default: {
         ...actualAxios,
--- a/api.planx.uk/lib/hasura/metadata/index.ts
+++ b/api.planx.uk/lib/hasura/metadata/index.ts
@@ -1,4 +1,5 @@
-import Axios, { AxiosResponse, isAxiosError } from "axios";
+import type { AxiosResponse } from "axios";
+import axios, { isAxiosError } from "axios";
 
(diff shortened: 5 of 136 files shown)


DafyddLlyr (commented):
Just catching up here @freemvmt, thanks so much for the detailed write up - much appreciated.

We're currently facing a few Jest issues in the frontend also and are considering the switch to [Vitest](https://vitest.dev/). I don't want to muddy the waters too much or add another big config heavy task to your plate, but it might be worth a timeboxed investigation to see if a switch could solve a few of your issues without too much work?

freemvmt:
As per #3453, we resolved the issue with Jest which this PR was intended to help investigate by migrating our entire test suite for `api.planx.uk` to Vitest, which had immediate results!

Implementing the `verbatimModuleSyntax` module may still be a desirable change, to make our TypeScript code as explicit as possible and safeguard against any issues with per-file transpilation, but is no longer a dependency for the Microsoft single sign on work, so this is a decision we can take in the fullness of time.

DafyddLlyr (changes requested):
I know that this was originally introduced to resolve jest pain points, but now we've resolved that with vitest we need to decide what to do here as it's no longer a strict pre-requisite.

The authority I trust on all things TypeScript [recommends using `verbatimModuleSyntax`](https://www.totaltypescript.com/tsconfig-cheat-sheet) so I'm all in favour of this change - it's more explicit and clear so let's go ahead and make this change. Especially as you've done all the groundwork here! 

Once the PR is rebased I'll be happy to approve and merge this one.

DafyddLlyr (approved):

PR #3529, opened 2024-08-17
chore: renew Southwark SSL certs

jessicamcinchak:
:crossed_fingers: these look all good in AWS cert manager 

Note that Southwark sends us a "Root.crt" & "ServerCertificate.crt" in addition to the "Intermediate.crt" (chain), and the "ServerCertificate" is what's used here ("Root" still throws "does not match key" error!).

Also learned through this that Southwark's internal email rules were blocking any emails with the zip of certs attached, therefore had some confusion around them thinking they'd already sent things to eachother/us when nothing had actually come through - in the end Sam was able to send directly in Slack fine.


jessicamcinchak:
[Renew Southwark SSL certificate](https://trello.com/c/ieMjU3Pb/2983-renew-southwark-ssl-certificate)

DafyddLlyr (approved):

PR #3535, opened 2024-08-19
feat: send a Slack notification when flow status is updated on production

jessicamcinchak:
Replaces #3466 

Now a much simpler approach than trying to hook into existing `/webhooks/hasura/send-slack-notification` & submission events !

**Changes:**
- Adds a new API endpoint which can be triggered by `teamEditors` (and `platformAdmins` implicitly) which accepts a simple `message` body and posts to #planx-notifications if we're on production env
- Calls this endpoint from flow status `onSubmit` if "save" has been successful

**To test:**
- Navigate to flow settings -> flow status in the editor on this pizza & open your "Network" tab
- Toggle the flow status and confirm a request has kicked off with response `200` - message should indicate it's skipping the actual Slack post because not production, but still display the message that _would have_ posted - confirm the team, flow slug and status are correct
  - If you're in a team that we want to skip (eg "testing"), confirm that no notification request kicks off at all but you can still update your flow status as normal

**Open questions** noted in original PR still apply here, will raise to Silvia before merging as she's main audience of this! :heavy_check_mark: Answers in thread here https://opensystemslab.slack.com/archives/C5Q59R3HB/p1724140076950159


jessicamcinchak:
[Can we set up notifications for admins, to know when a new service goes 'online'?](https://trello.com/c/pznE5Sqk/2940-can-we-set-up-notifications-for-admins-to-know-when-a-new-service-goes-online)

editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings.tsx
```diff
@@ -102,6 +113,36 @@ const ServiceSettings: React.FC = () => {
...
+  const sendFlowStatusSlackNotification = async (status: FlowStatus) => {
+    const skipTeamSlugs = [
+      "open-digital-planning",
+      "opensystemslab",
+      "planx",
+      "templates",
+      "testing",
+      "wikihouse",
+    ];
+    if (skipTeamSlugs.includes(teamSlug)) return;
```
jessicamcinchak, line 125: The API handles "environment" logic because that should consistently apply to any scenario we're using this endpoint for, but the "skipable teams" logic is specific to flow status notifications, so direclty handled here. 

Open to alternate suggestions !

api.planx.uk/modules/slack/controller.ts
```diff
@@ -0,0 +1,46 @@
...
+    message: z.string(),
+  }),
+});
+
+export type SendSlackNotificationController = ValidatedRequestHandler<
+  typeof slackNotificationSchema,
+  SendSlackNotificationResponse
+>;
+
+export const sendSlackNotifcationController: SendSlackNotificationController =
```
jamdelion, line 21:
```suggestion
export const sendSlackNotificationController: SendSlackNotificationController =
```

An easy word to mispell! 😓

api.planx.uk/modules/slack/routes.ts
```diff
@@ -0,0 +1,18 @@
+import { Router } from "express";
+import { validate } from "../../shared/middleware/validate.js";
+import { useTeamEditorAuth } from "../auth/middleware.js";
+import {
+  sendSlackNotifcationController,
```
jamdelion, line 5: Just flagging that my typo suggestion fix won't update imports I think!

jamdelion (approved):

api.planx.uk/modules/slack/routes.ts
jessicamcinchak: Thank you for spotting ! Properly updated throughout now :heavy_check_mark:

PR #3514, opened 2024-08-15
chore: Add "no-nested-ternary" ESLint rule to projects

DafyddLlyr:
## What?
- Adds "no-nested-ternary" rules to ESLint configs ([docs](https://eslint.org/docs/latest/rules/no-nested-ternary))
- Refactors nested ternaries to satisfy rule

## Why?
- I saw [this comment](https://github.com/theopensystemslab/planx-new/pull/3483/files#r1715103291) on #3483 and agreed - as we grow as a team we likely want to move more towards codifying these conventions we aim to follow in the codebase
- Hopefully some of the "before" snippets refactored in this PR demonstrate that nested ternaries are actually very hard to follow - understanding some of these took a surprising amount of mental capacity for this time of day 😅

PR #3536, opened 2024-08-20
Revert "chore: Add "no-nested-ternary" ESLint rule to projects"

jessicamcinchak:
Reverts theopensystemslab/planx-new#3514

Also running regression tests against this branch here: https://github.com/theopensystemslab/planx-new/actions/runs/10470664105

We've had issues with SWR requests validating correctly since this change merged, reverting this to see if it's connected (seems doubtful, but we're really scratching our heads with this bug!) - full context in this thread: https://opensystemslab.slack.com/archives/C04EEBR1E0P/p1724054282014549

### DIFF
--- a/api.planx.uk/.eslintrc
+++ b/api.planx.uk/.eslintrc
@@ -29,8 +29,7 @@
           "supertest.**.expect"
         ]
       }
-    ],
-    "no-nested-ternary": "error"
+    ]
   },
   "globals": {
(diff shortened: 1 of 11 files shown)


jessicamcinchak:
Closing this one as confirmed not source of SWR bug (regression tests pass at same spot as elsewhere, requests fail to validate on pizza) !

PR #3537, opened 2024-08-20
fix: ensure loading indicator is visible when fetching Planning Constraints

jessicamcinchak:
Quick followup from #3514 - which refactored this bit of logic

While the condition looked fine on paper, I was getting some goofy UI behavior where instead of the loading indicator, we'd show only gis responses OR only road responses OR the no-info error summary, but then the full collated correct responses would fully load

This clears things up that bug & hopefully makes a marginal readability improvement


DafyddLlyr (approved):
Thanks for picking this up - both fixed and more readable ✅ ✅

PR #3538, opened 2024-08-20
fix: when proposing a new address, map should display at full width and lower zoom

jessicamcinchak:
Fixes issues raised by Tewkesbury & West Berks linked here: https://trello.com/c/TAY09Mpw/2980-diagnose-and-fix-map-center-zoom-issue-when-displaying-team-boundaries

**Quick but key changes here:**
- FindProperty map gets a `FullWidthWrapper` similar to our other interactive maps (eg DrawBoundary) for more real-estate
- Lowered `zoom` from `14` to `10` which seems to better accomodate more spacious rural councils while still working for London boroughs like Lambeth too
- Removed `drawGeojsonDataBuffer` prop - we don't need this for initial render, and when coming "back" after having plotted a new address, the map defaults will still be smartly zoomed into & centered on your existing point

No changes necessary to how we store boundaries or calculate bounding-boxes - that's all correct, was purely a viewport / display issue here I believe! 

The viewport is still "clipped" to the council boundary, and now that we're starting at a lower zoom, don't expect to be able to zoom out much from the starting location, but it should feel more "centered" on the full council now (rather than the center of the boundary like before).

**Before & afters:** 
- Tewkesbury
Screenshot from 2024-08-20 15-22-02
Screenshot from 2024-08-20 15-21-28

- West Berks
Screenshot from 2024-08-20 15-24-30
Screenshot from 2024-08-20 15-26-20

- Lambeth (did not report issue, but is quite a different scope boundary than above examples, so want to confirm it still displays reasonably too!)
Screenshot from 2024-08-20 15-40-09
Screenshot from 2024-08-20 15-27-21

### DIFF
--- a/editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
+++ b/editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
@@ -13,6 +13,7 @@ import {
 } from "@planx/components/shared/Preview/MapContainer";
 import { useStore } from "pages/FlowEditor/lib/store";
 import React, { useEffect, useRef, useState } from "react";
+import FullWidthWrapper from "ui/public/FullWidthWrapper";
 import InputLabel from "ui/public/InputLabel";
 import ErrorWrapper from "ui/shared/ErrorWrapper";
 import Input from "ui/shared/Input";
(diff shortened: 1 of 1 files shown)


jessicamcinchak:
[Diagnose and fix map center/zoom issue when displaying team boundaries](https://trello.com/c/TAY09Mpw/2980-diagnose-and-fix-map-center-zoom-issue-when-displaying-team-boundaries)

DafyddLlyr (commented):
> No changes necessary to how we store boundaries or calculate bounding-boxes - that's all correct, was purely a viewport / display issue here I believe!

Thanks for the clear explanation and investigation here, good to see that this is just a simple viewport issue 👍 

Code reads as correct and all looks good - but having some issues with the Pizza. Maybe unrelated to changes here, can you verify?

When I load a map I get the error "Error: Unimplemented type: 3" which is caused when a 404 is returned by the vector tile provider.

image

Also getting an issue when loading up West Berks' LDC service - https://3538.planx.pizza/west-berkshire/apply-for-a-lawful-development-certificate/published?analytics=false

image

DafyddLlyr (approved):
Approved - issues not consistently re-creatable, and have appeared independently of this PR in the past. Will re-test once merged to `main` 👍

PR #3539, opened 2024-08-21
chore: Use CI build for E2E frontend

DafyddLlyr:


DafyddLlyr:
Getting the following (legitimate!) error in CI here, showing this isn't really the issue or a likely direction to a solution. 

```sh
Treating warnings as errors because process.env.CI = true.
Most CI servers set it automatically.

Failed to compile.
```

PR #3544, opened 2024-08-21
feat: Replace craco with vite

DafyddLlyr:
## What does this PR do?
- Replaces craco with vite
- Replaces jest with vitest

## Motivations
- create-react-app is pretty much deprecated, so it makes sense to look for alternatives
- For a while now, we've been getting inconsistent issues when the app was being built that we've not been able to narrow down successfully. This branch originally started off as a test to see if a build using vite would resolve this - it seems to be the case that it does!

✅  Regression tests passing against branch here - https://github.com/theopensystemslab/planx-new/actions/runs/10560558494


jessicamcinchak:
[Replace Jest with Vitest in Editor to fix CI build issues](https://trello.com/c/GlouIBWM/3010-replace-jest-with-vitest-in-editor-to-fix-ci-build-issues)

editor.planx.uk/src/components/Header.test.tsx
```diff
@@ -108,7 +109,7 @@ describe("Header Component - Editor Route", () => {
 
   it("displays breadcrumbs", () => {
     setup(<Header />);
-    expect(screen.getByText("Plan✕")).toBeInTheDocument();
+    // expect(screen.getByText("Plan✕")).to.be.
```
jessicamcinchak, line 112: Oops this looks like a mistake! Also should be able to comment-back-in a11y tests in this file - looks like I missed these ones!

editor.planx.uk/src/lib/graphql.ts
```diff
@@ -37,8 +37,10 @@ const customFetch = async (
   return fetchResult;
 };
 
+console.log(import.meta.env.VITE_APP_HASURA_URL);
```
jessicamcinchak, line 40: This looks like a local testing remnant, can we remove the console log before merging?

editor.planx.uk/src/testUtils.tsx
```diff
@@ -28,12 +19,21 @@ export const setup = (
...
  * Required as this frequently updates following the final "expect()" call of a test,
  * leading to multiple "an update was not wrapped in act(...)" warnings
  * Docs: https://testing-library.com/docs/example-react-transition-group/
  */
-export const mockFade = jest.mock("@mui/material/Fade", () =>
-  jest.fn(({ children }: FadeProps) => <div>{children}</div>),
-);
+// export const mockFade = vi.mock("@mui/material/Fade", () =>
+//   vi.fn(({ children }: FadeProps) => <div>{children}</div>),
+// );
```
jessicamcinchak, line 39: This looks like a mock we might want to / need to migrate still?

editor.planx.uk/tsconfig.json
```diff
@@ -16,8 +16,9 @@
     "skipLibCheck": true,
     "strict": true,
     "target": "es6",
-    "typeRoots": ["src/types"]
+    "typeRoots": ["src/types"],
+    "types": ["vite/client", "vite-plugin-svgr/client"]
   },
-  "include": ["**/*.ts", "**/*.tsx"],
+  "include": ["**/*.ts", "**/*.tsx", "./src/setupTests.ts"],
```
jessicamcinchak, line 22: nit: I explicitly added `"./src/setupTests.ts"` at some point here when troubleshooting, but I don't actually think we need it / it should already be picked up by `"**/*.ts"` ? Can you remove and double-check please?

jessicamcinchak (commented):
Took another pass through all 100+ file changes with fresh eyes & a couple quick spots to tidy up before merge ! Thanks again for queuing up this big migration, feels good :raised_hands:

editor.planx.uk/src/components/Header.test.tsx
DafyddLlyr: Thanks - not sure if this is a merge conflict issue or one that's been missed, but will resolve now!

editor.planx.uk/src/lib/graphql.ts
DafyddLlyr: Good catch thanks, removed 🧹 🗑️

editor.planx.uk/tsconfig.json
DafyddLlyr: Working locally without this, just pushed up the change 👍

editor.planx.uk/src/testUtils.tsx
DafyddLlyr: Just pushed up [f32e69c](https://github.com/theopensystemslab/planx-new/pull/3544/commits/f32e69cf4f8528057bb6c17fde92407e2e16a1af) which addresses this.

`mockFade` has had to move to the test setup file as vitest does not allow mocks to be exported.

jessicamcinchak (approved):

editor.planx.uk/src/pages/FlowEditor/components/Settings/tests/ServiceSettings/PublicLinks.test.tsx
```diff
@@ -121,7 +121,7 @@ describe("A team with a subdomain has an online, published service.", () => {
 
     user.click(copyButton);
 
-    expect(await screen.findByText("copied")).toBeVisible();
+    expect((await screen.findAllByText("copied"))[0]).toBeVisible();
```
DafyddLlyr, line 124: CI was throwing this error - 
image

This looks like it was catching the tooltip as well as the actual text.

image

PR #3545, opened 2024-08-21
chore: update Confirmation banner description to use rich text & reword details

jessicamcinchak:
**Changes:**
- Updates banner `description` to be a rich text input instead of plain text
- Refactors `details` to be handled directly in the `Confirmation` component rather than at the `Node` level
  - There was an original request to make the "details" property names customisable while keeping the existing values, but after chatting with August we decided against this approach
  - We don't let editors change the text of other static pages (eg Application Saved) or Gov Notify email templates, and we ultimately want to ensure all these fields have consistent naming between these various places - so customisation in this one place doesn't feel like the right solution and would likely lead to Confirmation page language that is out of sync/disjointed from other parts of the service
  - We did however update the existing language to be slightly more generic and accurate - eg `Planning application reference` &rarr; `Application reference` (now matches emails), `Submitted` &rarr; `Paid at` (matches actual data :see_no_evil:)
  
Screenshot from 2024-08-21 19-56-09

**Follow-ups:**
- I'll create a _new_ Trello ticket to pick back up the larger discussion of if we want to drop "application" language throughout, everything we need to touch to do that fully, and how we might achieve it (eg do we just agree on another word or read dynamically from schema application types where possible etc) - see thread here: https://opensystemslab.slack.com/archives/C07F7215W7P/p1724253793464279

### DIFF
--- a/editor.planx.uk/src/@planx/components/Confirmation/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/Confirmation/Public.tsx
@@ -5,7 +5,8 @@ import { QuestionAndResponses } from "@opensystemslab/planx-core/types";
 import Card from "@planx/components/shared/Preview/Card";
 import { SummaryListTable } from "@planx/components/shared/Preview/SummaryList";
 import { PublicProps } from "@planx/components/ui";
-import { useStore } from "pages/FlowEditor/lib/store";
+import { objectWithoutNullishValues } from "lib/objectHelpers";
+import { Store, useStore } from "pages/FlowEditor/lib/store";
 import React, { useEffect, useState } from "react";
 import Banner from "ui/public/Banner";
 import FileDownload from "ui/public/FileDownload";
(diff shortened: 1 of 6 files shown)


jessicamcinchak:
[Allow more customisation within the "Confirmation" component](https://trello.com/c/SkRYq2bn/2981-allow-more-customisation-within-the-confirmation-component)

editor.planx.uk/src/pages/Preview/Node.tsx
```diff
@@ -114,43 +112,10 @@ const Node: React.FC<any> = (props: Props) => {
...
-      };
 
-      return (
-        <Confirmation
-          {...allProps}
-          details={objectWithoutNullishValues(details)}
-          color={{ text: "#000", background: "rgba(1, 99, 96, 0.1)" }}
-        />
-      );
-    }
```
jessicamcinchak, line 153: I personally think this was quite a strange pattern & unnecessarily extracted/pre-processed component content - and much prefer having a presentational component `Details` now that reads from the store.

The one trade-off is a less meanginful Confirmation Storybook because `details` can't be mocked as a prop any longer - but I think we can solve this with smarter state-mocking in future within the story itself?

Do say so if any objections to this approach or if you think the Storybook trade-off is too big !

editor.planx.uk/src/@planx/components/Confirmation/Public.tsx
```diff
@@ -85,18 +74,68 @@ export default function ConfirmationComponent(props: Props) {
...
+        <React.Fragment key={`detail-${i}`}>
+          <Box component="dt">{k}</Box>
+          <Box component="dd">{v}</Box>
+        </React.Fragment>
+      ))}
+    </SummaryListTable>
+  );
+}
+
+// TODO - Retire in favor of ODP Schema application type descriptions or fallback to flowName
```
jessicamcinchak, line 133: Don't want to do this yet because we know our ODP Schema enums are a bit in flux, but imagine it will be the better / more consistent way forward ! This function only does special handling for LDCs - which is fine for now as they're the only live statutory service.

editor.planx.uk/src/@planx/components/Confirmation/Public.tsx
```diff
@@ -40,6 +68,23 @@ export default function ConfirmationComponent(props: Props) {
...
+  );
+}
+
+interface PresentationalProps extends Props {
+  sessionId: string;
+  applicableDetails: Record<string, string>;
+  data: QuestionAndResponses[];
+}
+
+export function Presentational(props: PresentationalProps) {
```
jessicamcinchak, line 87: The addition of a `Presentational` component here is purely a convenience so we can continue mocking `applicableDetails` for Storybook (eg similar to Review page which is another component that relies heavily on store data). Could just as easily return this via `Confirmation` directly, but then Storybook would only ever display a single row `"Application reference"` for details section as that's gauranteed to always be in state.

DafyddLlyr (approved):
One very small comment, but all looking good. The move to just more generic language is a good one I think, thanks for cross-checking with @augustlindemer on this one 👌 

Always nice to see some quality tidy ups as well!

PR #3547, opened 2024-08-21
chore: Downgrade esbuild

DafyddLlyr:
Don't think this is the solution we want, but interested to see the impact of rolling this back to the previous version we were using.


DafyddLlyr:
Regression tests failing despite this, closing so as to not introduce any new variables here.

PR #3548, opened 2024-08-21
chore: Use esbuild 0.20

DafyddLlyr:


DafyddLlyr:
Regression tests failing despite this, closing so as to not introduce any new variables here.

PR #3550, opened 2024-08-23
fix: Correct logic for `slotsSchema` validation

DafyddLlyr:
## What does this PR do?
Addresses issues outlined in this ticket - https://trello.com/c/zYOIK4d1/2996-allow-users-to-bypass-upload-and-label-exemptions-not-working-correctly

> Component requires either all files to be optional or files uploaded to be > 0. Recommended files are however not treated as optional, effectively being treated as if required. The allFilesOptional exemption should really be a noFilesRequired exemption, as the only time a users should be blocked from proceeding without uploading anything is when one or more files are required. Related issue: Even when all files are set to optional, with no uploaded files the component still always triggers the “Remove files which failed to upload” error.

This is now fixed (with added tests). The "Remove files which failed to upload” error was over-eager and giving false positives, the allFilesOptional test has been updated to make the syntax a little clearer, but it's still logically the same.

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/schema.test.ts
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/schema.test.ts
@@ -185,6 +185,19 @@ describe("slotSchema", () => {
     expect(result).toBe(false);
   });
 
+  it("rejects slots which failed to upload", async () => {
+    const mockSlots = [
+      { status: "error" },
+      { status: "success" },
+    ] as FileUploadSlot[];
+
+    const result = await slotsSchema.isValid(mockSlots, {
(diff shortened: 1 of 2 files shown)


DafyddLlyr:
[Allow users to bypass Upload and Label (exemptions not working correctly)](https://trello.com/c/zYOIK4d1/2996-allow-users-to-bypass-upload-and-label-exemptions-not-working-correctly)

jessicamcinchak (approved):
Perfect, thanks!

jessicamcinchak:
I'll go ahead and merge this one & flag to Jonny at standup this morning that it's ready to test on staging :+1:

augustlindemer:
Working perfectly.

PR #3552, opened 2024-08-23
fix: Update store on flow status change

DafyddLlyr:
Small UX improvement I noticed when turning a flow online/offline for testing.

By directly updating the store, Zustand will reactively update our UI so we don't have to reload the component (e.g. via navigation) to see the update.

https://github.com/user-attachments/assets/bf976ac5-50ae-4995-ad24-9f02c0eabba9


RODO94 (approved):
Looks great, good improvement.

PR #3554, opened 2024-08-23
refactor: Move `submission_email` from `teams` to `team_settings`

RODO94:
## What does this PR do?

This PR sets up work contained in this Trello Ticket: https://trello.com/c/dVsTQpZZ/2984-allow-editors-to-see-update-their-teams-submission-email-in-team-settings-form

With `submission_email` moving to the **General Settings** tab of a Team's settings, it makes sense to move `submission_email` to the `team_settings` table for future reference and consistency. 

For this to happen, I had to add a new column to `team_settings` + update permissions to include the new column + add a script to move values from `teams` table to `team_settings` using `teams.id = team_settings.team_id` as the WHERE clause.

### DIFF
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -1768,25 +1768,24 @@
       permission:
         check: {}
         columns:
-          - id
+          - boundary_bbox
+          - boundary_url
+          - email_reply_to_id
           - external_planning_site_name
           - external_planning_site_url
-          - homepage
           - help_email
           - help_opening_hours
           - help_phone
-          - email_reply_to_id
-          - team_id
-          - boundary_bbox
+          - homepage
+          - id
           - reference_code
-          - boundary_url
+          - submission_email
+          - team_id
       comment: ""
   select_permissions:
     - role: api
       permission:
         columns:
-          - id
-          - team_id
           - boundary_bbox
           - boundary_url
           - email_reply_to_id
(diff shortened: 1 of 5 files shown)


hasura.planx.uk/migrations/1724408755595_populate_team_settings_submissions_email/up.sql
```diff
@@ -0,0 +1,4 @@
+UPDATE team_settings
+SET submission_email = teams.submission_email
+FROM teams
+WHERE team_settings.team_id = teams.id;
```
jessicamcinchak, line 4: This all looks good! 

nit: I would prefer to see this `UPDATE...` statement squashed together into the same migration as the new column - then it will make more sense that this one doesn't have a corresponding `down.sql` !

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -1830,14 +1833,13 @@
           - homepage
           - id
           - reference_code
+          - submission_email
```
jessicamcinchak, line 1836: I don't think the `public` role should have access to this column? Please double-check existing permissions on current `teams.submission_email` column - it's specifically excluded from public role "select" access!

jessicamcinchak (commented):
Looking like a good start - please see one permissions-related comment!

hasura.planx.uk/metadata/tables.yaml
RODO94: Ah sorry! Good catch, I'll update now

jessicamcinchak (approved):
Ready to go, thanks for adjustments !

RODO94:

PR #3555, opened 2024-08-24
[api] migrate testing framework (Jest > Vitest)

freemvmt:
This PR is motivated by Jest's failure to handle ESM modules sufficiently well, which was a blocker on the Microsoft SSO work ([ticket](https://trello.com/c/owvV9UZh)). In the end we decided to migrate away from Jest, with Vitest being the obvious alternative, because a quick experiment with it showed promise (see [this comment](https://github.com/theopensystemslab/planx-new/pull/3528#pullrequestreview-2254037472)).

Ticket: https://trello.com/c/N1egz8at

So, this PR does the following:

- Add vitest package and configure (e.g. `vitest.config.ts`, setup files, `package.json` scripts etc)
- Adapts all tests in `api.planx.uk` to use `vitest` instead of `jest` - this was the most laborious step and I will document some of the thornier aspects in my own review shortly (Vitest's [dedicated migration guide](https://vitest.dev/guide/migration.html#migrating-from-jest) was very useful)
- Ensures coverage works similarly (e.g. relies on istanbul, keeps existing thresholds, displays in browser etc)
- Removes all things Jest

## A note on coverage

The `text-summary` coverage results are slightly different in this branch compared with `main`.

### main (jest)

```
Statements   : 78.74% ( 2489/3161 )
Branches     : 54.96% ( 299/544 )
Functions    : 67.67% ( 312/461 )
Lines        : 77.95% ( 2305/2957 )
```

### api-migrate-jest-vitest (vitest)

```
Statements   : 71.94% ( 1726/2399 )
Branches     : 54.92% ( 413/752 )
Functions    : 67.47% ( 305/452 )
Lines        : 71.75% ( 1659/2312 )
```

I haven't seriously investigated the cause of this discrepancy because it doesn't seem a substantial change. My instinct is that it's due to the default exclusions of each framework - e.g. as far as I can see, vitest has a [long list of exclusions](https://vitest.dev/config/#coverage-exclude), whereas [jest just ignores `node_modules`](https://jestjs.io/docs/configuration#coveragepathignorepatterns-arraystring) (or perhaps also excludes the [istanbul defaults](https://github.com/istanbuljs/schema/blob/master/default-exclude.js)). But it might be some other implementation detail.

If we wanted to try and create a culture of always-increasing coverage, we could adopt the `autoUpdate` flag, as per my comment in `vitest.config.ts`. But for now I've stuck with the singular minimum coverage threshold of 55% of functions.


api.planx.uk/lib/hasura/metadata/index.test.ts
```diff
@@ -1,32 +1,40 @@
...
-import Axios, { AxiosError } from "axios";
+import axios from "axios";
+import type { Mocked } from "vitest";
 
-jest.mock("axios", () => ({
-  ...jest.requireActual("axios"),
-  post: jest.fn(),
-}));
-const mockAxios = Axios as jest.Mocked<typeof Axios>;
+describe("Creation of scheduled event", () => {
```
freemvmt, line 5: Wrapped tests which I adapted in a `describe` if it didn't already exist 🫡

api.planx.uk/lib/hasura/metadata/index.test.ts
```diff
@@ -1,32 +1,40 @@
...
+  vi.mock("axios", async (importOriginal) => {
```
freemvmt, line 6: vitest has this `importOriginal` argument which is passed to the factory constructor in a `mock` call - we can use this to import the actual module (replacing `jest.requireActual`), as per [the docs](https://vitest.dev/guide/migration.html#importing-the-original-of-a-mocked-package).

Could also have used `vi.importActual` here, but using the arg seemed neater.

api.planx.uk/lib/hasura/metadata/index.test.ts
```diff
@@ -1,32 +1,40 @@
...
+    const actualAxios = await importOriginal<typeof import("axios")>();
+    return {
+      default: {
```
freemvmt, line 9: Vitest does not assume, like jest, that a returned object is the `default` export of given module, so we have to be explicit about it, as per [the docs](https://vitest.dev/guide/migration.html#module-mocks).

api.planx.uk/modules/admin/session/csv.test.ts
```diff
@@ -23,7 +23,9 @@ jest.mock("@opensystemslab/planx-core", () => {
 });
 
 describe("CSV data admin endpoint", () => {
-  afterEach(() => jest.clearAllMocks());
+  afterEach(() => {
```
freemvmt, line 26: hooks like `afterEach` have to return `null` or `undefined`, so have remove implicit return, as per [docs](https://vitest.dev/guide/migration.html#hooks)

api.planx.uk/modules/gis/service/classifiedRoads.test.ts
```diff
@@ -18,7 +18,7 @@ it("returns an error if required query param is missing", async () => {
 // "Success" test commented out due to reliance on external API calls and fallibility of nocks
 //   Please comment in and run locally if making changes to /roads functionality
 describe.skip("fetching classified roads data from OS Features API for any local authority", () => {
-  jest.setTimeout(10000);
+  vi.setConfig({ testTimeout: 1000 });
```
freemvmt, line 21: See relevant [docs](https://vitest.dev/guide/migration.html#timeout)

api.planx.uk/modules/gis/service/helpers.test.js
```diff
@@ -151,7 +151,6 @@ describe("rollupResultLayer helper function", () => {
     // Assert
     expect(result).toHaveProperty(key);
     layersToRollup.forEach((layer) => {
-      // Jest can handle paths using dot notation, so keys with a dot need to be wrapped in []
```
freemvmt, line 154: I assume this was supposed to say 'cannot handle...' ?

api.planx.uk/modules/pay/service/inviteToPay/createPaymentSendEvents.test.ts
```diff
@@ -3,9 +3,10 @@ import app from "../../../../server.js";
 import { createScheduledEvent } from "../../../../lib/hasura/metadata/index.js";
 import { queryMock } from "../../../../tests/graphqlQueryMock.js";
 import { flowWithInviteToPay } from "../../../../tests/mocks/inviteToPayData.js";
+import { MockedFunction } from "vitest";
```
freemvmt, line 6: no `vi` namespace so we have to explicitly import types like this one, as per [docs](https://vitest.dev/guide/migration.html#types)

api.planx.uk/modules/slack/index.test.ts
```diff
@@ -12,14 +12,16 @@ const invalidBody = {
   wrong: "message",
 };
 
-const mockSend = jest.fn();
-jest.mock<typeof SlackNotify>("slack-notify", () =>
```
freemvmt, line 16: tsc/vitest seem smart enough not to need the type declaration anymore 🤔

api.planx.uk/package.json
```diff
@@ -62,9 +62,9 @@
     "lint": "eslint '**/*.{js,ts}' && prettier -c .",
     "lint:fix": "eslint --fix '**/*.{js,ts}' && prettier -w .",
     "check": "tsc --noEmit && pnpm lint",
-    "test": "TZ=Europe/London NODE_OPTIONS='$NODE_OPTIONS --experimental-vm-modules' jest --no-cache",
-    "test:coverage": "TZ=Europe/London NODE_OPTIONS='$NODE_OPTIONS --experimental-vm-modules' jest; open ./coverage/lcov-report/index.html",
-    "test:watch": "TZ=Europe/London NODE_OPTIONS='$NODE_OPTIONS --experimental-vm-modules' jest --coverage=false --watch",
+    "test": "TZ=Europe/London vitest run",
+    "test:coverage": "TZ=Europe/London vitest run --coverage && open coverage/lcov-report/index.html",
+    "test:watch": "TZ=Europe/London vitest --ui --coverage",
```
freemvmt, line 67: `watch` mode now throws up a nice UI like below

Screenshot from 2024-08-24 16-39-50

freemvmt (commented):
Some comments for easier digestion!

api.planx.uk/modules/slack/index.test.ts
DafyddLlyr: Ah that's awesome because this step always felt awkward and brittle in jest imho 🙌

api.planx.uk/.eslintrc
```diff
@@ -1,12 +1,11 @@
 {
   "root": true,
   "parser": "@typescript-eslint/parser",
-  "plugins": ["@typescript-eslint", "jest"],
+  "plugins": ["@typescript-eslint"],
   "extends": [
     "eslint:recommended",
     "plugin:@typescript-eslint/recommended",
-    "prettier",
-    "plugin:jest/recommended"
```
DafyddLlyr, line 9: Super happy for this to be a follow up PR - it might be worth looking at something like https://www.npmjs.com/package/eslint-plugin-vitest to replace this?

api.planx.uk/modules/gis/service/helpers.test.js
DafyddLlyr: I actually think "can" was correct here - the key here would have been something like `a.b.c` which Jest would have tried to default to a nested path, as opposed to the object in question which is likely `{ "a.b.c": "x" }`

Either way, it's a good call to remove this comment - it looks like vitest [handles this more gracefully](https://vitest.dev/api/expect.html#tohaveproperty) - the test actually passes with or without the wrapping `[]` on line 154.

api.planx.uk/vitest.config.ts
DafyddLlyr, line 1: Really appreciate the comments in this config file 💪

api.planx.uk/vitest.config.ts
```diff
@@ -0,0 +1,30 @@
...
+    globalSetup: ["./tests/setup/global.ts"],
+    // runs before each test file
+    setupFiles: ["./tests/setup/graphQL.ts"],
+    coverage: {
+      provider: "istanbul",
+      // html reporter required to inspect coverage in Vitest UI dashboard
+      reporter: ["lcov", "html", "text-summary"],
+      thresholds: {
+        functions: 55,
+        // TODO: could add autoUpdate flag here so that function coverage is only allowed to increase
```
DafyddLlyr, line 17: Let's do this and leave this better than we found it 👍

DafyddLlyr (approved):
Fantastic PR - thanks so much for the documentation links and explanations, super helpful.

DafyddLlyr:
Merged in @freemvmt 's absence. 

I'll pick up the two (very small) tasks suggested above in the comments - 
 - Use the `autoUpdate` flag for API tests
 - Replace ESLint `"plugin:jest/recommended"` with a vitest equivalent

PR #3556, opened 2024-08-26
chore: add error feedback for editors if using List with "map" field without FindProperty first

jessicamcinchak:
The List component's new "map" field requires address data from state to position the map viewport - this means the component needs to come after a "FindProperty". Currently in testing, if you don't remember this, the larger List will load but the map field will show an empty/non-loading map. 

Now just throw a quick "graph error" (same pattern as other postion-dependent geo components like PlanningConstraints)!
Screenshot from 2024-08-26 10-30-12

Tests for the "map" fields still outstanding / to-be picked up in follow-up PR


jessicamcinchak:

DafyddLlyr (approved):
Working as expected - perfect!

PR #3549, opened 2024-08-22
feat: add `vitest-axe` & update frontend tests

jessicamcinchak:
**What is working here**
- Re-adds editor package.json script `test:silent` which CI Github workflows expect
- Swaps out `jest-axe` for `vitest-axe` & updates all a11y tests in `src/@planx/components` & beyond
  - Please note we have to install the pre-release version in order to access `toHaveNoViolations` ([see thread](https://github.com/chaance/vitest-axe/issues/13))
  - Adds `import "vitest-axe/extend-expect";` to `src/setupTests.ts`
  - Imports `axe` directly from `vitest-axe` and drops special config from `src/testUtils.ts` - does this seem okay? I don't think I've hit the SVG issue described in the code comment there yet / guessing it's old & since been resolved?
- Updates LOTS of other `jest` -> `vi` mocks along the way
- Adds per-test timeouts where necessary (see especially some public List component tests or editor Pay tests)
- Un-comments tests & `.skip()`s problematic ones so they're easier to globally search for in the final migration stretch! 

Screenshot from 2024-08-23 20-37-59

**What's not working yet / let's address in a follow up PR ?**
- All tests relying on `mockedAxios` are skipped - getting a `mockResolvedValueOnce` function not defined error (FileUploadAndLabel, Send)
- All tests relying on `vi.spyOn()` for the "default" function are skipped ([see thread](https://github.com/vitest-dev/vitest/issues/855), haven't quite sorted solution yet) (PlanningConstraints)
- Airbrake notifier tests - again a more complicated mock / spy to figure out!

Some unexpected linting stuff picked up with these changes has added extra noise to an already big PR, sorry!

PR #3558, opened 2024-08-26
test: Setup and manage Airbrake mocks and spies in vitest

DafyddLlyr:
Please see https://github.com/theopensystemslab/planx-new/pull/3549 for context

### DIFF
--- a/editor.planx.uk/src/airbrake.test.ts
+++ b/editor.planx.uk/src/airbrake.test.ts
@@ -1,68 +1,73 @@
-// import { Notifier } from "@airbrake/browser";
-import { vi } from "vitest";
+import { Notifier } from "@airbrake/browser";
+import { MockInstance, vi } from "vitest";
 
 vi.mock("@airbrake/browser");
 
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/airbrake.test.ts
```diff
@@ -1,68 +1,73 @@
...
+    vi.resetModules();
+  });
+
+  afterEach(() => {
+    logSpy.mockRestore();
+    vi.mocked(Notifier).mockClear();
+  });
+
+  afterAll(() => {
+    vi.unstubAllEnvs();
```
jessicamcinchak, line 27: Really enjoying how clear `vi` built-in utilities are for steps like this :+1:

jessicamcinchak (approved):
Perfect, thanks !

editor.planx.uk/src/airbrake.test.ts
DafyddLlyr: 💯

PR #3565, opened 2024-08-27
Production deploy

jessicamcinchak:
If we're happy wtih the state of the Vitest migration on the React app side, let's do a prod deploy! Notably, there's a small fix to planning constraints loading here that was raised last Friday afternoon in `#help-issues` that would be nice to get across: https://opendigitalplanning.slack.com/archives/C0241GWFG4B/p1724419199641959


DafyddLlyr (approved):
Yep, happy to go ahead here. We'll flag at standup that we're using a new build toolchain for the frontend and just ask George / August / Jonny to flag any issues they might hit 👍

PR #3568, opened 2024-08-27
feat: display static map for inactive List "map" fields

jessicamcinchak:
Last piece in order to have an end-to-end List option demo ready for tmrw morning - shows a static map when the List item is inactive for any "map" field types. 

Hit a funny quirk here testing with "points" where I realised a limitation in the map repository - our `geojsonData` layer has only ever accounted for polygons before and doesn't currently have a "point" style associated with it ! So will make that update in `@opensytemslab/map` for a future release, and in the meantime we just slightly adjust our rendering rules based on `drawType` here :+1: 

Screenshot from 2024-08-27 12-24-09

(Also sorry Rory @RODO94 I know we bookmarked this task to your name last week, but trying to keep momentum ahead of demos & UR deadline! will be plenty more followup tasks when you're ready to come back into works to trees !)

### DIFF
--- a/editor.planx.uk/src/@planx/components/List/utils.tsx
+++ b/editor.planx.uk/src/@planx/components/List/utils.tsx
@@ -43,6 +43,60 @@ export function formatSchemaDisplayValue(
       );
       return matchingOption?.data.text;
     }
+    case "map": {
+      const feature = value[0] as string as any; // won't be necessary to cast once we're only setting "geojsonData" prop in future
+      const drawType = field.data.mapOptions?.drawType;
+
+      switch (drawType) {
+        case "Point":
+          // Our "geojsonData" layer doesn't have a "point" style yet, so make due with center marker for now!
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/@planx/components/List/utils.tsx
```diff
@@ -43,6 +43,60 @@ export function formatSchemaDisplayValue(
       );
       return matchingOption?.data.text;
     }
+    case "map": {
+      const feature = value[0] as string as any; // won't be necessary to cast once we're only setting "geojsonData" prop in future
+      const drawType = field.data.mapOptions?.drawType;
+
+      switch (drawType) {
+        case "Point":
+          // Our "geojsonData" layer doesn't have a "point" style yet, so make due with center marker for now!
```
DafyddLlyr, line 52: Ha! Totally hadn't appreciated this, smart solution here!

DafyddLlyr (approved):

PR #3571, opened 2024-08-27
fix: add submission_email to team_settings seed script

RODO94:
## What does this PR do?

Previous PR: https://github.com/theopensystemslab/planx-new/pull/3554 - added `submission_email` to the `team_settings` table but it did not make the change in the scripts which add data to the local database instances. 

This PR fixes that and adds the relevant changes to the scripts.

PR #3575, opened 2024-08-27
fix: submission_email being populated from PROD

RODO94:
Seen from my previous fix: https://github.com/theopensystemslab/planx-new/pull/3571, that my database was being populated with Production data, which is not aligned with how it is scripted for the `teams` table.

I have altered that now so we don't send anything to council submission emails by accident when working on our local branches


ianjon3s (approved):
Well spotted, could've caused a couple of headaches

RODO94:

PR #3576, opened 2024-08-27
feat: Setup types for search data facets

DafyddLlyr:
## What does this PR do?
- Sets up types which define keys, or other accessors, for data fields across PlanX components
- Groundwork for a follow up PR which will implement these to visually display all component types as search results

## Next steps...
As seen in the very rough https://github.com/theopensystemslab/planx-new/pull/3476 there's a bit of a matrix of complexity here - we need to map each component type, for each data field, for 4 or 5 display fields.

The next PR will implement these types, likely pulling this logic out from nested conditional logic into an object or some data structure to hold this.

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/facets.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/facets.ts
@@ -1,0 +1,29 @@
+import { IndexedNode } from "@opensystemslab/planx-core/types";
+import { FuseOptionKey } from "fuse.js";
+
+type SearchFacets = Array<FuseOptionKey<IndexedNode>>;
+
+const generalData: SearchFacets = ["data.fn", "data.val"];
+
+const fileUploadAndLabelData: SearchFacets = ["data.fileTypes.fn"];
+
+const calculateData: SearchFacets = [
+  "data.output",
+  {
+    name: "formula",
+    getFn: (node: IndexedNode) => Object.keys(node.data?.defaults || {}),
+  },
+];
+
+const listData: SearchFacets = [
+  "data.schema.fields.data.fn",
+  "data.schema.fields.data.options.data.val",
+];
+
+/** Data fields used across PlanX components */
+export const DATA_FACETS = [
+  ...generalData,
+  ...fileUploadAndLabelData,
+  ...calculateData,
+  ...listData,
+];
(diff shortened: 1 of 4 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/facets.ts
```diff
@@ -0,0 +1,29 @@
+import { IndexedNode } from "@opensystemslab/planx-core/types";
+import { FuseOptionKey } from "fuse.js";
+
+type SearchFacets = Array<FuseOptionKey<IndexedNode>>;
```
jamdelion, line 4: It isn't really obvious to me what 'facets' are but I see that they are already a named thing in the codebase and possibly a standard word in search components! 😅

jamdelion (approved):

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/facets.ts
```diff
@@ -0,0 +1,29 @@
...
+  "data.schema.fields.data.options.data.val",
+];
+
+/** Data fields used across PlanX components */
+export const DATA_FACETS = [
+  ...generalData,
+  ...fileUploadAndLabelData,
+  ...calculateData,
+  ...listData,
+];
```
jessicamcinchak, line 29: Reminder `DrawBoundary` is another "special" case here I think ! 
```
const drawBoundaryData: SearchFacets = ["dataFieldBoundary", "dataFieldArea"];
```

jessicamcinchak (approved):

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/facets.ts
DafyddLlyr: It was also a new term to me! https://en.wikipedia.org/wiki/Faceted_search

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/facets.ts
DafyddLlyr: Very good catch thank you - missed this one!

Picked up in [bc27773](https://github.com/theopensystemslab/planx-new/pull/3576/commits/bc27773258a5482c38ce61c0006ce90fb0b01d06)

PR #3577, opened 2024-08-27
feat: add Lost/Gained list schema to GLA residential units

augustlindemer:
Add Lost/Gained list schema to GLA residential units in addition to current New/Retained/Rebuilt/Removed distinction. The current structure attempts to capture both national and GLA requirements in one structure. GLA councils do however not capture the nationally required information in their current form submissions. This is an alternative and simplified mode of capturing the GLA information.


DafyddLlyr:
Closed in favour of https://github.com/theopensystemslab/planx-new/pull/3588

PR #3578, opened 2024-08-27
refactor: Move editor console to tab

ianjon3s:
## What does this PR do?

🚧 **Edit sidebar icon replacement pt2** 🚧

Replaces console icon with a dedicated tab for the console. This is given a 'dark' theme with corresponding prop for dark tabs.

I've also moved `StyledTab` to its own file, I was unsure whether this should remain in the current folder, or be moved to the wider UI components folders, as it is only used in this place. Open to suggestions on this one.

Example:
https://3578.planx.pizza/lambeth/apply-for-planning-permission

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
@@ -11,7 +11,6 @@ import DialogContent from "@mui/material/DialogContent";
 import DialogTitle from "@mui/material/DialogTitle";
 import Link from "@mui/material/Link";
 import { styled } from "@mui/material/styles";
-import Tab, { tabClasses } from "@mui/material/Tab";
 import Tabs from "@mui/material/Tabs";
 import Tooltip from "@mui/material/Tooltip";
 import Typography from "@mui/material/Typography";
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
```diff
@@ -415,29 +390,12 @@ const Sidebar: React.FC<{
...
-            <StyledTab
-              disableFocusRipple
-              disableTouchRipple
-              disableRipple
-              value="Search"
-              label="Search"
-            />
+            <StyledTab value="Search" label="Search" tabTheme="light" />
           )}
+          <StyledTab value="Console" label="Console" tabTheme="dark" />
```
jessicamcinchak, line 398: Much tidier :raised_hands:

jessicamcinchak (approved):
I like this, feels much more like a "console" and always great to reduce overlapping windows / a lot of vertical scrolling ! 

Separate file for `StyledTab` feels very clean to maintain :heavy_plus_sign:

DafyddLlyr (approved):
Solid improvement! 

Small nit / personal opinion would be that the black is maybe too dark. I'm not sure if a colour close to the header would work a bit better maybe?

I think there's a whole redesign we could do here anyway, and using a lib like https://react-json-view.microlink.io/ to make the whole "console" concept a whole lot more useful!

ianjon3s:
Thanks both, I agree the dark is somewhat overpowering, I've changed to the same `background.dark` colour used in the header. I'll look at introducing the JSON formatting in a follow-up PR.

PR #3582, opened 2024-08-28
feat: API changes for handling new `submission_email` location

RODO94:
## What does this PR do?

This PR looks to alter references from `teams.submission_email` to `team_settings.submission_email`.

I have ran through all references to `submission_email`, `sendEmail`, `submissionEmail` and repointed them to the new reference.

### DIFF
--- a/api.planx.uk/modules/send/email/index.ts
+++ b/api.planx.uk/modules/send/email/index.ts
@@ -11,7 +11,7 @@ import {
 export async function sendToEmail(
   req: Request,
   res: Response,
-  next: NextFunction
+  next: NextFunction,
 ) {
   req.setTimeout(120 * 1000); // Temporary bump to address submission timeouts
 
@@ -64,7 +64,7 @@ export async function sendToEmail(
       localAuthority,
       sendToEmail,
       config,
-      response
+      response,
     );
 
     return res.status(200).send({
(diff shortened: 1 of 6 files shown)


RODO94:
Quick Pizza link for testing: https://3582.planx.pizza/buckinghamshire/submission-email-test

api.planx.uk/modules/send/email/index.ts
```diff
@@ -27,10 +27,11 @@ export async function sendToEmail(
   }
 
   try {
-    // Confirm this local authority (aka team) has an email configured in teams.submission_email
-    const { sendToEmail, notifyPersonalisation } =
+    // Confirm this local authority (aka team) has an email configured in team_settings.submission_email
+    const { notifyPersonalisation } =
```
DafyddLlyr, line 31: nit: If we're only using the `sendToEmail` field in this file, and not the rest of the `notifyPersonalisation` object, it would be cleaner / neater to destructure one level deeper here, and then not have to repeatedly write `notifyPersonalisation.sendToEmail` throughout the file 👍 

```suggestion
    const { notifyPersonalisation: { sendToEmail } } =
```

DafyddLlyr (approved):
Looking good!

api.planx.uk/modules/send/email/index.ts
RODO94: I think we are also using `...notifyPersonalisation` for part of the code below. I'll keep as is now.

RODO94:

PR #3581, opened 2024-08-28
chore: Use vite to build Storybook

DafyddLlyr:
## What does this PR do?
 - Follows on from https://github.com/theopensystemslab/planx-new/pull/3544
 - Replaces webpack with vite to build storybook
 - Upgrades `.js` files to `.ts` files
 - Updates syntax to match Vite & ESM

Pretty much all handled by `npx storybook automigrate` which was just great 🙌 

**Demo** - https://storybook.3581.planx.pizza

## Next steps...

The following was also generated by `npx storybook automigrate` - 

```
🩺 The doctor is checking the health of your Storybook..
╭ Incompatible packages found ───────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                                    │
│   The following packages are incompatible with Storybook 7.6.7 as they depend on different major versions of Storybook packages:   │
│   - @storybook/react-vite@7.6.20                                                                                                   │
│    Repo: https://github.com/storybookjs/storybook/tree/next/code/frameworks/react-vite                                             │
│   - @storybook/test@7.6.20                                                                                                         │
│    Repo: https://github.com/storybookjs/storybook/tree/next/code/lib/test                                                          │
│   - storybook@8.1.10                                                                                                               │
│    Repo: https://github.com/storybookjs/storybook/tree/next/code/lib/cli                                                           │
│   - storybook-addon-material-ui@0.9.0-alpha.24                                                                                     │
│    Repo: https://github.com/react-theming/storybook-addon-material-ui                                                              │
│                                                                                                                                    │
│                                                                                                                                    │
│   Please consider updating your packages or contacting the maintainers for compatibility details.                                  │
│   For more on Storybook 8 compatibility, see the linked GitHub issue:                                                              │
│   https://github.com/storybookjs/storybook/issues/26031                                                                            │
│                                                                                                                                    │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

You can always recheck the health of your project by running:
npx storybook doctor

Full logs are available in /Users/dafyddllyr/Code/planx-new/editor.planx.uk/doctor-storybook.log

```

Not updated here, but it would make sense to pick up these updates shortly also - I'll open another PR once this is tested and approved.

PR #3584, opened 2024-08-28
chore: Bump Storybook dependencies to v8

DafyddLlyr:
## What does this PR do?
 - Bumps all Storybook dependencies to v8
 - Follow on from https://github.com/theopensystemslab/planx-new/pull/3581


editor.planx.uk/.storybook/manager.ts
```diff
@@ -1,4 +1,4 @@
-import { addons } from "@storybook/addons";
+import { addons } from "@storybook/manager-api";
```
DafyddLlyr, line 1: Storybook now uses manager-api instead of addons

jessicamcinchak (approved):

PR #3585, opened 2024-08-28
feat: Setup Storybook for `MapAndLabel`

DafyddLlyr:
## What does this PR do?
 - Sets up `MapAndLabel.stories.tsx`
 - Splits up store logic from the rest of the component for simpler testing - the same pattern used in multiple other components for import to Storybook

## Motivation
I started looking at [the next item here to pick up](https://trello.com/c/JafdzU94/2855-epic-allow-applicants-to-work-to-trees-to-see-and-add-more-information-about-protected-trees-on-a-map) and was hitting a few issues setting up state before implementing tests. It seemed worthwhile to handle this once now before we start writing tests and then having to refactor them.

Also, having Storybook should prove helpful to avoid the `FindProperty` step beforehand when working on this and testing!

### DIFF
--- a/editor.planx.uk/.storybook/preview.tsx
+++ b/editor.planx.uk/.storybook/preview.tsx
@@ -4,8 +4,13 @@ import { DndProvider } from "react-dnd";
 import { HTML5Backend } from "react-dnd-html5-backend";
 import { defaultTheme } from "../src/theme";
 import React from "react";
+import { MyMap } from "@opensystemslab/map";
 
 import { reactNaviDecorator } from "./__mocks__/react-navi";
+
+if (!window.customElements.get("my-map")) {
+  window.customElements.define("my-map", MyMap);
+}
 
 export const decorators = [
   (Story) => (
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/MapAndLabel.stories.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/MapAndLabel.stories.tsx
@@ -1,0 +1,37 @@
+import { Meta, StoryObj } from "@storybook/react";
+import React from "react";
+
+import Wrapper from "../../fixtures/Wrapper";
+import Editor from "../Editor";
+import { Trees } from "../schemas/Trees";
+import { Presentational as Public, PresentationalProps } from ".";
+
+const meta = {
+  title: "PlanX Components/MapAndLabel",
+  component: Public,
+} satisfies Meta<typeof Public>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+const props: PresentationalProps = {
+  title: "Map and label the works to trees for this property",
+  description: "Add one or many trees below",
+  schemaName: "Trees",
+  fn: "MockFn",
+  schema: Trees,
+  basemap: "OSM",
+  drawColor: "#00FF00",
+  drawType: "Point",
+  longitude: -0.1629784,
+  latitude: 51.5230919,
+};
+
+export const Basic: Story = {
+  args: props,
+};
+
+export const WithEditor = () => {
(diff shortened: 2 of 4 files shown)


editor.planx.uk/.storybook/preview.tsx
```diff
@@ -4,9 +4,14 @@ import { DndProvider } from "react-dnd";
 import { HTML5Backend } from "react-dnd-html5-backend";
 import { defaultTheme } from "../src/theme";
 import React from "react";
+import { MyMap } from "@opensystemslab/map";
 
 import { reactNaviDecorator } from "./__mocks__/react-navi";
 
+if (!window.customElements.get("my-map")) {
+  window.customElements.define("my-map", MyMap);
+}
```
DafyddLlyr, line 13: Required to view the web component in Storybook.

editor.planx.uk/src/@planx/components/MapAndLabel/Public/MapAndLabel.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+type Story = StoryObj<typeof meta>;
+
+const props: PresentationalProps = {
+  title: "Map and label the works to trees for this property",
+  description: "Add one or many trees below",
+  schemaName: "Trees",
+  fn: "MockFn",
+  schema: Trees,
+  basemap: "OSM",
```
DafyddLlyr, line 24: When set to `MapboxSatellite` I get permission errors (works fine when running the application).

I think we need to allow `localhost:6006` (Storybook) on MapBox for this to work.

jessicamcinchak: Good catch! Added `http://localhost:6006` to allowed URLs for our "Planx local dev" token (ends with `NW1Q`) if you can check again? Hosted Storybook should be properly captured under the second key for our planx subdomains (token ends with `au8A`). 

These are are available named tokens for future ref:
Screenshot from 2024-08-28 20-45-40

editor.planx.uk/.storybook/preview.tsx
jessicamcinchak: :bulb: this is great! I actually thought this was a limitation of _React_ Storybook and not possible to render native web components - excited to backfill and correctly configure other geocomponents too :raised_hands:

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
```diff
@@ -221,11 +201,47 @@ const Root = () => {
...
   return (
-    <MapAndLabelProvider {...props}>
-      <Root />
-    </MapAndLabelProvider>
+    <Presentational
+      {...props}
+      latitude={latitude}
+      longitude={longitude}
+      boundaryBBox={teamSettings.boundaryBBox}
+    />
```
jessicamcinchak, line 244: Nice refactor :heavy_plus_sign:

jessicamcinchak (approved):
Thanks for this!!

editor.planx.uk/src/@planx/components/MapAndLabel/Public/MapAndLabel.stories.tsx
DafyddLlyr: Still getting this issue locally, the key I'm using does match what you've provided above. Not a silly issue with a trailing whitespace or slash or something? (seems unlikely!)

Either way, I don't think this is necessarily blocking just now but would be nice to resolve.

jessicamcinchak: Will double check this one with fresh eyes this morning & test locally - sure we can get it sorted going forward :+1:

PR #3589, opened 2024-08-29
feat: Basic validation for `MapAndLabel`

DafyddLlyr:
## What does this PR do?
- Wires up Formik validation on the `MapAndLabel` component
  - Validation is triggered when hitting continue
  - The minimum number of items is checked
  - The maximum number of items is checked
  - Errors in tabs are highlighted - certainly not advocating for this to be the final style but I just wanted to get the data flow set up here. 
 - Removes some List specific functionality from `MapAndLabelContext`
 - Starts using the language of "features" instead of "items" - feels a little clearer and less generic, but super open to feedback here.
 - No tests! Interaction with the Map webcomponent is proving tricky - I'll have a PR later today with my various attempts documented, and I'll work on mocking the GeoJSON change event as an alternative.

## Demo

https://github.com/user-attachments/assets/035be7c6-50d7-4a78-abb1-79501b7078fe

### DIFF
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
@@ -44,7 +44,7 @@ function a11yProps(index: number) {
 const VerticalFeatureTabs: React.FC<{ features: Feature[] }> = ({
   features,
 }) => {
-  const { schema, activeIndex, formik, editItem, isTabInvalid } =
+  const { schema, activeIndex, formik, editFeature, isFeatureInvalid } =
     useMapAndLabelContext();
 
   return (
@@ -62,7 +62,7 @@ const VerticalFeatureTabs: React.FC<{ features: Feature[] }> = ({
           variant="scrollable"
           value={activeIndex.toString()}
           onChange={(_e, newValue) => {
-            editItem(parseInt(newValue, 10));
+            editFeature(parseInt(newValue, 10));
           }}
           // TODO!
           aria-label="Vertical tabs example"
@@ -74,7 +74,7 @@ const VerticalFeatureTabs: React.FC<{ features: Feature[] }> = ({
               value={i.toString()}
               label={`${schema.type} ${feature.properties?.label}`}
               {...a11yProps(i)}
-              {...(isTabInvalid(i) && {
+              {...(isFeatureInvalid(i) && {
                 sx: (theme) => ({
                   backgroundColor: theme.palette.action.focus,
                 }),
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
@@ -44,7 +44,8 @@ function a11yProps(index: number) {
 const VerticalFeatureTabs: React.FC<{ features: Feature[] }> = ({
   features,
 }) => {
-  const { schema, activeIndex, formik, editItem } = useMapAndLabelContext();
+  const { schema, activeIndex, formik, editItem, isTabInvalid } =
+    useMapAndLabelContext();
 
   return (
     <Box
@@ -73,6 +74,11 @@ const VerticalFeatureTabs: React.FC<{ features: Feature[] }> = ({
               value={i.toString()}
               label={`${schema.type} ${feature.properties?.label}`}
               {...a11yProps(i)}
+              {...(isTabInvalid(i) && {
+                sx: (theme) => ({
+                  backgroundColor: theme.palette.action.focus,
+                }),
+              })}
             />
           ))}
         </Tabs>
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
@@ -17,6 +17,7 @@ import { FONT_WEIGHT_SEMI_BOLD } from "theme";
 import SelectInput from "ui/editor/SelectInput";
 import FullWidthWrapper from "ui/public/FullWidthWrapper";
 import InputLabel from "ui/public/InputLabel";
+import ErrorWrapper from "ui/shared/ErrorWrapper";
 
 import Card from "../../shared/Preview/Card";
 import CardHeader from "../../shared/Preview/CardHeader";
@@ -171,7 +172,8 @@ const VerticalFeatureTabs: React.FC<{ features: Feature[] }> = ({
 };
 
 const Root = () => {
-  const { validateAndSubmitForm, mapAndLabelProps } = useMapAndLabelContext();
+  const { validateAndSubmitForm, mapAndLabelProps, errors } =
+    useMapAndLabelContext();
   const {
     title,
     description,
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
```diff
@@ -61,31 +56,46 @@ const VerticalFeatureTabs: React.FC<{ features: Feature[] }> = ({
         maxHeight: "fit-content",
       }}
     >
-      <TabContext value={activeTab}>
+      <TabContext value={activeIndex.toString()}>
```
DafyddLlyr, line 59: I feel like I'm slightly missing something here!

MUI allows `TabContext` and `Tab` to take `string | number` for `value`, but `TabPanel` only accepts `string`?

What's happening here is we're using the `activeIndex`, as a string, to control the tab state instead of managing two variables.

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
```diff
@@ -61,31 +56,46 @@ const VerticalFeatureTabs: React.FC<{ features: Feature[] }> = ({
...
               key={`tab-${i}`}
-              value={feature.properties?.label}
+              value={i.toString()}
               label={`${schema.type} ${feature.properties?.label}`}
               {...a11yProps(i)}
+              {...(isFeatureInvalid(i) && {
+                sx: (theme) => ({
+                  backgroundColor: theme.palette.action.focus,
+                }),
+              })}
```
DafyddLlyr, line 81: This is how we get yellow tabs on error - styling very much TBC here.

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
```diff
@@ -94,44 +104,60 @@ const VerticalFeatureTabs: React.FC<{ features: Feature[] }> = ({
...
-                        key={option.properties?.label}
-                        value={option.properties?.label}
-                      >
-                        {`${schema.type} ${option.properties?.label}`}
-                      </MenuItem>
-                    ))}
-                  </SelectInput>
-                </InputLabel>
-              </Box>
+              {features.length > 1 && (
```
DafyddLlyr, line 114: Only show the copy select once we have multiple features.

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
```diff
@@ -94,44 +104,60 @@ const VerticalFeatureTabs: React.FC<{ features: Feature[] }> = ({
...
+                      title={"Copy from"}
+                      labelId={`select-label-${i}`}
+                      value={""}
+                      onChange={() =>
+                        console.log(`TODO - Copy data from another tab`)
+                      }
+                      name={""}
+                      style={{ width: "200px" }}
+                    >
+                      {/* Iterate over all other features */}
```
DafyddLlyr, line 129: ...and only list all _other_ features in the dropdown.

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
```diff
@@ -179,7 +208,12 @@ const Root = () => {
     return function cleanup() {
       map?.removeEventListener("geojsonChange", geojsonChangeHandler);
     };
-  }, [setFeatures]);
+  }, [setFeatures, addFeature]);
+
+  const rootError: string =
+    (errors.min && `You must provide at least ${schema.min} response(s)`) ||
+    (errors.max && `You can provide at most ${schema.max} response(s)`) ||
+    "";
```
DafyddLlyr, line 216: I've copied this verbatim from the List component - I'm not sure the concept of "response(s)" really holds up so well in this context?

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
```diff
@@ -94,44 +104,60 @@ const VerticalFeatureTabs: React.FC<{ features: Feature[] }> = ({
...
                   {feature.geometry.type === "Point"
                     ? ` (${feature.geometry.coordinates.map((coord) =>
-                      coord.toFixed(5),
-                    )})`
-                    : ` (area ${feature.properties?.["area.squareMetres"] || 0
-                    } m²)`}
+                        coord.toFixed(5),
+                      )})`
+                    : ` (area ${
+                        feature.properties?.["area.squareMetres"] || 0
```
RODO94, line 110: Guessing this was just a stray return...

RODO94 (approved):
All looks good to me, tried it out a few times and all worked as expected.

Only one thing is weird, and I have noted it in the Trello... when I add multiple trees, there's situations where all the labels change.

https://github.com/user-attachments/assets/deb650d8-830c-4caa-aa41-5c9435d694d9

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
jessicamcinchak: I think we'll want this language to be something like `"You must plot at least ${schema.min} ${schema.name} on the map"` - but very easy to fix forward !

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
jessicamcinchak: Hit this as well with initial tab implementation - it's a strange one - managing one variable definitely feels cleaner tho :ok_hand:

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
jessicamcinchak: I think a simple style option here is to make the default `borderRight` on the tab our `error` color and double the thickness when invalid? But again, super easy to fix forward / adjust later - thanks for highlighting placeholder here !

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
jessicamcinchak: a11y question: would it be better to be always visible _but disabled_ until we have multiple features? Maybe with conditional `visuallyHidden` text that announces something like "Add another feature first before copying" 

Just thinking if we conditionally render we'll have to probably sort out aria announcements which might get complicated/awkward quickly - eg:
- Plot a point, add tab one which is active tab by default, no copy
- Plot another point, add tab two which is now active tab, has copy - announce that tab one (the inactive tab) also has copy available now??

jessicamcinchak (approved):
Super solid foundation to keep building on - thank you!!

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
DafyddLlyr: That would work well! I've added a styling item to https://trello.com/c/JafdzU94/2855-epic-allow-applicants-to-work-to-trees-to-see-and-add-more-information-about-protected-trees-on-a-map to make sure this gets picked up here.

Going with the fix-forward approach due to slightly limited time this week and other PRs which are based on this branch 👍

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
DafyddLlyr: This is a very good point which I've overlooked, I'll pick this up in https://github.com/theopensystemslab/planx-new/pull/3594

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
DafyddLlyr: Yep this is a lot better! Will fix-forward in next PR 👍

DafyddLlyr:
> Only one thing is weird, and I have noted it in the Trello... when I add multiple trees, there's situations where all the labels change.

@RODO94 Thanks for catching and adding to Trello! 🙌

PR #3591, opened 2024-08-29
style: Simplify editor sidebar tabs

ianjon3s:
## What does this PR do?

Simplify the sidebar tabs so that they don't require a colour theme prop, and also take up less visual space.

Before:
image

After:
image

Preview:
https://3591.planx.pizza/lambeth/apply-for-planning-permission

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/StyledTab.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/StyledTab.tsx
@@ -19,7 +19,7 @@ const StyledTab = styled((props: TabProps) => (
   [`&.${tabClasses.selected}`]: {
     fontWeight: FONT_WEIGHT_SEMI_BOLD,
     color: theme.palette.text.primary,
-    boxShadow: `inset 0 -3px 0 ${theme.palette.warning.main}`,
+    boxShadow: `inset 0 -3px 0 ${theme.palette.info.main}`,
   },
 }));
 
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Sidebar/StyledTab.tsx
```diff
@@ -3,36 +3,23 @@ import Tab, { tabClasses, TabProps } from "@mui/material/Tab";
...
-        : theme.palette.background.default,
-    borderColor: theme.palette.border.main,
-    borderBottomColor: theme.palette.common.white,
-    color:
-      tabTheme === "dark"
-        ? theme.palette.common.white
-        : theme.palette.text.primary,
+    fontWeight: FONT_WEIGHT_SEMI_BOLD,
+    color: theme.palette.text.primary,
+    boxShadow: `inset 0 -3px 0 ${theme.palette.warning.main}`,
```
jessicamcinchak, line 22: nit: not sure I like the use of the "warning" color for the underline now that we've been incorporating it for actual warnings (eg publish validation checks) - would the original blue be better / more neutral?

jessicamcinchak (approved):
Code looks great! One small visual opinion, very much non-blocking / you're the designer so please take it or leave it !

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/StyledTab.tsx
ianjon3s: Good shout, I was going for something to satisfies the following:
1.  Is somewhere visually between and contrasty to both black & white, so is visible as a marker for both dark & light tabs
2. Is not the primary action colour, as we use this to mark something that should be interacted with (rather than something that is "active")

The warning colour was an easy/lazy choice here (also leaning a bit too far into Github's tab styling 😀). I'll update to be the status agnostic "info" light blue colour that we use in the PlanX public interface, as this also satisfies the above.

PR #3592, opened 2024-08-29
chore: add basic UI components for Map & Label "remove" & "copy"

jessicamcinchak:
"Copy from" and "Remove" don't work yet! Just stubbing them out so easier for anyone to come in and "wire up" or update styles next.

Screenshot from 2024-08-29 17-13-05

### DIFF
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
@@ -1,18 +1,23 @@
+import DeleteIcon from "@mui/icons-material/Delete";
 import TabContext from "@mui/lab/TabContext";
 import TabPanel from "@mui/lab/TabPanel";
 import Box from "@mui/material/Box";
+import Button from "@mui/material/Button";
+import MenuItem from "@mui/material/MenuItem";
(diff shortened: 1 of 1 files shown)


ianjon3s (approved):
Great additions and demonstration scope of functionality well!

RODO94:
@jessicamcinchak using this, I wonder if field level select would work better (e.g you can select from a list of species / proposed works that have already been entered) thinking is:

1. When there's a lot of trees, I found it hard to look through all the options and remember which one is Tree(n)
2. Could be more flexible if you have 10 spruce trees but they all have different works proposed

Not attached to this idea, just my thought when using it in the pizza.

jessicamcinchak:
@RODO94 good questions - but I'm having a hard time picturing how a field-level "Copy from" would play out in the UI and limited real estate of tabs without becoming quite busy? 

Let's keep thinking on this and add to Trello in a similar way of v1/v2 validation complexity (eg v1/v2 copy complexity)?

The idea here is that "Copy from" is just a quick way to autopopulate all values & save repeat typing, but then you can of course keep editing/overwrite if a certain field is different for this particular tree - which hopefully with larger map label text is easier to orient to/remember !

PR #3593, opened 2024-08-29
refactor: add useToast hook for toast notifications

jamdelion:
## What is this?

Replacing everywhere we use:

```js
<Snackbar onClose={handleCloseToast} autoHideDuration={6000} open={true}>
      <Alert onClose={handleCloseToast} severity={'success'} sx={{ width: "100%" }}>
        {message}
      </Alert>
    </Snackbar>
```

With `toast.success(message)` 🎉 

## Testing

An easy place to test this is in the Team Members page, when you've successfully added a new editor. You should see the usual toast message pop up on success.


editor.planx.uk/src/contexts/ToastContext.tsx
```diff
@@ -0,0 +1,70 @@
...
+const initialState: ToastState = {
+  toasts: [],
+};
+
+export const ToastContextProvider = ({
+  children,
+}: Readonly<{ children: ReactNode }>) => {
+  const [state, dispatch] = useReducer(toastReducer, initialState);
+  const addToast = (type: ToastType, message: string) => {
+    const id = Math.floor(Math.random() * 10_000_000);
```
jessicamcinchak, line 33: nit: we also have the `uuid` package available on the frontend which would let you generate a unique uuid id like `const id = uuidV4()`(consistent with how we generate session ids) - happy for this to work either way though / chances of unintentionally colliding/non-unique IDs seems veryyy low here :slightly_smiling_face:

editor.planx.uk/src/reducers/toastReducer.ts
```diff
@@ -0,0 +1,24 @@
...
+      );
+      return {
+        ...state,
+        toasts: updatedToasts,
+      };
+    }
+    default:
+      // @ts-ignore
+      //   Typescript complains because action is of type 'never' here
+      throw new Error(`Unhandled action type: ${action.type}`);
```
jessicamcinchak, line 22: :+1:

jessicamcinchak (approved):
Nice refactor / very handy :raised_hands:

editor.planx.uk/src/contexts/ToastContext.tsx
jamdelion: Nice tip thanks! I'll include that in the next PR 👍

PR #3594, opened 2024-08-30
feat(map-and-label): Copy feature

DafyddLlyr:
## What does this PR do?
 - Moves the `CopyFeature` select to its own component
 - Wires up this behaviour via a method in the provider

## To Do
- [x] Populate value & name props on the `SelectInput`
- [x] A11y: Don't add/remove, just disable / apply visually hidden modifier

## Demo
https://github.com/user-attachments/assets/1ff9a34b-b369-4d64-b8a0-634724cd14fb

### DIFF
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/CopyFeature.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/CopyFeature.tsx
@@ -8,29 +8,39 @@ import InputLabel from "ui/public/InputLabel";
 import { useMapAndLabelContext } from "./Context";
 
 interface Props {
-  index: number;
+  destinationIndex: number;
   features: Feature[];
 }
 
-export const CopyFeature: React.FC<Props> = ({ index: i, features }) => {
-  const { schema } = useMapAndLabelContext();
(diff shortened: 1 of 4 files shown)


editor.planx.uk/src/@planx/components/MapAndLabel/Public/CopyFeature.tsx
```diff
@@ -0,0 +1,59 @@
...
+          required
+          title={"Copy from"}
+          labelId={`select-label-${destinationIndex}`}
+          value={""}
+          onChange={(e) => {
+            const label = e.target.value as string;
+            // Convert text label to zero-indexed integer
+            const sourceIndex = parseInt(label, 10) - 1;
+            copyFeature(sourceIndex, destinationIndex);
+          }}
```
DafyddLlyr, line 43: This feels a little more complex than it needs to, but feels like the most readable version of what I'm aiming for.

If we use indices, it falls apart as we're "skipping" the current feature in the select. The alternatives are either something horrible like `value={ j > i ? j + 1 : i}` in the select to dynamically "skip", or keeping the current item in the select and making it disabled (not great for a11y).

jessicamcinchak (approved):
Thank you!!

PR #3595, opened 2024-08-30
feat(map-and-label): Remove features

DafyddLlyr:
## What does this PR do?
- Handles the "remove" functionality for the Map and Label component (mostly!)
  - Remove feature from formik form state
  - Remove feature from "features" array
  - **Does not** remove features from the map display
- Abstracts map interaction away into a reusable `useGeoJSONChange()` hook (see https://github.com/theopensystemslab/planx-new/pull/3595#discussion_r1738073599)

This PR is now ready to handover, for the outstanding task of removing the features from the map to be tackled 👍 

## Demo
https://github.com/user-attachments/assets/15a0f150-cf10-4786-b04e-c3dab8e8b45e

### DIFF
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/Context.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/Context.tsx
@@ -24,6 +24,7 @@ interface MapAndLabelContextValue {
   isFeatureInvalid: (index: number) => boolean;
   addFeature: () => void;
   copyFeature: (sourceIndex: number, destinationIndex: number) => void;
+  removeFeature: (index: number) => void;
   mapAndLabelProps: PresentationalProps;
   errors: {
     min: boolean;
@@ -106,6 +107,17 @@ export const MapAndLabelProvider: React.FC<MapAndLabelProviderProps> = (
     const sourceFeature = formik.values.schemaData[sourceIndex];
     formik.setFieldValue(`schemaData[${destinationIndex}]`, sourceFeature);
   };
+  
+  const removeFeature = (index: number) => {
+    resetErrors();
+
+    setActiveIndex(-1);
+
+    formik.setFieldValue(
+      "schemaData",
+      formik.values.schemaData.filter((_, i) => i !== index),
+    );
+  };
 
   return (
     <MapAndLabelContext.Provider
@@ -118,6 +130,7 @@ export const MapAndLabelProvider: React.FC<MapAndLabelProviderProps> = (
         validateAndSubmitForm,
         addFeature,
         copyFeature,
+        removeFeature,
         isFeatureInvalid,
         errors: {
           min: minError,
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/lib/gis.ts
```diff
@@ -0,0 +1,41 @@
...
+
+type UseGeoJSONChange = (
+  callback: (event: GeoJSONChangeEvent) => void,
+) => [Feature[] | undefined, Dispatch<SetStateAction<Feature[] | undefined>>];
+
+/**
+ * Hook for interacting with @opensystemslab/map
+ * Assign a callback function to be triggered on the "geojsonChange" event
+ */
+export const useGeoJSONChange: UseGeoJSONChange = (callback) => {
```
DafyddLlyr, line 21: I initially move this logic out of the components, and into the provider. 

I then realised it's actually possibly fairly generic / reusable and not MapAndLabel specific at all.

Not totally sold on location within the codebase - maybe we need a helpers folder for the map component or something?

editor.planx.uk/src/@planx/components/MapAndLabel/Public/Context.tsx
```diff
@@ -17,12 +19,13 @@ import { PresentationalProps } from ".";
...
 interface MapAndLabelContextValue {
   schema: Schema;
+  features?: Feature[];
   activeIndex: number;
   editFeature: (index: number) => void;
   formik: FormikProps<SchemaUserData>;
   validateAndSubmitForm: () => void;
   isFeatureInvalid: (index: number) => boolean;
-  addFeature: () => void;
+  removeFeature: (index: number) => void;
```
DafyddLlyr, line 28: I did consider making this something like `removeActiveFeature()` which would not require an argument and just use `activeIndex` internally.

I think passing an argument is a little more flexible, and would allow us to call this from other places if the remove button moved from the active tab (e.g. to the tab list).

editor.planx.uk/src/lib/gis.ts
jessicamcinchak: This is a nice idea! 

One nit on first read through: I think we need the `id` on L32 below to be a param as this vary throughout our various implementations (map&label, drawboundary, etc)!

DafyddLlyr: Ah yep, very good catch again thank you! Fixed in [f5bb2f4](https://github.com/theopensystemslab/planx-new/pull/3595/commits/f5bb2f4e84670f032f2472ff2d765994710038ee)

jessicamcinchak:
Thanks for groundwork here ! Happy to take this one over next week & do final map display step

Will review & merge with fresh eyes then :slightly_smiling_face:

DafyddLlyr:
Perfect - thanks @jessicamcinchak 🙌

jessicamcinchak:
Closing in favor of #3625

PR #3596, opened 2024-08-30
fix(map-and-label): Add spacing between schema fields

DafyddLlyr:
## What does this PR do?
 - Adds spacing between schema fields on Map and Label component

### Context
`SchemaFields` is intentionally unstyled - this should be left to the wrapping component (thoughts welcome!). By adding the MUI `sx` prop we allow a degree of flexibility here - unstyled out of the box, but a familiar API to actually configure this when needed.

| Before | After |
|--------|--------|
|image|image|

### DIFF
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
@@ -13,11 +13,11 @@ import { SchemaFields } from "@planx/components/shared/Schema/SchemaFields";
 import { Feature, GeoJsonObject } from "geojson";
 import { useStore } from "pages/FlowEditor/lib/store";
 import React, { useEffect, useState } from "react";
+import { FONT_WEIGHT_SEMI_BOLD } from "theme";
 import SelectInput from "ui/editor/SelectInput";
 import FullWidthWrapper from "ui/public/FullWidthWrapper";
 import InputLabel from "ui/public/InputLabel";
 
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
```diff
@@ -108,30 +116,49 @@ const VerticalFeatureTabs: React.FC<{ features: Feature[] }> = ({
...
                   </SelectInput>
                 </InputLabel>
               </Box>
             </Box>
             <SchemaFields
+              sx={(theme) => ({
+                display: "flex",
+                flexDirection: "column",
+                gap: theme.spacing(2),
+              })}
```
DafyddLlyr, line 146: This is the only change in this file, the rest is just linting (need to resolve this sooner rather than later!)

editor.planx.uk/src/@planx/components/shared/Schema/SchemaFields.tsx
```diff
@@ -15,6 +17,7 @@ interface SchemaFieldsProps {
   /** Formik instance generated from config provided by useSchema hook */
   formik: FormikProps<SchemaUserData>;
   schema: Schema;
+  sx?: SxProps<Theme>;
```
jessicamcinchak, line 20: super simple flexible solution :100:

jessicamcinchak (approved):

PR #3601, opened 2024-08-30
feat[map-and-label]: repopulate list map when clicking back

RODO94:
## What does this PR do?

This PR looks to ensure the `<my-map>` component is given a prop, `drawGeojsonData`, which will keep the feature on the map when a user navigates back to it on a Public form.


jessicamcinchak:
This required a change to `@opensystemslab/map` first (sorry for missing that bug / not catching this sequencing error!) - now addressed in https://github.com/theopensystemslab/planx-new/pull/3612

RODO94:
@jessicamcinchak is this good to close with your new PR up? 

https://github.com/theopensystemslab/planx-new/pull/3612

PR #3602, opened 2024-08-30
feat(map-and-label): Better tab error styling

DafyddLlyr:
Please see https://github.com/theopensystemslab/planx-new/pull/3589#discussion_r1738010699 for context

image

### DIFF
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
@@ -76,7 +76,9 @@ const VerticalFeatureTabs: React.FC<{ features: Feature[] }> = ({
               {...a11yProps(i)}
               {...(isFeatureInvalid(i) && {
                 sx: (theme) => ({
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
```diff
@@ -76,7 +76,9 @@ const VerticalFeatureTabs: React.FC<{ features: Feature[] }> = ({
               {...a11yProps(i)}
               {...(isFeatureInvalid(i) && {
                 sx: (theme) => ({
-                  backgroundColor: theme.palette.action.focus,
+                  borderRight: `12px solid ${theme.palette.error.main}`,
+                  // Appear over tab indicator
+                  zIndex: 2,
```
DafyddLlyr, line 81: This is a little hacky as the tabs are buttons which are slightly hidden / obscured by the TabPanel itself (see below).

image

@ianjon3s Would love another pair of eyes on this if you get a chance!

jessicamcinchak (approved):

PR #3606, opened 2024-08-31
Revert "feat: API changes for handling new `submission_email` location"

DafyddLlyr:
Reverts theopensystemslab/planx-new#3582 due to failing regression tests

✅ Regression test passing on this branch here - https://github.com/theopensystemslab/planx-new/actions/runs/10643974774

I suspect that this is due to the field `TeamSettings.submissionEmail` still being valid [here](https://github.com/theopensystemslab/planx-core/blob/main/src/types/team.ts#L32) - as it's not removed, there may be a reference to it somewhere that wasn't caught in theopensystemslab/planx-new#3582?

### DIFF
--- a/api.planx.uk/modules/send/downloadApplicationFiles/index.ts
+++ b/api.planx.uk/modules/send/downloadApplicationFiles/index.ts
@@ -17,10 +17,10 @@ export async function downloadApplicationFiles(
 
   try {
     // Confirm that the provided email matches the stored team settings for the provided localAuthority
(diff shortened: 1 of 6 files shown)


RODO94 (approved):
Ran this locally and all tests are passing for me.

RODO94:
I worked on an additive fix for this regression failure, but I couldn't come up with one today. 

We will merge this to allow other work to move along and look to add a bug fix later

PR #3608, opened 2024-09-02
fix: handle no existing team members case when adding a new editor

jamdelion:
Fixes case where the 'add' button wasn't visible for teams where there were no team editors already.

| Before                      | After                  |
| ------------------------ | ------------------------ |
|  |  |

Trello ticket: https://trello.com/c/piMcGsJa/2973-add-user-in-editor

Fix for comment raised in Slack: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1724945320946359?thread_ts=1724942587.700829&cid=C5Q59R3HB

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
@@ -56,15 +56,64 @@ export const MembersTable = ({
 
   if (members.length === 0) {
     return (
-      <Table>
-        <TableHead>
-          <TableRow>
-            <TableCell>
-              <strong>No members found</strong>
-            </TableCell>
-          </TableRow>
-        </TableHead>
-      </Table>
+      <>
+        <Table>
+          <TableHead>
+            <TableRow>
+              <TableCell>
+                <strong>No members found</strong>
+              </TableCell>
+            </TableRow>
+          </TableHead>
+          {showAddMemberButton && (
+            <TableRow>
+              <TableCell colSpan={3}>
+                <AddButton onClick={() => setShowModal(true)}>
+                  Add a new editor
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
```diff
@@ -56,15 +56,64 @@ export const MembersTable = ({
...
+            onClose={handleCloseErrorToast}
+          >
+            <Alert
+              onClose={handleCloseErrorToast}
+              severity="error"
+              sx={{ width: "100%" }}
+            >
+              Failed to add new user, please try again
+            </Alert>
+          </Snackbar>
```
jamdelion, line 106: This is unwieldy, should be removed in upcoming `useToast` PR

jessicamcinchak (approved):
Code changes look great here - thanks for fix :+1: 

Looks like this pizza has an API & ShareDB issue (just flakey & unrelated to your changes I suspect!). Since this is such a straightforward & tested change, happy for this to simply merge to staging and double check there once deployed, or would suggest logging into Vultr, destroying the instance & DNS, then re-running CI from scratch and I'd guess new pizza will be better?

jamdelion:
> Code changes look great here - thanks for fix 👍
> 
> Looks like this pizza has an API & ShareDB issue (just flakey & unrelated to your changes I suspect!). Since this is such a straightforward & tested change, happy for this to simply merge to staging and double check there once deployed, or would suggest logging into Vultr, destroying the instance & DNS, then re-running CI from scratch and I'd guess new pizza will be better?

Thanks @jessicamcinchak - I've destroyed the instance (I assume DNS goes along with it?) but now getting an earlier fail in CI on the upsert vultr instance job. Have I missed a step?

PR #3615, opened 2024-09-03
fix: List component input spacing

ianjon3s:
## What does this PR do?

Adds padding to list component inputs so that labels are correctly visually associated with corresponding inputs.

**Before:**
image

**After:**
image

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/Schema/SchemaFields.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Schema/SchemaFields.tsx
@@ -1,5 +1,6 @@
 import Box from "@mui/material/Box";
 import { SxProps, Theme } from "@mui/material/styles";
+import { styled } from "@mui/material/styles";
 import { FormikProps } from "formik";
 import React from "react";
 import InputRow from "ui/shared/InputRow";
@@ -20,6 +21,13 @@ interface SchemaFieldsProps {
   sx?: SxProps<Theme>;
 }
 
+const ListRows = styled(Box)(({ theme }) => ({
+  padding: theme.spacing(1, 0),
+  display: "flex",
+  flexDirection: "column",
+  gap: theme.spacing(1.5),
+}));
+
 /**
  * Display a set of fields for the provided schema
  */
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/@planx/components/shared/Schema/SchemaFields.tsx
```diff
@@ -20,6 +21,13 @@ interface SchemaFieldsProps {
   sx?: SxProps<Theme>;
 }
 
+const ListRows = styled(Box)(({ theme }) => ({
+  padding: theme.spacing(1, 0),
+  display: "flex",
+  flexDirection: "column",
+  gap: theme.spacing(1.5),
+}));
```
jessicamcinchak, line 29: Can you please take a look at this PR that merged last week - https://github.com/theopensystemslab/planx-new/pull/3596

Since these fields are now used in two separate places (List active cards & MapAndLabel tabs) I think we want to set this as an `sx` prop on List-specific instane of this componponent so MapAndLabel can maintain it's own unique spacing too !

ianjon3s: Ideally we want the spacing to be the same across the two, as we're essentially showing [small variations of] the same form inputs in different wrappers. To me it makes more sense to control this at the wrapper level rather than have to define twice.

jessicamcinchak: I personally anticipate that we're going to be getting feedback from two direction soon even though these are similar from our perspective and the flexibility is going to come in handy. 

But trust your expertise on this - if you'd like to just style once here via the `ListRows` styled component then that's fine, but please then revert/remove `sx` prop recently added to generic `Box` otherwise MapAndLabel implementation currently confusing has "double" styles applied here and hard to unpick which overrides other?

Happy to approve after that !

ianjon3s: Taken on board, for simplicity I've reintroduced the `sx` to `SchemaFields.tsx` and included specific styling in the component to fix the List Component spacing. If we find the styles are replicated after feedback I'll just in and make the styles global.

jessicamcinchak (approved):

PR #3618, opened 2024-09-03
fix: initial works to trees feedback

jessicamcinchak:
Working through feedback here: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1725376092471139

- [x] List option schema should order the "map" field first, rather than last (not removing original schema yet)
- [x] "DateInput" field no longer throws a type error when tabbing through
- [x] "Cancel" button should be hidden on initial List item


ianjon3s (approved):
All fixes working as expected 🙌

PR #3620, opened 2024-09-04
fix: display schema field descriptions if set

jessicamcinchak:
Fixes a regression flagged here: https://opensystemslab.slack.com/archives/C01E3AC0C03/p1725446829984269

**Locally edited schema to include descriptions:** 
Screenshot from 2024-09-04 13-10-43

**And how descriptions display with validation errors:**
Screenshot from 2024-09-04 13-11-14

### DIFF
--- a/editor.planx.uk/src/@planx/components/List/Public/index.test.tsx
+++ b/editor.planx.uk/src/@planx/components/List/Public/index.test.tsx
@@ -39,6 +39,12 @@ describe("Basic UI", () => {
     expect(textInput).toBeInTheDocument();
     expect(textInput.type).toBe("text");
 
+    // Input field descriptions are generated from schema if they exist...
+    const textInputDescription = getByText(
+      /Please make it cute/,
+    ) as HTMLElement;
+    expect(textInputDescription).toBeInTheDocument();
+
     // Props are correctly read
     const emailInput = getByLabelText(
       /What's their email address?/,
--- a/editor.planx.uk/src/@planx/components/shared/Schema/InputFields/shared.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Schema/InputFields/shared.tsx
@@ -1,0 +1,19 @@
+import Typography from "@mui/material/Typography";
+import React from "react";
+import ReactMarkdownOrHtml from "ui/shared/ReactMarkdownOrHtml";
+
+interface FieldInputDescriptionProps {
+  description: string;
+}
+
+export const FieldInputDescription: React.FC<FieldInputDescriptionProps> = ({
+  description,
+}) => (
+  <Typography variant="body2" mb={1.5}>
+    <ReactMarkdownOrHtml
+      source={description}
+      openLinksOnNewTab
+      // TODO a11y props? eg `id={DESCRIPTION_TEXT}` but unique ??
+    />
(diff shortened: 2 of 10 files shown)


augustlindemer:
Looking good on Storybook pizza.

editor.planx.uk/src/@planx/components/shared/Schema/InputFields/shared.tsx
```diff
@@ -0,0 +1,14 @@
...
+  description: string;
+}
+
+export const FieldInputDescription: React.FC<FieldInputDescriptionProps> = ({
+  description,
+}) => (
+  <Typography variant="body2" mb={1.5}>
+    {description}
+  </Typography>
+);
```
jessicamcinchak, line 14: **Two notes about this -** 
- When schema generation moves to the editor in the future, `description` will likely be rich text and we'll want to update this, but we don't need to account for this just yet 
- I'm also not 100% sure what a11y markup we need here - this is text that sits between the input label & input which are properly associated to each other, but the description itself is not currently associated to either of them
  - Maybe the input needs `aria-describedby` referencing a unique `id` set on this description?? 

I think the future probably looks more like this, but unless anyone knows for certain now let's revisit when we do a broader accessibility sweep of these new components soon!
```tsx
<Typography>
   <ReactMarkdownOrHtml
      source={description}
      openLinksOnNewTab
      // TODO a11y props? eg `id={DESCRIPTION_TEXT}` like CardHeader but unique because many fields on same page ??
    />
</Typography>
```

editor.planx.uk/src/@planx/components/List/Public/index.test.tsx
```diff
@@ -39,6 +39,12 @@ describe("Basic UI", () => {
     expect(textInput).toBeInTheDocument();
     expect(textInput.type).toBe("text");
 
+    // Input field descriptions are generated from schema if they exist...
+    const textInputDescription = getByText(
+      /Please make it cute/,
+    ) as HTMLElement;
+    expect(textInputDescription).toBeInTheDocument();
```
jessicamcinchak, line 46: Descriptions now checked in basic setup test so we proactively catch this next time!

editor.planx.uk/src/@planx/components/shared/Schema/InputFields/shared.tsx
RODO94: For the note on `aria-describedby`, just for my knowledge. You use that to say this thing is described by element with `id = "..." ` so for this the description will have an `id` and the input would have the `aria-describedby`?

RODO94 (approved):
Made a change locally for testing and it works as expected.

The screenshots look right and the code looks good to me. Thanks for updating the testing as well!

editor.planx.uk/src/@planx/components/shared/Schema/InputFields/shared.tsx
jessicamcinchak: Yes - I often reference MDN docs for clarification about ARIA properties, there's a lot of them & the differences are subtle! https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby

PR #3621, opened 2024-09-04
feat[mal]: Map and Label Error and Validation Tests

RODO94:
## What does this PR do? 

This PR follows work completed in: #3604 

I am writing tests for the **Error and Validation** `describe()` as part of larger work to write tests for all of the `test.todos()`.

My plan is to push up a PR for each `describe()` section so I can deal with feedback as I go and hopefully improve / fix any structure or logic issues.

For error checking at the Tab, I went with checking for the `border-left` styling as I couldn't see anything more specific to work from. From here I checked that this is removed when a new feature is added but is reapplied when you click "Continue" without adding new inputs.

The original `test.todos` there was one for Maximum number as an inverse to the minimum `it()` block I have written, but I didn't think there was a maximum for this component so I removed it.

### DIFF
--- a/editor.planx.uk/src/@planx/components/List/schemas/Trees.ts
+++ b/editor.planx.uk/src/@planx/components/List/schemas/Trees.ts
@@ -16,6 +16,7 @@ export const Trees: Schema = {
       type: "text",
       data: {
         title: "Proposed work",
+        description: "This is the wee descy",
         fn: "work",
         type: TextInputType.Short,
       },
--- a/editor.planx.uk/src/@planx/components/List/schemas/Trees.ts
+++ b/editor.planx.uk/src/@planx/components/List/schemas/Trees.ts
@@ -16,7 +16,6 @@ export const Trees: Schema = {
       type: "text",
       data: {
         title: "Proposed work",
-        description: "This is the wee descy",
         fn: "work",
         type: TextInputType.Short,
       },
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
@@ -46,14 +46,14 @@ describe("Basic UI", () => {
 
     await waitFor(() =>
       expect(
-        queryByText("Plot a feature on the map to begin")
-      ).not.toBeInTheDocument()
+        queryByText("Plot a feature on the map to begin"),
+      ).not.toBeInTheDocument(),
     );
   });
 
   it("renders the schema name as the tab title", async () => {
     const { queryByText, getByRole, getByTestId } = setup(
-      <MapAndLabel {...props} />
+      <MapAndLabel {...props} />,
     );
     expect(queryByText(/Tree 1/)).not.toBeInTheDocument();
 
@@ -68,7 +68,7 @@ describe("Basic UI", () => {
 
   it("should not have any accessibility violations", async () => {
     const { queryByText, getByTestId, container } = setup(
-      <MapAndLabel {...props} />
+      <MapAndLabel {...props} />,
     );
     expect(queryByText(/Tree 1/)).not.toBeInTheDocument();
 
@@ -86,7 +86,7 @@ describe("Basic UI", () => {
 describe("validation and error handling", () => {
   it("shows all fields are required", async () => {
     const { getAllByTestId, getByTestId, getByRole, user } = setup(
-      <MapAndLabel {...props} />
+      <MapAndLabel {...props} />,
     );
     const map = getByTestId("map-and-label-map");
     expect(map).toBeInTheDocument();
@@ -96,21 +96,57 @@ describe("validation and error handling", () => {
     expect(getByRole("tab", { name: /Tree 1/ })).toBeInTheDocument();
 
     const continueButton = getByRole("button", { name: /Continue/ });
-
-    user.click(continueButton);
+    expect(continueButton).toBeInTheDocument();
+    await user.click(continueButton);
 
     const errorMessages = getAllByTestId(/error-message-input/);
 
     // 5 error message inputs + 3 for date inputs (one per input and one wrapper)
-    expect(errorMessages).toHaveLength(8);
+    expect(errorMessages).toHaveLength(4);
 
     errorMessages.forEach((message) => {
-      expect(message).toBeEmptyDOMElement();
+      expect(message).not.toBeEmptyDOMElement();
     });
   });
+
+  // <button
+  //   class="MuiButtonBase-root MuiTab-root MuiTab-textColorPrimary css-1mb4v11-MuiButtonBase-root-MuiTab-root"
+  //   tabindex="-1"
+  //   type="button"
+  //   role="tab"
+  //   aria-selected="false"
+  //   id="vertical-tab-1"
+  //   aria-controls="vertical-tabpanel-1"
+  // >
+  //   Tree 2
+  // </button>;
+
+  // <button
+  //   class="MuiButtonBase-root MuiTab-root MuiTab-textColorPrimary Mui-selected css-1mb4v11-MuiButtonBase-root-MuiTab-root"
+  //   tabindex="0"
+  //   type="button"
+  //   role="tab"
+  //   aria-selected="true"
+  //   id="vertical-tab-0"
+  //   aria-controls="vertical-tabpanel-0"
+  // >
+  //   Tree 1
+  // </button>;
+
+  // <div
+  //   aria-labelledby="vertical-tab-1"
+  //   class="MuiTabPanel-root css-1iolky5-MuiTabPanel-root"
+  //   hidden=""
+  //   id="vertical-tabpanel-1"
+  //   role="tabpanel"
+  //   data-testid="vertical-tabpanel-1"
+  // ></div>;
+
   // it shows all fields are required in a tab
   it("should show all fields are required, for all feature tabs", async () => {
-    const { getByTestId, getByRole, user } = setup(<MapAndLabel {...props} />);
+    const { getByTestId, getByRole, user, debug } = setup(
+      <MapAndLabel {...props} />,
+    );
     const map = getByTestId("map-and-label-map");
     expect(map).toBeInTheDocument();
 
@@ -130,6 +166,7 @@ describe("validation and error handling", () => {
     const secondTabPanel = getByTestId("vertical-tabpanel-1");
 
     expect(firstTabPanel.childElementCount).toBeGreaterThan(0);
+    expect(secondTabPanel.childElementCount).toBe(0);
 
     // default is to start on first tab panel, second is hidden
     expect(firstTabPanel).toBeVisible();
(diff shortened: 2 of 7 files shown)


editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -84,44 +84,150 @@ describe("Basic UI", () => {
...
+    const continueButton = getByRole("button", { name: /Continue/ });
+    // await user.click(continueButton);
+
+    // error messages appear
+    const errorMessagesOne =
+      within(firstTabPanel).getAllByTestId(/error-message-input/);
+    expect(errorMessagesOne).toHaveLength(8);
+
+    // click to go to the second tab
+    await user.click(secondTab);
```
RODO94, line 148: @DafyddLlyr failing from this point

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -84,44 +96,156 @@ describe("Basic UI", () => {
...
+  // it shows all fields are required in a tab
+  it("should show all fields are required, for all feature tabs", async () => {
+    const { getByTestId, getByRole, user, debug } = setup(
+      <MapAndLabel {...props} />,
+    );
+    const map = getByTestId("map-and-label-map");
+    expect(map).toBeInTheDocument();
+    debug();
+
+    addMultipleFeatures([point1, point2]);
```
RODO94, line 133: new function for adding multiple features added here

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -84,44 +96,194 @@ describe("Basic UI", () => {
...
+
+    expect(tabOne).toHaveStyle("border-left: 5px solid #D4351C");
+
+    addFeaturesToMap(map, [point1, point2]);
+
+    const tabTwo = getByRole("tab", { name: /Tree 2/ });
+
+    expect(tabTwo).toBeInTheDocument();
+
+    expect(tabOne).not.toHaveStyle("border-left: 5px solid #D4351C");
```
RODO94, line 231: Testing functionality that when `tabOne` is in an error state, adding a new tree will remove this error state and reset things.

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -84,44 +96,194 @@ describe("Basic UI", () => {
...
+
+    await user.click(continueButton);
+
+    expect(tabOne).toHaveStyle("border-left: 5px solid #D4351C");
+    expect(tabTwo).toHaveStyle("border-left: 5px solid #D4351C");
```
RODO94, line 236: Then from here, I am retesting the error state being applied again.

@jessicamcinchak just wanted to check this is the intended functionality for now or is it likely to change soon?

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
```diff
@@ -128,6 +128,7 @@ const VerticalFeatureTabs: React.FC = () => {
             sx={{ width: "100%" }}
             aria-labelledby={`vertical-tab-${i}`}
             id={`vertical-tabpanel-${i}`}
+            data-testid={`vertical-tabpanel-${i}`}
```
RODO94, line 131: This felt more explicit for any future changes to this component as it is directly referencing it's use for testing. If this is OTT lemme know and I can switch to the `id` or `aria-labelledby`, but tried to account for the component still being worked on.

editor.planx.uk/src/@planx/components/List/schemas/Trees.ts
```diff
@@ -16,6 +16,7 @@ export const Trees: Schema = {
       type: "text",
       data: {
         title: "Proposed work",
+        description: "This is the wee descy",
```
jessicamcinchak, line 19: This schema is in use in the "real" flows that we're user-testing - we don't want to add this fake `description` directly here? Please see `List` component schemas for examples of how we keep separate mock schemas for testing purposes.

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -84,44 +84,194 @@ describe("Basic UI", () => {
...
+
+    // side tab validation
+    expect(firstTab).toBeInTheDocument();
+    expect(secondTab).toBeInTheDocument();
+
+    // form for each tab
+    const firstTabPanel = getByTestId("vertical-tabpanel-0");
+    const secondTabPanel = getByTestId("vertical-tabpanel-1");
+
+    // default is to start on seond tab panel, second is hidden
```
jessicamcinchak, line 135: nit: please clarify this comment - should be "first is hidden"

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -84,44 +84,194 @@ describe("Basic UI", () => {
...
+      within(secondTabPanel).getAllByTestId(/error-message-input/);
+    expect(errorMessagesTabTwo).toHaveLength(4);
+
+    // error messages are empty but visible before error state induced
+    // this ensures they contain the error message text
+    errorMessagesTabTwo.forEach((input) => {
+      expect(input).not.toBeEmptyDOMElement();
+    });
+
+    // click to go to the second tab
```
jessicamcinchak, line 157: nit: again outdated comment does not match code

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -84,44 +84,194 @@ describe("Basic UI", () => {
...
+
+    const continueButton = getByRole("button", { name: /Continue/ });
+
+    await user.click(continueButton);
+
+    const errorWrapper = getByTestId(/error-wrapper/);
+
+    const errorMessage = within(errorWrapper).getByText(/You must plot /);
+
+    expect(errorMessage).not.toBeEmptyDOMElement();
```
jessicamcinchak, line 182: nit: I think `toBeInTheDocument()` would be a stronger check _for visibility_ here? Per a11y requirements, we have many error wrappers that are _always_ in the DOM, and their _visibility_ is toggled depending on component state.

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
jessicamcinchak: I don't think this is actually partcularly meaningful / intentional behavior - the tabs are based off the number of features on the map, and when the length changes we currently _re-render all tabs_ (hence error state for tab 1 "disappears" when adding tab 2 because it's been entirely re-rendered) - but I don't actually think that's performant or ideal behavior. 

Personally, I think testing error style of tab 1 only is sufficient here.

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
jessicamcinchak: See above comment - I think we just need to test error style of a single tab to satisfy this test.

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -84,44 +84,194 @@ describe("Basic UI", () => {
...
+    expect(getByRole("tab", { name: /Tree 1/ })).toBeInTheDocument();
+
+    const continueButton = getByRole("button", { name: /Continue/ });
+    expect(continueButton).toBeInTheDocument();
+    await user.click(continueButton);
+
+    const errorMessages = getAllByTestId(/error-message-input/);
+
+    // 5 error message inputs + 3 for date inputs (one per input and one wrapper)
+    expect(errorMessages).toHaveLength(4);
```
jessicamcinchak, line 105: nit: comment out of date with code / mock without date input here

editor.planx.uk/src/@planx/components/MapAndLabel/test/mocks/Trees.ts
```diff
@@ -1,12 +1,19 @@
+import { Schema } from "@planx/components/shared/Schema/model";
+
 import { PresentationalProps } from "../../Public";
 import { Trees } from "../../schemas/Trees";
 
+const mockTreeSchema: Schema = {
+  ...Trees,
+  fields: Trees.fields.filter((field) => field.type !== "date"),
+};
```
jessicamcinchak, line 9: Per above comment, this would be better place to set `description` for testing-only please !

jessicamcinchak (commented):
Looks like a great start on more test coverage - a few comments to address/clean up before merging please!

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -84,44 +84,194 @@ describe("Basic UI", () => {
 
 // Schema and field validation is handled in both List and Schema folders - here we're only testing the MapAndLabel specific error handling
 describe("validation and error handling", () => {
-  test.todo("all fields are required");
-  test.todo("all fields are required, for all feature tabs");
-  test.todo("an error displays if the minimum number of items is not met");
-  test.todo("an error displays if the maximum number of items is exceeded");
```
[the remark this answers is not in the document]
jessicamcinchak: We should keep this `todo` please - all our schemas support a `max`, we just aren't enforcing yet or using in the initial "Trees" schema

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
jessicamcinchak: If this suggested change worked as expected here - can you also please update all other similar instances of `DOMElement` checks as well - otherwise I think these are quite flakey tests _not_ actually testing _visibility_ of error messages like we intend !!

RODO94: @jessicamcinchak for this, would it be the case that for the Input fields in the `<TabPanels>` that the errors are always in the document but just don't have any text in them so `toBeInTheDocument()` would come back yes even if there's no error messages displaying? 

Would something like this be more specific?

```ts
    errorMessages.forEach((message) => {
      expect(message).toHaveTextContent(/(Enter|Select)/);
    });
```

RODO94: Example here of error element before "Continue" is clicked:

```html

<p class="..." id="error-message-input-text-species" data-testid="error-message-input-text-species"></p>
```

jessicamcinchak (approved):

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
jessicamcinchak: Looking through other error handling tests in other components now and realising checking DOM element is used elsewhere too, let's leave this as-is right now. Sorry for confusion. 

There's still many outstanding to-do's here that are more valuable to move onto than get hung up on this.

PR #3624, opened 2024-09-04
feat: add parking space schemas

augustlindemer:
Add parking space list schemas for GLA and non-GLA councils.


jessicamcinchak (approved):
Looks good ! Thanks again for writing these :+1:

PR #3612, opened 2024-09-02
feat: enable "back" & "edit" navigation for List map fields and MapAndLabel using `@opensystemslab/map` v1.0.0-alpha.2

jessicamcinchak:
**A few key changes in here!** 
- Bumps `@opensystemslab/map` to v1.0.0-alpha.2
  - Fixes shifty labels when modifying features; now when you modify it should keep the same label
  - Enables processing a GeoJSON "FeatureCollection" passed into `drawGeojsonData` prop which is required for "back" navigation when `drawMany` is enabled (previously supported single "Feature")
- **[List]** Pre-populates drawing layer on the map when you click "Edit" for an existing item 
  - Replaces #3601
- **[List]** Fix bug preventing existing/pre-populated features from being successfully "cleared"
- **[MapAndLabel]** Updates `handleSubmit` to produce GeoJSON data
- **[MapAndLabel]** Pre-populates map drawing layer and tab fields when you come "back" or "change" from the Review page
- **[MapAndLabel]** Adds a very basic Review page summary
- **[MapAndLabel]** Fixes bug where no tabs were active when coming "back"/"changing" - always default to tab 0 as active

**Testing:** 
- https://3612.planx.pizza/testing/works-to-trees-prototype (toggle on `"TREES"` feature flag first!) 

**Known bugs:**
- **[MapAndLabel]** "Remove" button is still not hooked up, but introduction of `drawGeojsonData` here is precursor to being able to complete https://github.com/theopensystemslab/planx-new/pull/3595

PR #3625, opened 2024-09-04
feat(map-and-label): remove individual features

jessicamcinchak:
Rebased version of #3595

**My changes (builds on 3595 description):**
- [x] Remove feature from map
- [x] Shift labels on "remove" so they stay incremental
- [x] Ensure previously working `handleSubmit` & "back"/"change" navigation still works 
  - This is honestly still a bit wonky but I think less vital for UR than that actual "remove" functionality !!
  - See #3612 for original implementation, which isn't ideal in combo with new `GeoJSONChange` hook here

**Open questions:** 
- When removing a feature, I'm forcing a re-render of the web component by changing a `key` prop on its' parent container which does cause a visible flash - is this acceptable? Other less hacky feeling ideas?
  - Tried many other variations but was having a hard time getting the web component to re-render based on changing `features` alone - I _think_ this is most likely due to Lit's "shallow change detection" 
  - Spent some time reading up on [Lit Events](https://lit.dev/docs/components/events/), but not totally clear to me if an event listener would be another way to solve this?


editor.planx.uk/src/@planx/components/MapAndLabel/Public/Context.tsx
```diff
@@ -119,7 +150,18 @@ export const MapAndLabelProvider: React.FC<MapAndLabelProviderProps> = (
   const isFeatureInvalid = (index: number) =>
     Boolean(get(formik.errors, ["schemaData", index]));
 
-  const addFeature = () => {
+  const addFeatureToMap = (geojson: GeoJSONChange) => {
+    resetErrors();
+    setFeatures(geojson["EPSG:3857"].features);
+    setActiveIndex((features && features?.length - 2) || activeIndex + 1);
```
jessicamcinchak, line 156: There's some ugly active tab index management going on! Very much need to revisit this / open to suggestions. 

In many instances, I want the active tab index to be the last in the list. Otherwise there can be some strange behavior, for example on staging:
- Add trees 1, 2, 3 on the map
- Navigate to Tab 1 (it's active)
- Add another tree (#4) to the map - but now tab 2 is active when we really want 4 to be active

DafyddLlyr: Yeah this is an issue in the List component which is avoided by only allowing a single active item at a time (it's not possible to add when there's an active item). That solution won't work here due to the map driven nature of the component. 

It's going to be a little tricky as `setFeatures()` is async, but we can take the length from `geojson["EPSG:3857"].features` reliably I believe.

Something like this might work?

```
const newFeatures = geojson["EPSG:3857"].features;
setFeatures(newFeatures);
setActiveIndex(newFeatures.length - 1)
```

I've not tried this so I might be off the mark a little here and treading over old ground you've already worked through.

An alternative might a `useEffect()` to catch changes to `features.length` and tie the update to `setActiveIndex` to that?

DafyddLlyr (approved):
Working as expected and code is clear and straightforward 👍  

A few small issues below which may already be on your radar from last week - 

**Sorting**
 10 is coming before 2 - 

image



**Map re-render**
Updating the key works, but I agree is a bit hacky.

I _think_ what we want to do is add a custom change detection handler to `drawGeojsonData` that checks the feature length instead of shallow object equality ([docs](https://lit.dev/docs/components/properties/#haschanged)). This looks like it should work? Not a blocker for UR as it's a relatively small UX issue and the map is often out of view when hitting the remove button.

**Active tab handling**
A few unexpected things here - it's not a simple one. It feels odd to remove feature 2 of 10, and then be jumped onto editing feature 10 (the last in the list). I'd expect to just "stay in place" on the list as I removed items. However, this might be a little odd as due the the label updates you'd still be on "Feature 2" (the previous feature 3).

**Editing**
When an item is edited via the map, it looks like the active index is not being set correctly. I guess I'd expect either nothing to happen with tabs or to make the edited item active. Right now it's removing the schema fields. I don't think this is introduced here but wanted to flag!


https://github.com/user-attachments/assets/f4ea6b57-d62a-4526-85e9-21133ca494c9

jessicamcinchak:
Thanks @DafyddLlyr for thorough review here - merging to staging for UR deadline, but will open follow PRs today to address these comments / fix-forward !

PR #3626, opened 2024-09-04
fix(map-and-label): when modifying features on the map, tabs should stay ordered by label not recently changed

jessicamcinchak:
Follows on from #3612 & https://github.com/theopensystemslab/map/pull/493 - we stabilised labels directly on the map features themselves, but not the tabs

Previously: add three points to the map, modify point #2, tabs 2 & 3 swap order

Now: map _and_ tab order are stable when modifying


jamdelion (approved):
Nice fix :)

PR #3630, opened 2024-09-05
test[e2e]: add more components to create-flow test

jamdelion:
### In this PR

✅ I think this is the end of the 'straightforward' components, at least from the adding to a flow side. 

- Add tasklist, next steps and review components
- Add find property, draw boundary, file upload and planning constraints components
- Refactor to use an enum instead of magic strings!

⚔️ There are other components still to cover, but are difficult or skipped here because:
- they have a required field which is hard to access from playwright (e.g. content, see question below)
- I don't understand the component and need to ask someone (e.g. filter, portals)
- I think they are covered by other tests (but still need to check) - e.g. sections, pay, send

### Still to do 📝 
- Implement the remaining components
- Write a test to cover the 'using this flow' side (i.e. member of public's point of view)

### ❓ Questions

- There are a lot of input boxes identified by `data-placeholder` rather than `placeholder` which makes it harder to grab them in tests. Is there a way round this/can they be changed from data-placeholder to a normal placeholder attribute?


RODO94:

e2e/tests/ui-driven/src/create-flow/create-flow.spec.ts
```diff
@@ -175,6 +182,33 @@ test.describe("Navigation", () => {
       "some data field",
     );
 
+    nextNode = page.locator(".hanger > a").nth(12);
```
DafyddLlyr, line 185: This looks like a good use case for a helper function and `last()` ([docs](https://playwright.dev/docs/api/class-locator#locator-last)).

If it's not as simple as using `last()`, we could also maybe chain a `filter()` to narrow it down a little.

```ts
const getNextNode = () => page.locator(".hanger > a").last()
let nextNode = getNextNode();
```

e2e/tests/ui-driven/src/helpers/addComponent.ts
```diff
@@ -1,9 +1,27 @@
 import { Locator, Page } from "@playwright/test";
 
-const createBaseInput = async (
+enum PlanXEditorComponent {
```
DafyddLlyr, line 3: I think it would be more maintainable to import the existing component type enum, and select by value in the select element here. There's already too many places where we need to update/sync this list so I'd be keen to avoid adding another where possible.

```ts
import { ComponentType } from "@opensystemslab/planx-core/types";

...

page.locator("select").selectOption({ value: type });
```

As a bonus, we could later make this an exhaustive switch which would force us to include a new test case here when adding a new component type.

e2e/tests/ui-driven/src/helpers/addComponent.ts
```diff
@@ -44,20 +62,68 @@ const createBaseInput = async (
...
 
   // convert type name to lowercase, with dashes if there are spaces
-  const buttonName = type.toLowerCase().replace(/\s/g, "-");
+  // or custom name if it doesn't fit the pattern
+  const buttonName =
+    type === PlanXEditorComponent.FIND_PROPERTY
+      ? "find-property-merged"
+      : type.toLowerCase().replace(/\s/g, "-");
+
   await page
```
DafyddLlyr, line 127: This seems like a bit of complexity we can remove - the button has `type="submit"` which we can access via a Playwright locator, or we can apply a `data-testid` to the button.

DafyddLlyr (commented):
This is looking great 🙌 

I'd recommend trying out [the Fixture pattern Playwright uses](https://playwright.dev/docs/test-fixtures) - this would like remove a lot of the boilerplate of passing around `page` and `locationNode` and allow us to better model our pages.

e2e/tests/ui-driven/src/create-flow/create-flow.spec.ts
jamdelion: Great suggestion, that works well! 👍

api.planx.uk/.husky/pre-commit
```diff
@@ -2,3 +2,5 @@
 . "$(dirname -- "$0")/_/husky.sh"
 cd api.planx.uk
 pnpm dlx lint-staged
+cd ../e2e
+pnpm dlx lint-staged
```
jamdelion, line 6: I've found that I've needed to run `pnpm run lint:fix` on every commit from the `e2e` directory so hopefully this fixes that

DafyddLlyr: I'm not sure this is the right solution - this seems to be running e2e linting each time a commit is made in the api directory. There should be a pre-commit file in the e2e directory which does this (clearly it's not working though!)

I think we need to actually install lint-staged in the root ([docs](https://github.com/lint-staged/lint-staged?tab=readme-ov-file#how-to-use-lint-staged-in-a-multi-package-monorepo)) and then we can have a per-project husky working as expected?

DafyddLlyr (approved):

PR #3631, opened 2024-09-05
fix(a11y): Fieldset and legend for list/map date input

ianjon3s:
## What does this PR do?

Addresses accessibility issue in the date input generated by schema, in that a label (`Expected completion date` in this case) was not correctly associated with a form field.

I've introduced the `<fieldset>` with child `<legend>` pattern to create this association.


jessicamcinchak (approved):
Thanks for catching this!

PR #3637, opened 2024-09-06
style: Layout for inactive list card with map

ianjon3s:
## What does this PR do?

Updates layout of the inactive list card so that cards with maps are arranged in a horizontal layout on desktop.

Also general style improvements to the list component as a whole (button contrast, table column widths).

**Preview:**
image

### DIFF
--- a/editor.planx.uk/src/@planx/components/List/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/List/Public/index.tsx
@@ -124,9 +124,7 @@ const InactiveListCard: React.FC<{
   const { schema, formik, removeItem, editItem, isPageComponent } =
     useListContext();
 
-  const mapPreview = schema.fields.find(
-    (field) => field.data.fn === "features",
-  );
+  const mapPreview = schema.fields.find((field) => field.type === "map");
 
   return (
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/@planx/components/List/Public/index.tsx
```diff
@@ -115,39 +124,58 @@ const InactiveListCard: React.FC<{
   const { schema, formik, removeItem, editItem, isPageComponent } =
     useListContext();
 
+  const mapPreview = schema.fields.find(
+    (field) => field.data.fn === "features",
+  );
```
jessicamcinchak, line 129: nit: we decided to make the `fn` user-configurable, so I think we'll want this to be based on `field.type === "map"` rather than relying on `fn = "features"`

jessicamcinchak (approved):
So tidy & compact :nail_care:

editor.planx.uk/src/@planx/components/List/Public/index.tsx
ianjon3s: Thanks for spotting, have updated 👍

PR #3639, opened 2024-09-06
chore: remove `TREES` feature flag

jessicamcinchak:
This will make it easier for Nomensa to test on staging next week!

Map & Label still appears with a "(Testing only)" caveat when selecting from components list


freemvmt (approved):
Makes sense to me :)

PR #3643, opened 2024-09-09
chore: `MapAndLabel` tidy ups

jessicamcinchak:
A few tidy ups & fixes now that all major functionality is merged & open for testing ! Follows on from #3625 

- [x] Tabs correctly sorted - previous bug where 10 was getting sorted before 2 because geojson `properties` require labels to be stored as strings, but we want to sort them as numbers 
- [ ] `activeTabIndex` more consistently managed based on length of `features` when adding & removing features
- [ ] When editing a feature, correctly set the right tab index
- [ ] After removing a feature, make zoom / map viewport repositioning less jumpy
- [ ] When coming "back"/"changing", ensure latest tab is set as active one by default

### DIFF
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
@@ -70,8 +70,12 @@ const VerticalFeatureTabs: React.FC = () => {
     throw new Error("Cannot render MapAndLabel tabs without features");
   }
 
-  // Features is inherently sorted by recently added/modified, order tabs by stable labels
-  const sortedFeatures = sortBy(features, ["properties.label"]);
+  // Features is inherently sorted by recently added/modified, order tabs by stable labels (labels are integers stored as strings)
+  const sortedFeatures = sortBy(features, [
+    function (f) {
+      return Number(f.properties?.label);
+    },
+  ]);
 
   return (
     <Box


editor.planx.uk/src/@planx/components/MapAndLabel/Public/Context.tsx
```diff
@@ -212,8 +213,9 @@ export const MapAndLabelProvider: React.FC<MapAndLabelProviderProps> = (
     resetErrors();
     removeFeatureFromForm(index);
     removeFeatureFromMap(index);
+
     // Set active index as highest tab after removal, so that when you "add" a new feature the tabs increment correctly
-    setActiveIndex((features && features.length - 2) || activeIndex - 1);
+    setActiveIndex((features && features.length - 2) || 0);
```
jessicamcinchak, line 218: If this is `- 1` rather than `- 2`, then there's a known bug:
- Add three features
- Remove feature 3 - active tab is tab 2
- Remove feature 2 - active tab is NOT tab 1 as expected

One to keep thinking on, still not completely satisfied with this solution !

editor.planx.uk/src/@planx/components/MapAndLabel/Public/Context.tsx
```diff
@@ -152,13 +152,14 @@ export const MapAndLabelProvider: React.FC<MapAndLabelProviderProps> = (
...
-    setActiveIndex((features && features?.length - 2) || activeIndex + 1);
+    const newFeatures = geojson["EPSG:3857"].features;
+    setFeatures(newFeatures);
+    setActiveIndex(newFeatures.length - 1);
   };
 
   const addInitialFeaturesToMap = (features: Feature[]) => {
     setFeatures(features);
-    // TODO: setActiveIndex ?
+    // setActiveIndex(features.length - 1);
```
jessicamcinchak, line 162: This _looks_ like the right solution to me, and it does indeed mean the latest tab will be active/open when coming "back", but it unexpectedly make all other tabs un-clickable - so noted in PR description as outstanding bug

DafyddLlyr (approved):

PR #3460, opened 2024-07-26
feat(wip): Navigation from search

DafyddLlyr:
⚠️ _Rough and incomplete (very much a work in progress!), hoping to tidy up a little before show and tell_

https://github.com/user-attachments/assets/f3a8622b-a7b7-4887-9b6e-0486b1bbd0e8

To do...
 - [ ] Tests!
 - [ ] Handle internal portals
 - [ ] Handle nodes with `_root` in path
 - [ ] Handle all node types
 - [ ] Tidy up code

PR #3646, opened 2024-09-10
feat: Data value search

DafyddLlyr:
## What does this PR do?
 - Ensures that searching via the "data" facet works for all component types
 - For each component which has a unique way of handling data variables, we need to ensure we can still access display details for these components. These components and their associated exceptions are encoded in the `DataDisplayMap` data structure
 - Removes temporary guard / filter on search result cards which previously hid these exceptional components

https://github.com/user-attachments/assets/db5aa7e9-ace9-4d2e-8fa0-3281b4955d8f

Next up...
- Navigate to nodes on click of card (Prior art - https://github.com/theopensystemslab/planx-new/pull/3460)
- For nodes within portals, visually wrap with a "portal wrapper" as per Figma

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/SearchResultCard/DataDisplayMap.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/SearchResultCard/DataDisplayMap.tsx
@@ -1,4 +1,7 @@
 import { ComponentType, IndexedNode } from "@opensystemslab/planx-core/types";
+import { Calculate } from "@planx/components/Calculate/model";
+import { FileUploadAndLabel } from "@planx/components/FileUploadAndLabel/model";
+import { List } from "@planx/components/List/model";
 import { SearchResult } from "hooks/useSearch";
 import { capitalize, get } from "lodash";
 import { SLUGS } from "pages/FlowEditor/data/types";
@@ -34,7 +37,7 @@ const DISPLAY_DATA: Partial<ComponentMap> = {
       displayKey: "Option (data)",
       getTitle: ({ item }) => {
         const parentNode = useStore.getState().flow[item.parentId];
-        return parentNode!.data.text!;
+        return parentNode.data.text;
       },
       getHeadline: ({ item, key }) => get(item, key)?.toString(),
     },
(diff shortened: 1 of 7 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/SearchResultCard/DataDisplayMap.tsx
```diff
@@ -0,0 +1,116 @@
...
+    },
+  },
+  // FileUploadAndLabel has data values nested in FileTypes
+  [ComponentType.FileUploadAndLabel]: {
+    default: {
+      displayKey: "File type (data)",
+      getHeadline: ({ item, refIndex }) =>
+        (item["data"] as unknown as FileUploadAndLabel)["fileTypes"][refIndex][
+          "fn"
+        ],
```
DafyddLlyr, line 52: There's a fair amount of type casting in this data structure as `IndexedNode` (and our flow as a whole) are not strongly typed. 

We can't use a type guard as the types of `IndexedNode["data"]` and various components have no overlap - we hit the `"A type predicate's type must be assignable to its parameter's type"` error.

This means that casting is unfortunately necessary, but it's better than using `any` or just optional chaining the whole way down.

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/SearchResultCard/index.tsx
```diff
@@ -22,42 +20,13 @@ export const Root = styled(ListItemButton)(({ theme }) => ({
...
-    getDisplayDetailsForResult(result); // TODO - display portal wrapper
+  // TODO - display portal wrapper
+  const { iconKey, componentType, title, key, headline } =
+    getDisplayDetailsForResult(result);
+  const Icon = ICONS[iconKey];
 
   const handleClick = () => {
-    console.log("todo!");
-    console.log({ nodeId: result.item.id });
+    console.log({ result });
```
DafyddLlyr, line 29: The next PR should be handling this!

jessicamcinchak:
@DafyddLlyr FYI data sync failed here & can't log into pizza! https://hasura.3646.planx.pizza/console/data/default/schema/public/tables/users/browse

jessicamcinchak (approved):
Working as expected on pizza & tests are clear here ! Displays values feel very clear :raised_hands:

PR #3652, opened 2024-09-11
fix: Switch to $client for API mutation and replace fetch with Store value in feedback comp

RODO94:
Addressing the bug: https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1726047860980599

I found there was an issue with moveFlow in the API which was using the Public role to getTeamBySlug(). Found that this was unchecked for `submission_email` on Hasura.

Tried locally and it fixes the issue.

### DIFF
--- a/api.planx.uk/modules/flows/moveFlow/service.ts
+++ b/api.planx.uk/modules/flows/moveFlow/service.ts
@@ -4,6 +4,7 @@ import { $public, getClient } from "../../../client/index.js";
 import { Team } from "@opensystemslab/planx-core/types";
 
 export const moveFlow = async (flowId: string, teamSlug: string) => {
+  const $client = getClient();
   const team = await $public.team.getBySlug(teamSlug);
   if (!team)
     throw Error(
--- a/api.planx.uk/modules/flows/moveFlow/service.ts
+++ b/api.planx.uk/modules/flows/moveFlow/service.ts
@@ -5,7 +5,7 @@ import { Team } from "@opensystemslab/planx-core/types";
 
 export const moveFlow = async (flowId: string, teamSlug: string) => {
   const $client = getClient();
-  const team = await $public.team.getBySlug(teamSlug);
+  const team = await $client.team.getBySlug(teamSlug);
   if (!team)
     throw Error(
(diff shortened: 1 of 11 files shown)


DafyddLlyr (changes requested):
Unfortunately, I don't think this is the right solution here.

`submission_email` should not be exposed via the pubic role as it's used to access "send to email" payloads (see https://trello.com/c/EeJb0QVC/2614-harden-download-application-files-endpoint?search_id=8c2b8fda-b422-445a-a594-ae0c9b0360d7 and https://www.notion.so/opensystemslab/Inadequate-hardening-on-Send-to-email-endpoint-02cb855f339b4a1bab49fc0ff8e0dd72?pvs=4)


As we know that "move Flow" is an authorised user only feature, a better solution here would be to ensure that the API is using the correct permission level for this action 👍

RODO94:
> Unfortunately, I don't think this is the right solution here.
> 
> `submission_email` should not be exposed via the pubic role as it's used to access "send to email" payloads (see https://trello.com/c/EeJb0QVC/2614-harden-download-application-files-endpoint?search_id=8c2b8fda-b422-445a-a594-ae0c9b0360d7 and https://www.notion.so/opensystemslab/Inadequate-hardening-on-Send-to-email-endpoint-02cb855f339b4a1bab49fc0ff8e0dd72?pvs=4)
> 
> As we know that "move Flow" is an authorised user only feature, a better solution here would be to ensure that the API is using the correct permission level for this action 👍

Ah yeah okay, this makes sense. It be better for the moveFlow to use `$client` rather than `$public` like in:

```ts
export const getUserAndTeam = async ({
  userEmail,
  teamSlug,
}: {
  userEmail: string;
  teamSlug: string;
}) => {
  const $client = getClient();

  const team = await $client.team.getBySlug(teamSlug);
  if (!team) throw Error(`Unable to find team matching slug ${teamSlug}`);

  const user = await $client.user.getByEmail(userEmail);
  if (!user) throw Error(`Unable to find team matching email ${userEmail}`);

  return { team, user };
};
```
 @DafyddLlyr

DafyddLlyr:
@RODO94 Yep, that should work without changing the column permission structure on Hasura 👍

PR #3655, opened 2024-09-11
fix: Move Flow and Feedback Lib bug fixes

RODO94:
Cleaner branch to see all of the fixes for the `moveFlow` API call and the Feedback library fetch.

Used the `$client` instead of `$public `for the moveFlow `getBySlug()` query.

Removed the fetch for `teamId` in the lib/feedback.ts and replaced it with a store value

### DIFF
--- a/editor.planx.uk/src/lib/feedback.ts
+++ b/editor.planx.uk/src/lib/feedback.ts
@@ -25,10 +25,9 @@ export async function getInternalFeedbackMetadata(): Promise<FeedbackMetadata> {
     breadcrumbs,
     currentCard: node,
     computePassport,
-    fetchCurrentTeam,
+    teamId,
     id: flowId,
   } = useStore.getState();
-  const { id: teamId } = await fetchCurrentTeam();
   const userData = {
     breadcrumbs: breadcrumbs,
     passport: computePassport(),
--- a/editor.planx.uk/src/lib/feedback.ts
+++ b/editor.planx.uk/src/lib/feedback.ts
@@ -25,9 +25,10 @@ export async function getInternalFeedbackMetadata(): Promise<FeedbackMetadata> {
     breadcrumbs,
     currentCard: node,
     computePassport,
-    teamId,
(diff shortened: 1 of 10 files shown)


editor.planx.uk/src/lib/feedback.ts
```diff
@@ -25,10 +25,9 @@ export async function getInternalFeedbackMetadata(): Promise<FeedbackMetadata> {
     breadcrumbs,
     currentCard: node,
     computePassport,
-    fetchCurrentTeam,
+    teamId,
```
jessicamcinchak, line 28: This appears to be `undefined`, so I don't actually think this solution works and we need to do the fetch as before.

On the pizza, I determined this by: 
- Noticing the Feedback Log was empty, even though the front end interaction appeared to succeed (`teamId` is required to properly filter/populate view)
- Inspecting the graphql mutation `InsertFeedback` in my network tab - you'll see it's missing a `teamId` variable (graphql omits variables if undefined)
- double-checking `window.api.getState().teamId` in the console, which also returned `undefined` when on a /published link

jessicamcinchak (commented):
Thanks for this fix, but not quite there yet! 

- :white_check_mark: "Move" is working
- :white_check_mark: Send to email is working
- :x: Feedback is NOT working - please see comment

Also please make note of "GitHub Action" warning picking up an unused import! We can ignore these for files that haven't been changed, but good to pay attention to when flagging something in your changes!

editor.planx.uk/src/lib/feedback.ts
RODO94: @jessicamcinchak using the updated `getBySlug()` with `submission_email` removed seems to fix this for me locally, with team_id being in the database.

Going to check it on the pizza now

RODO94: Pizza is working for me

jessicamcinchak (approved):
All working as expected for me now :white_check_mark:

PR #3665, opened 2024-09-12
chore: Improve logging on Notify errors

DafyddLlyr:
First attempt to get a better insight into https://planx.airbrake.io/projects/329753/groups/3890967064109750041?tab=overview


jessicamcinchak (approved):
Thanks for this, also noticed that confirmation template bug. 

One small thought: it's been really handy to extract Hasura send event details out into the submissions log view to save paging through the event logs themselves - I wonder if it's worth doing the same for email events (just internal view, no reason to expose in editor yet)?

PR #3666, opened 2024-09-12
feat: Add internal portal wrapper to search cards

DafyddLlyr:
- **chore: Add icon for internal portals**
- **feat: Add internal portal header to search cards**


editor.planx.uk/src/@planx/components/ui.tsx
```diff
@@ -6,6 +6,7 @@ import CloudUpload from "@mui/icons-material/CloudUpload";
 import ContactPage from "@mui/icons-material/ContactPage";
 import CopyAll from "@mui/icons-material/CopyAll";
 import Create from "@mui/icons-material/Create";
+import DoorFrontOutlined from "@mui/icons-material/DoorFrontOutlined";
```
jessicamcinchak, line 9: nit: could [a simple folder icon](https://mui.com/material-ui/material-icons/?query=folder&selected=FolderOpen) be a more straightforward option here to communicate "this is nested inside something else"? Took me a while to make out exactly what the door was as the size it's used. 

Maybe also more in line with future "what do we call things" conversations where internal portals become "folders" and external portals become "subscribed flows" or similar!

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/SearchResultCard/index.tsx
```diff
@@ -13,72 +13,104 @@ import { FONT_WEIGHT_SEMI_BOLD } from "theme";
...
+  return (
+    <HeaderRoot>
+      {Icon && <Icon />}
+      <Typography
+        variant="body2"
+        fontSize={14}
+        fontWeight={FONT_WEIGHT_SEMI_BOLD}
+        ml={1}
+      >
+        Internal portal - {portalName}
```
jessicamcinchak, line 49: nit: wonder if we can actually drop `Internal portal - ` prefix here? I _think_ the black border style & icon are sufficient? 

The `{type} - ` prefix makes sense to me when it's used to say "this is the type of modal you're going to open if you click this", but in the the case of the internal portal here you're never navigating directly to the internal portal node modal, just simply communicating an organisation structure? 

Also easier to start without & simply add back if comes up as point of confusion in testing?

jessicamcinchak (approved):
Functionally works exactly as expected ! 

A few small styling nits, please take them or leave them!

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/SearchResultCard/index.tsx
DafyddLlyr: This works well for me - we can easily add back in, but pretty much no users will suggest moving visual clutter / noise 👍 Implemented in [da1675c](https://github.com/theopensystemslab/planx-new/pull/3666/commits/da1675c5fdfaa4167b78ccfbc1bb643c2a694c55)

editor.planx.uk/src/@planx/components/ui.tsx
DafyddLlyr: Yep this is a fair question!

I'm just re-using and formalising the existing icon here, as seen in the flow graph.

image

I'll keep as-is for now, but let's re-evaluate this in line with our "naming things" chat 👍

PR #3670, opened 2024-09-13
chore: extend `SiteAddress` type to account for ranges

jessicamcinchak:
It came up in Idox feedback that we aren't currently supporting address ranges in the ODP Schema. I looked through the OneApp JSON example we have and don't see any instances of it handling ranges either, but nonetheless it's low effort for us to start more carefully checking for these. 

So, this PR adds a few extra fields from the OS Places API "LPI" data source to our internal model which we can later decide how to send along in payloads / the ODP Schema (https://github.com/theopensystemslab/digital-planning-data-schemas/pull/236). 

Only additive changes here! 
- `saoEnd`
- `paoEnd`
- `ward`
- `parish`

Links: 
- OS Docs (ctrl + f "LPI Output"): https://osdatahub.os.uk/docs/places/technicalSpecification
- Example response: https://api.editor.planx.dev/proxy/ordnance-survey/search/places/v1/postcode?postcode=SE5+0HU&dataset=LPI&maxResults=100&output_srs=EPSG%3A4326&lr=EN&offset=0

### DIFF
--- a/editor.planx.uk/src/@planx/components/FindProperty/Public/Autocomplete.tsx
+++ b/editor.planx.uk/src/@planx/components/FindProperty/Public/Autocomplete.tsx
@@ -79,7 +79,19 @@ export default function PickOSAddress(props: PickOSAddressProps): FCReturn {
           latitude: selectedAddress.LAT,
           longitude: selectedAddress.LNG,
           organisation: selectedAddress.ORGANISATION || null,
-          sao: selectedAddress.SAO_TEXT,
+          sao: [
+            selectedAddress.SAO_START_NUMBER,
+            selectedAddress.SAO_START_SUFFIX,
+            selectedAddress.SAO_TEXT, // populated in cases of building name only, no street number
+          ]
+            .filter(Boolean)
+            .join(""),
+          saoEnd: [
+            selectedAddress.SAO_END_NUMBER,
+            selectedAddress.SAO_END_SUFFIX,
+          ]
+            .filter(Boolean)
+            .join(""),
           pao: [
             selectedAddress.PAO_START_NUMBER,
             selectedAddress.PAO_START_SUFFIX,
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/FindProperty/Public/Autocomplete.tsx
```diff
@@ -79,17 +79,37 @@ export default function PickOSAddress(props: PickOSAddressProps): FCReturn {
...
           longitude: selectedAddress.LNG,
           organisation: selectedAddress.ORGANISATION || null,
-          sao: selectedAddress.SAO_TEXT,
+          sao: [
+            selectedAddress.SAO_START_NUMBER,
+            selectedAddress.SAO_START_SUFFIX,
+            selectedAddress.SAO_TEXT, // populated in cases of building name only, no street number
+          ]
+            .filter(Boolean)
+            .join(""),
```
jessicamcinchak, line 88: This updates existing `sao` field to be handled same as `pao`, whereas previously we only accounted for `SAO_TEXT` and would miss `SAO_START_NUMBER` or `SAO_START_SUFFIX`

editor.planx.uk/src/@planx/components/FindProperty/model.ts
```diff
@@ -48,15 +48,19 @@ export interface MinimumSiteAddress {
...
   organisation?: string | null;
   sao?: string | null;
+  saoEnd?: string | null;
   pao?: string;
+  paoEnd?: string;
   street?: string;
   town?: string;
   postcode?: string;
+  ward?: string;
+  parish?: string;
```
jessicamcinchak, line 63: Generally a lot more "optional" types in here than I'd like - but OS Docs are really hard to make sense of about what's required versus not sometimes - for example, `PARISH_CODE` has a "multiplicity" of 1 (not 0..1) which I understand to mean required yet it's not one of the returned fields in the example response linked above.

DafyddLlyr: Helpful thanks!

DafyddLlyr (approved):

PR #3671, opened 2024-09-14
feat: Pesist `FlowEditor` state on route changes

DafyddLlyr:
- **fix: Stop re-render of most of the app on each rout change**
- **feat: Single instance of flow editor to persist state**
- **chore: Move settings pages to parent scope to escape flow editor view**
- **chore: Remove legacy /service-flags and /data routes and associated components**
- **fix: Don't flash login screen on load**


editor.planx.uk/src/pages/Login.tsx
```diff
@@ -46,6 +48,9 @@ const LoginButton = styled(Button)(({ theme }) => ({
 }));
 
 const Login: React.FC = () => {
+  const isLoading = useLoadingRoute();
+  if (isLoading) return <DelayedLoadingIndicator />;
```
DafyddLlyr, line 52: Without this, the login page flashed on initial load of the application.

editor.planx.uk/src/pages/layout/FlowEditorLayout.tsx
```diff
@@ -1,9 +1,13 @@
 import ErrorFallback from "components/ErrorFallback";
+import FlowEditor from "pages/FlowEditor";
 import React, { PropsWithChildren } from "react";
 import { ErrorBoundary } from "react-error-boundary";
 
 const FlowEditorLayout: React.FC<PropsWithChildren> = ({ children }) => (
-  <ErrorBoundary FallbackComponent={ErrorFallback}>{children}</ErrorBoundary>
+  <ErrorBoundary FallbackComponent={ErrorFallback}>
+    <FlowEditor />
```
DafyddLlyr, line 8: This is the single `FlowEditor` instance, attached to the layout, which persists regardless of route.

editor.planx.uk/src/routes/team.tsx
```diff
@@ -74,6 +74,12 @@ const routes = compose(
       return import("./flow");
     }),
 
+    "/:flow/feedback": lazy(() => import("./feedback")),
+
+    "/:flow/service": lazy(() => import("./serviceSettings")),
+
+    "/:flow/submissions-log": lazy(() => import("./submissionsLog")),
+
```
DafyddLlyr, line 82: These needed to be moved as they do not require `FlowEditor` as part of their layout. They now exist within the same parent view.

jessicamcinchak: One strange thing I'm noticing here that I suspect is related to this:
- Open a team > Open a flow > Navigate to `/:flow/submissions-log` from menu side icons
- Refresh page from submissions log > now Team settings menu is visible, not the flow side icons! 

A bit disorienting / unclear how to "go back" ! 
Screenshot from 2024-09-16 10-23-12

jessicamcinchak: `:flow/service` does the same thing, while `:flow/feedback` throws a "Not found" error page refresh (not new I think).

jessicamcinchak (commented):
Big one! Thanks for spending time with this. 

Code is looking good, one note/question about some unexpected pizza behavior though

editor.planx.uk/src/routes/team.tsx
DafyddLlyr: Thanks - very well spotted. I can recreate this as described, taking a look.

DafyddLlyr: This should now be fixed by [4351014](https://github.com/theopensystemslab/planx-new/pull/3671/commits/4351014b22e4817e17c6c9690f44265f6d0ebb81)

`EditorNavMenu` relies on `flowSlug`, which wasn't getting set on a refresh.

A wider refactor to move these routes to `/:flow/settings/:page` might be a more robust approach, but the above solution works for me - we can always take another look at this if the number of pages grow, or offload the responsibility to the nav menu itself.

jessicamcinchak (approved):
Previous issue now fixed and everything else working as expected, thanks !!

DafyddLlyr:
Moved to draft, hit a bug here where I can't route to internal portals. Will re-open once fixed.

DafyddLlyr:
I have an implementation of a loading bar here - https://github.com/theopensystemslab/planx-new/pull/3684 

Thoughts welcome!

PR #3673, opened 2024-09-14
fix: Add flow.settings back to public queries

DafyddLlyr:


jessicamcinchak (approved):
Thanks for catching this!! Feels like a strong contender for future e2e test coverage?

DafyddLlyr:
@jessicamcinchak Yep had this same thought! Quite a bad oversight on our part and really only properly testable in an E2E environment (mocks would have failed here also more likely than not).

PR #2947, opened 2024-04-01
feat: expanded graph styling for biennale exhibit

jessicamcinchak:
Builds on all the good stuff started here #2912 

This is a nice flow to visualise changes: https://2947.planx.pizza/opensystemslab/permitteddevelopment-solar-panels,vLhJmdJp1E

No intention or urgency to merge this one, exclusively spinning up pizza for demo purposes - need to think about "print" next !

PR #3674, opened 2024-09-15
feat: display data fields on graph

jessicamcinchak:
Cherry picked from our prior Biennale experimental / maximalist graph styling #2947 

Wondering if displaying data fields in the graph makes sense alongside preparing to launch search for "data fields only"? Eg, how familiar do we think average editors are with the concept of "data fields" ?

Just an idea, very much think this one is up to Daf, Ian and content team to decide if useful! While I like this style for increased visibility, it definitely stretches out overall size of graphs which may prove harder to navigate?

**Before:**
Screenshot from 2024-09-15 20-22-02

**After:**
Screenshot from 2024-09-15 20-16-58

** Does _not_ display data fields (yet?) for any of the "special" component types like DrawBoundary, FileUploadAndLabel etc that search results _are_ handling.

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Checklist.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Checklist.tsx
@@ -86,8 +86,13 @@ const Checklist: React.FC<Props> = React.memo((props) => {
           onContextMenu={handleContext}
           ref={drag}
         >
-          {Icon && <Icon />}
-          <span>{props.text}</span>
+          <div className="question-icon-wrapper">
+            {Icon && <Icon />}
+            <span className="question-text">{props.text}</span>
+          </div>
+          {props.data?.fn && (
+            <div className="question-schema">{props.data.fn}</div>
+          )}
         </Link>
         {groupedOptions ? (
           <ol className="categories">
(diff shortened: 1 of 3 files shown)


DafyddLlyr (approved):
Approved from a code POV - agree there's a bigger design question here to be decided on.

From a search / design perspective - I wonder if exposing this would make Editors expect there's more of a visual link between search results and Graph nodes (e.g. would there be an expectation that the data fields be "highlighted" in the graph the same way a search card's headline is?). Personally, I'd be against this for a few reasons - 
 - Complexity and overhead!
 - It breaks "Search" out of the right hand panel and sort of makes it a globally scoped feature
 - There's not a 1-to-1 relationship between graph node and data value, so it's never going to perfectly marry up. Likewise, this feature wouldn't work once search looks at non-data fields
 
So in summary, I think it's a useful and valuable addition to the graph (maybe we can have it toggle-able?), but I'd vote to not have these values dynamically linked to search results.

> Does not display data fields (yet?) for any of the "special" component types like DrawBoundary, FileUploadAndLabel etc that search results are handling.

Some of these are fairly complex as there's not a 1-to-1 relationship, I think it's totally fair to not expose the complexity and just show the "high level" data field which most components have.

jessicamcinchak:
Closing in favor of #3789 which introduces toggle button & React based "DataField" rather than scss like here

PR #3676, opened 2024-09-16
feat(page): Editor modal

DafyddLlyr:
## What does this PR do?
- Sets up Editor modal for Page component
- Copies "Advert Consent" schema from List component - this is being used as a page there

image

### DIFF
--- a/editor.planx.uk/src/@planx/components/Page/model.ts
+++ b/editor.planx.uk/src/@planx/components/Page/model.ts
@@ -1,10 +1,27 @@
+import { cloneDeep } from "lodash";
+
 import { MoreInformation, parseMoreInformation } from "../shared";
+import { Schema } from "../shared/Schema/model";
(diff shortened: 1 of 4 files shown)


editor.planx.uk/src/@planx/components/Page/model.ts
```diff
@@ -1,10 +1,27 @@
+import { cloneDeep } from "lodash";
+
 import { MoreInformation, parseMoreInformation } from "../shared";
+import { Schema } from "../shared/Schema/model";
+import { PAGE_SCHEMAS } from "./Editor";
+
+export interface PageSchema extends Schema {
+  min: 1;
+  max: 1;
+}
```
DafyddLlyr, line 10: This enforces that Pages can only have one item, whilst allowing us to re-use all logic from `useSchema()`

jessicamcinchak (approved):
Code all looks good here, happy for this to merge once Vultr sorted

PR #3678, opened 2024-09-16
feat(page): Handle submit and "back" functionality

DafyddLlyr:
## What does this PR do?
 - Handles submission of payload data
 - Sets up state when navigating "back"
 - Tests for Page component
 - Bumps planx-core to latest `main` commit across projects

image


editor.planx.uk/src/@planx/components/Page/Public.tsx
```diff
@@ -1,19 +1,59 @@
...
 
 type Props = PublicProps<Page>;
 
 function PageComponent(props: Props) {
+  const { formikConfig } = useSchema({
+    schema: props.schema,
+    previousValues: getPreviouslySubmittedData(props),
+  });
+
+  const onSubmit: FormikConfig<SchemaUserData>["onSubmit"] = (data) => {
```
DafyddLlyr, line 22: As with the `List` component, we're preversing both forms of data here as opposed to writing to and parsing from a single format.

This seems like an ok approach, especially as it's following an established pattern?

RODO94: By "preversing both forms of data" you're referring to `defaultPassportData` and `flattenedData`?

jessicamcinchak (approved):
Data structures make sense to me here, agree List is right precedent/pattern to follow here :+1:

editor.planx.uk/src/@planx/components/Page/Public.tsx
DafyddLlyr: Yes, that's right. We could parse/map/serialise both into a single structure and then unpack that on "back" navigation, but I'm proposing we allow both just now 👍

PR #3679, opened 2024-09-16
feat(wip): SummaryList view of Page component

DafyddLlyr:
## What does this PR do?
- Adds a view of the Page component inside SummaryList (Review component)

image

Would love some thoughts here @ianjon3s - very much not sold on this UI. 

Breaking this out into its own `dl` seems like the better solution here, but that would require a bit of a wider refactor due to how we're currently inserting components into cells. So I'm offering this up as a very rough version which at least handles getting the data to the right place, and to open up the discussion on the user interface

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
@@ -9,12 +9,15 @@ import {
 } from "@opensystemslab/planx-core/types";
 import { PASSPORT_UPLOAD_KEY } from "@planx/components/DrawBoundary/model";
 import { PASSPORT_REQUESTED_FILES_KEY } from "@planx/components/FileUploadAndLabel/model";
+import type { Page } from "@planx/components/Page/model";
 import { ConfirmationDialog } from "components/ConfirmationDialog";
 import format from "date-fns/format";
 import { useAnalyticsTracking } from "pages/FlowEditor/lib/analytics/provider";
 import { Store, useStore } from "pages/FlowEditor/lib/store";
 import React, { useState } from "react";
 import { FONT_WEIGHT_SEMI_BOLD } from "theme";
+
+import { SchemaUserResponse } from "../Schema/model";
 
 export default SummaryListsBySections;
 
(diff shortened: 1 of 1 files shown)


ianjon3s:
I agree @DafyddLlyr, not ideal visually, and tricky as you say with the setup of injecting data into a pre-defined layout.

However it does communicate the information in a way that is clear about the structure of the questions, and collapses to a single column correctly on smaller screens, so is more than adequate at this point.

editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
```diff
@@ -618,6 +620,25 @@ function FileUploadAndLabel(props: ComponentProps) {
...
+  const fields = (props.node.data as Page).schema.fields;
+
+  return (
+    <>
+      <Box component="dt">{props.node.data.title}</Box>
+      <Box component="dd" sx={{ gap: 2, display: "flex", flexDirection: "column" }}>
+        {fields.map((field) => (
+          <Box>
+            <Typography fontWeight={FONT_WEIGHT_SEMI_BOLD}>{field.data.title}</Typography>
+            <Typography>{answers[0][field.data.fn]}</Typography>
```
jessicamcinchak, line 634: nit: I think this needs to account for a `"map"` type field separately - either by showing a static map or a simple summary (eg "2 features") ??

jessicamcinchak (approved):
Simple & clean MVP solution :+1: 

One note around `"map"` field types, but that can be fixed-forward I think !

editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
DafyddLlyr: Nice catch! Fixed in [11aa28a](https://github.com/theopensystemslab/planx-new/pull/3679/commits/11aa28ae166ea1506ae6e1e806f6a887ecd69fea)

jessicamcinchak: Perfect, thanks !

PR #3683, opened 2024-09-16
test[e2e]: refactor, and add to public-facing editor flow test

jamdelion:
Trello ticket: https://trello.com/c/dBmTyu9F/3017-test-each-planx-component-via-an-e2e-test

## In this PR

- Refactor to use [page object model](https://playwright.dev/docs/pom) for editor
- Split out the create-flow test into two files to reduce length/speed up local test runs
- Add to the 'public facing' part of the 'create-flow' test and added helpers that matched the format of the existing ones

## Next steps

You might have noticed I've stopped at the `drawBoundary` component! I think I might shift some of the 'add component' steps around so that I can tackle easier components first (there is a[ separate ticket for geospatial components](https://trello.com/c/pjzQLPd2/3018-setup-end-to-end-tests-for-geospatial-components)). 😅 

❓ Perhaps I can change the test to not be order-dependent - i.e. create many parallel flows that have different components in them, rather than one long one. This might be more like unit testing though, and perhaps would lose some of the benefit of trying to have a more realistic flow (with many different components in it).


editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
jamdelion, line 1: All files below here are just linting changes, not sure why there are so many - sorry!

e2e/tests/ui-driven/src/create-flow/create-flow.spec.ts
```diff
@@ -321,5 +201,66 @@ test.describe("Navigation", () => {
     await expect(
       page.locator("h1", { hasText: "Sorry, this is a test" }),
     ).toBeVisible();
+    await clickContinue({ page });
```
jamdelion, line 204: Adding to the 'user-facing' part of the test

e2e/tests/ui-driven/src/create-flow/create-flow.spec.ts
```diff
@@ -52,166 +43,55 @@ test.describe("Navigation", () => {
...
-    });
-    await expect(teamSlugInHeader).toBeVisible();
-
-    await page.reload();
-    await expect(teamSlugInHeader).toBeVisible();
-
-    await page.goBack();
-    await expect(teamSlugInHeader).toBeHidden();
-  });
-
```
jamdelion, line 108: Split out into a 'page-refresh' test file to reduce the length of this one

e2e/tests/ui-driven/src/mocks/osPlacesResponse.ts
```diff
@@ -58,3 +60,14 @@ export const mockOSPlacesResponse = {
...
+  const ordnanceSurveryPlacesEndpoint = new RegExp(
+    /proxy\/ordnance-survey\/search\/places\/v1\/postcode\/*/,
+  );
+  await page.route(ordnanceSurveryPlacesEndpoint, async (route) => {
+    await route.fulfill({
+      status: 200,
+      body: JSON.stringify(mockOSPlacesResponse),
+    });
+  });
+}
```
jamdelion, line 73: Moved from userActions file to here

e2e/tests/ui-driven/src/create-flow/create-flow.spec.ts
```diff
@@ -52,166 +43,55 @@ test.describe("Navigation", () => {
...
+    await editor.createChecklist();
+    await editor.createTextInput();
+    await editor.createNumberInput();
+    await editor.createDateInput();
+    await editor.createAddressInput();
+    await editor.createContactInput();
+    await editor.createTaskList();
+    await editor.createFindProperty();
+    await editor.createDrawBoundary();
+    await editor.createPlanningConstraints();
```
jessicamcinchak, line 72: Per bigger picture approach questions in your PR description: 
- I'm definitely in favor of splitting out geo components into separate tests/flows as own Trello ticket suggests - makes perfect sense that draw boundary & other map-based ones are going to be a bit of a stumper here / need special handling 
  - This matches current separate tests for Pay & Send - which I wouldn't expect to necessarily be in this list either ! 
- Otherwise though I like the idea of as many component types in a single flow as we can for realistic testing - if you're finding it's very long running maybe we simply mark as `@regression` so it's picked up on the nightly run, but not per commit?

jessicamcinchak (approved):
Changes all look good here to me! Nice to see this list of components growing :tada:

PR #3684, opened 2024-09-16
feat: Route loading progress bar

DafyddLlyr:
An implementation of a routing loading bar as opposed to the FOUC and `DelayedLoadingIndicator`.

Follows on from #3671 

**Questions** -
 - Is 50ms of delay too short? (is it showing too often and flashing and dissapearing?)
 - Style 
   - should it be slimmer?
   - should it be a more subtle colour
 - Do we want to consider something similar on the public interface?

### DIFF
--- a/editor.planx.uk/src/components/RouteLoadingIndicator.tsx
+++ b/editor.planx.uk/src/components/RouteLoadingIndicator.tsx
@@ -1,0 +1,36 @@
+import Box from "@mui/material/Box";
+import LinearProgress from "@mui/material/LinearProgress";
+import { styled } from "@mui/material/styles";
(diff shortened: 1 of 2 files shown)


jessicamcinchak (approved):
Code wise this all looks good ! Will defer design questions to Ian, but I like your defaults here / feels like "just enough" visual feedback :+1:

DafyddLlyr:
Thanks @jessicamcinchak 🙌 

Keen to progress this and draw a line so I'll mere both PRs to `main` - confident enough it's a solid improvement even if not visually perfect. I'll then flag at standup and in `planx-design` just to make sure a design eye is cast over it 👍

PR #3689, opened 2024-09-17
fix: Move "invalid graph" warning to `MapFieldInput`

DafyddLlyr:
## What?
Moves the "invalid graph" warning from a high level component, to the lower level `MapFieldInput`

## Why?
I noticed when working on [11aa28a](https://github.com/theopensystemslab/planx-new/pull/3679/commits/11aa28ae166ea1506ae6e1e806f6a887ecd69fea) that I also needed to handle this error in the `Page` component. The actual component responsible for this dependency is the `MapFieldInput`, so it makes sense to do the error handling here once as opposed to remembering to do it in each parent which renders this component.


editor.planx.uk/src/@planx/components/shared/Schema/InputFields/MapFieldInput.tsx
```diff
@@ -11,6 +12,19 @@ import { getFieldProps, Props } from ".";
...
+  const { longitude, latitude } = useStore(
+    (state) =>
+      (state.computePassport()?.data?.["_address"] as SiteAddress) || {},
+  );
+
+  if (!longitude || !latitude) {
+    throw Error(
+      'Edit this flow so that this component is positioned after "FindProperty"; an address is required for schemas that include a "map" field.',
+      { cause: "Invalid graph" },
+    );
```
jessicamcinchak, line 25: nit: I'm not sure I'm sold on throwing an error here rather than rendering a component as before - do we want this noise coming through Airbrake? 

Understand we can't easily render the whole component any longer from this shared position, but just wondering if there's a mechanism to exclude from Airbrake here etc?

jessicamcinchak (approved):
Works exactly as expected, just one question below about how much visibility we actually need here if mostly intended for editor feedback !

I think you've also missed _removing_ the existing `ErrorSummaryContainer` from Map & Label, therefore still falling back to component first rather than field-level `throw` here !

RODO94 (approved):
Tested on the pizza and I still get the correct error messages. Although the format of the error is different to the error with the Map and Label component. I see this as separate to the purpose of this PR so happy to approve.

Should these be the same? 

- List Comp Error
Screenshot 2024-09-17 at 12 02 03

- Map and Label Comp
Screenshot 2024-09-17 at 12 01 54

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.test.tsx
DafyddLlyr, line 1: There's some basic tests here currently missing, I'll see if I can set this up in a follow up PR.

editor.planx.uk/src/components/Error/ErrorFallback.tsx
```diff
@@ -1,12 +1,15 @@
...
+import { logger } from "airbrake";
 import React from "react";
 
-import { logger } from "../airbrake";
+import { GraphErrorComponent, isGraphError } from "./GraphError";
 
-function ErrorFallback(props: { error: Error }) {
-  logger.notify(props.error);
+const ErrorFallback: React.FC<{ error: Error }> = ({ error }) => {
+  if (isGraphError(error)) return <GraphErrorComponent error={error} />;
```
jessicamcinchak, line 10: This is a neat solution & I think we'll appreciate not having extra noise from these, thanks!

jessicamcinchak (approved):

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.test.tsx
DafyddLlyr: https://github.com/theopensystemslab/planx-new/pull/3700

PR #3690, opened 2024-09-17
feat: Map and Label Happy Path Tests

RODO94:
## What does this PR do?

Running through the Happy Path test to-dos. Tried to do a nice amount of refactoring to keep the testing concise but explicit, so I have some simple functions handling "Clicking Continue" and "Checking for Error Messages are Populated or Empty" since these are repeatable situations without a high amount of meaning or context for the test. 

I would be interested to know if I should be waiting for a `handleSubmit()` to be called to confirm the happy path is successful rather than just checking the absence of error messages.

### DIFF
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
@@ -11,6 +11,9 @@ import { props } from "../test/mocks/Trees";
 import {
   addFeaturesToMap,
   addMultipleFeatures,
+  checkErrorMessagesEmpty,
+  checkErrorMessagesPopulated,
+  clickContinue,
   fillOutFirstHalfOfForm,
   fillOutForm,
   fillOutSecondHalfOfForm,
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -214,16 +189,137 @@ describe("validation and error handling", () => {
...
+    expect(thirdTab).toHaveAttribute("aria-selected", "true");
+    expect(secondTab).toHaveAttribute("aria-selected", "false");
+    expect(firstTab).toHaveAttribute("aria-selected", "false");
+  });
+
+  it("a user can input details on multiple features and submit", async () => {
+    const { getByTestId, getByRole, user, getAllByTestId } = setup(
+      <MapAndLabel {...props} />,
+    );
+    const map = getByTestId("map-and-label-map");
```
DafyddLlyr, line 248:
```suggestion
  it("a user can input details on multiple features and submit", async () => {
    const { getByTestId, getByRole, user } = setup(
      <MapAndLabel {...props} />,
    );
```

nit: A few unused variables here and there in this test file, have a quick scan through 👍

DafyddLlyr (approved):
A few very small nit comments to take a look through here.

PR is great through - clean and easy to ready, love it 👍 

On the `handleSubmit()` - it's worth explicitly testing that this has not been called when validation errors are shown. I'd suggest that rather than adding this to a random test or each test, just add a new test for just this instance - a little more atomic and explicit.

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -186,6 +186,22 @@ describe("validation and error handling", () => {
   });
   // shows the error state on a tab when it's invalid
 });
+
+it("does not trigger handleSubmit when errors exist", async () => {
```
DafyddLlyr, line 188: 👌

DafyddLlyr (approved):

PR #3668, opened 2024-09-13
refactor: Split up Sidebar child components

DafyddLlyr:
## What does this PR do?
 - Quick refactor / tidy up before taking a look at maintaining the tabs and search value in state as part of search
 - Makes `DebugConsole` its own file / component
 - Moves `Publish` to it's own folder and components
 - Doesn't unnecessarily pass a URL as props (see https://github.com/theopensystemslab/planx-new/pull/3483#discussion_r1705918426)
 - Declares URLs in a data structure instead of dynamically creating the URLs across the component (see https://github.com/theopensystemslab/planx-new/pull/3483#discussion_r1705915756)

No "new" code or functionality here

PR #3692, opened 2024-09-17
fix: Flow URLs should re-compute on re-render of Sidebar

DafyddLlyr:
## What's the problem?
 - Bug introduced in https://github.com/theopensystemslab/planx-new/pull/3668
 - URLs are not being re-generated when the route changes
 - This is because by defining `urls` outside of the lifecycle scope of a component, this is never re-generated
 - To recreate - 
   - Navigate to a flow in the Editor, checkout icon hrefs in Sidebar
   - Navigate to another flow
   - Icon hrefs have not updated ❌ 

## What's the solution?
- Calculate within the `Sidebar` component
- Pass URL as prop, instead of importing it

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Publish/PublishFlowButton.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Publish/PublishFlowButton.tsx
@@ -14,7 +14,6 @@ import React, { useState } from "react";
 import { useAsync } from "react-use";
 import Input from "ui/shared/Input";
 
-import { urls } from "..";
 import {
   AlteredNode,
   AlteredNodesSummaryContent,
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Publish/PublishFlowButton.tsx
```diff
@@ -147,7 +148,7 @@ export const PublishFlowButton = () => {
                 <Box pb={2}>
                   <Typography variant="body2">
                     {`Preview these content changes in-service before publishing `}
-                    <Link href={urls.preview} target="_blank">
+                    <Link href={previewURL} target="_blank">
```
RODO94, line 151: nit: should we be using `rel="noopener"` in combo with the `target="_blank"` - I've always done it for security reasons, but I don't know enough to determine whether it's actually useful in all scenarios

RODO94 (approved):
Works as expected for me! Added one wee nit / question

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Publish/PublishFlowButton.tsx
DafyddLlyr: Good spot! When linking internally this isn't required, but is [best practice when using any third-party resources](https://mui.com/material-ui/react-link/#security).

PR #3693, opened 2024-09-17
fix: hide add/edit user in Templates team and for archived members

jessicamcinchak:
Per #planx-design thread: https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1726564743563709

**Changes:**
- Hide the add/edit buttons in the "Templates" team as membership to this team is specially handled via a DB trigger
- Don't show an edit button for archived members


DafyddLlyr (approved):
Perfect!

PR #3695, opened 2024-09-17
feat: Map and Label Copy Tests

RODO94:
## What does this PR do?

This continues work on the tests for the Map and Label component, building up tests for the _Copy from_ feature

### DIFF
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
@@ -1,6 +1,6 @@
 import { MyMap } from "@opensystemslab/map";
 import { Presentational as MapAndLabel } from "@planx/components/MapAndLabel/Public";
-import { waitFor, within } from "@testing-library/react";
+import { screen, waitFor, within } from "@testing-library/react";
 import React from "react";
 import { setup } from "testUtils";
 import { vi } from "vitest";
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -332,18 +332,108 @@ describe("basic interactions - happy path", () => {
...
+  it("is enabled once multiple features are present", async () => {
+    const { getByTitle } = setup(<MapAndLabel {...props} />);
+
+    addMultipleFeatures([point1, point2]);
+
+    const copyTitle = getByTitle("Copy from");
+
+    const copyInput = within(copyTitle).getByRole("combobox");
+
+    expect(copyInput).not.toHaveAttribute("aria-disabled", "true");
```
RODO94, line 357: This was the best option I could find to prove if something is disabled or not. Happy to know if there's a better option

DafyddLlyr: This is the best way - `toBeDisabled()` won't work with MUI as it's using a combobox + input instead of a native select 👍

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -332,18 +332,108 @@ describe("basic interactions - happy path", () => {
...
+
+  it("should not have any accessibility violations", async () => {
+    const { getByTitle, user, container } = setup(<MapAndLabel {...props} />);
+    addMultipleFeatures([point1, point2]);
+
+    const copyTitle = getByTitle("Copy from");
+
+    const copyInput = within(copyTitle).getByRole("combobox");
+
+    expect(copyInput).not.toHaveAttribute("aria-disabled", "true");
```
DafyddLlyr, line 430: This seems like a redundant check here - it's already tested elsewhere?

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -332,18 +332,108 @@ describe("basic interactions - happy path", () => {
...
+    const speciesInputTwo = getByLabelText("Species");
+    const workInputTwo = getByLabelText("Proposed work");
+    const justificationInputTwo = getByLabelText("Justification");
+    const urgencyDivTwo = getByTitle("Urgency");
+    const urgencySelectTwo = within(urgencyDivTwo).getByRole("combobox");
+
+    expect(speciesInputTwo).toHaveDisplayValue("Larch");
+    expect(workInputTwo).toHaveDisplayValue("Chopping it down");
+    expect(justificationInputTwo).toHaveDisplayValue("Cause I can");
+    expect(urgencySelectTwo).toHaveTextContent("Low");
```
DafyddLlyr, line 396: This is really just testing that `fillOutForm()` is working as expected, which if not already tested, will get covered when it comes to looking at the payloads.

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -332,18 +332,108 @@ describe("basic interactions - happy path", () => {
...
+
+    await user.click(copyInput);
+
+    // Current item would be Tree 2 since we added two points
+    const listItemTwo = queryByRole("option", { name: "Tree 2" });
+
+    expect(listItemTwo).not.toBeInTheDocument();
+  });
+
+  it("copies all data from one feature to another", async () => {
```
DafyddLlyr, line 378: This test is slightly harder to read as we're not clearly defining the original inputs anywhere - they're nested within `filloutForm()`. 

If we had these values (as an object?) we could more easily compare that the values from the second tab match those used in `fillOutForm()`.

DafyddLlyr (approved):
A few minor suggestions, but all good!

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.test.tsx
```diff
@@ -332,18 +333,97 @@ describe("basic interactions - happy path", () => {
...
+    await user.click(copyInput);
+
+    const listItemTwo = getByRole("option", { name: "Tree 2" });
+
+    await user.click(listItemTwo);
+
+    const urgencyDiv = getByTitle("Urgency");
+    const urgencySelect = within(urgencyDiv).getByRole("combobox");
+
+    expect(getByLabelText("Species")).toHaveDisplayValue(mockTreeData.species);
```
RODO94, line 404: @DafyddLlyr is this the sort of thing you meant?

DafyddLlyr: Perfect!

DafyddLlyr (approved):

PR #3697, opened 2024-09-17
fix: realign members table cell borders

jamdelion:
| Before                      | After                  |
| ------------------------ | ------------------------ |
|  |  |


jessicamcinchak (approved):
Much better, thanks!

PR #3700, opened 2024-09-18
test: Basic coverage for `PlanningConstraints` public interface

DafyddLlyr:
## What?
Sets up very basic coverage of the public interface of the `PlanningConstraints` component.

## Why?
 - Follow up to https://github.com/theopensystemslab/planx-new/pull/3689#discussion_r1764504784
 - Upcoming work may focus on the Editor interface, it's worth having a good test setup here which can be expanded alongside this work

### DIFF
--- a/editor.planx.uk/src/@planx/components/PlanningConstraints/Public.test.tsx
+++ b/editor.planx.uk/src/@planx/components/PlanningConstraints/Public.test.tsx
@@ -1,21 +1,27 @@
 import ErrorFallback from "components/Error/ErrorFallback";
+import { useStore } from "pages/FlowEditor/lib/store";
 import React from "react";
+import { act } from "react-dom/test-utils";
 import { ErrorBoundary } from "react-error-boundary";
+import swr from "swr";
 import { setup } from "testUtils";
 import { vi } from "vitest";
 import { axe } from "vitest-axe";
 
 import classifiedRoadsResponseMock from "./mocks/classifiedRoadsResponseMock";
 import digitalLandResponseMock from "./mocks/digitalLandResponseMock";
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/@planx/components/PlanningConstraints/Public.test.tsx
```diff
@@ -55,18 +58,137 @@ describe("error state", () => {
...
+
+    // Planning data is called first
+    const swrURL = swrMock.calls[0][0]();
+    const swrResponse = swrMock.results[0].value;
+
+    expect(swrURL).toContain("/gis");
+    expect(swrResponse).toEqual({ data: digitalLandResponseMock });
+  });
+
+  it("fetches classified roads only when we have a siteBoundary", () => {
```
jessicamcinchak, line 137: nit: roads fetch relies on USRN, not `siteBoundary` - I think description is a bit misguided/outdated here ?  

I _think_ better test scenarios are:
- If user selected an OS address & has a USRN, we can check/fetch roads
- If user proposed a new address & only has a lat/long, we cannot check/fetch roads (but we can still check planning data)

jessicamcinchak (approved):
Long overdue, thanks for getting foundational mocks setup here :+1: 

One comment below re fetch logic, but otherwise looks good to go

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.test.tsx
DafyddLlyr: Fixed in [e0dd5f8](https://github.com/theopensystemslab/planx-new/pull/3700/commits/e0dd5f821625e4d15d4b8b12f199925ab069cf22)

PR #3706, opened 2024-09-18
feat: Remove a User

RODO94:
[WIP]


editor.planx.uk/src/pages/FlowEditor/components/Team/tests/TeamMembers.archiveUser.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+import { setupTeamMembersScreen } from "./helpers/setupTeamMembersScreen";
+import { mockTeamMembersData } from "./mocks/mockTeamMembersData";
+import { mockPlatformAdminUser } from "./mocks/mockUsers";
+
+vi.mock("pages/FlowEditor/components/Team/queries/updateUser.tsx", () => ({
+  updateTeamMember: vi.fn().mockResolvedValue({
+    id: 1,
+  }),
+}));
```
RODO94, line 13: Stuck here, need help

editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
```diff
@@ -130,57 +150,64 @@ export const MembersTable = ({
...
-                        setShowAddModal(true);
-                      }}
-                    >
-                      Add a new editor
-                    </AddButton>
-                  </TableCell>
-                </TableRow>
-              </Permission.IsPlatformAdmin>
+              <TableRow>
+                <TableCell colSpan={3}>
```
ianjon3s, line 189: nit: Changing the colSpan to 5 here will line up the horizontal borders

RODO94:
The lock files have a huge amount of changes where `'` become `"` - is there a way I can stop this?

editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
```diff
@@ -130,57 +150,64 @@ export const MembersTable = ({
...
+                  </AddButton>
+                </TableCell>
+              </TableRow>
             )}
           </TableBody>
         </Table>
       </TableContainer>
       {showAddModal && (
-        <EditorUpsertModal
+        <SettingsModal
```
RODO94, line 206: `SettingsModal` has been used here to separate out the logic of what content goes inside a `<Dialog>` - felt right considering the 3 different options which contain two different types of dialog content, but all reusing similar props

RODO94:
[Remove user in Editor](https://trello.com/c/0jbcXque/3012-remove-user-in-editor)

editor.planx.uk/src/pages/FlowEditor/components/Team/types.ts
```diff
@@ -15,15 +16,17 @@ export interface AddNewEditorModalProps {
   setShowModal: React.Dispatch<SetStateAction<boolean>>;
 }
 export interface AddNewEditorFormValues {
-  email: string;
+  email: string | null;
   firstName: string;
   lastName: string;
 }
```
DafyddLlyr, line 22: I understand the intention, but this seems like the wrong approach - in order to add a user, we do need an email. It seems incorrect that and interface called `AddNewEditorFormValues` could have a `null` email address.

editor.planx.uk/src/pages/FlowEditor/lib/store/team.ts
```diff
@@ -157,4 +158,12 @@ export const teamStore: StateCreator<
   setTeamMembers: async (teamMembers: TeamMember[]) => {
     set(() => ({ teamMembers }));
   },
+  deleteUser: async (userId: number) => {
+    const { $client } = get();
+    const response = await $client.user.delete(userId);
+    if (!response) {
+      throw new Error("Unable to remove user");
+    }
```
DafyddLlyr, line 166: It would be more valuable to catch the actual error here - in isolation if Airbrake had "Unable to remove user" it wouldn't be that easy to understand the root cause.

editor.planx.uk/src/pages/FlowEditor/components/Team/types.ts
```diff
@@ -1,8 +1,9 @@
 import { Role, User } from "@opensystemslab/planx-core/types";
 import React, { SetStateAction } from "react";
 
-export type TeamMember = Omit<User, "teams" | "isPlatformAdmin"> & {
+export type TeamMember = Omit<User, "teams" | "isPlatformAdmin" | "email"> & {
```
DafyddLlyr, line 4: As with the previous comment on `AddNewEditorFormValues` - we're unintentionally looseing our types here (both sematically, and in terms of type-safety).

Is might be the case that an `ArchivedTeamMember` is what you're intending to create / model?

```ts
type TeamMember = ActiveTeamMember | ArchivedTeamMember
```

editor.planx.uk/src/pages/FlowEditor/components/Team/components/SettingsModal.tsx
```diff
@@ -0,0 +1,46 @@
+import Dialog from "@mui/material/Dialog";
+import React from "react";
+
+import { EditorModalProps } from "../types";
+import { EditorUpsertModal } from "./EditorUpsertModal";
+import { RemoveUserModal } from "./RemoveUserModal";
+
+export const SettingsModal = ({
```
DafyddLlyr, line 8: I'm not sure this is the right abstraction here (though I agree on your instincts to try and share the styles etc).

This looks like a generic modal that could be used anywhere (going by the component name and location within the codebase), but it infact takes in a `EditorModalProps` and needs the user related  ``data-testid={`dialog-${actionType}-user`}``.

Here's what I think we should be aiming for - 
 - Shared modal styles
 - The ability to pass a `children` prop which is of type `EditorUpsertModal`
 
 What this would achieve is -
  - A truly generic and visually consistent `SettingsModal`
  - The ability to nest anything within in (user details, something to do wih settings, or themes etc)
  
There's a few ways to achieve this, but one of the most powerful and simple is likely via the MUI theme. It's likely there's opportunity to tidy up a few `Dialog` components by this method.

editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditorUpsertModal.tsx
```diff
@@ -58,12 +56,15 @@ export const EditorUpsertModal = ({
   const handleSubmitToAddNewUser = async () => {
     const { teamId, teamSlug } = useStore.getState();
 
+    if (formik.values.email === null) {
+      return;
+    }
```
DafyddLlyr, line 61: This looks like a type guard, which I think we can solve via a more semantically meaningful method (see previous comment).

editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditorUpsertModal.tsx
DafyddLlyr, line 1: Please see comment about the `SettingsModal` - if we're changing the responsibility of this component (no longer a modal, but just the content), we should try and give it a meaningful component name in line with this change.

DafyddLlyr (changes requested):
A few small nits and comments on PR, please see below. I think the main area to focus on is how the refactor of the `SettingsModal` is handled.

Pizza working as expected, but I hit a bug related to the cache of existing users persisting - I think we encountered and resolved something similar before?

https://github.com/user-attachments/assets/7211cc16-ae0f-4a00-8a51-f2151e150178

> The lock files have a huge amount of changes where ' become " - is there a way I can stop this?

There's some linting inconsistencies across our devices - clearly husky or prettier is not configured correctly. Let's try and solve this IRL this week 👍

RODO94:
@DafyddLlyr  for the bug, to solve the caching issue, would I need to trigger a re-rendering of the component? Would there be a reason this doesn't happen when I set state in the global store here: 

```ts
export const optimisticallyUpdateExistingMember = async (
  values: UpdateEditorFormValues,
  userId: number
) => {
  const existingMembers = useStore.getState().teamMembers;
  const updatedMembers: TeamMember[] = existingMembers.map((member) => {
    if (member.id === userId) {
      return { ...values, id: userId, role: member.role };
    }
    return member;
  });
  await useStore.getState().setTeamMembers(updatedMembers);
};

```

See the bug is present when editing a users details as well, but not when adding, so I wonder why this is performing differently. Does Zustand not see a difference between the old array of TeamMembers and the new?

api.planx.uk/pnpm-lock.yaml
DafyddLlyr, line 1: Let's make sure we fix this `'` vs `"` issue before merging 👌

editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx
```diff
@@ -67,7 +67,7 @@ export const TeamMembers = () => {
         />
       </SettingsSection>
       {archivedMembers.length > 0 && (
-        <SettingsSection>
+        <SettingsSection testId="archived-members">
```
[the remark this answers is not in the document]
RODO94: Will do follow on PR to pick up this change

editor.planx.uk/src/pages/FlowEditor/lib/store/team.ts
RODO94: Will follow up with TryCatch

editor.planx.uk/src/pages/FlowEditor/components/Team/components/SettingsModal.tsx
RODO94: Will do Styled Component in this PR and do a follow up refactor PR with using the theme.ts style overrides

editor.planx.uk/src/ui/editor/SettingsSection.tsx
```diff
@@ -21,26 +14,16 @@ const Root = styled(Box, {
...
-  background?: boolean;
-  testId?: string;
-}) {
-  return (
-    <Root data-testid={testId} background={background}>
-      {children}
-    </Root>
-  );
+export default function SettingsSection(
+  props: { children: ReactNode } & Partial<BoxProps>,
```
RODO94, line 26: @DafyddLlyr change that we discussed on our call but didn't resolve, I've now got it all working and it is showing up in the DOM as expected.

editor.planx.uk/src/pages/FlowEditor/components/Team/styles.ts
```diff
@@ -16,3 +17,14 @@ export const StyledTableRow = styled(TableRow)(({ theme }) => ({
     background: theme.palette.background.paper,
   },
 }));
+
+export const SettingsDialog = styled(Dialog)(({ theme }) => ({
```
RODO94, line 21: This works for now, but a future PR should refactor this into a global Editor style override in the `theme.ts` file

editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
```diff
@@ -130,59 +163,68 @@ export const MembersTable = ({
...
+                        data-testid={`edit-button-${member.id}`}
                       >
                         Edit
                       </EditUserButton>
-                    </TableCell>
+                    )}
+                  </Permission.IsPlatformAdmin>
+                </TableCell>
+                <TableCell>
+                  <Permission.IsPlatformAdmin>
```
RODO94, line 181: Permission level moved inside `<TableCell>` to ensure consistent UI across different permission levels

editor.planx.uk/src/ui/editor/SettingsSection.tsx
```diff
@@ -3,14 +3,7 @@ import { styled } from "@mui/material/styles";
...
 
-interface RootProps extends BoxProps {
-  background?: boolean;
-  "data-testid"?: string;
-}
-
-const Root = styled(Box, {
-  shouldForwardProp: (prop) => prop !== "background",
-})<RootProps>(({ background, theme }) => ({
+const Root = styled(Box)(({ bgcolor, theme }) => ({
```
RODO94: I had to revert `bgcolor` for `background` since this was acting as a boolean to conditionally style the `<SettingsSection>` component

editor.planx.uk/src/ui/editor/SettingsSection.tsx
```diff
@@ -3,14 +3,9 @@ import { styled } from "@mui/material/styles";
...
 
-interface RootProps extends BoxProps {
-  background?: boolean;
-  "data-testid"?: string;
-}
-
 const Root = styled(Box, {
   shouldForwardProp: (prop) => prop !== "background",
-})<RootProps>(({ background, theme }) => ({
+})<BoxProps & { background?: boolean }>(({ background, theme }) => ({
```
RODO94, line 8: Had to revert `bgcolor` since `background` was a boolean to conditionally style the `ServiceSettings` component

RODO94: I found this when running tests and seeing the other places this component was used

DafyddLlyr (approved):
Looking good - thanks for those changes 👍

PR #3711, opened 2024-09-19
fix: add flow name to S3 Slack notifications & webhook message

jessicamcinchak:
Barnet is expanding their use of the S3/Power Automate webhook "Send" method to more services, but our logging for this method isn't currently easily filter-able on flow name.

**Two quick changes:** 
- Add `service` (aka flow name) property to the message we send to the webhook
- Access that `service` name again (from the audit table record) for the Slack message so it's more consistent with other notifications like send to email 
  - Before: ` New S3 + Power Automate submission 3ec46727-88b4-4f39-bb95-a89ee33f0b52 [barnet] [Exempt]`
  - After:  ` New S3 + Power Automate submission "Apply for a Lawful Development Certificate" 3ec46727-88b4-4f39-bb95-a89ee33f0b52 [barnet] [Exempt]`

**Testing:**
I've tested locally & confirmed with Kev. Please do _not_ send an S3 test submission on this pizza because it's just a single environment on Barnet's side (so we want to track & flag any local or pizza session IDs to them).

Here's local confirmation:
Screenshot from 2024-09-19 10-43-56

### DIFF
--- a/api.planx.uk/modules/send/s3/index.ts
+++ b/api.planx.uk/modules/send/s3/index.ts
@@ -49,6 +49,7 @@ const sendToS3 = async (req: Request, res: Response, next: NextFunction) => {
 
     const session = await $api.session.find(sessionId);
     const passport = session?.data?.passport as Passport;
+    const flowName = session?.flow?.name;
 
     // Generate the ODP Schema JSON, skipping validation if not a supported application type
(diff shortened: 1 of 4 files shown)


DafyddLlyr (approved):
Nice simple solution 🙂

PR #3713, opened 2024-09-19
fix: hide "Add a new editor" button from non-admin users

jessicamcinchak:
It's the second planning meeting in a row where Ian G has asked about / been a bit confused by adding and editing users and I think it's because he can see the add button as a `teamEditor`, but doesn't have permission to actually use it. 

This change hides the "Add a new editor" from non-`platformAdmins` similar to the "Edit" button. Database-level permissions already look correct (eg only `platformAdmin` role has "insert" permission on `team_members` table).

We've thrown this previous error: https://opensystemslab.slack.com/archives/C01E3AC0C03/p1725542317058179
Screenshot from 2024-09-19 15-38-11

Good reminder to ourselves to trigger our own `platformAdmin` status "off" sometimes when testing/code reviewing pizzas!

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
@@ -148,7 +148,7 @@ export const MembersTable = ({
               </StyledTableRow>
             ))}
             {showAddMemberButton && (
-              <Permission.IsNotPlatformAdmin>
+              <Permission.IsPlatformAdmin>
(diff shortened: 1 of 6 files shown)


jamdelion (approved):
Clear and concise! ⭐ 
Thanks for sorting this, it was my bad for not covering this in the original ticket! I think I got a bit confused about the roles available/who was supposed to see what... 🙈 It's defo good advice to toggle off platform admin when testing pizzas, thanks 🍕

DafyddLlyr (approved):

PR #3716, opened 2024-09-20
feat: Add tags to Node data (take 2)

DafyddLlyr:
## What does this PR do?
- Adds `NodeTags` to `Store.Node["data"]`
- Final batch of TS updates in order to use NodeTags across all component types (I believe!)

Most of this is type updates, the two actual changes are highlighted below - 

https://github.com/theopensystemslab/planx-new/blob/0a1cbc68da8a5f3867b41756083aecd2dda7bd31/editor.planx.uk/src/pages/FlowEditor/lib/store/index.ts#L39-L41

https://github.com/theopensystemslab/planx-new/blob/0a1cbc68da8a5f3867b41756083aecd2dda7bd31/editor.planx.uk/src/pages/Preview/Node.tsx#L67-L71

## Next steps...
- Add form field for tags to each component (hopefully via `EditorProps<T>` or similar)
- Placeholder UI to add tag in-Editor

PR #3726, opened 2024-09-23
chore: Create model file for `Question`

DafyddLlyr:
## What?
 - Moves `Question` to its own `model.ts` file

## Why?
https://github.com/theopensystemslab/planx-new/pull/3716 has slightly descended into a large messy type fixing PR which I was trying to avoid. I'm splitting it up into smaller, more atomic, PRs in order to better highlight and track the actual important changes.


RODO94 (approved):
Looks good to me

PR #3734, opened 2024-09-27
fix: Remove additional Editor scroll

DafyddLlyr:
Small CSS fix I've been meaning to take a look at for a while.

When the "test environment" banner is visible, it takes the FlowEditor over 100vh which we don't want. This PR fixes that 👍 

## Before
https://github.com/user-attachments/assets/2707add5-ce5d-4f88-bde6-2a72b241ff45

## After
https://github.com/user-attachments/assets/5a27456b-9923-40c3-a87e-382296aa1d28

### DIFF
--- a/editor.planx.uk/.eslintrc
+++ b/editor.planx.uk/.eslintrc
@@ -5,15 +5,13 @@
     "react-hooks",
     "simple-import-sort",
     "jsx-a11y",
-    "testing-library",
-    "jest"
+    "testing-library"
   ],
   "extends": [
     "eslint:recommended",
     "plugin:jsx-a11y/recommended",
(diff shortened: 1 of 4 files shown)


editor.planx.uk/.eslintrc
DafyddLlyr, line 1: I was hitting some linting issues here. Next PR I'll look at this a bit deeper and see if I can find vitest specific replacements here.

DafyddLlyr: Please see https://github.com/theopensystemslab/planx-new/pull/3735

jessicamcinchak (approved):

PR #3731, opened 2024-09-26
chore: CVE-2024-47068

DafyddLlyr:
Fixes - 
 - https://github.com/theopensystemslab/planx-new/security/dependabot/131
 - https://github.com/theopensystemslab/planx-new/security/dependabot/132
 - https://github.com/theopensystemslab/planx-new/security/dependabot/133

PR #3735, opened 2024-09-27
chore: Replace jest linting with vitest linting

DafyddLlyr:
## What does this PR do?
- Replaces legacy jest eslint plugins with a vitest specific plugin
- Makes the few requisite updates to meet the new linting rules
- Add `pnpm lint:error` script which just identifies errors, ignoring warnings

I think that https://github.com/theopensystemslab/planx-new/pull/3731 introduced this issue which dropped jest as a peer dependency.

### DIFF
--- a/editor.planx.uk/.eslintrc
+++ b/editor.planx.uk/.eslintrc
@@ -5,13 +5,15 @@
     "react-hooks",
     "simple-import-sort",
     "jsx-a11y",
-    "testing-library"
+    "testing-library",
+    "@vitest"
   ],
   "extends": [
     "eslint:recommended",
     "plugin:jsx-a11y/recommended",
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Settings/tests/ServiceSettings/PublicLinks.test.tsx
DafyddLlyr, line 1: Heavy use of helper functions in this file which wrap `expect()` functions, so `eslint-disable-next-line @vitest/expect-expect` is needed throughout. Adding this once to the whole file, instead of for each test, seemed like a riskier approach.

jessicamcinchak (approved):

PR #3740, opened 2024-09-30
chore: optimise planning constraints requests

jessicamcinchak:
Joann at Gloucester is still reporting partial GIS queries: https://opendigitalplanning.slack.com/archives/C05NQJK5S9L/p1726495149134779

**Double-checking a few of our configs here to see if there's anything we've missed:**
- [x] Can we filter Planning Data metadata response fields similar to entity? 
  - `/dataset` endpoint does _not_ currently support include/exclude field params so nothing we can do here https://www.planning.data.gov.uk/docs
- [x] Can we optimise `/roads` OGC query?
  - Confirmed this is working as expected, there's not much way around a slow string-matching USRN query! Shared details with Steve at Planning Data about if/how roads might be a candidate for adding to the Planning Data platform in the future which could simplify a bit (eg consistent performance/caching/uptime, but still need to handle non-geospatial query method!)
- [x] Is loading logic correct throughout? Are we properly handling scenarios where one of two requests errors (ui feedback & passport vars)?
  - Again, properly looked this over & we're doing exactly what we want! If one of the two requests fails, you only see results from the other and any "failed" variables are not added to passport at all, therefore prompting users to manually see questions

**Only changes:**
- Split up components in `Public.tsx` & renamed `PlanningConstraintsContent` &rarr; `Presentational` as this is a more established pattern elsewhere among frontend components


editor.planx.uk/src/@planx/components/PlanningConstraints/Presentational.tsx
```diff
@@ -0,0 +1,167 @@
...
+    setInaccurateConstraints,
+  } = props;
+
+  const error = constraints.error || undefined;
+  const showError = error || !Object.values(constraints)?.length;
+  if (showError) return <ConstraintsFetchError error={error} {...props} />;
+
+  const positiveConstraints = Object.values(constraints).filter(
+    (v: Constraint, _i) => v.text && v.value,
+  );
```
RODO94, line 50: question: is `_i` here used for readability?

RODO94 (commented):
@jessicamcinchak trying to track and follow the code here. Are most changes on moving the jsx content to `Presentational.tsx` file from the `Public.tsx` file?

I see big diff around the `handleSubmit()` but it looks mostly like it has only changed position in the code rather than any logical changes?

Should I be testing any logic?

jessicamcinchak:
@RODO94 sorry for confusion - this ended up being a pure quick refactor for readability/smaller component files, no logic changes

editor.planx.uk/src/@planx/components/PlanningConstraints/Presentational.tsx
jessicamcinchak: I can remove this - I think the unused index (aka `_i`) is an old remnant from when we used to filter positive & negative constraints _and then also_ only show the top three of each/hide the rest or something like that

RODO94 (approved):
@jessicamcinchak thanks for clarifying, looks good to me. Thanks for picking up all the comms on this with Planning Data

PR #3741, opened 2024-09-30
feat(search): Add word break after special characters in search cards

DafyddLlyr:
## What does this PR do?
Adds a `<wbr/>` element ([docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/wbr)) after a headline character if it's a special character.

## Why?
The headline element is split into a series of spans (one per character), so applying a regular word break css rule to the parent won't work. By inserting a `<wbr/>` each time we hit a non-alphanumeric character, we allow the browser to break where it sees fit.

| Before | After |
|--------|--------|
|image|image|

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/Headline.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/Headline.tsx
@@ -15,17 +15,21 @@ export const Headline: React.FC<Props> = ({ text, matchIndices, variant }) => {
   return (
     <>
       {text.split("").map((char, index) => (
-        <Typography
-          component="span"
-          variant={variant}
-          key={`headline-character-${index}`}
-          sx={(theme) => ({
-            fontWeight: isHighlighted(index) ? FONT_WEIGHT_BOLD : "regular",
-            fontSize: theme.typography.body2.fontSize,
-          })}
-        >
-          {char}
-        </Typography>
+        <>
+          <Typography
+            component="span"
+            variant={variant}
+            key={`headline-character-${index}`}
+            sx={(theme) => ({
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/Headline.tsx
```diff
@@ -15,17 +15,21 @@ export const Headline: React.FC<Props> = ({ text, matchIndices, variant }) => {
...
+            key={`headline-character-${index}`}
+            sx={(theme) => ({
+              fontWeight: isHighlighted(index) ? FONT_WEIGHT_BOLD : "regular",
+              fontSize: theme.typography.body2.fontSize,
+            })}
+          >
+            {char}
+          </Typography>
+          {/* Add wordbreak after special characters */}
+          {char.match(/\W/) && <wbr />}
```
jessicamcinchak, line 31: I don't fully understand why this is _outside_ of the `<Typography>` block and only within the fragment? Don't we want all characters to share the styles above (eg indexed/matching characters _after_ a special character should still have bold highlighting?)

jessicamcinchak (approved):
Fundamentally works/looks better but one quick question about the code here!

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/Headline.tsx
DafyddLlyr: Good question! This is a little unexpected as the headline breaks each word into characters and inserts a "word" break after them. The `<wbr/>` is just a non-visual element so that it splits in the right place. 

The reason there's only one word highlighted is that the search is configured to just return the first match from a string, not all.

DafyddLlyr: image

PR #3747, opened 2024-10-01
feat: Handle editor sidebar visibility toggle on sidebar

ianjon3s:
## What does this PR do?

Moves toggle for show/hiding the editor sidebar to be placed on the sidebar itself.

Also stores the toggle position in localStorage so that the position is maintained when opening graph nodes or switching screens.

Example:
https://3747.planx.pizza/barking-and-dagenham/retrofit-guide

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
@@ -11,7 +11,7 @@ import Link from "@mui/material/Link";
 import { styled } from "@mui/material/styles";
 import Tabs from "@mui/material/Tabs";
 import Tooltip from "@mui/material/Tooltip";
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { rootFlowPath } from "routes/utils";
 import Permission from "ui/editor/Permission";
 import Reset from "ui/icons/Reset";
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
```diff
@@ -23,19 +26,24 @@ import StyledTab from "./StyledTab";
 
 type SidebarTabs = "PreviewBrowser" | "History" | "Search" | "Console";
 
-const Root = styled(Box)(({ theme }) => ({
+const SIDEBAR_WIDTH = 500;
+const SIDEBAR_WIDTH_MINIMISED = 20;
```
jessicamcinchak, line 30: nit: should these simply be set directly as strings to avoid the extra templating below?

jessicamcinchak, line 30: nit: should these be set directly as strings to avoid the extra templating step below?

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
```diff
@@ -118,97 +167,106 @@ const Sidebar: React.FC = React.memo(() => {
   };
 
   return (
-    <Root>
-      <Header>
-        <Box width="100%" display="flex">
-          <input type="text" disabled value={urls.preview} />
+    <Root isMinimised={isSidebarMinimised}>
```
jessicamcinchak, line 170: :+1: nice & clear !

jessicamcinchak (approved):
Works great! Feels a lot more intuitive than old header icon :raised_hands:

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
ianjon3s: Great spot, updated 👍

PR #3749, opened 2024-10-01
feat: Use `BaseNodeData` to add `tags` property to all component types

DafyddLlyr:
## What does this PR do?
 - Replaces the `MoreInformation` interface with `BaseNodeData`
   - Most of this PR is then just type changes that follow along
   - Removes duplicate properties (where `MoreInformation` was previously not being used)

## Why?
 - Part of the work to add tags to nodes
 - This new interface serves as a base for all Node types - meaning we always get both `NodeTags` and `MoreInformation`
 - Hopefully this pattern can be followed up fairly shortly to also pull in and standardise other common fields (`title`, `description`, maybe `fn`)

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/schema.ts
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/schema.ts
@@ -10,7 +10,7 @@ import {
 } from "yup";
 
 import { FileUploadSlot } from "../FileUpload/model";
-import { MoreInformation } from "../shared";
+import { BaseNodeData, MoreInformation } from "../shared";
 import {
   checkIfConditionalRule,
   Condition,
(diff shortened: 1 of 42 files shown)


editor.planx.uk/src/@planx/components/shared/index.ts
DafyddLlyr, line 1: This is the important file in this PR! 👇

editor.planx.uk/src/@planx/components/FileUploadAndLabel/schema.ts
```diff
@@ -31,6 +31,10 @@ const moreInformationSchema: SchemaOf<MoreInformation> = object({
   definitionImg: string(),
 });
 
+const baseNodeDataSchema: SchemaOf<BaseNodeData> = object({
+  tags: array(mixed().oneOf(["placeholder"])),
```
jessicamcinchak, line 35: nit: surprised to see `["placeholder"]` hardcoded specifically in this component file - can we define & export/import list of available tags someplace once?

jessicamcinchak (approved):
One small comment, but overall looks good to me :+1:

editor.planx.uk/src/@planx/components/FileUploadAndLabel/schema.ts
DafyddLlyr: Good catch thanks - now added 👍

PR #3750, opened 2024-10-01
feat: allow team editors to download a zip of application data from the editor submissions log

jessicamcinchak:
Folks using the existing Uniform connector have been having issues accessing application files in their DMS and more often requesting that a zip be sent to them via Slack, etc - this should make application data easier to self-serve from the editor ! See this thread for more context: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1727791569016769

**Changes:**
- Shows a download button in the submissions log (_if_ it's a sucessful submission within last 28 days, you have team editor privileges to this team, a submission email has been set in team settings)
  - This button downloads the same zip that folks access if using "Send to email" - it includes standard PlanX documents (application.csv, application.json, overview.html, redactedOverview.html, locationPlan.html, locationPlan.geojson) as well as any user-uploaded files

**Test flow:**
- https://3750.planx.pizza/testing/submission-log-download-test/submissions-log

Screenshot from 2024-10-01 20-23-48

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
@@ -1,3 +1,4 @@
+import CloudDownload from "@mui/icons-material/CloudDownload";
 import KeyboardArrowDown from "@mui/icons-material/KeyboardArrowDown";
 import KeyboardArrowUp from "@mui/icons-material/KeyboardArrowUp";
 import Payment from "@mui/icons-material/Payment";
@@ -17,6 +18,7 @@ import Typography from "@mui/material/Typography";
 import DelayedLoadingIndicator from "components/DelayedLoadingIndicator";
 import ErrorFallback from "components/Error/ErrorFallback";
 import { format } from "date-fns";
+import { useStore } from "pages/FlowEditor/lib/store";
 import React, { useState } from "react";
 import ErrorSummary from "ui/shared/ErrorSummary";
 
@@ -77,7 +79,8 @@ const EventsLog: React.FC<GetSubmissionsResponse> = ({
             <TableCell sx={{ width: 350 }}>
               <strong>Session ID</strong>
             </TableCell>
-            <TableCell sx={{ width: 60 }}></TableCell>
+            <TableCell sx={{ width: 50 }}></TableCell>
+            <TableCell sx={{ width: 50 }}></TableCell>
           </TableRow>
         </TableHead>
         <TableBody>
@@ -92,6 +95,16 @@ const EventsLog: React.FC<GetSubmissionsResponse> = ({
 
 const CollapsibleRow: React.FC<Submission> = (submission) => {
   const [open, setOpen] = useState<boolean>(false);
+
+  const [teamSlug, canUserEditTeam, submissionEmail] = useStore((state) => [
+    state.teamSlug,
+    state.canUserEditTeam,
+    state.teamSettings?.submissionEmail, // TODO teamSettings are undefined on submissions log route
+  ]);
+
+  // TODO and submission event created within last 30 days ??
+  const showDownloadButton =
+    canUserEditTeam(teamSlug) && submission.status === "Success";
 
   return (
     <React.Fragment key={`${submission.eventId}-${submission.createdAt}`}>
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/pages/FlowEditor/lib/store/team.ts
```diff
@@ -108,6 +108,19 @@ export const teamStore: StateCreator<
...
+              referenceCode: reference_code
+              helpEmail: help_email
+              helpPhone: help_phone
+              helpOpeningHours: help_opening_hours
+              emailReplyToId: email_reply_to_id
+              homepage: homepage
+              externalPlanningSiteName: external_planning_site_name
+              externalPlanningSiteUrl: external_planning_site_url
+              submissionEmail: submission_email
+            }
```
jessicamcinchak, line 123: In order to re-use the existing `/download-application-files/:sessionId` endpoint, we need team settings like `submissionEmail` to be available in the editor when the team store is initialised - is this okay? Am I forgetting any specific reasons that settings were omitted from this query and _not_ initialised?

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
```diff
@@ -121,6 +139,24 @@ const CollapsibleRow: React.FC<Submission> = (submission) => {
...
+            <IconButton
+              aria-label="download application"
+              size="small"
+              onClick={() => {
+                const zipUrl = `${
+                  import.meta.env.VITE_APP_API_URL
+                }/download-application-files/${
+                  submission.sessionId
+                }?localAuthority=${teamSlug}&email=${submissionEmail}`;
+                window.open(zipUrl, "_blank");
```
jessicamcinchak, line 153: This is really simple! Is `window.open()` actually _too_ simple / let's get council folks to test on staging to confirm that their council browser settings won't block this? 

We also know that we want to "harden" `/download-application-files/:sessionId` endpoint in the near future ([this Trello ticket](https://trello.com/c/EeJb0QVC/2614-harden-download-application-files-endpoint)) - so we'll ideally want a solution that works for both places here so that we can ensure send to email & submission log zips are totally consistent for end uers ! I've attached this PR to the Trello ticket for future reference.

jessicamcinchak:
[Harden `/download-application-files` endpoint](https://trello.com/c/EeJb0QVC/2614-harden-download-application-files-endpoint)

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
```diff
@@ -93,6 +97,20 @@ const EventsLog: React.FC<GetSubmissionsResponse> = ({
...
+    state.canUserEditTeam,
+    state.teamSettings?.submissionEmail,
+  ]);
+
+  // Only show an application download button if certain conditions are met
+  const submissionDataExpirationDate = addDays(new Date(), DAYS_UNTIL_EXPIRY);
+  const showDownloadButton =
+    canUserEditTeam(teamSlug) &&
+    submission.status === "Success" &&
+    submissionEmail &&
```
jessicamcinchak, line 111: It's admittedly a bit of a weird requirement that `submissionEmail` needs to be set to be able to download application data that was originally sent to any back-office system (because `submissionEmail` is "send to email" specific) - but now that this is self-service in the editor, I think it's ultimately okay and won't be a blocker?

This MVP re-uses the `/download-application-files/:sessionId` endpoint, but as other comments here also say we know we want to change/"harden" this soon so will be able to theoretically drop this requirement in next iteration!

ianjon3s:
Possible quick usability win here: wrapping the download icon in a tooltip. I'll branch off to add this.

editor.planx.uk/src/pages/FlowEditor/lib/store/team.ts
RODO94: I think this is cool, I remember I had a bit of trouble with the query, but think that was due to an issue on timing of changes with submissionEmail than anything code related that would stop this happening. All the settings should be defined in the database and able to be read, nothing gets generated after this... unless you edit a setting.

RODO94 (approved):
Works as expected for me, although it is hard for me to review the quality of the files, only testing the initialising of the team settings, and the use of the button to trigger a download.

@jessicamcinchak do you think that's something I should do before approving or it's unrelated to the PR?

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
DafyddLlyr: Yeah it's a bit odd as it stands but I totally get it!

I think a better approach would be to hit the API and use the user's JWT permissions to control access as opposed to the fact they have access to the `submission_email` column. 

The reason for this is that `team_settings` doesn't have access controls between teams, but the `lowcal_sessions` table could / should.

What I think this would look like - 
 - Set up permissions on `lowcal_sessions` table so that `teamEditors` can access sessions from their team
 - Make a new route, with middleware to check it's an authenticated request, but use the same controller (might need to look at which client is making the request?)
 - Hit the new route here

```ts
router.get("/download-submission-files/:sessionId", useTeamEditorAuth, downloadApplicationFiles);
```

This works fine for now, but when we revisit this I think it's worth tightening this up a little 👍

DafyddLlyr (approved):

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx
DafyddLlyr: Only one way to find out! This is the approach I would have gone with initially also 👍

jessicamcinchak:
[Add a download button from editor submissions log for easier file access](https://trello.com/c/Pa7sU62p/3072-add-a-download-button-from-editor-submissions-log-for-easier-file-access)

PR #3751, opened 2024-10-02
fix: clear MapAndLabel breadcrumbs if FindProperty address has changed

jessicamcinchak:
Fixes a buggy observation in user testing ! When navigating "back" or "changing" a MapAndLabel component, we only want to retain your previous answer if FindProperty has _not_ also changed because MapAndLabel relies on the geographic position of the site address.

**New behavior:**
- If you change your site address, you should now also be re-prompted to fill out MapAndLabel fresh (similar to DrawBoundary, PlanningConstraints, etc), without any previous answers retained
- If you're coming "back" to MapAndLabel _without_ changing your site address, your answers should still be retained


jessicamcinchak:

RODO94 (approved):
nit: I had an issue in a weird edge case where if you don't actually change the address after clicking to change it, and move to the next tab, the data is maintained but the tab isn't selected. 

Screenshot 2024-10-02 at 15 57 04

Don't expect it to happen, and my instinct was to click next or click on the tab which resolved it, just initially confusing

jessicamcinchak:
@RODO94 thanks for reviewing & description of this edge casey bug - think minimised tab should be addressed by #3753 ! Will get this merged to `main` alongside other fixes, then see if still recreatable :+1:

PR #3752, opened 2024-10-02
feat: Add Character Count to Text Input

RODO94:
## What does this PR do?

This PR adds a new component `<CharacterCounter>` and conditionally renders it within the `<Input>` component. It will only render for `text` inputs which have a `long` or `extraLong` type added.

`Aria-Live` regions was useful but tricky to implement with React and other components re-rendering and causing a lot of noise. I tried to reduce this by using `aria-hidden` at different points and different booleans to trigger these as and when they are required to be announced. Like below:

```diff
 <Typography
        id={"character-hint"}
        sx={visuallyHidden}
++        aria-hidden={"true"}
      >
        {`You can enter up to ${props.characterLimit} characters`}
      </Typography>
      
      ....
      
            <Typography
        aria-live="polite"
        aria-atomic="true"
        sx={visuallyHidden}
        data-testid="screen-reader-count"
      >
++        {showReaderCount && screenReaderCountText}
      </Typography>
```

This seemed to work well during my testing, but I recognise different screen readers may handle stuff differently and I'm probably not the ideal user for screen readers (bit of a n00b). 

On the positioning of the logic around `TEXT_LIMITS[]` and `TextInputType` - I have split it between the `<CharacterCount>` and the related text input component. I have simplified the props being passed into the `<CharacterCount>` so now it takes in `data.type || props.type` and type checking is done within the `<CharacterCount>`. 

I have kept the logic to show or hide `<CharacterCount>` within the Text Input component itself, so it is clearer what the component does and why things show or don't show. 

In order to maintain type safety, I have used the below function to manage the `TextInputType` enums. 

```ts
export function isLongTextType(
  type: TextInputType,
): type is TextInputType.Long | TextInputType.ExtraLong {
  return type === TextInputType.Long || type === TextInputType.ExtraLong;
}

...

  function getLongTextLimit(type: TextInputType): number {
    if (isLongTextType(type)) {
      return TEXT_LIMITS[type];
    } else {
      return 0;
    }
  }
```
This ensures we can properly use `TEXT_LIMITS[type]`.

There's scope for improving this with the below suggestion from @DafyddLlyr 

```ts
export const TEXT_LIMITS = {
  [TextInputType.Short]: {
    showCharacterCount: false,
    limit: 150,
  },
  [TextInputType.Long]: {
    showCharacterCount: true,
    limit: 250,
  },
  [TextInputType.ExtraLong]: {
    showCharacterCount: true,
    limit: 750,
  },
} as const;
```

And carrying this logic into the Character Count component. This would ensure you could just update `TEXT_LIMITS` once and using `TEXT_LIMITS[TextInputType.Short].showCharacterCount || .limit` to replace current logic. I tried this and ran into a couple errors, and it felt like the gain is low given how much time has already been put into it. It could definitely be a future PR if we wanted to start extending our text limits

Struggles I still have:

- [x] The screen reader sometimes states the character remaining count twice
- [x] The screen reader sometimes doesn't alway read out the whole passage when you stop typing
- [ ] The screen reader is re-reading the input label multiple times due to re-rendering

Managed to find improvements, but still found some screen reader functionality not working as I'd hoped. Current issue is the label being re-read multiple times, I have a potential solution but would rather this is a separate PR.

### DIFF
--- a/editor.planx.uk/src/@planx/components/TextInput/Public.test.tsx
+++ b/editor.planx.uk/src/@planx/components/TextInput/Public.test.tsx
@@ -8,6 +8,9 @@ import { axe } from "vitest-axe";
 import { ERROR_MESSAGE } from "../shared/constants";
 import { TextInputType } from "./model";
 import TextInput from "./Public";
+
+const twentyFiveCharacterTest = "25 characters for me.....";
+
 test("requires a value before being able to continue", async () => {
   const handleSubmit = vi.fn();
 
@@ -163,4 +166,55 @@ it("should change the role of the ErrorWrapper when an invalid input is given", 
 
   expect(errorWrapper).not.toBeEmptyDOMElement();
   expect(errorWrapper).toHaveAttribute("role", "alert");
+});
+
+test("character limit counter should appear for long text inputs", async () => {
+  setup(<TextInput title="hello" type={TextInputType.Long} />);
+
+  const characterCounter = await screen.findByText(
+    "You have 250 characters remaining",
+  );
+
+  expect(characterCounter).toBeInTheDocument();
+});
+
+test("character limit counter should not appear for short text inputs", async () => {
+  setup(<TextInput title="hello" type={TextInputType.Short} />);
+
+  const characterCounter = screen.queryByText(
+    "You have 120 characters remaining",
+  );
+
+  expect(characterCounter).not.toBeInTheDocument();
+});
+
+test("character limit counter should change when typed", async () => {
+  const { user } = setup(<TextInput title="hello" type={TextInputType.Long} />);
+
+  const textArea = screen.getByRole("textbox", {
+    name: /hello you can enter up to 250 characters/i,
+  });
+
+  await user.type(textArea, twentyFiveCharacterTest);
+
+  const newCharacterCounter = await screen.findByText(
+    "You have 225 characters remaining",
+  );
+
+  expect(newCharacterCounter).toBeInTheDocument();
+});
+
+test("character limit counter shows error state when over limit", async () => {
+  const { user } = setup(<TextInput title="hello" type={TextInputType.Long} />);
+  const textArea = screen.getByRole("textbox", {
+    name: /hello you can enter up to 250 characters/i,
+  });
+
+  await user.type(textArea, `${twentyFiveCharacterTest.repeat(10)}.`);
+
+  const errorCharacterCounter = await screen.findByText(
+    "You have 1 characters too many",
+  );
+
+  expect(errorCharacterCounter).toHaveStyle({ color: "#D4351C" });
 });
--- a/editor.planx.uk/src/@planx/components/TextInput/Public.test.tsx
+++ b/editor.planx.uk/src/@planx/components/TextInput/Public.test.tsx
@@ -15,7 +15,7 @@ test("requires a value before being able to continue", async () => {
   const handleSubmit = vi.fn();
 
   const { user } = setup(
-    <TextInput title="hello" handleSubmit={handleSubmit} />,
+    <TextInput title="hello" handleSubmit={handleSubmit} />
   );
 
   expect(screen.getByRole("heading")).toHaveTextContent("hello");
@@ -34,7 +34,7 @@ test("requires a valid email before being able to continue", async () => {
       title="hello"
       type={TextInputType.Email}
       handleSubmit={handleSubmit}
-    />,
+    />
   );
 
   expect(screen.getByRole("heading")).toHaveTextContent("hello");
@@ -59,7 +59,7 @@ test("recovers previously submitted text when clicking the back button", async (
           [nodeId]: "Previously submitted text",
         },
       }}
-    />,
+    />
   );
 
   await user.click(screen.getByTestId("continue-button"));
@@ -86,7 +86,7 @@ test("recovers previously submitted text when clicking the back button even if a
           "text-input-key": "Previously submitted text",
         },
       }}
-    />,
+    />
   );
 
   await user.click(screen.getByTestId("continue-button"));
@@ -116,7 +116,7 @@ examplePhoneNumbers.forEach((number) => {
         title="phone"
         type={TextInputType.Phone}
         handleSubmit={handleSubmit}
-      />,
+      />
     );
 
     fireEvent.change(screen.getByLabelText("phone"), {
@@ -131,7 +131,7 @@ examplePhoneNumbers.forEach((number) => {
 
 it("should not have any accessibility violations", async () => {
   const { container } = setup(
-    <TextInput title="phone" type={TextInputType.Phone} />,
+    <TextInput title="phone" type={TextInputType.Phone} />
   );
   const results = await axe(container);
   expect(results).toHaveNoViolations();
@@ -143,7 +143,7 @@ it("should always an empty error message element in the DOM", () => {
       title="Short Text"
       type={TextInputType.Short}
       id="testId"
-    ></TextInput>,
+    ></TextInput>
   );
   const errorMessage = screen.getByTestId(`${ERROR_MESSAGE}-testId`);
   expect(errorMessage).toBeEmptyDOMElement();
@@ -158,7 +158,7 @@ it("should change the role of the ErrorWrapper when an invalid input is given", 
       type={TextInputType.Short}
       handleSubmit={handleSubmit}
       id="testId"
-    ></TextInput>,
+    ></TextInput>
   );
 
   const [errorWrapper, ..._rest] = screen.getAllByTestId("error-wrapper");
@@ -171,19 +171,15 @@ it("should change the role of the ErrorWrapper when an invalid input is given", 
 test("character limit counter should appear for long text inputs", async () => {
   setup(<TextInput title="hello" type={TextInputType.Long} />);
 
-  const characterCounter = await screen.findByText(
-    "You have 250 characters remaining",
-  );
-
+  const characterCounter = await screen.findByTestId("screen-reader-count");
+  screen.logTestingPlaygroundURL();
   expect(characterCounter).toBeInTheDocument();
 });
 
 test("character limit counter should not appear for short text inputs", async () => {
   setup(<TextInput title="hello" type={TextInputType.Short} />);
-
-  const characterCounter = screen.queryByText(
-    "You have 120 characters remaining",
-  );
+  screen.logTestingPlaygroundURL();
+  const characterCounter = screen.queryByTestId("screen-reader-count");
 
   expect(characterCounter).not.toBeInTheDocument();
 });
@@ -192,13 +188,13 @@ test("character limit counter should change when typed", async () => {
   const { user } = setup(<TextInput title="hello" type={TextInputType.Long} />);
 
   const textArea = screen.getByRole("textbox", {
-    name: /hello You have 0 characters remaining/i,
+    name: /hello/i,
   });
 
   await user.type(textArea, twentyFiveCharacterTest);
 
   const newCharacterCounter = await screen.findByText(
-    "You have 225 characters remaining",
+    "You have 225 characters remaining"
   );
 
   expect(newCharacterCounter).toBeInTheDocument();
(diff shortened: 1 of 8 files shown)


editor.planx.uk/src/@planx/components/TextInput/Public.test.tsx
```diff
@@ -8,6 +8,9 @@ import { axe } from "vitest-axe";
 import { ERROR_MESSAGE } from "../shared/constants";
 import { TextInputType } from "./model";
 import TextInput from "./Public";
+
+const twentyFiveCharacterTest = "25 characters for me.....";
```
RODO94, line 12: Felt easier cause I could use  a `.repeat()` to trigger the error on the textarea

editor.planx.uk/src/@planx/components/TextInput/Public.test.tsx
```diff
@@ -164,3 +167,54 @@ it("should change the role of the ErrorWrapper when an invalid input is given",
   expect(errorWrapper).not.toBeEmptyDOMElement();
   expect(errorWrapper).toHaveAttribute("role", "alert");
 });
+
+test("character limit counter should appear for long text inputs", async () => {
```
RODO94, line 171: testing follows a similar pattern to what I seen above, thought this homogeny would help readability.

editor.planx.uk/src/@planx/components/TextInput/Public.tsx
```diff
@@ -60,9 +74,10 @@ const TextInputComponent: React.FC<Props> = (props) => {
             onChange={formik.handleChange}
             errorMessage={formik.errors.text as string}
             id={props.id}
+            textLength={props.type}
             inputProps={{
               "aria-describedby": [
-                props.description ? DESCRIPTION_TEXT : "",
+                props.description ? `${DESCRIPTION_TEXT} character-hint` : "",
```
RODO94, line 80: added a hint for screen readers to introduce character count when you first click on it

editor.planx.uk/src/@planx/components/TextInput/model.ts
```diff
@@ -12,6 +12,10 @@ export enum TextInputType {
   Phone = "phone",
 }
 
+export const shortTextLimit = 120;
```
RODO94, line 15: I thought it would be more maintainable if I could have these as exported variables I can use in the below conditions and also in the `<Input>` component

editor.planx.uk/src/ui/shared/CharacterCounter.tsx
```diff
@@ -0,0 +1,76 @@
...
+  characterCount: number;
+  error: boolean;
+};
+
+export const CharacterCounter: React.FC<Props> = (props) => {
+  const [characterLimitAnnouncement, setCharacterLimitAnnouncement] =
+    useState<number>(0);
+
+  const hintTextBool = useRef(true);
+  const updateCharacterCount = useCallback(
```
RODO94, line 17: Not super familiar with best practices on `useCallback` and `debounce` but I wanted it to only be called once on mounting in React and then called after a certain time period of typing to reduce noise in the screen reader.

Keen for alternates / thoughts

editor.planx.uk/src/ui/shared/CharacterCounter.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <Typography
+        aria-live="polite"
+        sx={{
+          position: "absolute",
+          height: "1px",
+          width: "1px",
+          overflow: "hidden",
+        }}
+      >
+        {characterLimitAnnouncementText}
```
RODO94, line 72: This hidden character counter should only be updated and announced after the `debounce()` has ran so you don't get the noice when typing each letter. I don't think I got it quite right in implementation, but I still found an improvement.

editor.planx.uk/src/ui/shared/Input.tsx
```diff
@@ -32,6 +38,7 @@ export interface Props extends InputBaseProps {
   large?: boolean;
   bordered?: boolean;
   errorMessage?: string;
+  textLength?: string;
```
RODO94, line 41: Added this prop so you can set character limit based on `type`

editor.planx.uk/src/ui/shared/Input.tsx
```diff
@@ -120,19 +145,30 @@ export default forwardRef((props: Props, ref): FCReturn => {
...
+            "aria-labelledby": ariaLabelledBy,
+          }}
+          id={id}
+          ref={container}
+          {...restProps}
+        />
+        {showCharacterCountBool && (
+          <CharacterCounter
+            characterCount={
+              typeof props.value === "string" ? props.value.length : 0
```
RODO94, line 165: `typeOf()` used here to get around the `value` having a type of `unknown`

DafyddLlyr (changes requested):
A few suggestions!

Fundamentally solid, getting the expected behaviour as far as I can tell here 👍

editor.planx.uk/src/ui/shared/Input.tsx
RODO94: I still need something to create logic on excluding `"Short"` text input types from the character count logic, this is something I would need from the props when specifying the "type"? 

`props.type` in this instance will only be "text" for the type of input we want

editor.planx.uk/src/ui/shared/CharacterCounter.tsx
```diff
@@ -0,0 +1,71 @@
...
+      >
+        {characterLimitText}
+      </Typography>
+      <Typography
+        aria-live="polite"
+        aria-atomic="true"
+        sx={visuallyHidden}
+        data-testid="screen-reader-count"
+      >
+        {showReaderCount && screenReaderCountText}
```
RODO94, line 67: This is hiding and redisplaying the screen reader announcement, it works to ensure the screen reader only reads relevant info to the user, but it leads to the Label being re-announced multiple times.

A way around this would be to separate out a label as I have done for the `"character-hint"` while hiding the current label. This would allow re-rendering of the announcement without the issue of extra noise.

Interesting to know if I am over-thinking this and adding too many layers of complexity as the who point of the label for an input is to ease this type of thing, but I couldn't think of anything else.

editor.planx.uk/src/ui/shared/Input.tsx
DafyddLlyr: We can either remove this prop and handle the character count within `TextInput` directly (see other comments), or just pass this value if the types match - 

```ts
textLength={[TextInputType.Long, TextInputType.ExtraLong].includes(type) && TEXT_LIMIT[type]
```

editor.planx.uk/src/ui/shared/CharacterCounter.tsx
DafyddLlyr: I'd expect that because of the debounce, we don't need the variable `showReaderCount` - it's just a single value which is updated in place. Can we remove this guard and just keep `screenReaderCountText` in place?

DafyddLlyr (changes requested):
A few small issues, happy to catch up tomorrow to talk things through if you'd like to 🙂

editor.planx.uk/src/ui/shared/Input.tsx
```diff
@@ -115,24 +123,32 @@ export default forwardRef((props: Props, ref): FCReturn => {
...
+          id={id}
+          ref={container}
+          {...restProps}
+        />
+        {showCharacterCount && (
+          <CharacterCounter
+            count={typeof props.value === "string" ? props.value.length : 0}
+            limit={TEXT_LIMITS[textLength]}
+          />
+        )}
```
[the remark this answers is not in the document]
RODO94: Yeah, I started off with this strategy but went for a DRY purist view of putting it here to stop repitition. Thinking was that if we ever had a new text field input somewhere, we wouldn't need to think of adding a character count because it is a standard feature of any long text input we use, so embedding the use as standard. 

I do worry about it bloating the <Input> though, so would appreciate the second opinion on it!

editor.planx.uk/src/@planx/components/shared/utils.ts
```diff
@@ -63,3 +64,18 @@ export const getPreviouslySubmittedData = ({
 
   return data;
 };
+
+export const characterCounterSwitch = (type: TextInputType | undefined) => {
```
RODO94, line 68: Ran round a few circles here trying to get a DRY, maintainable way of filtering out the different `TextInputType` keys, and found this to be the cleanest. 

I could see benefit in adding more complex `TEXT_LIMITS[]` or `props.type` logic at the `<TextInput>` level, but I couldn't get it to work without being cumbersome, or looking like a lot. This felt more readable

editor.planx.uk/src/@planx/components/shared/Schema/InputFields/TextFieldInput.tsx
```diff
@@ -33,13 +36,22 @@ export const TextFieldInput: React.FC<Props<TextField>> = (props) => {
...
           ]
             .filter(Boolean)
             .join(" "),
         }}
       />
+      {characterCountLimit && (
+        <CharacterCounter
+          limit={characterCountLimit}
+          count={fieldProps.value.length}
+          error={fieldProps.errorMessage ? true : false}
```
RODO94, line 52: This is to manage the style changes when an error is passed to the `<ErrorWrapper>` of the `<Input>` comp. Couldn't find a better way of managing this without there being strange styling.

Mostly comes when you go over the character limit, delete characters so you are within limit, but you don't submit. The alignment ends up being off between the Input error and the character counter error

editor.planx.uk/src/ui/shared/CharacterCounter.tsx
```diff
@@ -0,0 +1,73 @@
...
+  );
+
+  const currentCharacterCount = TEXT_LIMITS[limit] - count;
+  const showCharacterLimitError = currentCharacterCount < 0;
+
+  useEffect(() => {
+    if (count !== screenReaderCount) {
+      setShowReaderCount(false);
+    }
+    if (count > 0) {
```
RODO94, line 33: This reduced the amount of noise coming when you enter the textarea, felt low key enough to sneak in...

editor.planx.uk/src/ui/shared/CharacterCounter.tsx
RODO94: Tried to remove this now `<CharacterCounter>` sits outwith the `<Input>` but still found it announcing it too many times when typing.

editor.planx.uk/src/@planx/components/TextInput/Public.tsx
```diff
@@ -31,6 +36,8 @@ const TextInputComponent: React.FC<Props> = (props) => {
     }),
   });
 
+  const characterCountLimit = characterCounterSwitch(props.type);
```
DafyddLlyr, line 39: I'd argue that this is logic which is should belong to the character count component.

 Could we run this check with the component, returning `null` if the condition is not met?

editor.planx.uk/src/@planx/components/TextInput/Public.tsx
```diff
@@ -62,13 +69,22 @@ const TextInputComponent: React.FC<Props> = (props) => {
...
               ]
                 .filter(Boolean)
                 .join(" "),
             }}
           />
+          {characterCountLimit && (
+            <CharacterCounter
+              count={formik.values.text.length}
+              limit={characterCountLimit}
+              error={formik.errors.text ? true : false}
```
DafyddLlyr, line 85: nit: Maybe a bit cleaner?

```suggestion
              error={Boolean(formik.errors.text)}
```

editor.planx.uk/src/@planx/components/shared/utils.ts
DafyddLlyr: This is a very fair comment 👍 

This works just fine, but I think the more terse `const showCharacterCount = [TextInputType.ExtraLong, extInputType.Long].includes(type)` is just as clear imho - especially if we were to move this to the `CharacterCount` component.

Alternatively, we could encode this into `TEXT_LIMITS` - 

```ts
export const TEXT_LIMITS = {
  [TextInputType.Short]: {
    showCharacterCount: false,
    limit: 150,
  },
  [TextInputType.Long]: {
    showCharacterCount: true,
    limit: 250,
  },
  [TextInputType.ExtraLong]: {
    showCharacterCount: true,
    limit: 750,
  },
} as const;
```

I think the first (`.includes()`) method would be my preference here, but no strong feelings, this approach works perfectly also 👍

DafyddLlyr (approved):
A few comments, none that should be blocking though 👍

editor.planx.uk/src/ui/shared/CharacterCounter.tsx
```diff
@@ -0,0 +1,92 @@
...
+import React, { useCallback, useEffect, useState } from "react";
+import { FONT_WEIGHT_SEMI_BOLD } from "theme";
+
+export type Props = {
+  textInputType: TextInputType;
+  count: number;
+  error: boolean;
+};
+
+export function isLongTextType(
```
RODO94, line 14: @DafyddLlyr added this for type refining the types

editor.planx.uk/src/ui/shared/CharacterCounter.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const updateCharacterCount = useCallback(
+    debounce((count: number) => {
+      setScreenReaderCount(count);
+      setShowReaderCount(true);
+    }, 500),
+    [],
+  );
+
+  function getLongTextLimit(type: TextInputType): number {
```
RODO94, line 36: added this for displaying the right limit

editor.planx.uk/src/ui/shared/CharacterCounter.tsx
```diff
@@ -0,0 +1,92 @@
...
+    if (isLongTextType(type)) {
+      return TEXT_LIMITS[type].limit;
+    } else {
+      return 0;
```
RODO94, line 40: return a 0 for the situation where nothing happens, but assuming there's so much type guarding and logic before this, it feels like an unlikely situation

editor.planx.uk/src/ui/shared/CharacterCounter.tsx
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+
+  const currentCharacterCount = getLongTextLimit(textInputType) - count;
```
RODO94, line 44: So now I pass in the function with the `textInputType`

editor.planx.uk/src/@planx/components/shared/Schema/InputFields/TextFieldInput.tsx
```diff
@@ -12,6 +14,7 @@ export const TextFieldInput: React.FC<Props<TextField>> = (props) => {
   const { data, formik } = props;
   const { id, errorMessage } = fieldProps;
 
+  const characterCountLimit = data.type && isLongTextType(data.type);
```
RODO94, line 17: For me, it made sense to keep this here so when you look into the TextFieldInput, you understand why things show and why things don't without ever having to go into the CharacterCount component. To understand the TextFieldInput, you don't need to go to another file, in theory

PR #3755, opened 2024-10-03
fix: correct submissions log date comparison (hide download if > 28 days old)

jessicamcinchak:
Noticed when merged to staging - eg buttons should be hidden here because > 28 days old https://editor.planx.dev/southwark/apply-for-planning-permission/submissions-log (should help minimise prevalence of "cannot find published flow" error on staging, but that's another topic!)

Was my mistake - just a quick swap of comparison dates here so we're correctly calculating the expiry date _from_ the submission event, not now as before.


DafyddLlyr (approved):
Good catch!

PR #3759, opened 2024-10-03
[api] remove `/logout` route (do not implement front channel logout)

freemvmt:
Relates to [this Trello ticket](https://trello.com/c/0gYN7cdt) and follows on from #3453.

After some investigation, I came to the conclusion that implementing [front channel logout](https://learn.microsoft.com/en-us/entra/identity-platform/v2-protocols-oidc#single-sign-out) does not make sense in our context.

Before I realised that, and while trying to figure out how best to implement it, I realised that the API's `/logout` route is never hit, so serves no purpose. Therefore I have removed it in this PR.

## Some context

As per the Open ID Connect spec, a front channel logout is intended to log a user out of any server on which they are authenticated by a provider (e.g. Microsoft), when they log out of said provider centrally (or via another service).

For example, if I log out of my Microsoft account on Microsoft.com, I'd expect to be simultaneously logged out of PlanX, assuming I'd logged in via Microsoft SSO.

However, a PlanX user being 'logged in' is currently achieved by the presence of a `jwt` cookie stored on the browser, and does not relate to any session or state stored on the API server or in the database (this client-side session data is cleared when the user clicks `Logout` - see [here](https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/src/routes/index.tsx#L33)).

Therefore, there is no server to communicate a logout instruction to. The JWT on the user's browser cannot be revoked from without, will expire in due course (??), and cannot be renewed without signing in again.

I think this is a reasonable situation. If we wanted Microsoft to be able to log a user out, we'd have to keep session server-side, which would be a significant change.

## Further resources

- https://openid.net/specs/openid-connect-frontchannel-1_0.html#OPLogout

### DIFF
--- a/api.planx.uk/modules/auth/controller.ts
+++ b/api.planx.uk/modules/auth/controller.ts
@@ -7,14 +7,6 @@ export const failedLogin: RequestHandler = (_req, _res, next) =>
     message: "User failed to authenticate",
   });
 
-export const logout: RequestHandler = (req, res) => {
-  // TODO: implement dual purpose as Microsoft frontend logout channel
-  req.logout(() => {
-    // do nothing
-  });
-  res.redirect(process.env.EDITOR_URL_EXT!);
-};
-
 export const handleSuccess = (req: Request, res: Response) => {
   if (!req.user) {
     return res.json({
--- a/api.planx.uk/modules/auth/docs.yaml
+++ b/api.planx.uk/modules/auth/docs.yaml
@@ -6,13 +6,6 @@ tags:
   - name: auth
     description: Authentication related requests
 paths:
-  /logout:
-    get:
-      summary: Logout from the PlanX service
-      tags: ["auth"]
-      responses:
-        "302":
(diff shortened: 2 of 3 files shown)


DafyddLlyr (changes requested):
Thanks for this, as well as the clear and comprehensive update in the PR description.

One small change required - we also need to update `api.planx.uk/modules/auth/docs.yaml` to remove the `/logout` route - this is currently not dynamically generated by Express and is a bit of a painful manual step. This generates our Swagger API docs hosted here - 

https://api.3759.planx.pizza/docs/#/auth/get_logout

freemvmt:
> One small change required - we also need to update `api.planx.uk/modules/auth/docs.yaml` to remove the `/logout` route [...]

@DafyddLlyr nice catch, have done! I took the opportunity to document the new `/auth/microsoft` routes in the same file, since I'd missed that before.

DafyddLlyr (approved):
Perfect! 👏

PR #3762, opened 2024-10-04
feat(ui): Node tags

DafyddLlyr:
## What does this PR do?
 - Creates a `ComponentTagSelect` component, and adds this to all components which have more info and/or internal notes (I think just portals are excluded)
 - This is contained within a `ModalFooter` component, which now wraps internal notes, more information, and tags
 - Displays tags on the graph
 
<p align="center">
image

image
</p>

## Next steps / questions
A few great comments and questions from show and tell which I've listed below, with my suggestions on how to approach them.

- Can we have more tags (e.g. automation, sensitive)?
  - It might be worthwhile adding a few extra initially here to make the most of this feature. I can compile a small list and add to PlanX core

- Should tags have an associated colour?
   - Probably! We can match them and then update chips within the modal to re-use this colour scheme - seems a change worth making

- Should tags display on graphs? Is this a templates only thing? How should more than one tag display?
  - Great question, this one might need some more input from @ianjon3s - I guess if we have multiple tags per component, we need an alternate display method. For now I can not display on the graph in order to progress this PR and we can come back here shortly

- Can I search by tags or view them any other way?
  - Yes, this will be added as a facet to search shortly

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/SelectMultipleFileTypes.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/SelectMultipleFileTypes.tsx
@@ -163,7 +163,7 @@ export const SelectMultipleFileTypes = (props: SelectMultipleProps) => {
   };
 
   return (
-    <SelectMultiple<Option>
+    <SelectMultiple
       getOptionLabel={(option) => option.name}
       groupBy={(option) => option.category}
       id={`select-multiple-file-tags-${uploadedFile.id}`}
--- a/editor.planx.uk/src/@planx/components/Notice/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/Notice/Editor.tsx
@@ -7,6 +7,7 @@ import { ICONS, InternalNotes, MoreInformation } from "@planx/components/ui";
 import { useFormik } from "formik";
 import React from "react";
 import ColorPicker from "ui/editor/ColorPicker";
+import { ComponentTagSelect } from "ui/editor/ComponentTagSelect";
 import ModalSection from "ui/editor/ModalSection";
 import ModalSectionContent from "ui/editor/ModalSectionContent";
 import RichTextInput from "ui/editor/RichTextInput";
@@ -104,6 +105,13 @@ const NoticeEditor: React.FC<NoticeEditorProps> = (props) => {
           });
         }}
         value={props.value.notes}
+      />
+      <ComponentTagSelect
+        onChange={(value) => props.onChange({
+          ...props.value,
+          tags: value,
+        })}
+        value={props.value.tags}
       />
     </>
   );
--- a/editor.planx.uk/src/pages/FlowEditor/floweditor.scss
+++ b/editor.planx.uk/src/pages/FlowEditor/floweditor.scss
@@ -137,6 +137,7 @@ $fontMonospace: "Source Code Pro", monospace;
     border: $nodeBorderWidth dashed red;
   }
 
+  & > div > a,
   & > a {
     position: relative;
     display: flex;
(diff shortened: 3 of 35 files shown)


editor.planx.uk/src/@planx/components/FileUploadAndLabel/SelectMultipleFileTypes.tsx
```diff
@@ -163,7 +163,7 @@ export const SelectMultipleFileTypes = (props: SelectMultipleProps) => {
   };
 
   return (
-    <SelectMultiple<Option>
+    <SelectMultiple
```
DafyddLlyr, line 166: This type can be removed as it's redundant - it's already being inferred correctly from the `options` prop.

editor.planx.uk/src/@planx/components/Notice/Editor.tsx
```diff
@@ -105,6 +106,13 @@ const NoticeEditor: React.FC<NoticeEditorProps> = (props) => {
         }}
         value={props.value.notes}
       />
+      <ComponentTagSelect
+        onChange={(value) => props.onChange({
+          ...props.value,
+          tags: value,
+        })}
+        value={props.value.tags}
+      />
```
DafyddLlyr, line 115: There's a few components where a like-for-like swap to use `ModalFooter` isn't simple. 

Generally this is because either formik is not used (as in this component), or Formik is used via a context wrapper (see `Pay`) instead of the `useFormik()` hook.

A refactor is worth doing I think, but is outside the scope of this PR.

editor.planx.uk/src/pages/FlowEditor/floweditor.scss
```diff
@@ -137,6 +137,7 @@ $fontMonospace: "Source Code Pro", monospace;
     border: $nodeBorderWidth dashed red;
   }
 
+  & > div > a,
```
DafyddLlyr, line 140: Another reminder that it would be good to update these styles and incorporate them into MUI components sometime soon!

editor.planx.uk/src/ui/editor/ComponentTagSelect.tsx
```diff
@@ -0,0 +1,47 @@
...
+  false,
+  "div"
+>["renderOption"] = (props, tag, { selected }) => (
+  <ListItem {...props}>
+    <CustomCheckbox aria-hidden="true" className={selected ? "selected" : ""} />
+    { TAG_DISPLAY_VALUES[tag].displayName }
+  </ListItem>
+);
+
+export const ComponentTagSelect: React.FC<Props> = ({ value, onChange }) => {
```
DafyddLlyr, line 30: Better names very welcome here!

editor.planx.uk/src/pages/FlowEditor/floweditor.scss
jessicamcinchak: :heavy_plus_sign:

editor.planx.uk/src/ui/editor/ModalFooter.tsx
```diff
@@ -0,0 +1,45 @@
...
+      />
+    }
+    {showInternalNotes &&
+      <InternalNotes
+      name="notes"
+      onChange={formik.handleChange}
+      value={formik.values.notes}
+      />
+    }
+    { showTags &&
```
jessicamcinchak, line 38: nit: unnecessary leading space here inconsistent with preceding conditions (surprised precommit hook didn't catch!)

jessicamcinchak (approved):
Works for me !

editor.planx.uk/src/ui/editor/ModalFooter.tsx
DafyddLlyr: Thanks, fixed in [cda35f8](https://github.com/theopensystemslab/planx-new/pull/3762/commits/cda35f88d404a3812193b79fd1edeb7c3e1e36d1)

PR #3767, opened 2024-10-04
feat: Use zustand middleware to persist to local storage

DafyddLlyr:
This PR is a rethink of two Friday quick fixes - 
 - https://github.com/theopensystemslab/planx-new/pull/3765
 - https://github.com/theopensystemslab/planx-new/pull/3766

Originally, I was just resolving a React error and then tidying up some unused store variables.

In this approach, I'm using Zustand's persist middleware pattern ([docs](https://zustand.docs.pmnd.rs/integrations/persisting-store-data)) to handle the variable stores in local storage. Here's the benefits to this approach as I see them - 

 - Less boilerplate in components that would need local or session storage - we set this up once in a store
 - An easy pattern to follow if we needs something like this elsewhere (e.g. the staging banner, session (!) or notifications when we get there)
 - Using a store means the variable could still be controlled by multiple React components
 - Less complex styled components (passing props, remembering to use `shouldForwardProp` etc)
   - I've also used a MUI Collapse component to save passing these variables into the styled component

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
@@ -5,6 +5,7 @@ import OpenInNewIcon from "@mui/icons-material/OpenInNew";
 import OpenInNewOffIcon from "@mui/icons-material/OpenInNewOff";
 import Box from "@mui/material/Box";
 import Button from "@mui/material/Button";
+import Collapse from "@mui/material/Collapse";
 import Container from "@mui/material/Container";
 import Link from "@mui/material/Link";
 import { styled } from "@mui/material/styles";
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
```diff
@@ -167,102 +151,114 @@ const Sidebar: React.FC = React.memo(() => {
...
-            </Permission.IsPlatformAdmin>
+    <Root>
+      <Collapse 
+        in={showSidebar} 
+        orientation="horizontal" 
+        collapsedSize={SIDEBAR_WIDTH_MINIMISED} 
+        sx={{ height: "100%" }}
+        easing={"ease-in-out"}
+        timeout={200}
+      >
```
DafyddLlyr, line 162: Diff here is a bit weird - I've just wrapped this component in a MUI Collapse ([docs](https://mui.com/material-ui/api/collapse/)) instead of controlling this via variables passed to a styled component.

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
```diff
@@ -132,16 +126,14 @@ const TabList = styled(Box)(({ theme }) => ({
...
+  const [resetPreview, isFlowPublished, toggleSidebar, showSidebar] = useStore((state) => [
     state.resetPreview,
     state.isFlowPublished,
+    state.toggleSidebar,
+    state.showSidebar,
   ]);
 
   const [activeTab, setActiveTab] = useState<SidebarTabs>("PreviewBrowser");
-  const [isSidebarMinimised, setIsSidebarMinimised] = useState<boolean>(() => {
-    const savedState = localStorage.getItem("isSidebarMinimised");
```
RODO94, line 142: @DafyddLlyr is the gain that we swap this for the code above it: 

```ts
  const [resetPreview, isFlowPublished, toggleSidebar, showSidebar] = useStore((state) => [
    state.resetPreview,
    state.isFlowPublished,
    state.toggleSidebar,
    state.showSidebar,
  ]);
```

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -45,29 +46,35 @@ const send = (ops: Array<any>) => {
...
 
-  isTestEnvBannerVisible: !window.location.href.includes(".uk"),
+    isTestEnvBannerVisible: !window.location.href.includes(".uk"),
 
-  hideTestEnvBanner: () => set({ isTestEnvBannerVisible: false }),
-});
+    hideTestEnvBanner: () => set({ isTestEnvBannerVisible: false }),
+  }),
+  {
+    name: "editorUIStore",
```
RODO94, line 74: Is Local Storage the default here or have I missed where you specified this?

RODO94 (approved):
Checked the Application tab and all seemed to go well for me with the Store object in there.

```ts
{"state":{"showSidebar":true},"version":0}
```

Only checked info I seen in this PR in terms of code changes, nothing outwith what I seen here

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
DafyddLlyr: Yep, we also remove the `togglePreview()` function below as it's now handled by Zustand.

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
DafyddLlyr: That's correct, local storage is just default so we don't need to specify anything here.

PR #3772, opened 2024-10-07
fix: Restore loading bar and `FlowEditor` state on Editor-only routes

DafyddLlyr:
## What does this PR do?
 - This is a belated follow up to https://github.com/theopensystemslab/planx-new/pull/3671 and https://github.com/theopensystemslab/planx-new/pull/3714
 - Re-ordered routes so that the loading progress bar is restored, and the public facing pages maintain their loading spinner
 - The previous fix (#3714) wrapped `"*": editorRoutes` in the `loadingView` in order to restore the loading spinner to public facing pages, here I'm (finally!) fixing this properly.

### How can I test this?
 - All Editor routes should have a loading progress bar, when navigating between routes. There is still an initial loading spinner on first load.
 - All non-Editor routes should have a loading spinner

### DIFF
--- a/editor.planx.uk/src/routes/index.tsx
+++ b/editor.planx.uk/src/routes/index.tsx
@@ -57,38 +57,48 @@ const editorRoutes = mount({
   ),
 });
 
-const mountPayRoutes = () =>
-  map(async () => {
-    compose(withView(loadingView));
-    return lazy(() => import("./pay"));
-  });
+const loadPayRoutes = () =>
+  compose(
+    withView(loadingView),
+    lazy(() => import("./pay")),
+  );
+
+const loadPublishedRoutes = () =>
+  compose(
+    withView(loadingView),
+    lazy(() => import("./published")),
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/routes/index.tsx
```diff
@@ -57,38 +57,48 @@ const editorRoutes = mount({
...
-        "/:team/:flow/published": lazy(() => import("./published")), // XXX: keeps old URL working, but only for the team listed in the domain.
-        "/:flow": lazy(() => import("./published")),
-        "/:flow/pay": mountPayRoutes(),
-        // XXX: We're not sure where to redirect `/` to so for now we'll just return the default 404
-        // "/": redirect("somewhere?"),
-      }),
-    )
-  : compose(
-      withView(loadingView),
+const loadDraftRoutes = () =>
```
jamdelion, line 78: Nice refactor - much easier to read :)

PR #3763, opened 2024-10-04
feat(search): Map all node keys to display values

DafyddLlyr:

PR #3773, opened 2024-10-08
fix: Correctly construct URL by re-ordering portals

DafyddLlyr:
## What's the problem?
 - Identified here by @jessicamcinchak - https://opensystemslab.slack.com/archives/C5Q59R3HB/p1727767192939239?thread_ts=1726217456.551129&cid=C5Q59R3HB (OSL Slack)
 - When constructing the URL for a node, and then exiting, the path was incorrect

## What's the solution?
 - The order of portals in the URL is hierarchical (grandparent → parent → child), but the path generated by `getPathForNode()` is in opposite order (current node → parent → grandparent).
 - This means just need to call `reverse()` on the list of internal portals 👌

## To test - 
 - Search for "mansard" in the following flow - https://3773.planx.pizza/opensystemslab/permitteddevelopment
 - Close modal
 - Your URL is now https://3773.planx.dev/opensystemslab/permitteddevelopment,7GeYLgmnm9,cw73eSHjy2
 - You should be able to see the "mansard" node

## Next steps...
It's a little unexpected that the navigation happens after you close the modal, I think I'd expect it on click of the result card when the modal is opened. This may not be an easy change, but I'll timebox a look into this one.

This really highlights to me how valuable it would be to locate your node in the X/Y space of the flow editor. I'll add a card to the backlog for this, and might try to pick it up before a Friday show + tell.

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/getURLForNode.test.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/getURLForNode.test.ts
@@ -54,14 +54,14 @@ describe("constructing URLs for nodes", () => {
   test("a question node, deeply within internal portals", () => {
     const url = getURLForNode("tarantinoMovieQuestion");
     expect(url).toEqual(
-      "/testTeam/testFlow,tarantinoPortal,moviesPortal/nodes/tarantinoPortal/nodes/tarantinoMovieQuestion/edit",
+      "/testTeam/testFlow,moviesPortal,tarantinoPortal/nodes/tarantinoPortal/nodes/tarantinoMovieQuestion/edit",
     );
   });
 
   test("an answer node, deeply within internal portals", () => {
     const url = getURLForNode("reservoirDogsAnswer");
     expect(url).toEqual(
-      "/testTeam/testFlow,tarantinoPortal,moviesPortal/nodes/tarantinoPortal/nodes/tarantinoMovieQuestion/edit",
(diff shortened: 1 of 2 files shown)


jessicamcinchak (commented):
Hey thanks for digging into this one! 

Getting some unexpected behavior on pizza: 
- If I start here: https://3773.planx.pizza/opensystemslab/permitteddevelopment
- And do a data-only search for `internal.mezzanine` (which is the data `val` on one of the options in the root-level Checklist)
- My results are nearly _all_ for non-data-field content in internal portals  and the root-level option is never picked up? 
Screenshot from 2024-10-08 14-58-21

Haven't compared this to staging behavior yet, but doesn't feel expected - anything I might be missing here?

DafyddLlyr:
Ah! This is a feature flag issue sorry - I've kept the button disabled but not correctly handled the initial values.

I'm actually removing the flag in this PR - https://github.com/theopensystemslab/planx-new/pull/3763

In the meantime, we can wait for the above to merge, or test this PR after toggling the feature flag (`window.featureFlags.toggle("DATA_ONLY_SEARCH")`)

jessicamcinchak (approved):
Aha thanks for confirming feature flag issue - happy to ignore that then here.

URLs are correct, thanks for pinpointing sort issue here! Indeed agree navigation is a bit unexpected (both that it happens _after_ closing modal, and that it maintains previous scroll x/y position which no longer matches _new_ result) - but correct URLs feel most important priority!

PR #3776, opened 2024-10-08
feat(search): Strip HTML tags from search results

DafyddLlyr:
To do - 
 - [ ] Test with HTML


editor.planx.uk/src/hooks/useSearch.ts
```diff
@@ -27,6 +33,7 @@ export const useSearch = <T extends object>({
       useExtendedSearch: true,
       includeMatches: true,
       minMatchCharLength: 3,
+      ignoreLocation: true,
```
DafyddLlyr, line 36: By default, just the first 60 characters are searched which was giving me some unexpected false negative results.

The testing I've done doesn't make this feel much slower performance-wise - will flag this when this PR is tested.

Docs: https://www.fusejs.io/api/options.html#ignorelocation

DafyddLlyr:
Replaced by https://github.com/theopensystemslab/planx-new/pull/3783

PR #3777, opened 2024-10-08
fix: Center align text env banner vertically

DafyddLlyr:
## What does this PR do?
 - Ensures that the content of the test environment banner are vertically aligned correctly
 - It's always been off as `alignItems: "center"` was used without `display: "flex"`, it's just more noticeable now following https://github.com/theopensystemslab/planx-new/pull/3734  which made the banner taller

| Before | After |
|--------|--------|
|image|image|


ianjon3s (approved):
Good spot!

PR #3781, opened 2024-10-08
chore: CVE-2024-45296

DafyddLlyr:
Fixes https://github.com/theopensystemslab/planx-new/security/dependabot/115

Been [watching this one for a while](https://github.com/vercel/serve/issues/811), hoping that [serve](https://github.com/vercel/serve) update their dependencies. No movement so far, so let's just add an override and see if we hit any issues. If so, we can find an alternative to serve I'm sure 👍 

🚀 I'm running regression tests against this branch here - https://github.com/theopensystemslab/planx-new/actions/runs/11243267703


DafyddLlyr:
Closed - E2E tests failing (which rely on `serve`) on multiple runs.

PR #3783, opened 2024-10-09
feat: Strip HTML from rich text search results

DafyddLlyr:
## What does this PR do?
 - Strips HTML tags from rich text fields
 - Splits HTML elements with newlines to display in the `SearchResultCard`

| Before | After |
|--------|--------|
| image | image |


editor.planx.uk/src/hooks/useSearch.ts
```diff
@@ -27,6 +33,7 @@ export const useSearch = <T extends object>({
       useExtendedSearch: true,
       includeMatches: true,
       minMatchCharLength: 3,
+      ignoreLocation: true,
```
DafyddLlyr, line 36: By default, just the first 60 characters are searched which was giving me some unexpected false negative results.

The testing I've done doesn't make this feel much slower performance-wise - will flag this when this PR is tested.

Docs: https://www.fusejs.io/api/options.html#ignorelocation

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/SearchResultCard/getDisplayDetailsForResult.tsx
```diff
@@ -307,7 +241,7 @@ const defaultFormatter: SearchResultFormatter = {
   getIconKey: ({ item }) => item.type,
   getTitle: ({ item }) =>
     (item.data?.title as string) || (item.data?.text as string) || "",
-  getHeadline: ({ item, key }) => get(item, key)?.toString() || "",
+  getHeadline: ({ matchValue }) => matchValue,
```
DafyddLlyr, line 244: This is a nice simplification! Rather than try to reverse engineer the match back via multiple different `getHeadline()` methods, we can just use the original `matchValue` which fuse.js used.

jamdelion:
Looks nice and clean without the markdown in the search result :) I do find it a bit odd that several seemingly unrelated strings are sometimes thrown up in bold (e.g `aph` in 'paragraph' in your example), but I think I gathered from how you talked about this a few weeks ago that this whole feature has been quite tricky to get right! At least it certainly seems to throw up the most likely search results first, in my testing anyway. :)

PR #3785, opened 2024-10-09
fix(e2e): Ensure exact match for team slug

DafyddLlyr:
🚀 Regression tests running here on CI, passing locally - https://github.com/theopensystemslab/planx-new/actions/runs/11257964868

Best guess is that this was introduced by https://github.com/theopensystemslab/planx-new/pull/3772

The selector `page.getByRole("link", { name: context.team.slug })` is matching both the header and the team "card" in the list - I guess prior to the loading bar this was actually a false positive?


jessicamcinchak (approved):
Thanks for fix !!

PR #3786, opened 2024-10-09
fix: Add permission guard to "Add a new Editor" button

DafyddLlyr:
## What's the problem?
- Airbrake caught the following error, which is a teamEditor failing to add a new user https://planx.airbrake.io/projects/329753/groups/3909794420521801267
- They aren't able to do this as they must be a platformAdmin for the action to work
- This means that the button must be visible

I've also messaged the user in question to see if we can help them add a user 👍 

## What's the solution?
- Add the correct permission guard wrapper to the button
- Update the test to use a non-template team to correctly catch this behaviour

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
@@ -194,17 +194,19 @@ export const MembersTable = ({
               </StyledTableRow>
             ))}
             {showAddMemberButton && (
-              <TableRow>
-                <TableCell colSpan={5}>
-                  <AddButton
-                    onClick={() => {
-                      addUser();
-                    }}
-                  >
-                    Add a new editor
-                  </AddButton>
-                </TableCell>
-              </TableRow>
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
```diff
@@ -194,17 +194,19 @@ export const MembersTable = ({
...
+                    <AddButton
+                      onClick={() => {
+                        addUser();
+                      }}
+                    >
+                      Add a new editor
+                    </AddButton>
+                  </TableCell>
+                </TableRow>
+              </Permission.IsPlatformAdmin>
```
jessicamcinchak, line 209: We must have had a bad rebase somewhere because I swear I made this exact change a couple weeks ago ?? :eyes: 

Eg https://github.com/theopensystemslab/planx-new/pull/3713/files#diff-b5928b9d92361e26a33f6c0b9af0253b319eab4128cee43689cce0eb88cc87c5R151

jessicamcinchak (approved):
Again, thanks for this fix !

PR #3789, opened 2024-10-09
feat: toggle node images and data fields on editor graph

jessicamcinchak:
Adds ability to toggle visibility of images & data fields on the graph. Follows on from #3674 (toggle tags) & #2947 (maximalist biennale graph styling).

In cases where all attributes are set & toggled "on", proposed heirarchy of info from top to bottom is:
 - Image
 - Flag band
 - Text / title (sometimes including icon)
 - Data field
 - Tag 
 
 Does that feel right? Toggle button group is organised in same vertical order.

Screenshot from 2024-10-16 10-01-27

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Checklist.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Checklist.tsx
@@ -15,6 +15,7 @@ import { getParentId } from "../lib/utils";
 import Hanger from "./Hanger";
 import Node from "./Node";
 import { Tag } from "./Tag";
+import { Thumbnail } from "./Thumbnail";
 
 type Props = {
   type: TYPES;
@@ -96,6 +97,12 @@ const Checklist: React.FC<Props> = React.memo((props) => {
             <span>{props.text}</span>
           </Link>
           {props.tags?.map((tag) => <Tag tag={tag} key={tag} />)}
+          {props.data?.img && (
+            <Thumbnail
+              imageSource={props.data?.img}
+              imageAltText={props.data?.text}
+            />
+          )}
         </Box>
         {groupedOptions ? (
           <ol className="categories">
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Checklist.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Checklist.tsx
@@ -12,6 +12,7 @@ import { Link } from "react-navi";
 
 import { useStore } from "../../../lib/store";
 import { getParentId } from "../lib/utils";
+import { DataField } from "./DataField";
 import Hanger from "./Hanger";
 import Node from "./Node";
 import { Tag } from "./Tag";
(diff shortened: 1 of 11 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Checklist.tsx
```diff
@@ -92,9 +94,18 @@ const Checklist: React.FC<Props> = React.memo((props) => {
...
+                imageSource={props.data?.img}
+                imageAltText={props.data?.text}
+              />
+            )}
             {Icon && <Icon />}
             <span>{props.text}</span>
           </Link>
+          {props.data?.fn && (
+            <DataField value={props.data.fn} variant="parent" />
+          )}
```
jessicamcinchak, line 108: Small styling bug outstanding here actually - we need to get all attributes wrapped in the `Link` I think in order to avoid disjointed "visited" green border styling - navigating a mix of scss classnames & React component styles here though now which is a bit tricky! 
Screenshot from 2024-10-10 11-13-00

ianjon3s: Updating line 113 of editor.planx.uk/src/pages/FlowEditor/floweditor.scss to this should do the trick:
`&.wasVisited > div a,
  &.wasVisited > a {`

jessicamcinchak: Thanks! Rebasing to that change fixed borders :heavy_plus_sign:

editor.planx.uk/src/pages/FlowEditor/components/FlowEditor/ToggleDataFieldsButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+          onClick={toggleShowDataFields}
+          size="large"
+          sx={(theme) => ({
+            padding: theme.spacing(1),
+            color: showDataFields
+              ? theme.palette.text.primary
+              : theme.palette.text.disabled,
+          })}
+        >
+          {showDataFields ? <DataFieldIcon /> : <DataFieldOffIcon />}
```
jessicamcinchak, line 39: Open to MUI icon suggestions for this one especially - there's only so many that have "on"/"off" pairings ! 

But I think there's something sweet and hopefully not too far of a stretch about "auto" wand &rarr; "auto-answering" &rarr; data fields ?

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Question.tsx
```diff
@@ -80,9 +82,18 @@ const Question: React.FC<Props> = React.memo((props) => {
...
+              imageSource={props.data?.img}
+              imageAltText={props.data?.text}
+            />
+          )}
           {Icon && <Icon titleAccess={iconTitleAccess} />}
           <span>{props.text}</span>
         </Link>
+        {props.type !== TYPES.SetValue && props.data?.fn && (
+          <DataField value={props.data.fn} variant="parent" />
+        )}
```
jessicamcinchak, line 96: `props.data.fn` will _not_ catch all data fields on nodes (eg especially individual file types in FileUpload&Label, List component, etc) but it should catch all the major "top-level" data fields - which feels "good enough" for this experiment! 

Would be a nice exercise to apply some of Daf's thorough search mappings here in the future for more accurate coverage.

editor.planx.uk/src/pages/FlowEditor/components/FlowEditor/ToggleDataFieldsButton.tsx
ianjon3s: I'd say we need to be more literal as the magic element is an implied function of the data itself, rather than describing the function of the button.

Not perfect from a technical sense (code vs data), but `CodeIcon` and `CodeOffIcon` is a really simple icon combo that gives a stronger literal link to the 'thing' being toggled.

jessicamcinchak: Good shout - that's definitely a clearer option - updated!

jamdelion (approved):
Looking good to me! ✨

PR #3764, opened 2024-10-04
feat:

jessicamcinchak:
**High-level goals here:**
- `upcomingCardIds` is responsible for graph navigation based on _visited_ nodes only and does not check future node's data values for automation potential or queue up nodes based on optimistic automations
- Question, Checklist, and Filter components are automated _when_ they are reached, not any earlier. Each should leave an `auto: true` breadcrumb (Filters don't leave any breadcrumb currently)

**Todo:**
- [x] `upcomingCardIds`
- [ ] new help method which takes current node & passport and outputs ids of responses that can be automated
- [x] Questions
- [ ] Checklists
- [ ] Filters
- [ ] Update tests

**Test flow:** `testing/auto-answer-next`

PR #3797, opened 2024-10-10
feat: Filters support choosing any flagset category

jessicamcinchak:
**Changes:**
- Filter component supports picking from any available flagset categories & renders it correctly on the graph
  - It defaults to the "Planning permission" category so existing components should work without needing an "Update"
  - Uses same bare bones `<select />` styling as External Portal and Result modals, let's revisit styles later
- I have NOT touched the `collectedFlags` store method here, which means that all non-"Planning permission" filters won't actually "work" yet - in that they aren't correctly reading collected flags and still exclusively route down the "No flag result" branch
  - This will be addressed once rebased into #3764 (trying to make small PRs for review!)

Screenshot from 2024-10-10 13-43-02

Screenshot from 2024-10-10 13-32-27

### DIFF
--- a/editor.planx.uk/src/@planx/components/Filter/Editor.test.tsx
+++ b/editor.planx.uk/src/@planx/components/Filter/Editor.test.tsx
@@ -64,14 +64,13 @@ test("Adding a filter and selecting a flagset category", async () => {
   );
 });
 
-test.todo("Updating an existing filter to another category", async () => {
+test("Updating an existing filter to another category", async () => {
   const handleSubmit = vi.fn();
 
   setup(
     <Filter
       id="filterNodeId"
-      category="Listed building consent"
-      fn="flag"
+      node={mockExistingFilterNode}
       handleSubmit={handleSubmit}
     />,
   );
@@ -99,6 +98,15 @@ test.todo("Updating an existing filter to another category", async () => {
     ),
   );
 });
+
+const mockExistingFilterNode = {
+  data: {
+    fn: "flag",
+    category: "Listed building consent",
+  },
+  type: 500,
+  edges: ["flag1", "flag2", "flag3", "flag4", "blankFlag"],
+};
 
 const mockDefaultFlagOptions = [
   ...flatFlags.filter((flag) => flag.category === DEFAULT_FLAG_CATEGORY),
--- a/editor.planx.uk/src/@planx/components/Filter/Editor.test.tsx
+++ b/editor.planx.uk/src/@planx/components/Filter/Editor.test.tsx
@@ -1,0 +1,3 @@
+test.todo("Adding a filter");
+
+test.todo("Updating a filter");
--- a/editor.planx.uk/src/@planx/components/Filter/Editor.test.tsx
+++ b/editor.planx.uk/src/@planx/components/Filter/Editor.test.tsx
@@ -1,3 +1,55 @@
-test.todo("Adding a filter");
+import {
+  ComponentType as TYPES,
+  DEFAULT_FLAG_CATEGORY,
+  flatFlags,
+} from "@opensystemslab/planx-core/types";
+import { fireEvent, screen, waitFor } from "@testing-library/react";
+import React from "react";
+import { setup } from "testUtils";
+import { vi } from "vitest";
 
-test.todo("Updating a filter");
+import Filter from "./Editor";
+
+test("Adding a filter for the default flagset", async () => {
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/@planx/components/Filter/Editor.tsx
```diff
@@ -1,52 +1,83 @@
...
+            through the left-most matching flag option only.
+          </Typography>
+        </ModalSectionContent>
+        <ModalSectionContent title="Pick a flagset category">
+          <select
+            data-testid="flagset-category-select"
+            name="category"
+            value={formik.values.category}
+            onChange={formik.handleChange}
+          >
```
jessicamcinchak, line 72: My current thinking is that we can merge this to `main` with a simple `disabled` prop added here? 

Then I can rebase into #3764, correctly adjust filter automations, and un-disable? Does this sound solid, anything I might be forgetting?

Please test flows on this pizza with existing filter components to confirm they work correctly before any type of node "Update" !

jessicamcinchak:
[Try refactoring the Filter component to be a Question that auto-answers based on flag, rather than passort variable](https://trello.com/c/zb5ZcBhx/2244-try-refactoring-the-filter-component-to-be-a-question-that-auto-answers-based-on-flag-rather-than-passort-variable)

editor.planx.uk/src/@planx/components/Filter/model.ts
```diff
@@ -1,5 +0,0 @@
-export interface Filter {
-  id?: string;
-  handleSubmit?: (d: any, children: any) => void;
-  node?: any;
-}
```
jessicamcinchak, line 5: This was unused! Removed to avoid confusion, we're simply reading props defined at the top of Editor.tsx.

editor.planx.uk/src/@planx/components/Filter/Editor.tsx
```diff
@@ -1,52 +1,83 @@
...
+            {Array.from(categories).map((category) => (
```
RODO94, line 73: First time seeing this, its so slick with the new Set() above!

editor.planx.uk/src/@planx/components/Filter/Editor.tsx
RODO94, line 1: nit: I'd expect the Editor model to be similar to the Result component, where I can see the filter options before updating. I am an uneducated user, so it is feasible that this is redundant for people who know what to expect from the filters.

RODO94 (approved):
I ran through a prior approval check and from my understanding I ended up where I expected with the existing filters in place. Although it would be nice to have a session on this, _I'll put something in Slack_

Only one comment on the Editor Modal and one on using `fireEvent` in the tests. Both are nits and don't require action, more feedback on my expectations.

Rebasing on to the other PR mentioned sounded right and I couldn't think of anything missed here

editor.planx.uk/src/@planx/components/Filter/Editor.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const handleSubmit = vi.fn();
+
+  setup(<Filter handleSubmit={handleSubmit} />);
+
+  expect(screen.getByTestId("flagset-category-select")).toHaveValue(
+    DEFAULT_FLAG_CATEGORY,
+  );
+
+  fireEvent.change(screen.getByTestId("flagset-category-select"), {
+    target: { value: "Community infrastructure levy" },
```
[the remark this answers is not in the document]
jessicamcinchak: I'm using `fireEvent` throughout this file because the modal submit buttons like "Create filter" / "Update filter" aren't easily accessible to `user.click()` - therefore `fireEvent.submit()` is handy to have (this is commonly used throughout other "Editor" test files as well). 

I think the select interaction specifically could be migrated to `userEvent` going forward - which would mean we'd be testing the interaction more realistically to how the user would experience it in the browser.

editor.planx.uk/src/@planx/components/Filter/Editor.tsx
jessicamcinchak: Filter options are _not_ editable, they are hard-coded only still - therefore in my opinion there's no reason to display them in the modal here (they'd have to be disabled fields?). 

In the future, when we may have customisable filter options (aka flags and flagsets), then I imagine it will make sense to expose them directly here rather select from a pre-existing/pre-configured category ! But that's out of scope of these changes.

PR #3798, opened 2024-10-10
feat: Whole number option on Number Input

RODO94:
### What does this PR do?

This PR adds a switch option to the `<NumberInput>` editor modal to only allow whole numbers. 

A `test()` has been added to the validation schema to test that when `onlyWholeNumbers` is true, then `value` can only be an integer.

```ts
    .test({
      name: "check for a whole number",
      message: "Enter a whole number",
      test: (value: string | undefined) => {
        if (!value) {
          return false;
        }
        if (input.onlyWholeNumbers && !Number.isInteger(Number(value))) {
          return false;
        }
        return true;
      },
    });
```

### DIFF
--- a/editor.planx.uk/src/@planx/components/NumberInput/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/NumberInput/Editor.tsx
@@ -92,12 +92,9 @@ export default function NumberInputComponent(props: Props): FCReturn {
             <FormControlLabel
               control={
                 <Switch
-                  checked={formik.values.onlyWholeNumbers}
+                  checked={formik.values.isInteger}
                   onChange={() =>
-                    formik.setFieldValue(
-                      "onlyWholeNumbers",
-                      !formik.values.onlyWholeNumbers,
-                    )
+                    formik.setFieldValue("isInteger", !formik.values.isInteger)
                   }
                 />
               }
--- a/editor.planx.uk/src/@planx/components/NumberInput/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/NumberInput/Editor.tsx
@@ -80,7 +80,7 @@ export default function NumberInputComponent(props: Props): FCReturn {
                   onChange={() =>
                     formik.setFieldValue(
                       "allowNegatives",
-                      !formik.values.allowNegatives
+                      !formik.values.allowNegatives,
                     )
                   }
                 />
@@ -96,7 +96,7 @@ export default function NumberInputComponent(props: Props): FCReturn {
                   onChange={() =>
                     formik.setFieldValue(
                       "onlyWholeNumbers",
-                      !formik.values.onlyWholeNumbers
+                      !formik.values.onlyWholeNumbers,
                     )
                   }
                 />
--- a/editor.planx.uk/src/@planx/components/NumberInput/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/NumberInput/Editor.tsx
@@ -3,10 +3,7 @@ import Switch from "@mui/material/Switch";
 import { ComponentType as TYPES } from "@opensystemslab/planx-core/types";
 import type { NumberInput } from "@planx/components/NumberInput/model";
 import { parseNumberInput } from "@planx/components/NumberInput/model";
-import {
-  EditorProps,
-  ICONS,
-} from "@planx/components/ui";
+import { EditorProps, ICONS } from "@planx/components/ui";
 import { useFormik } from "formik";
 import React from "react";
 import { ModalFooter } from "ui/editor/ModalFooter";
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/NumberInput/Editor.tsx
```diff
@@ -83,14 +80,30 @@ export default function NumberInputComponent(props: Props): FCReturn {
...
                 />
               }
               label="Allow negative numbers to be input"
             />
           </InputRow>
+          <InputRow>
+            <FormControlLabel
+              control={
+                <Switch
+                  checked={formik.values.onlyWholeNumbers}
```
RODO94, line 95: Repeated `allowNegatives` switch styling

editor.planx.uk/src/@planx/components/NumberInput/Editor.tsx
```diff
@@ -83,14 +80,30 @@ export default function NumberInputComponent(props: Props): FCReturn {
...
+                  onChange={() =>
+                    formik.setFieldValue(
+                      "onlyWholeNumbers",
```
RODO94, line 98: Not wedded to the variable name or way I'm framing this option. Would appreciate alternative suggestions

editor.planx.uk/src/@planx/components/NumberInput/model.ts
```diff
@@ -52,6 +54,19 @@ export const numberInputValidationSchema = (input: NumberInput) =>
...
       },
+    })
+    .test({
+      name: "check for a whole number",
+      message: "Enter a whole number",
+      test: (value: string | undefined) => {
+        if (!value) {
+          return false;
+        }
+        if (!Number.isInteger(Number(value))) {
```
RODO94, line 65: This seemed most reliable in terms of validation while being related to assessing a number. I did have a string check in as well but this felt weird for a number check...

editor.planx.uk/src/@planx/components/NumberInput/model.ts
```diff
@@ -52,6 +54,19 @@ export const numberInputValidationSchema = (input: NumberInput) =>
         }
         return value === "0" ? true : Boolean(parseNumber(value));
       },
+    })
+    .test({
+      name: "check for a whole number",
+      message: "Enter a whole number",
+      test: (value: string | undefined) => {
+        if (!value) {
```
RODO94, line 62: Added this to cover the `string | undefined' type without mudding conditional check on the `Number.isInteger`

editor.planx.uk/src/@planx/components/NumberInput/model.ts
```diff
@@ -28,6 +29,7 @@ export const parseNumberInput = (
   fn: data?.fn || "",
   units: data?.units,
   allowNegatives: data?.allowNegatives || false,
+  onlyWholeNumbers: data?.onlyWholeNumbers || false,
```
RODO94, line 32: Not wedded to this variable naming, happy to receive suggestions

editor.planx.uk/src/@planx/components/NumberInput/model.ts
```diff
@@ -52,6 +54,19 @@ export const numberInputValidationSchema = (input: NumberInput) =>
...
       },
+    })
+    .test({
+      name: "check for a whole number",
+      message: "Enter a whole number",
+      test: (value: string | undefined) => {
+        if (!value) {
+          return false;
+        }
+        if (input.onlyWholeNumbers && !Number.isInteger(Number(value))) {
```
RODO94, line 65: I tried a few different ways of validating the whole number. Using the `Number.isInteger()` seemed the most relevant to what I wanted to check, rather than a string check for a decimal point or anything else.

RODO94:
[Add a toggle to NumberInput to enforce whole numbers and prevent decimals (eg "How many bedrooms?")](https://trello.com/c/ovbVCDgw/3019-add-a-toggle-to-numberinput-to-enforce-whole-numbers-and-prevent-decimals-eg-how-many-bedrooms)

editor.planx.uk/src/@planx/components/NumberInput/Public.test.tsx
```diff
@@ -84,6 +84,56 @@ test("allows negative numbers to be input when toggled on by editor", async () =
...
+
+  const textArea = screen.getByLabelText("What's the temperature?");
+
+  await user.type(textArea, "10.06");
+  await user.click(screen.getByTestId("continue-button"));
+
+  expect(screen.getByText("Enter a whole number")).toBeInTheDocument();
+  expect(handleSubmit).not.toHaveBeenCalledWith({
+    data: { fahrenheit: 10.06 },
+  });
```
jessicamcinchak, line 111: nit: `not.toHaveBeenCalled()` might be more meaningful & definitive check here? Because we want to check if the error _blocks_ submission, not that it was possibly automatically rounded to `fahrenheit: 10` and successfully submitted for example!

editor.planx.uk/src/@planx/components/NumberInput/model.ts
jessicamcinchak: `Number.isInteger()` makes sense to me !

jessicamcinchak (approved):
All working as expected on pizza & code looks good - nice & simple solution :raised_hands: 

One small test suggestion.

editor.planx.uk/src/@planx/components/NumberInput/model.ts
DafyddLlyr: Maybe `format: "integer" | "float"` is a little more explicit? Or even `isInteger: boolean`?

PR #3803, opened 2024-10-12
[api] security hardening for microsoft auth

freemvmt:
Ticket: https://trello.com/c/0gYN7cdt

Some changes to improve the security posture of the Microsoft auth flow.

### DIFF
--- a/api.planx.uk/modules/auth/controller.ts
+++ b/api.planx.uk/modules/auth/controller.ts
@@ -39,6 +39,7 @@ function setJWTCookie(returnTo: string, res: Response, req: Request) {
     maxAge: new Date(
       new Date().setFullYear(new Date().getFullYear() + 1),
     ).getTime(),
+    // pizzas rely on staging API for auth (due to static redirect URIs), so we have to allow cross-site
     sameSite: "none",
     secure: true,
   };
--- a/api.planx.uk/modules/auth/strategy/microsoft-oidc.ts
+++ b/api.planx.uk/modules/auth/strategy/microsoft-oidc.ts
@@ -1,3 +1,4 @@
+import crypto from "crypto";
 import type {
   Client,
   ClientMetadata,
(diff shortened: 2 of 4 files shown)


api.planx.uk/server.ts
```diff
@@ -112,7 +112,8 @@ assert(process.env.UNIFORM_SUBMISSION_URL);
 // needed for storing original URL to redirect to in login flow
 app.use(
   cookieSession({
-    maxAge: 24 * 60 * 60 * 100,
```
freemvmt, line 115: value was previously set at 2.4hrs but I think was *supposed* to be 24hrs (assuming `100` was intended as `1000`), so new value is only a slight reduction, and also clearer to read

api.planx.uk/modules/auth/strategy/microsoft-oidc.ts
```diff
@@ -41,18 +42,30 @@ export const getMicrosoftOidcStrategy = (client: Client): Strategy<Client> => {
 };
 
 const verifyCallback: StrategyVerifyCallbackReq<Express.User> = async (
-  req: Http.IncomingMessageWithSession,
+  req: Http.IncomingMessageWithCookies,
   tokenSet,
   done,
 ): Promise<void> => {
   // TODO: use tokenSet.state to pass the redirectTo query param through the auth flow
+  // TODO: validate id_token sig with the public key from the jwks_uri (...v2.0/keys)
```
freemvmt, line 50: It's not especially clear how to do this using `openid-client` v5x, but they completely rewrote the package for v6 and that might offer a clearer route to fetching the keys and verifying the signature

Related task in icebox: https://trello.com/c/7BMpYwyt

api.planx.uk/modules/auth/controller.ts
```diff
@@ -39,7 +39,8 @@ function setJWTCookie(returnTo: string, res: Response, req: Request) {
     maxAge: new Date(
       new Date().setFullYear(new Date().getFullYear() + 1),
     ).getTime(),
-    sameSite: "none",
+    // the JWT/auth cookies should be sent only between the API server and editor
+    sameSite: "strict",
```
freemvmt, line 43: Please correct me if this doesn't make sense!

freemvmt (commented):
Some notes

api.planx.uk/modules/auth/controller.ts
DafyddLlyr: I think this needs to be "none" for login to work between the Pizzas and Staging API. Worth rebasing this once https://github.com/theopensystemslab/planx-new/pull/3890 is merged and testing?

api.planx.uk/server.ts
DafyddLlyr: Good catch!

api.planx.uk/modules/auth/controller.ts
freemvmt: Riiight, that's the motivation for `none`, I hadn't thought about that case. Although this change should affect both means of authenticating, and Google seems to work on the [pizza](https://3803.planx.pizza/) 🤔

freemvmt: Will do as you suggest and see what happens

DafyddLlyr:
@freemvmt - one thing I just realised here! 

The changes to the API files here aren't being run when we're logging into the pizza as we're hitting the _staging_ API for logging in. So that means in order to test this, we need to merge to `main` and then try to log in on a Pizza.

I'm struggling to find the PR but I'm pretty sure that `sameSite: "none"` is required from previously trying this out.

freemvmt:
> The changes to the API files here aren't being run when we're logging into the pizza as we're hitting the _staging_ API for logging in. So that means in order to test this, we need to merge to `main` and then try to log in on a Pizza.
> 
> I'm struggling to find the PR but I'm pretty sure that `sameSite: "none"` is required from previously trying this out.

@DafyddLlyr that's a great point... I think I'll just take your word for it!

I was testing out `lax` because we should be sent to staging API via a top level, user-requested navigation step, which I'd expect it to allow, but who knows if that actually works when it's React 🤷🏼 

And perhaps now isn't the time to find out. I'll swap out for `none` and we can merge :)

freemvmt:
Closed in favour of perfect copy #3895 - @DafyddLlyr please approve that one instead!

PR #3804, opened 2024-10-14
[docs] Add how-to for creating a team secret

jamdelion:


jessicamcinchak (approved):
Perfect thank you! 

Also fun trick (but one I always forget about!) - when updating docs only, if you include `[skip pizza]` in your commit message it'll skip Vultr steps for speedier CI - useful when changes don't require a build! :slightly_smiling_face:

PR #3811, opened 2024-10-16
feat: add detailed steps on adding public analytics links to Hasura

zz-hh-aa:
- Breaks down the previous step on adding public links to Hasura into detailed sub-steps 
- Uploads screenshots too


DafyddLlyr (approved):
Nice - thanks!

PR #3812, opened 2024-10-16
refactor: Simplify editor toggle button group

ianjon3s:
## What does this PR do?

- Updates editor toggle button group to have a fixed position and contain inner buttons to prevent having to set a fixed bottom/left distance value and replicated styling for each button
- Allows further buttons to be added without needing specific styling
- Unifies heights for each button

**Before (left) vs after (right):**
image

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/index.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/index.tsx
@@ -27,6 +27,18 @@ const EditorContainer = styled(Box, {
     : `calc(100vh - ${HEADER_HEIGHT_EDITOR}px)`,
 }));
 
+const EditorVisualControls = styled(ButtonGroup)(({ theme }) => ({
+  position: "fixed",
+  bottom: theme.spacing(2.5),
+  left: theme.spacing(7.5),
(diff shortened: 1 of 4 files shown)


editor.planx.uk/src/pages/FlowEditor/index.tsx
```diff
@@ -27,6 +27,18 @@ const EditorContainer = styled(Box, {
...
+  position: "fixed",
+  bottom: theme.spacing(2.5),
+  left: theme.spacing(7.5),
+  zIndex: theme.zIndex.appBar,
+  border: `1px solid ${theme.palette.border.main}`,
+  borderRadius: "3px",
+  background: theme.palette.border.main,
+  gap: "1px",
+  overflow: "hidden",
+}));
```
jessicamcinchak, line 40: Much nicer, thank you !!

jessicamcinchak (approved):

editor.planx.uk/src/pages/FlowEditor/index.tsx
ianjon3s: Thanks @jessicamcinchak

PR #3815, opened 2024-10-16
feat: add metabase token and base url placeholders to dotenv template

zz-hh-aa:
- Adds `METABSE_API_KEY` and `METABASE_BASE_URL` to `api.planx.uk/.env.test.example` template


DafyddLlyr (changes requested):
Two small comments!

- I think we can group these within the `# Metabase analytics` section above
- Can `METABASE_BASE_URL` can be set to `METABASE_URL_EXT` (which matches the API and Editor) and derived as `METABASE_URL_EXT=http://localhost:${METABASE_PORT}`
  - This is quite important as this value needs to change across local dev and pizzas (docker) and then on Pulumi (staging & production)
  - If you take a look at other references to `API_URL_EXT` you should see it has different formats in different environments

zz-hh-aa:
Thanks @DafyddLlyr, I _think_ I've made those changes, if I've understood you correctly?

.env.example
```diff
@@ -51,6 +51,8 @@ SLACK_WEBHOOK_URL=👻
 
 # Metabase analytics
 METABASE_PORT=5000
+METABASE_API_KEY=👻
+METABASE_URL_EXT=👻
```
DafyddLlyr, line 55: This is actually a genuine environmental variable and not a secret, so I think we're fine to just do the following here - 
```suggestion
METABASE_URL_EXT=http://localhost:${METABASE_PORT}
```

DafyddLlyr (changes requested):

.env.example
DafyddLlyr: (Sorry if that wasn't clear from above!)

This just means that we have one fewer secret to manage which is always a bonus 👍

zz-hh-aa:
Gotcha, now updated!

DafyddLlyr (approved):
Perfect - thanks for making those changes! 🙌

PR #3816, opened 2024-10-16
feat: Description text for map schema date input

ianjon3s:
## What does this PR do?

- Adds a description/hint text to map date inputs `For example, 16 04 2027`, in line with [GOV.UK example](https://design-system.service.gov.uk/components/date-input/).

Identified in user-research, 2/5 householder participants tried to enter months using text (i.e. "Nov").

A first instinct was to add a script to insert an example date [for example] 6 months in the future, so that the suggested date would be relevant(ish) and never "expire". Seems like something that can be solved by setting a date well into the future to save unnecessary functionality. Interested to hear other opinions on this.

**Preview:**
image

### DIFF
--- a/editor.planx.uk/src/@planx/components/List/schemas/Trees.ts
+++ b/editor.planx.uk/src/@planx/components/List/schemas/Trees.ts
@@ -57,6 +57,7 @@ export const Trees: Schema = {
       type: "date",
       data: {
         title: "Expected completion date",
+        description: "For example, 16 04 2027",
         fn: "completionDate",
       },
     },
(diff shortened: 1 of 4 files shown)


DafyddLlyr (approved):
Good catch!

We could take this a step further by either making the description field required, or setting a default value.

If we wanted to use a default value we could set this in `DateFieldInput.tsx`.

To make it required (in List/Schema contexts) we'd need to update the `DateField` type in `/Schema/model.ts`, something like - 

```ts
export type DateField = {
  type: "date";
  data: DateInput & { fn: string, description: string };
};
```

ianjon3s:
PR was closed in error by accidentally deleting branch.

jessicamcinchak:
Merging this one as-is - will make a new Trello ticket to further discuss "required" placeholders for Date Inputs :+1:

PR #3820, opened 2024-10-16
feat: allow options to assign many flags

jessicamcinchak:
Following on from #3764 

**Gist:** 
- Question & Checklist "Options" currently support a single data field `val` and single `flag` each - as we think about the future of automations & filtering, there's actually no reason a single Option shouldn't be able to support _multiple_ flags
  - For example, answering "No" might flag up "Planning permission needed" AND "Listed building consent required" and it's ideal for editors to be able to set _both_ in a single go (each Options still supports a single data field `val`)
- Editors should ideally only select up to one flag _per category_
  - In a followup PR we could build nice validation error handling for this
  - But the cool thing is that it's actually harmless if they select more than one in the same category, it will simply be treated like flags within the same category are always/already treated - only the left-most, highest order flag matters! 

**Changes:**
- [x] Update `Option` node data type `flag` from `string` to `Array<Flag["value"]>`
- [x] Implement `MultipleSelect` with grouping by category for editor "Option" modals within Questions & Checklists
- [x] Update graph node styling with color band _per_ flag
- [x] Update preview store methods & associated tests
  - [x] `autoAnswerableFlag` for Filter node navigation
  - [x]  `getResultData`/`resultData` for Result node navigation and analytics tracking
  - [x] `collectedFlags` (no component or navigation dependencies)
- [x] Expose `collectedFlags` in editor Console side panel

**Open question:**
** All of the above work with _existing_ Option's `flag` (`string`) and new/updated Option's `flag` (`string[]`) without _requiring_ a content migration of existing flow content (or all existing mock test data!). 
  - But are we happy with slightly semantically incorrect type "flag" for a _list of values_? 
  - Do we still want to consider a content migration script across ALL flow's to rename `flag` to `flags` and format legacy values as nice-to-have? (Might be good practice for clones?)


jessicamcinchak:
[Collect flags in the passport](https://trello.com/c/3nYW4gyD/2243-collect-flags-in-the-passport)

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -382,7 +374,74 @@ export const previewStore: StateCreator<
 
   resultData(flagSet, overrides) {
     const { breadcrumbs, flow } = get();
-    return getResultData(breadcrumbs, flow, flagSet, overrides);
```
jessicamcinchak, line 385: Decided to simplify this into a single function rather than this previous pattern of `resultData` calling `getResultData` - cleaning up another `reduce` along the way (we aren't even dealing with an _array_ of anything in this method because a single flagset category is passed in :see_no_evil:) !

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -885,3 +871,43 @@ export const removeNodesDependentOnPassport = (
   }, [] as string[]);
   return { removedNodeIds, breadcrumbsWithoutPassportData: newBreadcrumbs };
 };
+
+const collectedFlagValuesByCategory = (
```
jessicamcinchak, line 875: No logical change here - this has simply been extracted out of `autoAnswerableFlag` so we can re-use it in `collectedFlags` and `resultData` too

jessicamcinchak:
[Logic followups](https://trello.com/c/dboRkua3/3112-logic-followups-jess-to-add-checklist)

editor.planx.uk/src/@planx/components/Checklist/Editor.tsx
```diff
@@ -295,14 +292,14 @@ export const ChecklistComponent: React.FC<ChecklistProps> = (props) => {
               ...values,
               ...(groupedOptions
                 ? {
-                    categories: groupedOptions.map((gr) => ({
-                      title: gr.title,
-                      count: gr.children.length,
-                    })),
-                  }
+                  categories: groupedOptions.map((gr) => ({
```
jamdelion, line 295: nit: what do you think about changing these acronyms (`gr`) to something a bit more meaningful? 

I just find it slightly increases the mental load unnecessarily to read something that isn't immediately obvious (like I understood it was a group on second reading, but I think making code as frictionless and easy to read as possible is something to strive for!)

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Option.tsx
```diff
@@ -36,7 +46,7 @@ const Option: React.FC<any> = (props) => {
             imageAltText={props.data.text}
           />
         )}
-        <div className="band" style={{ background, color }}></div>
+        {flags ? flags.map((flag) => <FlagBand key={`${props.id}-${flag.value}`} flag={flag} />) : <NoFlagBand />}
```
jamdelion, line 49: Nice clear component names 😍

jamdelion (approved):
I know Daf said he would review this too but I felt nosy 😆 Looks good to me but obviously feel free to wait for his review as he probably has more context!

editor.planx.uk/src/@planx/components/Checklist/Editor.tsx
jessicamcinchak: Fully agree - updated now !

DafyddLlyr (approved):
Working as expected, nice and clear to follow 👍 

A few minor suggestions, no showstoppers.

On the whole flag/flags thing I think we can live with this as is for now (`flag` pointing to an array) but it would be nice to handle a migration here for sure - it's nice simple use case as well.

One other comment would be on the validation - I think it's worth making the effort to enforce one per category just from the point of view of explaining this to users. I think it could be as simple as disabling the `renderOption` if `flag` from the same category is already in the group. We could also handle this in the validation schema. I'm aware that August flagged this as a blocker so happy for this to be a follow up here for sure 👍

editor.planx.uk/src/@planx/components/shared/FlagsSelect.tsx
```diff
@@ -0,0 +1,78 @@
+import { AutocompleteChangeReason, AutocompleteProps } from "@mui/material/Autocomplete";
+import Chip from "@mui/material/Chip";
+import ListItem from "@mui/material/ListItem";
+import { Flag, flatFlags } from "@opensystemslab/planx-core/types";
+import React, { useMemo } from "react";
+import InputRow from "ui/shared/InputRow";
+import { CustomCheckbox, SelectMultiple } from "ui/shared/SelectMultiple";
+
+interface Props {
+  value?: Array<Flag["value"]>;
```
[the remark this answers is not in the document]
jessicamcinchak: Annoying yes because `Flag["value"]` is optional, despite it always be populated/required in practice. Will try to do a more thorough revisit of types as a followup :+1:

PR #3823, opened 2024-10-17
feat: add Boundary to Map and Label comp

RODO94:
Coming from user feedback contained in the Trello Ticket: https://trello.com/c/liDbYC96/3094-both-options-show-static-boundary-or-address-point-for-reference-using-geojsondata-prop?filter=member:rorydoak2

We have added the boundary to map components so users can clearly see their property boundary when adding trees. 

Main changes are using the `geoJsonData` prop to populate a readonly boundary using passport data

https://github.com/user-attachments/assets/ad6067a3-24b4-4404-8c0c-13b7bfcddc53

### DIFF
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
@@ -240,6 +240,10 @@ const Root = () => {
       `You must plot at most ${schema.max} ${schema.type}(s) on the map`) ||
     "";
 
+  const [passport] = useStore((state) => [state.computePassport()]);
+
+  console.log(passport);
+
   return (
     <Card handleSubmit={validateAndSubmitForm} isValid>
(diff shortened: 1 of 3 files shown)


jessicamcinchak (approved):
Works as expected for me !

editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
```diff
@@ -227,6 +227,8 @@ const Root = () => {
     previouslySubmittedData,
   } = mapAndLabelProps;
 
+  const [passport] = useStore((state) => [state.computePassport()]);
```
DafyddLlyr, line 230: fyi: If we're just grabbing a single value, we can use the following notation - 

```suggestion
  const passportData = useStore((state) => state.computePassport().data);
```

DafyddLlyr (approved):

PR #3825, opened 2024-10-18
fix: Map and Label breaking on back action

RODO94:
The initial bug centred on navigating back after completing the Map and Label component, and finding the active tab not selected, then when adding a new point, the tabs would break. My solution was to change how the ActiveIndex was assigned and handled, and when the previous data from the passport was added to `features`. From this, I found a couple other bugs on how we handle labels when copying data.

Should all be fixed now with testing around navigating back. There is no test coverage right now for navigating back, adding new data, continuing forward, then navigating back again.

I would be interested to know if I should be rendering a different component within the tests so I can test clicking "Continue" then clicking "Back" - at the moment, I just recreate rendering a component with the `previouslySubmittedData`. Would be great to have feedback on whether this would be overkill.


jessicamcinchak (commented):
Looks like you're on the right track for this bug! Know this is still draft, but want to flag what's seeming like a possible new side-effect bug of your changes: 

Here's steps to recreate:
- Navigate forwards to Map&Label & draw two features
- Click "reset all" button the map -> features _and_ tabs should both clear
- Draw two features again
- "Continue" forwards this time
- Navigate "back" to Map&Label
- Confirm I can add another feature to map and tab is added (works now!)
- Click "reset all" button on the map -> map is cleared _but the tabs are not_ (new bug!) 

Can you recreate too?

First guess would be to double check that any changes you're making to the map context methods stay in sync with their form method counterpart !

editor.planx.uk/src/@planx/components/MapAndLabel/Public/Context.tsx
```diff
@@ -99,11 +99,14 @@ export const MapAndLabelProvider: React.FC<MapAndLabelProviderProps> = (
   });
 
   const [activeIndex, setActiveIndex] = useState<number>(
-    previousFormData?.length - 1 || -1,
+    previousFormData ? previousFormData?.length - 1 : 0,
   );
 
   const [minError, setMinError] = useState<boolean>(false);
   const [maxError, setMaxError] = useState<boolean>(false);
+  const [loadPreviousValues, setLoadPreviousValues] = useState<boolean>(
```
RODO94, line 107: Went round the blocks with this trying to find an elegant solution, this felt like the best option to ensure previous values only loaded once

editor.planx.uk/src/@planx/components/MapAndLabel/Public/Context.tsx
```diff
@@ -198,7 +206,10 @@ export const MapAndLabelProvider: React.FC<MapAndLabelProviderProps> = (
 
   const copyFeature = (sourceIndex: number, destinationIndex: number) => {
     const sourceFeature = formik.values.schemaData[sourceIndex];
-    formik.setFieldValue(`schemaData[${destinationIndex}]`, sourceFeature);
+    formik.setFieldValue(`schemaData[${destinationIndex}]`, {
+      ...sourceFeature,
+      label: `${destinationIndex + 1}`,
```
RODO94, line 211: The label was being assigned based on the object you were copying from rather than the object we were copying to

editor.planx.uk/src/@planx/components/MapAndLabel/test/mocks/Trees.ts
```diff
@@ -20,3 +24,13 @@ export const props: PresentationalProps = {
   longitude: -0.1629784,
   latitude: 51.5230919,
 };
+
+export const previousSingleDataProps: PresentationalProps = {
+  ...props,
+  previouslySubmittedData: previouslySubmittedSingleData,
```
RODO94, line 30: The bug was different if there was a single data point added or if there was more than one, thus I added two separate bug tests

editor.planx.uk/src/@planx/components/MapAndLabel/test/public.test.tsx
RODO94, line 1: This file was becoming quite long, so it felt right to move the back navigation tests to separate file, and relocate these files to the `test` folder on the Map and Label component

editor.planx.uk/src/@planx/components/MapAndLabel/Public/Context.tsx
jessicamcinchak: Good catch but I don't think this is quite working in all scenarios correctly yet! 

For example: 
- I add two trees on the map
- I fill out tree two tab
- On tab 1, I "copy from" tree 2
- I "continue" & my labels are correct
- I come "back" and "remove tree" from tab 1 - this shifts the map & tab labels for previous "tree 2" to 1 now
- I "continue" again and my passport data has a single feature with `label: 2` when it should be `label: 1` to reflect map/tab changes after a point was removed !

editor.planx.uk/src/@planx/components/MapAndLabel/test/public.backNavigation.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    const firstTab = queryByRole("tab", { name: /Tree 1/ });
+    const secondTab = queryByRole("tab", { name: /Tree 2/ });
+    expect(firstTab).toBeInTheDocument();
+    expect(secondTab).toBeInTheDocument();
+
+    const secondTabPanel = getByTestId("vertical-tabpanel-1");
+    expect(secondTabPanel).toBeVisible();
+
+    addMultipleFeatures([point1, point2, point3]);
```
jessicamcinchak, line 68: nit: aren't `point1` and `point2` already on the map? Am we actually mocking adding "multiple" features here or just one? I understand the mock data needs to reflect the full feature collection, but I wonder if there's a way we cna make this more clear in helper function names here etc?

editor.planx.uk/src/@planx/components/MapAndLabel/test/public.backNavigation.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+    window.customElements.define("my-map", MyMap);
+  }
+
+  const ResizeObserverMock = vi.fn(() => ({
+    observe: vi.fn(),
+    unobserve: vi.fn(),
+    disconnect: vi.fn(),
+  }));
+
+  vi.stubGlobal("ResizeObserver", ResizeObserverMock);
```
jessicamcinchak, line 25: nit: is `ResizeObserverMock` actually required by these tests / called in this file? Do these tests still pass if `beforeAll` only does the first definition step for `my-map` ?

jessicamcinchak (commented):
On the right track - pizza is working more smoothly, but a few questions/comments while reading through the code!

editor.planx.uk/src/@planx/components/MapAndLabel/Public/Context.tsx
RODO94: From what I could understand, `previouslySubmittedData` is always present once you navigate backwards, so when you add a new point the code will run again and it will add the new point to `features` then hit the `if()` statement then reassign `features` to be `previouslySubmittedData.data...`

I initially tried to place it in a `useEffect()` to only run the `if` statement on initial mount then whenever `previouslySubmittedData` changed, but I couldn't get it to work so easily, and it felt like a larger change than the state variable.

Have I missed a simpler solution?

editor.planx.uk/src/@planx/components/MapAndLabel/Public/Context.tsx
RODO94: Thanks so much for catching this, I'll add some adjustments

editor.planx.uk/src/@planx/components/MapAndLabel/test/public.backNavigation.test.tsx
RODO94: It fails when it isn't defined, not fully sure why. I'll need to look back at Daf setting up the test.todos

editor.planx.uk/src/@planx/components/MapAndLabel/test/public.backNavigation.test.tsx
RODO94: This is correct, it should just be adding one feature at a time. I'll adjust and switch the functions

editor.planx.uk/src/@planx/components/MapAndLabel/test/public.backNavigation.test.tsx
jessicamcinchak: No worries about tracking down more detailed answer for this - thanks for simply confirming it's necessary !

editor.planx.uk/src/@planx/components/MapAndLabel/test/public.backNavigation.test.tsx
RODO94: Found that this was correct and just adding a single point deletes the others from the map, so I need them added like this. I've added a comment above each line explaining this need

editor.planx.uk/src/@planx/components/MapAndLabel/Public/Context.tsx
jessicamcinchak: Thanks for explanation - happy to keep this approach then but can we please rename to `loadPreviousValuesOnMap` for clarity? Because it's never actually handling the form/tabs part based on how I'm reading it !

jessicamcinchak (approved):
Working as expected for me now - thanks for iterating on this one !

PR #3826, opened 2024-10-18
fix: Correctly set parent flow name in GovPayMetadata

DafyddLlyr:
## What's the problem?
- The Editor interface of the Pay component only ever knows about its own context - `flowName` will always be the flow it's held within
- At runtime, as an applicant hits the public interface of the component, the context can be different - `flowName` is the root ("parent" flow)
- This means that we send the incorrect flow name to GovPay

## What's the solution?
- Be a bit more smart about setting the defaults at runtime in the Public interface
- Always overwrite with the _current_ flow name, which is always correct

## Testing
I didn't find a good way to test this (without setting up a new E2E test). Open to ideas!

You can test this manually here - 
 - Public interface - https://3826.planx.pizza/buckinghamshire/test-pay-metadata/published?analytics=false
 - Editor interface - https://3826.planx.pizza/buckinghamshire/test-pay-metadata

### DIFF
--- a/editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
@@ -67,14 +67,14 @@ function Component(props: Props) {
     state.flowSlug,
   ]);
   const fee = props.fn ? Number(passport.data?.[props.fn]) : 0;
+
   const defaultMetadata = [
     { key: "source", value: "PlanX" },
     { key: "flow", value: flowSlug },
     { key: "paidViaInviteToPay", value: "@paidViaInviteToPay" },
   ];
-  const metadata = props.govPayMetadata?.length
-    ? props.govPayMetadata
-    : defaultMetadata;
+
+  const metadata = [...(props.govPayMetadata || []), ...defaultMetadata];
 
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
```diff
@@ -67,14 +67,14 @@ function Component(props: Props) {
...
   const defaultMetadata = [
     { key: "source", value: "PlanX" },
     { key: "flow", value: flowSlug },
     { key: "paidViaInviteToPay", value: "@paidViaInviteToPay" },
   ];
-  const metadata = props.govPayMetadata?.length
-    ? props.govPayMetadata
-    : defaultMetadata;
+
+  const metadata = [...(props.govPayMetadata || []), ...defaultMetadata];
```
DafyddLlyr, line 77: Before - defaults were only applied if nothing was set in the Editor.

Now - defaults overwrite any value set in the Editor, and are always correct and up to date.

jessicamcinchak (approved):
Pay-in-a-portal is definitely a new-to-me content pattern! Thanks for getting to bottom of this one, works for me :heavy_check_mark:

PR #3828, opened 2024-10-21
fix: Do not validate submitted or deleted sessions

DafyddLlyr:


jessicamcinchak (approved):
Thanks again for looking into this one - agree it's very strange !

PR #3837, opened 2024-10-22
refactor: reorganise feedback component folder

jamdelion:
Was finding all the random storybook files quite confusing, so essentially all I've done is move each component into its own folder: 

Screenshot 2024-10-22 at 14 58 59

Also took the opportunity to shorten the length of `index.ts` by extracting the types and styled components out to separate files. 

Wasn't sure of the distinction between index and FeedbackPhaseBanner from the storybook point of view so have left those in the original folder for now.


jamdelion:
Waiting for https://github.com/theopensystemslab/planx-new/pull/3838 to merge before I sort the Vultr job here, as I think it is trying to use the same pizza...

DafyddLlyr (approved):

PR #3841, opened 2024-10-23
feat: add basic in-page feedback component

jamdelion:
This is currently just adding the first pass at the Feedback component in storybook.

Screenshot 2024-10-24 at 17 22 49

### Next steps 
In future PRs here are the todos:
- Add in the Editor side
- Form validation
- Error handling

### Links
Trello ticket: https://trello.com/c/jEOLoubW/3034-can-we-introduce-in-house-end-of-service-feedback?filter=member:johumphrey6

Figma design: https://www.figma.com/design/bnUUrsVRG6qPwDkTmVKACI/PlanX-Design?node-id=16904-1689&t=mx6FzCppbhIfPsBM-4


DafyddLlyr (changes requested):
Nice to see a Storybook first approach! 🙌 

One small nit is that the hover shadows should match the border - 

image

When it comes to the next steps of configuring the component, there's [docs on this here](https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/docs/adding-a-new-component.md#L0-L1).

DafyddLlyr (changes requested):
One small little snag - getting some weird hover behaviour here. It looks like when I hover over the button group, the first item has a hover style applied. Can you recreate this?


https://github.com/user-attachments/assets/bdc73b26-0c59-4382-ab5e-f8ca5ae5b46a

jamdelion:
> One small little snag - getting some weird hover behaviour here. It looks like when I hover over the button group, the first item has a hover style applied. Can you recreate this?
> 
>  Screen.Recording.2024-10-28.at.14.44.28.mov

Nice catch, I'm seeing that on my side now as well! Will investigate... 👀

editor.planx.uk/src/@planx/components/Feedback/Public.tsx
```diff
@@ -0,0 +1,119 @@
...
+        experience will help us to improve it.
+      </Typography>
+      <Typography pt={2}>
+        Don't share any personal or financial information in your feedback. If
+        you do we will act according to our{" "}
+        <Link href={props.privacyPolicyLink ?? ""}>privacy policy</Link>.
+      </Typography>
+
+      <Box pt={2} mb={3}>
+        <InputLabel label="How would you rate your experience with this service?" />
```
jamdelion, line 59: The hover issue was caused by having more than one child inside the label component, as discussed in this issue: https://github.com/mui/material-ui/issues/14543

I've made the `children` prop optional in the `InputLabel` component to get round this.

DafyddLlyr: Ah nice one 👍 

Alternatively, I _think_ we could make children a `ReactElement` which would force a single element or fragment here. This will likely have cascading type issues but may point at other places where we're hitting a similar hover issue?

The a11y tests are still passing in CI and Storybook without the wrapping label, so this is all good as-is.

DafyddLlyr (approved):
Perfect! Thanks for the quick fixes here.

PR #3843, opened 2024-10-24
feat: Remove duplicated `Destination` enums, add `Send` test coverage

DafyddLlyr:
## What does this PR do?
 - Imports and uses `SendIntegration` type from `planx-core` (please see https://github.com/theopensystemslab/planx-core/pull/540)
 - Implements new types in the API, removing the `Destination` enum
 - Implements new types in the frontend, removing the `Destination` enum
 - Fixes breadcrumb generation for `Send` component - currently this does not work on staging - due to an incorrect type the breadcrumb remains empty
   - Demo - https://editor.planx.dev/buckinghamshire/test-send
   - Please see https://github.com/theopensystemslab/planx-new/pull/3843/files#r1814522369

### Also...
Picked up a few associated refactors in the frontend `Send` component, alongside adding some test coverage for the `Send` component.

## What's the motivation here?
As part of the GOSS integration I'm aiming to write some docs on "how to set up a send integration". This work is aiming to simplify this process so that it hopefully boils down to just the following - 
 - Add to `SendIntegration` array in `planx-core`
 - Write API to handle integration
 - Add secrets
 - Add label in `planx-new` Editor component

Hopefully then TypeScript will do a lot of the heavy lifting for the places where there are references to `SendIntegrations` that need to be updated.

There's still some room for improvement here I think - there's repeated logic in a few places still, but I'll pick up additional changes in another PR if they feel necessary. This is a big enough PR for now!


editor.planx.uk/src/@planx/components/Send/Public.tsx
```diff
@@ -61,10 +67,9 @@ const CreateSendEvents: React.FC<Props> = ({
   ]);
 
   // Send makes a single request to create scheduled events in Hasura, then those events make the actual submission requests with retries etc
-  const url = `${
-    import.meta.env.VITE_APP_API_URL
-  }/create-send-events/${sessionId}`;
-  const request: any = useAsync(async () => {
```
DafyddLlyr, line 67: This `any` was the cause of a bug where event ids are not logged to the breadcrumb.

The list below has `request.value.bops`, `request.value.uniform` etc, but it should be `request.value.data.bops`, `request.value.data.uniform` etc

editor.planx.uk/src/@planx/components/Send/mocks/hasuraEventsResponseMock.ts
DafyddLlyr, line 1: These ids will be UUIDs, just correctly updating mock here.

editor.planx.uk/src/@planx/components/Send/Public.test.tsx
```diff
@@ -1,36 +1,228 @@
...
+
+describe("Uniform overrides for Buckinghamshire", () => {
+  it("converts property.localAuthorityDistrict to the correct format", async () => {
+    act(() => setState({ 
+      teamSlug: "buckinghamshire", 
+      flow,
+      breadcrumbs: {
+        findProperty: {
+          data: {
+            "property.localAuthorityDistrict": ["Some local authority district"]
```
jessicamcinchak, line 147: nit: more realistic mock here would be `["Buckinghamshire", "Historic district name"]` - eg because I think we rely on being able to _filter council name out of_ `property.localAuthorityDistrict` array !

editor.planx.uk/src/@planx/components/Send/Public.test.tsx
```diff
@@ -1,36 +1,228 @@
...
+  await waitFor(() => expect(mockAxios.post).toHaveBeenCalledTimes(1));
+  expect(handleSubmit).toHaveBeenCalledTimes(1);
+
+  const breadcrumb = handleSubmit.mock.calls[0][0];
+
+  expect(breadcrumb.data).toEqual(expect.objectContaining({
+    bopsSendEventId: hasuraEventsResponseMock.bops.event_id,
+    uniformSendEventId: hasuraEventsResponseMock.uniform.event_id,
+  }));
+});
```
jessicamcinchak, line 219: Test coverage is long overdue, clear & thorough here, thank you!!

jessicamcinchak (approved):
Nice simplifications throughout here, thanks ! 

Code looks good, but I admittedly have _not_ tried manual submissions for each destination on the pizza - are we trusting E2E coverage there or waiting until staging or what do you think?

editor.planx.uk/src/@planx/components/Send/Public.test.tsx
DafyddLlyr: Ah of course - very good catch - best to be as explicit and through as possible here.

Updated in [480d86b](https://github.com/theopensystemslab/planx-new/pull/3843/commits/480d86bc07da13387ab82dde0119f1982e390aba)

DafyddLlyr:
> Code looks good, but I admittedly have not tried manual submissions for each destination on the pizza - are we trusting E2E coverage there or waiting until staging or what do you think?

Likewise - I've not manually tested each submission here, just the event creation which this should only be touching.

I've got a good degree of confidence that current test coverage is good here 👍

PR #3846, opened 2024-10-24
docs: add note about envs to team-secret doc

jamdelion:
Another doc update based on more Jo-misunderstandings 😅


DafyddLlyr (approved):
Always good to see updated docs - thanks!

PR #3847, opened 2024-10-25
chore: Add `VERBOSITY=terse` flag to db sync script in CI

DafyddLlyr:
Please see https://opensystemslab.slack.com/archives/C01EVPXFU7K/p1729769117634399 for context (OSL Slack)

Using the psql flag `VERBOSITY=terse` will prevent error `DETAILS` being caught in CI logs, but would still appear when running locally ([docs](https://www.postgresql.org/docs/current/runtime-config-logging.html#:~:text=TERSE%20excludes%20the%20logging%20of,number%20that%20generated%20the%20error.)).

### DIFF
--- a/.github/workflows/sync-staging-db.yml
+++ b/.github/workflows/sync-staging-db.yml
@@ -36,7 +36,15 @@ jobs:
       - name: Run sync script
         run: | 
           {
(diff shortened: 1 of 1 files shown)


jessicamcinchak (approved):
Nice solution !

PR #3854, opened 2024-10-25
fix: Scroll position behaviour

DafyddLlyr:


editor.planx.uk/src/pages/FlowEditor/lib/useScrollControlsAndRememberPosition.ts
```diff
@@ -15,97 +16,96 @@ import { rootFlowPath } from "../../../routes/utils";
 const useScrollControlsAndRememberPosition = (
   scrollContainerRef: React.RefObject<HTMLDivElement>,
 ) => {
+  const currentPath = rootFlowPath(true);
+  const isLoading = useLoadingRoute();
```
DafyddLlyr, line 20: The fix here is to make these dependencies of the `useEffect()`.

- `currentPath` ensures that each portal and flow retains a scroll position in session storage
- `isLoading` fixes a timing issue to ensure that `scrollTo()` is called once the route has finished loading

editor.planx.uk/src/pages/FlowEditor/lib/useScrollControlsAndRememberPosition.ts
```diff
@@ -15,97 +16,96 @@ import { rootFlowPath } from "../../../routes/utils";
...
   useEffect(() => {
-    let scrollListener: Subscription;
-    let dragListener: Subscription;
+    if (isLoading) return;
 
-    const storageKey = ["scrollPos", rootFlowPath(true)].join(":");
+    const storageKey = ["scrollPos", currentPath].join(":");
 
     const container = scrollContainerRef.current;
+    if (!container) return;
```
DafyddLlyr, line 28: Added as a guard clause to drop a level of nesting below, the rest of the diff below is just - 
 - Indentation 
 - Removed `any`s, added types

DafyddLlyr:
[Flows should be centered within the window on initial load](https://trello.com/c/wIFWO8dr/3083-flows-should-be-centered-within-the-window-on-initial-load)

jessicamcinchak (approved):

PR #3864, opened 2024-10-28
feat: Use Hasura column presets to populate `creator_id`

RODO94:
This follows work on the demo workspace ticket: https://trello.com/c/EfaEQRzK/2441-create-a-demo-workspace

We found that the `creator_id` was not populating when we create a flow in the editor. To fix this, we decided to use Hasura Column Presets to reduce the number of places the change had to be made, and to make it more maintainable in the future.

We do sacrifice readability in the code base, but we found it was a better win for maintainability for a data field most people would expect to be populated by the user creating it. 

[Column Preset Docs](https://hasura.io/docs/2.0/auth/authorization/permissions/column-presets/)

### DIFF
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -157,7 +157,7 @@
       definition:
         enable_manual: false
         insert:
-          columns: '*'
+          columns: "*"
       retry_conf:
         interval_sec: 30
         num_retries: 1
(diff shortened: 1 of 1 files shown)


RODO94:

DafyddLlyr:
@RODO94 Could you please also run E2E regression tests against this branch just to check that this method holds up with the mocking we've done there?

RODO94:
@DafyddLlyr Regression tests have passed

DafyddLlyr (changes requested):
Bit of a nitpick sorry, but there's no linting set up for this file afaik - it looks like it's your local IDE config that's changes from `"` to `'` here. 

Could you please change this back so that the only diff you've made for the column presets shows here?

DafyddLlyr (approved):
Nice! Thanks

PR #3865, opened 2024-10-28
refactor: split out ui file

jamdelion:
Sorry for the long diff - most of these are file import changes.

**Summary of main changes:**

The `ui.tsx` file seemed very random - filled with shared components such as `MoreInformation`, miscellaneous shared types (e.g. `PublicProps`) and the long icons mapping constant.

- I have split this into `shared/icons`, `shared/types.tsx` and moved the components into their own files, which are under `/src/ui/editor`.
- Also I've split out `types.tsx` from `Checklist.tsx`


editor.planx.uk/src/@planx/components/InternalNotes.tsx
DafyddLlyr, line 1: I think this should move to `/src/ui/editor/` to be located alongside other Editor modal components.

editor.planx.uk/src/@planx/components/MoreInformation/MoreInformation.tsx
DafyddLlyr, line 1: As above, this should go in `/src/ui/editor`.

The intention here would be for `src/@planx/components` to be the _graph_ components. There's a tricky language thing here where these are the lego building blocks of the graph, but the word "component" in the lego block sense clashes with "component" in the web framework sense.

editor.planx.uk/src/@planx/components/sharedTypes.tsx
DafyddLlyr, line 1: Maybe `/shared/types.tsx` would be more appropriate here?

DafyddLlyr (changes requested):
Great little refactor, love to see it!

A few comments on co-locating the Editor modal components 👍

editor.planx.uk/src/@planx/components/InternalNotes.tsx
jamdelion: Thanks, that makes sense! I think I've got a few holes in my understanding of the context around these groupings so that's helpful to have that framing :)

editor.planx.uk/src/@planx/components/MoreInformation/MoreInformation.tsx
jamdelion: I wonder if it would make sense to rename one of the folders to `graphComponents` in that case? 🤔

DafyddLlyr: Just discussed in dev call - let's stick with `/components` for now 👍

DafyddLlyr (approved):

PR #3867, opened 2024-10-29
feat: Adding a new user in the demo team automatically assigns the `demoUser` role

DafyddLlyr:
## What does this PR do?
 - If a user is added to the demo team, assigns the `demoUser` role instead of the `teamEditor` role

## To test
 - Add a user to the "Demo" team
 - User appears in table with `demoUser` role applied
 - Hasura `team_members` table confirms this

### Next steps...
 - Enforce via login API that users with the `demoUser` role can only access staging
 - Update team members UI to me more role agnostic - some language like "Team Editors" is not quite right (`demoUser` and `teamViewer` users are displayed)
 - Hide certain UI elements from `demoUser` role


RODO94:

PR #3868, opened 2024-10-29
feat: add `application.type` to Metabase Analytics

RODO94:
add variables to allow list


api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
```diff
@@ -15,6 +15,9 @@ import type { Operation } from "../sanitiseApplicationData/types.js";
  *  so that Metabase picks up the new columns
  */
 const ALLOW_LIST = [
+  "pp.full.householder",
+  "pp.full.minor",
+  "pp.full.major",
```
jessicamcinchak, line 20: Realise this is still draft, but want to flag early that these are _not_ expected allow list variables - expect this to be `application.type` as the Trello ticket is written with _possible values_ `pp.full.householder`, etc 

Allow list cares about the passport key/property names, _not_ the possible values ! 

Does this make sense?

RODO94:
Closed due to messy PR, reopening

PR #3870, opened 2024-10-29
fix: Update publishing details on flow change

DafyddLlyr:
Please see https://opensystemslab.slack.com/archives/C01E3AC0C03/p1730197158271359

## What's the problem?
- The "last published" text does not update on flow change
- The diff count badge does not update on flow change

## What's the solution
- Add `flowId` to the dependency arrays of the functions responsible for updating these values
- This likely wasn't an issue in the past pre-route fixes and the component was always re-rendering - now that's not the case

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Publish/PublishFlowButton.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Publish/PublishFlowButton.tsx
@@ -96,14 +96,14 @@ export const PublishFlowButton: React.FC<{ previewURL: string }> = ({
     const user = await lastPublisher(flowId);
 
     setLastPublishedTitle(formatLastPublishMessage(date, user));
-  });
+  }, [flowId]);
 
   const _validateAndDiffRequest = useAsync(async () => {
     const newChanges = await validateAndDiffFlow(flowId);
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Publish/PublishFlowButton.tsx
```diff
@@ -96,14 +96,14 @@ export const PublishFlowButton: React.FC<{ previewURL: string }> = ({
...
-  });
+  }, [flowId]);
 
   const _validateAndDiffRequest = useAsync(async () => {
     const newChanges = await validateAndDiffFlow(flowId);
     setAlteredNodes(
       newChanges?.data.alteredNodes ? newChanges.data.alteredNodes : [],
     );
-  });
+  }, [flowId]);
```
DafyddLlyr, line 106: _Technically_ this should update each time there's an operation, but I think this would be calling this expensive API call a lot more than we need to.

Having it update once when navigating into a flow feels fine as a fix for now.

jamdelion: Neat :)

jamdelion (approved):

PR #3852, opened 2024-10-25
feat: add demoUser role to Hasura with permissions

RODO94:
## What does this PR do? 

For the Demo Workspace, a `demoUser` may only see flows created by them, or flows within the Open Systems Lab, Open Digital Planning, or Templates team. Also, they cannot see team settings or any other setting outwith flow settings for their own flows.

I have treated them as a `teamEditor` within their flows, and restricted data they can see outwith this. 

### Introspection testing

Introspection testing has been added to help communicate the relevant permissions changes and ensure future maintainability.

PR #3871, opened 2024-10-29
feat: add `application.type' to allow list for Metabase

RODO94:
### What has changed?

I have added `application.type` to the `ALLOW_LIST` in the relevant api and editor repos, and I have added a migration to add a `application_type` to the `analytics_summary` and the `submission_services_summary` 

Ran through Southwarks Apply for Planning Permission service and tested the output, and received `["pp.full.householder"]` in the `analytics_summary` and the same in the `submission_services_summary`

If this is all good, I'll work up a how-to doc for future ref

### DIFF
--- a/api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
+++ b/api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
@@ -15,6 +15,7 @@ import type { Operation } from "../sanitiseApplicationData/types.js";
  *  so that Metabase picks up the new columns
  */
 const ALLOW_LIST = [
+  "application.type",
   "application.declaration.connection",
   "application.information.harmful",
   "application.information.sensitive",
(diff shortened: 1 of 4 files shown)


hasura.planx.uk/metadata/tables.yaml
RODO94, line 1: In a previous PR: https://github.com/theopensystemslab/planx-new/pull/3852/files#diff-078286215f2fbfc71afde3de4775cf69ece92a85c1e0a2cf75c7bdaa29b2c010

I accidentally switched from `'` to `"` and this PR is correctly undoing that mistake when tables.yaml is recreated by Hasura

api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
```diff
@@ -15,6 +15,7 @@ import type { Operation } from "../sanitiseApplicationData/types.js";
  *  so that Metabase picks up the new columns
  */
 const ALLOW_LIST = [
+  "application.type",
```
jessicamcinchak, line 18: nit: both `ALLOW_LIST` lists are currently sorted alphabetically, it'd be nice to maintain that order please !

jessicamcinchak (approved):
Looks good & all working as expected, thanks for picking this one up!

PR #3874, opened 2024-10-29
fix: eslint errors from auto-answering changes

jessicamcinchak:
Follows on from #3764 

Running `pnpm check` in the editor subdirectory now returns 0 errors for me, sorry I missed these on initial merge !

### DIFF
--- a/editor.planx.uk/.eslintrc
+++ b/editor.planx.uk/.eslintrc
@@ -87,7 +87,15 @@
         }
       }
     ],
-    "no-nested-ternary": "error"
+    "no-nested-ternary": "error",
+    "@vitest/expect-expect": [
+      "error",
+      {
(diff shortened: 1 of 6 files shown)


editor.planx.uk/.eslintrc
```diff
@@ -87,7 +87,15 @@
...
-    "no-nested-ternary": "error"
+    "no-nested-ternary": "error",
+    "@vitest/expect-expect": [
+      "error",
+      {
+        "assertFunctionNames": [
+          "expect"
+        ]
+      }
+    ]
```
jessicamcinchak, line 98: This rule is consistent with our API's `eslintrc` too

jamdelion (approved):
Thanks for quick turnaround!

PR #3876, opened 2024-10-30
chore: Update copy on "Team Members" page to be less Editor specific

DafyddLlyr:
## What does this PR do?
- Update copy on "Team Members" page to be less Editor specific, for example changing the titles from "Team Editors" to "Team Members"
- Updates language used to code to also reflect this

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx
@@ -41,12 +41,12 @@ export const TeamMembers = () => {
 
   return (
     <Container maxWidth="contentWrap">
-      <SettingsSection data-testid="team-editors">
+      <SettingsSection data-testid="team-members">
         <Typography variant="h2" component="h3" gutterBottom>
-          Team editors
+          Team members
         </Typography>
         <Typography variant="body1">
-          Editors have access to edit your services.
+          Editors have access to edit your services, whilst viewers can only browse your services.
         </Typography>
         <MembersTable
           members={activeMembers}
--- a/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
@@ -102,7 +102,7 @@ export const MembersTable = ({
                       addUser();
                     }}
                   >
-                    Add a new editor
+                    Add a new member
                   </AddButton>
                 </TableCell>
               </TableRow>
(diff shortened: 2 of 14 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Team/TeamMembers.tsx
```diff
@@ -41,12 +41,12 @@ export const TeamMembers = () => {
...
     <Container maxWidth="contentWrap">
-      <SettingsSection data-testid="team-editors">
+      <SettingsSection data-testid="team-members">
         <Typography variant="h2" component="h3" gutterBottom>
-          Team editors
+          Team members
         </Typography>
         <Typography variant="body1">
-          Editors have access to edit your services.
+          Editors have access to edit your services, whilst viewers can only browse your services.
```
DafyddLlyr, line 49: We don't need to explain the `demoUser` role here as only `platformAdmins` will be able to interact with this.

editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
```diff
@@ -98,14 +101,14 @@ export const MembersTable = ({
...
-                  Add a new editor
+                  Add a new member
                 </AddButton>
               </TableCell>
             </TableRow>
           )}
         </Table>
         {showModal && (
-          <EditorUpsertModal
+          <UserUpsertModal
```
DafyddLlyr, line 111: There's still a distinction between a `user` and `team_members` that I think we want to maintain. This modal adds a `user` record to the db, and then assigns them a role.

editor.planx.uk/src/pages/FlowEditor/components/Team/components/UserUpsertModal.tsx
```diff
@@ -136,7 +136,7 @@ export const EditorUpsertModal = ({
         >
           <Box sx={{ mt: 1, mb: 4 }}>
             <Typography variant="h3" component="h2" id="dialog-heading">
-              Add a new editor
+              Add a new { isDemoTeam ? "demo user" : "editor" }
```
RODO94, line 139: Should `"editor"` here not be `"member"`?

jessicamcinchak (approved):
All clear, thanks !

editor.planx.uk/src/pages/FlowEditor/components/Team/components/UserUpsertModal.tsx
RODO94: Or is it now that we're adding a user, we can now specify the type of role they can receive?

DafyddLlyr: Yep exactly - this is specifically adding a user with the `teamEditor` role, so "editor" is correct.

I think there's certainly scope for a role picker within this modal, but I think this is a separate task so I'm keeping thing simple here!

RODO94: Only grievance I would have with this is that you click "Add a team member" but receive a modal saying "Add a team editor" - feels like there's room for confusion. Could it be worth keeping the modal heading the same but alter the info text of the modal to give guidance on what permissions the team member will have?

This is a small nitty comment, so happy for it to be parked for future

PR #3877, opened 2024-10-30
feat: add demo user to api hasura roles

RODO94:
Altered the JWT token builder in the api to set the correct role rather than falling back to only `teamEditor`.

Now it will correctly select `platformAdmin | teamEditor | teamViewer | demoUser`


DafyddLlyr (approved):

RODO94:

PR #3878, opened 2024-10-30
dp/demo user trigger update

DafyddLlyr:
- **feat: Update trigger**
- **test(e2e): Add test coverage for new trigger behaviour**


RODO94:

hasura.planx.uk/migrations/1730277860890_run_sql_migration/up.sql
```diff
@@ -0,0 +1,36 @@
...
+    
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;
+
+DROP TRIGGER IF EXISTS grant_new_user_template_team_access ON users;
+
+CREATE CONSTRAINT TRIGGER grant_new_user_template_team_access
+AFTER INSERT ON users 
+DEFERRABLE INITIALLY DEFERRED
```
DafyddLlyr, line 33: `DEFERRABLE INITIALLY DEFERRED` is new to me!

This ensures that the the trigger will run post _commit_, not post _insert_.

The insert of a new user and the associated `demoUser` role are grouped into a single transaction via [the GraphQL mutation here](https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/src/pages/FlowEditor/components/Team/queries/createAndAddUserToTeam.tsx#L31). 

Without `DEFERRABLE INITIALLY DEFERRED`, the trigger will run after the user has been inserted and then hit a race condition against insertion of the `team_member` record (it may check for the record before it's been inserted).

This will now wait until both the `users` record and `team_members` record has been inserted before running.

e2e/tests/api-driven/src/hasuraTriggers/helpers.ts
```diff
@@ -1,6 +1,48 @@
...
+    lastName: "Test",
+    email: TEST_EMAIL,
+    teamId: demoTeamId,
+    role: "demoUser",
+  };
+
+  const response = await safely(() =>
+    $admin.client.request<{ insertUsersOne: { id: number } }>(
+      gql`
+        mutation CreateAndAddUserToTeam(
```
DafyddLlyr, line 23: This matches the query on the frontend - might be a good reason to wrap up on `planx-core`?

RODO94 (approved):
Working as intended for me. The `DEFERRABLE INITIALLY DEFERRED` bit was interesting, nice new bit of SQL

PR #3880, opened 2024-10-30
fix: Grant additional permissions to demoUser role

DafyddLlyr:
A few issues when testing the `demoUser` role end-to-end - 
 - All users must be able to access their own record in the `team_members` table in order to log in
 - All users must have read access to the `team_settings` table


RODO94:

PR #3883, opened 2024-10-30
fix: Use integer for userId in JWT

DafyddLlyr:
Regression tests running here - https://github.com/theopensystemslab/planx-new/actions/runs/11600394961


DafyddLlyr:
This isn't the right way forward here, according to the docs - 

> Session variables are currently expected to only be strings and should be encoded based on Postgres's literals for the relevant type.

Source: https://hasura.io/docs/2.0/auth/authorization/roles-variables/

Hasura should be casting this correctly, the insert error I'm getting when trying to use the demoUser role must be coming from somewhere else. Looking into it atm!

PR #3884, opened 2024-10-31
fix: Remove manual population of creator ID

DafyddLlyr:
## What's the problem?
 - The `/flows/:flowId/copy` endpoint is broken - throwing a permission error due to a missing field `creator_id`
 - All roles which write to this table have write permissions for this column
 
 ## What's the cause?
When we added a column preset for `creator_id` (https://github.com/theopensystemslab/planx-new/pull/3864) it removed the columns from the insert schema for the flows table.
 
 > The respective fields will also be removed from the generated GraphQL schema for that role.

Source: https://hasura.io/docs/2.0/auth/authorization/permissions/column-presets/

image

 
## What's the solution?
- Let Hasura handle this and populate it from the JWT, remove our custom code for fetch the user ID from the JWT
- It's actually a pretty nice solution / pattern imho!
- I've checked for other references to `creator_id` and `creatorId` - this is the only one in this repo and `planx-core`

### DIFF
--- a/api.planx.uk/helpers.ts
+++ b/api.planx.uk/helpers.ts
@@ -4,6 +4,7 @@ import type { Flow, Node } from "./types.js";
 import type { FlowGraph } from "@opensystemslab/planx-core/types";
 import { ComponentType } from "@opensystemslab/planx-core/types";
 import { $public, getClient } from "./client/index.js";
+import { userContext } from "./modules/auth/middleware.js";
 
 export interface FlowData {
   slug: string;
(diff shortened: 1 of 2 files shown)


api.planx.uk/helpers.ts
RODO94, line 1: I don't want to overcomplicate this but, would this be a good time to switch this to the `CreateFlow` mutation from `planx-core` so it matches the editor? @DafyddLlyr

DafyddLlyr: Yep, makes sense to standardise this I think and it could have avoided this issue.

As this is currently blocking work, I'd vote for moving this forward just now and getting to prod and picking up this refactor as a follow up?

RODO94 (approved):
Fixes the bug for me, thanks for picking this up!

PR #3889, opened 2024-10-31
feat: demo workspace e2e

RODO94:
## What does this PR do?

This PR follows on for demo workspace work here: https://trello.com/c/EfaEQRzK/2441-create-a-demo-workspace

I have written a suite of api-driven e2e tests to ensure `demoUser` restrictions and capabilities are maintained during future code changes.

### Strategy

To ensure a future dev could understand the tests and the nuances around database permissions, I have explicitly written out what data should be in the database during the tests in the `Given` statements within the `Background` of the `.feature` file. With this, I am able to populate the db directly with the `datatable` information.

To support the range of scenarios and different situations, I have split out the `.steps` file out into `background | navigation | verification` with each file handling either the `Given` steps, the `When` steps, or the `Then | But` steps.

This felt better from a readability point of view as putting them all in the same file became unwieldy and collating `When` or `Then` conditions seem to abstract the point of some of the tests. 

The `helper.ts` file contains a lot of GraphQL requests because I had to return specific data or send specific data and found the shared requests didn't give me enough control on things like `id & creator_id`. This is similar for the types like `Flow` etc.

### Struggles

I did struggle a bit with the right design patterns here and thinking about what code should sit within which kind of step, what goes in a `When` compared with a `Then` but I hope my use of state helps this. 

I seem to hit a randomly occurring error as well around the creation of flows, I couldn't pinpoint how it is happening or what is causing it. The addition or removal of a `console.error` seems to trigger the error sometimes, so I thought it may be a race condition around a `Promise.all()` but I couldn't root it out... if anyone finds something, let me know, for now it is all passing.


e2e/tests/api-driven/src/demo-workspace/demoUser.feature
```diff
@@ -0,0 +1,94 @@
...
+        Examples:
+            | TEAM  |
+            | demo  |
+
+
+    @demo-user-permissions
+    Scenario Outline: I can only view specific teams
+        When I query the teams table
+        Then I can access the teams with slug: "<SLUG>"
+        But I should not access the Other Team
```
jamdelion, line 42: nit: could this be reworded to `I should not be able to access the Other Team`?

e2e/tests/api-driven/src/demo-workspace/demoUser.feature
```diff
@@ -0,0 +1,94 @@
...
+            | SLUG                      |
+            | open-systems-lab          |
+            | templates                 |
+            | open-digital-planning     |
+            | demo                      |
+
+    @demo-user-permissions 
+    Scenario Outline: Creating a new flow
+        When I insert a flow into the team: "<TEAM>"
+        Then I should not succeed
```
jamdelion, line 54: Could this line be written a bit more explictly, so that it's more obvious what is not succeeding when you come across it in the steps file?

e.g. `I should not be able to create a flow`

jamdelion: Same for the line below actually, maybe `I should be able to create a flow in the Demo team`

e2e/tests/api-driven/src/demo-workspace/steps/background_steps.ts
```diff
@@ -0,0 +1,134 @@
...
+  otherTeam!: DataTableRecord;
+  insertFlowTeamId!: number;
+  demoFlowSlug!: string;
+  adminFlowSlug!: string;
+  demoTeamsArray!: Team[];
+  currentTeamId!: number;
+  teamFlows!: Flow[];
+}
+
+Before("@demo-user-permissions", async function () {
```
jamdelion, line 31: Ah actually I see you using the tag for cleanup, so my suggestion of removing the duplicates would interfere with this! Feel free to ignore

e2e/tests/api-driven/src/demo-workspace/steps/navigation_steps.ts
```diff
@@ -0,0 +1,48 @@
+import { When } from "@cucumber/cucumber";
+import { $admin } from "../../client";
+import { strict as assert } from "node:assert";
+import { getFlowBySlug, getTeamAndFlowsBySlug, getTeams } from "../helper";
+import { CustomWorld } from "./background_steps";
+
+When<CustomWorld>("I am in the {string} team", async function (this, string) {
```
jamdelion, line 7: naming: Wondering if there is a better variable name other than `string` as this could look a little confusing as a name in a typescript repo... maybe just `teamName`? Or `teamSlug`

e2e/tests/api-driven/src/demo-workspace/steps/navigation_steps.ts
```diff
@@ -0,0 +1,48 @@
...
+});
+
+When<CustomWorld>("I query the teams table", async function (this) {
+  const teamsArray = await getTeams(this.demoClient);
+  this.demoTeamsArray = teamsArray;
+  assert.ok(teamsArray, "Teams not fetched correctly");
+});
+
+When<CustomWorld>(
+  "I insert a flow into the team: {string}",
```
jamdelion, line 26: Same comment as above re `string`

e2e/tests/api-driven/src/demo-workspace/steps/verification_steps.ts
```diff
@@ -0,0 +1,114 @@
...
+  async function (this) {
+    const doesContainOtherFlows = this.teamFlows.find(
+      (flow) => flow.creator_id !== this.demoUser.id,
+    );
+    assert.ok(!doesContainOtherFlows, "Other flows are in the array");
+  },
+);
+
+Then<CustomWorld>(
+  "I can access the teams with slug: {string}",
```
jamdelion, line 32: naming: `teamSlug` again?

e2e/tests/api-driven/src/demo-workspace/steps/verification_steps.ts
```diff
@@ -0,0 +1,114 @@
...
+  const flow = await getFlowBySlug(this.demoClient, this.adminFlowSlug);
+  assert.equal(
+    flow.slug,
+    this.adminFlowSlug,
+    "Incorrect flow has been fetched",
+  );
+});
+
+Then<CustomWorld>(
+  "I should be able to {string} the flow",
```
jamdelion, line 81: naming: maybe `action` instead of `string`?

e2e/tests/api-driven/src/jwt.ts
```diff
@@ -31,8 +31,7 @@ const getAllowedRolesForUser = (user: User): Role[] => {
   const teamRoles = user.teams.map((teamRole) => teamRole.role);
   const allowedRoles: Role[] = [
     "public", // Allow public access
-    "teamEditor", // Least privileged role for authenticated users - required for Editor access
-    ...teamRoles, // User specific roles
+    ...teamRoles, // User specif ic roles
```
jamdelion, line 34:
```suggestion
    ...teamRoles, // User specific roles
```

jamdelion (commented):
Loved how you split the steps up into the different types - really clear to read through. 

I've mainly added some wording/variable name change suggestions to make it even clearer!

e2e/tests/api-driven/src/demo-workspace/steps/verification_steps.ts
RODO94: Did not realise you could change the "string" name... this makes way more sense

jamdelion (approved):
Great! ✨

PR #3890, opened 2024-10-31
route Microsoft auth requests from pizzas to staging API

freemvmt:
Ticket: https://trello.com/c/WdZicC3b

Resolves the problem in exactly the same way as we do for Google - we do the auth whole flow via the staging API (`api.editor.planx.dev`), and are returned to the pizza afterwards (rather than the staging client).


DafyddLlyr (approved):
Perfect!

PR #3899, opened 2024-11-01
fix: ensure "no flag" band always displays if no flags are present

jessicamcinchak:
Flagged here: https://opensystemslab.slack.com/archives/C01E3AC0C03/p1730462810086459

To recreate bug on staging/prod: 
- Add a question and assign flags to an option and another option without flags -> "create" -> color bands display as expected
- Update that node and remove the flags from the option -> "update" -> no color band displays on updated option
  - Fix is to explicitly check for `flags.length > 0` so that `flags = []` doesn't evaluate to `FlagBand` option

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Option.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Option.tsx
@@ -5,10 +5,10 @@ import { Link } from "react-navi";
 
 import { useStore } from "../../../lib/store";
 import { DataField } from "./DataField";
+import { FlagBand, NoFlagBand } from "./FlagBand";
 import Hanger from "./Hanger";
 import Node from "./Node";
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Option.tsx
```diff
@@ -46,9 +48,17 @@ const Option: React.FC<any> = (props) => {
             imageAltText={props.data.text}
           />
         )}
-        {flags ? flags.map((flag) => <FlagBand key={`${props.id}-${flag.value}`} flag={flag} />) : <NoFlagBand />}
+        {flags && flags.length > 0 ? (
```
jessicamcinchak, line 51: This line is the only logical change, otherwise linting !

DafyddLlyr (approved):

PR #3903, opened 2024-11-04
fix: ensure questions that set a data field with no option data values are not automated

jessicamcinchak:
Fixes edge case flagged here: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1730455691984869

**Two changes:**
- `autoAnswerableOptions` condition is corrected so that _existing_ nodes will behave as expected (put to user)
- Formik validation added to editor modal for Questions & Checklists so that future nodes aren't allowed to do this at all 

Screenshot from 2024-11-04 10-15-31

**Testing:**
- Production https://editor.planx.uk/testing/self-automation
- Pizza https://3903.planx.pizza/testing/self-automation

### DIFF
--- a/editor.planx.uk/src/@planx/components/Checklist/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/Checklist/Editor.tsx
@@ -5,7 +5,7 @@ import FormControlLabel from "@mui/material/FormControlLabel";
 import IconButton from "@mui/material/IconButton";
 import Switch from "@mui/material/Switch";
 import { ComponentType as TYPES } from "@opensystemslab/planx-core/types";
-import { useFormik } from "formik";
+import { FormikErrors, FormikValues, useFormik } from "formik";
 import adjust from "ramda/src/adjust";
 import compose from "ramda/src/compose";
 import remove from "ramda/src/remove";
@@ -313,7 +313,7 @@ export const ChecklistComponent: React.FC<ChecklistProps> = (props) => {
       }
     },
     validate: ({ options, ...values }) => {
-      const errors: Record<string, string> = {};
+      const errors: FormikErrors<FormikValues> = {};
       if (values.fn && !options?.some((option) => option.data.val)) {
         errors.fn =
           "At least one option must set a data value when the checklist has a data field";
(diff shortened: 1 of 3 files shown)


jessicamcinchak:
[Automation & filtering logic followups](https://trello.com/c/dboRkua3/3112-automation-filtering-logic-followups)

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -514,11 +514,12 @@ export const previewStore: StateCreator<
...
 
-    // Only proceed if the user has seen at least one node with this fn before
+    // Only proceed if the user has seen at least one node with this fn (or `output` in case of Calculate nodes) before
     const visitedFns = Object.entries(breadcrumbs).filter(
-      ([nodeId, _breadcrumb]) => flow[nodeId].data?.fn === data.fn,
+      ([nodeId, _breadcrumb]) =>
+        [flow[nodeId].data?.fn, flow[nodeId].data?.output].includes(data.fn),
     );
-    if (!visitedFns) return;
+    if (!visitedFns.length) return;
```
jessicamcinchak, line 522: A `filter` is always going to return an `[]` so the previous condition incorrectly always returned true!

DafyddLlyr: Great catch 👍

editor.planx.uk/src/@planx/components/Checklist/Editor.tsx
```diff
@@ -312,7 +312,14 @@ export const ChecklistComponent: React.FC<ChecklistProps> = (props) => {
...
     },
-    validate: () => {},
+    validate: ({ options, ...values }) => {
+      const errors: FormikErrors<FormikValues> = {};
+      if (values.fn && !options?.some((option) => option.data.val)) {
+        errors.fn =
+          "At least one option must set a data value when the checklist has a data field";
+      }
+      return errors;
+    },
```
DafyddLlyr, line 322: nit: It would be nice to make these validation schemas for ease of expansion in future, but this could well be a a YAGNI thing!

DafyddLlyr (approved):
Nice simple solution 👏

PR #3905, opened 2024-11-04
fix: bump timeout on `/diff` endpoint

jessicamcinchak:
Large flows are timing out on "Check for changes to publish" as reported here: https://opensystemslab.slack.com/archives/C01E3AC0C03/p1730715895034839

### DIFF
--- a/api.planx.uk/modules/flows/validate/service/index.ts
+++ b/api.planx.uk/modules/flows/validate/service/index.ts
@@ -6,10 +6,10 @@ import type {
 import * as jsondiffpatch from "jsondiffpatch";
 
 import { dataMerged, getMostRecentPublishedFlow } from "../../../../helpers.js";
-// import { validateFileTypes } from "./fileTypes.js";
-// import { validateInviteToPay } from "./inviteToPay.js";
-// import { validateSections } from "./sections.js";
-// import { validateProjectTypes } from "./projectTypes.js";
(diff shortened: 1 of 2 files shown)


DafyddLlyr (approved):

api.planx.uk/modules/flows/validate/service/index.ts
```diff
@@ -49,29 +49,30 @@ const validateAndDiffFlow = async (
...
+  //   .concat(warningChecks)
+  //   .concat(passingChecks)
+  //   .concat(notApplicableChecks);
 
   return {
     alteredNodes,
     message: "Changes queued to publish",
-    validationChecks: sortedValidationChecks,
+    // validationChecks: sortedValidationChecks,
+    validationChecks: undefined,
```
DafyddLlyr, line 75: I think this needs to `[]` for the following line to not throw on error (`Cannot read properties of undefined (reading 'filter')`) - 

https://github.com/theopensystemslab/planx-new/blob/21755b044fa6d94c68933dc790ef3044b9ebe4f4/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Publish/PublishFlowButton.tsx#L180

DafyddLlyr (changes requested):

DafyddLlyr (approved):

PR #3908, opened 2024-11-04
feat: ensure `application_type` is added to views as a text field

RODO94:
I have added a new migration to add `application_type` to the `submission_servies_summary` and `analytics_summary` views as a text field so Metabase can filter based on their values rather than their existence (currently you can only filter on whether the data exists or not, instead of what the data is)

Main change is this line:

```sql
   (al.allow_list_answers -> 'application.type' -> 0)::text AS application_type
```


hasura.planx.uk/migrations/1730730955039_application_type_as_text_in_view/up.sql
```diff
@@ -0,0 +1,162 @@
...
+    sa.s3_applications,
+    (ls.allow_list_answers -> 'usedFOIYNPP'::text) AS used_foiynpp,
+    (ls.allow_list_answers -> 'propertyInformation.action'::text) AS property_information_action,
+    (ls.allow_list_answers -> 'planningConstraints.action'::text) AS planning_constraints_action,
+    (ls.allow_list_answers -> '_overrides'::text) AS overrides,
+    (ls.allow_list_answers -> 'rab.exitReason'::text) AS rab_exit_reason,
+    (ls.allow_list_answers -> 'service.type'::text) AS pre_app_service_type,
+    (ls.allow_list_answers -> 'application.information.harmful'::text) AS pre_app_harmful_info,
+    (ls.allow_list_answers -> 'application.information.sensitive'::text) AS pre_app_sensitive_info,
+    (ls.allow_list_answers -> 'application.type' -> 0)::text AS application_type
```
DafyddLlyr, line 90: How come this one is different to all others? 

Shouldn't this fix apply to all columns, or is the data captured differently here?

RODO94: I wasn't sure on the data in the other columns and the request was only for `application.type` which I added the other week. My fear was thread pulling... 

Given it was highlighted as a priority, thought it best to keep this PR small and load up another in future for the other fixes? @DafyddLlyr

RODO94: Put the change up now, on second thought, seemed worth it to change and see what happens

zz-hh-aa:
Responding to @DafyddLlyr's point: the values display fine, it's just that Metabase won't let you _filter_ on a JSONB value and we want councils to be able to filter on this one for the HHPP dashboard. 

With other JSONB columns, my other workaround has been creating a SQL query that does the reformatting, but we haven't needed to do it so often. The more graphs and councils we have and the more instances in which that fix needs to be made, the less sustainable it feels...

I asked Mike ages ago about getting the other JSONB values recast as text to make them easier to handle in Metabase, and he explained why it was better to leave them as is--but I have to say I don't  remember the rationale!

RODO94:
I think the jsonb storage helps for data compression and size compared with text, maybe this was the reason?

PR #3910, opened 2024-11-04
fix: add missing postgres indexes 

jessicamcinchak:
Follows on from publishing debugging today: https://opensystemslab.slack.com/archives/C01E3AC0C03/p1730715895034839

**Approached this from two directions:**
1. Which indexes are we missing? Where is postgres is doing more sequence scans than index scans?
  - This blog has a useful sample query: https://erikrw.hashnode.dev/how-to-identify-missing-indexes-in-postgresql
  - From these results, I then cross-checked with how we commonly query those tables (eg `where` params)
  - Screenshot from 2024-11-04 19-05-38
2. Which existing indexes are unused? Can any be removed?
  - Ran `select * from pg_stat_user_indexes where schemaname = 'public' order by idx_scan asc;` for this one
  - We do have multiple cases of existing indexes that have never been hit, but they're all direct remnants of "Primary keys" or "Unique keys" created via Hasura and therefore don't make sense to drop
  - Screenshot from 2024-11-04 18-45-36

Wish that `published_flows` & `flows` would have come up higher on the "need to improve" checks, but the truth is they didn't !

### DIFF
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -157,7 +157,7 @@
       definition:
         enable_manual: false
         insert:
-          columns: "*"
+          columns: '*'
       retry_conf:
         interval_sec: 30
         num_retries: 1
@@ -171,7 +171,7 @@
         query_params:
           type: bops-submission
         template_engine: Kriti
-        url: "{{$base_url}}/webhooks/hasura/send-slack-notification"
+        url: '{{$base_url}}/webhooks/hasura/send-slack-notification'
         version: 2
 - table:
     name: document_template
@@ -225,7 +225,7 @@
       definition:
         enable_manual: false
         insert:
-          columns: "*"
+          columns: '*'
       retry_conf:
         interval_sec: 30
         num_retries: 1
@@ -239,7 +239,7 @@
         query_params:
           type: email-submission
         template_engine: Kriti
-        url: "{{$base_url}}/webhooks/hasura/send-slack-notification"
+        url: '{{$base_url}}/webhooks/hasura/send-slack-notification'
         version: 2
 - table:
     name: feedback
@@ -483,9 +483,9 @@
             forward_client_headers: false
             headers:
               - name: authorization
-                value: "{{HASURA_PLANX_API_KEY}}"
+                value: '{{HASURA_PLANX_API_KEY}}'
             timeout: 10
-            url: "{{HASURA_PLANX_API_URL}}/webhooks/hasura/validate-input/jsonb/clean-html"
+            url: '{{HASURA_PLANX_API_URL}}/webhooks/hasura/validate-input/jsonb/clean-html'
           type: http
     - role: demoUser
       permission:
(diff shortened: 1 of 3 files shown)


hasura.planx.uk/migrations/1730742914548_create_index_bops_applications_session_id/up.sql
```diff
@@ -0,0 +1,14 @@
...
+CREATE  INDEX "email_applications_session_id_idx" on
+  "public"."email_applications" using hash ("session_id");
+CREATE  INDEX "s3_applications_session_id_idx" on
+  "public"."s3_applications" using hash ("session_id");
+CREATE  INDEX "feedback_flow_id_team_id_idx" on
+  "public"."feedback" using btree ("team_id", "flow_id");
+CREATE  INDEX "reconciliation_requests_session_id_idx" on
+  "public"."reconciliation_requests" using hash ("session_id");
+CREATE  INDEX "published_flows_created_at_idx" on
+  "public"."published_flows" using btree ("created_at");
```
jessicamcinchak, line 14: A couple notes about this migration:
- Why a mix of `hash` versus `btree` index types? 
  - I'm using `hash` indices when it's a column that is exclusively queried via an equality check and `btree` indices when a) it's a column that is queried via equality _or range_ checks or b) multiple columns are combined into the same index
  - Postgres docs on this are helpful https://www.postgresql.org/docs/current/indexes-types.html
- I'm adding an index to `published_flows.created_at` because that is how all of our current queries are ordering, but I have _not_ actually checked/compared yet if there's a difference to sort by published flow id (or jogged my memory about why we don't do this already)

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -157,7 +157,7 @@
       definition:
         enable_manual: false
         insert:
-          columns: "*"
+          columns: '*'
```
jessicamcinchak, line 160: This file is only formatting changes :cyclone:

jessicamcinchak:
[Publishing timeout errors](https://trello.com/c/ebykktwn/3117-publishing-timeout-errors)

hasura.planx.uk/migrations/1730742914548_create_index_bops_applications_session_id/up.sql
DafyddLlyr: It may also be worth adding a `DESC` in here as we're always searching in this order ([docs](https://www.postgresql.org/docs/current/indexes-ordering.html)).

DafyddLlyr (approved):

hasura.planx.uk/migrations/1730742914548_create_index_bops_applications_session_id/up.sql
jessicamcinchak: Cool good shout - that's now updated :+1: Worth noting order isn't possible to specify in Hasura GUI for adding indexes, but worked no problem when added via "SQL" panel directly
Screenshot from 2024-11-05 09-13-40

PR #3911, opened 2024-11-05
fix: Hide `/:teams` `EditorNavMenu` from `demoUser` role

DafyddLlyr:
## What?
Ensure `demoUser` role cannot view Team settings, team members, or team design pages.

## Why?
Feedback from Show & Tell - users cannot use pages anyway, no benefit to showing them.

### DIFF
--- a/editor.planx.uk/src/components/EditorNavMenu.tsx
+++ b/editor.planx.uk/src/components/EditorNavMenu.tsx
@@ -9,7 +9,7 @@ import TuneIcon from "@mui/icons-material/Tune";
 import Box from "@mui/material/Box";
 import IconButton from "@mui/material/IconButton";
 import { styled } from "@mui/material/styles";
-import Tooltip, { tooltipClasses, TooltipProps } from "@mui/material/Tooltip";
+import Tooltip from "@mui/material/Tooltip";
 import Typography from "@mui/material/Typography";
 import { Role } from "@opensystemslab/planx-core/types";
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/components/EditorNavMenu.tsx
```diff
@@ -243,11 +242,7 @@ function EditorNavMenu() {
 
   const { routes, compact } = getRoutesForUrl(url.href);
 
-  const visibleRoutes = routes.filter(({ accessibleBy }) => {
-    if (user?.isPlatformAdmin) return accessibleBy.includes("platformAdmin");
-    if (canUserEditTeam(teamSlug)) return accessibleBy.includes("teamEditor");
-    return accessibleBy.includes("teamViewer");
-  });
+  const visibleRoutes = routes.filter(({ accessibleBy }) => role && accessibleBy.includes(role));
```
RODO94, line 245: This is checking they can only see routes with "demoUser" in the `accessibleBy` rather than restricting the paths they shouldn't see?

RODO94 (approved):
Works for me, I can't see any settings until I'm in a flow that I have created on the Demo team

editor.planx.uk/src/components/EditorNavMenu.tsx
DafyddLlyr: Yes that's right - this is just a visual control for the menu 👍

PR #3925, opened 2024-11-07
style: Feedback component style refinement 

ianjon3s:
Small styling refinements to @jamdelion's work on the feedback component:

- Updates rating buttons to be responsive to width
- Updates selected rating button to be consistent with styling of selected radio grid (blue border)
- Small type and spacing refinements

**Before:**
image

**After:**
image

### DIFF
--- a/editor.planx.uk/src/@planx/components/Feedback/Public/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/Feedback/Public/Public.tsx
@@ -1,5 +1,6 @@
 import Box from "@mui/material/Box";
 import Grid from "@mui/material/Grid";
+import Typography from "@mui/material/Typography";
 import { Disclaimer } from "@planx/components/shared/Disclaimer";
 import Card from "@planx/components/shared/Preview/Card";
 import { CardHeader } from "@planx/components/shared/Preview/CardHeader/CardHeader";
@@ -83,15 +84,17 @@ const FeedbackComponent = (props: PublicProps<Feedback>): FCReturn => {
         id={"DESCRIPTION_TEXT"}
         openLinksOnNewTab
       />
-      <Box pt={2} mb={3}>
+      <Box my={4}>
         {props.ratingQuestion && (
           <InputLabel
             label={
-              <ReactMarkdownOrHtml
-                source={props.ratingQuestion}
-                id={"RATING_QUESTION"}
-                openLinksOnNewTab
-              />
+              <Typography component="span" fontWeight="700">
+                <ReactMarkdownOrHtml
+                  source={props.ratingQuestion}
+                  id={"RATING_QUESTION"}
+                  openLinksOnNewTab
+                />
+              </Typography>
             }
           />
         )}
@@ -102,7 +105,12 @@ const FeedbackComponent = (props: PublicProps<Feedback>): FCReturn => {
           onChange={handleFeedbackChange}
           aria-label="feedback score"
         >
-          <Grid container columnSpacing={15} component="fieldset">
+          <Grid 
+            container 
+            columnSpacing={2} 
+            component="fieldset" 
+            direction={{ xs: "column", formWrap: "row" }}
+          >
             <FaceBox
               value="1"
               testId="feedback-button-terrible"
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/@planx/components/Feedback/Public/Public.tsx
```diff
@@ -83,15 +84,17 @@ const FeedbackComponent = (props: PublicProps<Feedback>): FCReturn => {
...
-                id={"RATING_QUESTION"}
-                openLinksOnNewTab
-              />
+              <Typography component="span" fontWeight="700">
+                <ReactMarkdownOrHtml
+                  source={props.ratingQuestion}
+                  id={"RATING_QUESTION"}
+                  openLinksOnNewTab
+                />
+              </Typography>
```
DafyddLlyr, line 97: I think this will produce invalid HTML.

`ReactMarkdownOrHTML` serves content from a rich text input and could contain a number of elements such as a `h1` or `h2`, leading to a structure like this - 

```html
<span>
   <div>
     <h1></h1>
  </div>
</span>
```

Another example - 

image

A few options in terms of resolving this - we could change the `component` prop here to `div` (if allowed by `Typography` ?), or we could change the EditorModal to actually make this just a text field and not a rich text field.

I'd lean towards making this just a text input, but it's worth checking with @jamdelion and the content team if there's a valid reason to make this rich text 👍

editor.planx.uk/src/@planx/components/Feedback/Public/Public.tsx
```diff
@@ -139,18 +147,19 @@ const FeedbackComponent = (props: PublicProps<Feedback>): FCReturn => {
         {props.freeformQuestion && (
           <InputLabel
             label={
-              <ReactMarkdownOrHtml
-                source={props.freeformQuestion}
-                id={"RATING_QUESTION"}
-                openLinksOnNewTab
-              />
+              <Typography component="span" fontWeight="700">
```
DafyddLlyr, line 150: Same as above 👍

editor.planx.uk/src/@planx/components/Feedback/components/FaceBox.tsx
```diff
@@ -20,36 +20,37 @@ export const FaceBox = ({
...
-    <Grid item xs={2} key={label}>
+    <Grid item xs={2.4} key={label}>
       <ToggleButton
         value={value}
         data-testid={testId}
         sx={{
           px: 0,
+          width: "100%",
+          textTransform: "none",
+          "&[aria-pressed='true'] > div": {
```
DafyddLlyr, line 31:
```suggestion
          [`&.${toggleButtonClasses.selected}`]: {
            div: {
```

It's unlikely that MUI will change how they use `aria-pressed` but we're a bit safer relying on their classes to monitor state within a component in instances like this. 

Docs: https://mui.com/material-ui/api/toggle-button/#classes

editor.planx.uk/src/@planx/components/shared/Disclaimer.tsx
```diff
@@ -7,7 +7,7 @@ import ReactMarkdownOrHtml from "ui/shared/ReactMarkdownOrHtml/ReactMarkdownOrHt
 export const Disclaimer = ({ text }: { text: string }) => (
   <WarningContainer>
     <ErrorOutline />
-    <Typography variant="body1" component="div" ml={2} mb={1}>
+    <Typography variant="body2" component="div" ml={2} sx={{ "& p:first-of-type": { marginTop: 0, } }}>
```
DafyddLlyr, line 10: Nice!

DafyddLlyr (changes requested):
Very nice refining here, great to see 💅 

A few small comments on code below!

editor.planx.uk/src/@planx/components/Feedback/Public/Public.tsx
ianjon3s: I've removed the `<Typography>` wrapper and instead wrapped the `defaultContent` for these labels with `<strong>` tags so that if we keep the rich text editor there will be a styled default.

Agree on the preference for text input for input labels 👍

editor.planx.uk/src/@planx/components/Feedback/components/FaceBox.tsx
ianjon3s: Thanks for this, I've updated

DafyddLlyr (approved):
Perfect!

jamdelion:
Thanks Ian and Daf, merging this now so I can work with these changes in my PR 😄

PR #3929, opened 2024-11-08
feat: Use shared `Switch` component instead of MUISwitch directly

DafyddLlyr:
## What does this PR do?
 - Makes a single `Switch` component, with label, to be used across PlanX
 - This stops repetition, and should help with drift in other components in future
 - Ensures just switch and label is clickable, not entire row
 - Adds `.eslint` rule to enforce that the shared component should be imported

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings/FlowStatus/index.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings/FlowStatus/index.tsx
@@ -1,9 +1,6 @@
 import Box from "@mui/material/Box";
 import Button from "@mui/material/Button";
-import FormControlLabel, {
-  formControlLabelClasses,
-} from "@mui/material/FormControlLabel";
-import Switch from "@mui/material/Switch";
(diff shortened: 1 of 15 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings/FlowStatus/index.tsx
```diff
@@ -104,28 +102,25 @@ const FlowStatus = () => {
...
+        <Switch
+          label={statusForm.values.status as Capitalize<string>}
+          formControlLabelProps={{
+            sx: {
+              marginBottom: 0.5,
+              [`& .${formControlLabelClasses.label}`]: {
+                fontWeight: FONT_WEIGHT_BOLD,
+                textTransform: "capitalize",
+                fontSize: 19,
+              },
```
DafyddLlyr, line 114: Switches in these forms are visually different - the label is also really the title of the card.

We could make this a variant, or could attempt to standardise here, but this seemed the best way to handle this initially.

jessicamcinchak (approved):

PR #3930, opened 2024-11-11
fix (take 2): ensure questions that set a data field with no option data values are not automated

jessicamcinchak:
Follows on from #3903 (which was reverted via theopensystemslab/planx-new#3915)

See thread here https://opensystemslab.slack.com/archives/C01E3AC0C03/p1730884655361309

Original solution was failing to automate `property.localAuthorityDistrict` & `property.region` questions because these data fields are not stored under `fn` in FindProperty.


DafyddLlyr (approved):
Nice and simple!

PR #3932, opened 2024-11-11
fix: safer destructuring in `autoAnswerableOptions` to prevent error on initial editor load

jessicamcinchak:
It appears that when we're loading flows, we're sometimes attempting to run `autoAnswerableOptions` before the whole flow has loaded (eg on `_root` node rather than true current card??) leading to errors like the one screenshot below & reported by Ian Grace in #help-issues here: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1731336388546549

This tries destructuring the single `node` first before immediately assuming it has `{ type, data, edges }` too, similar with passport to account for cases where it's empty (eg no `data` yet)
Screenshot from 2024-11-11 16-33-56


jessicamcinchak:
[Automation & filtering logic followups](https://trello.com/c/dboRkua3/3112-automation-filtering-logic-followups)

DafyddLlyr (approved):

PR #3933, opened 2024-11-11
feat: Add multiline to text inputs

RODO94:
I have added the `multiline` prop to certain text inputs to allow multiline text in things like Options in Checkboxes.

Adjustment had to me made with the `textarea` component as it used a `height: auto` rather than a `height: 100%` the change hasn't broken any styling I could find, so hope its okay!

Tested by August here: https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1730291082584539


jessicamcinchak (approved):
Works for me, thanks!

PR #3938, opened 2024-11-12
feat: demo team integrations to fail gracefully

RODO94:
Adding some failure screens with Send Components and Planning Constraints to gracefully catch errors and inform the `demoUsers` of certain restrictions in the Public interface, rather than the Editor one. Found this better as it would still show our range of components (things like Planning Constraints) while removing a potentially confusing experience for users.

For demoUsers, they may still see these components normally when in other teams, like ODP, but they just can't see them when they are within the Demo team.

This is because for the Public interface of the components, we don't have access to a User object which would help us manage it specifically for a `demoUser` role which sits within the `teams` section of a `user` object, as it is applied to the `team_members` table.

Feedback on this component has been given here: https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1731576342896299

### DIFF
--- a/editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
@@ -75,11 +75,6 @@ function Component(props: Props) {
 
   const metadata = [...(props.govPayMetadata || []), ...defaultMetadata];
 
-  const errorMessage =
-    teamSlug !== "demo"
-      ? "GOV.UK Pay is not enabled for this local authority"
-      : "GOV.UK Pay is not enabled demo users";
-
   // Handles UI states
   const reducer = (_state: ComponentState, action: Action): ComponentState => {
     switch (action) {
@@ -302,7 +297,10 @@ function Component(props: Props) {
               : "Retry payment"
           }
           error={
-            (state.status === "unsupported_team" && errorMessage) ||
+            (teamSlug === "demo" &&
+              "GOV.UK Pay is not enabled for demo users") ||
+            (state.status === "unsupported_team" &&
+              "GOV.UK Pay is not enabled for this local authority") ||
             (state.status === "undefined_fee" &&
               "We are unable to calculate your fee right now") ||
             undefined
--- a/editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
@@ -78,7 +78,7 @@ function Component(props: Props) {
   const errorMessage =
     teamSlug !== "demo"
       ? "GOV.UK Pay is not enabled for this local authority"
-      : "GOV.UK Pay is not enabled for the Demo team";
+      : "GOV.UK Pay is not enabled for services created in the Demo team";
 
   // Handles UI states
   const reducer = (_state: ComponentState, action: Action): ComponentState => {
--- a/editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
@@ -78,7 +78,7 @@ function Component(props: Props) {
   const errorMessage =
     teamSlug !== "demo"
       ? "GOV.UK Pay is not enabled for this local authority"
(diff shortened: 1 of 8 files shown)


RODO94:

editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
```diff
@@ -76,6 +75,11 @@ function Component(props: Props) {
 
   const metadata = [...(props.govPayMetadata || []), ...defaultMetadata];
 
+  const errorMessage =
+    teamSlug !== "demo"
+      ? "GOV.UK Pay is not enabled for this local authority"
+      : "GOV.UK Pay is not enabled for the Demo team";
```
jessicamcinchak, line 81: nit: does "Demo team" have meaning to outsiders? Would something like "GOV.UK Pay is not enabled for demonstrations" or "GOV.UK Pay is not enabled in this trial workspace" be more generic and less internal-jargony?

editor.planx.uk/src/@planx/components/Send/Public.tsx
```diff
@@ -52,6 +55,20 @@ const SkipSendWarning: React.FC<Props> = (props) => (
   </Card>
 );
 
+const DemoTeamWarning: React.FC<Props> = (props) => (
+  <Card handleSubmit={props.handleSubmit}>
+    <ErrorSummaryContainer role="status">
+      <Typography variant="h4" ml={2} mb={1}>
+        Send is disabled for the Demo team
```
jessicamcinchak, line 62: Similar to above! Is "Demo team" public-facing language we want to use? Also let's pick either "disabled" or "not enabled" for consistency!

jessicamcinchak (commented):
A few small language questions/suggestions - but Pay & Send are working as expected :heavy_check_mark: 

Re: Planning Constraints, while we can "turn this on" in the demo team, it WILL return "false negatives". 

We allow applicants to enter any postcode in all of England or Wales per the OS Places API in FindProperty, _but_ Planning Data only hosts constraints for ODP-onboarded councils. Planning Data's API returns the same response for constraints _that have been checked but don't intersect_ and for constraints _that do not have underlying source data to check in the first place_ - therefore why we have the manual toggle in the first place.

I personally think it's a lot more confusing/inaccurate to show folks "working" constraints that are going to say something doesn't apply when they may personally know that it does (but they haven't submitted their datasets to Planning Data yet) and have to explain the concept of "false negatives" - therefore would prefer an error message in that case as well please !!

editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
RODO94: I guess it depends what you mean by outsiders, I'm imagining it is the `demoUser` testing their services themselves or testing them with colleagues because they shouldn't be testing this with the actual public. 

For these people, they would access their flows and create new flows through the "Demo" team, so they would have a notion of what the `Demo` team is. 

Demonstrations or "trial workspace" might bring new terminology here in amongst being a demo user inside the demo team, so I'd rather keep the demo team wording.  

For instance, if you were to do a demonstration for someone, you would probably want to show planning constraints, successful GOV.UK Pay stuff etc... so it's different to what the `Demo` team is used for, which is a space for new councils to experiment and test the platform? 

I agree that the wording should change though... I'll have a think and maybe ask the Content Team or Sylvia

jessicamcinchak (approved):
All working as expected - one final note for a slightly cleaner Planning Constraints error please!

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
```diff
@@ -209,6 +211,22 @@ function Component(props: Props) {
...
+  if (teamSlug === "demo") {
+    return (
+      <Card handleSubmit={props.handleSubmit}>
+        <ErrorSummaryContainer role="status">
+          <Typography variant="h4" ml={2} mb={1}>
+            Planning Constraints are not enabled for demo users.
+          </Typography>
+          <Typography variant="body2" ml={2}>
+            Click continue to skip planning constraints and proceed with your
+            application for testing.
```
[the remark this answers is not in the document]
RODO94: Yeah good shout on both, I'll add these in!

PR #3939, opened 2024-11-12
feat: can toggle feedback required and show validation errors

jamdelion:
### In this PR:

- Added validation schema for public feedback component (make fields required if 'feedback required' toggle is true)
- Show error messages if required fields not filled in
- Added test coverage for these scenarios, including splitting `public.test.tsx` into two test files (one for accessibility, one for submission tests)

Screenshot 2024-11-12 at 14 45 36


editor.planx.uk/src/@planx/components/Feedback/model.ts
```diff
@@ -24,3 +27,21 @@ export const parseFeedback = (
...
+    feedbackScore: feedbackRequired
+      ? number()
+          .integer()
+          .min(1, "Feedback score must be at least 1")
+          .max(5, "Feedback score cannot exceed 5")
+          .required("Please provide a feedback score")
+      : number()
+          .integer()
+          .min(1, "Feedback score must be at least 1")
+          .max(5, "Feedback score cannot exceed 5"),
```
jessicamcinchak, line 45: nit: while this validation schema is totally correct, are `min` & `max` actually useful user-facing errors here since the user is interacting with emoji expressions rather than number scores? Basically I think I'd vote `.required()` or undefined is plenty here? 

Also perhaps `"Please rate your experience"` would better align to default label text rather than `"Please provide a feedback score"` ?

jessicamcinchak (approved):
One minor comment but overall works as expected & thanks for clear test coverage !!

PR #3943, opened 2024-11-13
feat: `ALLOW_LIST` docs

RODO94:


doc/how-to/how-to-add-allow-list-variables.md
```diff
@@ -0,0 +1,111 @@
...
+      "service.type",
+      "usedFOIYNPP",
+      "user.role",
+      *your variable*,
+    ] as const;
+    ```
+
+  Add your variable to this array.
+
+  *It is worth noting here that variables should be added in alphabetical order and relate to a respective passport variable*
```
jessicamcinchak, line 34: nit: it'd be great to add a little more context here like "Only passport variables that store non-personally-identifiable values should be added to the allow list. Sensitive or personally-identifiable data should _never_ be tracked or exposed via analytics."

doc/how-to/how-to-add-allow-list-variables.md
```diff
@@ -0,0 +1,111 @@
...
+  With both arrays now populated with the new variable we can now add them to our database views in Hasura
+### Step 2 - Adding variables to Hasura Views
+
+  For this part, you will need to pull the variable out of the passport jsonb in the SQL script and put them into two database views,`analytics_summary` and `submission_services_summary`. Examples can be found already in the current view script for how to do this.
+
+  Example to look for:
+
+```sql
+  /* Example from submission_services_summary */
+  (((ls.allow_list_answers -> 'application.type'::text) -> 0))::text AS application_type
```
jamdelion, line 45: Could you make clear here which part is `your_variable`?

doc/how-to/how-to-add-allow-list-variables.md
```diff
@@ -0,0 +1,111 @@
...
+    WHERE ((f.slug IS NOT NULL) AND (t.slug IS NOT NULL));
+```
+
+  The values here are being pulled from the table `lowcal_sessions.allow_list_answers` or `analytics_logs.allow_list_answers`
+
+  **At the end of your SQL script after the view creation/replacement, it is important to add another line after each one.**
+
+  `GRANT SELECT ON public.{VIEW_NAME} TO metabase_read_only`
+
+  So, your migration may end with something like this:
```
jamdelion, line 94: Maybe make this a bit more definite, like `We currently have two views, so you should add these two lines to the end of the migration file:`

jamdelion (commented):
Thanks for putting together such a thorough doc! 🎉  Just some minor wording changes to look at :)

jessicamcinchak (approved):

PR #3945, opened 2024-11-13
feat: add `description` and `templated_from` to `flows` table

RODO94:
To help lay the groundwork for the Templates feature, I have added two new columns to the `flows` table:

`description` and `templated_from`

I have updated the necessary permissions so admins, editors, and demoUsers can all create, update, and see `description`.

No permissions have been enabled for `templated_from`.

I'll follow this PR with another one to create an input for `description` in service settings of the flowEditor

### DIFF
--- a/hasura.planx.uk/migrations/1731498447017_add template from column to flows/up.sql
+++ b/hasura.planx.uk/migrations/1731498447017_add template from column to flows/up.sql
@@ -1,0 +1,1 @@
+alter table "public"."flows" add column "templated_from" timestamp;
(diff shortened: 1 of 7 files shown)


hasura.planx.uk/migrations/1731498447017_add template from column to flows/up.sql
```diff
@@ -0,0 +1 @@
+alter table "public"."flows" add column "templated_from" timestamp;
```
jessicamcinchak, line 1: Sorry think this might have been my typo on Trello but `templated_from` should actually be `uuid` type, not timestamp, same as `copied_from` !

jessicamcinchak (commented):
Permissions all look good here! One small comment re: data type (was my mistake on Trello). Also, not crucial, but please consider squashing migrations when single PR and commands on same table !

jessicamcinchak (approved):
Looks good, thanks for updates!

PR #3948, opened 2024-11-13
feat: feedback analytics and previously submitted answers

jamdelion:
### What's in this PR?

- Add `_feedback` to allow_lists
- Update `submission_services_summary` and `analytics_summary` views to include `feedback_score`.
- fix: access previously submitted answers when the user presses back

Verified that feedback_score is an `int` and displaying in `analytics_summary` ✅ 
Screenshot 2024-11-19 at 12 13 32

🎟️ Trello ticket: https://trello.com/c/jEOLoubW/3034-can-we-introduce-in-house-end-of-service-feedback

### DIFF
--- a/api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
+++ b/api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
@@ -22,7 +22,6 @@ const ALLOW_LIST = [
   "application.type",
   "drawBoundary.action",
   "_feedback",
-  "_feedback.feedbackScore",
   "findProperty.action",
   "_overrides",
   "planningConstraints.action",
--- a/api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
+++ b/api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
@@ -22,6 +22,7 @@ const ALLOW_LIST = [
   "application.type",
   "drawBoundary.action",
   "_feedback",
+  "_feedback.feedbackScore",
   "findProperty.action",
   "_overrides",
   "planningConstraints.action",
@@ -63,7 +64,7 @@ export const trackAllowListAnswers: Operation = async () => {
 
     const id = await updateLowcalSessionAllowListAnswers(
       sessionId,
-      allowListAnswers,
+      allowListAnswers
     );
     if (id) updatedSessionIds.push(id);
   }
@@ -103,7 +104,7 @@ export const getSubmittedUnAnalyzedSessionIds = async (): Promise<string[]> => {
  */
 export const updateLowcalSessionAllowListAnswers = async (
   sessionId: string,
-  allowListAnswers: Passport["data"],
+  allowListAnswers: Passport["data"]
 ): Promise<string> => {
   try {
     const mutation = gql`
@@ -126,7 +127,7 @@ export const updateLowcalSessionAllowListAnswers = async (
     return id;
   } catch (error) {
     throw new Error(
-      `Error updating allow_list_answers for lowcal_session ${sessionId}`,
+      `Error updating allow_list_answers for lowcal_session ${sessionId}`
     );
   }
 };
--- a/api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
+++ b/api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
@@ -1,6 +1,7 @@
 import { gql } from "graphql-request";
 
 import { Passport } from "@opensystemslab/planx-core";
+
 import { $api } from "../../../../client/index.js";
 import type { Operation } from "../sanitiseApplicationData/types.js";
 
@@ -20,6 +21,7 @@ const ALLOW_LIST = [
   "application.information.sensitive",
   "application.type",
   "drawBoundary.action",
+  "_feedback",
   "findProperty.action",
   "_overrides",
   "planningConstraints.action",
--- a/api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
+++ b/api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
@@ -64,7 +64,7 @@ export const trackAllowListAnswers: Operation = async () => {
 
     const id = await updateLowcalSessionAllowListAnswers(
       sessionId,
-      allowListAnswers
+      allowListAnswers,
     );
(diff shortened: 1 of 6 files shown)


hasura.planx.uk/migrations/1731498724636_run_sql_migration/up.sql
```diff
@@ -0,0 +1,160 @@
...
+    ((ls.allow_list_answers -> 'usedFOIYNPP'::text))::text AS used_foiynpp,
+    ((ls.allow_list_answers -> 'propertyInformation.action'::text))::text AS property_information_action,
+    ((ls.allow_list_answers -> 'planningConstraints.action'::text))::text AS planning_constraints_action,
+    ((ls.allow_list_answers -> '_overrides'::text))::text AS overrides,
+    ((ls.allow_list_answers -> 'rab.exitReason'::text))::text AS rab_exit_reason,
+    ((ls.allow_list_answers -> 'service.type'::text))::text AS pre_app_service_type,
+    ((ls.allow_list_answers -> 'application.information.harmful'::text))::text AS pre_app_harmful_info,
+    ((ls.allow_list_answers -> 'application.information.sensitive'::text))::text AS pre_app_sensitive_info,
+    (((ls.allow_list_answers -> 'application.type'::text) -> 0))::text AS application_type,
+    ((ls.allow_list_answers -> '_feedback'::text))::text AS feedback
```
jessicamcinchak, line 146: Question: since `_feedback` is an obejct itself, should the views actually drill down one level further to, for example, expose feedback score as its own column? Otherwise I think we'll hit up against usual "Metabase can't process JSONB types" error ! 

Score seems like obvious first priority for analytics, I'd leave it up to you if we want to expose other `_feedback` fields too now/up front or wait to expose them on request later if it comes up!

api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
```diff
@@ -20,6 +21,8 @@ const ALLOW_LIST = [
   "application.information.sensitive",
   "application.type",
   "drawBoundary.action",
+  "_feedback",
+  "_feedback.feedbackScore",
```
jamdelion, line 25: Not entirely sure whether `_feedback.feedbackScore` is needed here, and if it is, whether it is in the right format... The `sql` file does not have `_feedback.feedbackScore` in it directly, but seemed to work as it is!

e.g. this is from local hasura, analytics table:
Screenshot 2024-11-14 at 17 06 11

hasura.planx.uk/migrations/1731498724636_run_sql_migration/up.sql
jamdelion: Have attempted to implement this, but wasn't really very sure of it!

Also struggled to test it - eventually realised it was only populating the analytics tables when I finished a flow - is that right? 😅

api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
jessicamcinchak: Sorry for confusion here! Here's what I'd expect:
- _only_ `"_feedback"` should be included in this list. Dot-separated variables only work when it's the "whole" property name and doesn't require destructuring a value, which `"_feedback.feedbackScore"` would
- This means that `feedback` column in the raw analytics table will look exactly like you've posted above - a jsonb object ! 
- then in the SQL to generate the view you should be able to do something like `((ls.allow_list_answers -> '_feedback' -> 'feedbackScore'::int))::int AS feedback_score` so that what we're exposing to Metabase is the feedback score as an integer (Metabase can't process jsonb! And free-text `userComments` are nice to track in raw table but aren't very chart-able)

Does this make sense?

hasura.planx.uk/migrations/1731498724636_run_sql_migration/up.sql
jessicamcinchak: Full comment below - let's try something like `((ls.allow_list_answers -> '_feedback' -> 'feedbackScore'::int))::int AS feedback_score` to extract the view-level column data !

api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
jamdelion: Thanks Jess, the `int` version you put above (and variations I have tried) caused syntax errors `"message": "invalid input syntax for type integer: \"feedbackScore\"",` - I assume the reason it shouldn't be text is because Metabase can make nice charts with numbers and can't convert from string numbers to actual numbers?

Have added a column for userComment in the table, and I assume that the feedback column (displaying jsonb) is completely redundant then, given that we now have separate columns for `feedback_score` and `user_comment`. Figuring out a way to drop that column from the view if so!

jamdelion: Screenshot 2024-11-18 at 21 02 44

hasura.planx.uk/migrations/1731498724636_run_sql_migration/up.sql
```diff
@@ -0,0 +1,159 @@
...
+    sa.s3_applications,
+    ((ls.allow_list_answers -> 'usedFOIYNPP'::text))::text AS used_foiynpp,
+    ((ls.allow_list_answers -> 'propertyInformation.action'::text))::text AS property_information_action,
+    ((ls.allow_list_answers -> 'planningConstraints.action'::text))::text AS planning_constraints_action,
+    ((ls.allow_list_answers -> '_overrides'::text))::text AS overrides,
+    ((ls.allow_list_answers -> 'rab.exitReason'::text))::text AS rab_exit_reason,
+    ((ls.allow_list_answers -> 'service.type'::text))::text AS pre_app_service_type,
+    ((ls.allow_list_answers -> 'application.information.harmful'::text))::text AS pre_app_harmful_info,
+    ((ls.allow_list_answers -> 'application.information.sensitive'::text))::text AS pre_app_sensitive_info,
+    ((ls.allow_list_answers -> '_feedback') ->> 'feedbackScore'::text)::int AS feedback_score
```
jessicamcinchak, line 145: tricky casting, but works exactly as expected :100:

jessicamcinchak (approved):
Code looks great! Would suggest logging into Vultr & destroying instance & re-running latest failed CI to see if you can clear up errors? Looks like docker-related errors due to editing migrations I'd guess?

PR #3953, opened 2024-11-14
fix: Always center flows on load

DafyddLlyr:
## What does this PR do?
 - Ensures flow is always centered, after load
 - Alternate implementation of https://github.com/theopensystemslab/planx-new/pull/3854/files#r1816927924
 - This does not change or remove the current session storage of x and y values, looking at this separately 

Following https://github.com/theopensystemslab/planx-new/pull/3854 there was some valuable feedback on how this _should_ behave which I'll pick up in follow up PRs.


DafyddLlyr:
[Flows should be centered within the window on initial load](https://trello.com/c/wIFWO8dr/3083-flows-should-be-centered-within-the-window-on-initial-load)

PR #3954, opened 2024-11-14
feat: add a flow description settings section

RODO94:
## What does this PR do?

This PR introduces a new section in the **Service Settings** page of the Editor. This houses the flow `description` which was added to the `flows` table in a previous PR: https://github.com/theopensystemslab/planx-new/pull/3945

I have created relevant mutations and updated current queries to update, create, and select `description` in `planx-core` in this PR: https://github.com/theopensystemslab/planx-core/pull/564

This new section has been added to the bottom of the **Service Settings** page as not to disrupt the UI too much, although it may be worth reviewing how these things are structured, but I imagined this is most aligned with the "Rethinking the Editor" work rather than this wee PR. 

I've called the setting section "Service Information" in case we want to put anything else in there in the future, type of service, creator, templated dependency etc... and it also allowed me to stop constantly using the word "description"

>[!Note]
> The input structure follows the pattern of going for an Input Label over a placeholder due to the accessibility needs, I'll add a ticket to update the rest of the settings (Footer / Disclaimer) in line with this

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings/FlowDescription/FlowDescription.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings/FlowDescription/FlowDescription.tsx
@@ -15,6 +15,7 @@ const FlowDescription = () => {
     state.flowDescription,
     state.updateFlowDescription,
   ]);
+
   const toast = useToast();
 
   const formik = useFormik<{ description: string }>({
@@ -34,6 +35,9 @@ const FlowDescription = () => {
         );
       }
     },
+    validateOnBlur: false,
+    validateOnChange: false,
+    enableReinitialize: true,
   });
 
   return (
@@ -59,6 +63,7 @@ const FlowDescription = () => {
           <>
             <InputLabel label="Description" htmlFor="description">
               <Input
+                multiline
                 name="description"
                 onChange={(event) => {
                   formik.setFieldValue("description", event.target.value);
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings/FlowDescription/FlowDescription.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings/FlowDescription/FlowDescription.tsx
@@ -1,0 +1,78 @@
+import Box from "@mui/material/Box";
+import Typography from "@mui/material/Typography";
+import { useFormik } from "formik";
+import { useToast } from "hooks/useToast";
+import React from "react";
+import InputLabel from "ui/editor/InputLabel";
+import SettingsSection from "ui/editor/SettingsSection";
+import Input from "ui/shared/Input/Input";
+
+import { useStore } from "../../../../lib/store";
+import { SettingsForm } from "../../shared/SettingsForm";
+
+const FlowDescription = () => {
+  const [flowDescription, updateFlowDescription] = useStore((state) => [
+    state.flowDescription,
+    state.updateFlowDescription,
+  ]);
+  const toast = useToast();
+
+  const formik = useFormik<{ description: string }>({
+    initialValues: {
+      description: flowDescription || "",
+    },
+    onSubmit: async (values, { resetForm }) => {
+      const isSuccess = await updateFlowDescription(values.description);
+      if (isSuccess) {
+        toast.success("Description updated successfully");
+        resetForm({ values });
+      }
+      if (!isSuccess) {
+        formik.setFieldError(
+          "description",
+          "We are unable to update the service description, check your internet connection and try again",
+        );
+      }
+    },
+  });
+
+  return (
+    <Box mb={2}>
+      <SettingsSection>
+        <Typography variant="h2" component="h3" gutterBottom>
+          Service Information
+        </Typography>
+        <Typography variant="body1">
+          Useful information about this service.
+        </Typography>
+      </SettingsSection>
+      <SettingsForm
+        legend="Service Description"
+        formik={formik}
+        description={
+          <>
+            A short blurb on what this service is, how it should be used, and if
+            there are any dependencies related to this service.
+          </>
+        }
+        input={
+          <>
+            <InputLabel label="Description" htmlFor="description">
+              <Input
+                name="description"
+                onChange={(event) => {
+                  formik.setFieldValue("description", event.target.value);
+                }}
+                value={formik.values.description ?? ""}
+                errorMessage={formik.errors.description}
+                id="description"
+              />
+            </InputLabel>
+          </>
+        }
+      />
+    </Box>
+  );
+};
+
+export default FlowDescription;
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings/FooterLinksAndLegalDisclaimer.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings/FooterLinksAndLegalDisclaimer.tsx
@@ -1,12 +1,11 @@
 import Box from "@mui/material/Box";
 import Button from "@mui/material/Button";
-import { formControlLabelClasses } from "@mui/material/FormControlLabel";
+// eslint-disable-next-line no-restricted-imports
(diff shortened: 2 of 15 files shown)


DafyddLlyr:
@RODO94 Just tried testing but can't log into the Pizza - the db sync has failed here. Could you please rebase to pick up https://github.com/theopensystemslab/planx-new/pull/3957, and then recreate Vultr instance to resolve this?

editor.planx.uk/src/pages/FlowEditor/lib/store/settings.ts
```diff
@@ -51,6 +54,20 @@ export const settingsStore: StateCreator<
...
+
+  setFlowDescription: (flowDescription: string) => set({ flowDescription }),
+
+  updateFlowDescription: async (newDescription: string) => {
+    const { id, $client } = get();
+    const result = await $client.flow.setDescription({
+      flow: { id },
+      description: newDescription,
+    });
+    set({ flowDescription: newDescription });
```
DafyddLlyr, line 67: Despite this change, I'm not seeing the value updated when I toggle pages - 


https://github.com/user-attachments/assets/986b3c7c-a1e6-439d-9795-606d9098a2c1

DafyddLlyr: This is happening because Apollo is overwriting Zustand's state on route change. 

I think this is a much broader problem we need to solve across the app rather than just patching it here there and everywhere, but in the meantime we shouldn't introduce buggy behaviour.

editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings/FlowDescription/FlowDescription.tsx
```diff
@@ -0,0 +1,78 @@
...
+        description={
+          <>
+            A short blurb on what this service is, how it should be used, and if
+            there are any dependencies related to this service.
+          </>
+        }
+        input={
+          <>
+            <InputLabel label="Description" htmlFor="description">
+              <Input
```
DafyddLlyr, line 61: Following the pattern we've established in https://github.com/theopensystemslab/planx-new/pull/3933 we should make this a larger text input if as we're expecting more than a few words here (we are!).

Two other question (maybe already discussed / agree and I've missed them) -
 - should this be just plain text, not rich text?
 - should there be a limit on the length of the input?

editor.planx.uk/src/pages/FlowEditor/lib/store/settings.ts
RODO94: Ah yes, I forgot this from the team settings stuff, I'll put up a fix now! thanks for catching this

editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings/FlowDescription/FlowDescription.tsx
RODO94: Great shout, added the multiline prop now

For the questions, my initial thought was this was just laying the groundwork and these types of answers would come once we have a better idea when / how we display it to the user, so for now I just kept it to it's simplest implementation. Do you think we need to take it further in this PR?

DafyddLlyr: I think this is fine for now - just checking it wasn't a decision somewhere I'd missed. This will be easy to improve in future as you say if required.

editor.planx.uk/src/pages/FlowEditor/lib/store/settings.ts
```diff
@@ -51,6 +59,55 @@ export const settingsStore: StateCreator<
...
+    return Boolean(result?.id);
+  },
+
+  getFlowSettings: async (flowSlug, teamSlug) => {
```
DafyddLlyr, line 76: I found the naming here is a little unclear / unexpected - we already have a `flowSettings` property and this is not a getter for that property, but rather a function which fetches settings, status and description of a flow.

I think I'd stick to the existing pattern of -
 - Fetch within the route
 - Set store variables
 - Child components read from stores
 
 Always using the route `req` object over store values also ensures we're always fetching the right data and there's no room for any race condition or unexpected behaviour. This relies on a user always navigating via a route which sets these.

DafyddLlyr: https://github.com/user-attachments/assets/001fee03-b878-4deb-a82d-1f00b24631f1

editor.planx.uk/src/routes/serviceSettings.tsx
```diff
@@ -1,48 +1,10 @@
...
-    variables: {
-      slug: req.params.flow,
-      team_slug: req.params.team,
-    },
-  });
-
-  useStore.getState().setFlowSettings(settings);
-  useStore.getState().setFlowStatus(status);
-};
+const { flowSlug, teamSlug, getFlowSettings } = useStore.getState();
```
DafyddLlyr, line 7: Please see previous comment - this likely isn't the most robust way of handling this request.

Just a general note on this pattern though - we're fetching variables from the store, to pass back into a store function. It's cleaner and neater just to call `const { flowSlug, teamSlug } = get()` within the store function when we need to do this 👍

DafyddLlyr (changes requested):

editor.planx.uk/src/pages/FlowEditor/lib/store/settings.ts
RODO94: Yeah good thinking, I did keep the function name the same for consistency, but probably need some altered naming here. It is complicated a bit by the current `flowSettings` not really being all the settings for a flow, and rather an object of footer, privacy, and help rich text info. I'll look to update the naming to make it clearer 

The `req` bit is a great catch, sorry for missing this. Added the change now. 

Would there be a reason why these settings maintain a different pattern to General Settings and Design Settings where we fetch within the index.tsx of the component? 

Think I've fallen into a trap of trying to partially improve code while trying to debug and creating wee errors, thanks for all the help and time

editor.planx.uk/src/pages/FlowEditor/utils.ts
```diff
@@ -1,4 +1,17 @@
+import { FlowStatus } from "@opensystemslab/planx-core/types";
 import { formatDistanceToNow } from "date-fns";
+import { FlowSettings } from "types";
+
+export interface FlowInformation {
+  id?: string;
```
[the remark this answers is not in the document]
RODO94: Ah no, was mostly to get the getTeamSettings working, updated now

editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings/FooterLinksAndLegalDisclaimer.tsx
```diff
@@ -1,12 +1,11 @@
 import Box from "@mui/material/Box";
 import Button from "@mui/material/Button";
-import { formControlLabelClasses } from "@mui/material/FormControlLabel";
+// eslint-disable-next-line no-restricted-imports
```
DafyddLlyr, line 3: Good catch!

editor.planx.uk/src/pages/FlowEditor/lib/store/settings.ts
DafyddLlyr: > Would there be a reason why these settings maintain a different pattern to General Settings and Design Settings where we fetch within the index.tsx of the component?

Honestly, I think it's just history! This has been around for a while and we've not yet taken the dive to standardise and update it to match the rest - not a pressing concern right now, but would be a super nice tidy up in future.

> Think I've fallen into a trap of trying to partially improve code while trying to debug and creating wee errors, thanks for all the help and time

Very easily done - especially in sections of the code like this which are too low on test coverage.

DafyddLlyr (approved):
All working as expected 👌

RODO94:
[Templates groundwork](https://trello.com/c/YjhjFaIQ/3111-templates-groundwork)

PR #3955, opened 2024-11-14
feat: Center previous portal when navigating back

DafyddLlyr:


jessicamcinchak (approved):
Was nice to see this demo'd last week, feels like solid solution at this stage !

PR #3957, opened 2024-11-14
fix: sync all columns in `flows` table

jessicamcinchak:
`pnpm run up` on a fresh database is currently failing with the following error message (because we `select * from flows` but are missing recently added column defs)
Screenshot from 2024-11-14 15-19-47

### DIFF
--- a/scripts/seed-database/write/flows.sql
+++ b/scripts/seed-database/write/flows.sql
@@ -60,8 +60,8 @@ SET
   copied_from = EXCLUDED.copied_from,
   analytics_link = NULL,
   status = EXCLUDED.status,
-  name = EXCLUDED.name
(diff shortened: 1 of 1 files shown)


jamdelion (approved):

scripts/seed-database/write/flows.sql
```diff
@@ -54,7 +60,9 @@ SET
   copied_from = EXCLUDED.copied_from,
   analytics_link = NULL,
   status = EXCLUDED.status,
-  name = EXCLUDED.name;
+  name = EXCLUDED.name
+  templated_from = EXCLUDED.templated_from
```
DafyddLlyr, line 64: Not sure if trailing commas are required here?

DafyddLlyr (approved):

PR #3959, opened 2024-11-14
feat: remove `onDelete` for Placeholder tag for non Platform admiins

RODO94:
## What does this PR do?

This PR adds a role access constraint to the `Placeholder` tag so only `PlatformAdmins` can edit it, but everyone can see it. 

To do this, I have restricted the `renderOption` function so it only shows `Placeholder` tag in the MultiSelect for `PlatformAdmin` users. I have also added the constraint to the `renderTags` function, so it removes the `onDelete` from the `Chip` component when a user is not a `PlatformAdmin`

With the changes to `TAG_DISPLAY_VALUES` we can now use `Role` types to add further permission layers in future if required. To ensure this happens, new logic would need to be introduced to the `canEdit()` in `ComponentTagSelect()`

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Tag.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Tag.tsx
@@ -8,11 +8,12 @@ import { FONT_WEIGHT_SEMI_BOLD } from "theme";
 
 export const TAG_DISPLAY_VALUES: Record<
   NodeTag,
-  { color: keyof Palette["nodeTag"]; displayName: string }
+  { color: keyof Palette["nodeTag"]; displayName: string; isEditable?: boolean }
 > = {
   placeholder: {
     color: "blocking",
     displayName: "Placeholder",
+    isEditable: useStore.getState().user?.isPlatformAdmin,
   },
   toReview: {
     color: "nonBlocking",
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Tag.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Tag.tsx
@@ -8,12 +8,12 @@ import { FONT_WEIGHT_SEMI_BOLD } from "theme";
 
 export const TAG_DISPLAY_VALUES: Record<
   NodeTag,
-  { color: keyof Palette["nodeTag"]; displayName: string; isEditable?: Role }
+  { color: keyof Palette["nodeTag"]; displayName: string; isEditable?: Role[] }
 > = {
   placeholder: {
     color: "blocking",
     displayName: "Placeholder",
-    isEditable: "platformAdmin", // if new roles are added, we should update the canEdit() in ComponentTagSelect.tsx
+    isEditable: ["platformAdmin"], // if new roles are added, we should update the canEdit() in ComponentTagSelect.tsx
   },
   toReview: {
     color: "nonBlocking",
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Tag.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Tag.tsx
@@ -8,12 +8,12 @@ import { FONT_WEIGHT_SEMI_BOLD } from "theme";
 
 export const TAG_DISPLAY_VALUES: Record<
   NodeTag,
-  { color: keyof Palette["nodeTag"]; displayName: string; isEditable?: Role[] }
+  { color: keyof Palette["nodeTag"]; displayName: string; editableBy?: Role[] }
 > = {
   placeholder: {
     color: "blocking",
     displayName: "Placeholder",
-    isEditable: ["platformAdmin"], // if new roles are added, we should update the canEdit() in ComponentTagSelect.tsx
+    editableBy: ["platformAdmin"], // if new roles are added, we should update the canEdit() in ComponentTagSelect.tsx
   },
   toReview: {
     color: "nonBlocking",
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Tag.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Tag.tsx
@@ -1,6 +1,6 @@
 import Box from "@mui/material/Box";
 import { Palette, useTheme } from "@mui/material/styles";
-import { NodeTag } from "@opensystemslab/planx-core/types";
+import { NodeTag, Role } from "@opensystemslab/planx-core/types";
 import { useStore } from "pages/FlowEditor/lib/store";
 import React from "react";
 import { getContrastTextColor } from "styleUtils";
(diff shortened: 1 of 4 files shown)


editor.planx.uk/src/ui/editor/ComponentTagSelect.tsx
```diff
@@ -16,42 +17,65 @@ interface Props {
...
-  true,
-  false,
-  "div"
->["renderOption"] = (props, tag, { selected }) => (
-  <ListItem {...props}>
-    <CustomCheckbox aria-hidden="true" className={selected ? "selected" : ""} />
-    {TAG_DISPLAY_VALUES[tag].displayName}
-  </ListItem>
-);
+export const ComponentTagSelect: React.FC<Props> = ({ value, onChange }) => {
```
DafyddLlyr, line 20: Zustand has a non-hook API for use outside of React components, which means you don't have to nest these within `ComponentTagSelect` - 

```ts
const isPlatformAdmin = useStore.getState().user?.isPlatformAdmin;
```

editor.planx.uk/src/ui/editor/ComponentTagSelect.tsx
```diff
@@ -16,42 +17,65 @@ interface Props {
...
-        backgroundColor: theme.palette.nodeTag[TAG_DISPLAY_VALUES[tag].color],
-        color: getContrastTextColor(
-          theme.palette.nodeTag[TAG_DISPLAY_VALUES[tag].color],
-          "#FFF",
-        ),
-      })}
-    />
-  ));
+  const showPlaceholderTag = (tagName: string) =>
+    tagName.toLowerCase() === "placeholder" && !isPlatformAdmin ? true : false;
```
DafyddLlyr, line 24:
```suggestion
  const showPlaceholderTag = (tagName: string) =>
    tagName.toLowerCase() === "placeholder" && !isPlatformAdmin
```

DafyddLlyr (changes requested):
There might be a more generalisable pattern here than just toggling controls for a single named tag.

Could we set rules in an object, maybe a renamed/repurposed `TAG_DISPLAY_VALUES` that controls the behaviour?

```ts
export const TAG_DISPLAY_VALUES: Record<
  NodeTag,
  { color: keyof Palette["nodeTag"]; displayName: string, isEditable?: boolean }
> = {
  placeholder: {
    color: "blocking",
    displayName: "Placeholder",
    isEditable: useStore.getState().user?.isPlatformAdmin
  },
  ...
```

I think this would give us a more robust, scalable, and data driven solution.

Also, it would be great to see some tests here describing the change please 🙏

editor.planx.uk/src/ui/editor/ComponentTagSelect.tsx
```diff
@@ -17,65 +16,55 @@ interface Props {
...
-
-    if (showPlaceholderTag(tagName)) return;
+const renderOption: AutocompleteProps<
+  NodeTag,
+  true,
+  true,
+  false,
+  "div"
+>["renderOption"] = (props, tag, { selected }) => {
+  if (TAG_DISPLAY_VALUES[tag].isEditable) return null;
```
DafyddLlyr: This reads as incorrect to me - if it's editable, don't return anything. 

I'd expect the opposite behaviour - if a tag is _not_ editable, don't display it.

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Tag.tsx
```diff
@@ -1,18 +1,19 @@
...
 
 export const TAG_DISPLAY_VALUES: Record<
   NodeTag,
-  { color: keyof Palette["nodeTag"]; displayName: string }
+  { color: keyof Palette["nodeTag"]; displayName: string; isEditable?: Role }
 > = {
   placeholder: {
     color: "blocking",
     displayName: "Placeholder",
+    isEditable: "platformAdmin", // if new roles are added, we should update the canEdit() in ComponentTagSelect.tsx
```
DafyddLlyr, line 16: I'd expect a variable starting with `is` to be a boolean. This would be better with a different name (and maybe an array of roles?) if we want to just list roles here. 

It seems like a simpler option would be a boolean. It looks like you hit a test issue when trying this when I check out [the commit history](https://github.com/theopensystemslab/planx-new/pull/3959/commits/55c3236b1fbcec69fc71ee304d675ccbeffa1c95). Did you manage to work out what the issue here was? It might be a relatively simple case of just mocking the useStore import in these tests.

editor.planx.uk/src/ui/editor/ComponentTagSelect.test.tsx
DafyddLlyr, line 1: Thanks for adding this 👍

editor.planx.uk/src/ui/editor/ComponentTagSelect.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    act(() =>
+      setState({
+        user: {
+          ...mockUser,
+          isPlatformAdmin: true,
+        },
+      }),
+    ),
+  );
+  it("renders all tags with none selected", async () => {
```
DafyddLlyr, line 33: nit: Easier to read with more whitespace (imho!)

```suggestion
  );
  
  it("renders all tags with none selected", async () => {
```

editor.planx.uk/src/ui/editor/ComponentTagSelect.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const options = within(optionsList).getAllByRole("option");
+
+    const tagDisplayNames = Object.values(TAG_DISPLAY_VALUES).map(
+      (tag) => tag.displayName,
+    );
+    const optionTexts = options.map((option) => option.textContent);
+
+    expect(optionTexts).toEqual(expect.arrayContaining(tagDisplayNames));
+  });
+  it("renders all tags with Placeholder selected as a button", async () => {
```
DafyddLlyr, line 53:
```suggestion
  });
  
  it("renders all tags with Placeholder selected as a button", async () => {
```

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Tag.tsx
DafyddLlyr: Ah fair enough, this makes sense! In which case, I think that updating variable name and maybe making this a `Role[]` is the way to go here 👍

DafyddLlyr (changes requested):
This still feels a bit more complex and hard to follow than it needs to be.

I'd really recommend changing the variable name from `isEditable` (which denotes a boolean) to something more indicative of a `Role[]` - maybe `editableBy`?

Have a look at how `EditorNavMenu` solves a very similar problem.

RODO94:
> This still feels a bit more complex and hard to follow than it needs to be.
> 
> I'd really recommend changing the variable name from `isEditable` (which denotes a boolean) to something more indicative of a `Role[]` - maybe `editableBy`?
> 
> Have a look at how `EditorNavMenu` solves a very similar problem.

Nah you're definitely right, this slipped my mind when making the changes, sorry!

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Tag.tsx
```diff
@@ -8,12 +8,12 @@ import { FONT_WEIGHT_SEMI_BOLD } from "theme";
...
 export const TAG_DISPLAY_VALUES: Record<
   NodeTag,
-  { color: keyof Palette["nodeTag"]; displayName: string; isEditable?: Role[] }
+  { color: keyof Palette["nodeTag"]; displayName: string; editableBy?: Role[] }
 > = {
   placeholder: {
     color: "blocking",
     displayName: "Placeholder",
-    isEditable: ["platformAdmin"], // if new roles are added, we should update the canEdit() in ComponentTagSelect.tsx
+    editableBy: ["platformAdmin"], // if new roles are added, we should update the canEdit() in ComponentTagSelect.tsx
```
DafyddLlyr, line 16: If we use the existing `state.getUserRoleForCurrentTeam()` we won't need this comment or to maintain and update `canEdit` which will share the same functionality 👌

DafyddLlyr (changes requested):

DafyddLlyr (approved):

PR #3967, opened 2024-11-15
fix: Only trigger `email_user_submission_confirmation` event once per session

DafyddLlyr:
## What's the problem?
 - When an application is submitted to multiple locations, the `email_user_submission_confirmation` event trigger is called multiple times, resulting in multiple emails being sent to applicants from PlanX.
 - This is recreatable locally, and can be observed as a pattern via GovNotify or the Hasura event log - we can see grouped events and emails for the same PlanX session

## What's the solution?
 - Adds a check within the query, so that the trigger is only called once

### DIFF
--- a/api.planx.uk/modules/saveAndReturn/service/utils.ts
+++ b/api.planx.uk/modules/saveAndReturn/service/utils.ts
@@ -217,17 +217,26 @@ const softDeleteSession = async (sessionId: string) => {
 
 /**
  * Mark a lowcal_session record as submitted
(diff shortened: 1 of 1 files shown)


DafyddLlyr:
[Fix multiple confirmation mails being sent for a single submission](https://trello.com/c/nIohH4Xj/3113-fix-multiple-confirmation-mails-being-sent-for-a-single-submission)

jessicamcinchak (approved):
Perfect, thank you!

PR #3968, opened 2024-11-17
feat: editors can select which planning constraints to check

jessicamcinchak:
**Changes & testing notes:**
- PlanningConstraints editor modal now allows you select/de-select datasets to check (you must select at least one)
  - The public component only fetches and displays the datsets that you've selected
  - The passport (`property.constraints.planning` & `_nots["property.constraints.planning"]`) only include data values for the datasets that you've selected
- Existing PlanningConstraints components fetch all datasets by default
- Validation check prevents you from publishing a flow with more than one PlanningConstraints component

Further context in Slack here https://opensystemslab.slack.com/archives/C5Q59R3HB/p1732703826935819


jessicamcinchak:
[Give 'Planning Constraints' component datasets a 'hide from user' toggle](https://trello.com/c/Vjb3omgj/2845-give-planning-constraints-component-datasets-a-hide-from-user-toggle)

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
```diff
@@ -56,8 +60,7 @@ function Component(props: Props) {
 
   // PlanningConstraints must come after at least a FindProperty in the graph
   const showGraphError = !longitude || !latitude;
-  if (showGraphError)
-    throw new GraphError("mapInputFieldMustFollowFindProperty");
+  if (showGraphError) throw new GraphError("nodeMustFollowFindProperty");
```
jessicamcinchak, line 63: This is a small unrelated bug fix - this was incorrectly displaying the Map & Label "schema" graph error, when it should be a more generic sequence error

api.planx.uk/modules/gis/service/digitalLand.ts
```diff
@@ -71,15 +71,30 @@ async function go(
...
+    // else default to the internally maintained list of all "active" datasets
+    Object.keys(baseSchema).forEach((key) => {
+      if (baseSchema[key]["active"]) {
+        activeDataValues.push(key);
+        baseSchema[key]["digital-land-datasets"]?.forEach((dataset: string) => {
+          activeDatasets.push(dataset);
+        });
+      }
+    });
+  }
```
jessicamcinchak, line 97: All _new_ Planning Constraint requests will include `vals` param, but we also want to support existing "legacy" GIS requests that have been shared via ODP Schema application payloads (eg BOPS uses to "re-fetch" constraints). 

Therefore, both of these will work: 
- (with `vals` only fetches selected datasets) http://localhost:7002/gis/opensystemslab?geom=MULTIPOLYGON+%28%28%28-0.076303+51.485923%2C+-0.076307+51.48591%2C+-0.075586+51.485849%2C+-0.075574+51.485849%2C+-0.075569+51.485903%2C+-0.076293+51.485963%2C+-0.076303+51.485923%29%29%29&vals=article4
- (without `vals` defaults to all "active" datasets as before) http://localhost:7002/gis/opensystemslab?geom=MULTIPOLYGON+%28%28%28-0.076303+51.485923%2C+-0.076307+51.48591%2C+-0.075586+51.485849%2C+-0.075574+51.485849%2C+-0.075569+51.485903%2C+-0.076293+51.485963%2C+-0.076303+51.485923%29%29%29

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
```diff
@@ -71,79 +74,80 @@ function Component(props: Props) {
...
     ...data?.metadata,
     ...roads?.metadata,
   };
 
   const handleSubmit = () => {
     // `_constraints` & `_overrides` are responsible for auditing
-    const _constraints: Array<
-      EnhancedGISResponse | GISResponse["constraints"]
-    > = [];
+    const _constraints: Array<EnhancedGISResponse> = [];
```
jessicamcinchak, line 139: There's a couple follow-on type tidy ups here that I missed when retiring custom GIS in #3977

api.planx.uk/modules/flows/validate/service/planningConstraints.ts
```diff
@@ -0,0 +1,38 @@
...
+  );
+  if (numberOfPlanningConstraintNodes > 1) {
+    return {
+      title: "Planning Constraints",
+      status: "Fail",
+      message:
+        "When using Planning Constraints, your flow must have exactly ONE Planning Constraints component",
+    };
+  } else if (numberOfPlanningConstraintNodes === 1) {
+    // In future, add extra `hasPlanningData` validation step here to ensure integration is available for this team
```
jessicamcinchak, line 23: This is a nice-to-have, but will require team settings context as well as the flattened flow data structure to be available here. The editor preview panel already shows a warning if the integration isn't available for your team, so definitely not urgent!

api.planx.uk/modules/gis/service/digitalLand.ts
```diff
@@ -71,15 +71,30 @@ async function go(
...
   const activeDatasets: string[] = [];
-  Object.keys(baseSchema).forEach((key) => {
-    if (baseSchema[key]["active"]) {
-      baseSchema[key]["digital-land-datasets"]?.forEach((dataset: string) => {
-        activeDatasets.push(dataset);
-      });
-    }
-  });
+  let activeDataValues: string[] = [];
+  if (extras?.vals) {
```
DafyddLlyr, line 77: It's tricky as these aren't 1:1 Planning.data datasets, but I think we should use a more meaningful param name here - maybe `datasets`, but this isn't ideal given the context here and the existing meaning `datasets` has.

We also need to add this new param to Swagger docs 👍

api.planx.uk/modules/gis/service/digitalLand.ts
DafyddLlyr: I guess this actually isn't _totally_ stable just now in the same way that the Planning Data `sourceRequest` URL is. If in future there are new datasets added, the old URLs will pick up the full current list and not the one used at the time.

Bit of an edge case, but something to flag 👍

DafyddLlyr (approved):
Looking great! A few minor comments but no show stoppers 👍

editor.planx.uk/src/@planx/components/PlanningConstraints/Editor.tsx
```diff
@@ -84,77 +118,84 @@ function PlanningConstraintsComponent(props: Props) {
...
-                </TableBody>
-              </Table>
-            </TableContainer>
+            <ErrorWrapper error={formik.errors.dataValues}>
+              <TableContainer>
+                <Table size="small">
+                  <TableHead>
+                    <TableRow>
+                      <TableCell>
+                        <Checkbox
```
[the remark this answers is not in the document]
jessicamcinchak: Updated to use styled checkboxes, good shout ! 
Screenshot from 2024-11-29 10-30-35

editor.planx.uk/src/@planx/components/PlanningConstraints/Editor.tsx
```diff
@@ -36,8 +37,41 @@ function PlanningConstraintsComponent(props: Props) {
         data: newValues,
       });
     },
+    validate: (values) => {
+      const errors: FormikErrors<PlanningConstraints> = {};
+      if (!values.dataValues || values.dataValues?.length < 1) {
+        errors.dataValues = "Select at least one constraint";
+      }
+      return errors;
+    },
```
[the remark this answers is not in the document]
jessicamcinchak: Updated :heavy_check_mark:

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
```diff
@@ -36,6 +37,9 @@ export type InaccurateConstraints =
 export default Component;
 
 function Component(props: Props) {
+  // Existing components will not have dataValues prop so should default to all available datasets
+  const dataValues = props.dataValues || availableDatasets.map((d) => d.val);
```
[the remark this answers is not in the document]
jessicamcinchak: Not a bad idea, just a very tedious & high coordination task anytime we do it in my opinion ! 

But agree there's a few wishlist items that have been accumulating that are good fits for a data migration script:
- `flag` -> `flags` prop on answer-type nodes & ensuring values are array
- updating all policy links to strip `/made` param (per this thread https://opensystemslab.slack.com/archives/C07F7215W7P/p1731081559409279) 
- now constraints

I think we're at a point though now where we're no longer talking scripts that are manually ran across a subset of flows in a matter of minutes, but rather a full update of _every_ flow that runs via an overnight scheduled cron job?

editor.planx.uk/src/@planx/components/PlanningConstraints/model.ts
```diff
@@ -5,6 +5,7 @@ export interface PlanningConstraints extends BaseNodeData {
   description: string;
   fn: string;
   disclaimer: string;
+  dataValues?: string[] | undefined;
```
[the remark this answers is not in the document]
jessicamcinchak: Feel quite strongly that `datasets` is NOT a good name because we're not allowing users to say "I only want to check TPO _orders_ dataset from Planning Data, not TPO _zones_" - rather we're letting them select/deselect our pre-curated "options" (each with a title, passport value, and associated source datasets). 

Perhaps `options` is more generic and easier to grasp here though? In my mind, they're basically the PlanningConstraint node's equivalent of Question/Checklist "edges"/"options" ?

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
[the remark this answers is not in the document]
DafyddLlyr: Yeah I'm half hoping that a combination of templates and node history might solve this to a degree (e.g. add a notification or something if there's a new field and a node needs to be saved and updated). 

I'm happy keeping this on the backburner for now, just an observation really 👍

editor.planx.uk/src/@planx/components/PlanningConstraints/model.ts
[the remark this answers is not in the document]
jessicamcinchak: As for the "all" question - I think this becomes much simpler in a word of Templates (eg PlanningConstraints will _not_ be a "placeholder" node therefore just updated once per service template & changes propagated to children on publish), and if still too cumbersome then would favor data migration script like you say! 

Also no near-future requests for more datasets from Planning Data besides Planning History which I think will have a strong argument for it's own component or API outside of existing `/gis`.

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -577,38 +601,52 @@ export const previewStore: StateCreator<
...
+        if (blankOption?.id) { optionsThatCanBeAutoAnswered.push(blankOption.id) };
       }
-    } else {
-      // If we don't have any existing passport values for this fn but we do have a blank option,
-      //  proceed through the blank if every option's val has been visited before
-      const sortedOptionVals: string[] = sortedOptions.map(
-        (option) => option.data?.val,
+    }
+
+    if (data.fn === planningConstraintsFn && (foundPassportValues || nots)) {
```
jessicamcinchak, line 608: In addition to updated unit tests here, logic changes have been thoroughly tested by content team with lots of notes in this thread: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1732703826935819

Summary is:
- I've isolated planning constraints automation rules from larger automation rules in an attempt to have more clarity/flexibility around these - so constraints "rules" are now:
  - If there are intersecting constraints that exactly match the option, automate through that option
  - If there are no intersecting constraints that exactly match the option, but there is a "not" that exactly matches the option and a blank option available, automate through the blank
  - If there are no intersecting constraints or "nots" that exactly match an option (aka the option is something that we have not queried), put to user exactly once (future instances can be considered a "seen" option and therefore automated)
- (Extra credit bonus) Since planning constraints store all granularity levels in the passport (whereas normal behavior is only store most granular), a more isolated ruleset means we can only automate constraints-based questions/checklists now when it's an exact match (eg do not apply startsWith logic as a fallback if no direct matches like non-constraints passport values)
  - This is basically a more focused way to revisit this old pizza experiment https://github.com/theopensystemslab/planx-new/pull/3461 and fixes known issues like  listed.grade.I, listed.grade.II both automating when it's only listed.grade.II

Known edges cases / not handled:
- If we don't check conservation areas, and you manually answer "yes" that you are in one, we aren't going to additionally manually assign/capture the parent-level designated passport value as the API response formatting would handle
  - This might be best handled by slightly adjusting which line items are actually selectable in the editor modal to begin with - eg should "Land designations" be the thing you can toggle, and that automatically queries or doesn't query everything that is prefixed designated. ? This is essentially how others like listed, flood are already doing?
- As you mentioned before, "checklist with some values that can be automated and others that haven't been seen" still persists for planning constraints and beyond. But I think that will fit in really nicely with Jonny's topic for next logic camp (eg maybe instead of full auto-answered, put to user auto-populated with option(s) we do know for certain) :bookmark:

All of my non-constraint automation test suites are still passing without changes so not expecting any wider changes or regressions beyond planning constraint-related questions.

api.planx.uk/modules/gis/service/digitalLand.ts
jessicamcinchak: Went with `vals` because the response includes a `val` property for each item - agree not most semantically meaningful though ! Continue to strongly oppose "datasets" though as that already corresponds to Planning Data _source_, not our curated constraint item which rely on one or many source datasets.

PR #3970, opened 2024-11-18
feat: create new Metabase client

zz-hh-aa:
# What does this PR do?
- Creates Metabase client
    - Checks Metabase connection
    - Axios makes request to Metabase, and validates environment variables in `createMetabaseClient`
    - Sets up custom MetabaseError class
    - Retries for 5xx errors
- Creates an input type for the Metabase client
    - Creates `validateInput()` function to check that client input is the expected shape & data types


api.planx.uk/modules/analytics/metabase/shared/client.ts
```diff
@@ -0,0 +1,155 @@
+import axios from "axios";
+import type {
+  AxiosInstance,
+  AxiosError,
+  AxiosRequestConfig,
```
DafyddLlyr, line 5: GitHub's flagging this as an unused import 👇

api.planx.uk/modules/analytics/metabase/shared/client.ts
```diff
@@ -0,0 +1,155 @@
...
+  InternalAxiosRequestConfig,
+} from "axios";
+import dotenv from "dotenv";
+
+dotenv.config();
```
DafyddLlyr, line 10: I think this is a legacy from the script days of this class - this is already handled in the API. You should be able to remove this and still access `process.env` values.

api.planx.uk/modules/analytics/metabase/shared/client.ts
```diff
@@ -0,0 +1,155 @@
...
+
+// Custom error class for Metabase-specific errors
```
DafyddLlyr, line 12: nit: Redundant comment, we can pick this up from a nice and clear class name and the `extend` keyword below 👍

api.planx.uk/modules/analytics/metabase/shared/client.ts
```diff
@@ -0,0 +1,155 @@
...
+  baseURL: string;
+  apiKey: string;
+  timeout?: number;
+  retries?: number;
+}
+
+// Validate environment variables
+const validateConfig = (): MetabaseConfig => {
+  const baseURL = process.env.METABASE_BASE_URL;
+  const apiKey = process.env.METABASE_API_KEY;
```
DafyddLlyr, line 34: We'll need to add both of these env vars to the codebase. This can be done as part of this PR, or another one - whichever's best for you. It will need to merge either before or alongside this one though ideally 👍 

There's a guide on this here - https://github.com/theopensystemslab/planx-new/blob/main/doc/how-to/how-to-add-a-secret.md

Happy to talk this through sometime also if required.

api.planx.uk/modules/analytics/metabase/shared/client.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+  if (!apiKey) {
+    throw new Error("METABASE_API_KEY is not defined in environment variables");
+  }
+
+  return {
+    baseURL,
+    apiKey,
+    timeout: parseInt(process.env.METABASE_TIMEOUT ?? "30000"), // 30 second default timeout
+    retries: parseInt(process.env.METABASE_MAX_RETRIES ?? "3"), // 3 default retries
```
DafyddLlyr, line 49:
```suggestion
    timeout: METABASE_TIMEOUT,
    retries: METABASE_MAX_RETRIES,
```

I think these are really constants (`const METABASE_TIMEOUT = 30_000`) as opposed to env vars which can differ across environments.

api.planx.uk/modules/analytics/metabase/shared/client.ts
```diff
@@ -0,0 +1,155 @@
...
+  const apiKey = process.env.METABASE_API_KEY;
+
+  if (!baseURL) {
+    throw new Error(
+      "METABASE_BASE_URL is not defined in environment variables",
+    );
+  }
+  if (!apiKey) {
+    throw new Error("METABASE_API_KEY is not defined in environment variables");
+  }
```
DafyddLlyr, line 43: nit: We generally use a pattern of calling `assert()` for env vars (which also accepts a second message argument) across the code base.

I think it's worth keeping this pattern as I'd really like to do a find and replace sometime and replace it with an approach like this which would be super nice! https://www.youtube.com/watch?v=q1im-hMlKhM 🧙‍♂️

api.planx.uk/modules/analytics/metabase/shared/client.ts
```diff
@@ -0,0 +1,155 @@
...
+  // Request interceptor
+  client.interceptors.request.use(
+    (config) => {
+      // Could add request logging here
+      return config;
+    },
+    (error) => {
+      return Promise.reject(error);
+    },
+  );
```
DafyddLlyr, line 80: Is this actually doing anything currently?

api.planx.uk/modules/analytics/metabase/shared/client.ts
```diff
@@ -0,0 +1,155 @@
...
+      }
+
+      // Initialise retry count if not present
+      if (typeof originalRequest.retryCount === "undefined") {
+        originalRequest.retryCount = 0;
+      }
+
+      // Handle retry logic
+      if (error.response) {
+        // Retry on 5xx errors
```
DafyddLlyr, line 100: This looks good and robust - out of curiosity have you found that you've needed it so far? Have requests been timing out or failing?

api.planx.uk/modules/analytics/metabase/shared/client.ts
```diff
@@ -0,0 +1,155 @@
...
+// Health check function
+export const checkMetabaseConnection = async (): Promise<boolean> => {
+  try {
+    const response = await metabaseClient.get("/api/user/current");
+    return response.status === 200;
+  } catch (error) {
+    console.error("Metabase health check failed:", error);
+    return false;
+  }
+};
```
DafyddLlyr, line 153: This might be redundant - we have health checks set up for Metabase as part of our infra, with good visibility.

I think all we really care about here in the API is if a request fails, which looks to be well handled.

api.planx.uk/modules/analytics/metabase/shared/types.ts
DafyddLlyr, line 1: I'm not yet seeing where this file comes into play - should it be part of a different PR?

It looks like it's handling the validation of an incoming object into the API - when it's required we would be best following the existing `validate()` middleware pattern with Zod I think.

DafyddLlyr (changes requested):
Looking good!

A few minor comments and suggestions here and there 👍 

I think we should also test this client with unit tests in isolation - a lot of the functionality here could be hit using axios mocks for responses and requests. If you search `vi.mocked(axios, true)` in the codebase you'll find some existing examples.

You can run `pnpm test:coverage` in the API to check coverage as you go along.

api.planx.uk/modules/analytics/metabase/shared/client.ts
zz-hh-aa: Good shout, (hopefully!) handled on #3992

api.planx.uk/modules/analytics/metabase/shared/client.ts
zz-hh-aa: No, I think I used it before for debugging but will delete now as it's basically the same variety of redundant code as leftover `console.log`s

api.planx.uk/modules/analytics/metabase/shared/client.ts
zz-hh-aa: Hehe no I have to say that this was a Claude recommendation 😅 I had asked for help making my error handling more robust. Should it be kept? Or, if I haven't had issues with timeouts, is it best to remove?

api.planx.uk/modules/analytics/metabase/shared/types.ts
zz-hh-aa: You're right! I'm mixing up types for specific services and the client. This isn't being used, so will remove.

zz-hh-aa:
Thanks for the detailed review @DafyddLlyr 🌟  I believe I have addressed everything. One more outstanding question:

> I think we should also test this client with unit tests in isolation - a lot of the functionality here could be hit using axios mocks for responses and requests. If you search vi.mocked(axios, true) in the codebase you'll find some existing examples.

Do you think those unit tests should be added to `client.test.ts` as part of this PR?

api.planx.uk/modules/analytics/metabase/shared/client.ts
DafyddLlyr: No I think this looks good to keep! Just checking there wasn't an issue with timeouts or anything we might want to look into a little deeper 👍

DafyddLlyr (commented):
All changes look great thanks, marked all as resolved 👍 

> Do you think those unit tests should be added to client.test.ts as part of this PR?

I think it's worth doing it in this PR - this way we have an addition to the code base associated with all its requisite tests which is a nice way of working.

By purely testing the client itself (e.g. how it behaves without secrets, the retry logic, passing of API key, error handling) we'll end up with client-specific functionality covered here, and then the Metabase-specific functionality can be tested a little more cleanly also.

api.planx.uk/modules/analytics/metabase/shared/client.test.ts
```diff
@@ -1 +1,127 @@
...
+    const _client = createMetabaseClient();
+
+    expect(axiosCreateSpy).toHaveBeenCalledWith({
+      baseURL: process.env.METABASE_URL_EXT,
+      headers: {
+        "X-API-Key": process.env.METABASE_API_KEY,
+        "Content-Type": "application/json",
+      },
+      timeout: 30_000,
+    });
```
DafyddLlyr, line 35: I think we can improve this to make this test more meaningful - currently we're not actually testing the returned client. Technically `createMetabaseClient()` could be calling axios (so this test passes), but then returning something invalid for the client.

We can test the returned client like this, for each configuration we've customised - 

```ts
expect(client.defaults.headers["X-API-Key"]).toBe(process.env.METABASE_API_KEY);
```

This then actually tests we're getting back what we expected 👍

DafyddLlyr (changes requested):
Looking great! Two small comments and then I think we're good to go here 👌

api.planx.uk/modules/analytics/metabase/shared/client.test.ts
```diff
@@ -1 +1,127 @@
...
+  });
+
+  describe("Error handling", () => {
+    test("retries then succeeds on 5xx errors", async () => {
+      const baseURL = process.env.METABASE_URL_EXT;
+      if (!baseURL) {
+        throw new Error("METABASE_URL_EXT must be defined for tests");
+      }
+
+      const metabaseScope = nock(baseURL);
```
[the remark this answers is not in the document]
zz-hh-aa: It's actually in there 'cause I get a type error if I try another method, eg just going with `const metabaseScope = nock(process.env.METABASE_URL_EXT);` 
results in this error, which I haven't figured out a decent way of addressing:
``` Argument of type 'string | undefined' is not assignable to parameter of type 'string | RegExp | Url | URL'.
  Type 'undefined' is not assignable to type 'string | RegExp | Url | URL'.
  ```

File: api.planx.uk/modules/analytics/metabase/shared/client.ts [hunk 12]
```diff
@@ -0,0 +1,129 @@
...
+  return client;
+};
+
+// Export both client and instance with delayed instantiation for test purposes
+export let metabaseClient: AxiosInstance;
+
+export const initializeMetabaseClient = () => {
+  metabaseClient = createMetabaseClient();
+  return metabaseClient;
+};
```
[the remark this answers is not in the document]
zz-hh-aa: Seems I confused myself 😅 removed!

api.planx.uk/modules/analytics/metabase/shared/client.test.ts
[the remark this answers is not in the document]
DafyddLlyr: This works ok for me using `!` ([docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator))

image

Is this working locally but failing on CI? If so that might indicate an issue with the secrets on S3 and not the code itself. Happy to try and solve this on a call tomorrow 👍

api.planx.uk/modules/analytics/metabase/shared/client.test.ts
```diff
@@ -1 +1,117 @@
...
+  });
+
+  afterEach(() => {
+    vi.unstubAllEnvs();
+  });
+
+  test("returns configured client", async () => {
+    const client = createMetabaseClient();
+    expect(client.defaults.headers["X-API-Key"]).toBe(
+      process.env.METABASE_API_KEY,
```
DafyddLlyr, line 24: It's worth testing our other setup vars here also like the baseURL, timeout and content-type 👍

DafyddLlyr (approved):
Perfect! ✅

PR #3971, opened 2024-11-18
wip: feat: collection service & controller

zz-hh-aa:
Everything seems to work now--all of my tests pass and I have also successfully used Insomnia to check and update the PlanX DB (locally) and create collections in Metabase (staging). 

I think the next steps would be to break this up into two PRs with cleaner commit histories. Instinctively I would put everything that interacts with the PlanX DB into one (`getTeamAndMetabaseId` and `updateMetabaseId`) plus tests, and everything that only interacts with Metabase into another (`createCollection`). OR I could put the `controller` in one along with the `routes` update, and leave everything else in a bigger one? Feedback and thoughts very welcome!

# What does this PR do?
## PlanX DB
- Creates a migration adding a new column `metabase_id` to `teams`
- Updates `api` permissions to be able to update `metabase_id` and select others

## Controller and routes
- Creates a new route in PlanX API
- Creates a new controller that runs checkCollections service

## Service
- `checkCollections` is the main service, which runs other relevant functions
    - `getTeamAndMetabaseId` fetches `id` and `metabase_id` from `teams`
    - if there is no `metabase-id`, then `createCollection` runs, creating new collection and returning Metabase collection ID
    - `updateMetabaseId` runs updating `teams.metabase_id`

There are also tests!

# Questions
- Should `getTeamAndMetabaseId` be broken up into `getTeam` and `getMetabaseId`?
- Did I update the `api` permissions correctly 😅


api.planx.uk/modules/analytics/metabase/collection/types.ts
```diff
@@ -0,0 +1,54 @@
...
+  error?: string;
+};
+
+export const checkCollectionsSchema = z.object({
+  query: z.object({
+    name: z.string().min(1, "Name parameter is required"),
+  }),
+});
+
+export type CheckCollectionsRequest = ValidatedRequestHandler<
```
DafyddLlyr, line 15:
```suggestion
export type CheckCollections = ValidatedRequestHandler<
```

nit: These aren't requests, but request handlers - `ValidatedRequestHandler<Req, Res>`

api.planx.uk/modules/analytics/metabase/collection/types.ts
```diff
@@ -0,0 +1,54 @@
...
+  typeof checkCollectionsSchema,
+  ApiResponse<CheckCollectionResponse>
+>;
+
+export interface CheckCollectionResponse {
+  collectionId: number | false;
+}
```
DafyddLlyr, line 22: This is a bit of an unexpected response - I'd expect maybe `number | null` or `number | undefined`;

api.planx.uk/modules/analytics/metabase/collection/types.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export interface NewCollectionParams {
+  /** The name of the new collection */
```
DafyddLlyr, line 25: nit: Redundant

api.planx.uk/modules/analytics/metabase/collection/types.ts
```diff
@@ -0,0 +1,54 @@
...
+export type CheckCollectionsRequest = ValidatedRequestHandler<
+  typeof checkCollectionsSchema,
+  ApiResponse<CheckCollectionResponse>
+>;
+
+export interface CheckCollectionResponse {
+  collectionId: number | false;
+}
+
+export interface NewCollectionParams {
```
DafyddLlyr, line 24: Metbase docs suggest that the params `description`, `namespace` or `authority_level` are not required, and we likely won't be configuring them.

If this is the case we should remove them from the type and not worry about them.

api.planx.uk/modules/analytics/metabase/collection/types.ts
```diff
@@ -0,0 +1,54 @@
...
+  parent_id?: number;
+  namespace?: string;
+  authority_level?: null;
+}
+
+export const newCollectionSchema = z.object({
+  body: z.object({
+    name: z.string(),
+    description: z.string().optional(),
+    parent_id: z.number().optional(),
```
DafyddLlyr, line 40: I think it's safe to assume all new collections will be grouped here by default.

```ts
/** Metbase collection ID for the the "Council" collection **/
const COUNCIL_COLLECTION_ID = 58;

...
parent_id: z.number().default(COUNCIL_COLLECTION_ID)
```

api.planx.uk/modules/analytics/metabase/collection/types.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export interface NewCollectionParams {
+  /** The name of the new collection */
+  name: string;
+  description?: string;
+  /** Optional; if the collection is a child of a parent, specify parent ID here
+   * For council teams, parent collection should be 58
+   */
+  parent_id?: number;
```
DafyddLlyr, line 31: Our API should always accept requests in camelCase - if Metabase required snake_case we should convert when posting the body.

DafyddLlyr (changes requested):
A few minor comment and questions on the code.

I think a stronger pattern here would be to store Metabase collection IDs to the PlanX database. This will mean we don't have to query all Metabase collections and find the team each time we make a request. We can manually populate these values for existing teams.

When we add a new team, we want to automatically create a collection for them. This action should return an ID which we can persist in our database. When a new flow is created within this team, we'll then already have the collection ID to hand to ensure the dashboard is created in the correct collection.

I think this would be both simpler and a bit more robust (less room for failures) 👍

api.planx.uk/modules/analytics/metabase/shared/utils.ts
DafyddLlyr, line 1: For functions like this, I'd suggest importing from lodash (which we already use in the API) instead of rolling our own. It means we're using well tested and efficient utility functions, and there's not much overhead in terms of package size.

api.planx.uk/modules/analytics/metabase/collection/service.ts
```diff
@@ -0,0 +1,90 @@
...
+      );
+      return existingCollectionId;
+    }
+
+    // If no existing collection, create new one
+    const requestBody = toSnakeCase({
+      name,
+      description,
+      parent_id,
+    });
```
DafyddLlyr, line 37: Rather than casting to snake_case, the API should accept snake_case, and then we can cast or transform as necessary for the Metabase request.

Incoming request (camelCase) → Validate → Transform to snake_case → Make request to Metabase

api.planx.uk/modules/analytics/metabase/shared/types.ts
```diff
@@ -0,0 +1,40 @@
+export interface Input {
+  teamName: string;
+  originalDashboardId: number;
+  /** The name of the filter to be updated */
+  filter: string;
+  /** The value of the new filter; updateFilter() only supports strings right now */
+  filterValue: string;
+}
+
+export function validateInput(input: unknown): input is Input {
```
DafyddLlyr, line 10: This looks like it's currently unused - is that right?

In terms of validation, it would be best to follow the existing pattern in the API.

 → Incoming request of type `unknown` → 
 → Passes through existing `validate()` middleware in route → 
 → Using Zod, is checked against a provided schema → 
 → Validated payload is automatically passed along as `res.locals.parsedReq` → 
 → Controller and Service have both type safety and a validated payload

api.planx.uk/modules/analytics/metabase/collection/service.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    // Remove undefined properties
+    Object.keys(requestBody).forEach(
+      (key) => requestBody[key] === undefined && delete requestBody[key],
+    );
```
DafyddLlyr, line 42: Passing through the `validate()` middleware should handle this step already - no additional keys would be present.

api.planx.uk/modules/analytics/metabase/collection/service.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export async function authentication(): Promise<boolean> {
+  try {
+    const response = await client.get("/user/current");
+    return response.status === 200;
+  } catch (error) {
+    console.error("Error testing Metabase connection:", error);
+    return false;
+  }
+}
```
DafyddLlyr, line 15: This also looks unused currently.

I think this should be a middleware added to the route - before proceeding to make additional requests to Metabase, we can first check the API key is valid.

api.planx.uk/modules/analytics/metabase/collection/service.ts
zz-hh-aa: Ended up using `transform` instead of importing the Lodash method.

api.planx.uk/modules/analytics/metabase/collection/service.ts
```diff
@@ -0,0 +1,45 @@
...
+    throw error;
+  }
+}
+
+/**
+ * Retrieves info on a collection from Metabase, use to check a parent. Currently only used in tests but could be useful for other Metabase functionality
+ * @param id
+ * @returns
+ */
+// TODO: is this more suited to be part of the collection.test.ts?
```
jamdelion, line 41: Probably better in its own file `getCollection.ts` or something like that :)

api.planx.uk/modules/analytics/metabase/collection/service.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * First uses name to get teams.id and .metabase_id, if present.
+ * If metabase_id is null, return an object that includes its id. If not, create the collection.
+ * @params `name` is required, but `description` and `parent_id` are optional.
+ * @returns `response.data`, so use dot notation to access `id` or `parent_id`.
+ */
+export async function checkCollections(
+  params: NewCollectionParams,
+): Promise<any> {
```
jamdelion, line 17: Same comment as below - would be nice to type the response rather than have `<any>`

api.planx.uk/modules/analytics/metabase/collection/getTeamAndMetabaseId.ts
```diff
@@ -0,0 +1,43 @@
...
+
+interface GetMetabaseId {
+  teams: {
+    id: number;
+    name: string;
+    metabaseId: number | null;
+  }[];
+}
+
+export const getTeamAndMetabaseId = async (name: string) => {
```
jamdelion, line 12: I think this is fine as one gql request, maybe I'd change the function name to be `getTeamIdAndMetabaseId` just to be completely explicit

api.planx.uk/modules/analytics/metabase/collection/service.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const client = createMetabaseClient();
+
+/**
+ * First uses name to get teams.id and .metabase_id, if present.
+ * If metabase_id is null, return an object that includes its id. If not, create the collection.
+ * @params `name` is required, but `description` and `parent_id` are optional.
+ * @returns `response.data`, so use dot notation to access `id` or `parent_id`.
+ */
+export async function checkCollections(
```
jamdelion, line 15: Maybe I'd call this `createCollectionIfDoesNotExist` - "checking" is a bit of a vague term

jamdelion: Then maybe the controller could just be called `CollectionsController` or even `MetabaseCollectionsController`

jamdelion:
Added a few comments to try and answer some questions! :) I think your intuition about splitting on metabase/planX db lines is correct. 

I haven't tested this locally (think I'm missing the metabase `env` vars) but sounds like it was working from our chat yesterday 🙌 Maybe include some brief testing instructions when you have the PRs split out.

PR #3972, opened 2024-11-18
feat: add `user_clicked_save` to `analytics_summary` view

RODO94:
For the metabase ticket: https://trello.com/c/2j5CC6IG/3124-expose-userclickedsave-in-analyticssummary

Added extra data to the analytics_summary view to support the pilot work

Updated permissions to mirror `has_clicked_help`

I am only using the `SaveAndResumeButton` component to trigger this `trackEvent()`

>[!Note]
> It will not populate `analytics_logs` if it is on the first node, due to `lastVisibleNodeAnalyticsLogId` being `undefined` so it will fail `if (shouldSkipTracking()) return;` in `trackEvent()`. In normal cases, I'd imagine we'd have some form of intro node or something similar and this is the current behaviour for all cases so I wasn't going to begin pulling on threads here.


DafyddLlyr (changes requested):
I think there's a bit of a misunderstanding here about the underlying data model - specifically what the `has_user_saved` column indicates, and how the `lowcal_sessions` and `analytics` tables relate to each other.

**On relationships**
The analytics tables do not have a relationship with sessions - we collect anonymised analytics which cannot be linked to user-specific sessions.

When the join `LEFT JOIN lowcal_sessions ls ON ((ls.flow_id = a.flow_id))` is added to the view, this is joining analytics → flows → sessions, meaning there will be many sessions to one flow. We're not joining on session ID, and intentionally not able to do so.

If we want to access session data, we can do this using the `submission_services_summary` view.

**On the column**
`has_user_saved` is used to track if email reminder and expiry events have been set up for a session, and is a boolean value. There's a comment on the column explaining this, and a few references in code which should demonstrate how it's used. This does not tell us anything about where in the journey this happened (which could be none, one, or many places per session) just a binary true/false. 

So we're going to need a different approach! 🤔 

Here's how I'd read this request from Trello - 
 - We want to know when/where something happens (a user saves their session)
 - We want this exposed via Metabase so we can contextualise and understand it
 
We have an existing pattern for this - analytics events. I'd explore adding a new event on save, and then making sure Metabase has access to this. This meets the criteria laid out in this user story - 

> As a PlanX analyst
> I want user_clicked_save and node_title to exist in the same table
> so that I can measure where users most often click save in submission

cc @zz-hh-aa for visibility here also 👋

editor.planx.uk/src/pages/FlowEditor/lib/analytics/types.ts
```diff
@@ -32,6 +32,8 @@ export type AllowListKey = (typeof ALLOW_LIST)[number];
  */
 export type HelpClickMetadata = Record<string, string>;
 
+export type SaveClickMetadata = Record<string, string>;
```
DafyddLlyr, line 35: This seems redundant and unused. There's no metadata to gather - it's just a true/false.

Other analytics event gather metadata in order to store additional data, but this is just a straightforward boolean.

editor.planx.uk/src/pages/FlowEditor/lib/analytics/types.ts
```diff
@@ -108,6 +111,11 @@ type HelpClick = {
   metadata: HelpClickMetadata;
 };
 
+type SaveClick = {
```
DafyddLlyr, line 114: The other events have a line or two in jsdoc explaining their use. I appreciate that this seems simple and straightforward, but there's a decent level of complexity and context here so it's worth maintaining this pattern.

hasura.planx.uk/migrations/1732016907414_alter_table_public_analytics_logs_add_column_has_user_clicked_save/up.sql
```diff
@@ -0,0 +1,61 @@
+alter table "public"."analytics_logs" add column "has_user_clicked_save" boolean
```
DafyddLlyr, line 1: nit: I think just `has_clicked_save` is good here - 
 - It matches the current column names (e.g. `has_clicked_help`)
 - This is a table tracking user interactions so it should be implicit

hasura.planx.uk/migrations/1732016907414_alter_table_public_analytics_logs_add_column_has_user_clicked_save/up.sql
```diff
@@ -0,0 +1,61 @@
...
+    ((al.allow_list_answers -> 'usedFOIYNPP'::text))::text AS used_foiynpp,
+    ((al.allow_list_answers -> 'propertyInformation.action'::text))::text AS property_information_action,
+    ((al.allow_list_answers -> 'planningConstraints.action'::text))::text AS planning_constraints_action,
+    ((al.allow_list_answers -> '_overrides'::text))::text AS overrides,
+    ((al.allow_list_answers -> 'rab.exitReason'::text))::text AS rab_exit_reason,
+    ((al.allow_list_answers -> 'service.type'::text))::text AS pre_app_service_type,
+    ((al.allow_list_answers -> 'application.information.harmful'::text))::text AS pre_app_harmful_info,
+    ((al.allow_list_answers -> 'application.information.sensitive'::text))::text AS pre_app_sensitive_info,
+    (((al.allow_list_answers -> 'application.type'::text) -> 0))::text AS application_type,
+    al.has_user_clicked_save
```
DafyddLlyr, line 55: nit: I think in terms of readability and visibility, this is better grouped with `al.has_clicked_help` and other non-allow-list columns than just appended on the end.

DafyddLlyr (changes requested):
Quick turnaround - great to see!

A few small comments regarding naming and the metadata to take a look at 👍

hasura.planx.uk/migrations/1732016907414_alter_table_public_analytics_logs_add_column_has_user_clicked_save/up.sql
RODO94: I run into an SQL error here about altering the view, which I can get around if I remove and replace the view rather than `CREATE or REPLACE` but it seemed overly complicated for a data value like this.

Was this just a Hasura SQL based error though and you think it won't be a problem here in the migration?

RODO94: error:

```sql
"message": "cannot change name of view column \"allow_list_answers\" to \"has_user_clicked_save\"", "status_code": "42P16" },
```

DafyddLlyr: I think it's safe to assume any errors hit locally via Hasura will persist across environments.

No worries if this is the case! You would have to drop, recreate, and then sort out permissions - not worth it for a small nitpicky suggestion! 😅

editor.planx.uk/src/pages/FlowEditor/lib/analytics/types.ts
```diff
@@ -108,6 +109,16 @@ type HelpClick = {
...
 };
 
+/**
+ * Captured when a user clicks Save and Return.
+ * The mutation sets the "has_clicked_help: true" when Save and
+ * Return button is clicked
+ */
+
+type SaveClick = {
+  event: "saveClick";
```
DafyddLlyr, line 119: I think it's worth keeping `metadata: null` here to match other `EventData` interfaces :+1

DafyddLlyr (approved):

PR #3973, opened 2024-11-18
feat: add `applicant.researchOptIn` for metabase

RODO94:
This PR adds an extra field to the `ALLOW_LIST` to track what users "Opt In" to research.


hasura.planx.uk/migrations/1731944794522_add_research_opt_in_to_allow_list/up.sql
```diff
@@ -0,0 +1,160 @@
...
+    ((al.allow_list_answers -> 'usedFOIYNPP'::text))::text AS used_foiynpp,
+    ((al.allow_list_answers -> 'propertyInformation.action'::text))::text AS property_information_action,
+    ((al.allow_list_answers -> 'planningConstraints.action'::text))::text AS planning_constraints_action,
+    ((al.allow_list_answers -> '_overrides'::text))::text AS overrides,
+    ((al.allow_list_answers -> 'rab.exitReason'::text))::text AS rab_exit_reason,
+    ((al.allow_list_answers -> 'service.type'::text))::text AS pre_app_service_type,
+    ((al.allow_list_answers -> 'application.information.harmful'::text))::text AS pre_app_harmful_info,
+    ((al.allow_list_answers -> 'application.information.sensitive'::text))::text AS pre_app_sensitive_info,
+    (((al.allow_list_answers -> 'application.type'::text) -> 0))::text AS application_type,
+    (((al.allow_list_answers -> 'applicant.researchOptIn'::text) -> 0))::text AS applicant_research_opt_in
```
DafyddLlyr, line 52:
```suggestion
    al.allow_list_answers -> 'applicant.researchOptIn' ->> 0 AS applicant_research_opt_in
```

I think this is the same thing, just simpler? The value stored is the following as a JSONB `["true"]` or `["false"]`.

Does the above work or is there something I'm missing?

DafyddLlyr (commented):
One small question!

hasura.planx.uk/migrations/1731944794522_add_research_opt_in_to_allow_list/up.sql
RODO94: Very good question, I was wondering the same while looking at the `application.type` - I remember I had a bit of trouble there, that's why I repeated the pattern - but I agree it's a funny wee line

I'll fire up the change and see what happens

RODO94: @DafyddLlyr looks good with the change and I've extended it to `application.type`

DafyddLlyr: Yep the subtle difference is the `->` which just accesses and leaves as JSONB and `->>` which converts to type text.

Docs: https://www.postgresql.org/docs/current/functions-json.html

PR #3975, opened 2024-11-18
feat: add tags to portals

RODO94:
Following previous templates groundwork, this PR adds the ability to add tags to portals

### DIFF
--- a/editor.planx.uk/src/@planx/components/ExternalPortal/Editor.test.tsx
+++ b/editor.planx.uk/src/@planx/components/ExternalPortal/Editor.test.tsx
@@ -31,6 +31,7 @@ test("adding an external portal", async () => {
       type: TYPES.ExternalPortal,
       data: {
         flowId: "b",
+        tags: [],
       },
     }),
   );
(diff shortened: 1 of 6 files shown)


DafyddLlyr (approved):
Great!

DafyddLlyr (changes requested):
Sorry just quickly checked the Pizza!

It would be good to sort the layout here - 
image

RODO94:
> Sorry just quickly checked the Pizza!
> 
> It would be good to sort the layout here - image
> 
> I know these components aren't as well structured as others, but I think even just sorting out the margin here would be enough.

Yes, good catch! I did stash these, I'll put up a fix now.

DafyddLlyr (commented):
Code changes and visual output looking good! I think there's just some tests that need to be updated here and then we're good to go.

RODO94:
> Code changes and visual output looking good! I think there's just some tests that need to be updated here and then we're good to go.

Yeah, I did wonder why I stashed the UI changes... all good now

DafyddLlyr (approved):
Perfect - thanks for those changes!

PR #3977, opened 2024-11-19
chore: retire custom ESRI GIS integrations

jessicamcinchak:
In the past before Planning Data, we integrated directly with local ESRI GIS servers - notably for Braintree council and a demo for the Scottish government. Waiting for OK here: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1731954174803899

We no longer anticipate supporting this integration approach, so I'm fully removing these files as they add a fair amount of complexity to how we handle planning constraints. If we ever need to spin up demos again, the ESRI helpers especially should be find-able & easy to re-implement/revert :slightly_smiling_face:


editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
```diff
@@ -69,32 +69,22 @@ function Component(props: Props) {
...
-    version: "1",
-  };
 
   // Fetch planning constraints data for a given local authority
   const root = `${import.meta.env.VITE_APP_API_URL}/gis/${teamSlug}?`;
   const teamGisEndpoint: string =
     root +
     new URLSearchParams(
-      hasPlanningData ? digitalLandParams : customGisParams,
+      hasPlanningData ? planningDataParams : undefined,
```
DafyddLlyr, line 87: Are we relying on this failing because `geom` will be undefined when the request is made?

 If `!hasPlanningData` should be guard against making the request full stop?

api.planx.uk/modules/gis/service/index.js
```diff
@@ -61,37 +42,11 @@ export async function locationSearch(req, res, next) {
   }
 }
 
-// Digital Land is a single request with standardized geometry, so remove timeout & simplify query params
+// Planning Data is a single request with standardized geometry, so timeout is not necessary
 export function locationSearchWithoutTimeout(localAuthority, queryParams) {
```
DafyddLlyr, line 46: nit: Is it simpler to also drop the `withoutTimeout` suffix and comment as it no longer contextually relevant unless contrasted to `locationSearchWithTimeout()`

api.planx.uk/modules/gis/service/index.js
```diff
@@ -22,8 +20,7 @@ const localAuthorities = { digitalLand, braintree, scotland };
  *        description: Well-Known Text (WKT) formatted polygon or point
  */
 export async function locationSearch(req, res, next) {
-  // 'geom' param signals this localAuthority has data available via DigitalLand, "ready" teams are configured in PlanningConstraints component in editor
-  //   swagger doc intentionally doesn't cover legacy custom GIS hookups for Braintree and Scotland demo
+  // 'geom' param signals this localAuthority has data available via Planning Data
```
DafyddLlyr, line 23: Aha this answers my previous question - maybe a check against `hasPlanningData` in both places is a bit more robust / expected.

api.planx.uk/modules/gis/service/helpers.ts
DafyddLlyr, line 1: Thanks for the bonus `.js` → `.ts` here as well ✨

DafyddLlyr (approved):
Two small comments / questions, totally non-blocking.

Some satisfying deletions here - thanks!

DafyddLlyr:
A small follow up task here is to remove legacy custom GIS endpoints from UptimeRobot also.

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.tsx
jessicamcinchak: Already need to revisit these request conditions in #3968 (eg only make "roads" request _if_ the editor has specified they want to check roads and they have a USRN) - so going to take these comments forward into that PR is ok!

api.planx.uk/modules/gis/service/index.js
jessicamcinchak: Updated now!

PR #3981, opened 2024-11-19
chore: Replace `jest` with `vitest` for Hasura introspection tests

DafyddLlyr:
## What does this this PR do?
 - Removes jest, and replaces it with vitest

## What's the motivation?
 - Matches the rest of the project
 - Fixes https://github.com/theopensystemslab/planx-new/security/dependabot/143
 - Actually a super easy and quick change to make! This also caught [1f925ce](https://github.com/theopensystemslab/planx-new/pull/3981/commits/1f925ce279fd466eb7c47043d901e59f6074c9ce) which is nice. There's a small follow up task here in future to finally convert to TS here.


jamdelion (approved):
Nice one! 👍

PR #3984, opened 2024-11-19
fix: add default disclaimer text to shared component

jamdelion:
As discussed in [this slack thread](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1732029167504329), the disclaimer component got slightly broken in [my refactor ](https://github.com/theopensystemslab/planx-new/pull/3900). Namely, I extracted out the `Disclaimer` component from `Planning Constraints` but accidentally removed the fallback string. 🤦 

This has been re-added as an optional prop `defaultText` in the shared `Disclaimer` component.

### DIFF
--- a/editor.planx.uk/src/@planx/components/PlanningConstraints/Presentational.tsx
+++ b/editor.planx.uk/src/@planx/components/PlanningConstraints/Presentational.tsx
@@ -10,6 +10,7 @@ import { Disclaimer } from "../shared/Disclaimer";
 import { ErrorSummaryContainer } from "../shared/Preview/ErrorSummaryContainer";
 import SimpleExpand from "../shared/Preview/SimpleExpand";
 import ConstraintsList from "./List";
+import { DEFAULT_PLANNING_CONDITIONS_DISCLAIMER } from "./model";
 import { InaccurateConstraints } from "./Public";
 
 export type PresentationalProps = {
@@ -80,7 +81,10 @@ export function Presentational(props: PresentationalProps) {
               />
             </SimpleExpand>
           )}
-          <Disclaimer text={disclaimer} />
+          <Disclaimer
+            text={disclaimer}
+            defaultText={DEFAULT_PLANNING_CONDITIONS_DISCLAIMER}
+          />
         </>
       )}
       {positiveConstraints.length === 0 && negativeConstraints.length > 0 && (
@@ -110,7 +114,10 @@ export function Presentational(props: PresentationalProps) {
               setInaccurateConstraints={setInaccurateConstraints}
             />
           </SimpleExpand>
-          <Disclaimer text={disclaimer} />
+          <Disclaimer
+            text={disclaimer}
+            defaultText={DEFAULT_PLANNING_CONDITIONS_DISCLAIMER}
+          />
         </>
       )}
     </Card>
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/PlanningConstraints/Presentational.tsx
```diff
@@ -80,7 +81,10 @@ export function Presentational(props: PresentationalProps) {
               />
             </SimpleExpand>
           )}
-          <Disclaimer text={disclaimer} />
+          <Disclaimer
+            text={disclaimer}
+            defaultText={DEFAULT_PLANNING_CONDITIONS_DISCLAIMER}
+          />
```
jessicamcinchak, line 87: nit: what's your reasoning for a separate prop rather than simply passing `text={disclaimer || DEFAULT_PLANNING_CONDITIONS_DISCLAIMER}` to the existing prop?

Would that be a more accurate way to ensure the required "text" prop on the `Disclaimer` component is actually always populated?

editor.planx.uk/src/@planx/components/PlanningConstraints/Public.test.tsx
```diff
@@ -239,4 +239,21 @@ describe("following a FindProperty component", () => {
...
+
+  test("default disclaimer text should render if none provided", async () => {
+    const { queryByText } = setup(
+      <PlanningConstraints
+        title="Planning constraints"
+        description="Things that might affect your project"
+        fn="property.constraints.planning"
+        disclaimer=""
+        handleSubmit={vi.fn()}
+      />,
```
jessicamcinchak, line 251: nit: in August's case (and if you create a new Planning Constraints in the editor & remove the disclaimer text), we don't have `disclaimer=""` but rather the `disclaimer` prop isn't present at all ! 

That might be the more realistic case to test here?

editor.planx.uk/src/@planx/components/shared/Disclaimer.tsx
```diff
@@ -4,11 +4,24 @@ import { WarningContainer } from "@planx/components/shared/Preview/WarningContai
...
-      <ReactMarkdownOrHtml source={text} openLinksOnNewTab />
-    </Typography>
-  </WarningContainer>
-);
+export const Disclaimer = ({
+  text,
+  defaultText,
+}: {
+  text: string;
+  defaultText?: string;
```
jessicamcinchak, line 12: nit: see first comment - I'd personally favor a single `text` prop here that's always required & handle the `||` fallback condition at point of implementation!

jessicamcinchak (commented):
Realise I missed first review here, but a couple clarifing comments/questions to consider fixing-forward (no pressure) !

editor.planx.uk/src/@planx/components/PlanningConstraints/Presentational.tsx
jamdelion: Thanks @jessicamcinchak :) I did consider this but the since the disclaimer component is shared across a few places (for example, it appears in the Feedback component), it would be weird to have the `defaultText` related to planning conditions.
But you're right, since this is an optional prop, the problem that August encountered could still happen again if someone forgets to account for the case when `text` is `undefined`. 🤔 Perhaps it could be a non-optional prop, or I could come up with a more generic fallback to populate by default in the shared component?

jessicamcinchak: Not suggesting that the planning conditions default text is used as the _default_ for the shared component, just that the shared component _only_ needs a single required `text` prop as originally implemented and that this could have been a one-line fix here (in `Presentational.tsx`) I think by passing `text={disclaimer || DEFAULT_PLANNING_CONDITIONS_DISCLAIMER}` (aka _not_ impacting other uses of shared `Disclaimer`)!

Sorry first comment didn't articulate that so clearly

jamdelion: Ah sorry I misunderstood, that makes perfect sense!

PR #3986, opened 2024-11-20
fix: Only scroll to start node on initial load of flow

DafyddLlyr:
I had something to this effect in place on a previous PR, but seem to have lost it in a merge/rebase somewhere.

Currently (on staging), the start node scrolls to the center top on each load - it should only happen on initial load.

**Staging - incorrect behaviour**
_Scroll position reset on each route change_

https://github.com/user-attachments/assets/6b348956-f378-4b98-9187-dc92d47ac33a

**This PR - correct behaviour**
_Scroll position retained_

https://github.com/user-attachments/assets/cf789540-187f-469b-a3b1-1ceb121b3e56


RODO94 (approved):
Looks good to me

PR #3992, opened 2024-11-20
feat: setup Metabase env variables

zz-hh-aa:
Following the ['How to add a secret'](https://github.com/theopensystemslab/planx-new/blob/main/doc/how-to/how-to-add-a-secret.md) guide, this PR
- Adds Metabase staging & production keys to Pulumi files in `infrastructure/application`
- Adds vars to `api.planx.uk/.env.test.example`

Came out of feedback on #3970


api.planx.uk/.env.test.example
```diff
@@ -45,3 +45,7 @@ ORDNANCE_SURVEY_API_KEY=👻
 
 IDOX_NEXUS_TOKEN_URL=👻
 IDOX_NEXUS_SUBMISSION_URL=👻
+
+# Analytics
+METABASE_API_KEY=👻
```
DafyddLlyr, line 50: We also need to add this to `.env.example`.

`.env.example` is a public list of the required secrets needed to run PlanX, and serves as documentation.

`.env.test.example` is a per-project subset of the full env var list, which only includes variables required to run unit tests.

docker-compose.yml
```diff
@@ -145,6 +145,8 @@ services:
       IDOX_NEXUS_TOKEN_URL: ${IDOX_NEXUS_TOKEN_URL}
       JWT_SECRET: ${JWT_SECRET}
       MAPBOX_ACCESS_TOKEN: ${MAPBOX_ACCESS_TOKEN}
+      METABASE_API_KEY: ${METABASE_API_KEY}
+      METABASE_URL_EXT: ${METABASE_URL_EXT}
```
DafyddLlyr, line 149: We also need to pass this along to the Fargate API instance here - https://github.com/theopensystemslab/planx-new/blob/main/infrastructure/application/index.ts#L331-L456

DafyddLlyr (changes requested):
A few additional steps required! I've also improved the documentation a little here which hopefully helps to explain some of this setup 🙂  

https://github.com/theopensystemslab/planx-new/pull/3994/files?short_path=0285396#diff-02853966ac252e3f0d2457668dde4f04312e4e501fa8a05d8d4ac18256364f17

api.planx.uk/.env.test.example
zz-hh-aa: I think they're [in there already](https://github.com/theopensystemslab/planx-new/blob/06e01756676637d0fbb433a15c859d17454b5e5a/.env.example#L54C1-L57C51) from earlier PR #3815

docker-compose.yml
zz-hh-aa: Pushed a new commit with this, hope I have done / understood correctly?

api.planx.uk/.env.test.example
DafyddLlyr: Ah perfect - thanks!

PR #3998, opened 2024-11-21
refactor: added Int Portal restyling to Ext Portal

RODO94:
Following the thread on proposed changes and feedback, the approach for this PR is to add in reasonable changes to the styling and slight function, but not go too far into complete rework on the experience of the Modal.

This means we did not add a new way for people to filter flows by Team via a separate list of unique teams, but it maintains current functionality with adding a better Select component in the form of Material UI `Autocomplete`. 

Considering this is another use of the component, I have added refactoring to centralise the render components for the Autocomplete and extend the use of the `<Autocomplete/>` used for the Datafields work.


editor.planx.uk/src/@planx/components/ExternalPortal/Editor.tsx
```diff
@@ -51,20 +155,78 @@ const ExternalPortalForm: React.FC<{
...
+            fullWidth
+            popupIcon={PopupIcon}
+            ListboxProps={{
+              sx: (theme) => ({
+                paddingY: 0,
+                backgroundColor: theme.palette.background.default,
+              }),
+            }}
+            value={
+              flows.find((flow: Flow) => flow.id === formik.values.flowId) ||
```
RODO94, line 200: Without cascading more changes, I had to pull in the id and use that to define the value of each item, so I could relate it to the team. From what I could gauge, we need to ensure the id is assigned to formik so it can be passed in for existing portals.

Happy to consider a bigger change here, but my thinking was keeping the changes isolated to this file to maintain existing functionality

editor.planx.uk/src/@planx/components/ExternalPortal/Editor.test.tsx
```diff
@@ -9,22 +9,31 @@ import ExternalPortalForm from "./Editor";
...
+        { id: "b", name: "flow b", slug: "flow-b", team: "team" },
       ]}
       handleSubmit={handleSubmit}
     />,
   );
+  const autocompleteComp = screen.getByTestId("flowId");
+  const autocompleteInput = within(autocompleteComp).getByRole("combobox");
 
-  expect(screen.getByTestId("flowId")).toHaveValue("");
+  screen.debug(autocompleteInput);
```
jamdelion, line 24: nit: screen.debug left in 🧼 🫧

editor.planx.uk/src/@planx/components/ExternalPortal/Editor.test.tsx
```diff
@@ -41,24 +50,31 @@ test("adding an external portal", async () => {
...
-        { id: "b", text: "flow b" },
+        { id: "a", name: "flow a", slug: "flow-a", team: "team" },
+        { id: "b", name: "flow b", slug: "flow-b", team: "team" },
       ]}
       handleSubmit={handleSubmit}
     />,
   );
 
-  expect(screen.getByTestId("flowId")).toHaveValue("b");
+  const autocompleteComp = screen.getByTestId("flowId");
```
jamdelion, line 64: what does 'comp' mean?

editor.planx.uk/src/@planx/components/ExternalPortal/Editor.tsx
jamdelion, line 1: nit: Rather than growing this file with more sub-components and type definitions, could you consider splitting out into a separate `types.ts` file and `components` sub-folder? I think it makes the files much easier to read when they're just dealing with one component at a time.

editor.planx.uk/src/@planx/components/ExternalPortal/Editor.tsx
```diff
@@ -51,20 +155,78 @@ const ExternalPortalForm: React.FC<{
...
+              flows[0]
+            }
+            onChange={(_event, newValue: Flow) => {
+              formik.setFieldValue("flowId", newValue.id);
+            }}
+            options={
+              flows &&
+              flows.filter((flow) => {
```
jamdelion, line 208: You can write this more succinctly like this: 

```suggestion
              flows?.filter((flow) => {
```

editor.planx.uk/src/@planx/components/ExternalPortal/Editor.tsx
```diff
@@ -1,31 +1,135 @@
...
+  SelectMultiple,
+  StyledTextField,
+} from "ui/shared/SelectMultiple";
 
 import { ICONS } from "../shared/icons";
 
 interface Flow {
-  id: string;
-  text: string;
+  id: string | "select a flow";
```
jamdelion, line 29: it seems a bit weird to me that `Select a flow` could be the `id` of a flow. Perhaps I'd expect to see more of a mapping where like `{id: "xyz"}` is mapped to a fallback string ("select a flow") or something?

editor.planx.uk/src/@planx/components/ExternalPortal/Editor.tsx
```diff
@@ -51,20 +155,78 @@ const ExternalPortalForm: React.FC<{
             flow that it references.
           </span>
         </ModalSectionContent>
-        <ModalSectionContent title="Pick a flow">
-          <select
+        <ModalSectionContent
+          key={"team-section"}
+          title="Select a team (optional)"
+        >
+          <SelectMultiple
```
DafyddLlyr, line 162: There's an odd interaction here - if I have a flow selected, and then change the team the fist flow of this team gets selected which is quite unexpected.

editor.planx.uk/src/@planx/components/ExternalPortal/Editor.tsx
```diff
@@ -1,31 +1,135 @@
...
+      notched: false,
+    }}
+    key={params.id}
+  />
+);
+
+const renderGroup: FlowAutocompleteListProps["renderGroup"] = (params) => {
+  return (
+    <React.Fragment key={params.group}>
+      <AutocompleteSubHeader
```
DafyddLlyr, line 94: This looks quite different to other grouping we have - see `FileUploadAndLabel`

DafyddLlyr (changes requested):
A few comments and questions on the PR below, but a more general piece of feedback at this stage.

This one has turned out more complicated (and time consuming) than expected so I'd think that the best course of action here is to implement style changes, so that it's visually consistent with other Editor modals and not focus on the two selects and their interactions. I did think that this was the conclusion we got to in the Slack thread the other day and was expecting to see those changes implemented here - apologies if that wasn't clear though 👌 

So we'd end up with - 
 - Visually consistent and updated modal
 - A single select - this can certainly be grouped by team as you have here (if George/Jonny/August think this is an improvement) - I think it would be as we have a typeahead implemented
 - Trello ticket to gather Editor feedback / implement multiple selects with reference to these PRs so we can revisit this in future without losing design and dev work
 

Happy to talk this through if you'd like - just let me know.

editor.planx.uk/src/@planx/components/ExternalPortal/Editor.tsx
```diff
@@ -1,31 +1,135 @@
...
+    id={params.id}
+    InputProps={{
+      ...params.InputProps,
+      notched: false,
+    }}
+    key={params.id}
+  />
+);
+
+const renderGroup: FlowAutocompleteListProps["renderGroup"] = (params) => {
```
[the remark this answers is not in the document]
RODO94: Only thing stopping this is how we handle the `freeSolo` prop in the Autocomplete type. I've done an initial bit of work in standardising some components, types, and styles and locating them somewhere shared. I'd appreciate feedback on how deep to go here, to me it does feel like another PR to merge all of our Autocomplete uses and the nuances.

editor.planx.uk/src/ui/shared/Autocomplete/styles.ts
```diff
@@ -0,0 +1,86 @@
...
+      paddingY: 0,
+      transform: "translate(0px, -22px) scale(0.85)",
+    },
+  },
+}));
+
+export const StyledDataField = styled(StyledTextField)(({ theme }) => ({
+  [`& .${outlinedInputClasses.root}, input`]: {
+    cursor: "pointer",
+    // TODO extract as `format="data"` prop more like `Input` ?
```
jamdelion, line 57: Is this todo overlooked for this PR or a future thing you were going to come back to?

jamdelion:
Screenshot 2025-02-03 at 15 10 40

Just looked at the pizza again and functionality looks good. One thing I was thinking though is that the initial state of this input as a blank box is a bit disconcerting to me. Maybe I'd expect to see some sort of dropdown arrow in it? Or a placeholder? I feel like it needs some visual indication of what interaction is expected with it maybe...

RODO94:
> Screenshot 2025-02-03 at 15 10 40
> 
> Just looked at the pizza again and functionality looks good. One thing I was thinking though is that the initial state of this input as a blank box is a bit disconcerting to me. Maybe I'd expect to see some sort of dropdown arrow in it? Or a placeholder? I feel like it needs some visual indication of what interaction is expected with it maybe...

@jamdelion do you think a drop down arrow and maybe a more descriptive label would work? 

"Pick a flow or type to search"

jamdelion (approved):
Looking good to me!

DafyddLlyr (approved):

PR #3999, opened 2024-11-21
fix: Restore `EditorNavMenu` for platformAdmins

DafyddLlyr:
Please see https://opensystemslab.slack.com/archives/C01E3AC0C03/p1732200694468619 (OSL Slack)


jamdelion (approved):
Wondering if it's worth having a test to cover this, so that we expect that the editor nav menu is visible for platformAdmins?

DafyddLlyr:
> Wondering if it's worth having a test to cover this, so that we expect that the editor nav menu is visible for platformAdmins?

Yep very good shout - turns out this entire component doesn't have coverage. Added this in [5eadf02](https://github.com/theopensystemslab/planx-new/pull/3999/commits/5eadf0245bb4d3a4234194156c4a74f2f89c9808) 👍

PR #4000, opened 2024-11-21
feat: verify email before downloading application

jamdelion:


editor.planx.uk/src/routes/index.tsx
```diff
@@ -30,6 +31,13 @@ const editorRoutes = mount({
         }),
   ),
 
+  "/:team/:flow/:sessionId/verify-email": map((req) => {
+    return route({
+      title: makeTitle("Verify your email"),
+      view: <VerifyEmail params={req.params} />,
+    });
+  }),
```
jessicamcinchak, line 39: nit: for consistency, I wonder if this route name could be `/download-application` (then download link remains `/download-application-files`) ?

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -507,11 +506,6 @@ export const editorStore: StateCreator<
   async publishFlow(flowId: string, summary?: string) {
     const token = get().jwt;
 
-    const urlWithParams = (url: string, params: any) =>
```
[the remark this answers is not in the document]
jamdelion: Oops, I was using it in the new feature which is why I extracted it, but I must have stopped using it so I will revert the extraction

editor.planx.uk/src/pages/VerifyEmail/VerifyEmail.tsx
```diff
@@ -0,0 +1,127 @@
...
+        <DelayedLoadingIndicator />
+      ) : (
+        <Box width="100%">
+          <Card handleSubmit={formik.handleSubmit}>
+            <ErrorWrapper error={downloadApplicationError}>
+              <>
+                <CardHeader
+                  title="Verify your email address"
+                  description="We will use this to verify that you can download your application. Entering the correct email address will start the file download automatically."
+                />
```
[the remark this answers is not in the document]
jamdelion: Good shout, the wording is leftover from when I misunderstood the user story!

With the summary list from `InviteToPay`, do you mean the form that pops up on the Pay page?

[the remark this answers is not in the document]
jessicamcinchak: Yes ! "Summary list" is the Gov UK pattern language for those mini-forms/tables https://design-system.service.gov.uk/components/summary-list/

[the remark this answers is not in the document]
jamdelion: I've added a small summary table with the data we have easily available in this component - `sessionId` and `localAuthority`. I think getting other data will involve another API call, which I'll leave to a future PR if we think its a worthwhile enhancement ✨

editor.planx.uk/package.json
```diff
@@ -16,6 +16,7 @@
     "@mui/utils": "^5.15.11",
     "@opensystemslab/map": "1.0.0-alpha.4",
     "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#ccf9ac3",
+    "@tanstack/react-query": "^5.61.3",
```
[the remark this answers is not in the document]
jamdelion: As discussed in dev call, have got rid of react-query for now (and this work in the previous state saved off in another branch [jh/download-application-with-tanstack-react-query](https://github.com/theopensystemslab/planx-new/tree/jh/download-application-with-tanstack-react-query)) until we have figured out if it will work for us on a longer term basis!

editor.planx.uk/src/pages/SubmissionDownload/VerifySubmissionEmail.tsx
```diff
@@ -0,0 +1,121 @@
...
+                        formik.touched.email && formik.errors.email
+                          ? formik.errors.email
+                          : undefined
+                      }
+                      id="email"
+                      name="email"
+                      onBlur={formik.handleBlur}
+                      onChange={formik.handleChange}
+                      type="email"
+                      value={formik.values.email}
```
jessicamcinchak, line 96: nit: I noticed that this input is _not_ triggering an error when I press "enter" from within it before typing anything or when I just type "jess" (invalid email format) and press enter - which does not match our typical public-facing email input error handling !

jessicamcinchak (commented):
Thanks for updating to use `axios` here ! A few comments on re-review, overall all working nearly as expected ! 

Before a final review, I think this one is a particularly good candidate to post to wider team while still in pizza-state for feedback if not already on your radar. 

A few bigger-picture questions I have that would influence how/when it's merged: 
- Does the "Download" page look PlanX-y enough / will council's see it as a trustworthy page to download an application from? 
  - Do we need to a more-PlanX/Gov UK feeling header or something? I need to remember how Invite to Pay handles this? 
- Will this new workflow _break_ any existing integrations for councils using Send to Email? 
  - Eg I have a hunch that Doncaster & Camden's internal IT teams are perhaps running scripts to auto-download email links from PlanX submission emails? 
  - What's going to be the process for giving council's a headsup / trying to learn more about who might need to adjust things internally and helping them prepare for that rather than getting bug reports later?

editor.planx.uk/src/pages/SubmissionDownload/helpers/downloadZip.tsx
```diff
@@ -0,0 +1,19 @@
...
+  if (!data) {
+    console.error("No data to download");
+    return;
+  }
+  const blobData = new Blob([data], { type: "application/zip" });
+  try {
+    const href = URL.createObjectURL(blobData);
+    const link = document.createElement("a");
+    link.href = href;
+    link.setAttribute("download", "application.zip");
```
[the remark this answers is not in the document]
jamdelion: Oooh good catch, thanks!

editor.planx.uk/src/pages/SubmissionDownload/VerifySubmissionEmail.tsx
jamdelion: Ah wait I see what you mean now, pressing enter on the keyboard does not submit it (only pressing the continue button does), therefore the error handling is not triggered. Looking for a fix now...

jamdelion: @jessicamcinchak just looking at other inputs in the editor and it seems to be normal behaviour that just pressing 'enter' does not do anything. If you press tab it correctly finds the 'continue' button that you can then press enter on. Am I missing something else?

jessicamcinchak: Sorry the enter key part was my mistake - only the "Postcode" input on FindProperty does the on-enter-key validation (special a11y case because the next address input is conditional, not visible yet, I believe), was mis-remembering that we do that throughout ! This is working for me now :+1:

jamdelion: I do think now we have the summary table creating more distance between the 'Continue' button and the input, it makes it look a bit more confusing - it's not so clear that the continue button is for the input.

jessicamcinchak: Agree - maybe swap summary table to top & email input below? A nice specific design prompt for larger group maybe?

editor.planx.uk/src/routes/index.tsx
```diff
@@ -30,6 +31,13 @@ const editorRoutes = mount({
         }),
   ),
 
+  "/:team/:flow/:sessionId/download-application": map((req) => {
+    return route({
+      title: makeTitle("Download application"),
+      view: <VerifySubmissionEmail params={req.params} />,
```
DafyddLlyr, line 37: As a user I'd expect this to still be within an existing PlanX view (e.g. with header and footer).

There's a good example of this on the invite to pay pages - I think it's just a cast of wrapping this in the `standaloneView()` function.

editor.planx.uk/src/pages/SubmissionDownload/tests/VerifySubmissionEmail.test.tsx
DafyddLlyr, line 1: Good to see these `.todo()`s 👍

DafyddLlyr (changes requested):
Testing and working as expected!

One comment about the layout that's worth addressing I think 👍

editor.planx.uk/src/routes/index.tsx
jamdelion: Thanks for point that my way - looks loads better now! 🎉

editor.planx.uk/src/ui/shared/ErrorWrapper.tsx
```diff
@@ -7,7 +7,7 @@ import React, { ReactElement, useEffect } from "react";
 import { FONT_WEIGHT_SEMI_BOLD } from "theme";
 
 export interface Props {
-  error?: string;
+  error?: string | null;
```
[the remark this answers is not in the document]
jamdelion: Thanks, I've reverted this - think it was leftover from when I was trying to sort some typescript error but nothing has broken now that I've reverted it 🤞

DafyddLlyr (approved):

DafyddLlyr:
Closed in favour of #4926

PR #4002, opened 2024-11-22
test: Add retry to flaky react unit test

DafyddLlyr:
This test has been failing on a number of CI runs, and then usually passes on a retry.

image

I'm not able to recreate this locally, so I'm adding in a retry to see if this mitigates the issue.

Once this is up and running, I'll retry the react tests a bunch of time on CI to check that a single retry actually seems to resolve this.

Update: Ran this multiple times, no issues so far 🤞 

image


RODO94 (approved):
Thanks for picking this up, the change looks good to me. Hopefully just 1 retry is enough, but yeah it's a tough one to recreate the flake.

PR #4004, opened 2024-11-22
fix: External Portal node graph styling

DafyddLlyr:
Please see https://opensystemslab.slack.com/archives/C01E3AC0C03/p1732286143726329 (OSL Slack)

Fixed small visual regressions introduced in https://github.com/theopensystemslab/planx-new/pull/3975, and adds back missing `ref`s.

image

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Portal.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Portal.tsx
@@ -141,19 +141,22 @@ const InternalPortal: React.FC<any> = (props) => {
   return (
     <>
       <Hanger hidden={isDragging} before={props.id} parent={parent} />
-      <li className={classNames("card", "portal", { isDragging })} ref={ref}>
-        <Link
-          href={href}
-          prefetch={false}
-          ref={drag}
-          onContextMenu={handleContext}
-        >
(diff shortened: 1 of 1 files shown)


RODO94 (commented):
May I ask to include the Internal Portal here with this change? Realised I missed it when adding tags 

Should be:

```ts
  return (
    <>
      <Hanger hidden={isDragging} before={props.id} parent={parent} />
      <li ref={ref}>
        <Box className={classNames("card", "portal", { isDragging })}>
          <Link
            href={href}
            prefetch={false}
            ref={drag}
            onContextMenu={handleContext}
          >
            {Icon && <Icon />}
            <span>{props.data.text}</span>
          </Link>
          <Link href={editHref} prefetch={false} className="portalMenu">
            <MoreVert titleAccess="Edit Portal" />
          </Link>
        </Box>
        {props.tags?.map((tag: NodeTag) => <Tag tag={tag} key={tag} />)}
      </li>
    </>
  );

```

DafyddLlyr:
@RODO94 Yep! Added in [af8082e](https://github.com/theopensystemslab/planx-new/pull/4004/commits/af8082e2069ed678a86c6a13875d5da9f6078923)

RODO94 (approved):
Looks good to me

PR #4005, opened 2024-11-25
feat: Add External Portals to E2E ui-driven tests

RODO94:
## What does this PR do?

This PR adds the External Portal component to the ui-driven end to end testing. This required adding another flow to the `create-flow.spec` so that it could be added properly and adjusting how `flows` are handled in `context`. If have separated out the flows so we have `flow` which is intended to be the main one, then `externalPortalFlow` which is intended to only be for creating an External Portal. I found `mainFlow`, `targetFlow`, `parentFlow` unnecessary and making `externalPortalFlow` felt more explicit and easy to understand what it is

I also set up some test fixes on flaky parts of the test and added helpers to simplify the process of publishing, turning a service online, and navigating between services.


e2e/tests/ui-driven/src/helpers/context.ts
```diff
@@ -15,13 +23,7 @@ export interface Context {
...
   };
   team: { id?: number } & NewTeam;
-  flow?: {
-    id?: string;
-    publishedId?: number;
-    slug: string;
-    name: string;
-    data?: object;
-  };
+  flows?: Flow[];
```
RODO94, line 26: This was the main change to `flows` so that the context takes in an array of flows instead of one flow object.

e2e/tests/ui-driven/src/create-flow.spec.ts
```diff
@@ -158,15 +168,7 @@ test.describe("Flow creation, publish and preview", () => {
...
-    // Open flow settings
-    page.locator('[aria-label="Service settings"]').click();
-
-    // Toggle flow online
-    page.getByLabel("Offline").click();
-    page.getByRole("button", { name: "Save", disabled: false }).click();
-    await expect(
-      page.getByText("Service settings updated successfully"),
-    ).toBeVisible();
+    await turnServiceOnline(page);
```
jamdelion, line 171: Nice, I highly approve of extracting out these helpers! 👍

jamdelion: Could the previous line `await page.goto(`/${context.team.slug}/${serviceProps.slug}`);` be included in these helpers?

e2e/tests/ui-driven/src/create-flow.spec.ts
```diff
@@ -177,6 +179,54 @@ test.describe("Flow creation, publish and preview", () => {
...
+      externalPortalFlowData.title,
+      externalPortalFlowData.answers,
+    );
+    await expect(editor.nodeList).toContainText([
+      externalPortalFlowData.title,
+      externalPortalFlowData.answers[0],
+      externalPortalFlowData.answers[1],
+    ]);
+
+    await publishService(page);
```
jamdelion, line 216: question: why do we publish the service twice in this test? Can we not just create all the components and then publish/turn online at the end? I guess this would remove the need to call `navigateToService` as well.

e2e/tests/ui-driven/src/helpers/context.ts
```diff
@@ -6,6 +6,14 @@ import { log } from "./globalHelpers";
 
 type NewTeam = Parameters<CoreDomainClient["team"]["create"]>[0];
 
+export interface Flow {
```
jamdelion, line 9: nit: Personally I'd start extracting the types from this file into their own file, it starts making less sense when there are interfaces with different names from the name of the file.

e2e/tests/ui-driven/src/helpers/context.ts
```diff
@@ -65,23 +67,24 @@ export async function setUpTestContext(
     });
   }
   if (
-    context.flow?.slug &&
-    context.flow?.data &&
-    context.flow?.name &&
+    context.flows &&
+    context.flows[0].slug &&
```
jamdelion, line 71: suggestion: to make this easier to read you could make a const at the top called `firstFlow` or something like that and then use that everywhere that `context.flows[0]` is used. 

E.g.
```
const firstFlow = context.flows[0]
...
firstFlow.slug &&
firstFlow.data &&
...
```

You can probably think of a better name!

e2e/tests/ui-driven/src/create-flow.spec.ts
jamdelion: I think there are other instances in the test where `navigateToService` could be used, e.g. line 121

e2e/tests/ui-driven/src/helpers/navigateAndPublish.ts
```diff
@@ -0,0 +1,28 @@
+import { expect, Page } from "@playwright/test";
+import { contextDefaults } from "./context";
+
+export const navigateToService = async (page: Page, slug: string) => {
+  await page.goto(`/${contextDefaults.team.slug}/${slug}`);
+
+  await expect(page.getByRole("link", { name: slug })).toBeVisible();
+
+  return true;
```
jamdelion, line 9: nit: do we need to return anything from this function?

e2e/tests/ui-driven/src/invite-to-pay/agent.spec.ts
```diff
@@ -118,11 +119,11 @@ test.describe("Agent journey @regression", async () => {
...
-    await expect(
-      secondPage.getByRole("heading", {
-        name: "Sorry, you can't make changes to this application",
-      }),
-    ).toBeVisible();
+    const errorHeader = secondPage.getByRole("heading", {
+      name: "Sorry, you can't make changes to this application",
+    });
+
+    await expect(errorHeader).toBeVisible();
```
jamdelion, line 126: nice refactor 😍

jamdelion (approved):
A few questions and suggestions from me, mainly to improve readability/consistency. But nothing blocking so feel free to follow up in a separate PR if you like! 💪

e2e/tests/ui-driven/src/create-flow.spec.ts
RODO94: Yeah good shout, I was treading quite lightly with this.

e2e/tests/ui-driven/src/create-flow.spec.ts
```diff
@@ -177,6 +179,54 @@ test.describe("Flow creation, publish and preview", () => {
...
+
+    page.on("dialog", (dialog) =>
+      dialog.accept(externalPortalServiceProps.name),
+    );
+    await editor.addNewService();
+
+    // update context to allow new flow to be torn down
+    context.flows?.push({ ...externalPortalServiceProps });
+
+    await createQuestionWithOptions(
```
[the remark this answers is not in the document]
RODO94: Hmm yeah I agree, would keep things tidier and more consistent, I'll add the change

e2e/tests/ui-driven/src/create-flow.spec.ts
jamdelion: Not sure if we're talking about the same thing 😅 I think it's fair enough having a separate portal test like this. I was referring to the fact that in this test ('can create external portal') you have:
```js
createQuestionWithOptions()

publishService() // <--- this is the publish I have issue with! 
// Why do we publish after creating the question as well as at the end (after creating the portal)?

turnServiceOnline() // I think this should be moved to the bottom with the publish

navigateToService()

createExternalPortal()

publishService() // I think this should be the only publish
```

Happy to chat if you prefer 😁

e2e/tests/ui-driven/src/helpers/context.ts
```diff
@@ -91,12 +94,27 @@ export async function setUpTestContext(
...
     await deleteSession(adminGQLClient, context);
-    await deletePublishedFlow(adminGQLClient, context);
-    await deleteFlow(adminGQLClient, context);
+
+    for (const flow of context.flows) {
+      await deleteFlow(adminGQLClient, flow);
+      if (flow.publishedId) {
+        await deletePublishedFlow(adminGQLClient, flow);
+      }
+    }
```
DafyddLlyr, line 117: Have a look at using `$admin.flow._destroyAll()` here - I also think that published flows should be deleted via cascade when the source flow is deleted.

DafyddLlyr (changes requested):
Some great comments here by @jamdelion 🙌 

Great to see this being tested, a few comments to do with how `context` is handled - we should probably aim to keep this simple with a single `flow` if possible. Very open to thoughts here!

e2e/tests/ui-driven/src/create-flow.spec.ts
RODO94: Ahh because we need to publish the external portal service, then add that external portal to the main service and publish that as a change.

1. Create **External Portal service**
2. Publish **External Portal service**
3. Navigate to **Main service**
4. Add External Portal component to **Main service**
5. Publish changes to **Main service**

Otherwise it won't pick up the external portal when we answer the questions in the published link?

e2e/tests/ui-driven/src/helpers/context.ts
DafyddLlyr: There should be a good comparable example in the API e2e tests btw 👌

e2e/tests/ui-driven/src/create-flow.spec.ts
```diff
@@ -177,6 +179,54 @@ test.describe("Flow creation, publish and preview", () => {
...
+      externalPortalFlowData.answers[0],
+      externalPortalFlowData.answers[1],
+    ]);
+
+    await publishService(page);
+    await turnServiceOnline(page);
+
+    navigateToService(page, serviceProps.slug);
+
+    await createExternalPortal(page, page.locator("li:nth-child(6)"));
```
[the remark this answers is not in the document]
RODO94: I wanted it to be added in the middle of the flow rather than after the Review and Send components, so we would test going into an external portal and exiting an external portal. If I added it at the end with `getNextNode` then we wouldn't have that exit scenario convered?

e2e/tests/ui-driven/src/helpers/context.ts
RODO94: I think it is worthwhile having a fuller `Flow` type from `planx-core` but feel it sits outwith this PR, as I'd like to keep this focused on the tests. I'll queue up the change in a separate PR and add it to my list for this sprint

e2e/tests/ui-driven/src/create-flow.spec.ts
jamdelion: Ah ok that makes sense! Maybe just a comment next to each publish to say which service is being published then :)

RODO94: Yeah that's sick!

e2e/tests/ui-driven/src/helpers/context.ts
DafyddLlyr, line 1: A few code checks getting flagged for this file in CI - most unused functions now that we're using the `_deleteAll()` methods.

e2e/tests/ui-driven/src/helpers/types.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export interface TestContext {
+  user: Pick<
+    User,
+    "firstName" | "lastName" | "email" | "isPlatformAdmin" | "id"
+  >;
+  team: { id?: number } & NewTeam;
+  flow?: Flow;
+  externalPortalFlow?: Flow;
```
DafyddLlyr, line 21: This looks great and I think will match out general assumptions at this point 👍 Thanks for picking up this change.

e2e/tests/ui-driven/src/helpers/types.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface Flow {
+  id?: string;
+  publishedId?: number;
+  slug: string;
+  name: string;
+  data?: object;
+}
+
+export interface TestContext {
```
DafyddLlyr, line 14: nit: I'm not sure if `TestContext` is really much clearer or better than `Context` here when we're entirely within Playwright tests.

Not a strongly held opinion - was there a motivation for this change? Classes with `BrowserContext` maybe?

DafyddLlyr (changes requested):
One small tidy up around redundant functions and we're good to go here, last round of changes look good 👍

e2e/tests/ui-driven/src/helpers/types.ts
RODO94: I can't fully remember why, but it did come out of some frustration with the generalised context. Might have been to help express that as a Type it was generated within the Test, rather than bringing it in from external Types. If that makes sense?

DafyddLlyr (approved):
Great! ✅

PR #4010, opened 2024-11-27
feat(fee-breakdown): Initial setup and logic with `useFeeBreakdown()` hook

DafyddLlyr:
## What does this PR do?
Sets up basic logic for parsing fee details from the passport, and displaying them to the user via the `FeeBreakdown` component

image

## Design decisions
These are largely points up for discussion, I just want to explain some of my reasoning and outline potential next steps.

**If we cannot parse values from the passport, do not show the breakdown to the user**
Even if the `showFeeBreakdown` prop is set to `true`, the breakdown still won't show if we cannot display it to the user (e.g. missing or invalid values). If this happens it's a content issue which will be flagged via Airbrake for content editors to ultimately fix - the user journey won't be disrupted.

**Hardcode `application.fee` data field**
This is a rough MVP decision just to move forward initially, it's not a big additional task to read `data.fn` from the relevant Pay component here.

**Reductions and exemptions are combined to a generic "reduction" value**
The fee calculator doesn't associate a numerical value with exemptions and reductions, but just updates `application.fee.payable` and sets true/false values. There's a few options around how we could handle this - 
 - Update content to always set a passport value, and then remove it from `application.fee.payable`
 - List exemptions and reduction, and sum the totals
I'm very keen to avoid repeating any logic already handled in the `fee-calculator` flow - doing so puts us in an awkward code vs. low-code spot with regards to how this logic is handled. 

Again, I think we take a fallback approach here. If we cannot parse a list of exemptions or reductions from the passport, we can fallback to summed generic "reductions" amount?

**VAT is set via a Calculate component, and not re-calculated**
I'm not sure about this one - it allows room for error if an Editor incorrectly calculates this value. A more robust approach might be to allow the user to set a `application.fee.payable.includesVAT` variable, and display an amount based on this.
- PR here - https://github.com/theopensystemslab/planx-new/pull/4042

**Nothing is really explained to Editors**
I think this will need something - I think just adding a list of variables and how this words in the Editor would achieve this.

**Disabled for Invite to pay journeys**
This is something I hadn't initially considered - we have no passport when returning to the service via invite to pay so the current approach does not work. We'll need to persist the fee breakdown data to the `payment_requests` table in a similar way to how `govpay_metadata` is stored. I'll open another PR shortly to provide a solution here.

## Next steps...
Here's what I'd like to go for as I suspect the current implementation is too MVP - 
 - List reductions and exemptions, but with a single summed value each (exemptions = full amount, reductions = calculated - payable)
 - Handle invite to pay journeys
 - Add explanation / documentation within the Pay editor modal if `showFeeBreakdown` is checked


editor.planx.uk/src/@planx/components/Pay/Public/FeeBreakdown/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+  data["application.fee.calculated"]
+    ? data["application.fee.calculated"] - data["application.fee.payable"]
+    : 0;
+
+/**
+ * Transform Passport data to a FeeBreakdown shape
+ */
+export const toFeeBreakdown = (data: PassportFeeFields): FeeBreakdown => ({
+  applicationFee:
+    data["application.fee.calculated"] || data["application.fee.payable"],
```
DafyddLlyr, line 22: This essentially matches the logic here - 

https://github.com/theopensystemslab/planx-core/blob/main/src/export/digitalPlanning/model.ts#L996

editor.planx.uk/src/@planx/components/Pay/Public/FeeBreakdown/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+    .union([questionSchema, setValueSchema])
+    .transform(toNumber);
+
+  const schema = z
+    .object({
+      "application.fee.calculated": feeSchema.optional().default(0),
+      "application.fee.payable": feeSchema,
+      "application.fee.payable.vat": feeSchema.optional().default(0),
+    })
+    .transform(toFeeBreakdown);
```
DafyddLlyr, line 41: This parse & transform pattern might be a nice approach to try in `planx-core` passport → payload mappings.

editor.planx.uk/src/@planx/components/Pay/Public/FeeBreakdown/types.ts
```diff
@@ -0,0 +1,12 @@
+export interface FeeBreakdown {
+  applicationFee: number;
+  total: number;
```
jessicamcinchak, line 3: nit: introducing `applicationFee` and `total` language feels like a bit of extra mental load here and I wonder if this can re-use `calculated` (which if undefined, fallsback to payable) and `payable` instead?

jessicamcinchak: Before reading the tests, I would have also personally guessed that `total` = full `calculated` amount and `applicationFee` = `payable` but your passport mappings are the other way around !

editor.planx.uk/src/@planx/components/Pay/Public/FeeBreakdown/types.ts
```diff
@@ -0,0 +1,12 @@
...
+  reduction: number;
+  vat: number | undefined;
+}
+
+export interface PassportFeeFields {
+  "application.fee.calculated": number;
+  "application.fee.payable": number;
+  "application.fee.payable.vat": number;
```
jessicamcinchak, line 11: Question - are you expecting `application.fee.payable.vat` to be an _input_ (eg already set in passport _before_ Pay is reached; this interface & unit tests indicate yes?) OR is it value that is _calculated by_ Pay and written to passport via Pay's `handleSubmit`? 

I was actually thinking the latter based on `VAT_RATE` defined earlier in the component, and a user journey like: 
- As an editor I have an "Apply VAT" toggle in my Pay component that is off my default. If I toggle it on then VAT is both included in displayed fee breakdown _and_ `application.fee.payable.VAT` amount will be written to passport on Pay `handleSubmit` (or sooner so capturable by Invite to pay journey) ? But I never need to create a separate SetValue or Calculate or similar myself which sets `application.fee.payable.vat` ?

jessicamcinchak (approved):
This looks overall good - a few clarifying questions and language suggestions but nothing stopping merge at this point as this is all feature-flagged :+1:

editor.planx.uk/src/@planx/components/Pay/Public/FeeBreakdown/types.ts
DafyddLlyr: Initially (and in this PR) I was following [the pattern set here](https://editor.planx.dev/doncaster/pre-application-advice/nodes/_root/nodes/EIlRrlYGo6/edit) where Editors set this value before hitting the pay component though.

This is far from ideal though and on Monday we decided to - 
 - Do the calculation ourselves
 - Allow Editors to set a passport field of `application.fee.payable.includesVAT` which controls this

This is implemented here as a follow up PR - https://github.com/theopensystemslab/planx-new/pull/4042

editor.planx.uk/src/@planx/components/Pay/Public/FeeBreakdown/types.ts
DafyddLlyr: Useful feedback thanks - this was an intentional decision to try and make it a little clearer and more explicit (I did also consider gross/net as well at one point). 

Clearly not successful though 😂 Your point about consistency and mental load is a very valid one 👍

DafyddLlyr:
Thanks very much for the review 👍 

Due to brancher and other PRs which rely on this I'm going to fix these issues forward - all currently feature flagged.

PR #4017, opened 2024-11-28
fix: Clone styling for Questions with data variables or tags

DafyddLlyr:
## What does this PR do?
Updates structure of graph nodes so that is the `isClone` class always targets all children within a single `div` element.

| Before | After |
|--------|--------|
| image | image | 
| Border not applied to sibling data tag, 3px gap | Border applied to single parent div, 3px gap is correctly offsetting hanger above |

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Question.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Question.tsx
@@ -1,4 +1,5 @@
 import ErrorIcon from "@mui/icons-material/Error";
+import Box from "@mui/material/Box";
 import {
   ComponentType as TYPES,
   NodeTags,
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Question.tsx
```diff
@@ -76,25 +77,27 @@ const Question: React.FC<Props> = React.memo((props) => {
...
-          prefetch={false}
-          onContextMenu={handleContext}
-          ref={drag}
-        >
-          {props.data?.img && (
-            <Thumbnail
-              imageSource={props.data?.img}
-              imageAltText={props.data?.text}
-            />
+        <Box>
```
DafyddLlyr, line 80: Wrapping this in the box ensures that the `isClone > div` selector is now properly targeted, matching how checkboxes are already set up.

RODO94 (approved):
Changes look good to me and tested well on the pizza following the messages on the Slack thread

PR #4019, opened 2024-11-28
test: verify email component

jamdelion:
The long-awaited filling in of the `test.todo`s for the [verify submission email component](https://github.com/theopensystemslab/planx-new/pull/4000)!

I have added Mock Service Worker as a package ([recommended by react-testing-library](https://github.com/theopensystemslab/planx-new/pull/4000) as the best way to mock API calls, and something I have worked with before). I'm using here to intercept calls to `/download-application-file` so I can test behaviour in an error state, and check that files are downloaded without having to do complicated download mocks.


editor.planx.uk/src/setupTests.tsx
```diff
@@ -13,10 +14,16 @@ import { vi } from "vitest";
  */
 const mockFade = vi.fn(({ children }: FadeProps) => <div>{children}</div>);
 
-vi.mock("@mui/material/Fade", () => ({
+vi.doMock("@mui/material/Fade", () => ({
```
jamdelion, line 17: Errored after I added in the msw beforeEach stuff, so found this solution from https://vitest.dev/api/vi.html#vi-mock

> vi.mock is hoisted (in other words, moved) to top of the file. It means that whenever you write it (be it inside beforeEach or test), it will actually be called before that.

> This also means that you cannot use any variables inside the factory that are defined outside the factory.

> If you need to use variables inside the factory, try [vi.doMock](https://vitest.dev/api/vi.html#vi-domock). It works the same way but isn't hoisted. Beware that it only mocks subsequent imports.

DafyddLlyr: Appreciate the explanation and link to docs!

PR #4020, opened 2024-11-29
feat: support "address" fields in List schemas

jessicamcinchak:
**Changes:**
- Building on #4153, adds support for `type: "address"` fields in List & Page schemas, 
- `"address"` fields follow our `AddressInput` component pattern, with a mix of required and optional fields
- Adds test coverage
- Adds a preliminary "Owners" schema (very much anticapte content team will want to edit this!)

**Testing:** https://4020.planx.pizza/testing/owners-as-list

Screenshot from 2025-01-20 22-27-10


editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
```diff
@@ -637,6 +639,8 @@ function Page(props: ComponentProps) {
     if (isTextResponse(answer)) return answer;
     if (isNumberFieldResponse(answer)) return answer.toString();
     if (isMapFieldResponse(answer)) return `${answer.length || 0} features`;
+    if (isAddressFieldResponse(answer))
+      return formatAsSingleLineAddress(answer);
```
jessicamcinchak, line 643: Nice & tidy :+1:

editor.planx.uk/src/@planx/components/shared/Schema/InputFields/AddressFieldInput.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const AddressFieldInput: React.FC<Props<AddressField>> = (props) => {
+  const { data, formik } = props;
+  const { id, errorMessage, value, name } = getFieldProps(props);
+
+  /**
+   * Maps changes from AddressField inputs to their corresponding nested form fields
+   * @example
+   * The input for name=line1 will be mapped to the schemaData[0][yourFn]["line1"] formik field
+   */
```
jessicamcinchak, line 22: :+1:

jessicamcinchak:
[List component should support Address inputs and optional fields for better UX of "Ownership" questions & schema alignment](https://trello.com/c/wMVVmvnn/3172-list-component-should-support-address-inputs-and-optional-fields-for-better-ux-of-ownership-questions-schema-alignment)

editor.planx.uk/src/@planx/components/shared/Schema/model.ts
```diff
@@ -162,6 +175,11 @@ export const isChecklistFieldResponse = (
 ): response is ResponseValue<ChecklistField> =>
   Array.isArray(response) && !isMapFieldResponse(response);
 
+export const isAddressFieldResponse = (
+  response: unknown,
+): response is ResponseValue<AddressField> =>
+  typeof response === "object" && response !== null && "line1" in response;
+
```
DafyddLlyr, line 182: For context please see https://github.com/theopensystemslab/planx-new/pull/4153#discussion_r1919809762

editor.planx.uk/src/@planx/components/AddressInput/model.ts
```diff
@@ -1,25 +1,18 @@
...
-  postcode: string().required("Enter a postcode"),
-  country: string(),
-});
+export const addressValidationSchema = (): SchemaOf<Address> =>
+  object({
+    line1: string().required("Enter the first line of an address"),
+    line2: string(),
+    town: string().required("Enter a town"),
+    county: string(),
+    postcode: string().required("Enter a postcode"),
```
RODO94, line 13: nit: multi data entry can be filled with errors compared with the one at a time type, do we want to increase the guards due to this. On postcode, could be something like:

```ts
postcode: string().matches(/^[a-z]{1,2}\d[a-z\d]?\s*\d[a-z]{2}$/i,"Enter a valid UK postcode").required("Enter a postcode"),
```

Put as `nit` since it's an additional feature for the Address Input component, but just thought it could be useful

RODO94: Actually, thought about it more, and if the reason we don't do this is because addresses can be from anywhere, not just the UK, I completely get that

editor.planx.uk/src/@planx/components/List/Public/index.test.tsx
```diff
@@ -452,8 +452,23 @@ describe("Form validation and error handling", () => {
     );
 
     let errorMessages = getAllByTestId(/error-message-input/);
-    // One error per field, plus 3 for a date input (one per input)
-    const numberOfErrors = mockZooProps.schema.fields.length + 3;
+
+    let numberOfErrors = 0;
+    mockZooProps.schema.fields.forEach((field) => {
```
RODO94, line 457: I like this! Nice way to clearly manage the date error count

editor.planx.uk/src/@planx/components/shared/Schema/model.ts
RODO94: These type guard functions are sick, not looked as deeply into them until now

editor.planx.uk/src/@planx/components/AddressInput/Public.tsx
```diff
@@ -38,7 +30,7 @@ export default function AddressInputComponent(props: Props): FCReturn {
     },
     validateOnBlur: false,
     validateOnChange: false,
-    validationSchema: userDataSchema,
+    validationSchema: addressValidationSchema,
```
RODO94, line 33: question: significance of this switch is now the List schema now shares the same validation schema as this address input component?

RODO94: Thus, you gave it a more specific name

RODO94: And the list schema is now using the `<AddressFields/>` within the `<AddressFieldsInput/>`

RODO94 (approved):
Tested it locally and on the pizza, everything seems to be working for me as I'd expect. With all data tracking across nicely to the passport.

Only added one `nit` comment related to error handling and how strict to be with validating data in the components. 

The current Address Component doesn't have this and it can get complex if we handle addresses outside the UK, so happy for this to be left out at your discretion.

editor.planx.uk/src/@planx/components/shared/Schema/model.ts
DafyddLlyr: A really good alternative to casting. 

Casting lets you bypass the type system in "god mode", but type guards force you to explain yourself - much more future proof and intelligible to other devs (and your future self!).

editor.planx.uk/src/@planx/components/AddressInput/model.ts
DafyddLlyr: Exactly 👍 I don't think there's a requirement (from a planning perspective) that things like an agent's or landowner's address has to be in the UK so we intentionally loose here with validation. Somebody may enter a US zip code for example.

I do agree though, there might be an argument that the address component allows an Editor to select "UK address" or "International address" or something in future. This hasn't yet come up as a requirement as far as I know.

DafyddLlyr: We also use https://github.com/ideal-postcodes/postcode elsewhere to validate and format postcodes which has a nice API.

editor.planx.uk/src/@planx/components/AddressInput/Public.tsx
DafyddLlyr: Yes - specific names mean we don't have to import multiple instances of `userDataSchema` and cast them to a meaningful name. 

This approach is quite fiddly and error prone, and you need to remember the naming convention and repeat across files 👇 

```ts
import { userDataSchema as addressValidationSchema } from "@planx/components/AddressInput/model";
import { userDataSchema as textValidationSchema } from "@planx/components/TextInput/model";
```

This isn't actually applied consistently yet in this file - will push up a quick commit now to address this.

editor.planx.uk/src/@planx/components/AddressInput/Public.tsx
```diff
@@ -50,16 +42,36 @@ export default function AddressInputComponent(props: Props): FCReturn {
...
+}
+
+export interface AddressFieldsProps {
+  id?: string;
+  values: Address;
+  errors: FormikErrors<Address>;
+  handleChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+}
+
+export function AddressFields(props: AddressFieldsProps): FCReturn {
```
jamdelion, line 62: nit/personal opinion: I usually try to have one file per React component, so would prefer to see this in its own file, but that's just me!

editor.planx.uk/src/@planx/components/AddressInput/model.ts
```diff
@@ -1,25 +1,18 @@
...
-
-export const userDataSchema: SchemaOf<Address> = object({
-  line1: string().required("Enter the first line of an address"),
-  line2: string(),
-  town: string().required("Enter a town"),
-  county: string(),
-  postcode: string().required("Enter a postcode"),
-  country: string(),
-});
+export const addressValidationSchema = (): SchemaOf<Address> =>
```
jamdelion, line 7: Good to have renamed this 👍

editor.planx.uk/src/@planx/components/List/Public/index.test.tsx
```diff
@@ -167,7 +167,7 @@ describe("Building a list", () => {
     expect(addItemButton).not.toBeInTheDocument();
   });
 
-  test("Adding an item", { timeout: 20000 }, async () => {
+  test("Adding an item", { timeout: 35000 }, async () => {
```
jamdelion, line 170: question: what is the timeout for? Is this test flaky?

jamdelion: also nit: we can include a numeric separator to make these big numbers easier to read, e.g. `35_000`

editor.planx.uk/src/@planx/components/List/Public/index.test.tsx
jamdelion, line 1: This is a humungous file! Again personal preference but I'd be tempted to split this into more test files, e.g. `index.validation.test.tsx`, `index.payloadGeneration.test.tsx` so that it's easier to read/find the tests you want. 🤷

editor.planx.uk/src/@planx/components/AddressInput/Public.tsx
DafyddLlyr: Please see [93edced](https://github.com/theopensystemslab/planx-new/pull/4020/commits/93edcede48176ce3e155b1ccf78ed852ab611ab5)

editor.planx.uk/src/@planx/components/List/Public/index.test.tsx
DafyddLlyr: Great feedback thanks - timeouts now have numerical separators - please see [5435bf9](https://github.com/theopensystemslab/planx-new/pull/4020/commits/5435bf97c6b294d71ff1eae9db72529ce74850a4) ✅

The tests here are pretty slow! They pass locally without timeouts as generous as this but I think that's an issue with underpowered GHA runners needing a larger timeout than my local machine.

This issue here has come up in a few other PRs and it's not flakeiness but some sort of state update interplay between vitest, formik, and React context which leads to these being slow tests. I've not go to the bottom of this and it does seem to be a test-only issue. It might be the case that we need to revisit how we're using context providers for the schema - I've had a few passes over adding `useMemo()`s etc here and there without any seeming effect on the test speeds.

PR #4023, opened 2024-12-02
feat: adding application.type to feedback_summary hasura migration

zz-hh-aa:
Adds a new Hasura migration to add `application.type` to `feedback_summary` view.

We tested it in a flow with find property, application type etc. and everything worked.

### DIFF
--- a/hasura.planx.uk/migrations/1733134891363_add_application_type_to_feedback_summary/up.sql
+++ b/hasura.planx.uk/migrations/1733134891363_add_application_type_to_feedback_summary/up.sql
@@ -21,7 +21,7 @@ CREATE OR REPLACE VIEW "public"."feedback_summary" AS
     (((fb.user_data -> 'passport'::text) -> 'data'::text) ->> 'proposal.projectType'::text) AS project_type,
     (((fb.user_data -> 'passport'::text) -> 'data'::text) ->> 'property.constraints.planning'::text) AS intersecting_constraints,
     fb.feedback_score,
-    ((((fb.user_data -> 'passport'::text) -> 'data'::text) -> 'application.type'::text) ->> 0) AS application_type
+    fb.user_data -> 'passport' -> 'data' -> 'application.type' ->> 0 AS application_type
    FROM ((feedback fb
      LEFT JOIN flows f ON ((f.id = fb.flow_id)))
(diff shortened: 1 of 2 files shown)


hasura.planx.uk/migrations/1733134891363_add_application_type_to_feedback_summary/up.sql
```diff
@@ -0,0 +1,28 @@
...
+    (fb.node_data ->> 'description'::text) AS node_text,
+    (fb.node_data ->> 'info'::text) AS help_text,
+    (fb.node_data ->> 'policyRef'::text) AS help_sources,
+    (fb.node_data ->> 'howMeasured'::text) AS help_definition,
+    COALESCE(((((fb.user_data -> 'passport'::text) -> 'data'::text) -> '_address'::text) ->> 'single_line_address'::text), ((((fb.user_data -> 'passport'::text) -> 'data'::text) -> '_address'::text) ->> 'title'::text)) AS address,
+    ((((fb.user_data -> 'passport'::text) -> 'data'::text) -> '_address'::text) ->> 'uprn'::text) AS uprn,
+    (((fb.user_data -> 'passport'::text) -> 'data'::text) ->> 'proposal.projectType'::text) AS project_type,
+    (((fb.user_data -> 'passport'::text) -> 'data'::text) ->> 'property.constraints.planning'::text) AS intersecting_constraints,
+    fb.feedback_score,
+    ((((fb.user_data -> 'passport'::text) -> 'data'::text) -> 'application.type'::text) ->> 0) AS application_type
```
DafyddLlyr, line 24:
```suggestion
    fb.user_data -> 'passport' -> 'data' -> 'application.type' ->> 0 AS application_type
```

I think this is the same just much clearer, we don't need to cast this chain all the way down.

Could you try this out and check it still works?

DafyddLlyr (changes requested):

hasura.planx.uk/migrations/1733134891363_add_application_type_to_feedback_summary/up.sql
zz-hh-aa: Good catch, worked for me so pushed the change!

DafyddLlyr (approved):
Perfect! Once this is merged we can queue up a prod deployment so that you can get this across onto Metabase.

PR #4021, opened 2024-11-29
feat: visually indicate if a node has help text in the graph

jessicamcinchak:
Adds a simple toggle & icon to indicate if a node has help text defined without having to open its' modal

Screenshot from 2024-12-02 09-47-59

PR #4024, opened 2024-12-02
fix: Handle nodes without more info fields

DafyddLlyr:
Quick fix following on from #4021 

This account for nodes which don't have the "More information" fields - currently hitting this here on staging https://editor.planx.dev/templates/apply-for-a-rights-of-way-order

image


jessicamcinchak (approved):
Thanks for catching & quick fix !

PR #4025, opened 2024-12-02
feat: Add E2E Coverage for Property Information

RODO94:
Added coverage for the Property Information component in the E2E ui-driven tests. 

Covered adding the component in the Editor, Navigating it in the Public view, and changing the property type.

In order to see if the map is working, I have added a check to review the `geoJsonData` attribute and ensure correct data is being added here, which should indicate the map is working.

Added a new package here to bring in the same types as we use in the Editor for `Feature` so we can match it up with the mock

### DIFF
--- a/e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
+++ b/e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
@@ -145,8 +145,7 @@ test.describe("Flow creation, publish and preview", () => {
     await expect(
       page.getByRole("radio", { name: "Residential", checked: true }),
     ).toBeVisible();
-
-    // change button navigates the user back thro the service so we need to continue twice
+    
     await answerQuestion({
       page: page,
       title: "What type of property is it?",
--- a/e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
+++ b/e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
@@ -15,6 +15,8 @@ import {
 } from "./helpers/navigateAndPublish";
 import { TestContext } from "./helpers/types";
 import { serviceProps } from "./helpers/serviceData";
+import { mockMapGeoJson } from "./mocks/geoJsonMock";
+import { checkGeoJsonContent } from "./helpers/geospatialChecks";
 
 test.describe("Flow creation, publish and preview", () => {
   let context: TestContext = {
@@ -48,6 +50,8 @@ test.describe("Flow creation, publish and preview", () => {
 
     await editor.createFindProperty();
     await expect(editor.nodeList).toContainText(["Find property"]);
+    await editor.createPropertyInformation();
+    await expect(editor.nodeList).toContainText(["About the property"]);
     await editor.createInternalPortal();
     await editor.populateInternalPortal();
     await page.getByRole("link", { name: "start" }).click(); // return to main flow
@@ -103,6 +107,17 @@ test.describe("Flow creation, publish and preview", () => {
       page.locator("h1", { hasText: "Find the property" }),
     ).toBeVisible();
     await answerFindProperty(page);
+    await clickContinue({ page });
+
+    await expect(
+      page.getByRole("heading", { name: "About the property" }),
+    ).toBeVisible();
+
+    // Check map component has geoJson content
+    await checkGeoJsonContent(page, mockMapGeoJson);
+
+    // Check property info is being shown
+    await expect(page.getByText("Test Street, Testville")).toBeVisible();
     await clickContinue({ page });
 
     await expect(
--- a/e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
+++ b/e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
@@ -149,6 +149,8 @@ test.describe("Flow creation, publish and preview", () => {
     });
 
     await clickContinue({ page });
+    await expect(page.getByText("Commercial")).toBeVisible();
+
     await clickContinue({ page });
 
     await expect(
--- a/e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
+++ b/e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
@@ -118,6 +118,14 @@ test.describe("Flow creation, publish and preview", () => {
 
     // Check property info is being shown
     await expect(page.getByText("Test Street, Testville")).toBeVisible();
+    await expect(page.getByText("Residential - Semi Detached")).toBeVisible();
+    const changeButton = page.getByRole("button", {
+      name: "Change your Property type",
+    });
+
+    await changeButton.click();
+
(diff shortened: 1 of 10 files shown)


e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
```diff
@@ -105,6 +109,25 @@ test.describe("Flow creation, publish and preview", () => {
...
+    // Check property info is being shown
+    await expect(page.getByText("Test Street, Testville")).toBeVisible();
+    await expect(page.getByText("Residential - Semi Detached")).toBeVisible();
+    const changeButton = page.getByRole("button", {
+      name: "Change your Property type",
+    });
+
+    await changeButton.click();
+
+    await expect(page.getByText("Unknown")).toBeVisible();
```
jessicamcinchak, line 128: I don't think this "Unknown" error is actually meaningful to test here? 

The "change" link relies on the Preview store method `overrideAnswer` and notably _at least one node located before PropertyInformation that sets passport value `property.type`_ in order to successfully navigate as explained by other existing frontend tests (in "real" flows, this would be the entire `opensystemslab/property-types` portal).

I'd suggest:
- Adding "PropertyInformation" to the E2E tests _without_ the "change" link toggled on and confirming it's not visible (and relying on our existing editor test rather than e2e tests for navigation)
- Or correctly mocking this full journey, which is roughly:
  - The flow needs to create at least one node (Question or Checklist) that sets `property.type` in a location _before_ the PropertyInformation component and _after_ FindProperty (so it's auto-answered and never seen by user _unless_ the user clicks "change")
    - The node should have option passport values that are less-granular than the fetched property type from OS (eg `residential`, `commercial`) 
  - When you click "Change", we should test that you navigate _back_ to this node successfully and the `residential` option is auto-filled (eg not `Unknown` error !!)
  - You can successfully "change" the answer to another option
  - When you "Continue" forwards back to PropertyInformation, your new selection displays in type field

RODO94: This makes a lot more sense than the current way I was using it. I'll update the tests

RODO94:
[[E2E Tests] - Property Info](https://trello.com/c/XPqnKhWp/3135-e2e-tests-property-info)

e2e/tests/ui-driven/src/helpers/addComponent.ts
```diff
@@ -153,6 +157,21 @@ export const createQuestionWithOptions = async (
   );
 };
 
+export const createQuestionWithDataFieldOptions = async (
```
RODO94, line 160: Would appreciate feedback here, the `createBaseComponent` seems reliant on the `option: string[]` and my attempts to add in an option which contains a data value pulled on a lot of threads. These below lines seemed the easiest, lowest impact additions. 

Is there a pattern I should be using here with discriminated unions and overloaded functions in Typescript?

e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
```diff
@@ -48,6 +57,16 @@ test.describe("Flow creation, publish and preview", () => {
 
     await editor.createFindProperty();
     await expect(editor.nodeList).toContainText(["Find property"]);
+    // Find property will automate past this question at first
+    await editor.createQuestionWithDataFieldOptions(
+      "Assign property type",
```
jessicamcinchak, line 62: nit: a more realistic Question title would be "What type of property is it?"

e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
```diff
@@ -99,12 +118,41 @@ test.describe("Flow creation, publish and preview", () => {
...
+    // Check property info is being shown
+    await expect(page.getByText("Test Street, Testville")).toBeVisible();
+    await expect(page.getByText("Residential - Semi Detached")).toBeVisible();
+    const changeButton = page.getByRole("button", {
+      name: "Change your Property type",
+    });
+
+    await changeButton.click();
+
+    // change button navigates the user back thro the service so we need to continue twice
```
jessicamcinchak, line 144: When I initially land on this page, it'd be useful to check that the "Residential" option is selected by default based on the mocked property type value of "Residential - Semi Detached" above _before_ changing the answer to "Commercial". 

A key functionality/expectation of all PlanX "back" navigation is that previous answers are pre-populated/retained but we're failing to check that in this test.

e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
```diff
@@ -99,12 +118,41 @@ test.describe("Flow creation, publish and preview", () => {
...
+    await answerQuestion({
+      page: page,
+      title: "Assign property type",
+      answer: "Commercial",
+    });
+
+    await clickContinue({ page });
+    await expect(page.getByText("Commercial")).toBeVisible();
```
jessicamcinchak, line 152: nit: this would be slightly more clear to me if there were more verbose `expect` statements (or just more code comments) to confirm we've navigated _forwards_ to About the property _again_ before checking that the property type now displays our new selection of "Commercial" - eg
```
// "Continue" has navigated us back to PropertyInformation which should display updated selection
await expect(page.getByRole("heading", { name: "About the property" })).toBeVisible();
await expect(page.queryByText("Residential - Semi Detached")).toBeUndefined();
await expect(page.getByText("Commercial")).toBeVisible();
 ```

File: e2e/tests/ui-driven/src/helpers/addComponent.ts (hunk 2 above)
Comment (jessicamcinchak) [reply e2e/tests/ui-driven/src/helpers/addComponent.ts:160]: I'm fine with this approach of a new helper `createQuestionWithDataFieldOptions` - seems pragmatic and correct for moving this forward :+1:

**Review by jessicamcinchak — COMMENTED** [review]
Thanks for adjustments here - a lot closer but still a few comments to address please !

File: e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts [hunk 6]
```diff
@@ -99,18 +118,58 @@ test.describe("Flow creation, publish and preview", () => {
...
+    });
+
+    await changeButton.click();
+
+    // ensure residential is selected on back nav to test previouslySubmittedData is working
+    await expect(
+      page.getByRole("radio", { name: "Residential", checked: true }),
+    ).toBeVisible();
+
+    // change button navigates the user back thro the service so we need to continue twice
```
jessicamcinchak, line 149: nit: L149 looks like an outdated or misplaced comment !

jessicamcinchak (approved):
One lingering comment but otherwise happy for this one to merge - thanks for adjusting per feedback! 

Going forwards, would love to understand what would have helped make the expected steps of this test (eg how "change property type" works) more discoverable for you upfront without me catching that "Unknown" on first review.

e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
RODO94: great catch!

PR #4029, opened 2024-12-02
feat: conditionally show/hide "Continue" button on Confirmation using `isFinalCard`

jessicamcinchak:
Per request here: https://opensystemslab.slack.com/archives/C01E3AC0C03/p1733157241528129

**Changes:**
- The Confirmation component now determines if it is the final node in the flow
  - If it is the final node, it will not display a "Continue" button
  - If it is _not_ the final node, it **will** display a "Continue" button and leave a basic breadcrumb
  - In either scenario, the Confirmation component still _never_ displays a "back" button - which I think is correct/intended behavior as we would not allow someone to change answers post-Confirmation
- Hides the "Save and return..." link beneath "Continue" if there's a Send component present in our breadcrumbs

**Todo:** 
- :white_check_mark: Add some tests (honestly a little unsettling that this hasn't broken any :sweat_smile:)
  
**Testing questions:**
- Does this work as expected? Is there any sluggishness introduced when navigating between cards? 
- Is it okay that any node that comes _after_ Confirmation will _not_ be included in a submission payload? Does this impact any of our existing feedback/analytics processes for tracking user satisfaction, etc? 
- Do we need to hide the "Save and resume this application later" link under "Continue" on all nodes _after_ Confirmation?
  - :white_check_mark: See note above, now hidden if breadcrumbs include a Send node
- Do we need to introduce any publish validation rules for types of components that are allowed to come after Confirmation, eg no Pay or Send?

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
@@ -3,6 +3,7 @@ import Button from "@mui/material/Button";
 import Container from "@mui/material/Container";
 import Fade from "@mui/material/Fade";
 import { styled, Theme, useTheme } from "@mui/material/styles";
+import { ComponentType as TYPES } from "@opensystemslab/planx-core/types";
 import { useAnalyticsTracking } from "pages/FlowEditor/lib/analytics/provider";
 import { useStore } from "pages/FlowEditor/lib/store";
 import React, { useEffect } from "react";
@@ -52,8 +53,11 @@ const Card: React.FC<Props> = ({
     state.path,
     state.currentCard,
   ]);
+
   const showSaveResumeButton =
-    path === ApplicationPath.SaveAndReturn && handleSubmit;
+    path === ApplicationPath.SaveAndReturn &&
+    handleSubmit &&
+    visibleNode?.type !== TYPES.Confirmation;
   const { track } = useAnalyticsTracking();
 
   useEffect(() => {
--- a/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
@@ -49,15 +49,23 @@ const Card: React.FC<Props> = ({
   ...props
 }) => {
   const theme = useTheme();
-  const [path, visibleNode] = useStore((state) => [
+  const [path, visibleNode, breadcrumbs, flow] = useStore((state) => [
     state.path,
     state.currentCard,
+    state.breadcrumbs,
+    state.flow,
   ]);
 
+  // Check if we have a Send node in our breadcrumbs
+  //   In the frontend this is a better/more immediate proxy for "Submitted" because actual send events that populate lowcal_sessions.submitted_at are async
+  const hasSent = Object.keys(breadcrumbs)
+    .reverse()
+    .some(
+      (breadcrumbNodeId: string) => flow[breadcrumbNodeId]?.type === TYPES.Send,
+    );
+
   const showSaveResumeButton =
-    path === ApplicationPath.SaveAndReturn &&
-    handleSubmit &&
-    visibleNode?.type !== TYPES.Confirmation;
+    path === ApplicationPath.SaveAndReturn && handleSubmit && !hasSent;
   const { track } = useAnalyticsTracking();
 
   useEffect(() => {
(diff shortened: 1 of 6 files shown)


editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -670,9 +670,8 @@ export const previewStore: StateCreator<
   },
 
   isFinalCard: () => {
-    // Temporarily always returns false until upcomingCardIds is optimised
-    // OSL Slack explanation: https://bit.ly/3x38IRY
-    return false;
+    const { upcomingCardIds } = get();
+    return upcomingCardIds().length === 1;
```
jessicamcinchak, line 674: This one's been a long time coming ([original commit](https://github.com/theopensystemslab/planx-new/commit/f67560d483954829d8e4bd79b0a22a646bb8852d#diff-d93e4413e0cb16ffeec24f6df1499fd4f24074ba6837ece6c782bb90053e0f37R638)) & more promising now since `upcomingCardIds` has been recently refactored to only be responsible for navigation and no longer optimistic auto-answering

editor.planx.uk/src/pages/Preview/Node.tsx
```diff
@@ -75,16 +75,13 @@ interface Props {
...
-  );
-
-  const handleSubmit = isFinalCard ? undefined : props.handleSubmit;
+  const [childNodesOf, resetPreview, cachedBreadcrumbs] = useStore((state) => [
+    state.childNodesOf,
+    state.resetPreview,
+    state.cachedBreadcrumbs,
+  ]);
+
+  const handleSubmit = props.handleSubmit;
```
jessicamcinchak, line 84: Ideally, in the near future, we can/will re-introduce `isFinalCard` at the `Preview/Node` level, but we'll want to do more thorough performance reviews and add test coverage before doing this - so have adjusted this to keep current behavior in the meantime and instead _only_ reference `isFinalCard` within the Confirmation component itself rather than _any_ component

editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
```diff
@@ -48,12 +49,23 @@ const Card: React.FC<Props> = ({
...
     state.path,
     state.currentCard,
+    state.breadcrumbs,
+    state.flow,
   ]);
+
+  // Check if we have a Send node in our breadcrumbs
+  //   This is a better/more immediate proxy for "Submitted" because actual send events that populate lowcal_sessions.submitted_at are queued via Hasura
+  const hasSent = Object.keys(breadcrumbs)
+    .reverse()
```
DafyddLlyr, line 62: nit: As breadcrumb keys are sorted alphabetically not chronologically this doesn't offer any benefits (despite being a clever and nice optimisation!)

editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
DafyddLlyr: Very happy to see this commented back in 👍

editor.planx.uk/src/pages/Preview/Node.tsx
DafyddLlyr: Totally agree - this is the right approach here 👍

DafyddLlyr (approved):
PR description super helpful thanks. Great point about tests - would be super nice to see them added 👍 

The question around validation checks is a good one - 
 - We _probably_ don't want anything that sets a passport value coming after send (maybe this just needs to be an alert, people may want to do something after send around next steps etc)
 - We can't have Feedback be the final component of a service, as it only inserts a feedback record when "Continue" is hit

editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
jessicamcinchak: ah shoot yes always forget that alphabetical key storage ! I'll remove the `.reverse()` here and just call `.some()` directly (at least it stops at first find)

Very much assuming it's NOT worth the expense of sorting, then reversing, then calling `.some()` - but maybe one to revisit in future

jessicamcinchak:
[Allow Confirmation components to conditionally have a "Continue" button so that Feedback can be collected post-Confirmation](https://trello.com/c/bnQBMdFY/3173-allow-confirmation-components-to-conditionally-have-a-continue-button-so-that-feedback-can-be-collected-post-confirmation)

PR #4030, opened 2024-12-03
feat(fee-breakdown): Initial Editor UI

DafyddLlyr:
## What does this PR do?
- Updates the Editor UI to illustrate how the fee breakdown works
- Very open to suggestions here! To me something along these lines of "show not tell" is a bit more digestible / explainable and allows room for some rows to be added (e.g. service charge), or an input to be added (for configuration)
- Just UI behind a feature flag, no functional changes here

### Questions
I've hardcoded `application.fee` as a prefix here, but this is a dynamic key.

In reality we know that all active services (apart from one) currently pass through the osl/fee-calculator flow which rely on this variable.

Here's the question - should we keep this dynamic, or make the field disabled in the Pay component and always set to `application.fee` (after double checking services)? I think making it static leans more towards the "SetFee" component approach we discussed, but isn't without downsides. Maybe one for the dev call!

image

### DIFF
--- a/editor.planx.uk/src/@planx/components/Pay/Editor/FeeBreakdownSection.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Editor/FeeBreakdownSection.tsx
@@ -1,13 +1,134 @@
+import HelpIcon from "@mui/icons-material/Help";
 import ReceiptLongIcon from "@mui/icons-material/ReceiptLong";
+import Box from "@mui/material/Box";
+import Table from "@mui/material/Table";
+import TableBody from "@mui/material/TableBody";
+import TableCell, { tableCellClasses } from "@mui/material/TableCell";
+import TableContainer from "@mui/material/TableContainer";
+import TableHead from "@mui/material/TableHead";
+import TableRow from "@mui/material/TableRow";
+import Tooltip from "@mui/material/Tooltip";
+import Typography from "@mui/material/Typography";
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/Pay/Editor/FeeBreakdownSection.tsx
```diff
@@ -1,13 +1,134 @@
...
+          <TableCell align="right" sx={{ display: "flex", alignItems: "center"}}>
+            <HelpTooltip text="The fee will be the calculated amount (if set), or the payable amount" />
+            <DataField>£ application.fee.calculated</DataField>
+          </TableCell>
+        </TableRow>
+        <TableRow>
+          <TableCell>Reduction</TableCell>
+          <TableCell align="right" sx={{ display: "flex", alignItems: "center" }}>
+            <HelpTooltip text="This is the sum of all reductions applied to the application fee. The list of reductions will be listed below" />
+            <DataField>- £ (calculated - payable)</DataField>
```
DafyddLlyr, line 55: I'm not sure if this shorthand for `application.fee.calculated - application.fee.payable` is that clear?

The "full" version pretty much always overflows and feels less clear even though the above is not technically correct.

image

DafyddLlyr:
Closing in favour of a list based approach and not a illustrative table / fee breakdown. Will reference this PR as prior art!

PR #4032, opened 2024-12-03
fix: remove `alteredNodes` badge on publish button and `validateAndDiff` request on first flow load

jessicamcinchak:
I've noticed that loading flows in the editor feels particularly sluggish lately and leads to more errors in the Preview side panel than usual (and therefore Airbrake too) - eg `Error: id "eMxQAgygE6" not found`, `TypeError: Cannot read properties of undefined (reading 'data')` and so on.

This PR proposes removing the "badge" on the publish button which displays how many altered nodes are currently un-published. This badge requires calling `validateAndDiff` on initial flow load which is fairly expensive and regularly timesout (therefore badge fails to display at all).

Feels like it's an easy win to remove this feature for now (doubt it'll be noticed or missed), _only_ call `validateAndDiff` when the "Check for cahnges to publish" modal is actually opened and _not_ on first flow load as before, and re-visit this when we talk about publishing designs later this phase. Hopeful it'll free up more space for other requests on first flow load to run and ease a bit of noise in the errors channel :crossed_fingers:

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Publish/PublishFlowButton.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Publish/PublishFlowButton.tsx
@@ -1,4 +1,3 @@
-import Badge from "@mui/material/Badge";
 import Box from "@mui/material/Box";
 import Button from "@mui/material/Button";
 import Dialog from "@mui/material/Dialog";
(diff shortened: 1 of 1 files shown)


RODO94 (approved):
Looks good to me and agree it's a good change. Users can still review changes by clicking the "Check for changes to Publish" button, which could make it a more meaningful interaction as you'll get better feedback on the changes.

The thing we lose is the orange count badge on the button, and this means team members who click into a service that another colleague is working on, won't see if there's currently unpublished changes until they click the button?

Is the History tab a better place for this tracking / auditing to happen?

jessicamcinchak:
@RODO94 Because this request is timing out so often, the orange badge is commonly failing to load entirely so I'm quite certain no one is relying on it currently (never comes up in help-issues etc). 

"History" tab currently only covers _operations_, not publishing events - but what we can show on that timeline is definitely one to revisit in future publishing & templates design conversations.

DafyddLlyr (approved):

PR #4033, opened 2024-12-03
fix: `TypeError: Cannot read properties of undefined (reading 'data')` in `autoAnswerableFlag`

jessicamcinchak:
Flagged by Airbrake here: https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1733161837391839

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
@@ -505,8 +505,8 @@ export const previewStore: StateCreator<
     const node = flow[id];
     if (!node) return;
 
-    const { type, data, edges } = node;
     // Only Question & Checklist nodes that have an fn & edges are eligible for auto-answering
+    const { type, data, edges } = node;
     if (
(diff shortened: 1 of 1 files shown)


RODO94 (approved):
Code change looks good to me, although I wasn't able to recreate the bug to test it and I couldn't fully understand what Airbrake was telling me.

I could only get that it was happening in a Medway/Article4 service. How did you find the source?

jessicamcinchak:
@RODO94 I check Airbrake's "Backtrace" tab which will often include a function name (often helpful to use the left/right scroll in upper right to toggle through "Occurances" which have been grouped)
Screenshot from 2024-12-03 15-25-23

These are edge cases of likely-legacy nodes that don't have the same structure as new nodes created today so they're difficult to re-create in the editor directly. Instead by simply reading over the function code, I looked for any instances where `data` wasn't safely handled and fixed that directly (eg `const {type, data, edges} = flow[id]` would fail in instances where `flow[id] = undefined`, so now I'm first confirming that the node still exists in the flow before trying to destructure the node properties.)

PR #4034, opened 2024-12-03
fix: when next step is deleted, props.value doesn't update

RODO94:
Found that when you deleted a step from the NextSteps component, it wouldn't set `description` or `url` to "", but not set it at all. This meant previous values with the same index would persist until it is submitted then when it is reloaded, data would return to the formik saved values. 

Looked into trying to alter the data model, but got a bit caught up in the components and functions with the way they are passed down and back up. Open to suggestions on a more robust change, but this seemed to work.


jessicamcinchak (approved):
Thanks for picking this one up! Pizza working as expected for me, would be good to get this one merged & live to un-block content team if you and August are also happy with it :+1: 

In future, this bug/expected behavior also feels like a good candidate to write a unit test for (eg `NextSteps/Editor.test.ts` &rarr; `it("preserves existing steps values when one is deleted")`)

PR #4035, opened 2024-12-03
feat: E2E coverage for Draw Boundary

RODO94:
## What does this Test do?

For the Draw Boundary testing, I have focused only on:
-  the ability to add via the Editor modal
- path to navigate to the file upload route
- a map is present with correct geoJSON
- a map can be reset and new boundary drawn

I have left testing the file upload to the relevant unit tests or related testing elsewhere in the code base. I thought this was the easiest given we are yet to do the `FileUploadAndLabel` e2e tests. I can come back to add this in after that.

Rather than add in comments, in some parts I have just added custom text to the `expect()`. It made it easier for me to track when running the test and it was acting the same as the comments in the code. 

```js
    await expect(drawBoundaryTitle, "We have navigated back to the map component",).toBeVisible();
```

Screenshot 2024-12-10 at 16 45 01

### DIFF
--- a/e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
+++ b/e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
@@ -72,8 +72,7 @@ test.describe("Flow creation, publish and preview", () => {
     await editor.createInternalPortal();
     await editor.populateInternalPortal();
     await page.getByRole("link", { name: "start" }).click(); // return to main flow
-    await editor.createUploadAndLabel();
-    // TODO: editor.createPropertyInfo()
+    // await editor.createUploadAndLabel();
     await editor.createDrawBoundary();
     await editor.createPlanningConstraints();
     // await editor.createFileUpload();
@@ -81,7 +80,6 @@ test.describe("Flow creation, publish and preview", () => {
     await expect(editor.nodeList).toContainText([
       "Find property",
       "an internal portalEdit Portal",
-      "Upload and label",
       "Confirm your location plan",
       "Planning constraints",
       // "File upload",
@@ -169,7 +167,12 @@ test.describe("Flow creation, publish and preview", () => {
     ).toBeVisible();
     await clickContinue({ page });
 
+    await expect(page.getByRole('heading', { name: 'Confirm your location plan' })).toBeVisible()
+
+    await expect(page.getByTestId('upload-file-button')).toBeVisible()
+    await expect(page.getByText("490.37")).toBeVisible()
+
     // TODO: answer uploadAndLabel
-    // TODO: answerPropertyInfo, answerDrawBoundary, answerPlanningConstraints
+    // TODO: answerPropertyInfo, answerPlanningConstraints
   });
 });
--- a/e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
+++ b/e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
@@ -33,7 +33,10 @@ import {
   mockPropertyTypeOptions,
   mockTitleBoundaryGeoJson,
 } from "./mocks/geospatialMocks";
-import { setupOSMapsStyles, setupOSMapsVectorTiles } from "./mocks/osMapsResponse";
+import {
+  setupOSMapsStyles,
+  setupOSMapsVectorTiles,
+} from "./mocks/osMapsResponse";
 
 test.describe("Flow creation, publish and preview", () => {
   let context: TestContext = {
@@ -127,8 +130,8 @@ test.describe("Flow creation, publish and preview", () => {
       `/${context.team.slug}/${serviceProps.slug}/published?analytics=false`,
     );
 
-    setupOSMapsStyles(page)
-    setupOSMapsVectorTiles(page)
+    setupOSMapsStyles(page);
+    setupOSMapsVectorTiles(page);
 
     await expect(
       page.locator("h1", { hasText: "Find the property" }),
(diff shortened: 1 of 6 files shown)


RODO94:
[[E2E Tests] - Draw Boundary](https://trello.com/c/WhtNhVSN/3134-e2e-tests-draw-boundary)

e2e/tests/ui-driven/src/mocks/geospatialMocks.ts
```diff
@@ -37,3 +43,23 @@ export const mockMapGeoJson: Feature = {
     "organisation-entity": "13",
   },
 };
+
+export const mockChangedMapGeoJson: GeoJsonChangeHandler = {
```
RODO94, line 47: pulled this from the tests so I had a static mock to check against when running the tests

e2e/tests/ui-driven/src/mocks/geospatialMocks.ts
```diff
@@ -1,29 +1,35 @@
...
+      [-0.633455, 51.605606],
+      [-0.633788, 51.605643],
+      [-0.634429, 51.605799],
+      [-0.634429, 51.605767],
+      [-0.633498, 51.605485],
+    ],
+  ],
+];
 
 export const mockMapGeoJson: Feature = {
```
jessicamcinchak, line 30: nit: `mockTitleBoundaryGeoJson` would be a more meaningful name here, everything is GeoJSON with the map !

e2e/tests/ui-driven/src/mocks/geospatialMocks.ts
```diff
@@ -1,29 +1,35 @@
...
-          [-0.633788, 51.605643],
-          [-0.634429, 51.605799],
-          [-0.634429, 51.605767],
-          [-0.633498, 51.605485],
-        ],
-      ],
-    ],
-  },
   type: "Feature",
+  geometry: { type: "MultiPolygon", coordinates: mockCoordinates },
```
jessicamcinchak, line 32: nit: what's the point of extracting the `mockCoordinates` here, they don't seem to be referenced elsewhere?

e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
```diff
@@ -169,7 +176,51 @@ test.describe("Flow creation, publish and preview", () => {
     ).toBeVisible();
     await clickContinue({ page });
 
+    const drawBoundaryTitle = page.getByRole("heading", {
+      name: "Confirm your location plan",
+    });
+    await expect(
+      drawBoundaryTitle,
+      "We are in the Draw Boundary component",
+    ).toBeVisible();
```
jessicamcinchak, line 185: nit: `await expect(drawBoundaryTitle).toBeVisible();` is a very clear indicator that we're on the right page to me and I'm not sure I really understand the value of of the second `expect` argument here - feels quite noisy?

e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
```diff
@@ -169,7 +176,51 @@ test.describe("Flow creation, publish and preview", () => {
...
+
+    await checkGeoJsonContent(page, "drawgeojsondata", mockMapGeoJson);
+
+    const area = "490.37";
+
+    await expect(
+      page.getByText(area),
+      "We can see a value for area",
+    ).toBeVisible();
```
jessicamcinchak, line 194: nit: the component will have calculated the precise area in both metres squared & hectares, but we expect only the _rounded_ metres squared to be visible to the user - can this be simplified to:
```
const area = "The property boundary you have drawn is 490.37 m²",
await expect(page.getByText(area)).toBeVisible();
```
(Again `"We can see"` followed by `.toBeVisible()` feels very redundant to me!)

e2e/tests/ui-driven/src/helpers/geospatialChecks.ts
```diff
@@ -1,12 +1,83 @@
...
+    page.getByRole("heading", { name: "Upload a location plan" }),
+    "Should be in a page for uploading a file",
+  ).toBeVisible();
+
+  await expect(
+    page.getByRole("button", { name: "Drop file here or choose" }),
+    "A button for uploading files is visible",
+  ).toBeVisible();
+
+  const useMapButton = page.getByTestId("use-map-button");
```
jessicamcinchak, line 80: nit: before toggling right back to the map page, it might be additionally useful to click "Continue", confirm an error is visible that prevents you from proceeding without uploading, _then_ toggle back to map page. Error handling is important to check in my opinion!

jessicamcinchak (commented):
This looks like a solid start! Some initial comments below and a larger one: 
- I was expecting to find mocks for Ordnance Survey API calls as part of this work - when you load up the DrawBoundary here, is it making live calls to the OS Vector Tiles API ? Are OS mocks something you previously handled with FindProperty & PropertyInformation components before this ?

We definitely don't want our live OS APIs to be called repeated by E2E so let's catch up about this please to move this forward !

e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
RODO94: True, I do feel like it does add value with other checks where it may look like duplication. Like when we navigate back somewhere or when we're using a certain value in a certain place, but these do look a bit indulgent

jessicamcinchak (approved):
Thank you for adding mocks here, looks good to go to me :+1:

PR #4039, opened 2024-12-04
feat: Remove toggle for FeeBreakdown

DafyddLlyr:
Quick change following design call - this feature doesn't need to be toggleable. It's still behind a feature flag but once this is removed it will always be shown on the public facing Pay component.


jessicamcinchak (approved):
Nice simplification! Appreciate you taking in all our feedback at design call on this surprisingly complex one!

PR #4041, opened 2024-12-04
refactor: remove duplicate optionsEditor and move into shared component

jamdelion:
More refactoring! I noticed we had two large subcomponents that looked very similar - the `OptionsEditor` for checklists and questions. I've changed this so that there is one OptionsEditor component in the shared folder, that is used in both places with appropriate props. I think they work the same, but do shout if I've missed anything!

Not sure why it's not figuring out the diff to main correctly! The `jh/checklist-with-or` (a misnomer!) branch has been merged so the diff should be the same with main


editor.planx.uk/src/@planx/components/Question/Editor.tsx
```diff
@@ -247,7 +152,10 @@ export const Question: React.FC<Props> = (props) => {
               }) as Option
             }
             Editor={OptionEditor}
-            editorExtraProps={{ showValueField: !!formik.values.fn }}
+            editorExtraProps={{
+              showValueField: !!formik.values.fn,
+              showDescriptionField: true,
```
jamdelion, line 157: A new prop for toggling the description field - question options have it, while checklists do not

editor.planx.uk/src/@planx/components/shared/OptionsEditor.tsx
```diff
@@ -0,0 +1,125 @@
...
+              label: `Move to ${group || `group ${groupIndex}`}`,
+              onClick: () => {
+                props.onMoveToGroup &&
+                  typeof props.index === "number" &&
+                  props.onMoveToGroup(props.index, groupIndex);
+              },
+              disabled: groupIndex === props.groupIndex,
+            }))}
+          />
+        )}
```
jamdelion, line 88: This section comes from the original Checklist OptionsEditor

editor.planx.uk/src/@planx/components/shared/OptionsEditor.tsx
```diff
@@ -0,0 +1,125 @@
...
+          placeholder="Description"
+          multiline
+          onChange={(ev) => {
+            props.onChange({
+              ...props.value,
+              data: {
+                ...props.value.data,
+                description: ev.target.value,
+              },
+            });
```
jamdelion, line 70: This section comes from the Question OptionsEditor

jamdelion:
Realising now that I'm deeper into the ticket that these should be more separated than I thought. There are still common elements that I will extract out to reduce duplication

editor.planx.uk/src/@planx/components/Checklist/Editor/OptionsEditor.tsx
```diff
@@ -1,104 +1,46 @@
...
-              },
-            });
-          }}
+    <BaseOptionsEditor
+      value={value}
+      onChange={onChange}
+      showValueField={showValueField}
+    >
+      {typeof index !== "undefined" && groups && onMoveToGroup && (
+        <SimpleMenu
```
DafyddLlyr, line 31: The layout of this is not quite right on the Pizza, the menu has spilled over to the next row - https://4041.planx.pizza/bromley/article4/nodes/_root/nodes/wRCvCPZxVT/edit

image

DafyddLlyr (changes requested):
Thanks for tackling this - only one small issue I've spotted.

editor.planx.uk/src/@planx/components/Checklist/Editor/OptionsEditor.tsx
jamdelion: Good spot, thank you! Have just committed a fix for this :)

DafyddLlyr (approved):
Great!

PR #4042, opened 2024-12-05
feat(fee-breakdown): Calculate VAT using `useFeeBreakdown()`

DafyddLlyr:
## What does this PR do?
 - Changes logic to now rely on on `application.fee.includesVAT` (boolean) instead of `application.fee.vat` (number)
 - This means we are relying on Editors to tell us if the fee in inclusive of VAT, but does not rely on them to make the calculation themselves
 - Based on decision from Monday's fee breakdown show & tell

### Note
This doesn't cover the cases where teams require a VAT amount to be included within their GovUK Pay metadata. In these cases they will still need to calculate it as a passport variable, and add it to the metadata fields. It's probably worth getting teams to adopt this approach there as well but I'm intentionally avoiding this here in order to focus on the fee breakdown feature.


DafyddLlyr:
Closed - I'm going to port this along to planx-core shortly!

PR #4045, opened 2024-12-05
fix: account for expandable checklists in editor modal passport value validation

jessicamcinchak:
Flagged here https://opensystemslab.slack.com/archives/C01E3AC0C03/p1733408598813289

We don't currently have an `Editor.test.tsx` file for Checklists but happy to revisit that as a follow-up PR so we don't miss this again ! Probably nice to merge now though to minimise rebases for Jo :slightly_smiling_face:


jamdelion (approved):
Thanks for picking this up! ⭐

DafyddLlyr (approved):
Appreciate the quick fix here! 🙌

PR #4052, opened 2024-12-06
fix: add missing BOPS app type suffixes to Slack notification pilot criteria

jessicamcinchak:
Quick followup from this thread yesterday: https://opensystemslab.slack.com/archives/C0405SQDP8B/p1733407486961749

Should now add the apparently-very-relied-on :large_orange_diamond: to all BOPS app type alerts coming through #planx-notifications during the pilot testing

### DIFF
--- a/api.planx.uk/modules/webhooks/service/sendNotification/index.ts
+++ b/api.planx.uk/modules/webhooks/service/sendNotification/index.ts
@@ -36,7 +36,11 @@ export const sendSlackNotification = async (
   const pilotServices = [
     "uniform",
     "happ",
+    "hapr",
+    "hret",
     "ldc",
+    "minor",
+    "major",
     "apply for planning permission",
     "apply for a lawful development certificate",
   ];


api.planx.uk/modules/webhooks/service/sendNotification/index.ts
```diff
@@ -36,7 +36,11 @@ export const sendSlackNotification = async (
   const pilotServices = [
     "uniform",
     "happ",
+    "hapr",
+    "hret",
```
jessicamcinchak, line 40: `HRET` is what actually came through Slack, but `HAPR` also looks like a valid application type "suffix" in BOPS codebase so cautiously accounting for both! 
- https://github.com/unboxed/bops/blob/main/spec/factories/application_type.rb
- https://github.com/unboxed/bops/blob/main/db/migrate/20240321155240_add_householder_retrospective_type.rb#L14

api.planx.uk/modules/webhooks/service/sendNotification/index.ts
```diff
@@ -36,7 +36,11 @@ export const sendSlackNotification = async (
...
     "ldc",
+    "minor",
+    "major",
```
jessicamcinchak, line 43: again, I don't _think_ minors & majors are actually live in pilot, but just in case one is sent to BOPS we'll now properly flag it with the :large_orange_diamond:

api.planx.uk/modules/webhooks/service/sendNotification/index.ts
DafyddLlyr: 🕵️‍♀️

DafyddLlyr (approved):

PR #4053, opened 2024-12-06
fix: correct editor link validation

augustlindemer:
Tightens editor link validation from any link text including ''click'' or ''here'' to links set solely over ''click here'' or ''clicking here''. Includes trimming of white space and punctuation.

### DIFF
--- a/editor.planx.uk/src/ui/editor/RichTextInput/RichTextInput.tsx
+++ b/editor.planx.uk/src/ui/editor/RichTextInput/RichTextInput.tsx
@@ -200,8 +200,8 @@ export const emptyContent = "<p></p>";
 const linkSelectionError = (selectionHtml: string): string | null => {
   if (selectionHtml.startsWith("<p>") && selectionHtml.endsWith("</p>")) {
     const text = selectionHtml.slice(3, -4);
-    const lowercaseText = text.toLowerCase();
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/ui/editor/RichTextInput/RichTextInput.tsx
```diff
@@ -200,8 +200,8 @@ export const emptyContent = "<p></p>";
 const linkSelectionError = (selectionHtml: string): string | null => {
   if (selectionHtml.startsWith("<p>") && selectionHtml.endsWith("</p>")) {
     const text = selectionHtml.slice(3, -4);
-    const lowercaseText = text.toLowerCase();
-    if (lowercaseText.includes("click") || lowercaseText.includes("here")) {
+    const lowercaseText = text.toLowerCase().trim().replace(/[.,]/g, "");
+    if (lowercaseText == "click here" || lowercaseText == "clicking here") {
```
DafyddLlyr, line 204: nit: We should use `===` here instead of `==` ([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality))

DafyddLlyr (approved):
Great fix! 🏅 

This would currently be catching things like "atmosphere"  or "sphere" 🤦

PR #4054, opened 2024-12-07
fix: Convert Digital Planning payload to JSON file before upload

DafyddLlyr:
## What does this PR do?
 - Convert a `DigitalPlanningPayload` to an `Express.Multer.File` for upload
 - Addresses issues discussed here - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1733474436178079 (OSL Slack)

### Comments
Trickier than expected! TypeScript isn't catching the mismatch between `DigitalPlanningPayload` and `Express.Multer.File` even when the type is imported and assigned. I think it's to do with the combination of the type complexity plus the `Awaited<>` utility type which has a depth limit I believe.

Next up is some tests here for the entire module (in another PR), will also check the Pizza to make this is working as expected once it's up.

### DIFF
--- a/api.planx.uk/modules/file/service/uploadFile.ts
+++ b/api.planx.uk/modules/file/service/uploadFile.ts
@@ -7,7 +7,6 @@ import mime from "mime";
 import { customAlphabet } from "nanoid";
 import { isLiveEnv } from "../../../helpers.js";
 import { s3Factory } from "./utils.js";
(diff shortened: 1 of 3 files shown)


jessicamcinchak (approved):
Thanks for picking this up :+1: 

Pizza working as expected & JSON file is populated: https://4054.planx.pizza/barnet/test/submissions-log

PR #4057, opened 2024-12-09
feat(fee-breakdown): Display for "invite to pay" journeys

DafyddLlyr:


editor.planx.uk/src/pages/Pay/MakePayment.tsx
```diff
@@ -225,14 +227,19 @@ export default function MakePayment({
...
+            </Container>
+            <Confirm
+              fee={toDecimal(paymentAmount)}
+              onConfirm={readyAction}
+              buttonTitle={currentState.button!}
+              showInviteToPay={false}
+              hideFeeBanner={true}
+              paymentStatus={payment?.state.status}
+            />
+          </>
```
jessicamcinchak, line 242: nit: I think this should be feature-flagged ? I can see a breakdown here with & without feature flag: https://4057.planx.pizza/barnet/invite-to-pay-test/pay?paymentRequestId=f0babfb3-4a57-4c01-bdc2-cd344bac865c

Also, minor style nitpick: should we remove "Fee" line from top table, and then simply call bottom part "Fee" instead of "Fee breakdown" ? 
Screenshot from 2024-12-09 16-46-37

hasura.planx.uk/migrations/1733686124805_alter_table_public_payment_requests_add_column_fee_breakdown/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."payment_requests" add column "fee_breakdown" jsonb
+ null;
```
jessicamcinchak, line 2: :+1:

editor.planx.uk/src/pages/Pay/MakePayment.tsx
jessicamcinchak: Also, do we need to account for any active invitations that won't have saved breakdown details when this merges? Will the breakdown simply not display/be hidden?

DafyddLlyr: Yep - feature flag should be here thanks! 

Changed from "Fee breakdown" to "Fee".

This will just not show at all if there's no fee breakdown, so we should be ok for that transition period 👍 

I'd love a pass over this page by Ian at some point also - the two tables, one horizontal and one vertical just feel a bit weird back to back here.

jessicamcinchak (approved):

PR #4058, opened 2024-12-09
feat(fee-breakdown): Make Pay component data field a static value

DafyddLlyr:
## What?
Makes the `data.fn` value of the Pay component a hard-coded variable - `application.fee.payable`

## Why?
This is relied upon when working out the fee breakdown, as well as in the payload mapping. All real services are using this value and this matches the pattern for other variable like addresses where we need a single static passport key.

### DIFF
--- a/editor.planx.uk/src/@planx/components/Pay/Editor/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Editor/Editor.tsx
@@ -1,5 +1,5 @@
 import { ComponentType as TYPES } from "@opensystemslab/planx-core/types";
-import { parsePay, Pay, validationSchema } from "@planx/components/Pay/model";
+import { parsePay, PASSPORT_FN, Pay, validationSchema } from "@planx/components/Pay/model";
 import { Form, Formik } from "formik";
 import React from "react";
 import { ComponentTagSelect } from "ui/editor/ComponentTagSelect";
@@ -69,9 +69,9 @@ const Component: React.FC<Props> = (props: Props) => {
                 <Input
                   format="data"
                   name="fn"
-                  value={values.fn}
+                  value={PASSPORT_FN}
                   placeholder="Data field for calculated amount"
-                  onChange={handleChange}
+                  disabled
                 />
               </InputRow>
             </ModalSectionContent>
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/@planx/components/Pay/Editor/Editor.tsx
```diff
@@ -69,9 +69,9 @@ const Component: React.FC<Props> = (props: Props) => {
                 <Input
                   format="data"
                   name="fn"
-                  value={values.fn}
+                  value={PASSPORT_FN}
```
DafyddLlyr, line 72: This should mean that any existing services with a value other and `application.fee.payable` just need the node to be saved to update this.

In reality, these would just be test services so I don't think there's a need for anything more sophisticated here such as a validation check or data migration.

editor.planx.uk/src/@planx/components/Pay/Editor/Editor.tsx
```diff
@@ -69,9 +69,9 @@ const Component: React.FC<Props> = (props: Props) => {
...
                   placeholder="Data field for calculated amount"
```
jessicamcinchak, line 73: nit: can we also remove this confusing placeholder (eg "calculated" not "payable" language) since will never be seen now?

editor.planx.uk/src/@planx/components/Pay/model.ts
```diff
@@ -10,12 +10,14 @@ import { array, boolean, object, string } from "yup";
 
 import { type BaseNodeData, parseBaseNodeData } from "../shared";
 
+export const PASSPORT_FN = "application.fee.payable" as const;
```
jessicamcinchak, line 13: nit: can we rename to `PAY_FN` so more clear if/when exported/imported beyond this component? (Eg tests, etc). Already have a few `DEFAULT_FN` etc that I notice that we're commonly re-naming when importing elsewhere.

editor.planx.uk/src/@planx/components/Pay/Editor/Editor.tsx
jessicamcinchak: Confirmed that a legacy, un-updated Pay component still knows how to read/display correct amount: https://4058.planx.pizza/testing/pay-test :+1: 

And modal looks correct for next "Update"/"Save"

jessicamcinchak (approved):
In support of this :raised_hands:

editor.planx.uk/src/@planx/components/Pay/Editor/Editor.tsx
DafyddLlyr: Good catch! Removed 🗑️

PR #4062, opened 2024-12-10
feat: autocomplete data fields in the editor

jessicamcinchak:


editor.planx.uk/src/@planx/components/shared/DataFieldAutocomplete.tsx
```diff
@@ -0,0 +1,99 @@
...
+        getOptionLabel={(option) => {
+          let formattedOption = option;
+          // If a new option was added, strip out the value between ""
+          if (option.startsWith('Add "')) {
+            formattedOption = option.split('"')[1];
+          }
+          return formattedOption;
+        }}
+        renderOption={renderOptions}
+        freeSolo
```
jessicamcinchak, line 91: Explicitly passing the `freeSolo` prop is a bit strange because it's already handled in `AutocompleteProps<string, false, false, true, "div">`, but explicitly passing here seems to be the Stack Overflow consensus for how to clear up a console error related to `isOptionEqualToValue` which assumes "new" options will never be added/present

editor.planx.uk/src/@planx/components/shared/DataFieldAutocomplete.tsx
```diff
@@ -0,0 +1,99 @@
...
+    }
+  };
+
+  return (
+    <InputRow>
+      <AutocompleteInput
+        id="data-field-autocomplete"
+        key="data-field-autocomplete"
+        placeholder="Data field"
+        required={Boolean(props.required)}
```
jessicamcinchak, line 68: `required` will simply trigger browser default error handling which is quite consistent with current editor inputs ! Gov UK-looking error wrappers for autocomplete are larger future thing to revisit (eg would we simply wrap in our existing wrapper or account for new `error`-related props here?)
Screenshot from 2024-12-18 10-13-06

editor.planx.uk/src/ui/shared/AutocompleteInput.tsx
```diff
@@ -0,0 +1,130 @@
...
+    borderRadius: 0,
+    border: `1px solid${theme.palette.border.main} !important`,
+  },
+  "& fieldset": {
+    borderColor: theme.palette.border.main,
+  },
+  backgroundColor: theme.palette.background.paper,
+  [`& .${outlinedInputClasses.root}, input`]: {
+    cursor: "pointer",
+    // TODO extract as `format="data"` prop more like `Input` ?
```
jessicamcinchak, line 77: This `TODO` is good to note, but I don't think we need to pick it up until there's another use case/iteration of the autocomplete input - any objections?

editor.planx.uk/src/@planx/components/List/Editor.tsx
```diff
@@ -92,6 +94,8 @@ function ListComponent(props: Props) {
     validateOnChange: false,
   });
 
+  const dataFieldSchema = useStore().getFlowSchema()?.nodes;
```
jessicamcinchak, line 97: `List`, `Page` & `MapAndLabel` components have their own pre-existing concepts of "schemas", so the store function is more explicitly assigned to `dataFieldSchema` in these cases

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -594,11 +596,41 @@ export const editorStore: StateCreator<
...
+
+    Object.entries(flow).map(([_id, node]) => {
+      if (node.data?.fn) {
+        // Exclude Filter fn value as not exposed to editors
+        if (node.data?.fn !== "flag") nodes.add(node.data.fn)
+      };
+  
+      // TODO align to (reuse?) data facets from search
+      if (node.data?.output) nodes.add(node.data.output);
+      if (node.data?.dataFieldBoundary) nodes.add(node.data.dataFieldBoundary);
```
jessicamcinchak, line 622: This TODO goes away with #4083 & #4085

e2e/tests/ui-driven/src/helpers/addComponent.ts
```diff
@@ -55,11 +55,11 @@ const createBaseComponent = async (
       break;
     case ComponentType.AddressInput:
       await page.getByPlaceholder("Title").fill(title || "");
-      await page.getByPlaceholder("Data Field").fill(options?.[0] || "");
+      await page.locator("id=data-field-autocomplete").fill(options?.[0] || "");
```
jessicamcinchak: E2E still are ultimately failing because nodes with a `required` data field are not being correctly populated -I've tried a couple iterations here, but still not landed on the right solution. 

Opening for wider review in the meantime and any pointers very welcome! 

Here's what I've tried so far:
- The placeholder text has changed from `"Data Field"` to `"Data field"`, but `getByPlaceholder()` will now match more than one element (MUI Autocompletes have many nested elements)
Screenshot from 2024-12-18 21-46-20
- So, `"getByRole("combobox", { name: "Data field" })` _should_ exclusively be matching the `input` element
- On the first component, there will be no flow schema or suggestions yet and the `popupIcon` arrow is disabled, so there's no dropdown of options to `.click()` or open/expand - but I'm guessing I'm still using the wrong combo of `.fill`/`.click` here to mimic typing input ??

e2e/tests/ui-driven/src/helpers/addComponent.ts
```diff
@@ -55,11 +55,15 @@ const createBaseComponent = async (
       break;
     case ComponentType.AddressInput:
       await page.getByPlaceholder("Title").fill(title || "");
-      await page.getByPlaceholder("Data Field").fill(options?.[0] || "");
+      await page
+        .getByRole("combobox", { name: "Data field" })
+        .fill(options?.[0] || "");
```
jessicamcinchak, line 60: E2E still are ultimately failing because nodes with a `required` data field are not being correctly populated -I've tried a couple iterations here, but still not landed on the right solution. 

Opening for wider review in the meantime and any pointers very welcome! 

Here's what I've tried so far:
- The placeholder text has changed from `"Data Field"` to `"Data field"`, but `getByPlaceholder()` will now match more than one element (MUI Autocompletes have many nested elements)
Screenshot from 2024-12-18 21-46-20
- So, `"getByRole("combobox", { name: "Data field" })` _should_ exclusively be matching the `input` element
- On the first component, there will be no flow schema or suggestions yet and the `popupIcon` arrow is disabled, so there's no dropdown of options to `.click()` or open/expand - but I'm guessing I'm still using the wrong combo of `.fill`/`.click` here to mimic typing input ??

editor.planx.uk/src/@planx/components/Checklist/Editor/Options.tsx
```diff
@@ -28,6 +30,10 @@ export const Options: React.FC<{ formik: FormikHookReturn }> = ({ formik }) => {
 
   const exclusiveOrOptionManagerShouldRender =
     hasFeatureFlag("EXCLUSIVE_OR") && nonExclusiveOptions.length;
+  
+  const schema = useStore().getFlowSchema()?.options;
+  const initialOptions: Option[] | undefined = formik.initialValues.options || formik.initialValues.groupedOptions?.map((group: Group<Option>) => group.children)?.flat();
+  const initialOptionVals = initialOptions?.map((option) => option.data?.val);
```
jessicamcinchak, line 36: This is a bit of a lazy rebase right now while the exclusive-or work is still in-progress - I think `initialOptions` _should_ be able to be simplified/derived from `exclusiveOptions` or `nonExclusiveOptions` once those are handling "grouped" options

editor.planx.uk/src/@planx/components/shared/utils.ts
```diff
@@ -63,3 +64,21 @@ export const getPreviouslySubmittedData = ({
 
   return data;
 };
+
+export const getOptionsSchemaByFn = (fn?: string, defaultOptionsSchema?: string[], initialOptions?: (string | undefined)[]) => {
```
DafyddLlyr, line 68: Super clear and easy to follow 👍

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -594,11 +596,41 @@ export const editorStore: StateCreator<
...
+    const nodes: Set<string> = new Set();
+    const options: Set<string> = new Set();
+
+    Object.entries(flow).map(([_id, node]) => {
+      if (node.data?.fn) {
+        // Exclude Filter fn value as not exposed to editors
+        if (node.data?.fn !== "flag") nodes.add(node.data.fn)
+      };
+  
+      // TODO align to (reuse?) data facets from search
```
DafyddLlyr, line 620: Great comment!

We could probably use `DATA_FACETS` here (filtered to strings only) and then use this to get the data values from each node.

However, I think this is actually a more complex solution - there's a very different use case here and we'll have just as many exceptions and clauses. This works really well for now - we just need to make sure that we keep data values for new components, this function, and `DATA_FACETS` in sync.

DafyddLlyr (approved):
Great! Zero issues - all working as expected, code was clear and easy to follow.

Appreciate the detailed PR description and thought that's gone into this - great feature 😄

DafyddLlyr:
P.S. Happy to have a crack at Playwright later today and push a commit here 👍

e2e/tests/ui-driven/src/helpers/addComponent.ts
DafyddLlyr: This is working for me locally - [5114475](https://github.com/theopensystemslab/planx-new/pull/4062/commits/5114475b7486bcd2f0f004b1fd6e89f6f4371a34)

Running on CI now 🤞

DafyddLlyr: It failed on some still - to be fair I didn't run them all locally just the first few. Will pick this up again but I think adding the `.click()` is what we're after.

jessicamcinchak: Appreciate second eyes on this - it's a tedious tricky one! Once one node with a data field is successfully added, then future ones _will_ have suggested schema options - so might require combination of `fill` & `click` ?

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
jessicamcinchak: Yep - `fn` only basically felt like right "first" go at this since other `DATA_FACETS` like FileUploadAndLabel individual file fns are getting "overwritten" by ODP Schema enum suggestions anyways ! So it's a good question of whether or not we actually always want this to be perfectly in sync !

editor.planx.uk/src/@planx/components/shared/DataFieldAutocomplete.tsx
```diff
@@ -0,0 +1,99 @@
...
+  createFilterOptions
+} from "@mui/material/Autocomplete";
+import ListItem from "@mui/material/ListItem";
+import isNull from "lodash/isNull";
+import React, { useMemo } from "react";
+import AutocompleteInput from "ui/shared/AutocompleteInput";
+import InputRow from "ui/shared/InputRow";
+
+interface Props {
+  schema?: string[];
```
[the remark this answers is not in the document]
jessicamcinchak: :bulb: Good suggestion! Will aim to push this change up this afternoon

e2e/tests/ui-driven/src/helpers/addComponent.ts
jessicamcinchak: Sorted these all out now per [b94e92a](https://github.com/theopensystemslab/planx-new/pull/4062/commits/b94e92a711a114017009045089d3a88675242a3f)

It's been a minute since I've been deep in e2e tests and a couple observations worth noting to revisit in the future: 
- I noticed that our "create component" helper methods don't throw an error in the "create flow" tests if a certain component type fails to be created/added to the graph, rather the "view a published flow" test will fail to find the public-representation of that component - this made troubleshooting rather confusing initially!
- It seems quite common that ui-driven e2e failures are bringing ShareDB down in test containers - for example only the first attempt error is meaningful, and the retries will fail to load the page at all and the trace simply shows an infinite "Loading" spinner consistent with ShareDB errors - again making troubleshooting quite hard, have to rely on error log rather than trace !

PR #4063, opened 2024-12-10
feat: E2E Testing for Planning Constraints

RODO94:
## What I want to test

- Planning Constraints can be added for checking all constraints
- Planning Constraints can be added for checking some constraints
- It displays correct constraints on public side
- A user can click to challenge a constraint

I am not going to check UI elements like colours and accordion functionality, as I think this is what Storybook is better at, I'll focus on functionality.


RODO94:
[[E2E Tests] - Planning Constraints](https://trello.com/c/Nwth9Bpb/3169-e2e-tests-planning-constraints)

e2e/tests/ui-driven/src/helpers/context.ts
```diff
@@ -50,6 +50,24 @@ export async function setUpTestContext(
         submissionEmail: context.team.settings?.submissionEmail,
       },
     });
+    const integrations = await $admin.client.request(
```
RODO94, line 53: Couldn't find a better way to switch `has_planning_data` to `true` - I'll probably move the mutation out and add it to the `beforeAll` for the geospatial stuff since it's isolated to those tests

e2e/tests/ui-driven/src/mocks/geospatialMocks.ts
```diff
@@ -37,3 +37,165 @@ export const mockMapGeoJson: Feature = {
     "organisation-entity": "13",
   },
 };
+
+export const mockPropertyConstraints = {
```
RODO94, line 41: Seemed like an overzealous mock, but seemed the past way to ensure we get all the data pulled in that we need for showing constraints. Happy to refine if we think it is required

e2e/tests/ui-driven/src/helpers/context.ts
```diff
@@ -50,6 +50,26 @@ export async function setUpTestContext(
         submissionEmail: context.team.settings?.submissionEmail,
       },
     });
+
+    // has_planning_data needs to be altered manually for geospatial tests
+    await $admin.client.request(
+      gql`
+        mutation AddPlanningData($teamId: Int) {
```
jessicamcinchak, line 57: nit: I often associate "Add" with a database "insert", but this is an update - would `TogglePlanningData` be more accurate?

e2e/tests/ui-driven/src/mocks/geospatialMocks.ts
jessicamcinchak: I don't think this is a complete mock - I would expect it to additionally include `metadata` from Planning Data (like you've included in the roads mock)

e2e/tests/ui-driven/src/helpers/geoSpatialUserActions.ts
```diff
@@ -0,0 +1,33 @@
...
+export async function answerFindProperty(page: Page) {
+  await setupOSMockResponse(page);
+  await page.getByLabel("Postcode").fill("SW1 1AA");
+  await page.getByLabel("Select an address").click();
+  await page.getByRole("option").first().click();
+}
+
+export const userChallengesPlanningConstraint = async (page: Page) => {
+  const thisDoesNotApplyConstraintButton = page.getByRole("button", {
+    name: "I don't think this constraint",
```
jessicamcinchak, line 13: nit: can we check the full button text here / this substring is a bit confusing? Eg `"I don't think this constraint applies to this property`"

e2e/tests/ui-driven/src/helpers/geoSpatialUserActions.ts
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  await thisDoesNotApplyConstraintButton.click();
+
+  const constraintDoesNotApplyDialog = page.getByRole("heading", {
+    name: "I don't think this constraint",
+  });
+  constraintDoesNotApplyDialog.isVisible();
+
+  const noAddressSuppliedButton = page.getByTestId("42103309");
```
jessicamcinchak, line 23: nit: can we add/use a human-readable `data-testid` in the frontend then reference here to avoid this?

e2e/tests/ui-driven/src/mocks/geospatialMocks.ts
```diff
@@ -37,3 +37,165 @@ export const mockMapGeoJson: Feature = {
...
+  sourceRequest:
```
jessicamcinchak, line 42: This `sourceRequest` is not accurately mocked - eg it includes `dataset=conservation-area` despite the later check that `!splitUrl?.includes("designated.conservationArea")` ? 

Did you simply fail to manually mock this or did you copy this from the frontend which then might point to a bug?

e2e/tests/ui-driven/src/helpers/geoSpatialUserActions.ts
RODO94: Great catch, added a string infront to add some context

e2e/tests/ui-driven/src/mocks/propertyConstraintMetadata.json
RODO94, line 1: Separated this from the `geospatialMocks` object as it was quite long and not all of it was super relevant. Felt like the right way to keep things concise

e2e/tests/ui-driven/src/helpers/userActions.ts
```diff
@@ -194,13 +193,6 @@ export async function submitCardDetails(page: Page) {
   await page.locator("#confirm").click();
 }
 
-export async function answerFindProperty(page: Page) {
```
RODO94, line 197: Moved this to a separate `geoSpatialUserActions` doc to fit it better with the other ones

jessicamcinchak (approved):

PR #4067, opened 2024-12-11
feat: Display `Pay` component when fee is zero

DafyddLlyr:
## What does this PR do?
- Refactors `Confirm` component to more explicitly handle various states
  - Before this PR, the logic for information only, pay, and invite to pay was quite tricky to manage and understand - adding a "zero fee" option into the mix would add to this confusion.
- Renders `Confirm` in instances where fee is £0
- Removes fee breakdown title and description - if we find we need these we can add them back. There's already a clear title and simple description here.
- Pulls in changes made in https://github.com/theopensystemslab/planx-core/pull/637 which also renders any exemptions

There's still scope here for a content pass through which I'm going to address once this feature is technically complete (e.g. the title "pay for your application" is not very meaningful when we know the fee is £0.)

Notion documentation to sit alongside this feature can be found here - https://www.notion.so/opensystemslab/How-displaying-the-fee-breakdown-works-19035d469ad180e3bcd2eab977a39a56

### DIFF
--- a/editor.planx.uk/src/@planx/components/Pay/Public/Confirm.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Public/Confirm.tsx
@@ -55,6 +55,10 @@ const PayText = styled(Box)(({ theme }) => ({
   },
 }));
 
+/**
+ * Displayed when component is in an error state
+ * Relies on props.error being set
+ */
 const Error: React.FC<Props> = ({ onConfirm, error }) => {
   if (error?.startsWith(PAY_API_ERROR_UNSUPPORTED_TEAM)) {
     return (
@@ -79,6 +83,10 @@ const Error: React.FC<Props> = ({ onConfirm, error }) => {
   );
 };
 
+/**
+ * Main functionality
+ * Allows users to make a payment via GovUK Pay
+ */
 const PayBody: React.FC<PayBodyProps> = (props) => {
   const defaults = getDefaultContent();
 
@@ -117,6 +125,10 @@ const PayBody: React.FC<PayBodyProps> = (props) => {
   );
 };
 
+/**
+ * Display information only - does not allow users to pay
+ * Generally used at the end of guidance services as an illustrative example of what you could pay
+ */
 const InformationOnly: React.FC<Props> = (props) => {
   const defaults = getDefaultContent();
 
@@ -143,6 +155,10 @@ const InformationOnly: React.FC<Props> = (props) => {
   );
 };
 
+/**
+ * Displayed if the fee is calculated as £0
+ * Still displays component and fee breakdown, but allows user to continue without making a payment
+ */
 const ZeroFee: React.FC<Props> = (props) => (
   <Button
     variant="contained"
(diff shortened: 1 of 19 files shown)


jessicamcinchak (approved):
Overall all working as expected & now displaying when fee = 0 :heavy_check_mark: 

Spent quite a bit of time on the pizza and a few minor nit pick comments that I don't believe are introduced here:
- Should we capitalize the first letter of the type of reduction or exemption for consistency? (in interim until fetching from schema `@description`?)
- I was surprised to get this breakdown (reductions "-129", exemptions "-0") with this passport - I would expect no line item for reductions, and "-129" for exemptions?
Screenshot from 2024-12-16 10-43-12
Screenshot from 2024-12-16 10-43-49

Was testing on this flow: https://4067.planx.pizza/barnet/pay-testing

editor.planx.uk/src/@planx/components/Pay/Public/Confirm.tsx
```diff
@@ -138,59 +199,68 @@ export default function Confirm(props: Props) {
...
+        <PayText>
+          {
+            {
+              pay: <PayBody {...props} changePage={toggleToInviteToPayPage} />,
+              informationOnly: <InformationOnly {...props} />,
+              inviteToPay: <InviteToPayForm {...inviteToPayFormProps} />,
+              error: <Error {...props} />,
+              zeroFee: <ZeroFee {...props} />,
+            }[componentState]
+          }
```
jessicamcinchak, line 260: This is a really nice pattern & improvement :+1:

editor.planx.uk/src/@planx/components/Pay/Public/Pay.test.tsx
```diff
@@ -112,14 +133,56 @@ describe("Pay component when fee is undefined or £0", () => {
...
+        data: {
+          "application.fee.payable": ["-12"],
+        },
+      },
+    };
+
+    setState({
+      flow: flowWithNegativeFee,
+      breadcrumbs: negativeFeeBreadcrumbs,
+    });
```
jessicamcinchak, line 180: Question: rather than auto-answer/skip in this scenario - why wouldn't we display (and log) an explicit error like "Something has gone wrong calculating your application fee" ? Similar to how we currently do for the `"undefined_fee"` state?  

Is there a risk to auto-answering, submitting, then the council having to manually prompt for payment and the applicant being oblivious to an issue? 

The error page could very well still have an enabled "Continue" button still, but at the least the applicant would be aware something has gone wrong and that they should _expect_ to be contacted about paying after the fact?

jessicamcinchak (approved):
Okay spent a good amount of time with this pizza and code changes (and associated planx-core PR and the Notion page) - it's all looking very solid ! Pay component states feel more managable than ever, thanks for some nice refactors here :gem: 

I definitely think the best way forwards is to get lots more eyes testing this, especially services team to shake out any edge case journeys through the fee calculator that might be blindspots for us.

Comments aren't blocking, but I do have one rather significant one which is that I think we should avoid auto-answering "Pay" in "negative" fee scenarios (or really _all_ scenarios now!), and instead treat it like an "undefined" fee (display & log an error that something's went wrong, applicant has option to proceed at their own risk).

editor.planx.uk/src/@planx/components/Pay/Public/Pay.test.tsx
jessicamcinchak: In other words, should the Pay component _ever_ be auto-answered now?

DafyddLlyr: This is a great point - I've now updated this to behave the same as an undefined fee.

PR #4071, opened 2024-12-11
feat: analytics module gets and updates PlanX DB

zz-hh-aa:
# What's new in this PR
- Migration to add `metabase_id` column to `teams` in PlanX DB
- Update Hasura permissions for `api` to be able to read and update relevant values
- A `getTeamIdAndMetabaseId` function
- An `updateMetabaseId ` function
- Tests for the above

# Why
- Daf suggested that a better pattern than querying Metabase every time we want to check if a collection exists would be to store a `metabase_id` in the PlanX DB and check that instead
- This is the PlanX DB portion of big PR #3971

### DIFF
--- a/api.planx.uk/modules/analytics/metabase/collection/getTeamIdAndMetabaseId.ts
+++ b/api.planx.uk/modules/analytics/metabase/collection/getTeamIdAndMetabaseId.ts
@@ -1,0 +1,38 @@
+import { gql } from "graphql-request";
+import { $api } from "../../../../client/index.js";
+
+interface GetMetabaseId {
+  teams: {
+    id: number;
+    name: string;
+    metabaseId: number | null;
+  }[];
+}
+
+export const getTeamIdAndMetabaseId = async (name: string) => {
+  try {
+    const response = await $api.client.request<GetMetabaseId>(
+      gql`
+        query GetTeamAndMetabaseId($name: String!) {
+          teams(where: { name: { _ilike: $name } }) {
+            id
+            name
+            metabaseId: metabase_id
+          }
+        }
+      `,
+      {
+        name: name,
+      },
+    );
+
+    const result = response.teams[0];
+    return result;
+  } catch (e) {
+    console.error(
+      "Error fetching team's ID / Metabase ID from PlanX DB:",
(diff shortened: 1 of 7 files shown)


jamdelion (commented):
Just hopping on as I realised I'd forgotten to review these for you! This is looking pretty good to me, though we are getting an integration test failure:

```
 FAIL  teams.test.js > teams > api > cannot create, update, or delete teams
Error: 'teams' unexpectedly found in mutation(s) [update_teams,update_teams_by_pk,update_teams_many]
 ❯ teams.test.js:102:17
    100| 
    101|     test("cannot create, update, or delete teams", () => {
    102|       expect(i).toHaveNoMutationsFor("teams");
       |                 ^
    103|     });
    104|   });
```

This is because we have added a mutation (update metabase ID in the teams table), where previously the api role did not have any mutations. 

I guess we should update the test to instead expect the mutation (similar to the `can query teams` test in the same file)

api.planx.uk/modules/analytics/metabase/collection/getTeamIdAndMetabaseId.ts
```diff
@@ -0,0 +1,38 @@
...
+    metabaseId: number | null;
+  }[];
+}
+
+export const getTeamIdAndMetabaseId = async (name: string) => {
+  try {
+    const response = await $api.client.request<GetMetabaseId>(
+      gql`
+        query GetTeamAndMetabaseId($name: String!) {
+          teams(where: { name: { _ilike: $name } }) {
```
DafyddLlyr, line 17: `_ilike` _could_ match both "North Somerset" and "South Somerset" - it would be better to match using `_eq` on team slugs if possible which are guaranteed to be unique.

DafyddLlyr (changes requested):

api.planx.uk/modules/analytics/metabase/collection/getTeamIdAndMetabaseId.ts
zz-hh-aa: Great catch! Updated.

DafyddLlyr (approved):

PR #4072, opened 2024-12-11
feat: Metabase module controller and service

zz-hh-aa:
# What does this PR do
- Creates `createCollectionIfDoesNotExist` service
- Creates `createCollection` Metabase API call
- Adds necessary types
- Adds necessary tests
- Creates new route for analytics module
- Fixes an import that was missing from the Metabase client (in order to get this working) [Maybe this should have gone in another PR..?]

# Why
This is the second PR that breaks up #3971 (follows #4071)

### DIFF
--- a/api.planx.uk/modules/analytics/docs.yaml
+++ b/api.planx.uk/modules/analytics/docs.yaml
@@ -14,8 +14,8 @@ components:
   schemas:
     NewCollection:
       type: object
-      properties:  
-        description:  
+      properties:
+        description:
           type: string
           description: Optional description for the collection
         parentId:
@@ -43,19 +43,19 @@ paths:
         "204":
           $ref: "#/components/responses/AnalyticsResponse"
 
-  /metabase/collection/{slug}:  
+  /metabase/collection/{slug}:
     post:
       summary: Create new Metabase collection
       description: Creates a new collection in Metabase if it doesn't already exist
       tags:
         - metabase
-      parameters:  
-        - name: slug  
-          in: path  
-          required: true  
-          schema:  
-            type: string  
-          description: PlanX team slug  
+      parameters:
+        - name: slug
+          in: path
+          required: true
+          schema:
+            type: string
+          description: PlanX team slug
       requestBody:
         required: true
         content:
@@ -71,8 +71,8 @@ paths:
                 type: object
                 properties:
                   data:
-                    type: integer  
-                    description: Metabase collection ID  
+                    type: integer
+                    description: Metabase collection ID
         "400":
           description: Bad request or collection creation failed
           content:
--- a/api.planx.uk/modules/analytics/docs.yaml
+++ b/api.planx.uk/modules/analytics/docs.yaml
@@ -2,12 +2,32 @@ openapi: 3.1.0
 info:
   title: Plan✕ API
   version: 0.1.0
+
 tags:
   - name: analytics
+  - name: metabase
+
 components:
   responses:
     AnalyticsResponse:
       description: Successful response with no content. Not awaited from server as endpoint is called via the Beacon API
+
+  schemas:
+    NewCollection:
+      type: object
+      required:
+        - slug
+      properties:
+        slug:
+          type: string
+          description: Name of the collection
+        description:
+          type: string
+          description: Optional description for the collection
+        parentId:
+          type: integer
+          description: Optional ID of the parent collection
+
 paths:
   /analytics/log-user-exit:
     post:
@@ -18,6 +38,7 @@ paths:
       responses:
         "204":
           $ref: "#/components/responses/AnalyticsResponse"
+
   /analytics/log-user-resume:
     post:
       summary: Log user resume
@@ -27,3 +48,36 @@ paths:
       responses:
         "204":
           $ref: "#/components/responses/AnalyticsResponse"
+
+  /metabase/collection:
+    post:
+      summary: Create new Metabase collection
+      description: Creates a new collection in Metabase if it doesn't already exist
+      tags:
+        - metabase
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/NewCollection"
+      responses:
+        "201":
+          description: Collection created successfully
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  data:
+                    $ref: "#/components/schemas/NewCollection"
+        "400":
+          description: Bad request or collection creation failed
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
+                    description: Error message
--- a/api.planx.uk/modules/analytics/docs.yaml
+++ b/api.planx.uk/modules/analytics/docs.yaml
@@ -5,7 +5,6 @@ info:
 
 tags:
   - name: analytics
-  - name: metabase
 
 components:
   responses:
--- a/api.planx.uk/modules/analytics/docs.yaml
+++ b/api.planx.uk/modules/analytics/docs.yaml
@@ -15,13 +15,8 @@ components:
   schemas:
     NewCollection:
       type: object
-      required:
-        - slug
-      properties:
-        slug:
-          type: string
-          description: Name of the collection
-        description:
+      properties:  
+        description:  
           type: string
           description: Optional description for the collection
         parentId:
@@ -49,12 +44,19 @@ paths:
         "204":
           $ref: "#/components/responses/AnalyticsResponse"
 
-  /metabase/collection:
+  /metabase/collection/{slug}:  
     post:
       summary: Create new Metabase collection
       description: Creates a new collection in Metabase if it doesn't already exist
       tags:
         - metabase
+      parameters:  
+        - name: slug  
+          in: path  
+          required: true  
+          schema:  
+            type: string  
+          description: PlanX team slug  
       requestBody:
         required: true
         content:
@@ -70,7 +72,8 @@ paths:
                 type: object
                 properties:
                   data:
-                    $ref: "#/components/schemas/NewCollection"
+                    type: integer  
+                    description: Metabase collection ID  
         "400":
           description: Bad request or collection creation failed
           content:
--- a/api.planx.uk/modules/analytics/metabase/collection/collection.test.ts
+++ b/api.planx.uk/modules/analytics/metabase/collection/collection.test.ts
@@ -18,7 +18,7 @@ describe("createTeamCollection", () => {
       teams: [
         {
           id: 26,
-          name: "Barnet",
+          slug: "barnet",
           metabase_id: null,
         },
       ],
@@ -27,15 +27,15 @@ describe("createTeamCollection", () => {
     // Mock Metabase API calls
     const metabaseMock = nock(process.env.METABASE_URL_EXT!)
       .post("/api/collection/", {
-        name: "Barnet",
+        slug: "barnet",
       })
       .reply(200, {
         id: 123,
-        name: "Barnet",
+        slug: "barnet",
       });
 
     const collectionId = await createCollection({
-      name: "Barnet",
+      slug: "barnet",
     });
 
     expect(collectionId).toBe(123);
@@ -49,37 +49,37 @@ describe("createTeamCollection", () => {
         returning: [
           {
             id: 26,
-            name: "Barnet",
+            slug: "barnet",
             metabase_id: 123,
           },
         ],
       },
     });
 
-    const testName = "Example council";
+    const testSlug = "example-council";
     const metabaseMock = nock(process.env.METABASE_URL_EXT!);
 
     // Mock collection creation endpoint
     metabaseMock
       .post("/api/collection/", {
-        name: testName,
+        slug: testSlug,
         parent_id: 100,
       })
       .reply(200, {
         id: 123,
-        name: testName,
+        slug: testSlug,
         parent_id: 100,
       });
 
     // Mock GET request for verifying the new collection
     metabaseMock.get("/api/collection/123").reply(200, {
       id: 123,
-      name: testName,
+      slug: testSlug,
       parent_id: 100,
     });
 
     const collectionId = await createCollection({
-      name: testName,
+      slug: testSlug,
       parentId: 100,
     });
 
@@ -92,12 +92,12 @@ describe("createTeamCollection", () => {
     expect(metabaseMock.isDone()).toBe(true);
   });
 
-  test("returns collection correctly no matter collection name case", async () => {
+  test("returns collection correctly no matter collection slug case", async () => {
     vi.spyOn($api.client, "request").mockResolvedValueOnce({
       teams: [
         {
           id: 26,
-          name: "barnet",
+          slug: "barnet",
           metabaseId: 20,
         },
       ],
@@ -114,7 +114,7 @@ describe("createTeamCollection", () => {
 
     await expect(
       createCollection({
-        name: "Test Collection",
+        slug: "test-collection",
       }),
     ).rejects.toThrow("Network error occurred");
   });
@@ -126,7 +126,7 @@ describe("createTeamCollection", () => {
 
     await expect(
       createCollection({
-        name: "Test Collection",
+        slug: "test-collection",
       }),
     ).rejects.toThrow(MetabaseError);
   });
@@ -228,50 +228,50 @@ describe("edge cases", () => {
     vi.resetAllMocks();
   });
 
-  test("handles missing name", async () => {
+  test("handles missing slug", async () => {
     await expect(
       createTeamCollection({
-        name: "",
+        slug: "",
       }),
     ).rejects.toThrow();
   });
 
-  test("handles names with special characters", async () => {
-    const specialName = "@#$%^&*";
+  test("handles slug with special characters", async () => {
+    const specialSlug = "@#$%^&*";
 
     nock(process.env.METABASE_URL_EXT!).get("/api/collection/").reply(200, []);
 
     nock(process.env.METABASE_URL_EXT!)
       .post("/api/collection/", {
-        name: specialName,
+        slug: specialSlug,
       })
       .reply(200, {
         id: 789,
-        name: specialName,
+        slug: specialSlug,
       });
 
     const collection = await createCollection({
-      name: specialName,
+      slug: specialSlug,
     });
     expect(collection).toBe(789);
   });
 
-  test("handles very long names", async () => {
-    const longName = "A".repeat(101);
+  test("handles very long slugs", async () => {
+    const longSlug = "A".repeat(101);
 
     nock(process.env.METABASE_URL_EXT!).get("/api/collection/").reply(200, []);
 
     nock(process.env.METABASE_URL_EXT!)
       .post("/api/collection/", {
-        name: longName,
+        slug: longSlug,
       })
       .reply(400, {
-        message: "Name too long",
+        message: "Slug too long",
       });
 
     await expect(
       createTeamCollection({
-        name: longName,
+        slug: longSlug,
       }),
     ).rejects.toThrow();
   });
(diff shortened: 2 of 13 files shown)


api.planx.uk/modules/analytics/metabase/collection/service.ts
```diff
@@ -0,0 +1,36 @@
+import { createMetabaseClient } from "../shared/client.js";
+import { updateMetabaseId } from "./updateMetabaseId.js";
+import type { NewCollectionParams } from "./types.js";
+import { getTeamIdAndMetabaseId } from "./getTeamIdAndMetabaseId.js";
+import { createCollection } from "./createCollection.js";
+
+const client = createMetabaseClient();
```
jamdelion, line 7: I don't think this line is needed, the `client` doesn't appear to be used anywhere

jamdelion: A better pattern in all these places where client is used - see the `getClient` pattern in this file: https://github.com/theopensystemslab/planx-new/blob/47aba73d1eefdf8782b12493264a925e1a58969c/api.planx.uk/modules/flows/moveFlow/service.ts#L7

jamdelion:
We're also getting an API test failure here in a test suite unrelated to metabase: https://github.com/theopensystemslab/planx-new/blob/47aba73d1eefdf8782b12493264a925e1a58969c/api.planx.uk/modules/send/s3/index.test.ts

I think this is because there we are globally mocking axios and not accounting for the new use case we have in metabase/shared/client.ts: `axios.create`. I'm not immediately sure how to fix this, somehow I think we need to restrict the mock in this file so that it doesn't care about how we are using axios elsewhere

zz-hh-aa:
Thanks @jamdelion, I sought Claude's advice on restricting the mock in this file and tried a few things out but to no avail. 

It looked like the axios spy in `client.test.ts` wasn't actually being used but a.) that didn't solve the problem and b.) even if it did, it probably wouldn't belong in this PR anyway would it 😅. So guess I should remove commit [778b6d4](https://github.com/theopensystemslab/planx-new/pull/4072/commits/778b6d46449d8a8dd5eb3cff0a74b0d2d0f6e525)?

Maybe we can flag at dev call on Thursday?

zz-hh-aa:
Cheers @DafyddLlyr I've just resolved those--up for talking things through tomorrow if you've got time then but no rush :)

DafyddLlyr (changes requested):
I've had a quick pass through with some initial comments - always happy to catch up and talk through if you'd like.

I'll checkout the branch and see if I can get to the bottom of the issue with tests and point you in the right direction.

In my mind, we're aiming for something like this - does this match your mental model?

```mermaid
sequenceDiagram
    participant User
    participant Hasura
    participant API
    participant MetabaseClient

    Note over User, Hasura: User Integration
    User->>Hasura: Create Team
    Hasura->>Hasura: Add record to teams table
    Hasura-->>User: Success

    Note over Hasura, MetabaseClient: Async action
    Hasura->>Hasura: Event: New team created
    Hasura->>API: Create collection request
    API->>MetabaseClient: Create collection
    MetabaseClient-->>API: Success
    API->>Hasura: Update Team Table
    Hasura-->>API: Success
 ```
 
Once we have an API endpoint that we can call to create a collection for a team, the next step will be to call it automatically when a team is created, and to populate the current production db with the existing collection ids.

File: api.planx.uk/modules/analytics/metabase/collection/service.ts [hunk 2]
```diff
@@ -0,0 +1,33 @@
...
+import { getTeamIdAndMetabaseId } from "./getTeamIdAndMetabaseId.js";
+import { createCollection } from "./createCollection.js";
+
+/**
+ * First uses name to get teams.id and .metabase_id, if present.
+ * If metabase_id is null, return an object that includes its id. If not, create the collection.
+ * @params `name` is required, but `description` and `parent_id` are optional.
+ * @returns `response.data`, so use dot notation to access `id` or `parent_id`.
+ */
+export async function createCollectionIfDoesNotExist(
```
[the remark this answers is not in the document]
zz-hh-aa: I think that works! 

With not creating Metabase clutter in mind, I don't know that we need to run this every time a team is created, especially if PlanX will over time take on more non-planning uses? (eg we don't need collections for demo, sandbox and other agency--eg EA--spaces). But I think that's a question that's outside of the scope of this PR. (One other thing to note is that we are currently hard-coding a `parent_id` for the Metabase collection, which would be in our "Councils" collection.)

api.planx.uk/modules/analytics/metabase/collection/types.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/** Interface for request after transforming to snake case (Metabase takes snake while PlanX API takes camel) */
+export interface MetabaseCollectionParams {
+  name: string;
+  description?: string;
+  parent_id?: number;
+}
+
+/** Metbase collection ID for the the "Council" collection **/
+// const COUNCILS_COLLECTION_ID = 58;
```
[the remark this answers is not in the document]
zz-hh-aa: Yep, that's why it's commented out. Is there a better way to handle this in the meantime?

api.planx.uk/modules/analytics/metabase/collection/service.ts
```diff
@@ -0,0 +1,33 @@
...
+import type { NewCollectionParams } from "./types.js";
+import { getTeamIdAndMetabaseId } from "./getTeamIdAndMetabaseId.js";
+import { createCollection } from "./createCollection.js";
+
+/**
+ * First uses name to get teams.id and .metabase_id, if present.
+ * If metabase_id is null, return an object that includes its id. If not, create the collection.
+ * @params `name` is required, but `description` and `parent_id` are optional.
+ * @returns `response.data`, so use dot notation to access `id` or `parent_id`.
+ */
```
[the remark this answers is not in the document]
zz-hh-aa: That makes sense, I think it was inexperience that led me to writing some 'what' comments for myself :) 

In this case I think there was _some_ slightly better reasoning too: that it was hard to come up with a name that also communicated the `try ... getTeamIdAndMetabaseId()` functionality. I've tried to update the comment (28712c4f) to explain a bit more of the why rather than the what, let me know what you think!

zz-hh-aa:
Made this back into a draft because I realised that I've just made team slugs the name of the Metabase collections too--will fix that (and look into [`api.editors.planx.dev/metabase/:teamSlug/collection`](https://github.com/theopensystemslab/planx-new/pull/4072#discussion_r1891380404) routes question too)

api.planx.uk/modules/analytics/metabase/collection/createCollection.ts
```diff
@@ -0,0 +1,20 @@
...
+  params: NewCollectionParams,
+): Promise<any> {
+  const transformedParams = {
+    name: params.name,
+    parent_id: params.parentId,
+  };
+
+  const response = await client.post(`/api/collection/`, transformedParams);
+
+  console.log(
```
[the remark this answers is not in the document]
DafyddLlyr: Yep great question! Generally we only leave this in if we're debugging or want to get more visibility on a feature which is raising errors. If we leave them in as just status updates it can lead to very noisy logs and tests.

api.planx.uk/modules/analytics/metabase/collection/service.ts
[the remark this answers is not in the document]
DafyddLlyr: Great point - I _think_ that the non-council, non-Metabase teams, would the exceptions generally. We can always add a guard for these or remove collections at a later date if this is an issue.

Agree it's outside the scope of this PR and something to circle back to later.

api.planx.uk/modules/analytics/metabase/collection/types.ts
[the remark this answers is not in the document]
DafyddLlyr: I think there's three main approaches - 
 - We set up `COUNCILS_COLLECTION_ID` as an env var and pass in a valid ID for staging/pizza/localhost - this will differ per dev so not ideal but perfectly workable as an initial step
 - We commit to Metabase being prod only, and add a middleware to all `/metabase` routes on our API which just exits early if we're not on production - makes testing tricky!
 - We set up Metabase on staging and sync data down
 
We might end up on the third option, but let's work our way there via 1 and 2 first!

api.planx.uk/modules/analytics/routes.ts
```diff
@@ -18,5 +20,10 @@ router.post(
   validate(logAnalyticsSchema),
   logUserResumeController,
 );
+router.post(
+  "/collection",
```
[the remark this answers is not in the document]
zz-hh-aa: I think I have sorted this, though I went with `collection/:teamSlug`--does that order seem important?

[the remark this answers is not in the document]
DafyddLlyr: Nice! Order here is totally fine 👍

api.planx.uk/modules/analytics/docs.yaml
```diff
@@ -2,12 +2,27 @@ openapi: 3.1.0
 info:
   title: Plan✕ API
   version: 0.1.0
+
 tags:
   - name: analytics
+  - name: metabase
+
```
DafyddLlyr, line 9: Despite the plural "tags", Swagger only allows a route to be in one location - 

image

We generally want these to match the structure of out API modules - we can either being metabase out to it's own module, or keep it in analytics.

I'd argue it's it's own module (is there any shared types/code/logic?) but this isn't a blocker and we can make this decision later - it's just a case of moving file around.

For now, let's chose the easy path and just apply the analytics tag.

api.planx.uk/modules/analytics/metabase/shared/client.ts
```diff
@@ -120,10 +121,4 @@ export const createMetabaseClient = (): AxiosInstance => {
...
 };
 
-// // Export both client and instance with delayed instantiation for test purposes
-// export let metabaseClient: AxiosInstance;
-
-// export const initializeMetabaseClient = () => {
-//   metabaseClient = createMetabaseClient();
-//   return metabaseClient;
-// };
+export const $metabase = createMetabaseClient();
```
DafyddLlyr, line 124: Nice!

api.planx.uk/modules/analytics/metabase/collection/service.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    if (metabaseId) {
+      return metabaseId;
+    }
+
+    const { slug, ...rest } = params;
+    const metabaseParams = {
+      name,
+      ...rest,
+    } as const;
```
DafyddLlyr, line 31: I'm not sure this `as const` is needed here. 

🧙‍♂️ https://www.youtube.com/watch?v=6M9aZzm-kEc

api.planx.uk/modules/analytics/metabase/collection/service.ts
```diff
@@ -0,0 +1,41 @@
...
+      metabaseId,
+      name,
+      id: teamId,
+    } = await getTeamIdAndMetabaseId(params.slug);
+
+    if (metabaseId) {
+      return metabaseId;
+    }
+
+    const { slug, ...rest } = params;
```
DafyddLlyr, line 27: Please see linting comment on PR about unused var.

api.planx.uk/modules/analytics/metabase/collection/service.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * The `getTeamIdAndMetabaseId()` function is run here to first get teams.id and .metabase_id from PlanX db, if present,
+ * so that the service can figure out if it needs to run `createCollection()` or not.
+ * Instead of running `getTeamIdAndMetabaseId()` first in the controller, it is encapsulated in `createTeamCollection` to keep business logic in `service.ts` instead of `controller.ts`.
+ * @params `slug` is required, but `description` and `parent_id` are optional.
+ * @returns `response.data`, so use dot notation to access `id` or `parent_id`.
+ */
+export async function createTeamCollection(
+  params: NewCollectionParams,
```
DafyddLlyr, line 14:
```suggestion
  { slug, parentId, description }: NewCollectionParams,
```

We could use destructing here to save a big of complexity and mapping below.

DafyddLlyr (commented):
Code looking great!

Just looking a bit deeper at tests now to sort out CI. One initial issue is that we don't have the right env vars set up on AWS (again 🤦‍♂️). 

The good news is that I think I worked out the cause of our headaches here. There's been a few PRs like this recently which have update the env file - https://github.com/theopensystemslab/planx-new/pull/4094. I think that these have just pushed a new .env file without first pulling, thus overwriting your new variables. The timing seems to line up. Could you please update the `.env` files on S3 again please (pull then push)?

DafyddLlyr:
@zz-hh-aa Please see [55d089b](https://github.com/theopensystemslab/planx-new/pull/4072/commits/55d089bb20d9dedf94118f11179347fffa6887e7) which should resolve the test issues in the API files.

zz-hh-aa:
Thanks for the review once again @DafyddLlyr! Sadly things aren't quite working yet🫠 

I've done the tidying up tasks, hopefully those are now resolved. 

I also re-pulled / pushed the vars (`cd` into `scripts/` and ran `pull-secrets.sh planx-staging` followed by `push-secrets.sh planx-staging`), but it looks like that hasn't helped. Not sure if I have done something wrong here? 

API tests are also failing despite [55d089b](https://github.com/theopensystemslab/planx-new/pull/4072/commits/55d089bb20d9dedf94118f11179347fffa6887e7).  I don't quite understand [the issue](https://github.com/theopensystemslab/planx-new/actions/runs/12465771457/job/34792152673?pr=4072#step:10:78), since it looks like there aren't explicit error messages, just warnings. I also thought the 'code style' issues should be addressed through pre-commit scripts, but maybe I've misunderstood?

api.planx.uk/modules/analytics/metabase/collection/service.ts
```diff
@@ -0,0 +1,37 @@
...
+import { getTeamIdAndMetabaseId } from "./getTeamIdAndMetabaseId.js";
+import { createCollection } from "./createCollection.js";
+
+/**
+ * The `getTeamIdAndMetabaseId()` function is run here to first get teams.id and .metabase_id from PlanX db, if present,
+ * so that the service can figure out if it needs to run `createCollection()` or not.
+ * Instead of running `getTeamIdAndMetabaseId()` first in the controller, it is encapsulated in `createTeamCollection` to keep business logic in `service.ts` instead of `controller.ts`.
+ * @params `slug` is required, but `description` and `parent_id` are optional.
+ * @returns `response.data`, so use dot notation to access `id` or `parent_id`.
+ */
```
jamdelion, line 12: nit: This comment seems a bit verbose/unnecessary imo! Particularly the params - this is covered by the type definition. And the functionality is quite self-explanatory from reading the code. 

I understand it was good for your understanding to have an explanation alongside the code but I'd expect the code to be able to speak for itself in most instances :)

jamdelion, line 12:
```suggestion
```

api.planx.uk/modules/analytics/metabase/collection/types.ts
```diff
@@ -0,0 +1,49 @@
...
+  /** Optional; if the collection is a child of a parent, specify parent ID here */
+  parentId?: number;
+}
+
+/** Interface for request after transforming to snake case (Metabase takes snake while PlanX API takes camel) */
+export type MetabaseCollectionParams = Omit<NewCollectionParams, "slug"> & {
+  name: string;
+};
+
+/** Metbase collection ID for the the "Council" collection **/
```
jamdelion, line 22:
```suggestion
/** Metabase collection ID for the the "Council" collection **/
```

Might be worth just adding another line explaining why commented out as per the review thread with Daf :)

api.planx.uk/modules/send/s3/index.test.ts
```diff
@@ -23,6 +23,11 @@ vi.mock("../../file/service/uploadFile.js", () => ({
 
 vi.mock("../../client/index.js");
 
+// Mock the Metabase client which also relies on Axios which these tests need to mock
+vi.mock("../../analytics/metabase/shared/client.js", () => ({
+  createMetabaseClient: vi.fn(),
+}));
+
```
jamdelion, line 30: Nice 😁

api.planx.uk/modules/analytics/metabase/collection/collection.test.ts
jamdelion, line 1: Nice comprehensive test coverage here 👍

jamdelion (approved):
Couple of nits to do with comments but overall great work getting this in good shape! 👍

api.planx.uk/modules/analytics/metabase/collection/getCollection.ts
```diff
@@ -0,0 +1,14 @@
+import type { GetCollectionResponse } from "./types.js";
+import { $metabase } from "../shared/client.js";
+
+/**
+ * Retrieves info on a collection from Metabase, use to check a parent. Currently only used in tests but could be useful for other Metabase functionality
+ * @param id
+ * @returns
+ */
```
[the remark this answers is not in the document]
zz-hh-aa: Cheers, I'd mistakenly thought that was tsdoc convention--thanks for pointing it out 😄

api.planx.uk/modules/analytics/metabase/collection/service.ts
```diff
@@ -0,0 +1,35 @@
+import { updateMetabaseId } from "./updateMetabaseId.js";
+import type { NewCollectionParams } from "./types.js";
+import { getTeamIdAndMetabaseId } from "./getTeamIdAndMetabaseId.js";
+import { createCollection } from "./createCollection.js";
+
+/**
+ * The `getTeamIdAndMetabaseId()` function is run here to first get teams.id and .metabase_id from PlanX db, if present,
+ * so that the service can figure out if it needs to run `createCollection()` or not.
+ * Instead of running `getTeamIdAndMetabaseId()` first in the controller, it is encapsulated in `createTeamCollection` to keep business logic in `service.ts` instead of `controller.ts`.
+ */
```
DafyddLlyr, line 10: I actually think the code here is pretty clear and easy to follow - we can safely remove this comment. I found it a little harder to follow and reason with in English than in code 😂

api.planx.uk/modules/analytics/metabase/collection/types.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/** TODO: when running on production, turn below comment back into code
+ * the Metabase collection ID is for the "Council" collection
+ * see https://github.com/theopensystemslab/planx-new/pull/4072#discussion_r1892631692
+ **/
+// const COUNCILS_COLLECTION_ID = 58;
+
+export const createTeamCollectionSchema = z.object({
+  params: z.object({
+    slug: z.string(),
```
DafyddLlyr, line 26: Just seen the "long slug" test - if Metabase has a restriction here (presumably 100 based on test?), we can also catch this early here. 

```suggestion
    slug: z.string().max(100),
```

DafyddLlyr: There's no restriction on PlanX team names so it's _possible_ that an team name could exceed 100 characters (Llanfair­pwllgwyngyll­gogery­chwyrn­drobwll­llan­tysilio­gogo­goch 😉 ).

If this were to ever happen I'm sure we can limit and make unique in some way 👍

api.planx.uk/modules/analytics/metabase/collection/collection.test.ts
```diff
@@ -90,3 +223,59 @@ describe("updateMetabaseId", () => {
...
+  beforeEach(() => {
+    nock.cleanAll();
+    vi.clearAllMocks();
+    vi.resetAllMocks();
+  });
+
+  test("handles missing slug", async () => {
+    await expect(
+      createTeamCollection({
+        slug: "",
```
DafyddLlyr, line 237: We could also check this with zod using `.min(1)`

DafyddLlyr (approved):
Looking great - a few very minor suggestions but no blockers at all! 👏

api.planx.uk/modules/analytics/metabase/collection/types.ts
zz-hh-aa: Don't think Metabase actually has a limit, just thought it was something that might be worth testing / went with an arbitrary number, so will leave out of validation for now!

PR #4076, opened 2024-12-12
boost hasura-proxy Fargate container CPU / memory (staging)

freemvmt:

The 180s grace period introduced in #4051 didn't resolve the scaling issue, so trying a different tack. This PR boosts CPU and memory resources of the hasura-proxy container (and therefore of the Task overall).

Also moves the variables for controlling those resources to the Pulumi yaml (staging only for now).


freemvmt:
Increase in shared resources of Hasura task immediately evident

image

PR #4081, opened 2024-12-13
bump staging RDS instance to T3 Small

freemvmt:

This PR bumps the RDS PostgreSQL DB instance used on staging to `db.t3.small`, in order to increase the number of connections it can handle by doubling the memory (to 2GB).

Prod is still pinned to `db.t3.medium`, which has twice the memory again (4GB).

We only need to merge this if #4080 doesn't solve the scaling issue - also see that PR for further context and motivation for this change.

Also note that we'll have to manually roll out the upgrade on AWS via Pulumi CLI, since there is no CI setup for the `data` layer as there is for `application` (see the `push-main` workflow).

### DIFF
--- a/infrastructure/data/index.ts
+++ b/infrastructure/data/index.ts
@@ -16,7 +16,7 @@ const db = new aws.rds.Instance("app", {
   // $ aws rds describe-db-engine-versions --default-only --engine postgres
   engineVersion: "12.17",
   // Available instance types: https://aws.amazon.com/rds/instance-types/
-  instanceClass: env === "production" ? "db.t3.medium" : "db.t3.micro",
+  instanceClass: env === "production" ? "db.t3.medium" : "db.t3.small",
   allocatedStorage: env === "production" ? 100 : 20,
   allowMajorVersionUpgrade: true,
   dbSubnetGroupName: networking.requireOutput("subnetId"),


DafyddLlyr (approved):
Approved so that you can merge this should you wish to do so 👍 

> Also note that we'll have to manually roll out the upgrade on AWS via Pulumi CLI, since there is no CI setup for the data layer as there is for application (see the push-main workflow).

Very well remembered! We can run `pulumi up --stack staging` from the `/infrastructure/data` folder locally in order to achieve this - there's an intermediate step where you'll get told if there are any destructive action (the classic IaC `+, -, ~`).

I'd expect this to show a change `~` to instance class which would be non-destructive (i.e. not a `+ / -`). I think that the change will be auto-applied at the next maintenance window but I'm not totally sure, the might actually be a pulumi flag to force the change.

If the absolute worst happens (very very unlikely) there's nothing irreplaceable on the staging db - it's just a clone of a subset of data from production.

freemvmt:
@DafyddLlyr right you were!

image

However I apparently didn't have the right permissions to make this change from local:

image

So I actually just made the change directly in the AWS console as we discussed the other day. Will merge this PR to reflect that the change has been applied.

PR #4085, opened 2024-12-17
migrate: find+replace DrawBoundary `dataFieldBoundary` prop to `fn` and remove `dataFieldArea` prop

jessicamcinchak:
Queueing up per migration plan: https://docs.google.com/spreadsheets/d/1Vtxp5BLweDPDooQoNhgOCYjCIBPRYIcOuyArGJRqOkI/edit?gid=0#gid=0

Planx-core counterpart here https://github.com/theopensystemslab/planx-core/pull/594

### DIFF
--- a/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
@@ -58,7 +58,7 @@ export default function Component(props: Props) {
     passport.data?.["property.boundary"];
   const previousArea =
     props.previouslySubmittedData?.data?.[props.fn] ||
-    passport.data?.["proposal.site.area"];
+    passport.data?.["property.boundary.area"];
   const [boundary, setBoundary] = useState<Boundary>(previousBoundary);
   const [area, setArea] = useState<number | undefined>(previousArea);
   const [mapValidationError, setMapValidationError] = useState<string>();


editor.planx.uk/src/@planx/components/DrawBoundary/Editor.tsx
```diff
@@ -53,25 +53,13 @@ function DrawBoundaryComponent(props: Props) {
...
-          <InputGroup label="Area data field (square metres)">
-            <InputRow>
-              <Input
-                name="dataFieldArea"
-                placeholder="property.boundary.area"
-                format="data"
-                value={formik.values.dataFieldArea}
-                onChange={formik.handleChange}
+                value={formik.values.fn}
+                disabled
```
jessicamcinchak, line 62: This file is a good summary / convergence point of main significant changes throughout: 
- Only support a single data field prop (rather than two data fields)
- Change the underlying prop name for the data field so it's picked up in graph styling, search facets by default
- Disable the data field input similar to PlanningConstraints & Pay as it's actually very important for integrations that it's called exactly this
- Change the actual data field _value_ to better align to schema

**Before:**
Screenshot from 2024-12-17 17-59-28
Screenshot from 2024-12-17 17-59-04

**After:** 
Screenshot from 2024-12-17 17-58-28
Screenshot from 2024-12-17 17-58-02

RODO94 (approved):
Looks good to me, I really like these changes! Thanks so much for putting this together.

Couldn't see any mistakes in the changes or any overlooked changes.

PR #4089, opened 2024-12-18
fix: Feedback table alignment

ianjon3s:
## What does this PR do?

Quick one: fixes alignment of the feedback table dropdown, as another header row has been added.

**Before:**
image

**After:**
image

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackLog/components/CollapsibleRow.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackLog/components/CollapsibleRow.tsx
@@ -167,7 +167,7 @@ export const CollapsibleRow: React.FC<CollapsibleRowProps> = (item) => {
         </TableCell>
       </TableRow>
       <TableRow sx={{ background: (theme) => theme.palette.background.paper }}>
-        <TableCell sx={{ padding: 0, border: "none" }} colSpan={4}>
+        <TableCell sx={{ padding: 0, border: "none" }} colSpan={5}>
           <Collapse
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackLog/components/CollapsibleRow.tsx
```diff
@@ -167,7 +167,7 @@ export const CollapsibleRow: React.FC<CollapsibleRowProps> = (item) => {
         </TableCell>
       </TableRow>
       <TableRow sx={{ background: (theme) => theme.palette.background.paper }}>
-        <TableCell sx={{ padding: 0, border: "none" }} colSpan={4}>
+        <TableCell sx={{ padding: 0, border: "none" }} colSpan={5}>
```
DafyddLlyr, line 170: One to think about when we're revising these tables - ideally this would dynamically reflect the number of columns.

DafyddLlyr (approved):
Great catch thanks! 🙌

editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackLog/components/CollapsibleRow.tsx
ianjon3s: Agreed, there used to be a useful `colspan="0"` property that would span all available cells, which turned out to be a hack and was depreciated with HTML5. Other option is set an unrealistically high number (i.e. `colspan="9999"`), but that doesn't feel like a proper solution.

PR #4090, opened 2024-12-18
feat: Add Preview/Draft Layout Wrapper for Testing Links

RODO94:
## What does this PR do?

This PR adds a wrapper around the Preview and Draft view to initially show a **'This is a test environment'** node that a user would have to navigate past.

For this PR, I used the `Layout` pattern as a guide, but just used a `Page` rather than a `Layout` to wrap the `View`. I started with the `Layout` in place of the `Page` below but it didn't contain any extra logic or anything useful, so I removed it.

```jsx
  return (
    <PublicLayout>
      <TestWarningPage>
        <View />
      </TestWarningPage>
    </PublicLayout>
  );
```

I used the Save and Return journey as a guide and created a new component inside the `Card` component to control which `Or` route is used. This has all been decoupled from the `Card` component and added to the `OrNavigationButton` 

```tsx
       <OrNavigationButton handleSubmit={handleSubmit} />
```

Hopefully this is an elegant solution and I have overcomplicated this `<OrNavigationButton>`. An unintended consequence of this is that the `Go to live link` is now present as an `Or` option on the `Send` component. This is why I had to mock the react-navi import for the tests

> [!IMPORTANT]
> When clicking "Go to published..." the flow data is deleted from localStorage and window is reset similar to the `Restart` button while you are navigated to the published?analytics=false page

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
@@ -9,8 +9,7 @@ import { useStore } from "pages/FlowEditor/lib/store";
 import React, { useEffect } from "react";
 import { ApplicationPath } from "types";
 
-import NavigateToPublishedButton from "./NavigateToPublishedButton";
-import SaveResumeButton from "./SaveResumeButton";
+import OrNavigationButton from "./OrNavigationButton";
 
 interface Props {
   children: React.ReactNode;
@@ -18,6 +17,8 @@ interface Props {
   isTestWarningWrapper?: boolean;
   handleSubmit?: (data?: any) => void;
 }
+
+export type OrNavigationType = "save-resume" | "navigate-to-published";
 
 export const contentFlowSpacing = (theme: Theme): React.CSSProperties => ({
   marginTop: theme.spacing(2),
@@ -48,6 +49,7 @@ const Card: React.FC<Props> = ({
   children,
   isValid = true,
   handleSubmit,
+  isTestWarningWrapper,
   ...props
 }) => {
   const theme = useTheme();
@@ -57,21 +59,33 @@ const Card: React.FC<Props> = ({
     state.breadcrumbs,
     state.flow,
   ]);
-
-  // Check if we have a Send node in our breadcrumbs
-  //   This is a better/more immediate proxy for "submitted" in the frontend because actual send events that populate lowcal_sessions.submitted_at are queued via Hasura
-  const hasSent = Object.keys(breadcrumbs).some(
-    (breadcrumbNodeId: string) => flow[breadcrumbNodeId]?.type === TYPES.Send,
-  );
-
-  const showSaveResumeButton =
-    path === ApplicationPath.SaveAndReturn && handleSubmit && !hasSent;
   const { track } = useAnalyticsTracking();
+
+  const defineNavigationType = (): OrNavigationType | undefined => {
+    // Check if we have a Send node in our breadcrumbs
+    //   This is a better/more immediate proxy for "submitted" in the frontend because actual send events that populate lowcal_sessions.submitted_at are queued via Hasura
+    const hasSent = Object.keys(breadcrumbs).some(
+      (breadcrumbNodeId: string) => flow[breadcrumbNodeId]?.type === TYPES.Send,
+    );
+
+    const showSaveResumeButton =
+      path === ApplicationPath.SaveAndReturn && handleSubmit && !hasSent;
+
+    if (showSaveResumeButton && !isTestWarningWrapper) {
+      return "save-resume";
+    }
+
+    if (!showSaveResumeButton && isTestWarningWrapper) {
+      return "navigate-to-published";
+    }
+  };
 
   useEffect(() => {
     // The Card component is only rendered when there's content the user will see
     if (visibleNode?.id) track(visibleNode?.id);
   }, []);
+
+  const orNavigationType = defineNavigationType();
 
   return (
     <Fade
@@ -103,8 +117,9 @@ const Card: React.FC<Props> = ({
                 Continue
               </Button>
             )}
-            {showSaveResumeButton && <SaveResumeButton />}
-            {props.isTestWarningWrapper && <NavigateToPublishedButton />}
+            {orNavigationType !== undefined && (
+              <OrNavigationButton type={orNavigationType} />
+            )}
           </Box>
         </InnerContainer>
       </Container>
--- a/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
@@ -15,7 +15,7 @@ import SaveResumeButton from "./SaveResumeButton";
 interface Props {
   children: React.ReactNode;
   isValid?: boolean;
-  isTest?: boolean;
+  isTestWrapper?: boolean;
   handleSubmit?: (data?: any) => void;
 }
 
@@ -48,6 +48,7 @@ const Card: React.FC<Props> = ({
   children,
   isValid = true,
   handleSubmit,
+  isTestWrapper,
   ...props
 }) => {
   const theme = useTheme();
@@ -67,10 +68,6 @@ const Card: React.FC<Props> = ({
   const showSaveResumeButton =
     path === ApplicationPath.SaveAndReturn && handleSubmit && !hasSent;
   const { track } = useAnalyticsTracking();
-
-  const showGoToPublishedServiceButton =
-    window.location.pathname.endsWith("/draft") ||
-    (window.location.pathname.endsWith("/preview") && props.isTest);
 
   useEffect(() => {
     // The Card component is only rendered when there's content the user will see
@@ -108,7 +105,7 @@ const Card: React.FC<Props> = ({
               </Button>
             )}
             {showSaveResumeButton && <SaveResumeButton />}
-            {showGoToPublishedServiceButton && <NavigateToPublishedButton />}
+            {isTestWrapper && <NavigateToPublishedButton />}
           </Box>
         </InnerContainer>
       </Container>
--- a/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
@@ -9,11 +9,13 @@ import { useStore } from "pages/FlowEditor/lib/store";
 import React, { useEffect } from "react";
 import { ApplicationPath } from "types";
 
+import NavigateToPublishedButton from "./NavigateToPublishedButton";
 import SaveResumeButton from "./SaveResumeButton";
 
 interface Props {
   children: React.ReactNode;
   isValid?: boolean;
+  isTest?: boolean;
   handleSubmit?: (data?: any) => void;
 }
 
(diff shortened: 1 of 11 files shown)


editor.planx.uk/src/@planx/components/shared/Preview/NavigateToPublishedButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+const NavigateToPublishedButton: React.FC = () => {
+  const testEnvironment = window.location.pathname.endsWith("/draft")
+    ? "/draft"
+    : "/preview";
+
+  const editorLink = window.location.pathname.replace(
+    testEnvironment,
+    "/published",
+  ) as `/${string}`;
```
RODO94, line 15: Used a type assertion here to make it clear where the slash is on this part of the link

editor.planx.uk/src/@planx/components/shared/Preview/NavigateToPublishedButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const redirectLink = `${window.location.origin}${editorLink}`;
```
RODO94, line 16: Using this method to define the links feels a bit janky to me, but with the data available, it was the only way I thought about doing it. 

Hoping that the subdomain logic can be handled with the `window.location.origin`

editor.planx.uk/src/@planx/components/shared/Preview/NavigateToPublishedButton.tsx
```diff
@@ -0,0 +1,36 @@
+import Link from "@mui/material/Link";
+import Typography from "@mui/material/Typography";
+import React from "react";
+import { useNavigation } from "react-navi";
+
+import { InnerContainer } from "./SaveResumeButton";
```
DafyddLlyr, line 6: If this is now shared, it should be its own component. 

Generally we'd try to avoid feature to feature imports like this. The main reason being that when I make a change in future on this component it's not inherently obvious that it's going to have a knock on effect in an unrelated component.

A few options here - we could repeat the style, extract it to its own very simple component, or make a component which is the "or" text option which then takes in a callback function. I think the later is the best option here as it keeps an "or" text options consistent and we could control within the card that it only shows once and doesn't stack as is currently possible.

How about a component which took in a condition (to decide if it's shown or hidden) and a callback (to be called when clicked), which is then passed into the `Card`?

editor.planx.uk/src/@planx/components/shared/Preview/Card.tsx
```diff
@@ -9,11 +9,13 @@ import { useStore } from "pages/FlowEditor/lib/store";
 import React, { useEffect } from "react";
 import { ApplicationPath } from "types";
 
+import NavigateToPublishedButton from "./NavigateToPublishedButton";
 import SaveResumeButton from "./SaveResumeButton";
 
 interface Props {
   children: React.ReactNode;
   isValid?: boolean;
+  isTestWarningWrapper?: boolean;
```
DafyddLlyr, line 18: image

editor.planx.uk/src/@planx/components/shared/Preview/NavigateToPublishedButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+    : "/preview";
+
+  const editorLink = window.location.pathname.replace(
+    testEnvironment,
+    "/published",
+  ) as `/${string}`;
+
+  const handleClick = () => {
+    navigation.navigate(editorLink);
+  };
```
DafyddLlyr, line 22:
```suggestion
  const { navigate } = useNavigation();

  const handleClick = () => {
    navigate("published");
  };
```

DafyddLlyr: I think this would do the same thing but much simpler - we just get react-navi to take us to a new route - it's always "published".

editor.planx.uk/src/pages/Preview/TestWarningPage.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { CardHeader } from "@planx/components/shared/Preview/CardHeader/CardHeader";
+import React, { PropsWithChildren, useState } from "react";
+
+export const TestWarningPage = ({ children }: PropsWithChildren) => {
+  const [showChildren, setShowChildren] = useState(false);
+  return (
+    <>
+      {showChildren ? (
+        children
+      ) : (
```
DafyddLlyr, line 12:
```suggestion
  const [hasAcknowledgedWarning, setHasAcknowledgedWarning] = useState(false);
  if (hasAcknowledgedWarning) return children;
```

Something like this may be clearer? `showChildren` holds no semantic meaning and is describing the what the behaviour (which can already be understood from the code).

DafyddLlyr (changes requested):
Looking good!

A few comments to take a look at 👍

editor.planx.uk/src/@planx/components/shared/Preview/NavigateToPublishedButton.tsx
RODO94: Great comment, thanks for the suggestions here. @DafyddLlyr 

This may be an indulgence at the wrong time, but I've been reading a bit on decoupling code for maintainability and thus, taken a slightly different approach using the comment as inspo. 

I wanted to keep the handleClick logic within each `SaveAndResumeButton` and `NavigateToPublishedButton` so they could expand or become more complex in future without bloating out the `Card` and it means the components can focus on doing one thing well at a time. 

I have put in an extra component, `OrNavigationButton` to handle the logic of the displaying either button based on a `type` it receives from the `Card` component, so it again just does one wee thing well. The `Card` comp can then do a small function to define what type to show. 

Lemme know if I've led myself astray and you're suggestion is a cleaner way of doing what I was hoping to do!

editor.planx.uk/src/@planx/components/shared/Preview/NavigateToPublishedButton.tsx
```diff
@@ -0,0 +1,21 @@
...
+const NavigateToPublishedButton: React.FC = () => {
+  const { navigate } = useNavigation();
+  const handleClick = () => {
+    navigate("published");
+  };
+
+  return (
+    <Link onClick={handleClick} component="button">
+      <Typography variant="body1" textAlign="left">
+        Go to the live service
```
jessicamcinchak, line 15: The copy provided by Alice attached to the Trello ticket uses the language "Go to the published version of this service" - can we please use that here? Draft & preview links aren't technically _not live_ or offline per se so I don't think we want to mix that language here?

jessicamcinchak:
One question related to this point: 
> Important
> "Go to the live service" remains across each node and when you click it, it will navigate you to that node in the flow rather than back at the start with pre-filled in questions

What happens when the `/preview` or `/draft` links have a node that is not in the published version? (As is most commonly the case? think this could lead to very strange and opaque submisison data errors?) 

I don't think I actually agree that this is intended behavior, and I think users should be dropped at the beginning of a published service when clicking this link rather than midway?

editor.planx.uk/src/@planx/components/shared/Preview/NavigateToPublishedButton.tsx
RODO94: Yeah sorry, I'll add that in now

RODO94:
> One question related to this point:
> 
> > Important
> > "Go to the live service" remains across each node and when you click it, it will navigate you to that node in the flow rather than back at the start with pre-filled in questions
> 
> What happens when the `/preview` or `/draft` links have a node that is not in the published version? (As is most commonly the case? think this could lead to very strange and opaque submisison data errors?)
> 
> I don't think I actually agree that this is intended behavior, and I think users should be dropped at the beginning of a published service when clicking this link rather than midway?

@jessicamcinchak yeah, I wasn't sure on this, thanks for the clarification, I'll reset back to start on this

editor.planx.uk/src/@planx/components/shared/Preview/OrNavigationButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+  "save-resume": SaveResumeButton,
+  "navigate-to-published": NavigateToPublishedButton,
+} as const;
+
+const TEST_ENVIRONMENTS = new Set(["preview", "draft"]);
+
+const OrNavigationButton = ({
+  handleSubmit,
+}: {
+  handleSubmit: ((data?: any) => void) | undefined;
```
DafyddLlyr, line 31:
```suggestion
  handleSubmit: ((data?: unknown) => void) | undefined;
```

We should actually come back and type `handleSubmit` properly, but for now let's work to not introduce any new `any`s to the codebase.

editor.planx.uk/src/@planx/components/shared/Preview/OrNavigationButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+}) => {
+  const [path, breadcrumbs, flow] = useStore((state) => [
+    state.path,
+    state.breadcrumbs,
+    state.flow,
+  ]);
+
+  const endOfUrl = window.location.pathname.split("/").slice(-1)[0];
+
+  const isTestEnvironment: boolean = TEST_ENVIRONMENTS.has(endOfUrl);
```
DafyddLlyr, line 41: nit: Redundant type?

P.S. love to see a `Set` being used 👏

editor.planx.uk/src/@planx/components/shared/Preview/OrNavigationButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+    orNavigationType && BUTTON_COMPONENTS[orNavigationType];
+
+  return (
+    ButtonComponent && (
+      <InnerContainer>
+        <Typography variant="body1">or</Typography>
+        <ButtonComponent />
+      </InnerContainer>
+    )
+  );
```
DafyddLlyr, line 74:
```suggestion
  const orNavigationType = defineNavigationType();

   if (!orNavigationType) return null;

  const ButtonComponent = BUTTON_COMPONENTS[orNavigationType];

  return (
    <InnerContainer>
      <Typography variant="body1">or</Typography>
      <ButtonComponent />
    </InnerContainer>
  );
```

An early return here feels more idiomatic imho.

This is a bit more clear and explicit - if the condition it not met, don't show this component.

It also helps us with type narrowing - we no longer need to check for the `ButtonComponent` which must always be defined.

DafyddLlyr (changes requested):
This is looking great - really a big step up in clarity (and quality!) than last time I reviewed it. I appreciate the work and though that's gone into it 👍 

There's some very minor nit comments on the code below, no showstoppers.

Only thing blocking an approve from me is @jessicamcinchak's previous point about retaining the "or" option as the user progresses through their journey. This could lead to both weird submission behaviours and a very unexpected user journey.

I would suggest that we want would be - 
 - The `TestWarningPage` shows ✅ 
 - The user can go to the published version ✅ 
 - The user can continue ✅ 
   - If they do continue,  the "or" option goes away ❌ 

I think this is actually relatively simple to achieve. 

If the `TestWarningPage` sets `hasAcknowledgedWarning` to global state (via the `Zustand` store) instead of it's local React `useState()` then we can check this inside the `OrNavigationButton` when deciding what to show.

editor.planx.uk/src/@planx/components/shared/Preview/OrNavigationButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+      return "navigate-to-published";
+    }
+  };
+
+  const orNavigationType = defineNavigationType();
+
+  if (!orNavigationType) return null;
+
+  if (hasAcknowledgedWarning && orNavigationType === "navigate-to-published")
+    return null;
```
DafyddLlyr, line 70: This logic seems a little out of place.

What we really mean here is that in order for the `orNavigationType` to be `"navigate-to-published"`, `hasAcknowledgedWarning` must be `false`.

This logic would make more sense to me as part of `defineNavigationType()`. Splitting this up make it harder to reason with and expand upon in future.

```ts
if (!showSaveResumeButton && isTestEnvironment && !hasAcknowledgedWarning) {
  return "navigate-to-published";
}
```

DafyddLlyr (approved):
Great!

PR #4095, opened 2024-12-19
feat: add audit table for data migration scripts

jessicamcinchak:
Following on from data migration chat! This is a basic table structure that will help us both: queue up _which_ flows should be updated by a given data migration script and track which have been `updated` so far.

Exclusively the `platformAdmin` role can access and update this table.

I'm imagining: 
- We'll simply manually populate the table like so:
```psql
INSERT INTO temp_data_migrations_audit (flow_id, team_id)
SELECT id, team_id
FROM flows 
WHERE team_id IN (1,2,3); -- team_ids match councils that have "consented"
```
- Then kick off the migration script which will fetch/transform/update one flow/row at a time `where updated = false` until all records have been updated
- Finally after a successful migration, we can simply `TRUNCATE` the table and re-use the same structure for our next use-case (plus option to export as CSV like a receipt and store on gdrive before truncation if necessary)

### DIFF
--- a/hasura.planx.uk/migrations/1734623060683_create_table_public_temp_data_migrations_audit/up.sql
+++ b/hasura.planx.uk/migrations/1734623060683_create_table_public_temp_data_migrations_audit/up.sql
@@ -1,0 +1,8 @@
+CREATE TABLE "public"."temp_data_migrations_audit" (
+  "flow_id" uuid NOT NULL, 
+  "team_id" integer NOT NULL, 
+  "updated" boolean NOT NULL DEFAULT false, 
+  "updated_at" timestamptz NOT NULL DEFAULT now(), 
+  PRIMARY KEY ("flow_id") 
+);
+COMMENT ON TABLE "public"."temp_data_migrations_audit" IS E'Basic table structure that we can use to track status of data migrations. Contents of this table are ephemeral and may be truncated between migrations';
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -2465,6 +2465,9 @@
         filter: {}
       comment: ""
 - table:
+    name: temp_data_migrations_audit
+    schema: public
+- table:
     name: uniform_applications
     schema: public
   insert_permissions:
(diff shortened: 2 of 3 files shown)


hasura.planx.uk/migrations/1734623060683_create_table_public_temp_data_migrations_audit/up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE "public"."temp_data_migrations_audit" (
+  "flow_id" uuid NOT NULL, 
+  "team_id" integer NOT NULL,
```
DafyddLlyr, line 3: Would relationships here be helpful for queries in the migration scripts?

DafyddLlyr (approved):

hasura.planx.uk/migrations/1734623060683_create_table_public_temp_data_migrations_audit/up.sql
jessicamcinchak: I don't think so ! As I see it we'll simply return the `flow_id` from this table (it's the primary key so indexed) then pass into a GraphQL relational query which will fetch the flow and its' latest published version (so meaningful foreign key is at GraphQL & existing `flows` table level rather than on this one?). But please object if I'm missing a benefit of fkeys directly here!

jessicamcinchak: Oof slow brain today - answered my own question there - you're totally right, if we set a proper fkey to flows here then migration script will be able to make a single query rather than two :+1: update incoming!

jessicamcinchak: :white_check_mark: 
```graphql
query MyQuery {
  temp_data_migrations_audit(where: {updated: {_eq: false}}, limit: 1) {
    flow_id
    flows {
      data
      published_flows(order_by: {created_at: desc}, limit: 1) {
        data
      }
    }
  }
}
```

PR #4099, opened 2024-12-20
fix: Responsive type sizes

ianjon3s:
## What does this PR do?

MUI responsive type sizes provides a one-value (`factor: x`) scale for resizing type which aims to provide a global function, but removes the ability to scale or fine-tune by a particular type property.

In PlanX we use h2 sizes for most page/question titles and h1 for headline/banners, MUI was resizing these to be too small comparatively to be effective.

I've tried a number of methods to work with MUI's automatic scale, ideally we'd use an up-to-date method (such as `fontSize: "clamp(2.125rem, 3rem, 5rem)",`, but MUI needs to see a hardcoded value for each property.

This PR introduces manual media queries to increase larger header sizes at the smallest (sub 500px wide) screens. There is also the reluctant use of an `!important` rule as MUI trumps any media queries defined in the theme with its own, and there is an apparent bug in `h4` resizes in which MUI does not scale this correctly between 500px – 768px.

Before changes, largest viewport (left) vs smallest viewport (right):
image

After changes, largest viewport (left) vs smallest viewport (right):
image

### DIFF
--- a/editor.planx.uk/src/theme.ts
+++ b/editor.planx.uk/src/theme.ts
@@ -156,11 +156,17 @@ const getThemeOptions = ({
         fontSize: "3rem",
         letterSpacing: SPACING_TIGHT,
         fontWeight: FONT_WEIGHT_BOLD,
+        "@media (max-width: 500px)": {
+          fontSize: "2.125rem",
+        },
       },
       h2: {
         fontSize: "2.25rem",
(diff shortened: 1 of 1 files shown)


DafyddLlyr (approved):

DafyddLlyr:
Nice catch thanks!

PR #4102, opened 2024-12-20
fix: add `logs` column to `temp_data_migrations_audit` table and create relationship from `flow` to `lowcal_sessions`

jessicamcinchak:
A couple quick database migrations per our discussion !

### DIFF
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -448,6 +448,15 @@
           table:
             name: flow_document_templates
             schema: public
+    - name: lowcal_sessions
+      using:
+        manual_configuration:
+          column_mapping:
+            id: flow_id
+          insertion_order: null
+          remote_table:
+            name: lowcal_sessions
+            schema: public
     - name: operations
       using:
         foreign_key_constraint_on:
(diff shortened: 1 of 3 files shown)


hasura.planx.uk/metadata/tables.yaml
```diff
@@ -448,6 +448,15 @@
...
             schema: public
+    - name: lowcal_sessions
+      using:
+        manual_configuration:
+          column_mapping:
+            id: flow_id
+          insertion_order: null
+          remote_table:
+            name: lowcal_sessions
+            schema: public
```
jessicamcinchak, line 459: This allows us to relationally query the `lowcal_sessions` for a given `flow`

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -2468,7 +2477,7 @@
     name: temp_data_migrations_audit
     schema: public
   object_relationships:
-    - name: flows
+    - name: flow
```
jessicamcinchak, line 2480: This is an `object_relationship` and should therefore be a singular rather than plural name, my mistake yesterday !

DafyddLlyr (approved):

PR #3917, opened 2024-11-06
chore: extract Public component wrappers for auto-answering into preview Node

jessicamcinchak:
First step before moving onto auto-answering Input node types - this should be a more flexible, extensible pattern than per-`@planx/components` "visibility wrappers"

- [x] Filters
- [ ] Questions
- [ ] Checklists

PR #4108, opened 2025-01-05
chore: refactor public component wrappers for auto-answering into preview `Node`

jessicamcinchak:
Replaces outdated https://github.com/theopensystemslab/planx-new/pull/3917

First step before moving onto support auto-answering Input component types ([per logic followups outlined here](https://trello.com/c/dboRkua3/3112-automation-filtering-logic-followups)) - this refactor sets up a more flexible, extensible pattern than per-component public "visibility wrappers" as initially implemented. Now, all auto-answerable node IDs are computed within the parent `Node` component and passed down to relevant component types only.

Updates the following `@planx/components`:
- [x] Filters
- [x] Questions
- [x] Checklists

All existing tests still passing (Public component tests, automation logic unit tests, e2e).


editor.planx.uk/src/pages/Preview/Node.tsx
```diff
@@ -96,13 +104,16 @@ const Node: React.FC<Props> = (props) => {
   });
 
   switch (props.node.type) {
+    case TYPES.AddressInput:
+      return <AddressInputComponent {...getComponentProps<AddressInput>()} />;
+
```
jessicamcinchak, line 109: This file is a bit extra noisy because I've alphabetized our component types within the `switch` block - only `Checklist`, `Filter`, `Question` types have functional changes!

editor.planx.uk/src/@planx/components/Question/Public/Question.fixture.tsx
```diff
@@ -1,28 +0,0 @@
-import React from "react";
-
-import Question from "./Question";
-
-export default (
```
jessicamcinchak, line 5: I'm dropping this `fixture` file as I believe it's unused and not a component pattern we follow elsewhere - any objections? Dropping this also allows to eliminate `Public/package.json`

editor.planx.uk/src/@planx/components/Checklist/Public/Public.tsx
```diff
@@ -1,28 +1,156 @@
...
+import { GroupedChecklistOptions } from "./components/GroupedChecklistOptions";
+import { toggleNonExclusiveCheckbox } from "./helpers";
+import { useExclusiveOption } from "./hooks/useExclusiveOption";
+import { useSortedOptions } from "./hooks/useSortedOptions";
+
+export enum ChecklistLayout {
+  Basic,
+  Grouped,
+  Images,
+}
```
DafyddLlyr, line 26: I think this should be in model.ts or types.ts

editor.planx.uk/src/@planx/components/Checklist/model.ts
```diff
@@ -1,7 +1,7 @@
 import { array } from "yup";
 
 import { BaseNodeData, Option } from "../shared";
-import { ChecklistLayout } from "./Public/components/VisibleChecklist";
+import { ChecklistLayout } from "./Public/Public";
```
DafyddLlyr, line 4: See previous comment about importing this from types.ts - avoids this slightly unexpected pattern.

editor.planx.uk/src/@planx/components/Filter/Public.tsx
```diff
@@ -8,17 +7,12 @@ export type Props = PublicProps<Filter>;
 
 // Filters are always auto-answered and never seen by a user, but should still leave a breadcrumb
 export default function Component(props: Props) {
-  const autoAnswerableFlag = useStore((state) => state.autoAnswerableFlag);
-
-  let idThatCanBeAutoAnswered: string | undefined;
-  if (props.id) idThatCanBeAutoAnswered = autoAnswerableFlag(props.id);
```
DafyddLlyr, line 14: Nice! 💪

editor.planx.uk/src/@planx/components/Question/Public/package.json
DafyddLlyr, line 1: Thanks for spotting this - this looks like just a weird artifact of somebody calling `pnpm i` in this folder that we've missed at some point?

editor.planx.uk/src/pages/Preview/Node.tsx
```diff
@@ -155,6 +173,13 @@ const Node: React.FC<Props> = (props) => {
         />
       );
 
+    case TYPES.Filter: {
+      const filterProps = getComponentProps<Filter>();
+      const autoAnswer = nodeId ? autoAnswerableFlag(nodeId) : undefined;
```
DafyddLlyr, line 178: nit: We could calculate this at the top of the file and add it to both `getComponentProps()` and `PublicProps<T>`. 

Slightly cleaner / more consistent, and we could put a check for component type within `autoAnswerableFlag()` if it's an expensive / slow function.

editor.planx.uk/src/pages/Preview/Node.tsx
```diff
@@ -221,34 +265,11 @@ const Node: React.FC<Props> = (props) => {
...
-        <PropertyInformationComponent
-          {...getComponentProps<PropertyInformation>()}
-        />
-      );
-
-    case TYPES.Filter:
-      return <FilterComponent {...getComponentProps<Filter>()} />;
-
     // These types are never seen by users, nor do they leave their own breadcrumbs entry
+    case TYPES.Answer:
```
DafyddLlyr, line 269: Always appreciate seeing some alphabetising!

DafyddLlyr (approved):
Great! Thanks for getting back to this - exactly as expected.

Two very minor comments, no showstoppers!

editor.planx.uk/src/@planx/components/Checklist/Public/Public.tsx
jessicamcinchak: Yep good call - updated !

PR #4113, opened 2025-01-07
feat: add `has_send_component` to flows table

RODO94:
## What does this PR do?

To support the work on organising flows, we want to try understand which services are Submission and which are Guidance. One tactic for this is to look for a `Send` component in the flow data. This would say that this services submits data rather than providing guidance. 

Linked to this Trello tick: https://trello.com/c/7av8lsVr/3160-epic-as-an-editor-i-want-to-have-a-way-to-organise-my-flows-so-that-i-can-easily-find-what-im-looking-for

### DIFF
--- a/hasura.planx.uk/migrations/1736253268324_add_has_data_to_flows/up.sql
+++ b/hasura.planx.uk/migrations/1736253268324_add_has_data_to_flows/up.sql
@@ -1,0 +1,13 @@
+alter table "public"."flows" add column "has_send_component" boolean
+ null default 'false';
+
+/* The Send component has an enum = 650 */
+UPDATE "public"."flows"
+SET "has_send_component" = true
+WHERE EXISTS (
+    SELECT 1 
+    FROM jsonb_each(data) as objects(key, value)
+    WHERE 
+        key NOT IN ('_root')
+        AND (value->>'type')::integer = 650
+);
--- a/hasura.planx.uk/migrations/1736253268324_add_has_data_to_flows/up.sql
+++ b/hasura.planx.uk/migrations/1736253268324_add_has_data_to_flows/up.sql
@@ -1,2 +1,60 @@
 alter table "public"."published_flows" add column "has_send_component" boolean
  default 'false';
+
+CREATE OR REPLACE FUNCTION public.diff_latest_published_flow(source_flow_id uuid, since timestamp with time zone)
+ RETURNS published_flows
+ LANGUAGE sql
+ STABLE
+AS $function$
+WITH current_published_flow as (
+  SELECT
+    id, data, created_at, flow_id, publisher_id, summary, has_send_component
+  FROM
+    published_flows
+  WHERE
+    published_flows.flow_id = source_flow_id
+  ORDER BY
+    created_at desc
+  LIMIT
+    1
+),
+previous_published_flow as (
+  SELECT
+    flow_id, data
+  FROM
+    published_flows
+  WHERE
+    published_flows.flow_id = source_flow_id
+    AND
+    published_flows.created_at < since
+  ORDER BY
+    created_at desc -- the latest published version before "since"
+  LIMIT
+    1
+),
+data_diff as (
+  SELECT
+    flow_id,
+    ( SELECT
+        jsonb_object_agg(COALESCE(old.key, new.key), new.value)
+      FROM
+        jsonb_each(previous_published_flow.data) AS old
+      FULL OUTER JOIN 
+        jsonb_each(current_published_flow.data) AS new 
+      ON
+        new.key = old.key 
+      WHERE 
+        new.value IS DISTINCT FROM old.value
+    ) as data -- shallow diff where deleted keys have a 'null' value
+  FROM 
+    current_published_flow
+  JOIN
+    previous_published_flow USING (flow_id)
+)
+SELECT 
+    id, data_diff.data as data, created_at, flow_id, publisher_id, 'auto generated diff' as summary, has_send_component
+FROM 
+    current_published_flow
+FULL OUTER JOIN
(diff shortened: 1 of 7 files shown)


hasura.planx.uk/migrations/1736253268324_add_has_data_to_flows/up.sql
```diff
@@ -0,0 +1,13 @@
+alter table "public"."flows" add column "has_send_component" boolean
+ null default 'false';
+
+/* The Send component has an enum = 650 */
+UPDATE "public"."flows"
+SET "has_send_component" = true
+WHERE EXISTS (
+    SELECT 1
```
RODO94, line 8: My assumption is that this means it will stop searching upon first find, rather than going through them all

scripts/seed-database/write/flows.sql
```diff
@@ -32,7 +33,7 @@ INSERT INTO flows (
   status,
   name,
   templated_from,
-  description
+  description, has_send_component
```
jessicamcinchak, line 36: nit: please maintain one-line-per-column pattern here for readability !

hasura.planx.uk/migrations/1736253268324_add_has_data_to_flows/up.sql
jessicamcinchak: While this `UPDATE` script does exactly what I expect it to locally, I'm not sure it makes sense to introduce here before we have frontend logic to update this column on "edit" or "publish" as well? Otherwise flows are going to fall out of sync in the meantime and this will need to be re-ran at some point anyways? 

Is "always false" more clear in the meantime while the total feature is WIP and for this script to be included in a future/next PR that includes frontend update logic too?

hasura.planx.uk/migrations/1736253268324_add_has_data_to_flows/up.sql
```diff
@@ -0,0 +1,13 @@
+alter table "public"."flows" add column "has_send_component" boolean
+ null default 'false';
```
jessicamcinchak, line 2: Question: is live flow data the correct place for this value to live or should it be associated with `published_flows` ?

I think there's a few benefits to it being on `published_flows`: 
- It allows "parent" flows that have a Send component in an external portal to be correctly marked as "submission" (because Send component is only found in flattened data, not `flows.data`)
- It saves us having to check on every graph operation if a Send component is being added/removed and update this column accordingly, instead we check & update once at point of publish
- It means the service type "submission" or "guidance" exposed to editors matches whatever the state of content currently available to the public is, rather than what their unpublished graph may include

Does this make sense?

RODO94: Aaahh yeah, I guess I was initially thinking it was easier so you can see WIP submission services, but completely missed that we're only updating on publish so it makes this thinking redundant. 

Agree with this and thanks for the comment!

hasura.planx.uk/migrations/1736253268324_add_has_data_to_flows/up.sql
RODO94: "Always false" makes sense here, I'll update the script now

hasura.planx.uk/migrations/1736253268324_add_has_data_to_flows/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."published_flows" add column "has_send_component" boolean
```
RODO94, line 1: the `default false` hopefully ensures it's false always and I don't need to populate everything

jessicamcinchak (approved):

hasura.planx.uk/migrations/1736253268324_add_has_data_to_flows/up.sql
```diff
@@ -0,0 +1,60 @@
...
+ default 'false';
+
+CREATE OR REPLACE FUNCTION public.diff_latest_published_flow(source_flow_id uuid, since timestamp with time zone)
+ RETURNS published_flows
+ LANGUAGE sql
+ STABLE
+AS $function$
+WITH current_published_flow as (
+  SELECT
+    id, data, created_at, flow_id, publisher_id, summary, has_send_component
```
jessicamcinchak, line 11: :+1:

hasura.planx.uk/migrations/1736253268324_add_has_data_to_flows/up.sql
jessicamcinchak: nit: I don't mean this to sound harsh, but I find comments like this quite confusing to review - are you confident that `default 'false'` works as you expect it to (based on your local env and how pizza built) or are you asking a question that you need a second opinion on? 

"Hopefully ensures" sounds like you're _not_ confident in what you've done here and also have not double-checked the pizza's `published_flows` table to confirm, which is not the right attitude for database migrations !!

jessicamcinchak (changes requested):
The pizza here has an empty database that has failed to sync - because this is a database migration, I would expect to please be able to check the Hasura console for new values !

hasura.planx.uk/migrations/1736253268324_add_has_data_to_flows/up.sql
RODO94: Nah not harsh at all! I was planning to come back with confirmation that it was all good, apologies for the nature of the changes / issues on this one. Thanks for giving feedback even when you think it's harsh, always really appreciate it

jessicamcinchak (approved):
Thanks for splitting up database migration and sync script here so we can doublecheck on pizza !

PR #4121, opened 2025-01-09
chore: bump planx-core

jessicamcinchak:
See https://github.com/theopensystemslab/planx-core/pull/614

**Testing notes:**
- I've copied an "old" session from Bucks RAB production into this pizza database to confirm that the overview document maps display correctly
- Login to the pizza editor to generate an auth token then open this admin endpoint in your browser: https://api.4121.planx.pizza/admin/session/12c48901-178d-445d-a550-783a36ba4bc1/html
  - Confirm that you have the subtext "The user accepted the title boundary" and a grey code block "boundary" populated
  - There's CORS errors in the console related the the OS maps display that are consistent with production that we can fix forward later (this seems to be specific to the admin endpoint and local downloads are fine)
  - But the presence of the subtext & "boundary" block are confirmation that variables are linking up again when compared to prod endpoint https://api.editor.planx.uk/admin/session/12c48901-178d-445d-a550-783a36ba4bc1/html

I've additionally confirmed that sessions started _and_ submitted today (so only new data fields present) are correctly displaying documents and maps - download `115c6638-f77a-40d4-a653-2dc1ea342214` from here to confirm https://editor.planx.uk/gateshead/report-a-planning-breach/submissions-log

### DIFF
--- a/api.planx.uk/package.json
+++ b/api.planx.uk/package.json
@@ -13,7 +13,7 @@
     "@airbrake/node": "^2.1.8",
     "@aws-sdk/client-s3": "^3.696.0",
     "@aws-sdk/s3-request-presigner": "^3.701.0",
-    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#dc2386b",
+    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#e126742",
     "@types/isomorphic-fetch": "^0.0.36",
     "adm-zip": "^0.5.10",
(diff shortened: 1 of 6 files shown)


DafyddLlyr (approved):
Tested following the description in the PR (thanks!) and getting expected results.

Agree that the CORS issue is unrelated - it's possibly been there since day 1 of the `/html` route and will only effect PlanX devs testing.

These files were intended for local consumption so CORS aren't a consideration. If I download the file the map works as expected.

PR #4123, opened 2025-01-09
fix: add guard clause for flow[nodeId] in autoAnswerableOptions

RODO94:
Running through the session and recreating it locally, we found the issue to be in the `autoAnswerableOptions` store function.

The received error locally was:
Screenshot 2025-01-09 at 12 05 50

After some logging we found the issue lying in `flow[nodeId]` being `undefined` in the Filter function. 

@jamdelion and I did find it hard to track why flow[nodeId] could be undefined for a save and return path. We thought it would be due to the node being deleted, but that wouldn't be consistent with the way published routes are handled for save and return flows (with the user always returning to the published flow they left from), so we were stumped here.

>[!IMPORTANT]
> What is the root cause of `flow[nodeId]` being `undefined`?

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
@@ -223,19 +223,27 @@ export const previewStore: StateCreator<
           if (passportValue.length > 0) {
             const existingValue = acc.data?.[key] ?? [];
 
-            const combined = key === planningConstraintsFn 
-              ? passportValue.concat(existingValue) // Planning constraints uniquely store all-levels of granularity, rather than most granular only
-              : passportValue
-              .concat(existingValue)
-              .reduce(
-                (acc: string[], curr: string, _i: number, arr: string[]) => {
-                  if (!arr.some((x) => x !== curr && x?.startsWith(curr))) {
-                    acc.push(curr);
-                  }
-                  return acc;
-                },
-                [],
-              );
+            const combined =
+              key === planningConstraintsFn
+                ? passportValue.concat(existingValue) // Planning constraints uniquely store all-levels of granularity, rather than most granular only
+                : passportValue
+                    .concat(existingValue)
+                    .reduce(
+                      (
+                        acc: string[],
+                        curr: string,
+                        _i: number,
+                        arr: string[],
+                      ) => {
+                        if (
+                          !arr.some((x) => x !== curr && x?.startsWith(curr))
+                        ) {
+                          acc.push(curr);
+                        }
+                        return acc;
+                      },
+                      [],
+                    );
 
             passportData[key] = uniq(combined);
           }
@@ -414,8 +422,7 @@ export const previewStore: StateCreator<
         }));
         const hidden = !selections.some(
           (selection) =>
-            selection.data?.flags &&
-            selection.data.flags.includes(flag?.value)
+            selection.data?.flags && selection.data.flags.includes(flag?.value),
         );
 
         return {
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -520,10 +520,13 @@ export const previewStore: StateCreator<
 
     // Only proceed if the user has seen at least one node with this fn before
     const visitedFns = Object.entries(breadcrumbs).filter(
-      ([nodeId, _breadcrumb]) =>
-        flow[nodeId].data?.fn === data.fn ||
+      ([nodeId, _breadcrumb]) =>{
+        if (!flow[nodeId]) {
+          return;
+        }
+        return flow[nodeId].data?.fn === data.fn ||
```
jessicamcinchak, line 527: When written this way, it makes it impossible for the `||` clause to be hit (which does not rely on the breadcrumb node still existing in the flow). 

Would a 1-line fix like `flow[nodeId]?.data?.fn === data.fn ||` accomplish the same while still allowing the second "or" clause to be hit?

jessicamcinchak:
> with the user always returning to the published flow they left from

When resuming, the user returns the _most recent published version_ of their flow, NOT the version they started on/left from. This is handled by what we call a "reconciliation" process. If we let everyone resume to the version of the flow they started from, there w wouldn't be a need for reconciliation, but there would be the potential to submit invalid/out-of-date applications if legislation changed over your 28 days.

Therefore, it's valid and expected that their may be a breadcrumb which no longer has an associated node in the flow. 

We have a function `removeOrphansFromBreadcrumbs` which I believe we call before processing submission data, but not on each "continue"/`record()` (from memory! would need to double check this!).

jessicamcinchak (commented):
I think this change is correct (automation tests are all passing), but I am a bit confused re-creating session and want to double check my work against what you all did:
- I've copied the session to this pizza via Hasura console
- I can resume a session here using my OSL email https://4123.planx.pizza/barnet/apply-for-a-lawful-development-certificate/published?analytics=false&sessionId=0726e705-7e81-4120-b67c-93c2859a1095
- I see the "Resume your application" landing page, all my sections are complete until "Upload your documents", I can "continue" and answer the first question about materials
- But then when I continue onwards from that first question post-resume I get a blank screen and this console error:
Screenshot from 2025-01-09 15-07-29
- The `currentCard` of the blank screen is `uMU15fLt77`
- When testing yourselves, were you able to proceed successfully through more than one question after resuming?

RODO94:
> I think this change is correct (automation tests are all passing), but I am a bit confused re-creating session and want to double check my work against what you all did:
> 
> * I've copied the session to this pizza via Hasura console
> * I can resume a session here using my OSL email https://4123.planx.pizza/barnet/apply-for-a-lawful-development-certificate/published?analytics=false&sessionId=0726e705-7e81-4120-b67c-93c2859a1095
> * I see the "Resume your application" landing page, all my sections are complete until "Upload your documents", I can "continue" and answer the first question about materials
> * But then when I continue onwards from that first question post-resume I get a blank screen and this console error:
>   Screenshot from 2025-01-09 15-07-29
> * The `currentCard` of the blank screen is `uMU15fLt77`
> * When testing yourselves, were you able to proceed successfully through more than one question after resuming?

Good catch!

I just tried it now and it didn't work due to this line:

```ts
   visitedFns.forEach(([nodeId, _breadcrumb]) => {
      flow[nodeId].edges?.map((edgeId) => {
        if (flow[edgeId].type === TYPES.Answer && flow[edgeId].data?.val) {
          visitedOptionVals.push(flow[edgeId].data.val);
        }
      });
    });
```

I'll repeat the fix here for `flow[nodeId]`

jessicamcinchak (approved):
Works as expected now, thanks!

PR #4124, opened 2025-01-09
chore: bump planx-core in Editor

RODO94:
Bumped planx-core in line with the API and E2E here:

https://github.com/theopensystemslab/planx-new/pull/4121


jessicamcinchak (approved):
Thanks for catching

PR #4128, opened 2025-01-09
fix: Toggle button visual order

ianjon3s:
## What does this PR do?

- Corrects visual order of editor graph toggle buttons
- Adds `aria-pressed` attribute to ensure correct state is visible in the DOM


jessicamcinchak (approved):
Working exactly as expected now, thanks for quick fix!

PR #4129, opened 2025-01-09
feat: Read dbUrl from config file, not remote pulumi

DafyddLlyr:
- After restoring the production db to a point in time 9/1/25, we need to point to a new db instance
- This is currently controlled via an output from the data layer (`dbRootUrl`)
- Rather than run this layer and try to get it to update and point to the new db, we're just statically importing the db root url from the pulumi config file now

This does not affect any workflows as we don't have an automated link between the data and application layer.


jessicamcinchak (approved):
:muscle:

PR #4137, opened 2025-01-10
feat: add select for sorting flow list

RODO94:
## What does this PR do?

Working towards this EPIC: https://trello.com/c/7av8lsVr/3160-epic-as-an-editor-i-want-to-have-a-way-to-organise-my-flows-so-that-i-can-easily-find-what-im-looking-for

The main task was to add the ability to Sort flows in the Team page, the page which shows a Team's flows. 

### Approach

The Team page component is quite long, so I created a new component separate from this to handle the logic of sorting and also contain a Select component for users to interact with. 

This new component takes in state variables `flows` and `setFlows` from the Team page component, which felt right for me so the flow list content always lives in one place but is given to other components to augment. I'd want to repeat this pattern for filtering the flows.

My hope for the sort component  was to provide a maintainable, scalable component for sorting the list of flows on the Team page using keys we receive from `getFlows()` contained within the `FlowSummary` interface. It may need to be used across many different tables, so I have enabled a generic Type to be added when the component is used, so it will now received a `record` as an Array, a way of setting that record `setRecord`, and options for sorting the Array `sortOptions`. This should enable the Type safety I want for this specific use case while extending it for future ones

See this as just the building blocks of the sort component I can add into Ian's work on the new Team page and filter. 

### What's included?

Right now it covers: Name, Last Published, and Last Updated. I added an ascending or descending toggle as well. 

In order to make this work, I have extended the `getFlows()` query to pull in `published_flows` data and `status` data from Hasura. 

Future Improvement:

- If more items have to be added to the search params we could create a hook for it (useSearchParamState())

>[!IMPORTANT]
> This component is hidden using a FeatureFlag - `SORT_FLOWS`


editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -390,6 +400,13 @@ export const editorStore: StateCreator<
                 lastName: last_name
               }
             }
+            publishedFlows: published_flows(
+              order_by: { created_at: desc }
+              limit: 1
```
RODO94, line 405: Seen this was only used by the Team page, so though it felt better to expand this initial query

editor.planx.uk/src/ui/editor/SortFlowsSelect.tsx
```diff
@@ -0,0 +1,123 @@
...
+export const SortFlowsSelect = ({
+  flows,
+  setFlows,
+}: {
+  flows: FlowSummary[];
+  setFlows: React.Dispatch<React.SetStateAction<FlowSummary[] | null>>;
+}) => {
+  const [sortBy, setSortBy] = useState<SortObject>(sortArray[0]);
+  const [sortDirection, setSortDirection] = useState<SortDirection>("asc");
+  useEffect(() => {
```
RODO94, line 67: Went with a useEffect here to provide stability for adding more sorting types, would be interested to know if there's cleaner techniques

editor.planx.uk/src/ui/editor/SortFlowsSelect.tsx
```diff
@@ -0,0 +1,123 @@
...
+    if (sortKey === "publishedFlows") {
+      const sortedFlows = flows?.sort((a: FlowSummary, b: FlowSummary) => {
+        const { nestedKey } = sortBy;
+
+        // auto sort unpublished flows to bottom
+        if (!a[sortKey][0]) return 1;
+        if (!b[sortKey][0]) return -1;
+
+        const aValue = a[sortKey][0][nestedKey];
+        const bValue = b[sortKey][0][nestedKey];
```
RODO94, line 79: Although this possibly looks overly complicated, it allowed me to use the type safety to build the `sortArray` while not having to map sorting names to data fields with ternaries or conditional statements. 

I iterated this a bit and found that the extra code came a lot from the nested keys in the `flows` array, but the extra lines always provided better maintainability for adding or removing sorting criteria while remaining aligned with what data is provided to us here

DafyddLlyr (changes requested):
A few general comments on the code that might simplify things a lot here.

However, there's a more general design concept to consider here. We know that we're going to need this sort of sort component in a few places fairly soon. The logic of the component is very closely coupled with the incoming data - I'd recommend splitting these up.

Think about a pattern like this - 

- A "sort" component which takes in a generic `T[]` - the items to be sorted
- It can call a callback function to update state when the user interacts with the object
- Sort keys have to be `keyof typeof T` (I think type-fest allows deep keys also)
- We pass in a list of flows, and the setFlows function, and a list of keys

We now would have a component which could easily be used for the feedback, submission and flows pages with no extra code or unit tests required.

Right now, it would have a static UI (this is actually fine) but a future enhancement could to make it a bit more headless or to have variants (e.g. "compact")

Super happy to talk this through tomorrow if you'd like 👍

editor.planx.uk/src/ui/editor/SortFlowsSelect.tsx
```diff
@@ -0,0 +1,152 @@
...
+    navigation.navigate(
+      {
+        pathname: window.location.pathname,
+        search: `?sort=${sortKey}`,
+      },
+      {
+        replace: true,
+      },
+    );
+  };
```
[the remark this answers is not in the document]
DafyddLlyr: Just seen this is listed still a to do on the PR description 👌

editor.planx.uk/src/ui/editor/SortFlowsSelect.tsx
```diff
@@ -0,0 +1,152 @@
+import TrendingDownIcon from "@mui/icons-material/TrendingDown";
+import TrendingUpIcon from "@mui/icons-material/TrendingUp";
```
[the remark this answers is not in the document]
RODO94: Ah yeah! This was mostly for Ian to see it working and then we'll think about how to add it to the sort. Good feedback to have in tho

editor.planx.uk/src/ui/editor/SortFlowsSelect.tsx
```diff
@@ -0,0 +1,152 @@
...
+  sortKey: "publishedFlows";
+  nestedKey: keyof PublishedFlowSummary;
+}
+
+type SortObject = PublishedFlowSort | BasicSort;
+
+const sortArray: SortObject[] = [
+  { displayName: "Name", sortKey: "name" },
+  { displayName: "Last updated", sortKey: "updatedAt" },
+  { displayName: "Status", sortKey: "status" },
```
jessicamcinchak, line 37: nit: in additional to all of Daf's comments, a small thing here is that I don't believe `status` should be an attribute included in the "sort", but rather "filter" in the future (eg because it has categorical values)

editor.planx.uk/src/ui/editor/SortFlowsSelect.tsx
```diff
@@ -0,0 +1,152 @@
...
+};
+
+export const SortFlowsSelect = ({
+  flows,
+  setFlows,
+}: {
+  flows: FlowSummary[];
+  setFlows: React.Dispatch<React.SetStateAction<FlowSummary[] | null>>;
+}) => {
+  const [sortBy, setSortBy] = useState<SortObject>(sortArray[0]);
```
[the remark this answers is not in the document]
RODO94: I've taken the approach of slugifying the displayName and using that as a lookup key, since I had to keep the sortKey as part of an object connected with the name we use for displaying it in the select. 

There's a chance here that I've Wiley Coyote'd myself and ran off a cliff when the answer was much simpler, but it met what I wanted in terms of how you can define your sort options and how they are displayed in one place, and the component handles the logic of sorting / finding them.

editor.planx.uk/src/ui/editor/SortFlowsSelect.tsx
```diff
@@ -0,0 +1,152 @@
...
+  PublishedFlowSummary,
+} from "pages/FlowEditor/lib/store/editor";
+import React, { useEffect, useState } from "react";
+import { useNavigation } from "react-navi";
+
+import SelectInput from "./SelectInput/SelectInput";
+
+type SortDirection = "asc" | "desc";
+type SortKeys = keyof FlowSummary;
+type SortNestedKeys = keyof PublishedFlowSummary | keyof FlowSummaryOperations;
```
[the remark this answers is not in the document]
RODO94: This was a great one and I did have to add `type-fest` to the editor to get the method

editor.planx.uk/src/pages/Team.tsx
```diff
@@ -351,6 +364,13 @@ const Team: React.FC = () => {
         </Box>
         {showAddFlowButton && <AddFlowButton flows={flows} />}
       </Box>
+      {hasFeatureFlag("SORT_FLOWS") && flows && (
+        <SortControl<FlowSummary>
+          records={flows}
+          setRecords={setFlows}
+          sortOptions={sortArray}
```
RODO94, line 371: Switched naming to something more generic and reusable across the comp and remove refs to `flow` in the naming as well

editor.planx.uk/src/ui/editor/SortControl.tsx
```diff
@@ -0,0 +1,126 @@
...
+  sortOptions: SortableFields<T>[];
+}) => {
+  const [selectedSort, setSelectedSort] = useState<SortableFields<T>>(
+    sortOptions[0],
+  );
+  const [sortDirection, setSortDirection] = useState<SortDirection>("asc");
+
+  const navigation = useNavigation();
+  const route = useCurrentRoute();
+  const selectedDisplaySlug = slugify(selectedSort.displayName);
```
RODO94, line 51: This is probably the part I feel most unsure about, but I'm comfortable that it delivers a unifying parameter I can use to search for `sortOptions` while adding it to the search params in a nice way for the user to see and read. My issue lay in the lack of readability in the key naming for deeply nested keys, and the fact `displayName` could really contain any sort of spaces and the like.

editor.planx.uk/src/ui/editor/SortControl.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  const sortOptionsMap = useMemo(() => {
```
RODO94, line 53: UseMemo() here so it only runs when things change rather than on each re-render, and it will produce a lookup Object based on the slugified `displayName`. 

This should replace the Array.find() methods I had in the code

editor.planx.uk/src/ui/editor/SortFlowsSelect.tsx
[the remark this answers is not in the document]
DafyddLlyr: I think we've had it in and out a few times as a dependency, we also use it in the API and I _think_ planx-core.

editor.planx.uk/src/ui/editor/SortControl.tsx
DafyddLlyr: Yeah it's a good question - `fieldName` is guaranteed to be unique, but not necessarily readable. 

There's no guarantee that `slugify(selectedSort.displayName)` doesn't return duplicates through right? It's unlikely but possible.

This might just be a data structure question. If we were to make `sortOptions` something like this our problems go away - 

```ts
sortOptions: Record<string, SortableFields<T>>

...

  const sortArray: Record<string, SortableFields<FlowSummary>> = {
    name: { displayName: "Name", fieldName: "name" },
    lastUpdated: { displayName: "Last updated", fieldName: "updatedAt" },
    lastPublished: {
      displayName: "Last published",
      fieldName: `publishedFlows.0.publishedAt`,
    },
  }
  
  
```

This would - 
 - Force a unique key
 - Provide us with a map already
 
 What do you think?

editor.planx.uk/src/ui/editor/SortControl.tsx
DafyddLlyr: This `reduce()` just looks a bit like a complex `groupBy()`

```ts
Object.groupBy(sortOptions, ({ fieldName }) => fieldName)
```

DafyddLlyr: https://x.com/jaffathecake/status/1213077702300852224 ;)

editor.planx.uk/src/ui/editor/SortControl.tsx
RODO94: Yeah this was an iteration I considered, I was thinking this is dealing with a complexity which is within the component (so not obvious when you add it to your table comp) so it is better to handle it within the component? 

Imagining someone building this into their component and being confused why you have to add so many names, ultimately the same thing three times. 

I can see the risk of duplicates but thought it was quite edge casey and resolvable. 

But, I'm happy to integrate the change, I can see the benefits of the safety, so you don't need to try find your error in the naming.

DafyddLlyr: Yeah this is a fair point - lets aim to keep the complexity within the component I think 👍 

Your version is clearer and simpler for consumers. The risk of duplicates is very very low!

editor.planx.uk/src/pages/Team.tsx
```diff
@@ -351,6 +364,13 @@ const Team: React.FC = () => {
         </Box>
         {showAddFlowButton && <AddFlowButton flows={flows} />}
       </Box>
+      {hasFeatureFlag("SORT_FLOWS") && flows && (
```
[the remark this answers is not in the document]
RODO94: I'll add this in when I combine it with the Filters block

editor.planx.uk/src/pages/Team.tsx
```diff
@@ -311,15 +308,23 @@ const Team: React.FC = () => {
...
   const [flows, setFlows] = useState<FlowSummary[] | null>(null);
 
-  const sortArray: SortableFields<FlowSummary>[] = [
-    { displayName: "Name", fieldName: "name" },
-    { displayName: "Last updated", fieldName: "updatedAt" },
+  const sortOptions: SortableFields<FlowSummary>[] = [
+    {
+      displayName: "Name",
+      fieldName: "name",
+      directionNames: { asc: "A - Z", desc: "Z - A" },
```
DafyddLlyr, line 315: Nice! When we have more use cases for this we _may_ find we want these display values centralised within the component, and just allow consumers to define by type (e.g. `direction: "alphabetical" | "date"`). This is perfect for now 👍

DafyddLlyr (approved):
Thanks - changes here look great, and we've ended up with a solution which is much for robust and expandable - love to see it!

Not manually tested due to failing CI (due to cache issue - please run `gh cache delete --all` and then re-run CI).

RODO94:
@DafyddLlyr manual testing is probably best with the filters included

PR #4142, opened 2025-01-13
chore: Bump planx-core

DafyddLlyr:
Resolves https://opendigitalplanning.slack.com/archives/C0241GWFG4B/p1734625014177729 (ODP Slack)


RODO94 (approved):
Looks good to me, updates in all the right files and all contain the same hash

PR #4144, opened 2025-01-13
fix: grouped checklists bug fixes and basic editor-side test

jamdelion:
### In this PR:

- Split out `Public.test.tsx` into two files so that grouped and non-grouped are separated
- Fix a couple of bugs that August spotted - one to do with deleting the exclusive 'or' option, and one to do with re-ordering the options: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1736356593283979?thread_ts=1736175896.631909&cid=C5Q59R3HB
- Remove the feature flag `EXCLUSIVE_OR`(makes this live for councils! ⚠️ )
- Add Editor side tests as there was no coverage here. 
- Add `aria-describedby` to flag to users navigating with a screen reader that other options will be deselected when the exclusive "or" option is selected. I couldn't find any a11y examples doing anything else, but if you can think of a better solution I'd be v grateful to hear it!


editor.planx.uk/src/@planx/components/Checklist/Editor/Options.tsx
```diff
@@ -35,7 +35,7 @@ export const Options: React.FC<{ formik: FormikHookReturn }> = ({ formik }) => {
               const newCombinedOptions =
                 newOptions.length === 0
                   ? []
-                  : [...exclusiveOptions, ...newOptions];
+                  : [...newOptions, ...exclusiveOptions];
```
jamdelion, line 38: This was the cause of the re-ordering bug 😅

editor.planx.uk/src/@planx/components/Checklist/Editor/Editor.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    await waitFor(() => user.click(screen.getByLabelText("Expandable")));
+
+    const groupedOptionsEditor = await screen.findByPlaceholderText(
+      "Section Title",
+    );
+    expect(groupedOptionsEditor).toBeInTheDocument();
+  });
+
+  it.todo(
```
jessicamcinchak, line 39: Appreciate groundwork for `Editor`-side test coverage here! 

I wonder if the click issue you describe is perhaps a role-based permissions error failing silently/lazily (eg view-only modal but we haven't actually disabled buttons etc)? Have you tried to additionally mock a user with editing privileges like here https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/src/%40planx/components/FileUploadAndLabel/Editor.test.tsx ?

jamdelion: Thanks, that was totally the issue! 😍 Have added that `beforeEach` now, and it fixed the click problem 🌟

jessicamcinchak (approved):
Pizza working as expected for me! One minor comment re: Gov UK alignment

Also one small final style nit that stuck out to me: is it possible/easy to hide the "drag" indicator for the "or" option to better indicate it's position is static/always last?
Screenshot from 2025-01-20 10-48-25

editor.planx.uk/src/@planx/components/Checklist/Public/components/ExclusiveChecklistItem.tsx
```diff
@@ -18,14 +20,18 @@ export const ExclusiveChecklistItem = ({
...
+      </Box>
+      <p id="exclusive-option-explanation" style={visuallyHidden}>
+        Selecting this option will deselect all other options.
+      </p>
       <ChecklistItem
         onChange={changeCheckbox(exclusiveOrOption.id)}
         label={exclusiveOrOption.data.text}
         id={exclusiveOrOption.id}
         checked={formik.values.checked.includes(exclusiveOrOption.id)}
+        inputProps={{ "aria-describedby": "exclusive-option-explanation" }}
```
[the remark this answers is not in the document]
jamdelion: Good thought! But it wasn't allowed as an inputProp by the typing (which comes from `React.InputHTMLAttributes<HTMLInputElement>`). It didn't seem to take affect when trying to add as a prop in the `ChecklistItem` rather than inputProps so have not implemented this.

jamdelion:
> Pizza working as expected for me! One minor comment re: Gov UK alignment
> 
> Also one small final style nit that stuck out to me: is it possible/easy to hide the "drag" indicator for the "or" option to better indicate it's position is static/always last? Screenshot from 2025-01-20 10-48-25

Thanks for thoughts on this one. I originally held off as didn't want to interfere with this too much but actually only the GovPayMetadata section uses the `disableDragAndDrop` prop, and I think it is fine without the greyed out drag handle too: 

Screenshot 2025-01-20 at 11 55 19

So have fully deleted the dragHandle for cases when that disable prop is set, and renamed to `noDragAndDrop` to better reflect what it looks like.

PR #4146, opened 2025-01-14
fix: Add optional chaining to flag check

DafyddLlyr:
image

This should address an error raised by Greg (via email).

I haven't actually worked out the underlying issue here yet (why there's a node without data) but I'm currently looking into this by working through the session's breadcrumbs.


jessicamcinchak (approved):
Thank you! 

This method recently got repeated over in planx-core here which we'll also want to update to ensure that "result" in documents is also able to be generated successfully: https://github.com/theopensystemslab/planx-core/pull/611#discussion_r1905417112

PR #4149, opened 2025-01-14
fix: display `fn` validation errors in Question and Checklist editor modals

jessicamcinchak:
Per August's feedback here: https://opensystemslab.slack.com/archives/C01E3AC0C03/p1736435506266819

Fixes a small visual regression from introducing the `DataFieldAutocomplete`. The data field validation was still running and preventing invalid nodes from being created, but failing to visually display an error message. Now added back:
Screenshot from 2025-01-14 18-13-45

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/BaseOptionsEditor.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/BaseOptionsEditor.tsx
@@ -76,8 +76,9 @@ export const BaseOptionsEditor: React.FC<BaseOptionsEditorProps> = (props) => (
     )}
     {props.showValueField && (
       <DataFieldAutocomplete
+        key={`${props.value.id}-data-field-autocomplete`}
         schema={props.schema}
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/shared/BaseOptionsEditor.tsx
```diff
@@ -76,8 +76,9 @@ export const BaseOptionsEditor: React.FC<BaseOptionsEditorProps> = (props) => (
     )}
     {props.showValueField && (
       <DataFieldAutocomplete
+        key={`${props.value.id}-data-field-autocomplete`}
```
jessicamcinchak, line 79: This clears up another error where if there was an option(s) with a null data `val`, then the autocomplete inputs could unexpectedly shift or nullify when being dragged into a different order - see before: 
chrome-capture-2025-0-17

DafyddLlyr (approved):

PR #4150, opened 2025-01-14
feat: Metabase dashboard service and controller

zz-hh-aa:
# What does this PR do?
- Creates services `getDashboard`, `copyDashboard`, `updateFilter` and `generatePublicLink`
- Adds necessary types and tests
- Creates new route for analytics module
- Exports the Metabase client as a singleton instance (this also happens in [8f12f20](https://github.com/theopensystemslab/planx-new/pull/4072/commits/8f12f205708fb03b802f8f9330fce54319a21f93) in #4072 but there were other changes in the same commit so I've isolated the change here)
- Updates Swagger yaml

# Why?
Following on from #4072, this should complete all the services we need to automate Metabase dashboard creation

This PR became a bit bigger than I'd anticipated... let me know if it'd be better to break it up!

### DIFF
--- a/api.planx.uk/modules/analytics/docs.yaml
+++ b/api.planx.uk/modules/analytics/docs.yaml
@@ -22,6 +22,33 @@ components:
           type: integer
           description: Optional ID of the parent collection
 
+    NewDashboard:
+      type: object
+      properties:
+        templateId:
+          type: integer
+          description: ID of the template dashboard to copy from
+        description:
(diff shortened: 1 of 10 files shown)


DafyddLlyr:
Removed Vultr instance for this PR - please see https://opensystemslab.slack.com/archives/C01E3AC0C03/p1737021735567669 (OSL Slack)

PR #4152, opened 2025-01-14
fix: never fully remove `_root` node via graph operations

jessicamcinchak:
Per bug spot by August https://opensystemslab.slack.com/archives/C01E3AC0C03/p1736354419439469

When removing the last/final node in the graph currently, the flow data will be set to `{}` - which then on next render will display a blank screen & fail to load because of type errors:
```
index-vd9rjCRb.js:614 
TypeError: Cannot read properties of undefined (reading 'edges')
    at Object.filterFlowByType (index-vd9rjCRb.js:1565:3082)
    at Object.initNavigationStore (index-vd9rjCRb.js:1565:2543)
    at Object.setFlow (index-vd9rjCRb.js:1610:741)
    at n (index-vd9rjCRb.js:1500:92956)
    at Object.connectTo (index-vd9rjCRb.js:1500:93024)
    at async team-BRe-K93g.js:107:189
```

This change ensures that removing the last node still preserves the original `{ "_root": { edges: [] } }` flow structure to prevent type errors and not disrupt the editing workflow.

### DIFF
--- a/editor.planx.uk/src/@planx/graph/__tests__/remove.test.ts
+++ b/editor.planx.uk/src/@planx/graph/__tests__/remove.test.ts
@@ -26,7 +26,7 @@ test("with clones", () => {
   ]);
 });
 
-test("with id", () => {
+test("final node with id", () => {
   const [graph, ops] = remove(
     "a",
     "_root",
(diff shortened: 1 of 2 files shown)


DafyddLlyr (approved):
Thanks for getting round to this one!

PR #4153, opened 2025-01-15
feat(types): Stricter and more accurate Schema response types

DafyddLlyr:
## What does this PR do?
 - Correctly defines output values of a schema field, based on it's field type. This is the `ResponseValue<T>` type.
   - Please see https://github.com/theopensystemslab/planx-new/pull/4153#discussion_r1919809762 for details on why this is the approach I've gone for
 - Updates code that was relying on type casting, or type checking via disparate means, to use a standard set of type guards defined in the model file

## Why?
- Overdue technical debt from the list/schema work
- A barrier for https://github.com/theopensystemslab/planx-new/pull/4020 which is introducing another field and response value. One this PR is merged I'll rebase the above branch to incorporate this and make the required type changes there.

### DIFF
--- a/editor.planx.uk/src/@planx/components/List/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/List/Public/index.tsx
@@ -8,7 +8,7 @@ import TableBody from "@mui/material/TableBody";
 import TableCell from "@mui/material/TableCell";
 import TableRow from "@mui/material/TableRow";
 import Typography from "@mui/material/Typography";
-import { MapField, ResponseValue } from "@planx/components/shared/Schema/model";
+import { isMapFieldResponse } from "@planx/components/shared/Schema/model";
 import { SchemaFields } from "@planx/components/shared/Schema/SchemaFields";
 import { PublicProps } from "@planx/components/shared/types";
 import React, { useEffect, useRef } from "react";
@@ -129,7 +129,17 @@ const InactiveListCard: React.FC<{
 }> = ({ index: i }) => {
   const { schema, formik, removeItem, editItem } = useListContext();
 
-  const mapPreview = schema.fields.find((field) => field.type === "map");
+  const formattedMapResponse = (() => {
+    const mapField = schema.fields.find((field) => field.type === "map");
+    if (!mapField) return null;
+
+    const mapResponse = formik.values.schemaData[i][mapField.data.fn];
+    if (!mapResponse) return null;
+
+    return isMapFieldResponse(mapResponse)
+      ? formatSchemaDisplayValue(mapResponse, mapField)
+      : null;
+  })();
 
   return (
     <ListCard data-testid={`list-card-${i}`}>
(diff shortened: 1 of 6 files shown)


editor.planx.uk/src/@planx/components/shared/Schema/model.ts
```diff
@@ -118,10 +118,21 @@ export interface Schema {
...
+/**
+ * Value returned per field, based on field type
+ */
+export type ResponseValue<T extends Field> = T extends MapField
+  ? Feature[]
+  : T extends ChecklistField
+  ? string[]
+  : T extends NumberField
+  ? number
+  : string;
```
DafyddLlyr, line 130: Unfortunately the only way to make these conditional types are long ternaries like this.

There is an alternative of using [a mapped type](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html), but I tried a few variations on this and to me it read as more complex than this.

editor.planx.uk/src/@planx/components/List/utils.test.ts
```diff
@@ -63,7 +62,7 @@ describe("passport data shape", () => {
           identicalUnits: 2,
         },
       ],
-    } as unknown as Record<string, SchemaUserResponse[]>;
```
DafyddLlyr, line 66: A nice outcome here is that this is no longer required 👍

editor.planx.uk/src/@planx/components/List/utils.tsx
```diff
@@ -15,19 +24,27 @@ const List = styled("ul")(() => ({
...
-  field: Field,
-) {
+export const formatSchemaDisplayValue = <T extends Field>(
+  value: ResponseValue<T>,
+  field: T,
+) => {
   switch (field.type) {
-    case "number":
+    case "number": {
+      if (!isNumberFieldResponse(value)) return;
```
DafyddLlyr, line 33: These early `return`s are really just here for type narrowing and won't actually be hit- we know that a schema field will always match a response's field.

editor.planx.uk/src/@planx/components/List/Public/index.tsx
```diff
@@ -128,21 +129,26 @@ const InactiveListCard: React.FC<{
...
+    const mapField = schema.fields.find((field) => field.type === "map");
+    if (!mapField) return null;
+
+    const mapResponse = formik.values.schemaData[i][mapField.data.fn];
+    if (!mapResponse) return null;
+
+    return isMapFieldResponse(mapResponse)
+      ? formatSchemaDisplayValue(mapResponse, mapField)
+      : null;
+  })();
```
jessicamcinchak, line 142: :+1:

editor.planx.uk/src/@planx/components/shared/Schema/model.ts
```diff
@@ -130,6 +141,27 @@ export type SchemaUserData = {
...
+export const isMapFieldResponse = (
+  response: unknown,
+): response is ResponseValue<MapField> =>
+  Array.isArray(response) && response[0]?.type === "Feature";
+
+export const isChecklistFieldResponse = (
+  response: unknown,
+): response is ResponseValue<ChecklistField> =>
+  Array.isArray(response) && !isMapFieldResponse(response);
+
```
jessicamcinchak, line 164: :+1:

jessicamcinchak (approved):
Really appreciate you picking this one up, more invovled than I anticipated but this seems both robust and still easy to read/define :star2: 

You're more than welcome to rebase my AddressInput PR and carry forward if you've got momentum, or I'm happy to take from here !

PR #4156, opened 2025-01-16
feat: populate database for `has_send_comp` and add to api mutation

RODO94:
## What does this PR do?

As part of the Filter and Sorting epic: https://trello.com/c/7av8lsVr/3160-epic-as-an-editor-i-want-to-have-a-way-to-organise-my-flows-so-that-i-can-easily-find-what-im-looking-for

I have added a populate script to migrations, and added an extra field within the API when a flow is published to set `has_send_component` to the result of:

```ts
const hasSendComponent = hasComponentType(flattenedFlow, ComponentType.Send);
```

This will support the work to define whether a service is of type Submission or not

### DIFF
--- a/hasura.planx.uk/migrations/1736953485265_populate published_flows has_send_component/down.sql
+++ b/hasura.planx.uk/migrations/1736953485265_populate published_flows has_send_component/down.sql
@@ -1,0 +1,11 @@
+-- Could not auto-generate a down migration.
+-- Please write an appropriate down migration for the SQL below:
+-- UPDATE "public"."published_flows"
+-- SET has_send_component = true
+-- WHERE EXISTS (
+--     SELECT 1
+--     FROM jsonb_each(data) as objects(key, value)
+--     WHERE
+--         key NOT IN ('_root')
+--         AND (value->>'type')::integer = 650
+-- );
--- a/hasura.planx.uk/migrations/1736953485265_populate published_flows has_send_component/down.sql
+++ b/hasura.planx.uk/migrations/1736953485265_populate published_flows has_send_component/down.sql
@@ -1,11 +1,2 @@
--- Could not auto-generate a down migration.
--- Please write an appropriate down migration for the SQL below:
--- UPDATE "public"."published_flows"
--- SET has_send_component = true
--- WHERE EXISTS (
---     SELECT 1
---     FROM jsonb_each(data) as objects(key, value)
---     WHERE
---         key NOT IN ('_root')
---         AND (value->>'type')::integer = 650
--- );
+UPDATE "public"."published_flows"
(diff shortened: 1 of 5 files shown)


hasura.planx.uk/migrations/1736953485265_populate published_flows has_send_component/up.sql
DafyddLlyr, line 1: Running this could be pretty brutal 😅 Not sure if this will cause a timeout or other issues of prod (how is it running locally?)

We probably only care about correctly setting the most recent published flow to have this value set correctly. I guess the idea is to show the tag, we need to join flow > most recent published flow right?

DafyddLlyr (approved):
Looks good - once question/comment on how this migration will go.

You could test this locally by dropping the `DISTINCT` constraint from the `published_flows` sync script, pulling in all records and testing - maybe I'm worrying about nothing!

DafyddLlyr (changes requested):
Sorry - accidentally marked as approved - wrong radio. We should test this before running on prod data.

hasura.planx.uk/migrations/1736953485265_populate published_flows has_send_component/up.sql
RODO94: Yeah that would be right actually, I tested it locally but in the Hasura SQL console and it ran pretty quick... 

I think adding some distinct guard here could help it tho, the `published_flows` table is a bit of a beast

RODO94: Updated both scripts and tested locally with full production `published_flows` table. The down command was rapid and there was not much change in the performance of the up script, although it feels like a finer search so I updated it.

hasura.planx.uk/migrations/1736953485265_populate published_flows has_send_component/down.sql
```diff
@@ -1,2 +1,12 @@
...
+    ROW_NUMBER() OVER (PARTITION BY flow_id ORDER BY created_at DESC) as rn
+  FROM "public"."published_flows"
+)
 UPDATE "public"."published_flows"
 SET has_send_component = false
+WHERE id IN (
+  SELECT id 
+  FROM ranked_flows 
+  WHERE rn = 1
+);
```
DafyddLlyr, line 12: This looks like a hard way of writing `LIMIT 1`

Docs: https://www.postgresql.org/docs/current/queries-limit.html

DafyddLlyr (changes requested):
Using the `LIMIT` operator should make this simpler and more idiomatic 👍

DafyddLlyr (approved):

PR #4160, opened 2025-01-16
[infrastructure] add initial Locust load testing scripts

freemvmt:
Relevant Trello ticket: https://trello.com/c/oRzedbmJ

This PR introduces load testing scripts in `infrastructure/performance`.

We leverage the [Locust](https://locust.io/) package for Python to ensure our load testing suite is easily maintainable, intending that any PlanX dev can run similar tests in future againts local or staging (or pizza test envionments/prod, should we need to).

I'm not yet finished with the load testing work, but I'm just checking in the general setup and scripts for the aspects of the PlanX stack which I've already load tested to satisfaction / assessed a baseline for.

These are:
- The basic CloudFront-served HTML site (`https://editor.planx.dev/`) ✅ 
- The ECS Fargate Hasura service (server + proxy) - which in turns hits the RDS postgres database ✅ 

OSL team can see my [comprehensive notes in Notion](https://www.notion.so/opensystemslab/PlanX-Load-Testing-13735d469ad180b18c45f0d1db44511e), which provide full documentation of the process, as well as explaining how I arrived at this heuristic.

PR #4161, opened 2025-01-16
[hasura] finalise infrastructure config at sensible values

freemvmt:
Relevant Trello ticket: https://trello.com/c/oRzedbmJ

After gaining confidence that the Hasura setup currently on staging can handle the load tests published in #4160, this PR is just intended to polish the infrastructure specifics, and round down some of the values which were extreme for the purpose debugging.

See #4051, #4076, #4080 and #4096 for those initial changes.

I'll tag this up with a quick review to explain some of the decisions.

### DIFF
--- a/infrastructure/application/Pulumi.staging.yaml
+++ b/infrastructure/application/Pulumi.staging.yaml
@@ -43,7 +43,8 @@ config:
   application:hasura-planx-api-key:
     secure: AAABANHLs3ItPxkteh0chwMP2bKuHO3ovuRLi4FsIrCqerzXVIaTLFDqNR+4KBTeMPz4cnF5tCTwsrJv9GruZdXU+lg=
   application:hasura-proxy-cpu: "512"
-  application:hasura-proxy-memory: "2048"
+  application:hasura-proxy-memory: "1024"
+  application:hasura-scaling-maximum: "2"
   application:idox-nexus-client:
     secure: AAABABprDQomVM9wJQkTMTVtUKvj9lVVVJLdpEBR5p3ibZYvSMedTOb2jztPa0vm6UCH2hilyOV2fsd+akYd3sP8Up5G26mkEKSLSSN4Nc9fu/Hi3Apn1rXHnw==
   application:idox-nexus-submission-url: https://dev.identity.idoxgroup.com/agw/submission-api
--- a/infrastructure/application/services/hasura.ts
+++ b/infrastructure/application/services/hasura.ts
@@ -34,10 +34,6 @@ export const createHasuraService = async ({
     protocol: "HTTP",
     healthCheck: {
       path: "/healthz",
-      interval: 30,
-      timeout: 10,
-      healthyThreshold: 3,
-      unhealthyThreshold: 5,
     },
   });  
   const hasuraListenerHttp = targetHasura.createListener("hasura-http", { protocol: "HTTP" });
@@ -57,8 +53,8 @@ export const createHasuraService = async ({
     desiredCount: 1,
     deploymentMinimumHealthyPercent: 50,
     deploymentMaximumPercent: 200,
-    // extend service-level health check grace period to match hasura server migrations timeout
-    healthCheckGracePeriodSeconds: 600,
+    // service-level health check grace period should exceed proxy dependency timeout
+    healthCheckGracePeriodSeconds: 240,
     deploymentCircuitBreaker: {
       enable: true,
       rollback: true,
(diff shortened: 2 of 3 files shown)


infrastructure/application/services/hasura.ts
```diff
@@ -34,10 +34,6 @@ export const createHasuraService = async ({
     protocol: "HTTP",
     healthCheck: {
       path: "/healthz",
-      interval: 30,
```
freemvmt, line 37: The defaults for the ALB health checking the Hasura service are reasonable, so no need for this.

infrastructure/application/services/hasura.ts
```diff
@@ -172,8 +168,8 @@ export const createHasuraService = async ({
 
   // TODO: bump awsx to 1.x to use the FargateService scaleConfig option to replace more verbose config below
   const hasuraScalingTarget = new aws.appautoscaling.Target("hasura-scaling-target", {
-    // start conservative, can always bump max as required
-    maxCapacity: 3,
+    // maxCapacity should consider compute power of the RDS instance which Hasura relies on
+    maxCapacity: parseInt(config.require("hasura-planx-api-key")),
```
freemvmt, line 172: When I ran tried to push Hasura on staging to breaking point (i.e. hit it with a workload well beyond our satisfactory target), AWS responded by spinning up a 3rd server, which failed to start because the database couldn’t handle any more connections (Hasura servers maintain 20 or so connections even when they have no active queries).

For reference, this saturation point was ~200 connections for the `db.t3.small` instance we've provisioned on staging. There's no reason to put it under that strain.

By comparison, the prod RDS runs on `db.t3.medium`, which is twice as powerful.

infrastructure/application/Pulumi.staging.yaml
```diff
@@ -43,7 +43,8 @@ config:
   application:hasura-planx-api-key:
     secure: AAABANHLs3ItPxkteh0chwMP2bKuHO3ovuRLi4FsIrCqerzXVIaTLFDqNR+4KBTeMPz4cnF5tCTwsrJv9GruZdXU+lg=
   application:hasura-proxy-cpu: "512"
-  application:hasura-proxy-memory: "2048"
+  application:hasura-proxy-memory: "1024"
+  application:hasura-scaling-maximum: "2"
```
freemvmt, line 47: Am I right in assuming these values need to be strings? So we should parse this as an int at read time, as below?

infrastructure/application/services/hasura.ts
```diff
@@ -149,17 +145,17 @@ export const createHasuraService = async ({
               name: "HASURA_PLANX_API_KEY",
               value: config.require("hasura-planx-api-key"),
             },
-            // extend timeout for migrations during setup to 10 mins (default is 30s)
+            // extend timeout for migrations during setup to 3 mins (default is 30s)
             {
               name: "HASURA_GRAPHQL_MIGRATIONS_SERVER_TIMEOUT",
-              value: "600",
+              value: "180",
```
freemvmt, line 151: This is ample time for the Hasura migrations to happen, if they're ever going to! Most of the edits to values above are just a tightening up which flow from this point.

infrastructure/application/services/hasura.ts
```diff
@@ -189,10 +185,10 @@ export const createHasuraService = async ({
         predefinedMetricSpecification: {
             predefinedMetricType: "ECSServiceAverageCPUUtilization",
         },
-        // scale out quickly for responsiveness, but scale in more slowly to avoid thrashing
-        targetValue: 60.0,
+        // scale out early and quickly for responsiveness, but scale in more slowly to avoid thrashing
+        targetValue: 50.0,
```
freemvmt, line 189: We should scale out a bit sooner/earlier than we were to minimise latencies during the transition.

infrastructure/application/services/hasura.ts
freemvmt: Note that if "connection churn" ever becomes a real problem, we can put something like [PgBouncer](https://www.pgbouncer.org/) in front of it to do connection pool management (or AWS’s solution, [RDS Proxy](https://aws.amazon.com/rds/proxy/)).

jessicamcinchak: Oops typo here I think - I think you mean to 
```
maxCapacity: parseInt(config.require("hasura-scaling-maximum"))
```

infrastructure/application/Pulumi.staging.yaml
jessicamcinchak: This looks right to me! If it's not, we should get an obvious error on the staging deploy in the `pulumi up` step :ok_hand:

jessicamcinchak (commented):
Super appreciate the notes & links throughout here !

infrastructure/application/services/hasura.ts
freemvmt: @jessicamcinchak oops thank you, nice catch!

jessicamcinchak (approved):

PR #4162, opened 2025-01-16
chore: Additional logging for publish error

DafyddLlyr:
Looking into issue raised here - https://opendigitalplanning.slack.com/archives/C0241GWFG4B/p1737040722198679 (ODP Slack)

Nothing meaningful caught in Airbrake logs or AWS API logs, this might help get more more info.


jessicamcinchak (approved):
Thanks for this !

PR #4163, opened 2025-01-17
chore: Throw Uniform error to Airbrake, don't pass to `next()`

DafyddLlyr:
Quick change that came from a a double-submission failure - this now matches BOPS, which was caught by Airbrake.


jessicamcinchak (approved):
Thanks for adjusting this one !

PR #4166, opened 2025-01-19
feat: Allow optional fields in `List` and `Page` schemas

DafyddLlyr:
## What does this PR do?
- Introduced a new `required` property for all schema fields
- Based on this prop, makes the input conditionally optional or required. Required is the default and fallback value if not set.
- Standardises validation schema names to `${field}InputValidationSchema` (see https://github.com/theopensystemslab/planx-new/pull/4020#discussion_r1923853166)
- Test coverage for new functionality

## To test...
 - Pull down branch
 - Make schema changes to make fields required or not (Zoo schema has examples of each field type)
 - See `" (optional)" title appended to optional fields
 - Optional fields get validated correctly

### DIFF
--- a/editor.planx.uk/src/@planx/components/Checklist/model.ts
+++ b/editor.planx.uk/src/@planx/components/Checklist/model.ts
@@ -123,15 +123,14 @@ export const getLayout = ({
   return ChecklistLayout.Basic;
 };
 
-interface ValidationSchema {
-  data: Checklist;
-  required: boolean;
-}
-
 export const checklistValidationSchema = ({
   data: { allRequired, options, groupedOptions },
   required,
-}: ValidationSchema) => {
+}: {
+  // Cannot use FieldValidationSchema<ChecklistInput> as this is a simplified representation (i.e. no groups)
+  data: Checklist;
+  required: boolean;
+}) => {
   const flatOptions = getFlatOptions({ options, groupedOptions });
 
   return array()
--- a/editor.planx.uk/src/@planx/components/Checklist/model.ts
+++ b/editor.planx.uk/src/@planx/components/Checklist/model.ts
@@ -123,7 +123,7 @@ export const getLayout = ({
   return ChecklistLayout.Basic;
 };
 
-export const checklistValidationSchema = ({
+export const checklistInputValidationSchema = ({
   data: { allRequired, options, groupedOptions },
   required,
 }: {
--- a/editor.planx.uk/src/@planx/components/Checklist/model.ts
+++ b/editor.planx.uk/src/@planx/components/Checklist/model.ts
@@ -127,7 +127,7 @@ export const checklistInputValidationSchema = ({
   data: { allRequired, options, groupedOptions },
   required,
 }: {
-  // Cannot use FieldValidationSchema<ChecklistInput> as this is a simplified representation (i.e. no groups)
+  // Cannot use type FieldValidationSchema<ChecklistInput> as this is a simplified representation (i.e. no groups)
   data: Checklist;
   required: boolean;
 }) => {
--- a/editor.planx.uk/src/@planx/components/Checklist/model.ts
+++ b/editor.planx.uk/src/@planx/components/Checklist/model.ts
@@ -136,23 +136,14 @@ export const checklistInputValidationSchema = ({
   return array()
     .when([], {
       is: () => required,
-      then: array().required(),
+      then: array().min(1, "Select at least one option"),
       otherwise: array().notRequired(),
-    })
-    .test({
-      name: "atLeastOneChecked",
-      message: "Select at least one option",
-      test: (checked?: Array<string>) => {
-        if (!required && !checked?.length) return true;
-
-        return Boolean(checked && checked.length > 0);
-      },
     })
     .test({
       name: "notAllChecked",
       message: "All options must be checked",
       test: (checked?: Array<string>) => {
-        if (!required && !checked?.length) return true;
+        if (!checked?.length) return true;
         if (!allRequired) return true;
 
         const allChecked = checked && checked.length === flatOptions.length;
--- a/editor.planx.uk/src/@planx/components/Checklist/model.ts
+++ b/editor.planx.uk/src/@planx/components/Checklist/model.ts
@@ -144,7 +144,7 @@ export const checklistValidationSchema = ({
       name: "atLeastOneChecked",
       message: "Select at least one option",
       test: (checked?: Array<string>) => {
-        if (!required) return true;
+        if (!required && !checked?.length) return true;
 
         return Boolean(checked && checked.length > 0);
       },
@@ -153,7 +153,7 @@ export const checklistValidationSchema = ({
       name: "notAllChecked",
       message: "All options must be checked",
       test: (checked?: Array<string>) => {
-        if (!required) return true;
+        if (!required && !checked?.length) return true;
         if (!allRequired) return true;
 
         const allChecked = checked && checked.length === flatOptions.length;
--- a/editor.planx.uk/src/@planx/components/Checklist/model.ts
+++ b/editor.planx.uk/src/@planx/components/Checklist/model.ts
@@ -123,19 +123,29 @@ export const getLayout = ({
   return ChecklistLayout.Basic;
 };
 
+interface ValidationSchema {
+  data: Checklist;
+  required: boolean;
+}
+
 export const checklistValidationSchema = ({
-  allRequired,
-  options,
-  groupedOptions,
-}: Checklist) => {
+  data: { allRequired, options, groupedOptions },
+  required,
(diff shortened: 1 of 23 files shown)


jessicamcinchak:
[List component should support Address inputs and optional fields for better UX of "Ownership" questions & schema alignment](https://trello.com/c/wMVVmvnn/3172-list-component-should-support-address-inputs-and-optional-fields-for-better-ux-of-ownership-questions-schema-alignment)

editor.planx.uk/src/@planx/components/Checklist/model.ts
```diff
@@ -123,29 +123,38 @@ export const getLayout = ({
...
     })
     .test({
       name: "notAllChecked",
       message: "All options must be checked",
       test: (checked?: Array<string>) => {
-        if (!allRequired) {
-          return true;
-        }
+        if (!required && !checked?.length) return true;
+        if (!allRequired) return true;
```
DafyddLlyr, line 156: Despite the similar names I believe the interaction between these two properties is correct - 
 - If the schema field is not required, and none have been selected that's fine
 - Even if the schema field is marked as optional, if `allRequired` is set they have to all be checked

I think this permutation is unlikely to be used, but I'd like a sense check here that I'm not introducing some odd/unexpected behaviour.

editor.planx.uk/src/@planx/components/DateInput/model.ts
```diff
@@ -68,32 +69,44 @@ export const parseDate = (date?: string) => {
 /**
  * Validates that a given string is a date in the correct format
  */
-export const dateSchema = () => {
+export const dateSchema = ({ required }: { required: boolean }) => {
   return string()
     .test("missing day", "Date must include a day", (date?: string) => {
+      if (!required && !date) return true;
```
DafyddLlyr, line 75: I spent a while looking into this - there's not a hierarchical way to run this check once and first - it needs to be repeated. Pretty tedious / error prone!

editor.planx.uk/src/@planx/components/shared/Schema/InputFields/SelectInputField.tsx
```diff
@@ -10,17 +10,18 @@ import { FieldInputDescription } from "./shared";
 
 export const SelectFieldInput: React.FC<Props<QuestionField>> = (props) => {
   const { data, formik } = props;
-  const { id, errorMessage, name, value } = getFieldProps(props);
+  const { id, errorMessage, name, value, required, title } =
+    getFieldProps(props);
```
DafyddLlyr, line 14: We now need to grab `title` from the `getFieldProps()` helper function instead of `data.title` as this will be the formatted value (`" (optional)"` is appended).

I couldn't really find a simple way using TS (or anything else) to block a well-intentioned dev from incorrectly grabbing the raw `data`. We need it elsewhere and mistakes should be fairly obvious if we go wrong here.

editor.planx.uk/src/@planx/components/shared/Schema/model.ts
```diff
@@ -36,70 +36,104 @@ interface ChecklistInput {
...
       name: "atLeastOneFeature",
       message: `Draw at least one ${
         mapOptions?.drawType?.toLocaleLowerCase() || "feature"
       } on the map`,
       test: (features?: Array<Feature>) => {
-        return Boolean(features && features?.length > 0);
+        if (!required && !features?.length) return true;
+        if (!features || !features.length) return false;
+
+        const isGeoJSON = (input: Feature) => input?.type === "Feature";
```
DafyddLlyr, line 86: Pretty crude, there's probably some good libraries for GeoJson validation (turf probably has something) but I think this is fine for now. 

I was hitting an issue where any array (e.g an array of strings) that I passed in met the validation requirements so I couldn't test this.

editor.planx.uk/src/@planx/components/shared/Schema/model.ts
```diff
@@ -36,70 +36,104 @@ interface ChecklistInput {
...
+        return features.every(isGeoJSON);
       },
     });
 
-export type TextField = {
-  type: "text";
-  data: TextInput & { fn: string };
+type BaseField<T> = {
+  required?: boolean;
+  data: T & { fn: string };
```
DafyddLlyr, line 94: This is the fundamental change in this PR - 

 - A new `BaseField` interface with all other fields extends
 - `required` is set her, not at a lower level in `data` (more below)
 - `fn` is being made required at this level (no functional change, moving it here simply saves repetition)

**Why are we setting required at the highest level, and not as part of `data`?**
It's set here as we're making the entire "field" (which can consist of multiple inputs) optional. All PlanX components are currently required, and we're very much restricting this change to schema fields only - not their related components. Setting this here hopefully makes this distinction more explicit.

editor.planx.uk/src/@planx/components/Checklist/model.ts
```diff
@@ -123,29 +123,38 @@ export const getLayout = ({
...
     })
     .test({
       name: "notAllChecked",
       message: "All options must be checked",
       test: (checked?: Array<string>) => {
-        if (!allRequired) {
-          return true;
-        }
+        if (!checked?.length) return true;
+        if (!allRequired) return true;
```
DafyddLlyr, line 156: Despite the similar names I believe the interaction between these two properties is correct -

- If there are no checked items - don't run this test, defer to the required/not required `when()` statement above
- If `allRequired` is set, all values must be checked, even if the field is optional

This means there's two possible states for an optional checklist with `allRequired` - all empty, or all checked.

I think this permutation is unlikely to be used, but I'd like a sense check here that I'm not introducing some odd/unexpected behaviour.

editor.planx.uk/src/@planx/components/DateInput/model.ts
```diff
@@ -71,29 +72,41 @@ export const parseDate = (date?: string) => {
 export const dateSchema = () => {
   return string()
     .test("missing day", "Date must include a day", (date?: string) => {
+      if (!date) return true;
```
DafyddLlyr, line 75: I spent a while looking into this - there's not a hierarchical way to run this check once and first - it needs to be repeated. 

Pretty tedious / error prone! If this is not included, we'll run validations on optional fields (which will always fail).

jessicamcinchak (commented):
All working as expected locally for me - thanks again for picking this one up and sorting through so many tricky cascading type changes. 

A few minor comments throughout and a couple high-level questions/thoughts:
- I'm surprised to not see `/Schema/InputFields/RadioFieldInput` & `@planx/components/Question` model & model.test.ts changes included here with additions around `required` ? Is this intentional because it already has a consistently named validation schema and way of handling optionals, or was this simply accidently overlooked? 
- This implementation of `required` prop at canonical `@planx/component` level while first a bit surprising (expected it to be constrained to the schema "field" types), is now feeling quite sensible and I have lots of wheels turning around how logic functions outside of schema use-cases might hook into it in the future :crystal_ball:

DafyddLlyr:
> I'm surprised to not see /Schema/InputFields/RadioFieldInput & @planx/components/Question model & model.test.ts changes included here with additions around required ? Is this intentional because it already has a consistently named validation schema and way of handling optionals, or was this simply accidently overlooked?

Very good question - there's no required changes to `RadioFieldInput` and both this and `SelectFieldInput` are basically two interfaces for the question component. This is based on number of options, defined here - 

https://github.com/theopensystemslab/planx-new/blob/00dde47e50b473b1e60340f11ec7cb1167af8cdc/editor.planx.uk/src/%40planx/components/shared/Schema/InputFields/index.tsx#L53-L57

They share the (already renamed) `questionInputValidationSchema`. This is a simplified version of what's applied to the full `@planx/components/Question` component which handles a lot more (e.g. grouped options, exclusive or). This means there's no changes needed in this component as it does not share a validation schema.

Side note: There's a whole usability question here around optional radio fields - once checked they can't be unchecked. 

> This implementation of required prop at canonical @planx/component level while first a bit surprising (expected it to be constrained to the schema "field" types), is now feeling quite sensible and I have lots of wheels turning around how logic functions outside of schema use-cases might hook into it in the future 🔮

Yeah it's something I struggled with, as discussed at the dev call. The underlying issue is that if a schema field and a PlanX component share a validation schema (which we want), then we have to handle the `required` value in both locations. I felt that this was the most explicit way of saying "I have to set this here but it's not something an Editor can change, and the node data of this component will never know about it". In future, if we do want to make components optional, we probably would need to move this `required` prop down into `data`. To me it felt best to quite explicitly not do that here.

This is the same reason that tests cover this functionality - I've tried to co-locate all validation schema tests together. You're right though that this isn't super clear - as suggested I've now updated test names to highlight this (please see d8dd83c80).

jessicamcinchak (approved):
Thanks for explanation and adjusting test language - all makes sense !

PR #4169, opened 2025-01-20
feat: add `is_statutory_application_type` to `published_flows` table

RODO94:
## What does this PR do?

To enable users to see which services are _Statutory_ we need to define this in the flow data. The approach I will take is to add a check when a service is published to see if any `Question` || `Checklist` || `SetValue` component has an `fn = "application.type"` and then cross reference this value with the ODP Schema using  `getValidSchemaValues("ApplicationType");`

This will enable all new published services to be defined as Statutory or not. It will not automatically ascribe a non-statutory service as "Discretionary" service as a service may be a modular service used to build other services, making it neither _Discretionary_ or _Statutory_

For this PR, I have only focused on updating the database with the new column, adding permissions, and updating relevant functions.

In a following PR, I will make changes to the seed scripts and api to update the column on publish.


api.planx.uk/modules/flows/validate/service/applicationTypes.ts
```diff
@@ -0,0 +1,38 @@
...
+  const validApplicationTypes = getValidSchemaValues("ApplicationType");
+  const isStatutoryApplication = applicationTypes.some((value) =>
+    validApplicationTypes?.includes(value),
+  );
+  return isStatutoryApplication;
+};
+
+const getApplicationTypeVals = (flowGraph: FlowGraph): string[] => {
+  const applicationTypeChecklistNodes = Object.entries(flowGraph).filter(
+    (entry) => {
```
jessicamcinchak, line 19: nit: when iterating over a flow, we much more commonly refer to the iterator as a `node` rather than "entry". Know it's minor, but helps us build shared language for our data structures ! 

Also `(entry)` &rarr; `([nodeId, node])` might help make the final return condition more clear? Eg `entry[1].data?.fn` would become `node.data?.fn`

api.planx.uk/modules/flows/validate/service/applicationTypes.ts
```diff
@@ -0,0 +1,38 @@
...
+  const applicationTypes = getApplicationTypeVals(flattenedFlow);
+  const validApplicationTypes = getValidSchemaValues("ApplicationType");
+  const isStatutoryApplication = applicationTypes.some((value) =>
+    validApplicationTypes?.includes(value),
+  );
+  return isStatutoryApplication;
+};
+
+const getApplicationTypeVals = (flowGraph: FlowGraph): string[] => {
+  const applicationTypeChecklistNodes = Object.entries(flowGraph).filter(
```
jessicamcinchak, line 18: nit: would `applicationTypeNodes` or `nodesWithFnApplicationType` be more accurate here? You aren't exclusively looking at `Checklist` types ! 

Also do we need to return the whole node or could we simply return their ids to pass into the next function here?

api.planx.uk/modules/flows/validate/service/applicationTypes.ts
```diff
@@ -0,0 +1,38 @@
...
+  const answerVals: string[] = [];
+  applicationTypeChecklistNodes.map(([_nodeId, node]) =>
+    node.edges?.map((edgeId) => {
+      if (typeof flowGraph[edgeId]?.data?.val === "string") {
+        answerVals.push(flowGraph[edgeId]?.data?.val);
+      }
+    }),
+  );
+
+  return answerVals;
```
jessicamcinchak, line 37: nit: it might be nice to consider returning a list of _unique_ application type values found in this flow - this is currently probably going to be much longer with more duplicates than useful !

jessicamcinchak (commented):
Database changes look good here, but I'd personally appreciate if they were separate from the introduction of API helper functions that aren't being called (or tested!) yet. 

These API helper functions are also ripe for unit test coverage, let's talk through how to mock flow data and write test case for them if helpful !

scripts/seed-database/write/published_flows.sql
```diff
@@ -6,7 +6,7 @@ CREATE TEMPORARY TABLE sync_published_flows (
   summary text,
   publisher_id int,
   created_at timestamptz,
-  has_send_component boolean
+  has_send_component boolean
```
jessicamcinchak, line 9: Looks like sync scripts are referenceing the wrong column from a previous PR here !

api.planx.uk/modules/flows/validate/service/applicationTypes.ts
RODO94: Thanks, agree on shared language, it'll make it easier in the future to track what's happening

api.planx.uk/modules/flows/validate/service/applicationTypes.ts
RODO94: Yeah true, we could just return `node[1].id` and pass it in place of `edgesId`? 

Put up the proposed change

scripts/seed-database/write/published_flows.sql
RODO94: This has now been merged, so it's worth keeping in for consistency?

RODO94:
@jessicamcinchak agree on splitting out the API helper functions into a separate PR, I did do that for using them, but for some reason thought it would be good to keep in the helpers. I'm it made sense at some point in the week. 

Thanks for the review

scripts/seed-database/write/published_flows.sql
jessicamcinchak: That's fine to keep it in now that it's also merged, but sync scripts are still missing `is_statutory_application_type` (which is what THIS PR is supposed to be about) :slightly_smiling_face:

api.planx.uk/modules/flows/validate/service/applicationTypes.ts
jessicamcinchak: Appreciate addressing these comments - but still prefer that this whole API files gets cherry-picked out to it's own PR separate from migration per overall code review comment below! Both so unit tests can be added, and we can introduce the function when we're actually ready to _call_ the function

scripts/seed-database/write/published_flows.sql
RODO94: ah! In my mind, I kept the seed scripts out of this PR, as I thought you would need the column in production database before you could copy the data over into the table and test the seed scripts were working locally?

scripts/seed-database/write/published_flows.sql
```diff
@@ -29,7 +29,7 @@ SELECT
   publisher_id,
   created_at,
   has_send_component
-FROM sync_published_flows
+  FROM sync_published_flows
```
jessicamcinchak, line 32: nit: lots of trailing spaces and atypical sql indents introduced here which decrease readability - we don't have linting on .sql files so please try to pay closer attention to manual changes here to avoid this !

scripts/seed-database/write/published_flows.sql
jessicamcinchak: that's correct - but if the goal is to get these two migrations deployed on prod today, then we can update both columns in seed script here now.

my bigger-picture feedback is that it can be quite confusing to review something about _one_ column, and then see _another_ column show up unexplained/uncontextualised as part of the changes!

jessicamcinchak (approved):
Thanks for separating out API changes here, happy for this migration to merge :+1:

PR #4170, opened 2025-01-20
Revert "feat: populate database for `has_send_comp` and add to api mutation"

DafyddLlyr:
Reverts theopensystemslab/planx-new#4156

CI failed so this has not run on the staging DB.

I think a revert here is the simplest course of action given the data migrations ongoing this week - this will be one less thing to worry about or check, and holding it back won't block anything 👍


RODO94 (approved):
Looks good to me, thanks for running the revert

PR #4174, opened 2025-01-21
feat: add search to Team list

RODO94:
## What does this PR do?

This PR introduces a `SearchBox` component which pulls in a set of records and search keys, and outputs the results using a setState function.

This is tied into the larger epic here: https://trello.com/c/7av8lsVr/3160-epic-as-an-editor-i-want-to-have-a-way-to-organise-my-flows-so-that-i-can-easily-find-what-im-looking-for

I am using the `useSearch` hook developed and PoC'd here: https://github.com/theopensystemslab/planx-new/pull/3462

Although I took a slightly different approach to the PoC, due to the way my page was rendering and having to render a list in a parent component.


editor.planx.uk/src/ui/shared/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,105 @@
...
+interface SearchBoxProps<T> {
+  records: T[];
+  staticRecords: T[];
+  setRecords: React.Dispatch<React.SetStateAction<T[] | null>>;
+  searchKey: FuseOptionKey<T>[];
+}
+
+export const SearchBox = <T extends object>({
+  records,
+  staticRecords,
```
RODO94, line 24: Using `staticRecords` here to reset the array when search is cleared. This may become redundant when I add in Filters, and pull things together there, but for now it worked well.

editor.planx.uk/src/ui/shared/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  const { results, search } = useSearch({
+    list: records,
+    keys: formik.values.keys,
+  });
+
+  const debouncedSearch = useMemo(
+    () =>
+      debounce((recordsToUpdate: T[]) => {
+        setRecords(recordsToUpdate);
```
RODO94, line 43: I wanted there to be a slight delay when processing the `results` so you don't receive nothing when you start typing. Iterated around doing this around `search` instead but found this to be the most robust with the other rendering going on. It felt nicer to let `search` run when things change and then wait to pull the `results` out of that after the 500ms

editor.planx.uk/src/ui/shared/SearchBox/SearchBox.tsx
DafyddLlyr: React props are immutable - we can just call `setRecords(records)` to reset when there's no search pattern.

Passing in (and managing) both `records` and `staticRecords` makes things a lot more complex in this component.

DafyddLlyr (changes requested):
A few things here to take a second look at.

The structure is good, but the implementation is a little unclear. Fundamentally the responsibility of this component is to handle the search term, run the search, and then update it's parent's state. 

Something like this will probably get you closer to what you're looking for here - 

```ts
// When the search is reset, display the full list
useEffect(() => {
  if (!formik.values.pattern) setRecords(records)
}, [formik.values.pattern, setRecords, records]);

// When results are updated, update parent state with the filtered records
useEffect(() => {
  if (!results.length) return;
  const filteredRecords = results.map(result => result.item);
  setRecords(filteredRecords)
}, [results, setRecords]);
```

editor.planx.uk/src/ui/shared/SearchBox/SearchBox.tsx
RODO94: I found this tricky with all the re-rendering, but I have added in a state variable called `originalRecords` within the searchbox, I hope this meets this in the middle.

Happy to chat it through tho

editor.planx.uk/src/ui/shared/SearchBox/SearchBox.tsx
RODO94: Not met this so directly, but I have switched this so the search sits in the debounce!

My difficulty was in how to handle the adding of results to the setRecords state dispatch

editor.planx.uk/src/ui/shared/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,105 @@
...
+  staticRecords,
+  setRecords,
+  searchKey,
+}: SearchBoxProps<T>) => {
+  const [isSearching, setIsSearching] = useState(false);
+
+  const formik = useFormik({
+    initialValues: { pattern: "", keys: searchKey },
+    onSubmit: () => {},
+  });
```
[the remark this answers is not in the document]
RODO94: The reason it's like this is so they can be defined in the Parent then memo'ised by Formik within the component.

What do you mean by a form field (&prop)?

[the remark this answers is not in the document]
DafyddLlyr: The keys are never change, and can never change. We're defining a form field for a static value as well as a prop in this component.

We could simplify this with just the following and get the same functionality - 

```ts
  const keys = useMemo(() => ["name", "slug"], []);

  const { results, search } = useSearch({
    list: originalRecords || [],
    keys,
  });
```

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/index.tsx
```diff
@@ -17,7 +17,7 @@ import { ALL_FACETS, SearchFacets } from "./facets";
 import { SearchHeader } from "./SearchHeader";
 import { SearchResultCard } from "./SearchResultCard";
 
-const DEBOUNCE_MS = 500;
+export const DEBOUNCE_MS = 500;
```
DafyddLlyr, line 20: This might be better in a shared location - maybe `ui/shared/constants.ts`?

editor.planx.uk/src/ui/shared/SearchBox/SearchBox.tsx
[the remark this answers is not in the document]
RODO94: Ah so move it out the Formik?

editor.planx.uk/src/ui/editor/SimpleMenu.tsx
```diff
@@ -2,10 +2,11 @@ import MoreVert from "@mui/icons-material/MoreVert";
 import IconButton from "@mui/material/IconButton";
 import Menu from "@mui/material/Menu";
 import MenuItem from "@mui/material/MenuItem";
-import React, { useState } from "react";
+import React, { PropsWithChildren, useState } from "react";
 
 interface Props {
   className?: string;
+  children?: React.ReactNode;
```
DafyddLlyr, line 9: We're wrapping this in `PropsWithChildren` so we don't need this property.

editor.planx.uk/src/ui/editor/SimpleMenu.tsx
```diff
@@ -25,8 +30,10 @@ export default function SimpleMenu({ items, ...restProps }: Props): FCReturn {
         }}
         aria-label="Options"
         size="large"
+        disableRipple
```
DafyddLlyr, line 33: Thanks for sorting this one out 👍

editor.planx.uk/src/ui/shared/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,105 @@
...
+    <Box maxWidth={360}>
+      <InputRow>
+        <InputRowLabel>
+          <strong>Search</strong>
+        </InputRowLabel>
+        <InputRowItem>
+          <Box sx={{ position: "relative" }}>
+            <Input
+              sx={{
+                borderColor: (theme) => theme.palette.border.input,
```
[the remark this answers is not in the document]
DafyddLlyr: Sorry - should have been clearer here - I was specifically referring to the border colour that doesn't match.

image
image

This is actually inconsistent in a few places in the Editor atm and I'm keen not to introduce any more (cc @ianjon3s)

image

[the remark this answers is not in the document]
DafyddLlyr: Spinner is great btw 👌

editor.planx.uk/src/ui/shared/SearchBox/SearchBox.tsx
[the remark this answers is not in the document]
DafyddLlyr: Yep - this is what I meant - thanks for picking this up.

Keys aren't a form field - they're not user controlled and don't need validation etc, they're just an input into the component (prop). Reading back this comment wasn't super clear sorry! 😅

ianjon3s:
Good catch on the input border consistency @DafyddLlyr , something I've been meaning to address.

The input borders used in the editor modal are deliberately left lower contrast due to the density and close-proximity of input groupings, and when used on a single/isolated input (especially against a white background) the same border shade is too low-contrast to be effective, hence the darker border shade for the filters setup.

What's needed is a consistent border shade that works in both instances, looking at your examples it seems we already have a style used for data and select inputs that would be a universal fit. I'll go through and update to this in a separate PR.

DafyddLlyr (approved):
Great!

PR #4177, opened 2025-01-21
feat: display external portals on readme page + test coverage

jamdelion:
### In this PR:

- Display external portals list on Read Me Page (including refactoring component so usable across multiple places) 
- Add storybook file for ReadMePage
- Add test coverage for ReadMePage (one issue encountered, see below ⬇️ )

### Testing issues

In the test `it("displays data in the fields if there is already flow information in the database"`, it only worked if I set just one piece of state. If I set two fields in state:

```js
await act(async () =>
      setState({
        flowSummary: "This flow summary is in the db already",
        flowDescription: "Something else",
      })
    );
```
    
   ...it would find the first piece of text in the document but not the second. When I logged what was in the store on the component page, both pieces of text were in the store on a first render, and then both were `""` on the second render. With `screen.debug` only `serviceSummary` had a value displayed. 
   
 I'm wondering if this is a timing issue and too many re-renders, but wasn't really sure how to debug where the re-renders are coming from. I wasn't able to recreate this with the profiling tool in React dev tools (although it did show a high number of re-renders when the inputs are typed in, which makes sense as they re-render on every keystroke).


DafyddLlyr (approved):
Looking great - a few very minor comments.

Couldn't get to the bottom of the testing issue, but totally agree it's non-functional and something to do with the interacting of vitest and zustand - all working as expected for a user 👍

PR #4178, opened 2025-01-21
fix: ensure auto-answered `SetValue` nodes track allow list answers

jessicamcinchak:
Replaces #4007; see conversation here https://opensystemslab.slack.com/archives/C5Q59R3HB/p1737538391491109

**Before:**
- `analytics_logs` tracks every node you've traveled through with metadata about whether or not it's been auto-answered
- For nodes that have been put to a user, we also record `allow_list_answers` (eg answers to non-personal white-listed data fields that we can retain forever)
  - We aslo only record the `next_log_created_at` timestamp for nodes which have been put to a user, which means these timestamps can be used to understand how long users spend on _visible_ nodes
- For nodes that have been auto-answered, we won't record `allow_list_answers` even if the auto-answered node sets an allow listed data field
  - This generally makes sense because the auto-answered node is likely to be a less granular version of the thing that's already been put to a user - for example, I go through find property and my allow list answer is `property.type = residential.house.terrace`, then I'm automated through a question like "is it residential". Why would I want to duplicate a less-granular allow list answer like `property.type = residential`, or even stranger a "blank" if the automated question was "is it commercial" ?
  - But while that makes sense for Question and Checklist type nodes, it doesn't for SetValues because they can set a data field that will never otherwise be put to a user

**Now:**
- For nodes that have been auto-answered, if they are a SetValue type that is appending or replacing a data value (_not_ removing), we'll record it's allow list answer if it has an applicable data field
  - We will still omit recording allow list answers for auto-answered Questions or Checklists to reduce noise and duplication
  
**Testing:**
- This flow mocks the RAB structure as I understand it https://4178.planx.pizza/testing/rab-mock
- We're testing to ensure that we can now record `rab.exitReason = "permittedDevelopment"` as an allow list answer when it's set via SetValue, whereas previously we would have only captured `civilProject`and other reasons put to the user via non-auto-answered nodes
Screenshot from 2025-01-22 22-25-49

**Open questions:**
- The `analytics_summary` view has zero reference to `metadata.isAutoAnswered` - is this still fine or is there a use case to expose as a new column for easy filtering?

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analytics/mutations.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analytics/mutations.ts
@@ -125,11 +125,15 @@ export const UPDATE_ALLOW_LIST_ANSWERS = gql`
 
 export const UPDATE_AUTO_ANSWERED_ALLOW_LIST_ANSWERS = gql`
   mutation UpdateAllowListAnswers(
+    $analytics_id: bigint!
     $allow_list_answers: jsonb
     $node_id: String!
   ) {
     update_analytics_logs(
-      where: { node_id: { _eq: $node_id } }
+      where: {
+        analytics_id: { _eq: $analytics_id }
+        node_id: { _eq: $node_id }
+      }
       _set: { allow_list_answers: $allow_list_answers }
     ) {
       returning {
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analytics/mutations.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analytics/mutations.ts
@@ -123,6 +123,22 @@ export const UPDATE_ALLOW_LIST_ANSWERS = gql`
   }
 `;
 
+export const UPDATE_AUTO_ANSWERED_ALLOW_LIST_ANSWERS = gql`
+  mutation UpdateAllowListAnswers(
+    $allow_list_answers: jsonb
+    $node_id: String!
+  ) {
+    update_analytics_logs(
+      where: { node_id: { _eq: $node_id } }
+      _set: { allow_list_answers: $allow_list_answers }
+    ) {
+      returning {
+        id
+      }
+    }
+  }
+`;
+
 export const TRACK_INPUT_ERRORS = gql`
   mutation TrackInputErrors($id: bigint!, $error: jsonb) {
     update_analytics_logs_by_pk(
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analytics/provider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analytics/provider.tsx
@@ -420,10 +420,9 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
 
     updatedBreadcrumbKeys.forEach((breadcrumbKey) => {
       const breadcrumb = breadcrumbs[breadcrumbKey];
-
-      // track() is called from the Card component, so auto-answers are naturally omitted because they aren't rendered to a user
-      //   instead we manually insert their analytics_logs and track allow list answers as applicable here
-      if (breadcrumb?.auto) {
+      if (breadcrumb.auto) {
+        // track() is called from the Card component, so auto-answers are naturally omitted because they aren't rendered to a user
+        //   instead we manually insert their analytics_logs and track allow list answers as applicable here
         track(breadcrumbKey);
         updateAutoAnsweredNodeWithAllowListAnswers(breadcrumbKey, breadcrumb);
       } else {
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analytics/provider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analytics/provider.tsx
@@ -371,7 +371,7 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     //   Do not track allow list answers for Questions or Checklists because these will always be equal or less granular data values than put to the user
     if (
       flow[nodeId]?.type !== TYPES.SetValue ||
-      ["removeOne", "removeAll"].includes(flow[nodeId]?.data?.operation)
+      !["append", "replace"].includes(flow[nodeId]?.data?.operation)
     )
       return;
 
@@ -383,11 +383,12 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     );
     if (!allowListAnswers) return;
 
-    // We can't rely on lastVisibleNodeAnalyticsLogId variable here because we have an auto-answered node (not visible)
+    // We can't rely on lastVisibleNodeAnalyticsLogId variable here to mutate a single record because we have an auto-answered node (not visible), so instead we update entire session using analyticsId
     //   But this shouldn't matter in the case of SetValues because very nodeId has same data value (this would not hold for Questions and Checklists with different option paths)
     await publicClient.mutate({
       mutation: UPDATE_AUTO_ANSWERED_ALLOW_LIST_ANSWERS,
       variables: {
+        analytics_id: analyticsId,
         allow_list_answers: allowListAnswers,
         node_id: nodeId,
       },
--- a/editor.planx.uk/src/pages/FlowEditor/lib/analytics/provider.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/analytics/provider.tsx
@@ -353,7 +353,7 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     await publicClient.mutate({
       mutation: UPDATE_ALLOW_LIST_ANSWERS,
       variables: {
-        id: lastVisibleNodeAnalyticsLogId,
+        id: lastVisibleNodeAnalyticsLogId, // TODO ensure this is correct id when isAutoanswered = true ! Currently lagging ~2 logs (need last log, not *visible* node) else empty mutation response
         allow_list_answers: allowListAnswers,
         node_id: nodeId,
       },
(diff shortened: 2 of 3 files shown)


jessicamcinchak:
[Ensure auto answered nodes track allow list variables (Metabase RAB bug)](https://trello.com/c/x3ZyTBOZ/3190-ensure-auto-answered-nodes-track-allow-list-variables-metabase-rab-bug)

editor.planx.uk/src/pages/FlowEditor/lib/analytics/utils.ts
```diff
@@ -29,7 +29,10 @@ export function extractNodeTitle(node: Store.Node): string {
   const nodeTitle =
     node?.type === TYPES.Content
       ? getContentTitle(node)
-      : node?.data?.title ?? node?.data?.text ?? node?.data?.flagSet;
+      : node?.data?.title ??
+        node?.data?.text ??
+        node?.data?.flagSet ??
+        node?.data?.category;
```
jessicamcinchak, line 35: This is a small un-related change to ensure `Filter` node types capture `node_title` as their flagset category name, rather than undefined like now

editor.planx.uk/src/pages/FlowEditor/lib/analytics/mutations.ts
```diff
@@ -123,6 +123,22 @@ export const UPDATE_ALLOW_LIST_ANSWERS = gql`
...
+    update_analytics_logs(
+      where: { node_id: { _eq: $node_id } }
+      _set: { allow_list_answers: $allow_list_answers }
+    ) {
+      returning {
+        id
+      }
+    }
+  }
+`;
```
jessicamcinchak, line 140: I'm not attempting to backfill any data as part of this PR, we'll purely be collecting this _going forward_ ! 

BUT because of how this `UPDATE_AUTO_ANSWERED_ALLOW_LIST_ANSWERS` mutation works, if the `node_id` has been stable over time then historical `analytics_logs` will also get updated whenever it's updated the first time - is this okay or a problem for how content changes over time? 

I spent a fair bit of time trying to wrangle with `lastVisibleNodeAnalyticsLogId` in the context of auto-answers (eg _not visible_ nodes) in order to update a single record in the mutation to no avail ! (LogId was always lagging ~2 records behind). Am I forgetting any other ways to update the single, most recent record which matches the `where` param? (Can't pass a `limit` & `order_by` to a mutation!)

editor.planx.uk/src/pages/FlowEditor/lib/analytics/provider.tsx
```diff
@@ -387,8 +421,12 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
     updatedBreadcrumbKeys.forEach((breadcrumbKey) => {
       const breadcrumb = breadcrumbs[breadcrumbKey];
       if (breadcrumb.auto) {
+        // track() is called from the Card component, so auto-answers are naturally omitted because they aren't rendered to a user
+        //   instead we manually insert their analytics_logs and track allow list answers as applicable here
         track(breadcrumbKey);
+        updateAutoAnsweredNodeWithAllowListAnswers(breadcrumbKey, breadcrumb);
       } else {
+        // Visible nodes always track allow list answers independent of component type
```
DafyddLlyr, line 429: Explainer here is super helpful - we'll be grateful of this in future!

editor.planx.uk/src/pages/FlowEditor/lib/analytics/provider.tsx
```diff
@@ -360,6 +361,39 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
...
+    nodeId: string,
+    breadcrumb: Store.UserData,
+  ) {
+    if (shouldSkipTracking()) return;
+
+    // Only proceed if it's a SetValue component that is appending or replacing (not removing) a value
+    //   Do not track allow list answers for Questions or Checklists because these will always be equal or less granular data values than put to the user
+    if (
+      flow[nodeId]?.type !== TYPES.SetValue ||
+      ["removeOne", "removeAll"].includes(flow[nodeId]?.data?.operation)
```
DafyddLlyr, line 374: It seems unlikely, but this could fail if we add more operations to SetValue.

We might be better checking if it's not `"replace"` or `"append"`.

editor.planx.uk/src/pages/FlowEditor/lib/analytics/mutations.ts
RODO94: question: what you're saying here is that if the node has not been updated then it'll be caught up in these changes because we can't limit to the latest version of it? 

For the question at the end, I can't think of anything without using a computer field or something more complex.

editor.planx.uk/src/pages/FlowEditor/lib/analytics/provider.tsx
```diff
@@ -360,6 +361,39 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
...
+    )
+      return;
+
+    const allowListAnswers = getAllowListAnswers(
+      nodeId,
+      breadcrumb,
+      flow,
+      breadcrumbs,
+    );
+    if (!allowListAnswers) return;
```
RODO94, line 384: question: there's no need to check if this is an empty object because that's already caught within `getAllowListAnswers` in the `utils.ts`?

editor.planx.uk/src/pages/FlowEditor/lib/analytics/provider.tsx
```diff
@@ -360,6 +361,39 @@ export const AnalyticsProvider: React.FC<{ children: React.ReactNode }> = ({
...
+
+    // We can't rely on lastVisibleNodeAnalyticsLogId variable here because we have an auto-answered node (not visible)
+    //   But this shouldn't matter in the case of SetValues because very nodeId has same data value (this would not hold for Questions and Checklists with different option paths)
+    await publicClient.mutate({
+      mutation: UPDATE_AUTO_ANSWERED_ALLOW_LIST_ANSWERS,
+      variables: {
+        allow_list_answers: allowListAnswers,
+        node_id: nodeId,
+      },
+    });
```
RODO94, line 394: This whole way of structuring the code is super cool, really nice to read through

RODO94 (commented):
Put in some questions for the moment and will do some manual testing now to track how SetValues do with `ALLOW_LIST` variables. 

Some of the macro pieces of this did go over my head a bit, but I have tested the mutation in Hasura console and ran through the code for any logic typos. These look good to me, pending manual testing

editor.planx.uk/src/pages/FlowEditor/lib/analytics/mutations.ts
DafyddLlyr: Great questions! 🤔 

I think we can restrict this to the analytics id - it's not a real problem if we overwrite all historic records all the time as I think we'll maintain stable nodeIds but it's a very weird side effect, and unexpected. There might be some weird consequences associated with updating timestamps etc that may catch us out in future.

I think we can restrict to an analytics using something like this (build in GraphiQL playground in Hasura but not tested!) 

```graphql
mutation UpdateAllowListAnswers(
  $allow_list_answers: jsonb
  $node_id: String!
  $analyticsId: bigint
) {
  update_analytics_logs(
    where: {
      node_id: {_eq: $node_id}
     // Limit to this analytics "session"
      analytics_id: {_eq: $analyticsId}
    }, 
    _set: {
      allow_list_answers: $allow_list_answers
    }) {
    returning {
      id
    }
  }
}
```

If there are multiple records for this nodeId + analytics Id it will overwrite them all, but that should not be a problem (e.g. a user going backwards and forwards many times over this node).

editor.planx.uk/src/pages/FlowEditor/lib/analytics/provider.tsx
jessicamcinchak: Correct - the return type of `getAllowListAnswers` is `Partial<Record<AllowListKey, any>> | undefined`, which excludes an `{}`

editor.planx.uk/src/pages/FlowEditor/lib/analytics/mutations.ts
DafyddLlyr: However - this is a clever side effect and nice way to update historic missing records - we should do this for the gap we've created 👍

DafyddLlyr (changes requested):
One idea about the query - happy to talk it through and help test. Very keen to get this one sorted asap!

editor.planx.uk/src/pages/FlowEditor/lib/analytics/mutations.ts
jessicamcinchak: `analytics_id` :bulb: thanks for this suggestion! this is indeed a good straightforward way to limit the mutation to only this session instead of any/every past occurance of this node ID. Mutation now updated :+1: 

Re: timestamps - you're right this could get weird ! But a couple good clarifications after double-checking table:
- `analytics_log` only has an autogenerated `created_at`, there is no additional `updated_at` reflecting mutations
- there is a `next_log_created_at` which is important for _visible_ nodes, but the SetValue records will never set this because they are auto-answered

jessicamcinchak: Will revisit that question of historic records as part of allow list `property.type` migration perhaps (agree node id is a clever option when it comes to SetValues specifically) - but let's just merge this for now with correct tracking _for the future_ :heavy_plus_sign:

RODO94:
@jessicamcinchak manual testing worked for me:

Ran through this: https://4178.planx.pizza/buckinghamshire/report-a-planning-breach

With the intention to hit this: https://4178.planx.pizza/buckinghamshire/report-a-planning-breach/nodes/_root/nodes/qeFV90b0Gn/edit

And when I did, I got this record in the database:

Screenshot 2025-01-23 at 17 07 03

PR #4179, opened 2025-01-22
chore: Upgrade Metabase

DafyddLlyr:
Update Metabase to latest version, v0.50.26, as recommended by Jumpsec.


jamdelion (approved):
👍 Thanks for 'jumping' on these ;)

PR #4183, opened 2025-01-22
fix: Consistent text case in the editor

ianjon3s:
## What does this PR do?

Currently we use a mix of `Sentence case` and `Title Case` across the editor, with broad mix of the two being used for titles, legends and labels. Our style guides for PlanX state that sentence case should be used except for referring to proper nouns. PR adjusts all visible titles, legends and labels to be sentence case.

Labels for switches were programatically capitalised into title case, I've added a prop `capitalize` which can be used where needed (online/offline switch).


jessicamcinchak (approved):
Appreciate this attention to detail ! :nail_care:

PR #4171, opened 2025-01-21
fix(wip): `ListManager` reordering bug

DafyddLlyr:

PR #4186, opened 2025-01-23
fix: Option label reordering in `ListManager`

DafyddLlyr:
## Context
This is me trying to unpick and resolve the changes made in [https://github.com/theopensystemslab/planx-new/pulls whi](https://github.com/theopensystemslab/planx-new/pull/4171) which touch and fix a few things, but is failing E2E tests in a way I've not yet identified. I'm hoping that splitting up the fixes will help me narrow this down. Next up is the same issue for the `DataFieldAutocomplete`.

This PR intentionally does not try to address any issues with the `DataFieldAutocomplete`.

## What's the problem?
When using a the `ListManger` and re-ordering options, the associated tags are not correctly re-ordered and remain in their index position. This only happens on new, unsaved, list items. See demo below on staging - 

https://github.com/user-attachments/assets/4da620de-b0df-441a-8085-693802ed91de

I believe that this is caused by inputs going from controlled to uncontrolled ([docs](https://react.dev/learn/sharing-state-between-components#controlled-and-uncontrolled-components)), as the following error can be seen in the console on dev mode - 

image

## What's the solution?
By setting all fields to have a value (i.e. not `undefined`), no inputs should go from controlled to uncontrolled.

### DIFF
--- a/editor.planx.uk/src/@planx/components/Checklist/Editor/Options.tsx
+++ b/editor.planx.uk/src/@planx/components/Checklist/Editor/Options.tsx
@@ -38,15 +38,15 @@ export const Options: React.FC<{ formik: FormikHookReturn }> = ({ formik }) => {
               formik.setFieldValue("options", newCombinedOptions);
             }}
             newValueLabel="add new option"
-            newValue={() =>
-              ({
-                data: {
-                  text: "",
-                  description: "",
-                  val: "",
-                },
-              }) as Option
-            }
+            newValue={() => ({
+              id: "",
+              data: {
+                text: "",
+                description: "",
+                val: "",
+                flags: [],
+              },
+            })}
             Editor={ChecklistOptionsEditor}
             editorExtraProps={{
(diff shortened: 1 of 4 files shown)


editor.planx.uk/src/@planx/components/Checklist/Editor/Options.tsx
```diff
@@ -38,15 +38,15 @@ export const Options: React.FC<{ formik: FormikHookReturn }> = ({ formik }) => {
...
             }}
             newValueLabel="add new option"
-            newValue={() =>
-              ({
-                data: {
-                  text: "",
-                  description: "",
-                  val: "",
-                },
-              }) as Option
```
DafyddLlyr, line 48: This casting as masking the problem - `flags` was toggling between `undefined` and an actual value on change.

jessicamcinchak (approved):
Thanks for unpicking these - good reminders about "default"/undefined value handling here !

PR #4187, opened 2025-01-23
fix: Force `DataFieldAutocomplete` to be controlled

DafyddLlyr:
## Context
This is me trying to unpick and resolve the changes made in [https://github.com/theopensystemslab/planx-new/pulls whi](https://github.com/theopensystemslab/planx-new/pull/4171) which touch and fix a few things, but is failing E2E tests in a way I've not yet identified. I'm hoping that splitting up the fixes will help me narrow this down. 

## What's the problem?
When using a the `ListManger` and re-ordering options, the associated data fields are not correctly re-ordered and remain in their index position. This only happens on new, unsaved, list items.

I believe that this is caused by inputs going from controlled to uncontrolled ([docs](https://react.dev/learn/sharing-state-between-components#controlled-and-uncontrolled-components)), as the following error can be seen in the console on dev mode - 

image

## What's the solution?
By setting all fields to have a value (i.e. not `undefined`), no inputs should go from controlled to uncontrolled.

Note: There's an issue when blurring from data input fields - this only works when values are selected via the "Add xxx" button, or when hitting `Enter` on the keyboard. PR #4188 addresses this.

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/BaseOptionsEditor.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/BaseOptionsEditor.tsx
@@ -84,7 +84,7 @@ export const BaseOptionsEditor: React.FC<BaseOptionsEditorProps> = (props) => (
             ...props.value,
             data: {
               ...props.value.data,
-              val: targetValue ? targetValue : undefined,
+              val: targetValue ?? "",
             },
           });
         }}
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/@planx/components/shared/BaseOptionsEditor.tsx
```diff
@@ -84,7 +84,7 @@ export const BaseOptionsEditor: React.FC<BaseOptionsEditorProps> = (props) => (
             ...props.value,
             data: {
               ...props.value.data,
-              val: targetValue ? targetValue : undefined,
+              val: targetValue ?? "",
```
DafyddLlyr, line 87: Here's the issue - the input it originally `undefined` and then gets assigned to a string (or `null`) on change which toggles between controlled/uncontrolled.

jessicamcinchak (approved):

PR #4188, opened 2025-01-23
fix: Use `autoSelect` prop to capture blur events on `DataFieldAutocomplete`

DafyddLlyr:
## Context
This is me trying to unpick and resolve the changes made in [https://github.com/theopensystemslab/planx-new/pulls whi](https://github.com/theopensystemslab/planx-new/pull/4171) which touch and fix a few things, but is failing E2E tests in a way I've not yet identified. I'm hoping that splitting up the fixes will help me narrow this down.

## What's the problem?
When using the `DataFieldAutocomplete` field, values are only captured (the `handleChange()` function being called) only happens when I select a value with the mouse or hit enter. If I type and tab (or click) away, the input will display my value but Formik will not update - this leads to a number of issues.

## What's the solution?
The `autoSelect` prop controls this, so it's just a case of setting this to `true`. 

> **`autoSelect`**
>
> If `true`, the selected option becomes the value of the input when the Autocomplete loses focus unless the user chooses a different option or changes the character string in the input.
>
> When using the `freeSolo` mode, the typed value will be the input value if the Autocomplete loses focus without highlighting an option.

Docs - https://mui.com/material-ui/api/autocomplete/#props

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/DataFieldAutocomplete.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/DataFieldAutocomplete.tsx
@@ -1,7 +1,7 @@
 import {
   AutocompleteChangeReason,
   AutocompleteProps,
-  createFilterOptions
+  createFilterOptions,
 } from "@mui/material/Autocomplete";
 import ListItem from "@mui/material/ListItem";
 import isNull from "lodash/isNull";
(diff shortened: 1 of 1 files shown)


jessicamcinchak (approved):
Thanks for getting to bottom of which prop was missing/needed here - huge difference !

PR #4195, opened 2025-01-23
feat: Gracefully handle expired JWTs

DafyddLlyr:
## What does this PR do?
- Follow on to #4191 & #4193
- Handles an expired token in the following ways - 
  - API request - redirect to "/logout" using `res.redirect()`
  - GraphQL error - catch and parse error, set `window.location.href`

There's a notable exception here with requests made to planx-core. When these requests fail we currently call a `console.error()` (generally), which can't get caught in a meaningful way by an error boundary. When I've been navigating as an Editor it's not really an issue as most pages require an Apollo request or API request, so I may get a component not loading and then a redirect a few moments later. If this feels like a real usability issue we can work our way through these, throw errors, and catch in an `ErrorBoundary`.

## To test...
If you want to try this out locally, please pull down the branch and change the following line to a short timeout, e.g. `"2m"` - 

https://github.com/theopensystemslab/planx-new/blob/main/api.planx.uk/modules/auth/service.ts#L19

## To do...
- [x] Test coverage

### DIFF
--- a/api.planx.uk/.env.test.example
+++ b/api.planx.uk/.env.test.example
@@ -28,7 +28,7 @@ FILE_API_KEY_GATESHEAD=👻
 FILE_API_KEY_DONCASTER=👻
 
 # Editor
-EDITOR_URL_EXT=example.com
+EDITOR_URL_EXT=https://www.example.com
 
 # Hasura
 HASURA_GRAPHQL_URL=http://hasura:8080/v1/graphql
--- a/api.planx.uk/errors/requestHandlers.ts
+++ b/api.planx.uk/errors/requestHandlers.ts
@@ -1,0 +1,53 @@
+import type { ErrorRequestHandler } from "express";
+import { ServerError } from "./serverError.js";
+import airbrake from "../airbrake.js";
+
+/**
+ * Check for expired JWTs, redirect to /logout if found
+ */
+export const expiredJWTHandler: ErrorRequestHandler = (
+  errorObject,
+  _res,
+  res,
+  next,
+) => {
+  const isJWTExpiryError =
+    errorObject?.name === "UnauthorizedError" &&
+    errorObject?.message === "jwt expired";
+
+  if (!isJWTExpiryError) return next(errorObject);
+
+  const logoutPage = new URL("/logout", process.env.EDITOR_URL_EXT!).toString();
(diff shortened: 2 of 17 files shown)


api.planx.uk/errors/requestHandlers.ts
jamdelion, line 1: Nice, always pro splitting things like this into a different file 👍

jamdelion (commented):
Code looks good and I've tried this locally - changed the expiry to 1 minute and the next time I clicked around in the editor after 1 minute, I got successfully directed to the /logout screen. 

I have a slight concern about the user experience however. I could imagine a user happily using the editor right up the expiry time and then being really baffled when they are suddenly logged out without warning. I wonder if we could add something to explain what's happening - perhaps just a toast message saying something like "Your session has expired, please log in again"?

jessicamcinchak (approved):
Similarly working as expected for me locally :heavy_plus_sign: 

Onboard with plan to keep iterating on this one - 
- Auto redirect to `/logout` is way better than nothing like now!
- Let's keep thinking on friendlier UI and planx-core error handling next

api.planx.uk/.env.test.example
```diff
@@ -28,7 +28,7 @@ FILE_API_KEY_GATESHEAD=👻
 FILE_API_KEY_DONCASTER=👻
 
 # Editor
-EDITOR_URL_EXT=example.com
+EDITOR_URL_EXT=https://www.example.com
```
DafyddLlyr, line 31: Value also update on `.env.test` on S3.

api.planx.uk/tests/mockJWT.ts
```diff
@@ -1,7 +1,13 @@
 import type { Role } from "@opensystemslab/planx-core/types";
 import jwt from "jsonwebtoken";
 
-function getJWT({ role }: { role: Role }) {
+function getTestJWT({
```
DafyddLlyr, line 4: This name change is here just to make things super clear!

DafyddLlyr:
Thanks for the reviews both!

I did originally have a toast here (matching other errors in `graphql.ts`) but it's pretty flakey and doesn't have the planx-core issues I mentioned.

I've added this back, with a fallback, to mitigate this as much as we can before we solve this with a refresh token pattern.

PR #4196, opened 2025-01-23
chore: update historic `rab.exitReason` SetValue nodes

jessicamcinchak:
Follows on from https://github.com/theopensystemslab/planx-new/pull/4178

### DIFF
--- a/hasura.planx.uk/migrations/1737661526220_backfill_rab_exitReason_analytics_logs_allow_list/up.sql
+++ b/hasura.planx.uk/migrations/1737661526220_backfill_rab_exitReason_analytics_logs_allow_list/up.sql
@@ -1,0 +1,20 @@
+-- https://editor.planx.uk/open-digital-planning/report-a-planning-breach-scoping/nodes/tpjEjOO1Ne/nodes/zYRFfYQHFE/edit
+UPDATE analytics_logs
+SET allow_list_answers = jsonb_set(
+  coalesce(allow_list_answers, '{}'), '{rab.exitReason}', '["permittedDevelopment"]'
+)
+WHERE node_id = 'zYRFfYQHFE';
+
+-- https://editor.planx.uk/open-digital-planning/report-a-planning-breach-scoping/nodes/tFY60OO1Ne/nodes/DMfGTEtkSY/edit
+UPDATE analytics_logs
+SET allow_list_answers = jsonb_set(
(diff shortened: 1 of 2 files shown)


hasura.planx.uk/migrations/1737661526220_backfill_rab_exitReason_analytics_logs_allow_list/up.sql
```diff
@@ -0,0 +1,20 @@
+-- https://editor.planx.uk/open-digital-planning/report-a-planning-breach-scoping/nodes/tpjEjOO1Ne/nodes/zYRFfYQHFE/edit
+UPDATE analytics_logs
+SET allow_list_answers = jsonb_set(
+  coalesce(allow_list_answers, '{}'), '{rab.exitReason}', '["permittedDevelopment"]'
```
jessicamcinchak, line 4: A couple notes about these commands: 
- Because the current `allow_list_answers` column will be `null` (or possibly have an existing/newly tracked `rab.exitReason` already), we want to use `coalesce` to ensure `jsonb_set()` works correctly in both scenarios. [This thread](https://stackoverflow.com/questions/63391816/why-jsonb-functions-does-not-work-on-update-when-column-is-null) explains it well
- This `UPDATE` statement does not impact any of the timestamps associated with `analytics_logs`; this table does not have an auto `updated_at`

jessicamcinchak:
[Ensure auto answered nodes track allow list variables (Metabase RAB bug)](https://trello.com/c/x3ZyTBOZ/3190-ensure-auto-answered-nodes-track-allow-list-variables-metabase-rab-bug)

hasura.planx.uk/migrations/1737661526220_backfill_rab_exitReason_analytics_logs_allow_list/up.sql
jessicamcinchak: For testing: you probably won't have these specific node ids in your own `analytics_logs` table, but you can manually run the `UPDATE` command and swap in any `node_id` of your choice! 

Try it against a record with `null` `allow_list_answers` and a record that has other `allow_list_answer` values already (`rab.exitReason` should be _appended_ onto the existing values, not overwriting them!)

DafyddLlyr (approved):
Perfect - really appreciate the testing notes - tested locally and getting expected results ✅ 

Very nice to be able to backfill these historic records - a much better outcome.

PR #4200, opened 2025-01-24
feat: `is_statutory_application_type` helper functions added to API on publish

RODO94:
## What does this PR do?

This PR introduces API functions to update `is_statutory_application_type` and unit testing within the API that monitors how it is assigned when a flow is published at the `flow/:flowId/publish` end point.

I have also rearranged parts of the code in the API to match the way I'm importing certain functions and mocks


api.planx.uk/modules/flows/flowHelpers.ts
RODO94, line 1: Moved `isComponentType` out of validate folder as I was using it in the Publish work so it seemed necessary to put it in a file which covers the module rather than just a service within it

api.planx.uk/modules/flows/publish/helpers.ts
RODO94, line 1: This query mocks were coming up continually and it made it difficult to read through the file, hope the naming and arguments passed ensure naming and nuance is captured while reducing the size of the test file, that was my intent

api.planx.uk/modules/flows/publish/helpers.ts
```diff
@@ -0,0 +1,49 @@
...
+        data: alteredFlow,
+      },
+    },
+  });
+};
+
+export const mockCheckStatApplicationTypesFn = (vi: VitestUtils) => {
+  const publishMut = vi.spyOn(
+    applicationTypes,
+    "checkStatutoryApplicationTypes",
```
RODO94, line 46: This refers to the function I use to define whether an application is a Statutory application

api.planx.uk/modules/flows/publish/publish.test.ts
```diff
@@ -179,3 +145,99 @@ describe("publish", () => {
       });
   });
 });
+
+describe("how 'is_statutory_application_Type' is updated when a service is published", () => {
```
RODO94, line 149: For this test block, I ran through each component type I expected to give me a `true` result and tested these in isolation while adding a false test at the end. It is an `or` logic test so I need to test each true component but can pick up all the failing ones in a single test.

I did consider testing that another component does not give me a `true` result but I wasn't sure how to fence this in... and I wasn't sure it was adding enough value.

api.planx.uk/modules/flows/publish/publish.test.ts
```diff
@@ -179,3 +145,99 @@ describe("publish", () => {
...
+  beforeEach(() => {
+    const getStoreMock = vi.spyOn(userContext, "getStore");
+    getStoreMock.mockReturnValue({
```
RODO94, line 152: I ran into a number of `No user is attached to the query` fails, which this fixed. Would need more time to dig into exactly why this was throwing up an error, but code is minimal and makes it work

api.planx.uk/modules/flows/validate/helpers.ts
```diff
@@ -3,16 +3,7 @@ import type {
   FlowGraph,
   Node,
 } from "@opensystemslab/planx-core/types";
-import type { Entry } from "type-fest";
```
RODO94, line 6: This has been moved out of `validate/` and now sits in a helper at `flows/` level

api.planx.uk/tests/mocks/applicationTypeCheckMocks.ts
RODO94, line 1: Iterated on this a bit when coding, and found it easier to create `Base` types for `SetValue | Checklist | Question` then add in the nuance later for `val`. Thinking here is that, if we wanted to test a new application type, we would have to update a smaller object and understand less of the rest of the mock since there is a clear variable being changed (giving you licence to ignore other parts)

api.planx.uk/tests/mocks/applicationTypeCheckMocks.ts
```diff
@@ -0,0 +1,106 @@
+const setValueApplicationTypeBase = {
```
jessicamcinchak, line 1: nit: these mocks could be typed as `FlowGraph` to ensure they're the expected data structure.

I currently find them very strange to read because you _haven't_ mocked them as realistic flow data strucutres with a `_root` node, despite your function expecting a `FlowGraph` argument

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Publish/PublishFlowButton.tsx
```diff
@@ -52,6 +52,8 @@ export const PublishFlowButton: React.FC<{ previewURL: string }> = ({
     try {
       setLastPublishedTitle("Checking for changes...");
       const alteredFlow = await validateAndDiffFlow(flowId);
+      console.log(alteredFlow);
```
jessicamcinchak, line 55: Another `console.log` here - again, please take a quick scan through your own code before opening for review - I want to focus on the juicy changes only :slightly_smiling_face:

api.planx.uk/modules/flows/publish/publish.test.ts
jessicamcinchak: Let's talk about this tomorrrow ! I think being able to test the negative/inverse here (that we are _not_ unexpectedly updating this column) definitely adds value and should be covered

api.planx.uk/modules/flows/publish/service/applicationTypes.ts
```diff
@@ -0,0 +1,51 @@
+import {
```
jessicamcinchak, line 1: A couple quick reactions here: 
- I think this function could live within `flows/publish/helpers` - I think the `service/` subdirectory in addition to the existing `service.ts` file is quite confusing organisationally
- I'd like to see simple unit tests for this helper function _in addition to_ the full route-level `supertest` tests you've already written (eg `flows/publish/helpers.test.ts`). That would help us more easily/clearly test some of those cases you mention above like checking _other_ component types
- Can we chat tomorrow about how we might optimise this function and testing it?

api.planx.uk/modules/flows/publish/service/applicationTypes.ts
```diff
@@ -0,0 +1,51 @@
...
+  ComponentType,
+  type FlowGraph,
+} from "@opensystemslab/planx-core/types";
+import { getValidSchemaValues } from "@opensystemslab/planx-core";
+import { isComponentType } from "../../flowHelpers.js";
+
+export const checkStatutoryApplicationTypes = (flattenedFlow: FlowGraph) => {
```
jessicamcinchak, line 8: nit: it'd be helpful to have a return type here! Eg `boolean`? "check" doesn't typically indicate a boolean return for me semantically!

jessicamcinchak (commented):
Took a quick first pass - there's a few immediate things to tidy up here please! 

I'll come back tomorrow morning with fresher eyes to fully review the functionality of the new function and testing on the pizza - I think it might be nice to check-in or pair about how to generate realistic mock flow data for testing, and how to simplify the logic here a bit ?

api.planx.uk/modules/flows/publish/service/applicationTypes.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+    if (typeof flowGraph[node]?.data?.val === "string") {
+      answerVals.push(flowGraph[node]?.data?.val);
+    }
+  });
+
+  const uniqueAnswerVals = [...new Set(answerVals)];
+
+  return uniqueAnswerVals;
+};
```
jessicamcinchak, line 50: Okay back with fresh eyes here :sun_with_face: 

I have two main questions about the logic here: 
- One is more high-level about our intended outcome: do we expect/want the "statutory" tag in the editor to _be a subset of_ the "submission services" classification, or do we want to tag every flow "statutory" that sets an applicable `application.type` data field?
  - For example, many flows used only as external portals will have an `application.type` node for automations and get tagged as "statutory" by your current logic - see https://editor.planx.dev/barnet/apply-for-service-declarations
- The next is about what this function actually needs to do - if we've decided that we only care about determining that _there is at least one statutory application path_ within a flow (eg `some` not `every`), then do we actually need to get _every_ answer val - or could we simply `return true` at the first statutory one that we find? Can we combine these two functions into a single one and ensure we're not iterating over/joining more flow data than necessary since it's expensive?

As mentioned before, let's maybe pair on refactoring this? It's a good candidate to refine your mock flow data, write unit tests, then work backwards to refactor the function I think.

api.planx.uk/modules/flows/publish/helpers.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    ).toEqual(false);
+  });
+
+  test("returns true for a flow with Send and at least one statutory application.type value", () => {
+    expect(hasStatutoryApplicationType(mockStatutoryFlow)).toEqual(true);
+  });
+
+  test("returns true for a flow with Send and at least one statutory application.type in a SetValue Component", () => {
+    expect(
+      hasStatutoryApplicationType(mockSetValueStatutoryApplicationType),
```
RODO94, line 40: Added this to ensure it goes into the first nested `if()` condition, otherwise the test didn't have it covered

api.planx.uk/modules/flows/publish/helpers.test.ts
```diff
@@ -0,0 +1,139 @@
...
+  ...mockStatutoryFlow,
+  SetValueStatutoryAppType: {
+    data: {
+      fn: "application.type",
+      val: "ldc",
+      operation: "replace",
+    },
+    type: TYPES.SetValue,
+  },
+} as FlowGraph;
```
RODO94, line 135: Required the cast to ensure type error didn't pop up on the next `delete` line

api.planx.uk/modules/flows/publish/service.ts
```diff
@@ -26,11 +27,14 @@ export const publishFlow = async (flowId: string, summary?: string) => {
...
   const flattenedFlow = await dataMerged(flowId);
   const mostRecent = await getMostRecentPublishedFlow(flowId);
-  const hasSendComponent = hasComponentType(flattenedFlow, ComponentType.Send);
   const delta = jsondiffpatch.diff(mostRecent, flattenedFlow);
 
   if (!delta) return null;
 
+  const hasSendComponent = hasComponentType(flattenedFlow, ComponentType.Send);
+  const isStatutoryApplication =
+    hasSendComponent && hasStatutoryApplicationType(flattenedFlow);
```
RODO94, line 36: added this line so you don't need to go into the fn to understand it's initial condition or what defines a `statutory application`

api.planx.uk/modules/flows/publish/helpers.ts
```diff
@@ -0,0 +1,49 @@
...
+
+        return isStatutoryApplication;
+      }
+
+      // Check other Nodes which have Edges
+      if (nodeToCheck.edges) {
+        // Loop through each edge and check the value
+        nodeToCheck.edges.some((edge) => {
+          const edgeData = flattenedFlow[edge]?.data;
+          if (typeof edgeData?.val === "string") {
```
jessicamcinchak, line 36: I don't think there should never be an instance where `val` isn't a "string" type, is this `if` necessary? Was it throwing a type error without it?

api.planx.uk/modules/flows/publish/service.ts
```diff
@@ -48,13 +53,15 @@ export const publishFlow = async (flowId: string, summary?: string) => {
...
             has_send_component: $has_send_component
+            is_statutory_application_type: $is_statutory_application_type
           }
         ) {
           id
           flowId: flow_id
           publisherId: publisher_id
           createdAt: created_at
           data
+          is_statutory_application_type
```
jessicamcinchak, line 64: nit: do we actually need to return `is_statutory_application_type`, are we doing anything with it? Or can we simply _set_ it?

jessicamcinchak, line 64: nit: do we actually need to return `is_statutory_application_type`, are we doing anything with it? Or can we simply _set_ it?

jessicamcinchak, line 64: nit: do we actually need to return `is_statutory_application_type`, are we doing anything with it? Or can we simply _set_ it?

jessicamcinchak, line 64: nit: do we actually need to return `is_statutory_application_type`, are we doing anything with it? Or can we simply _set_ it?

jessicamcinchak, line 64: nit: do we actually need to return `is_statutory_application_type`, are we doing anything with it? Or can we simply _set_ it?

jessicamcinchak (approved):
Thanks for changes - this all feels much simpler, more readable, and better tested than the first iteration ! Two small comments, happy for this to merge :+1:

api.planx.uk/modules/flows/publish/helpers.ts
RODO94: I'll check, it would have been a type error I imagine to do with the `Value` type maybe?

PR #4207, opened 2025-01-25
fix: Drop expiry from internal only JWT

DafyddLlyr:
## What's the problem?
Internal requests from the API to Hasura are failing due to expired JWTs. See recent failures in the `#planx-notifications-internal` Slack channel (OSL).

This is happening as the `$api` client is a singleton, generated once, along with its associated JWT. Calls made by users via the API pass along their JWTs, generated at login, along with their associated roles. The internal JWT used by the API to access Haura is used for various cron jobs, admin endpoints, and other utilities.

Some cron processes are passing as their rely on the admin secret and not the JWT.

We haven't hit this (at time of writing) on staging or production as it's not yet been 24hrs since a deploy.

## What's the solution?
Remove the expiry on internal-only JWTs. This isn't actually required and it's only needed for user instantiated requests.

We could (should?) implement a process where this token is still time-limited, but can be automatically refreshed when expired. I'm just opting for a simple and pragmatic roll-back here just now.

### DIFF
--- a/api.planx.uk/modules/auth/service.ts
+++ b/api.planx.uk/modules/auth/service.ts
@@ -3,7 +3,9 @@ import { $api } from "../../client/index.js";
 import type { User, Role } from "@opensystemslab/planx-core/types";
 import type { HasuraClaims, JWTData } from "./types.js";
 
-export const buildJWT = async (email: string): Promise<string | undefined> => {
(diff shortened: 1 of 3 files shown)


jessicamcinchak (approved):
Thanks for spotting & resolving-for-now !!

PR #4209, opened 2025-01-25
chore: Improve email failure logging

DafyddLlyr:
Quick issue I spotted - it would be nice to expose this session id for debugging. It's currently possible to get this via Hasura events but it's a bit tricker that way round. 

With this id exposed we can re-trigger emails quickly and simply via the API docs.


jessicamcinchak (approved):
Have similarly been meaning to make a quick `emails_log` view for internal troubleshooting/auditing similar to submissions log to bypass paging through native event logs - I can pick that up early next week!

PR #4211, opened 2025-01-26
chore: create `email_events_log` view to access Hasura Event metadata

jessicamcinchak:
A quick followup to https://github.com/theopensystemslab/planx-new/pull/4209

It's a pain to paginate through our Hasura event logs. We already parse out relevant metadata about our submission events to `submissions_services_log` (aka the "Submissions log" in the editor), this does the same for email events now.

No plans to expose this via the editor right now, but hopefully a handy tool for dev debugging in the meantime :+1:

### DIFF
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -247,6 +247,49 @@
         url: '{{$base_url}}/webhooks/hasura/send-slack-notification'
         version: 2
 - table:
+    name: email_events_log
+    schema: public
+  select_permissions:
+    - role: demoUser
+      permission:
+        columns:
+          - retry
+          - response
+          - event_id
+          - event_type
+          - status
+          - created_at
+          - flow_id
+          - session_id
+        filter: {}
+      comment: ""
+    - role: platformAdmin
+      permission:
+        columns:
+          - retry
+          - response
+          - event_id
+          - event_type
+          - status
+          - created_at
+          - flow_id
+          - session_id
+        filter: {}
+      comment: ""
+    - role: teamEditor
+      permission:
+        columns:
+          - retry
+          - response
+          - event_id
+          - event_type
+          - status
+          - created_at
+          - flow_id
+          - session_id
+        filter: {}
+      comment: ""
+- table:
     name: feedback
     schema: public
   object_relationships:
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -249,46 +249,6 @@
 - table:
     name: email_events_log
     schema: public
-  select_permissions:
-    - role: demoUser
-      permission:
-        columns:
-          - retry
-          - response
-          - event_id
-          - event_type
-          - status
-          - created_at
-          - flow_id
-          - session_id
-        filter: {}
-      comment: ""
-    - role: platformAdmin
-      permission:
-        columns:
-          - retry
-          - response
-          - event_id
-          - event_type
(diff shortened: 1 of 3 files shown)


hasura.planx.uk/metadata/tables.yaml
```diff
@@ -246,6 +246,49 @@
...
+          - retry
+          - response
+          - event_id
+          - event_type
+          - status
+          - created_at
+          - flow_id
+          - session_id
+        filter: {}
+      comment: ""
```
jessicamcinchak, line 291: I matched "select" permissions to the `submission_services_log` view here - does this seem fine? Alternatively, can keep stricter until explicit plans to expose in editor. View does _not_ include any personal user data like email address or payment details, only session ID.

hasura.planx.uk/migrations/1737881459011_create_view_email_events_log/up.sql
```diff
@@ -0,0 +1,52 @@
...
+      WHEN ((webhook_conf)::text ~~ '%/send-email/expiry"'::text) THEN 'Session expiry'::text
+      WHEN ((webhook_conf)::text ~~ '%/send-email/reminder"'::text) THEN 'Session reminder'::text
+      WHEN ((webhook_conf)::text ~~ '%/send-email/confirmation"'::text) THEN 'Submission confirmation'::text
+      WHEN ((webhook_conf)::text ~~ '%/send-email/invite-to-pay"'::text) THEN 'Invitation to pay'::text
+      WHEN ((webhook_conf)::text ~~ '%/send-email/invite-to-pay-agent"'::text) THEN 'Invitation to pay - agent'::text
+      WHEN ((webhook_conf)::text ~~ '%/send-email/payment-expiry"'::text) THEN 'Payment expiry'::text
+      WHEN ((webhook_conf)::text ~~ '%/send-email/payment-expiry-agent"'::text) THEN 'Payment expiry - agent'::text
+      WHEN ((webhook_conf)::text ~~ '%/send-email/payment-reminder"'::text) THEN 'Payment reminder'::text
+      WHEN ((webhook_conf)::text ~~ '%/send-email/payment-reminder-agent"'::text) THEN 'Payment reminder - agent'::text
+      ELSE (webhook_conf)::text
```
DafyddLlyr, line 27: Great fallback to catch new templates or other issues 👍 

The fact we have to parse this from here (but it's very useful information) suggests we should maybe consider capturing this more explicitly / pro-actively as an audit log when sending emails.

This is a really solid and practical solution though - love it 👍

hasura.planx.uk/metadata/tables.yaml
DafyddLlyr: Great question!

I'd favour not exposing it all until we need to - if it's intended to be for platform admins only lets capture that here 👍 

If / when we come to exposing it, we set the right permissions then.

I'll checkout the demoUser / `submission_services_log` permissions just now - I would have expected that to have the normal demo user filter of "I can only select my own sessions" applied here.

DafyddLlyr (approved):
Great!

hasura.planx.uk/metadata/tables.yaml
DafyddLlyr: PR here - https://github.com/theopensystemslab/planx-new/pull/4212

jessicamcinchak: Yep makes sense - will rollback & keep as `platformAdmin`-select only

DafyddLlyr (approved):

PR #4212, opened 2025-01-27
fix: Restrict access by role to `submissions_services_summary`

DafyddLlyr:
Configures missing permissions, please see comments and https://github.com/theopensystemslab/planx-new/pull/4211#discussion_r1929697776 for context.

### DIFF
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -1824,6 +1824,25 @@
 - table:
     name: submission_services_summary
     schema: public
+  object_relationships:
+    - name: session
+      using:
+        manual_configuration:
+          column_mapping:
+            session_id: id
+          insertion_order: null
+          remote_table:
+            name: lowcal_sessions
+            schema: public
+    - name: team
+      using:
+        manual_configuration:
+          column_mapping:
+            team_slug: slug
+          insertion_order: null
+          remote_table:
+            name: teams
+            schema: public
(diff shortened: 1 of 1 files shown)


hasura.planx.uk/metadata/tables.yaml
```diff
@@ -1855,7 +1874,11 @@
           - team_slug
           - created_at
           - submitted_at
-        filter: {}
+        filter:
+          team:
+            flows:
+              creator_id:
+                _eq: x-hasura-user-id
```
DafyddLlyr, line 1881: Users with the `demoUser` role can only access the records for flows which they created.

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -1919,7 +1942,14 @@
...
           - submitted_at
-        filter: {}
+        filter:
+          team:
+            members:
+              _and:
+                - user_id:
+                    _eq: x-hasura-user-id
+                - role:
+                    _eq: teamEditor
```
DafyddLlyr, line 1952: Users with the `teamEditor` role can only access records which are associated with teams where they hold the `teamEditor` role.

jessicamcinchak (approved):
Thanks for this folowup ! Currently we were handling this permissions level in the editor-only (eg the "Submissions log" page/route is only accessble to `teamEditors` of said team and `platformAdmins`), but much more robust to have db select access in place too

PR #4213, opened 2025-01-27
fix: Add filetype validation to API

DafyddLlyr:
## What's the problem?
Described as follows by the pen test - 

> **Description**
> The application does not implement proper content filtering for file upload function allowing an attacker to upload a malicious file on the system.
>
> **Recommendation**
> Implement file content and extension filtering before saving the file on the server.

Currently, we only restrict this via the frontend. Files are sent to Scanii for checking, but are still written to the server first as this rightly points out.

## What's the solution?
Add restrictions, using the Multer `fileFilter` ([docs](https://github.com/expressjs/multer?tab=readme-ov-file#filefilter)) to stop the upload of files that don't meet the criteria. I've also added a file size restriction to match the frontend here - not a requirement for the pen test but seemed worthwhile to pick up at the same time.

## To test
 - Upload a HTML (or anything else) file on staging API docs - file upload is accepted
 - Upload a HTML (or anything else) file on Pizza API docs - file upload is rejected

### DIFF
--- a/api.planx.uk/modules/file/file.test.ts
+++ b/api.planx.uk/modules/file/file.test.ts
@@ -39,6 +39,9 @@ vi.mock("@aws-sdk/client-s3", async (importOriginal) => {
   };
 });
 
+const PRIVATE_ENDPOINT = "/file/private/upload";
+const PUBLIC_ENDPOINT = "/file/public/upload";
+
 describe("File upload", () => {
   beforeEach(() => {
     vi.clearAllMocks();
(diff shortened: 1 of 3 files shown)


jessicamcinchak (commented):
Solution here looks great, thanks (once again!) for super quick turnaround on audit issues ! 

One question/consideration (haven't checked pizza yet!): our API's S3 send module currently also shares the `uploadPrivateFile` method and expects to be able to upload a JSON - I believe that's going to be blocked now here?

DafyddLlyr:
> One question/consideration (haven't checked pizza yet, can confirm once built!): our API's S3 send module currently also shares the `uploadPrivateFile` method and expects to be able to upload a JSON - I believe that's going to be blocked now here?

This is a great question and honestly one I hadn't considered. I actually don't think this is an issue however as the restricting in place here just adds middleware to incoming requests to the API to check file extension and file types. 

`uploadPrivateFile()` is part of the service layer, so will accept a JSON file as it doesn't exit the API and make an new HTTP request that will pass through this validation layer.

jessicamcinchak (approved):
@DafyddLlyr thanks, that makes sense !

Swagger testing works exactly as expected for me here. Let's confirm the S3 question once merged to staging before going to prod (think we need to do it against a real council else we just throw "not enabled for this local authority" error early before upload would happen, but I can flag this to Kev & test this afternoon).

DafyddLlyr:
Perfect - thanks very much - sounds like a great plan! Will merge to staging once CI complete and we can test there 👍

PR #4214, opened 2025-01-27
chore: set up power automate tokens for Gloucester and Tewkesbury

jamdelion:
For both councils:
✅ File API key tokens generated and encrypted, and added to Hasura team_integrations
✅ Power automate API key tokens generated and encrypted, and added to Hasura team_integrations
✅ AWS S3 pizza-secrets updated with staging file API key values

Neither have provided us with the Power Automate Webhook URLs yet.


jessicamcinchak (approved):
Perfect! Thanks again for picking these up :raised_hands:

PR #4216, opened 2025-01-28
fix: Use current values to populate `DataAutocomplete` options, not just static initial values

DafyddLlyr:
## What's the problem?
The `DataAutocomplete` component is not aware of the current status of the form, just the initial status. As I enter new values across fields, I should be able to reference and select them within my component. Raised here as an issue on OSL Slack - https://opensystemslab.slack.com/archives/C5Q59R3HB/p1737660734834029?thread_ts=1737640916.451959&cid=C5Q59R3HB

## What's the solution?
Rather than just entering the `formik.initialValues` as the schema, dynamically pass the current `formik.values`. As I enter new values across the current component, I can see this change reflected across all my `DataAutocomplete` components.

## To test..
 Please watch video in Slack thread referenced above for context
 - I can open a new node (e.g. question or checklist)
 - As I enter data values, each autocomplete maintains a full copy of the current options, not just the initial options
 - This works on new and existing nodes

### DIFF
--- a/editor.planx.uk/src/@planx/components/Checklist/Editor/Options.tsx
+++ b/editor.planx.uk/src/@planx/components/Checklist/Editor/Options.tsx
@@ -6,7 +6,7 @@ import ListManager from "ui/editor/ListManager/ListManager";
 import ModalSectionContent from "ui/editor/ModalSectionContent";
 
 import { Option } from "../../shared";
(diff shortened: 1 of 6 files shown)


jessicamcinchak (approved):
Tricky one, thanks for resolving this ! Pizza working as expected !

PR #4218, opened 2025-01-28
fix: ensure reconciliation behaves as expected and does not result in orphaned breadcrumbs

jessicamcinchak:
In-progress investigation


api.planx.uk/modules/saveAndReturn/service/validateSession.ts
```diff
@@ -135,8 +135,8 @@ async function reconcileSessionData({
     }
   };
 
-  // remove all auto-answered breadcrumbs
-  // (auto-answers are reconstructed in the editor by `upcomingCards`)
+  // remove all auto-answered breadcrumbs because their automation may rely on the same data values as an altered node
+  //   (auto-answers can be reconstructed on forwards navigation if they are still auto-answerable on resume)
```
jessicamcinchak, line 139: See existing `test('auto-answered breadcrumbs with matching "data.val" values are also removed')` in `validateSession.test.ts` to better understand this. I ultimately believe this is _correct_ behavior

jessicamcinchak:
This one has sat open for too long! Would like to propose merging as-is with very minor comment adjustment then commiting to revisit the functional reconciliation questions here alongside upcoming templates / publishing changes.

DafyddLlyr (approved):

PR #4219, opened 2025-01-28
fix: Add file extension check to filename parameter in body

DafyddLlyr:
## What's the problem?
From Phil at Jumpsec - 

> I've looked at the file upload issue again and have noticed that whilst the restrictions have been implemented on the initial filetype uploaded, the "filename" parameter of the upload forms can be used to change the file extension of the file. I've attached a screenshot of this to demonstrate what I mean. I'm still exploring the exploitability of this, but would you be able to implement the same checks on the "filename" parameter file extension value as well to make sure this also matches the allowed list of file types? 

## What's the solution?
By adding the same validation to our Zod schema, we can share this validation check across the values in the body of the incoming http request, as well as the attached file.

## To test
- Upload a file with a valid extension / mime type (e.g. `myImage.png`)
- Use an invalid extension for the `filename` param (e.g. `myImage.exe`)
- Get a Zod validation error ✅ 

image

### DIFF
--- a/api.planx.uk/modules/file/controller.ts
+++ b/api.planx.uk/modules/file/controller.ts
@@ -5,6 +5,7 @@ import { getFileFromS3 } from "./service/getFile.js";
 import { z } from "zod";
 import type { ValidatedRequestHandler } from "../../shared/middleware/validate.js";
 import { ServerError } from "../../errors/index.js";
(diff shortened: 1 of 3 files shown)


jessicamcinchak (approved):
Swagger docs working exactly as expected for me here, happy for this to go once tests are updated/passing too !

PR #4220, opened 2025-01-28
chore: adjust row and column access permissions in Hasura

jessicamcinchak:
Per audit feedback:
- [x] `GetSubmissions` 
  - This is called via editor Apollo client (not via `$api` role so unchanged)
  - Queries `submission_services_log` view which previously did not have row-level permissions configured
    - Updated query definition to join to flows to access `team_id`, established "relationship" to `teams`, added row-level permissions for both `teamEditor` & `demoUser` roles
- [x] `GetFeedbackForFlow`
  - This is called via editor Apollo client (not via `$api` role so unchanged)
  - Queries `feedback_summary` view which DID have row-level permissions configured already
    - Updated column-level permissions to toggle-all rather than omit recently added column

Following changes are rolled back per PR feedback / comments below:
- `GetUsersForTeam`
  - This is called via editor Apollo client (not via `$api` role so unchanged)
  - Queries `users` table which does not have row-level permissions configured
    - Added row-level "select" permissions for both `teamEditor` & `demoUser` roles, toggled all columns (any reason no one was allowed to "select" `staging_only` before?)
- `GetTeamBySlug`
  - This is called via API "move flow" and defined in planx-core
  - Queries `teams` with relationship to `themes` and `settings`
    - `teams` does not have row-level "select" permissions configured for the `teamEditor` role - these are now added
    - `team_themes` has full select-all access across rows and row-level "update" permissions already - I think this is correct
    - `team_settings` handled in point below
- `GetTeamSettings`
  - This is defined in planx-core and called via editor
  - Queries `team_settings` table which does not have row-level "select" permissions configured (only "update") 
    - Added row-level "select" permissions for both `teamEditor` & `demoUser` roles
- `GetFlow`
  - This is called via editor
  - Queries `flows` table which only has row-level permissions configured for `demoUser`
    - Column-level permissions are out of date here and randomly omitting a lot of recently added fields (eg `templated_from`, `description`) - do we want this or select-all ? 

**Future todo:**
- Update all applicable introspection tests to match new configs - or add more E2E coverage for these scenarios !

### DIFF
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -2664,13 +2664,14 @@
       permission:
         check: {}
         columns:
-          - id
-          - first_name
-          - last_name
           - created_at
-          - updated_at
-          - is_platform_admin
           - email
+          - first_name
+          - id
+          - is_platform_admin
+          - is_staging_only
+          - last_name
+          - updated_at
   select_permissions:
     - role: api
       permission:
@@ -2687,36 +2688,49 @@
     - role: demoUser
       permission:
         columns:
-          - created_at
-          - email
-          - first_name
           - id
-          - is_platform_admin
+          - first_name
           - last_name
+          - created_at
           - updated_at
-        filter: {}
+          - is_platform_admin
+          - is_staging_only
+          - email
+        filter:
+          created_flows:
+            creator_id:
+              _eq: '"x-hasura-user-id"'
     - role: platformAdmin
       permission:
         columns:
           - id
           - first_name
           - last_name
-          - email
           - created_at
           - updated_at
           - is_platform_admin
+          - is_staging_only
+          - email
         filter: {}
     - role: teamEditor
       permission:
         columns:
-          - created_at
-          - email
-          - first_name
           - id
-          - is_platform_admin
+          - first_name
           - last_name
+          - created_at
           - updated_at
-        filter: {}
+          - is_platform_admin
+          - is_staging_only
+          - email
+        filter:
+          teams:
+            _and:
+              - user_id:
+                  _eq: '"x-hasura-user-id"'
+              - role:
+                  _eq: teamEditor
+      comment: ""
   update_permissions:
     - role: platformAdmin
       permission:
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -2721,7 +2721,7 @@
         filter:
           created_flows:
             creator_id:
-              _eq: '"x-hasura-user-id"'
+              _eq: x-hasura-user-id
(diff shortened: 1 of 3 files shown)


hasura.planx.uk/metadata/tables.yaml
```diff
@@ -2661,36 +2710,49 @@
...
+          - created_at
           - updated_at
-        filter: {}
+          - is_platform_admin
+          - is_staging_only
+          - email
+        filter:
+          created_flows:
+            creator_id:
+              _eq: '"x-hasura-user-id"'
```
DafyddLlyr, line 2724:
```suggestion
              _eq: x-hasura-user-id
```

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -2661,36 +2710,49 @@
...
           - updated_at
-        filter: {}
+          - is_platform_admin
+          - is_staging_only
+          - email
+        filter:
+          teams:
+            _and:
+              - user_id:
+                  _eq: '"x-hasura-user-id"'
```
DafyddLlyr, line 2752:
```suggestion
                  _eq: x-hasura-user-id
```

Hasura sneakily wraps their input in quotation marks - this has caught me out before.

image

DafyddLlyr (changes requested):
A few things I think we need to roll back here and explain to the pen testers in terms of our data model.

The changes for feedback and submissions are correct and ones we've missed though 👍 

`GetUsersForTeam`
> Added row-level "select" permissions for both teamEditor & demoUser roles, toggled all columns (any reason no one was allowed to "select" `staging_only` before?)

`staging_only` was just unchecked because it wasn't used - just used in the API in one place. We can allow access (or not) dependant on permissions - it's not a sensitive field.

I believe this was intentional that anybody logged in could access users - we call this table in order to populate the "last edited by" text in a few places. 

I think it's totally fine for our permission model that any logged in user can access the firstName and lastName of all users (as they can view their flows, edits, etc).

---

`GetTeamBySlug`
I think this was right as it was - it's possibly not been understood by the pen testers how our multi-tenancy works. `teamEditor` users should be able to select other teams.

---

`GetTeamSettings`
This was already correct I believe and should match the `public` role - the only _slightly_ sensitive column is `submission_email`. I don't think this is an issue for logged in users realistically, but we could move this to `team_integrations` to properly limit it.

---

`GetFlow`
This is and should be public ✅ Good question re:new readme columns - happy to leave as-is for now for simplicity and turn on when we need them but no strong opinions.

hasura.planx.uk/metadata/tables.yaml
jessicamcinchak: Sneaky indeed - thanks for catching these - didn't even think to check the autogenerated syntax !

PR #4226, opened 2025-01-29
feat: last edited timestamp in flows

jamdelion:


editor.planx.uk/src/components/Footer/Footer.tsx
```diff
@@ -48,14 +50,25 @@ export interface Props {
...
   return (
     <Root>
       <Container maxWidth={false}>
-        <ButtonGroup py={0.5}>
-          {items
-            ?.filter((item) => item.title)
-            .map((item) => <FooterItem {...item} key={item.title} />)}
-        </ButtonGroup>
+        <Box display={"flex"} justifyContent={"space-between"}>
+          {items && items.length > 0 && (
```
jamdelion, line 59: If no items, hide the `ButtonGroup` container so that the timestamp displays on the left

editor.planx.uk/src/routes/views/published/publishedView.tsx
```diff
@@ -0,0 +1,69 @@
...
+export const publishedView = async (req: NaviRequest) => {
+  const flowSlug = req.params.flow.split(",")[0];
+  const teamSlug =
+    req.params.team || (await getTeamFromDomain(window.location.hostname));
+  const data = await fetchSettingsForPublishedView(flowSlug, teamSlug);
+  const flow = data.flows[0];
+
+  const lastPublishedDate = await getLastPublishedAt(flow.id);
+  useStore.setState({ lastPublishedDate });
+
```
jamdelion, line 35: The only new lines in this file (I slightly reorganised hence the diff is bigger)

editor.planx.uk/src/routes/views/published/publishedView.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const state = useStore.getState();
+  // XXX: necessary as long as not every flow is published; aim to remove dataMergedHotfix.ts in future
+  // load pre-flattened published flow if exists, else load & flatten flow
+  state.setFlow({
+    id: flow.id,
+    flow: publishedFlow,
+    flowSlug,
+    flowStatus: flow.status,
+    flowName: flow.name,
+  });
```
DafyddLlyr, line 54: Question (genuinely not sure of the answer!): Would it be simpler / cleaner to add lastPublishedData as a flow property rather than it's own value in the store?

editor.planx.uk/src/routes/views/published/queries.ts
DafyddLlyr, line 1: Nice change 👌

DafyddLlyr (approved):
Looking great - simple, clean, short & sweet.

One small test issue, but good to go once sorted 🚀

editor.planx.uk/src/routes/views/published/queries.ts
```diff
@@ -0,0 +1,98 @@
...
+import { NotFoundError } from "navi";
+
+import { PublishedViewSettings } from "./publishedView";
+
+export const getLastPublishedAt = async (flowId: string): Promise<string> => {
+  try {
+    const { data } = await publicClient.query({
+      query: gql`
+        query GetLastPublishedFlow($id: uuid) {
+          flows(limit: 1, where: { id: { _eq: $id } }) {
```
jessicamcinchak, line 12: nit: I think `flows_by_pk` would accomplish the limit & where by default here

editor.planx.uk/src/routes/views/published/queries.ts
```diff
@@ -0,0 +1,98 @@
...
+              }
+              slug
+            }
+            settings
+            status
+            publishedFlows: published_flows(
+              limit: 1
+              order_by: { created_at: desc }
+            ) {
+              data
```
jessicamcinchak, line 80: question: is there a reason we can't simply also fetch/return `created_at` here and instead are now making two separate queries?

jessicamcinchak (commented):
Looks good ! Two small comments re: queries

editor.planx.uk/src/routes/views/published/queries.ts
jessicamcinchak: Since your only addition is a simple date, I'd go same query here (considering the existing one is already doing the heavy lifting by fetching `data`) ! If the other way around and your addition was a large piece of data, then maybe more justification to introduce as separate and not complicate an existing query - but I agree single query is probably our most performant bet here!

editor.planx.uk/src/routes/views/published/publishedView.tsx
jamdelion: Hadn't actually considered adding as a flow property, but looking at it again I think it was because I wanted to tie it up with the existing `lastPublished` function/query which is in the editor store - I added a line there to set `lastPublishedDate` as a result of that. 

Originally I tried just calling that existing function as we do here: https://github.com/theopensystemslab/planx-new/blob/9cd34319c3ef0cb32c2bc6a7c27cb27282414d2e/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Publish/PublishFlowButton.tsx#L95
But I think there was a mismatch between store state and db or something and it didn't work! 😅

PR #4227, opened 2025-01-29
feat: add "TEMPLATES" feature flag and skeleton UI

jessicamcinchak:
It's a basic start !
- Adds a `"TEMPLATES"` feature flag
- When toggled on, will show an additional "Start from a template" button to editors, which opens a modal with basic placeholder text, a select picker (to be populated for real in a future PR once we have a way of tagging & querying flows as a "source" template), and two buttons - one to simply close/go back and one to create the template you've selected (to be wired up next)

Goal here is to keep templates PRs small & quick, so that as much feature-flagged, "skeleton" UI can get to `main` as frequently as possible, allowing design refinement to happen in parallel to technical development.

Screenshot from 2025-01-29 21-51-46


DafyddLlyr (approved):
Perfect - short and simple, very much agree this is the way to go here 💪

PR #4230, opened 2025-01-30
feat: automatically publish newly created flows

jessicamcinchak:
Realising this will be nice to have in place _before_ templates - as I think we'll want a new `createTemplate` function to follow a similar pattern (offline by default, but is already published)!

Applies to both creating a new flow and copying a flow.

E2E tests rely on methods defined in `planx-core` which are updated here https://github.com/theopensystemslab/planx-core/pull/639

**Testing:**
- When creating via "Add a new service" button or "Copy" menu, ensure that on initial navigation the flow is published (
  - This is indicated underneath "Check for changes to publish" button in text and via console command `window.api.getState().isFlowPublished`
- Make changes, "Check for changes to publish", and ensure that you can manually publish the first time as normal/expected


jessicamcinchak:
[Set new services offline by default but automatically published](https://trello.com/c/M2SoPoPp/3129-set-new-services-offline-by-default-but-automatically-published)

api.planx.uk/helpers.ts
```diff
@@ -63,7 +65,7 @@ interface InsertFlow {
 }
 
 // Insert a new flow into the `flows` table
-const insertFlow = async (
+const createFlow = async (
```
jessicamcinchak, line 68: In my opinion, the editor store function `createFlow` _should_ be calling this API method instead of doing a frontend operation ! Have made a note of this for now, but intentionally choosing not to refactor here to keep momentum on templates as a whole - will revisit when adding methods for "createTemplatedFlow"

e2e/tests/ui-driven/src/create-flow.spec.ts
```diff
@@ -114,39 +114,7 @@ test.describe("Flow creation, publish and preview", () => {
     ]);
   });
 
-  test("Cannot preview an unpublished flow", async ({
```
jessicamcinchak, line 117: E2E tests always start from an empty database and create a flow from scratch which now _includes_ publishing - meaning they no longer have the concept of ever accessing an "unpublished" flow or needing to test initial publish action separately.

The "publishFlow" action is still called throughout this file to save/publish content changes.

api.planx.uk/helpers.ts
DafyddLlyr: Yep - agreed - we want to be narrowing this down to (ideally) a single implementation.

DafyddLlyr (approved):
[One comment](https://github.com/theopensystemslab/planx-core/pull/639#discussion_r1937381899) in https://github.com/theopensystemslab/planx-core/pull/639 which is worth thinking about / trying out, but no blockers! 👌

PR #4231, opened 2025-01-30
refactor: split Rich Text Input file into smaller files

jamdelion:
🔥 Blazing in to completely split this file up as it was my most hated file to look at (so far...). 

Using the excuse of the next ticket on my radar (https://trello.com/c/Sv8IfEVi/3206-add-validation-to-editor-modals-to-prevent-made-policy-reference-links-from-being-added-to-help-text) to do this now!

**Summary:**
- `RichTextInput.tsx` now mostly just contains the main input component and editor.
- `/components` sub-folder contains all other defined components that were in the main folder, along with newly extracted `RichTextInputButtons` (such as `H1Button`)
- Styled components extracted to `styles.ts`
- Type definitions extracted to `types.ts`
- Most of the helper functions extracted to `utils.ts`

Now much easier to read in my opinion! 🤭

### DIFF
--- a/editor.planx.uk/src/ui/editor/RichTextInput/RichTextInput.tsx
+++ b/editor.planx.uk/src/ui/editor/RichTextInput/RichTextInput.tsx
@@ -8,241 +8,37 @@ import FormatListNumbered from "@mui/icons-material/FormatListNumbered";
 import LinkIcon from "@mui/icons-material/Link";
 import Box from "@mui/material/Box";
 import IconButton from "@mui/material/IconButton";
-import { type Editor, type JSONContent } from "@tiptap/core";
-import Bold from "@tiptap/extension-bold";
-import BulletList from "@tiptap/extension-bullet-list";
-import Document from "@tiptap/extension-document";
-import HardBreak from "@tiptap/extension-hard-break";
-import Heading from "@tiptap/extension-heading";
+import { type Editor } from "@tiptap/core";
 import History from "@tiptap/extension-history";
-import Italic from "@tiptap/extension-italic";
-import Link from "@tiptap/extension-link";
-import ListItem from "@tiptap/extension-list-item";
 import Mention from "@tiptap/extension-mention";
-import OrderedList from "@tiptap/extension-ordered-list";
-import Paragraph from "@tiptap/extension-paragraph";
 import ExtensionPlaceholder from "@tiptap/extension-placeholder";
-import Text from "@tiptap/extension-text";
-import { generateHTML, generateJSON } from "@tiptap/html";
-import { EditorContent, ReactRenderer, useEditor } from "@tiptap/react";
-import { map } from "ramda";
+import { EditorContent, useEditor } from "@tiptap/react";
 import React, {
   ChangeEvent,
   type FC,
-  forwardRef,
   useCallback,
   useEffect,
-  useImperativeHandle,
   useRef,
   useState,
 } from "react";
-import tippy, { type Instance } from "tippy.js";
-import { create } from "zustand";
 
 import Input from "../../shared/Input/Input";
 import PublicFileUploadButton from "../../shared/PublicFileUploadButton";
-import CustomImage from "../RichTextImage";
 import { PopupError } from "./components/PopUpError";
+import { suggestion } from "./components/suggestion";
+import { RichContentContainer, StyledBubbleMenu } from "./styles";
+import { commonExtensions } from "./tiptapExtensions";
+import { Props } from "./types";
 import {
-  MentionItems,
-  MentionItemsButton,
-  MentionItemsEmpty,
-  RichContentContainer,
-  StyledBubbleMenu,
-} from "./styles";
-import { MentionListProps, Props } from "./types";
-
-const passportClassName = "passport";
-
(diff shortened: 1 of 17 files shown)


editor.planx.uk/src/ui/editor/RichTextInput/utils.ts
```diff
@@ -0,0 +1,161 @@
...
+    }
+    if (typeof val === "object") {
+      return map(modifyDeep(fn), val);
+    }
+    return val;
+  };
+
+// Generic value that `modifyDeep` works off of. Since it can be object, array or primitive, any typing more accurate than `any` is not compiling at the moment.
+// TODO: find a better typing alternative
+type Value = any;
```
RODO94, line 34: I think `planx-core` has a `Value` type which might be better than an any because it shouldn't be `undefined`? 

`planx-core/src/types/data.ts`

editor.planx.uk/src/ui/editor/RichTextInput/RichTextInput.tsx
RODO94, line 1: nit: could be worth destructuring the props to remove the dependency suggestion on line 109

editor.planx.uk/src/ui/editor/RichTextInput/components/MentionList.tsx
```diff
@@ -0,0 +1,98 @@
...
+          onClick={() => {
+            props.command({ id: props.query });
+            variablesStore.addVariable(props.query);
+          }}
+        >
+          + Add <span className={passportClassName}>@{props.query}</span>
+        </MentionItemsButton>
+      )}
+      {props.items.length > 0 ? (
+        props.items.map((item: any, index: number) => (
```
RODO94, line 81: The type for `item` should be `Placeholder`?

RODO94 (commented):
Had a code review with a few suggestions, was going to do some manual testing now

RODO94 (approved):
Other than those other comments, I like the structure and it all works for me after testing manually

editor.planx.uk/src/ui/editor/RichTextInput/utils.ts
DafyddLlyr, line 1: There's two groupings of functions here as I see it - 
 - Actual helpers/utility functions
 - Validation function (`getLinkNewTabError()`, `linkSelectionError()` etc)

I think this distinction is meaningful, particularly in light of the upcoming ticket which inspired this PR. It's probably worth splitting these into their own files 👍

editor.planx.uk/src/ui/editor/RichTextInput/types.ts
```diff
@@ -0,0 +1,24 @@
+// eslint-disable-next-line no-restricted-imports
+import type { InputBaseProps } from "@mui/material";
```
DafyddLlyr, line 2:
```suggestion
import type { InputBaseProps } from "@mui/material/InputBase";
```

editor.planx.uk/src/ui/editor/RichTextInput/RichTextInput.tsx
```diff
@@ -478,12 +131,13 @@ const RichTextInput: FC<Props> = (props) => {
       return;
     }
     const editorValue = internalValue.current || toHtml(editor.getJSON());
-    if (props.value !== editorValue) {
-      internalValue.current = stringValue;
-      const doc = fromHtml(stringValue);
-      setContentHierarchyError(getContentHierarchyError(doc));
-      editor.commands.setContent(doc);
+    if (props.value === editorValue) {
+      return;
```
DafyddLlyr, line 135: Love a guard clause 👍

editor.planx.uk/src/ui/editor/RichTextInput/RichTextInput.tsx
```diff
@@ -557,68 +212,12 @@ const RichTextInput: FC<Props> = (props) => {
...
-                }}
-              >
-                <FormatListNumbered />
-              </IconButton>
+              <H1Button editor={editor} />
+              <H2Button editor={editor} />
+              <BoldButton editor={editor} />
+              <ItalicButton editor={editor} />
+              <BulletListButton editor={editor} />
+              <OrderedListButton editor={editor} />
```
DafyddLlyr, line 220: So much better!

editor.planx.uk/src/ui/editor/RichTextInput/RichTextInput.tsx
jamdelion: I've destructured the props but found that unexpected things happen when implementing the suggested changes to this dep array so haven't changed anything else. There are other dep arrays that I've updated and all looks ok

DafyddLlyr (approved):
Appreciate this! Ventured into the file to take a look at the `multiline` / `rows` question and it was pretty tricky to understand what was what 🌀

editor.planx.uk/src/ui/editor/RichTextInput/utils.ts
jamdelion: Thanks - I delved into this one a little but the Value type is not quite robust enough 😞 Have added this to my list to come back to at some point!

PR #4234, opened 2025-01-31
feat: Do not allow template flows to be deleted / archived

DafyddLlyr:
## What does this PR do?
 - Removes "delete" button from the frontend UI - to be replaced by "Archive" [in this ticket](https://trello.com/c/fhoa1wCa/3166-add-ability-to-archive-flows-instead-of-deleting-them?filter=archive)
 - Adds `flows.deleted_at` column in order to allow a soft delete pattern

### Permissions
 - Removes "delete" permissions from all roles for the `flows` table
 - Allows all roles to set the `flows.deleted_at` column (within the parameters of their current permission levels)
 - Does not allow any roles to soft delete (set `flows.deleted_at`) for template flows

Some of this is pretty tricky to understand via the `tables.yaml` and took a bit of trial and error to figure out. Some e2e test coverage is needed here to explain this and act as a guard rail for future changes. I'm planning on introducing this in a future PR so that I'm not holding up the "archive flows" ticket @RODO94 will be shortly working on. 

### To manually test - 
 - Log into PlanX to get a JWT
 - Use this in the Hasura console, test out mutations and queries based on the permission level of your JWT
 - E2E tests incoming here on another PR 🚀

### DIFF
--- a/e2e/tests/api-driven/src/demo-workspace/helper.ts
+++ b/e2e/tests/api-driven/src/demo-workspace/helper.ts
@@ -277,10 +277,13 @@ export async function updateFlow(client, flowId: UUID): Promise<UUID> {
 }
 
 export async function deleteFlow(client, flowId: UUID): Promise<UUID> {
-  const { delete_flows_by_pk: response } = await client.request(
+  const { update_flows_by_pk: response } = await client.request(
     gql`
-      mutation deleteFlow($flowId: uuid!) {
-        delete_flows_by_pk(id: $flowId) {
+      mutation softDeleteFlow($flowId: uuid!) {
+        update_flows_by_pk(
+          pk_columns: { id: $flowId }
+          _set: { deleted_at: "now()" }
+        ) {
           id
         }
       }
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -811,6 +811,7 @@
           - created_at
           - creator_id
           - data
+          - deleted_at
           - description
           - id
           - limitations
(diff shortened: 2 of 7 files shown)


hasura.planx.uk/metadata/tables.yaml
```diff
@@ -649,7 +649,9 @@
           - version
         computed_fields:
           - data_merged
-        filter: {}
+        filter:
+          deleted_at:
+            _is_null: true
```
DafyddLlyr, line 654: This repeated pattern means that archived flows cannot be selected.

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -798,11 +812,20 @@
...
+                    _eq: x-hasura-user-id
+            - deleted_at:
+                _is_null: true
+        check:
+          _and:
+            - deleted_at:
+                _is_null: false
+            - templated_flows:
+                id:
+                  _is_null: false
```
DafyddLlyr, line 828: This repeated pattern is a post-update check which ensures that flows cannot be archived if they have associated templated flows derived from them.

e2e/tests/api-driven/src/demo-workspace/helper.ts
DafyddLlyr, line 1: This change converts this request to an `UPDATE` instead of a `DELETE` operation.

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -793,7 +804,12 @@
           - team_id
           - updated_at
           - version
-        filter: {}
+        filter:
+          _and:
+            - is_template:
+                _eq: false
```
DafyddLlyr, line 810: This repeated pattern means that this role cannot update a template.

jessicamcinchak (approved):
Works as expected for me, thanks :+1:

PR #4235, opened 2025-01-31
feat: Set up relationship between a template flow and children derived from it

DafyddLlyr:
Quick follow up to #4232 

Something I missed here is that I need a relationship established between a parent template flow, and all of its children (flows derived from it).

This is what we'll use to cascade changes from parent → child, and to block template flows which have "children" to be deleted/archived.

### DIFF
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -485,7 +485,7 @@
           table:
             name: published_flows
             schema: public
-    - name: template_children
+    - name: templated_flows
       using:
         manual_configuration:
           column_mapping:
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -485,6 +485,15 @@
           table:
             name: published_flows
             schema: public
+    - name: template_children
+      using:
+        manual_configuration:
+          column_mapping:
+            id: templated_from
+          insertion_order: null
+          remote_table:
+            name: flows
+            schema: public
   computed_fields:
     - name: data_merged
       definition:


hasura.planx.uk/metadata/tables.yaml
```diff
@@ -485,6 +485,15 @@
           table:
             name: published_flows
             schema: public
+    - name: template_children
```
DafyddLlyr, line 488: Not convinced that "children" is really the right term here - very open to suggestions.

jessicamcinchak: Are they `templated_flows` (again following `published_flows` pattern)?

Many published flows or templated flows to one "source" flow is my evolving mental model?

RODO94 (commented):
Would be cool to look through the pizza Hasura to manually check this.

Is this a new table in the database which captures the `templated_from` field in the `flows` table linking it with the `flow.id`?

Is there a need to update `seed-database` ? I couldn't run `pnpm run up` locally on this branch

hasura.planx.uk/metadata/tables.yaml
DafyddLlyr: Much better!

RODO94:
> > Would be cool to look through the pizza Hasura to manually check this.
> > Is this a new table in the database which captures the `templated_from` field in the `flows` table linking it with the `flow.id`?
> > Is there a need to update `seed-database` ? I couldn't run `pnpm run up` locally on this branch
> 
> There's no new column introduced in this PR so no update required to the seed script. This is a relationship set up in Hasura which is adding a relationship between the tables on the underlying PostreSQL database.
> 
> There's a `flows.templated_from` column already (added here - #3945)
> 
> I'm rebuilding Vultr just now so you can test on the Pizza. What was the local issue you were having @RODO94?

Yeah just seen that locally now, makes more sense.

Error was: 

psql:write/flows.sql:19: ERROR:  extra data after last expected column
CONTEXT:  COPY sync_flows, line 1: "3d9e67ad-f7b6-432a-9bc1-cbb74f20db87;1;london-data-hub;84;"{""_root"": {""edges"": [""6JctE36vQD"", ..."

DafyddLlyr:
@RODO94 Fixed here - https://github.com/theopensystemslab/planx-new/pull/4236

PR #3989, opened 2024-11-20
chore: Uprade aws-sdk to v3

jamdelion:
I'm not particularly familiar with configuring AWS so have found this quite tricky, despite the helpful [codemod script](https://github.com/aws/aws-sdk-js-codemod) recommended by AWS.

I've added some comments on the code below with some of the issues/questions I've had!

---

Picking up from the work @jamdelion has started here!

## What does this PR do?
 - Update aws-sdk from v2 (deprecated) to v3
 - Updates associated functions, types and mocks

PR #4237, opened 2025-01-31
[performance] add API load testing script

freemvmt:

Main thrust of this PR is the addition of the `test_api.py` load testing script (see final commit).

However, to get it working I had to make some other changes, such as adding a mechanism to bypass rate limiting, and exposing a file deletion endpoint. They are probably best considered on a per commit basis. I'm happy to split this up into multiple PRs if that's preferable :)

I'll also do a review and tag this up with some additional context.

Once this is merged we should be able to run the new script against the API on staging, to get an idea of whether it can handle the same load we are confident the data layer can (i.e. Hasura + RDS).

When we do that I'll track results [here](https://www.notion.so/opensystemslab/Testing-the-API-17d35d469ad180049b17ea1731c3b221).


DafyddLlyr:
Rebase to `main` - please see https://opensystemslab.slack.com/archives/C04EEBR1E0P/p1738574105777659

.env.example
```diff
@@ -39,6 +39,8 @@ FILE_API_KEY_DONCASTER=👻
 FILE_API_KEY_GLOUCESTER=👻
 FILE_API_KEY_TEWKESBURY=👻
 
+# Used to circumvent API rate limiting for development purposes (e.g. load testing)
+SKIP_RATE_LIMIT_SECRET=👻
```
freemvmt, line 43: I've already generated a 40char alphanumeric secret and pushed it up to S3, so it's ready to pull down

api.planx.uk/modules/auth/passport.ts
```diff
@@ -14,7 +19,29 @@ export default async (): Promise<Authenticator> => {
   const customPassport = new passport.Passport();
 
   // instantiate Microsoft OIDC client, and use it to build the related strategy
-  const microsoftIssuer = await Issuer.discover(MICROSOFT_OPENID_CONFIG_URL);
+  // we also keep said config as a fixture to enable offline local development
```
freemvmt, line 22: this change was literally just to allow me to run the API for development purposes (i.e. do `pnpm dev`) while I was on a train with no internet, so it doesn't necessarily need to make it into the codebase, but could be useful for others?

afaik fetching the config for the Microsoft OIDC stuff is the only call the code makes when spinning up, so it was possible to run `pnpm dev` to completion while offline before we added that SSO option, but now it isn't (without this change)

api.planx.uk/modules/file/routes.ts
```diff
@@ -36,16 +37,23 @@ router.post(
...
 router.get(
   "/file/private/:fileKey/:fileName",
   useNoCache,
   useFilePermission,
-  validate(downloadFileSchema),
+  validate(hostedFileSchema),
   privateDownloadController,
 );
 
+router.delete(
```
freemvmt, line 52: would definitely like to hear whether anyone thinks there's a risk associated with exposing this endpoint? I've made it accessible only to anyone with platform level admin powers.

the utility of it here is that we can upload as many files to staging S3 as we like during load testing without worrying about incurring additional cost, because we can also delete them immediately

api.planx.uk/modules/file/service/utils.ts
```diff
@@ -23,7 +23,7 @@ function useMinio() {
     // Points to Minio
     return {
       endpoint: `http://minio:${process.env.MINIO_PORT}`,
-      s3ForcePathStyle: true,
+      forcePathStyle: true,
```
freemvmt, line 26: this took me a good while to hunt down! a pesky little breaking change resulting from our move to v3 `@aws-sdk/client-s3` in #3989 that was causing 500 errors with the following msg:

```
ServerError: Failed to upload public file: getaddrinfo ENOTFOUND pizza-user-uploads.minio
```

api.planx.uk/rateLimit.ts
```diff
@@ -7,6 +7,19 @@ const apiLimiter = rateLimit({
   max: 250,
   standardHeaders: true,
   legacyHeaders: false,
+  skip: (req: Request, _res: Response) => {
+    // add a mechanism for skipping rate limit when load testing (on local or staging only)
+    if (process.env.APP_ENVIRONMENT == "production") return false;
```
freemvmt, line 12: without this (or a more convoluted solution like hosting Locust workers on remote servers with different IPs) I can't exceed the rate limit of 250 reqs per IP per 10 minute period, which of course is not sufficient for load testing purposes!

infrastructure/performance/test_api.py
```diff
@@ -0,0 +1,100 @@
...
+        headers={
+          "Authorization": f"Bearer {AUTH_TOKEN}",
+          "X-Skip-Rate-Limit-Secret": SKIP_RATE_LIMIT_SECRET,
+        },
+      ) as resp:
+        if resp.status_code != 204:
+          return resp.failure("Failed to delete file")
+        resp.success()
+
+  # TODO: add another task which hits the /upload-submission endpoint
```
freemvmt, line 97: I haven't managed to implement this part, but I think the above task should suffice for our purposes for this round of load testing

api.planx.uk/modules/auth/passport.ts
jessicamcinchak: I'd say no harm in keeping this around! Fixture may additionally come in handy for future E2E test mocks if we ever want Microsoft SSO coverage?

api.planx.uk/modules/file/routes.ts
jessicamcinchak: Exposing to `platformAdmins` seems fine by me ! 

"Public" files are pretty exclusively flow content (eg vector drawings like roof pitch examples) used throughout services uploaded by editors; all have original copies on OSL's Google Drive, so low risk here! (Whereas "/file/private/..." are uploaded by applicants, so it's ideal to keep deletion of those restricted to session sanitation jobs.)

api.planx.uk/modules/file/service/utils.ts
jessicamcinchak: Thanks for shaking some of these out - went unnoticed for a good while!

api.planx.uk/rateLimit.ts
jessicamcinchak: Only thing I'm not totally following here is why we need a request header _and_ env var? Setting both and checking their equality is just an extra layer of assurance that we're bypassing intentionally?

jessicamcinchak (approved):
Code looks great here - one minor comment to address around Pulumi secret-getting, but otherwise think this is good to merge & test on staging :slightly_smiling_face: 

Happy to have another go with the Python script instructions once merged :+1:

api.planx.uk/modules/auth/passport.ts
freemvmt: @jessicamcinchak ok great, that makes sense as well, would be good to have this tested!

infrastructure/application/index.ts
```diff
@@ -392,6 +392,10 @@ export = async () => {
             name: "FILE_API_KEY_TEWKESBURY",
             value: config.requireSecret("file-api-key-tewkesbury"),
           },
+          {
+            name: "SKIP_RATE_LIMIT_SECRET",
+            value: config.requireSecret("skip-rate-limit-secret"),
+          },
```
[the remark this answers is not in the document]
freemvmt: great catch!

api.planx.uk/rateLimit.ts
jessicamcinchak: Latest commit feels much clearer, and still safe/catious :+1:

PR #4241, opened 2025-02-03
fix: switch order of column select in `seed-database` flows scripts

RODO94:
Seed scripts were failing and showing the error with the input mistake as it was trying to add `<p>...</p>` to the `templated_from` column which expected a `uuid` type. Seems like this is a simple ordering issue of seed script changes happening close together and in fact, `<p>...</p>` should be trying to update the `description` column.

I have switched these round locally, and it is working as expected

#### For testing

Here's the problem `flow_id` : `9d70b18e-be0d-4f21-944c-a129da271ebd`

### DIFF
--- a/scripts/seed-database/write/flows.sql
+++ b/scripts/seed-database/write/flows.sql
@@ -13,8 +13,8 @@ CREATE TEMPORARY TABLE sync_flows (
   analytics_link text,
   status text,
   name text,
-  templated_from uuid,
   description text,
+  templated_from uuid,
   summary varchar(120),
   limitations text
   );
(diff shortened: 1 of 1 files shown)


jessicamcinchak (approved):
Thanks for fix here, pizza looks good! 

To be totally clear about what problem is:
- Our sync script does a `SELECT * FROM flows` to get data, but then the `write` step was trying to insert columns in a different order than they were selected. 

It's _not_ an issue with a single flow UUID (rather that id you have listed is probably just the first to fail the insertion) !

jamdelion (approved):
Checked the sync-data script locally and working as expected for me 👍

PR #4244, opened 2025-02-03
feat: additional error messages for duplicate flow name and moving a flow

RODO94:
## Where does this PR come from?

This PR is linked to this Trello ticket: https://trello.com/c/Lsjy4ZgQ/3195-add-descriptive-error-message-for-duplicate-slug-when-copying-a-service-and-moving-it-to-a-new-team

It comes from a situation where a flow was copied and the user tried to move it to another team, but the other team already had a flow with the same name. No useful error was shown to point the user towards how to resolve it. This is what I have added here.

## What has been added?

To guide users towards resolving errors themselves, I have added two new pathways / error messages.

1. For duplicate names within a team, I have added a new prompt pre-populated with the first attempted name

This should help users tweak their initial trial to push it through.

2. I have added a conditional alert for the `uniqueness violation` error messages which points towards a slug duplication within a team.

The struggle I had here was how to efficiently compare the slug name with that of the new team. This approach seemed the lightest without adding a new `getFlows()` with the new team ID, which could maybe be added at the api level or in the frontend, but this seemed to grow the boundaries of this task. I was also thinking of a future where we use the Dialog components we have created for user lists for services, changing the way users interact with these actions.

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
@@ -180,7 +180,7 @@ export interface EditorStore extends Store.Store {
   makeUnique: (id: NodeId, parent?: NodeId) => void;
   moveFlow: (
     flowId: string,
-    teamName: string,
+    teamSlug: string,
     flowName: string,
   ) => Promise<any>;
   moveNode: (
@@ -490,8 +490,7 @@ export const editorStore: StateCreator<
     send(ops);
   },
 
-  moveFlow(flowId: string, teamName: string, flowName: string) {
-    const teamSlug = slugify(teamName);
+  moveFlow(flowId: string, teamSlug: string, flowName: string) {
     const valid = get().canUserEditTeam(teamSlug);
     if (!valid) {
       alert(
@@ -517,7 +516,7 @@ export const editorStore: StateCreator<
         const { data } = response;
         if (data.error.toLowerCase().includes("uniqueness violation")) {
           alert(
-            `Failed to move this flow. ${teamName} already has a flow with name '${flowName}'. Rename the flow and try again`,
+            `Failed to move this flow. ${teamSlug} already has a flow with name '${flowName}'. Rename the flow and try again`,
           );
         } else {
           alert(
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
@@ -492,11 +492,10 @@ export const editorStore: StateCreator<
 
   moveFlow(flowId: string, teamName: string, flowName: string) {
     const teamSlug = slugify(teamName);
-    console.log(teamSlug);
     const valid = get().canUserEditTeam(teamSlug);
     if (!valid) {
       alert(
-        `You do not have permission to move this service into ${teamSlug}, try again`,
+        `You do not have permission to move this flow into ${teamSlug}, try again`,
       );
       return Promise.resolve();
     }
@@ -518,11 +517,11 @@ export const editorStore: StateCreator<
         const { data } = response;
         if (data.error.toLowerCase().includes("uniqueness violation")) {
           alert(
-            `${teamName} already has a service with name '${flowName}'. Rename the service and try again`,
+            `Failed to move this flow. ${teamName} already has a flow with name '${flowName}'. Rename the flow and try again`,
           );
         } else {
           alert(
-            "Failed to move this service. Make sure you're entering a valid team name and try again",
+            "Failed to move this flow. Make sure you're entering a valid team name and try again",
           );
         }
       });
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
@@ -515,7 +515,7 @@ export const editorStore: StateCreator<
         const { data } = response;
         if (data.error.toLowerCase().includes("uniqueness violation")) {
           alert(
-            `${teamSlug} already have a service with name: '${flowName}'. Enter a different name and try again `,
+            `${teamSlug} already have a service with name: '${flowName}'. Rename the service and try again `,
           );
         } else {
           alert(
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
@@ -492,10 +492,11 @@ export const editorStore: StateCreator<
 
   moveFlow(flowId: string, teamName: string, flowName: string) {
     const teamSlug = slugify(teamName);
+    console.log(teamSlug);
     const valid = get().canUserEditTeam(teamSlug);
     if (!valid) {
       alert(
-        `You do not have permission to move this flow into ${teamSlug}, try again`,
+        `You do not have permission to move this service into ${teamSlug}, try again`,
       );
       return Promise.resolve();
     }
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/pages/Team.tsx
```diff
@@ -279,6 +275,21 @@ const GetStarted: React.FC<{ flows: FlowSummary[] }> = ({ flows }) => (
   </Box>
 );
 
+const checkForDuplicateFlowName = (name: string, flows: FlowSummary[]) => {
```
RODO94, line 278: There is part of me which questions the position of this function in the file and if it should be placed in a file outside of the main component, but considering the other work happening right now with this file for Filters and Search, it seems better to consider this after these have been added and polished

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -507,11 +511,18 @@ export const editorStore: StateCreator<
         },
       )
       .then((res) => alert(res?.data?.message))
-      .catch(() =>
-        alert(
-          "Failed to move this flow. Make sure you're entering a valid team name and try again",
-        ),
-      );
+      .catch(({ response }) => {
+        const { data } = response;
```
RODO94, line 515: Iterated on this a few times to get the cleanest code, I settled on this as it balanced explicit variable naming with concise lines of code in what is quite a long function

RODO94: Would appreciate feedback here, other iterations included a similar conditional check for when to display the `'Failed to move this flow.'` alert, but I am, as ever, treading lightly.

jessicamcinchak (commented):
Unique error messages are a definite improvement here! 

A few initial comments & bugs - please try moving a flow to the valid team name "Barking & Dagenham" on the pizza :bug:

editor.planx.uk/src/pages/Team.tsx
jessicamcinchak: Also more to your original question - I don't mind this helper function being added into this existing file right now, but I do prefer helper functions to be added to the bottom rather than _in between_ component definitions.

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -507,11 +513,18 @@ export const editorStore: StateCreator<
...
+        if (data.error.toLowerCase().includes("uniqueness violation")) {
+          alert(
+            `The ${teamName} team already has a service with name '${flowName}'. Rename the service and try again `,
```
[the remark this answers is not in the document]
RODO94: Just spoke to Jonny about it and he agrees!

editor.planx.uk/src/pages/Team.tsx
RODO94: This is great feedback, thanks! I'll rename it and add the return type

editor.planx.uk/src/pages/Team.tsx
```diff
@@ -279,6 +275,21 @@ const GetStarted: React.FC<{ flows: FlowSummary[] }> = ({ flows }) => (
...
+  const newFlowSlug = slugify(name);
+  const duplicateFlowName = flows?.find((flow) => flow.slug === newFlowSlug);
+
+  if (duplicateFlowName) {
+    const updatedName = prompt(
+      `A service already exists with the name '${name}', enter another name`,
+      name,
+    );
```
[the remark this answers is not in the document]
RODO94: Given it is a list of **Services** and a flow is only the supporting data structure, should we be committing to `service` over `flow`?

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -485,7 +490,8 @@ export const editorStore: StateCreator<
     send(ops);
   },
 
-  moveFlow(flowId: string, teamSlug: string) {
+  moveFlow(flowId: string, teamName: string, flowName: string) {
+    const teamSlug = slugify(teamName);
```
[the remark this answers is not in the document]
RODO94: Yeah I switched it so we could take in a `teamName` so it can be shown in the error messages, which I got feedback on from Jonny, I had it like you suggested, but he did ask that we show the `teamName` as its added. I did find it hard cause at one point you want to show the user that what they've entered isn't working, but then it obscures the fact we actually use the `slug` for matching it.

On the `Barking & Dagenham` one, this is interesting because the slug uses `and` but the name uses `&` - but when you `slugify('Barking & Dagenham')` you get `barking-dagenham` not `barking-and-dagenham`. Has this been manually added in the database?

I tried the same thing in staging and I get the same error, so I don't think it's a bug I've introduced here? And I'm not sure how I'd resolve it, other than updating how we handle `&` in the database slug

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
jessicamcinchak: Not at all asking you to remove the other changes, I agree it's ultimately a nice user experience, just trying to gauge how requirements versus nice-to-haves were interpreted here. It does seem like the additional prompt added more time, questions, and complexity, so it's important to be able to identify that when reflecting on sprint tasks !

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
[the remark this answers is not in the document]
jessicamcinchak: This error does _not_ exist on staging because staging prompts us for the slug ! I can successfully move a flow to B&G, which does not work on your pizza if I have to enter the team _name_ and it gets automatically slugified
Screenshot from 2025-02-05 13-02-16

From memory, team slugs do _not_ follow the automatic slugify convention because they need to be able to coform to/match council subdomains.

In my opinion, the prompt should still take a team _slug_, which editors can always find in the URL.

editor.planx.uk/src/pages/Team.tsx
```diff
@@ -233,14 +226,16 @@ const FlowItem: React.FC<FlowItemProps> = ({
...
+                  const newTeam = prompt(
+                    "Add the new team's slug. A slug is a way to represent a team name, for example 'Barking & Dagenham' would be 'barking-and-dagenham'. ",
+                  );
                   if (newTeam) {
                     if (slugify(newTeam) === teamSlug) {
                       alert(
                         `This flow already belongs to ${teamSlug}, skipping move`,
                       );
                     } else {
-                      handleMove(slugify(newTeam));
```
RODO94, line 243: As we chatted about, I removed the adding of teamName here and reverted it back to original state with `slugify(teamName)`, but with `flow.name` which seems harmless.

We should consider a future PR improving this error state though

editor.planx.uk/src/pages/Team.tsx
```diff
@@ -233,14 +226,16 @@ const FlowItem: React.FC<FlowItemProps> = ({
               {
                 label: "Move",
                 onClick: () => {
-                  const newTeam = prompt("New team");
+                  const newTeam = prompt(
+                    "Add the new team's slug. A slug is a way to represent a team name, for example 'Barking & Dagenham' would be 'barking-and-dagenham'. ",
```
RODO94, line 230: Added more specificity here and using B&D as an example as it's probably the biggest source of error in this flow.

jessicamcinchak: nit: I would say "**Enter** the new team's slug. A slug is the URL name for a team, for example..." (then users know where to look to find theres!)

editor.planx.uk/src/pages/Team.tsx
```diff
@@ -483,3 +472,21 @@ const Team: React.FC = () => {
...
+  if (duplicateFlowName) {
+    const updatedName = prompt(
+      `A service already exists with the name '${name}', enter another name`,
+      name,
+    );
+    if (!updatedName) return;
+    return getUniqueFlow(updatedName, flows);
+  }
+  return { slug: newFlowSlug, name: name };
+};
```
jessicamcinchak, line 492: This is much clearer :heavy_plus_sign:

jessicamcinchak (approved):
Thanks for changes here ! All looks good and pizza behaving as expected

RODO94:
@jessicamcinchak at the end of the day yesterday, Jonny said "destination" team sounds better than "new" team, since "new" might indicate they are making a new team. It seems a good switch to me and quite minor still, so I've implemented it

jessicamcinchak:
@RODO94 my priority was reviewing/approving _code_ here, Jonny's final wording suggestion is fine and there's no reason to wait to bring this to dev call in my opinion - let's please get this one merged so it's marked as "done" ahead of planning this afternoon !

PR #4245, opened 2025-02-03
feat: Setup Slack notification for failed Fargate deployment

DafyddLlyr:
Follow up to https://github.com/theopensystemslab/planx-new/pull/4096

When a deployment fails on AWS, this will trigger a Slack webhook to ping the `#planx-notifications-internal` channel which is where notifications are currently set up for failed CI deployments.

I believe that this only applies to Hasura currently which has circuit breakers and rollbacks in place - other failures should fail a health check and trigger notifications via CI/Pulumi or UptimeRobot. However I've aimed to write the code in an agnostic way so that we could apply it to other services in future if needed to.

Currently there's just a single webhook for both staging and production - if we find we need to / want to differentiate here I can set up one per-environment if required. Once I've been able to test this I should be able to see if the notification natively includes this information or if I actually need two webhooks.

Helpful guide this is essentially based on - https://www.youtube.com/watch?v=CszzQcPAqNM

### DIFF
--- a/infrastructure/application/utils/failureNotification.ts
+++ b/infrastructure/application/utils/failureNotification.ts
@@ -35,7 +35,7 @@ export const setupFailureNotificationForDeployments = (
       alarmActions: [topic.arn],
       comparisonOperator: "GreaterThanThreshold",
       evaluationPeriods: 1,
-      metricName: "ECSServiceFailedTaskCount",
+      metricName: "UnhealthyTaskCount",
       namespace: "AWS/ECS",
       period: 300,
       statistic: "Sum",
--- a/infrastructure/application/utils/failureNotification.ts
+++ b/infrastructure/application/utils/failureNotification.ts
@@ -1,0 +1,50 @@
+"use strict";
+
+import * as pulumi from "@pulumi/pulumi";
+import * as aws from "@pulumi/aws";
+import * as awsx from "@pulumi/awsx";
+
+/**
+ * Send a Slack notification when a Fargate service rolls back it's deployment
+ * Any initial Pulumi errors already handle this via notifications from GHA
+ * This will catch instances where the container has successfully deployed, but may be unhealthy and is caught by a circuit breaker
+ */
+export const setupFailureNotificationForDeployments = (
+  serviceName: string,
+  cluster: awsx.ecs.Cluster,
+  service: awsx.ecs.FargateService
+) => {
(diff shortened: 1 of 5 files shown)


freemvmt (approved):
Very nice!

I did think while I was working on the Hasura infra stuff that we could do with some AWS alarms, so this is a great model from which to work should we need more in future :)

infrastructure/application/utils/failureNotification.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+  );
+
+  new aws.cloudwatch.MetricAlarm(
+    `${serviceName}-deployment-alarm-notification`,
+    {
+      alarmActions: [topic.arn],
+      comparisonOperator: "GreaterThanThreshold",
+      evaluationPeriods: 1,
+      metricName: "ECSServiceFailedTaskCount",
```
freemvmt, line 38: Only thing I'm a bit unsure about is this metric? I can't see it referenced in AWS docs, e.g. [here](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/available-metrics.html) or [here](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Container-Insights-metrics-ECS.html). Of course, that absolutely doesn't mean that it doesn't exist 😅

freemvmt: [This article](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/deployment-alarm-failure.html) suggests using metrics on the ALB 🤔

freemvmt: After a bit of poking around I think `UnhealthyTaskCount` might be the right metric? Based on the fact that a rollback should occur only in the case that the new task cannot be booted to a healthy state, and is therefore labelled as unhealthy

DafyddLlyr: Thanks - struggled with this one as I couldn't find an appropriate metric via the console and came across this somewhere when looking for a solution.

I've now updated to `UnhealthyTaskCount` - seems more reasonable!

freemvmt: @DafyddLlyr awesome, sounds good! Let's test this and see how it goes. It may be we need to reduce the period over which the unhealthy task count needs to be sustained (since it may be killed before 5 mins are up).

PR #4248, opened 2025-02-03
feat: add ownership list schemas

augustlindemer:
Adds list schemas to support the collection of ownership information for multiple other owners (and/or people with interest in the land).


editor.planx.uk/src/@planx/components/List/schemas/InterestInLandLDC.ts
```diff
@@ -1,55 +1,36 @@
...
       },
     },
+    {
+      type: "text",
+      data: {
+        title: "What's their interest in the land?",
+        fn: "interest",
+        type: TextInputType.Short,
+      },
+    },
```
jessicamcinchak, line 29: Good to note from schema perspective that for LDCs this is free text input now and not a picklist/enum :+1:

editor.planx.uk/src/@planx/components/List/schemas/InterestInLandLDC.ts
```diff
@@ -63,6 +44,15 @@ export const Owners: Schema = {
...
     },
+    {
+      type: "text",
+      required: false,
+      data: {
+        title: "If they have not been notified, explain why not",
+        fn: "noNoticeReason",
+        type: TextInputType.Short,
+      },
+    },
```
jessicamcinchak, line 55: Great to see "optional" in use - works just as expected on pizza :raised_hands:

jessicamcinchak (approved):
Looks great !

editor.planx.uk/src/@planx/components/List/schemas/InterestInLandLDC.ts
augustlindemer: Importantly only in the context of the applicant not being owner/lessee/occupier. This is still captured separately as singular `applicant.ownership.interest` : `owner`/`lessee`/`occupier`/`other` value.

`interest` thus now subsumes the previously separate `interestDescription` (will no longer exist) that followed on from `other`.

PR #4250, opened 2025-02-04
chore: Remove `FeeBreakdownSection` from Editor modal

DafyddLlyr:
We've decided that - 
 - the fee breakdown will always display where possible
 - we don't need to explain or configure anything in the modal - this can live in Notion without issue. If we get more feedback here we can link to Notion (like we do with Calculate → MathJS).

In reality, almost all services will be going through the `osl/fee-calculator` flow so this should "just work"!


editor.planx.uk/src/@planx/components/Pay/Editor/Editor.tsx
```diff
@@ -101,9 +100,8 @@ const Component: React.FC<Props> = (props: Props) => {
               </InputRow>
             </ModalSectionContent>
           </ModalSection>
-          <GovPayMetadataSection/>
-          <InviteToPaySection/>
-          <FeeBreakdownSection/>
+          <GovPayMetadataSection />
+          <InviteToPaySection />
```
jessicamcinchak, line 104: Small unrelated nit: have often thought that these final existing sections should be swapped ! Eg invite to pay first because impacts public form, then more information because similarly public facing, then Gov Pay metadata last (or before internal notes) because behind-the-scenes/internal only . Recognise this is totally outside of scope here though, maybe disorienting to change for current users !

jessicamcinchak (approved):

editor.planx.uk/src/@planx/components/Pay/Editor/Editor.tsx
DafyddLlyr: Great shout! Done ✅

PR #4251, opened 2025-02-04
fix: Visited clone styling

ianjon3s:
## What does this PR do?

Quick one: fixes clone styling so that visited clones are still given visible border treatment.

Before (left) vs after (right):

image

**Test:**
https://4251.planx.pizza/testing/clone-styling


DafyddLlyr (approved):
Perfect!

PR #4254, opened 2025-02-05
feat[api]: create flows from templates

jessicamcinchak:
**Changes:**
- Removes redundant `createFlow` GraphQL mutations from the frontend store and instead calls a new API route `flows/create` to access existing API-side mutations already used by copy, move, etc.
- Adds another new API route `/flows/create-from-template/:templateId` which creates a new flow in the target team with a populated `templated_from` reference and data matching the source template (nodes _are not_ renamed like on copy)
  - This route _cannot_ be called from the frontend until we have a way of identifying & selecting "source templates" from the skeleton modal - but it's possible to call it direct from Swagger docs in the meantime by passing any valid `flowId` as the `templateId`

### DIFF
--- a/api.planx.uk/modules/flows/createFlow/controller.ts
+++ b/api.planx.uk/modules/flows/createFlow/controller.ts
@@ -1,0 +1,48 @@
+import { z } from "zod";
+import type { ValidatedRequestHandler } from "../../../shared/middleware/validate.js";
+import { ServerError } from "../../../errors/index.js";
+import { createFlow } from "../../../helpers.js";
+import type { Flow } from "../../../types.js";
+
+interface CreateFlowResponse {
+  message: string;
+  id: Flow["id"];
+  slug: Flow["slug"];
+}
+
+export const createFlowSchema = z.object({
+  body: z.object({
+    teamId: z.number(),
+    slug: z.string(),
+    name: z.string(),
+    data: z.any(), // FlowGraph
+  }),
+});
+
+export type CreateFlowController = ValidatedRequestHandler<
+  typeof createFlowSchema,
+  CreateFlowResponse
+>;
+
+export const createFlowController: CreateFlowController = async (
+  _req,
+  res,
+  next,
+) => {
+  try {
+    const { teamId, slug, name, data } = res.locals.parsedReq.body;
+
+    // createFlow automatically handles the associated operation and initial publish
+    const { id } = await createFlow(teamId, slug, name, data);
+
+    res.status(200).send({
+      message: `Successfully created flow ${slug}`,
+      id,
+      slug,
+    });
+  } catch (error) {
+    return next(
+      new ServerError({ message: `Failed to create flow. Error: ${error}` }),
+    );
+  }
+};
--- a/api.planx.uk/modules/flows/createFlow/controller.ts
+++ b/api.planx.uk/modules/flows/createFlow/controller.ts
@@ -1,7 +1,8 @@
+import type { FlowGraph } from "@opensystemslab/planx-core/types";
 import { z } from "zod";
-import type { ValidatedRequestHandler } from "../../../shared/middleware/validate.js";
 import { ServerError } from "../../../errors/index.js";
 import { createFlow } from "../../../helpers.js";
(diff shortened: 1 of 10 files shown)


api.planx.uk/modules/flows/createFlow/controller.ts
```diff
@@ -0,0 +1,48 @@
...
+  id: Flow["id"];
+  slug: Flow["slug"];
+}
+
+export const createFlowSchema = z.object({
+  body: z.object({
+    teamId: z.number(),
+    slug: z.string(),
+    name: z.string(),
+    data: z.any(), // FlowGraph
```
jessicamcinchak, line 18: Was a little stumped here and landed on `z.any()`, is there a way for `zod` to access an externally defined type instead or a generic "json" type? Repeating planx-core's `FlowGraph` type as a `z.object({})` directly felt like the wrong direction?

jessicamcinchak:
[[Implementation] Make sure live services are always up to date (aka flows copied from Templates should receive updates)](https://trello.com/c/huXP4VBO/2859-implementation-make-sure-live-services-are-always-up-to-date-aka-flows-copied-from-templates-should-receive-updates)

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -297,59 +301,42 @@ export const editorStore: StateCreator<
...
-        flow_id: id,
-        data: {
-          m: { ts: 1592485241858 },
-          v: 0,
-          seq: 1,
-          src: "143711878a0ab64c35c32c6055358f5e",
-          create: {
-            data: { [ROOT_NODE_KEY]: { edges: [] } },
-            type: "http://sharejs.org/types/JSONv0",
-          },
```
jessicamcinchak, line 345: It's worth highlighting that our frontend mutation which created the operation associated with a new flow was defining more `data` than our API equivalant (which has been actively in use by `/flows/:flowId/copy` endpoint). 

We haven't ever hit issues with this before for copied flows, so I have not updated the existing API mutation. Any objections to this? 

Compare here: https://github.com/theopensystemslab/planx-new/blob/main/api.planx.uk/helpers.ts#L129

api.planx.uk/modules/flows/createFlow/controller.ts
jessicamcinchak: Ah of course, there's actually no good reason that `data` needs to be passed from the frontend anyways here ! Have updated to simply define `const initialFlowData: FlowGraph = {...}` API-side which negates my zod issue :slightly_smiling_face:

api.planx.uk/modules/flows/createFlowFromTemplate/service.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const createFlowFromTemplate = async (
+  templateId: string,
+  teamId: number,
+): Promise<{ id: string; slug: string }> => {
+  // Fetch the source template data
+  const sourceTemplate = await getFlowData(templateId);
+
+  const newSlug = sourceTemplate.name + "-template";
+  const newName = sourceTemplate.name + " (template)";
```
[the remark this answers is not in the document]
jessicamcinchak: I'm not sure we'll always want to keep this pattern, but it seems like the right starting point because:
- It matches "copy" 
- It means that councils will be able to keep, for example, their _existing_ "LDC" flows running alongside new "LDC (template)" prototypes initially without hitting uniqueness violations

api.planx.uk/modules/flows/createFlowFromTemplate/service.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  // Create the new templated flow, including an associated operation and initial publish, and templated_from reference to the source templateId
+  const { id } = await createFlow(
+    teamId,
+    newSlug,
+    newName,
+    sourceTemplate.data,
+    undefined,
+    templateId,
+  );
```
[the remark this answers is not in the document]
jessicamcinchak: Very fair point, added a quick comment !

DafyddLlyr (approved):

PR #4255, opened 2025-02-05
chore: Clean property type column

DafyddLlyr:
Opened on behalf of @christinelangston

A few questions below for @zz-hh-aa 👇

### DIFF
--- a/hasura.planx.uk/migrations/1738757992023_run_sql_migration/up.sql
+++ b/hasura.planx.uk/migrations/1738757992023_run_sql_migration/up.sql
@@ -1,0 +1,62 @@
+DROP VIEW "public"."analytics_summary";
+
+CREATE OR REPLACE VIEW "public"."analytics_summary" AS 
+SELECT a.id AS analytics_id,
+    al.id AS analytics_log_id,
+    f.slug AS service_slug,
+    t.slug AS team_slug,
+    a.type AS analytics_type,
+    al.created_at AS analytics_log_created_at,
+    a.created_at AS analytics_created_at,
+    (a.user_agent -> 'os'::text) ->> 'name'::text AS operating_system,
+    (a.user_agent -> 'browser'::text) ->> 'name'::text AS browser,
+    (a.user_agent -> 'platform'::text) ->> 'type'::text AS platform,
+    a.referrer,
+    al.flow_direction,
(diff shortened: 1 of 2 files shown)


hasura.planx.uk/migrations/1738757992023_run_sql_migration/up.sql
```diff
@@ -0,0 +1,62 @@
...
+    al.has_clicked_help,
+    al.input_errors,
+    date_part('epoch'::text, al.next_log_created_at - al.created_at)::numeric(10,1) AS time_spent_on_node_seconds,
+    a.ended_at AS analytics_ended_at,
+    (date_part('epoch'::text, a.ended_at - a.created_at) / 60::double precision)::numeric(10,1) AS time_spent_on_analytics_session_minutes,
+    al.node_id,
+    al.allow_list_answers,
+    (al.allow_list_answers -> 'proposal.projectType'::text)::text AS proposal_project_type,
+    (al.allow_list_answers -> 'application.declaration.connection'::text)::text AS application_declaration_connection,
+	(al.allow_list_answers -> 'property.type' ->> 0) as property_type,
```
DafyddLlyr, line 40: This is the core change here - it's currently outputting an array, e.g. ["someValue"].

This will just output "someValue".

**Questions for @zz-hh-aa**
 - Will this break anything - are we relying on this or already extracting it in Metabase?
 - We're relying on property type being singular all the time - is this assumption correct?

zz-hh-aa: 1. No this won't break anything, none of our dashboards use this at the moment!
2. Property type is not always singular but it only looks like there's one array that has multiple values in it (and it shows up 111 times) 
image

christinelangston: Design Decision - We can confirm after speaking with Jonny that the property type should only ever have one type, so it is appropriate to take the 0th element. At this time, there are historic data logs (110 records) that have two types related to solar panels. This has been changed and no longer possible in the tool.

jessicamcinchak:
Reminder to selves we should still migrate historic allow-list property type values to sync up to latest changes :bookmark:

PR #4157, opened 2025-01-16
feat: Metabase dashboard service and controller (re-pushed)

zz-hh-aa:
Re-pushed version of #4150 hopefully sidestepping the Vultr phishing issue ([Slack thread here](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1737021735567669)). 

>What does this PR do?
Creates services getDashboard, copyDashboard, updateFilter and generatePublicLink
Adds necessary types and tests
Creates new route for analytics module
Exports the Metabase client as a singleton instance (this also happens in [8f12f20](https://github.com/theopensystemslab/planx-new/pull/4072/commits/8f12f205708fb03b802f8f9330fce54319a21f93) in https://github.com/theopensystemslab/planx-new/pull/4072 but there were other changes in the same commit so I've isolated the change here)
Updates Swagger yaml
Why?
Following on from https://github.com/theopensystemslab/planx-new/pull/4072, this should complete all the services we need to automate Metabase dashboard creation
This PR became a bit bigger than I'd anticipated... let me know if it'd be better to break it up!

PR #4260, opened 2025-02-05
feat: handle Metabase API <-> PlanX API camelCase to snake_case conversion

zz-hh-aa:
# What does this PR do
Requests and responses are handled _once_ at the point of entry / exit. Each function handles any relevant transformations, and `Metabase` specific types are added too. 

# Why
Metabase API uses snake_case, PlanX camel. Carries on from [conversation](https://github.com/theopensystemslab/planx-new/pull/4157#discussion_r1932074730) in #4157 about how best to handle case conversions. 

# Questions
I didn't end up using the modules-based logical namespacing approach, is that necessary? I ended up keeping the Metabase types in their respective directories.


api.planx.uk/modules/analytics/metabase/collection/createCollection.ts
```diff
@@ -1,9 +1,14 @@
-import type { MetabaseCollectionParams } from "./types.js";
+import type { CreateCollectionParams, MetabaseCreateCollectionParams } from "./types.js";
```
jamdelion, line 1: Unused type here :)

```suggestion
import type { CreateCollectionParams } from "./types.js";
```

api.planx.uk/modules/analytics/metabase/collection/collection.test.ts
```diff
@@ -84,12 +84,14 @@ describe("createTeamCollection", () => {
       name: testName,
       parentId: 100,
     });
+    console.log({collectionId})
```
jamdelion, line 87:
```suggestion
```

api.planx.uk/modules/analytics/metabase/collection/collection.test.ts
```diff
@@ -84,12 +84,14 @@ describe("createTeamCollection", () => {
...
 
     // Check the ID is returned correctly
     expect(collectionId).toBe(123);
 
     // Verify the collection details using the service function
     const collection = await getCollection(collectionId);
+    console.log({collection})
```
jamdelion, line 94:
```suggestion
```

api.planx.uk/modules/analytics/metabase/collection/createCollection.ts
```diff
@@ -1,9 +1,14 @@
...
 export async function createCollection(
-  params: MetabaseCollectionParams,
+  params: CreateCollectionParams,
 ): Promise<number> {
-  const response = await $metabase.post(`/api/collection/`, params);
+  const response = await $metabase.post(`/api/collection/`, { 
+    name: params.name,
+    parent_id: params.parentId, 
+    description: params.description
+  } as MetabaseCreateCollectionParams);
```
jamdelion, line 11: I think you generally shouldn't be using `as` like this. The `as` keyword is kind of only meant to be used when the typescript compiler can't figure out the type of the item - i.e. as an override, to force something into a particular type.

I think instead you should do something like:

```typescript
const metabaseCopyDashboardParams: MetabaseCopyDashboardParams = {
    name: params.name,
    description: params.description,
    collection_id: params.collectionId,
  }
  const response = await $metabase.post(
    `/api/dashboard/${params.templateId}/copy`,
    metabaseCopyDashboardParams,
  );
  ```
  
  So you label that you expect the params to comply with the type without forcing it. This way it will display the type errors if any problems.

File: api.planx.uk/modules/analytics/metabase/dashboard/copyDashboard.ts [hunk 5]
```diff
@@ -0,0 +1,16 @@
...
+export async function copyDashboard(
+  params: CopyDashboardParams,
+): Promise<GetDashboardResponse["id"]> {
+  const response = await $metabase.post(
+    `/api/dashboard/${params.templateId}/copy`,
+    {
+      name: params.name,
+      description: params.description,
+      collection_id: params.collectionId,
+    } as MetabaseCopyDashboardParams,
```
jamdelion, line 13: Same comment as above here regarding `as` keyword :)

api.planx.uk/modules/analytics/metabase/dashboard/dashboard.test.ts
```diff
@@ -1 +1,183 @@
...
+              name: filterName,
+              type: "number/=",
+              value: [42],
+            },
+          ],
+        });
+
+      // nock(BASE_URL!).put(`/api/dashboard/${dashboardId}`).reply(400, {
+      //   message: "Invalid parameter type. Expected number, got string.",
+      // });
```
jamdelion, line 154: Forgotten to delete perhaps? :)

api.planx.uk/modules/analytics/metabase/dashboard/service.ts
```diff
@@ -0,0 +1,42 @@
...
+      name: newName,
+      templateId,
+      description,
+      collectionId,
+    });
+
+    await updateFilter({
+      dashboardId: copiedDashboardId,
+      filter: filter,
+      value: value,
```
jamdelion, line 32: More succinct!

```suggestion
      filter,
      value,
```

jamdelion (commented):
Just a quick review from me so far which has thrown up a few type issues and console logs/tidying to do 🧼 🫧

api.planx.uk/modules/analytics/metabase/collection/collection.test.ts
zz-hh-aa: 😓 I am so failing at my new year's resolution to stop leaving `console.log`s in my code...! Thanks for keen eyes as ever.

api.planx.uk/modules/analytics/metabase/collection/getCollection.ts
```diff
@@ -8,5 +8,14 @@ export async function getCollection(
...
+
+  const transformedCollection = {
+    name: response.data.name,
+    id: response.data.id,
+    description: response.data.description,
+    slug: response.data.slug,
+    parentId: response.data.parentId,
+  };
+
+  return transformedCollection;
```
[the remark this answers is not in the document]
zz-hh-aa: Super interesting, had no idea here!

PR #4263, opened 2025-02-06
refactor: break up `SimpleStyledMenu` in Team page

RODO94:


editor.planx.uk/src/pages/Team/FlowMenu/MenuComponents/Copy.ts
jamdelion, line 1: All these MenuComponents don't seem to strictly be components..more like `getConfig` functions. Perhaps I would rename these all to things like `getRenameFlowButtonConfig` to be a little clearer as to what's happening when they are called inside the `SimpleMenu` props.

Likewise I'd rename the MenuComponents folder, maybe `FlowMenuActions`?

editor.planx.uk/src/pages/Team/FlowMenu/MenuComponents/Rename.ts
jamdelion, line 1: So much cleaner having these in their own files 😍

editor.planx.uk/src/pages/Team/FlowMenu/MenuComponents/Rename.ts
```diff
@@ -0,0 +1,54 @@
...
+                    slug: { _eq: $slug }
+                  }
+                  _set: { slug: $newSlug, name: $newName }
+                ) {
+                  affected_rows
+                }
+              }
+            `,
+            variables: {
+              teamId: teamId,
```
jamdelion, line 41: This can be collapsed into shorthand property name:

```suggestion
              teamId,
```

editor.planx.uk/src/pages/Team/FlowMenu/FlowMenu.tsx
```diff
@@ -0,0 +1,39 @@
...
+}));
+
+interface FlowMenuProps {
+  flow: FlowSummary;
+  flows: FlowSummary[];
+  setDeleting: (value: React.SetStateAction<boolean>) => void;
+  refreshFlows: () => void;
+}
+
+export const FlowMenu = ({
```
jamdelion, line 23: suggestion: Perhaps `FlowActionsMenu` to be more specific?

editor.planx.uk/src/pages/Team/FlowMenu/MenuComponents/Move.ts
```diff
@@ -0,0 +1,29 @@
+import { useStore } from "pages/FlowEditor/lib/store";
+import { FlowSummary } from "pages/FlowEditor/lib/store/editor";
+import { slugify } from "utils";
+
+export const moveFlow = (flow: FlowSummary, refreshFlows: () => void) => {
```
jamdelion, line 5: Renaming this function (as suggested in another comment) will make it less confusing to have two functions named `moveFlow` in the same file.

editor.planx.uk/src/pages/Team/FlowMenu/MenuComponents/Move.ts
```diff
@@ -0,0 +1,29 @@
...
+  const { moveFlow, teamSlug } = useStore.getState();
+
+  const handleMove = (newTeam: string, flowName: string) => {
+    moveFlow(flow.id, newTeam, flowName).then(() => {
```
jamdelion, line 9: Locally I get an error for having three args passed to `moveFlow` instead of two - the definition in the store looks like it only needs `flowId` and `teamSlug`.

editor.planx.uk/src/pages/Team/FlowMenu/MenuComponents/Rename.ts
```diff
@@ -0,0 +1,54 @@
+import { gql } from "@apollo/client";
+import { client } from "lib/graphql";
+import { useStore } from "pages/FlowEditor/lib/store";
+import { FlowSummary } from "pages/FlowEditor/lib/store/editor";
+
+import { getUniqueFlow } from "../utils";
+
+export const renameFlow = (
+  flow: FlowSummary,
+  flows: FlowSummary[],
```
jamdelion, line 10: Seems a bit weird to me to pass in `flow` and all `flows`, even though they are both used. 🤔  Perhaps you could make the function params a bit more specific - e.g. in this function only `flow.name` and `flow.slug` are used so you could use bring those in maybe? Just a feeling, feel free to ignore!

editor.planx.uk/src/pages/Team/FlowMenu/styles.ts
jamdelion, line 1: This is empty, I guess you were planning to populate this with some of the styled components? There is only one in the `FlowMenu` file, so maybe you could move this up a level and extract all of the ones that are currently in the `Team` file as well.

jamdelion (commented):
Love to see this! `Team.tsx` is one of my most hated files that I've been itching to refactor 😈 

This is a good start, I've left some comments about what you've done but I also think this could go much further (either here or in another PR) by also extracting `FlowItem`, `Confirm` (and rename to `ConfirmDelete`!), `GetStarted`, `AddFlowButton` etc into their own files, grouping the type definitions, styles and utils into their own files as well as you've started here.

RODO94:
@jamdelion Thanks for the comments! I'm going to close this and reopen it after Filters. There's so much movement in the Team.tsx file over the last week that it would be better to start afresh

PR #4273, opened 2025-02-06
refactor: Simplify view when no services are present

ianjon3s:
## What does this PR do?

- Removes search, filters and sort when no services are present

Before changes:
image

After changes:
image

**Preview:**
https://4273.planx.pizza/a-new-team

### DIFF
--- a/editor.planx.uk/src/pages/Team.tsx
+++ b/editor.planx.uk/src/pages/Team.tsx
@@ -266,7 +266,7 @@ const Team: React.FC = () => {
             </Typography>
             {showAddFlowButton && <AddFlowButton flows={flows} />}
           </Box>
-          {teamHasFlows && hasFeatureFlag("SORT_FLOWS") && flows && (
+          {teamHasFlows && hasFeatureFlag("SORT_FLOWS") && (
             <SearchBox<FlowSummary>
               records={flows}
               setRecords={setSearchedFlows}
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/pages/Team.tsx
```diff
@@ -266,82 +266,86 @@ const Team: React.FC = () => {
             </Typography>
             {showAddFlowButton && <AddFlowButton flows={flows} />}
           </Box>
-          {hasFeatureFlag("SORT_FLOWS") && flows && (
+          {teamHasFlows && hasFeatureFlag("SORT_FLOWS") && flows && (
```
DafyddLlyr, line 269: It looks like the additional check for `flows` here is redundant as `teamHasFlows` is already checking this. Is it needed for type-narrowing, or can it be removed?

DafyddLlyr (approved):
I can't test this as there's no Pizza (I'll sort this out on GHA actions now) but all looks good.

On a similar note, I think we need a card or something to indicate there's no results for a search/filter operation.

editor.planx.uk/src/pages/Team.tsx
ianjon3s: Good spot, seems to be redundant, I've removed.

ianjon3s:
@DafyddLlyr 

> I think we also need a card or something to indicate there's no results for a search/filter operation.

When full functionality is built the "no results" output will look like this (you'll need to apply filters to see no results in the original demo branch): https://4120.planx.pizza/a-test-team?serviceType=submission&applicationType=statutory

DafyddLlyr:
Sorted out CI and end to end tests @ianjon3s 👍

ianjon3s:
Thanks @DafyddLlyr 👍

PR #4274, opened 2025-02-06
fix: Transparent focused checkboxes

DafyddLlyr:
Small fix for UI issue noticed on the filters PRs that's not yet been addressed.

When a checkbox is focused on the filters, it picks up a grey background - this resolves the issue.


jamdelion (approved):
Haven't been able to check with no pizza but believable that it works from the code!

ianjon3s:
Thanks for spotting and addressing this @DafyddLlyr !

PR #4279, opened 2025-02-07
feat: Embed using native Notion "Share > Embed this page" option

DafyddLlyr:
image

Turns out we can do this and save the $2 monthly fee - still need to check ours Cloudfront setup allow this without changes to CSP headers.


ianjon3s:
Wow, good find @DafyddLlyr , this simplifies things a lot! Either this is a very new feature or I wasn't looking hard enough.

ianjon3s (approved):

PR #4280, opened 2025-02-07
feat(templates): Add `flows.is_template` column

DafyddLlyr:
## What does this PR do?
 - Create a new `flows.is_template` column
 - Readable by all, but can only be set by `platformAdmins`
 - Updates `StartFromTemplateButton` component to display actual list of templates, wires up basic formik
 - Just logs out a value, doesn't interact with the API

Having this additional column will massively simplify some of the permissions in https://github.com/theopensystemslab/planx-new/pull/4234 which are trying to check the existance of a relationship, instead of a simple value.

### DIFF
--- a/editor.planx.uk/src/pages/Team/StartFromTemplateButton.tsx
+++ b/editor.planx.uk/src/pages/Team/StartFromTemplateButton.tsx
@@ -1,0 +1,96 @@
+import { gql, useQuery } from "@apollo/client";
+import Box from "@mui/material/Box";
+import Button from "@mui/material/Button";
+import Dialog from "@mui/material/Dialog";
+import DialogActions from "@mui/material/DialogActions";
+import DialogContent from "@mui/material/DialogContent";
+import DialogTitle from "@mui/material/DialogTitle";
+import MenuItem from "@mui/material/MenuItem";
+import { useFormik } from "formik";
+import React, { useState } from "react";
+import { AddButton } from "ui/editor/AddButton";
+import SelectInput from "ui/editor/SelectInput/SelectInput";
+import InputLabel from "ui/public/InputLabel";
+
+export interface TemplateOverview {
+  name: string;
+  slug: string;
+  id: string;
+}
+
+export const StartFromTemplateButton: React.FC<{
+  templates: TemplateOverview[];
+}> = ({ templates }) => {
+  const [dialogOpen, setDialogOpen] = useState<boolean>(false);
+
+  const formik = useFormik<{ templateId: string }>({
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -404,8 +333,23 @@ const Team: React.FC = () => {
...
+        id
+        slug
+        name
+      }
+    }
+  `);
+
   const teamHasFlows = flows && Boolean(flows.length);
   const showAddFlowButton = teamHasFlows && canUserEditTeam(slug);
+  const showAddTemplateButton =
```
DafyddLlyr, line 348: This is a bit convoluted but will get simplified when there are always templates and the feature flag is removed.

editor.planx.uk/src/pages/Team/StartFromTemplateButton.tsx
```diff
@@ -0,0 +1,96 @@
...
+export const StartFromTemplateButton: React.FC<{
+  templates: TemplateOverview[];
+}> = ({ templates }) => {
+  const [dialogOpen, setDialogOpen] = useState<boolean>(false);
+
+  const formik = useFormik<{ templateId: string }>({
+    initialValues: { templateId: templates[0].id },
+    onSubmit: ({ templateId }) => {
+      console.log(
+        `POST to /flows/create-from-template/${templateId}, then redirect`,
```
jessicamcinchak, line 30: :+1: Very happy to wire this up to my previous PR in a follow up

editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -404,8 +333,23 @@ const Team: React.FC = () => {
     fetchFlows();
   }, [fetchFlows]);
 
+  const { data: templateData } = useQuery<{ flows: TemplateOverview[] }>(gql`
```
jessicamcinchak, line 336: nit: maybe just `templates` or `sourceTemplates` ? (data feels a bit off!)

jessicamcinchak (approved):
Perfect ! Sorry didn't push harder to clarify this earlier, but agree it'll totally simplify the delete constraint (and lots of other pieces) !

DafyddLlyr:
> Perfect ! Sorry didn't push harder to clarify this earlier, but agree it'll totally simplify the delete constraint (and lots of other pieces) !

Honestly no need to apologise! We had previously talked about it, and written it down on Trello - I was clearly just having a slow morning when we talked about it 😅 ☕

PR #4283, opened 2025-02-09
feat: allow editors to summarise operations in "History" timeline

jessicamcinchak:
A small indepedent feature that we need to support _both_ templates and publishing tickets in this SOW (similar to flows being published & offline by default) is the ability for editors to insert "comments" into the current "History" timeline. 

The "History" timeline currently reads from `operations` and displays a human translation of the JSON OT. Operations will remain the only timeline items that can be "restored", but the "comments" will allow editors to insert notes which essentially act like bookends on the timeline between groups of operations to contextualise the bigger picture (eg "Implemented new CIL Form 1") _between_ publishes (which may still include 1,000s of individual node operations).

These will enable us to show more friendly change logs on publish as well as help templated flow users understand how the source template content has been changing. 

It's in our interest to introduce this feature early, _not_ under the "TEMPLATES" feature flag, so that editors can actually start using it in the meantime even though we are not adjusting the "Check for changes to publish" dialog just yet.

**Key changes**: 
- [Hasura] Creates new table `flow_comments` (including introspection tests)
- [Hasura] Creates new view `flow_history` which unions operations, comments and publishes from last 6 months into a single event timeline that we can subscribe to
- [Editor] Breaks out `editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory` into separate component files: `index.ts`, `Timeline`, `TimelineItems` and `AddCommentDialog`
- [Editor] Adds two new store methods `addFlowCmment` and `deleteFlowComment` which are straight-forward, permission-scoped GraphQL mutations against the `flow_comments` table

(Apologies that this single PR became _both_ database & frontend changes rather than split up, but hopefully clear to follow!)


jessicamcinchak:
[Add ability to insert comment into "History" timeline that can be pulled through in publish changelogs](https://trello.com/c/KjTjN9wq/3228-add-ability-to-insert-comment-into-history-timeline-that-can-be-pulled-through-in-publish-changelogs)

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
```diff
@@ -0,0 +1,171 @@
...
+    operationsToUndo.map((op) => operationsData.unshift(op.data));
+    const flattenedOperationsData: OT.Op[] = operationsData?.flat(1);
+
+    // Then undo all operations by applying a single new inverse operation on top
+    undoOperation(flattenedOperationsData);
+
+    // Also delete each comment in the undo scope
+    commentsToDelete.forEach((comment) => {
+      deleteFlowComment(comment.id);
+    });
```
jessicamcinchak, line 57: `handleUndo` still treats "operations" exactly as it did before (just some slight renaming here to clarify we're parsing many event types), but now there's an additional step of automatically deleting any comments in the undo scope.

Editors will only see the action to "add comments", they cannot manually edit or delete comments for this iteration.

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+          type
+          createdAt: created_at
+          actorId: actor_id
+          firstName: first_name
+          lastName: last_name
+          data(path: "op")
+          comment
+        }
+      }
+    `,
```
jessicamcinchak, line 68: I knew that GraphQL `subscription` only worked against a single database entity at a time (previously `operations` table), but I didn't know until this piece of work that we can do subscriptions against a _view_ ! 

It's a bit of a strange mental model for me, but works a charm for this scenario where the alternative would be three separate subscriptions to the `operations`, `flow_comments` and `published_flows` tables then having to concat & sort responses on the frontend !

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  return (
+    <Box>
+      {user?.id && canUserEditTeam(teamSlug) && <AddCommentDialog flowId={flowId} actorId={user.id} />}
+      {data?.history && <EditHistoryTimeline events={data.history} />}
+      {data?.history.length === 30 && (
+        <>
+          <Divider />
+          <Typography variant="body2" mt={2} color="GrayText">
+            {`History shows the last 50 edits made to this service within the last six months. If you have questions about restoring to an earlier point in time, please contact a developer.`}
```
jessicamcinchak, line 103: Note that previously we showed the last 15 operations with no time limit. Now we'll show 50 _or_ the last 6 months, whichever is shorter.

There's more incentive to show a longer timeline now to capture "commented" and "published" events alongside the "operations", and the 6 months helps keep our `flow_history` query smaller and more performative.

This piece is particularly hard to realistically test until on production though (because of limited data sync to other envs), so have agreed with August that we'll keep an eye on performance and tweak-forward these params if any lag in live environments.

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
```diff
@@ -0,0 +1,171 @@
...
+    const operationsData: Array<OT.Op[]> = [];
+    operationsToUndo.map((op) => operationsData.unshift(op.data));
+    const flattenedOperationsData: OT.Op[] = operationsData?.flat(1);
+
+    // Then undo all operations by applying a single new inverse operation on top
+    undoOperation(flattenedOperationsData);
+
+    // Also delete each comment in the undo scope
+    commentsToDelete.forEach((comment) => {
+      deleteFlowComment(comment.id);
```
DafyddLlyr, line 56: This might well be correct and we don't need to `await` this, but just flagging it's an `async` operation within a `forEach` so this won't get handled consistently.

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
```diff
@@ -0,0 +1,171 @@
...
+        },
+      }}
+    >
+      {events?.map((op: HistoryItem, i: number) => (
+        <TimelineItem key={`${op.type}-${op.id}`}>
+          <TimelineSeparator>
+            <TimelineDot
+              sx={{
+                bgcolor: (theme) =>
+                  inUndoScope(i) && isUndoType(op.type)
```
DafyddLlyr, line 91: nit: We're relying on this combination in line a few places - maybe a sign that `TimelineItem` should be a component with this as a single calculated value.

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/index.tsx
DafyddLlyr: This is great to confirm! Much simpler than the alternative and allows us to combine and then sort a number of models here which is exactly what we're after.

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  // Handle missing operations (e.g. non-production data)
+  if (!loading && !data?.history) return null;
+
+  return (
+    <Box>
+      {user?.id && canUserEditTeam(teamSlug) && <AddCommentDialog flowId={flowId} actorId={user.id} />}
+      {data?.history && <EditHistoryTimeline events={data.history} />}
+      {data?.history.length === 30 && (
```
DafyddLlyr, line 99: Typo? Should this be `50`?

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/AddCommentDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+                onChange={formik.handleChange}
+                bordered
+                multiline={true}
+                rows={3}
+              />
+            </InputLabel>
+          </Box>
+        </DialogContent>
+        <DialogActions sx={{ paddingX: 2 }}>
+          <Button disableRipple onClick={() => setDialogOpen(false)}>BACK</Button>
```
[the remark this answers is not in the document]
DafyddLlyr: This persists as I'd expect, sort of like GitHub comments - my draft copy is there and persists until I submit or remove it. It makes sense that this is a field somebody may come and go from, and exiting is easy (I don't want to click away and lose my work).

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/AddCommentDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+        maxWidth="sm"
+      >
+        <DialogTitle variant="h3" component="h1">
+          {`Add a comment`}
+        </DialogTitle>
+        <DialogContent>
+          {`History reflects the individual edits to your flow. Insert a custom comment into the timeline to add context. This will help other editors understand what's changed on next publish.`}
+          <Box mt={2} component="form" onSubmit={formik.handleSubmit}>
+            <InputLabel label="Comment">
+              <Input
```
DafyddLlyr, line 52: As this is a single input, we might want to autofocus this. Currently I have to click "add comment", then click on the input, to start typing.

DafyddLlyr (approved):
Looking great! Working as expected 👌 

PR nice and clear and super helpful description thank you.

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/AddCommentDialog.tsx
jessicamcinchak: Good spot, `autofocus` prop now added !

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/index.tsx
jessicamcinchak: Fixed!

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
jessicamcinchak: Async/await now added, my oversight - thanks for spotting!

PR #4287, opened 2025-02-10
feat: add components ShowingServices and FlowTag to Filters

RODO94:
Added an extra component to handle the logic of what header to show for the `Showing X services` message. 

I have also updated the `FlowTag` component in `FlowCard` to now show `statutory` application type tag

This should be merged after we move the contents of `Team.tsx` to `index.tsx` as specified here: https://github.com/theopensystemslab/planx-new/pull/4286


DafyddLlyr:
[Show a breakdown of planning fee + service charge to users](https://trello.com/c/B3XJVH3l/3067-show-a-breakdown-of-planning-fee-service-charge-to-users)

editor.planx.uk/src/pages/Team/components/ShowingServicesHeader.tsx
```diff
@@ -0,0 +1,36 @@
...
+export const ShowingServicesHeader = ({
+  matchedFlows,
+}: {
+  matchedFlows: FlowSummary[] | null;
+}) => {
+  const [numberOfFlows, setNumberOfFlows] = useState(matchedFlows?.length || 0);
+
+  useEffect(() => {
+    setNumberOfFlows(matchedFlows?.length || 0);
+  }, [matchedFlows]);
```
DafyddLlyr, line 15: Is this internal state actually required - can we just pass a count into the component?

```ts
<ShowingServicesHeader flowCount={matchingFlows.length} />
```

editor.planx.uk/src/pages/Team/components/ShowingServicesHeader.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  const showingServicesMessage =
+    numberOfFlows !== 1
+      ? `Showing ${numberOfFlows} services`
+      : `Showing ${numberOfFlows} service`;
```
DafyddLlyr, line 20: We handle 0 services at a higher level (checking `teamHasFlows`).

It would be more consistent (and simpler?) to keep this logic grouped together. If 0 flows, we can return `null`.

editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -155,52 +207,80 @@ const Team: React.FC = () => {
...
+              sx={{
+                display: "flex",
+                justifyContent: "space-between",
+                alignItems: "center",
+                gap: 2,
+              }}
+            >
+              {teamHasFlows && (
+                <ShowingServicesHeader matchedFlows={matchingFlows} />
+              )}
```
DafyddLlyr, line 250: See previous comment about splitting up the logic here.

DafyddLlyr (approved):
One very minor suggestion, plus a question!

editor.planx.uk/src/pages/Team/components/ShowingServicesHeader.tsx
RODO94: `teamHasFlows` is different to `matchingFlows.length === 0` though since one relates to if a team has flows and another shows if there's search / filter results?

PR #4289, opened 2025-02-10
feat: handle Metabase API <-> PlanX API camelCase to snake_case conversion (v2)

zz-hh-aa:
Replaces #4260 because my rebase got way too messy!

# What does this PR do
Requests and responses are handled once at the point of entry / exit. Each function handles any relevant transformations, and Metabase specific types are added too.

# Why
Metabase API uses snake_case, PlanX camel. Carries on from https://github.com/theopensystemslab/planx-new/pull/4157#discussion_r1932074730 in https://github.com/theopensystemslab/planx-new/pull/4157 about how best to handle case conversions.


jamdelion (approved):
Looks fine to me! I feel like I've slightly lost the context of what changes you're making so perhaps would recommend a third pair of eyes just in case!

PR #4120, opened 2025-01-08
wip: Team page filters and cards

ianjon3s:
Demo PR for collective work on team page & filters.

PR #4291, opened 2025-02-10
feat: Clear filters and search button

RODO94:
The final major feature for Filters and Search for the Team page is a clear all button, which clears search term and any collected filters resetting the results back to normal.

The feature should allow a user to filter the list, add a search term and in one click, clear both to reset the list. It should also work independently (if the list is only filtered, or if it is only searched).

This meets feature parity with our demo branch: https://github.com/theopensystemslab/planx-new/pull/4120

>[!IMPORTANT] 
> To test the pizza you need to toggle the feature flag "SORT_FLOWS"

### DIFF
--- a/editor.planx.uk/src/pages/Team/index.tsx
+++ b/editor.planx.uk/src/pages/Team/index.tsx
@@ -85,13 +85,16 @@ const Team: React.FC = () => {
   const [matchingFlows, setMatchingflows] = useState<FlowSummary[] | null>(
     null,
   );
-  const [clearTrigger, setClearTrigger] = useState<number>(0);
+  const [shouldClearFilters, setShouldClearFilters] = useState<boolean>(false);
 
   useEffect(() => {
     const diffFlows =
       searchedFlows?.filter((flow) => filteredFlows?.includes(flow)) || null;
     setMatchingflows(diffFlows);
-  }, [searchedFlows, filteredFlows]);
+    if (shouldClearFilters) {
+      setShouldClearFilters(false);
+    }
+  }, [searchedFlows, filteredFlows, shouldClearFilters]);
 
   const sortOptions: SortableFields<FlowSummary>[] = [
     {
@@ -214,7 +217,7 @@ const Team: React.FC = () => {
               records={flows}
               setRecords={setSearchedFlows}
               searchKey={["name", "slug"]}
-              clearSearch={clearTrigger}
+              clearSearch={shouldClearFilters}
             />
           )}
         </Box>
@@ -237,7 +240,7 @@ const Team: React.FC = () => {
                   records={flows}
                   setFilteredRecords={setFilteredFlows}
                   filterOptions={filterOptions}
-                  clearFilters={clearTrigger}
+                  clearFilters={shouldClearFilters}
                 />
               )}
             </Box>
@@ -262,7 +265,7 @@ const Team: React.FC = () => {
                     matchedFlowsCount={matchingFlows?.length || 0}
                   />
                   <Button
-                    onClick={() => setClearTrigger((prevId) => prevId + 1)}
+                    onClick={() => setShouldClearFilters(true)}
                     variant="link"
                   >
                     Clear filters
--- a/editor.planx.uk/src/pages/Team/index.tsx
+++ b/editor.planx.uk/src/pages/Team/index.tsx
@@ -258,24 +258,11 @@ const Team: React.FC = () => {
                 gap: 2,
               }}
             >
-              <Box
-                sx={{
-                  display: "flex",
-                  justifyContent: "flex-start",
-                  alignItems: "center",
-                  gap: 2,
-                }}
-              >
-                <Typography variant="h3" component="h2">
-                  Showing X services
-                </Typography>
-                <Button
-                  onClick={() => setTriggerClearFiltersAndSearch(true)}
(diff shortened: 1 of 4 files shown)


editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -246,9 +259,24 @@ const Team: React.FC = () => {
...
+                  }}
+                >
+                  <ShowingServicesHeader
+                    matchedFlowsCount={matchingFlows?.length || 0}
+                  />
+                  <Button
+                    onClick={() => setTriggerClearFiltersAndSearch(true)}
+                    variant="link"
+                  >
+                    Clear filters
```
RODO94, line 277: Tried also moving this inside the `ShowingServicesHeader` component but it didn't feel linked, and I felt it hid the clear filters feature. This current form looks the easiest to follow

editor.planx.uk/src/ui/editor/Filter/Filter.tsx
```diff
@@ -143,9 +146,16 @@ export const Filters = <T extends object>({
 
   useEffect(() => {
     if (values.filters) {
-      handleSubmit();
```
RODO94, line 146: Wasn't sure on the main difference between `handleSubmit` and `submitForm` but I chose the same as I used for `SearchBox` to maintain consistency. From what I could read from Formik docs, `handleSubmit` is used for passing into a component rather than triggering in a function

editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -84,12 +85,22 @@ const Team: React.FC = () => {
...
+  const [triggerClearFiltersAndSearch, setTriggerClearFiltersAndSearch] =
+    useState<boolean>(false);
 
   useEffect(() => {
     const diffFlows =
       searchedFlows?.filter((flow) => filteredFlows?.includes(flow)) || null;
     setMatchingflows(diffFlows);
-  }, [searchedFlows, filteredFlows]);
+    if (diffFlows?.length === flows?.length && triggerClearFiltersAndSearch) {
+      setTriggerClearFiltersAndSearch(false);
```
RODO94, line 96: if `triggerClearFiltersAndSearch` is true, and nothing has been filtered, we can reset the trigger, because it means it has been cleared.

editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -84,6 +85,7 @@ const Team: React.FC = () => {
   const [matchingFlows, setMatchingflows] = useState<FlowSummary[] | null>(
     null,
   );
+  const [clearTrigger, setClearTrigger] = useState<number>(0);
```
RODO94, line 88: Using a number to trigger the clear felt the most straight forward in terms of code, and reading it through. I did iterate with a boolean, but the need to toggle it added too much complexity.

jamdelion: This seems weird to me, I had a quick go with a boolean and it worked but with a delay - was this one of the issues you had? Happy to chat through it :)

jamdelion (commented):
Mostly looks good and works well for me! However I think the number used for the state is a bit unexpected - more in comment below!

editor.planx.uk/src/ui/editor/Filter/Filter.tsx
```diff
@@ -44,6 +44,8 @@ interface FiltersProps<T> {
   setFilteredRecords: React.Dispatch<React.SetStateAction<T[] | null>>;
   /** An array of objects to define how to filter the records - the FilterOptions type has more information */
   filterOptions: FilterOptions<T>[];
+  /** Optional prop for clearing filters from a parent component */
```
DafyddLlyr, line 47: nit: Redundant? This seems to be covered by the property name and type already.

DafyddLlyr (changes requested):
Not tested as I can't access the pizza - likely the data sync issue resolved here.

One small point on the boolean/number issue.

editor.planx.uk/src/pages/Team/index.tsx
RODO94: Yeah, I guess the nuance is in that it only needs to be falsey once, and then it can remain truthy after that. It's this that makes it feel unnecessary to use a boolean value since toggling it means that one value is always truthy and one is always falsey, but a number doesn't have this, it is falsey once then truthy then after.

The fact it doesn't need to toggle between true and false, the variable only needs to trigger an Effect and be falsey on first render, makes the number make more sense to me.

I'll implement the boolean version, you probably need to be deeper in it to understand it with the number, so it's better for future to have it as a boolean.

Thanks for the feedback!

DafyddLlyr (approved):
Changes look good - a few small UI issues I noticed just now. 

Happy for these to be fixes forward - they're unrelated to the changes here I believe.

**Jump in number of services**

https://github.com/user-attachments/assets/7ae1c1f7-f209-4e12-91ad-7eb85bf89748

**Very wide labels taking up full row width, not just text**

https://github.com/user-attachments/assets/4697c648-cc51-49cc-923e-4fb34e9e971d

Also it looks like either my fix from https://github.com/theopensystemslab/planx-new/pull/4274 didn't work, or this isn't on the same version of `main` - could you check that out please?

PR #4293, opened 2025-02-10
fix: get name based on user role in `applicant_research_opt_in_view`

jessicamcinchak:
See request here: https://opensystemslab.slack.com/archives/C01E3AC0C03/p1739179324695849

**Key changes:**
- Renames `id` to `planx_application_reference`
- Adds `site_address`
- Adds `user_role`
- Updates `first_name` and `last_name` to reference correct variables based on `user_role`

### DIFF
--- a/hasura.planx.uk/migrations/1739222061613_update_view_applicant_research_opt_in/up.sql
+++ b/hasura.planx.uk/migrations/1739222061613_update_view_applicant_research_opt_in/up.sql
@@ -1,0 +1,34 @@
+DROP VIEW IF EXISTS "public"."applicant_research_opt_in_view";
+CREATE OR REPLACE VIEW "public"."applicant_research_opt_in_view" AS 
+WITH data AS (
+    SELECT 
+        ls.id as planx_reference_number, 
+        ls.email, 
+        ls.created_at AS application_started_at, 
+        ls.submitted_at,
+        COALESCE(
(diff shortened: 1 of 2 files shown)


hasura.planx.uk/migrations/1739222061613_update_view_applicant_research_opt_in/up.sql
```diff
@@ -0,0 +1,34 @@
...
+WITH data AS (
+    SELECT 
+        ls.id as planx_reference_number, 
+        ls.email, 
+        ls.created_at AS application_started_at, 
+        ls.submitted_at,
+        COALESCE(
+            ls.data->'passport'->'data'->'_address'->>'single_line_address',
+            ls.data->'passport'->'data'->'_address'->>'title'
+        ) as site_address,
```
jessicamcinchak, line 12: This `COALESCE` passport field logic is a good one to know about and commonly applies everywhere we display a site address (email templates, resume application, invite to pay etc - see methods like `getSessionDetails`, `buildContentFromSessions` in the API as examples using `||`).

If you look at the `MinimumSiteAddress` interface in the `FindProperty` component, you'll see that _every_ address will have a `title`, but this logic favors the `single_line_address` in cases of OS addresses because it's more detailed (eg includes town & postcode) and fallsback to `title` in cases where a new address is being proposed.

DafyddLlyr (approved):

PR #4295, opened 2025-02-11
fix(e2e): Fix async and logical issues with demoUser tests

DafyddLlyr:
Regression tests passing here - https://github.com/theopensystemslab/planx-new/actions/runs/13262726107 ✅ 

This resolves test failures I was hitting here - https://github.com/theopensystemslab/planx-new/pull/4234

### DIFF
--- a/e2e/tests/api-driven/src/demo-workspace/steps/verification_steps.ts
+++ b/e2e/tests/api-driven/src/demo-workspace/steps/verification_steps.ts
@@ -83,18 +83,20 @@ Then<CustomWorld>("I should be able to see a flow", async function (this) {
   );
 });
 
-Then<CustomWorld>(
-  "I should be able to {string} the flow",
-  async function (this, action) {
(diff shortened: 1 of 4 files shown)


e2e/tests/api-driven/src/demo-workspace/steps/verification_steps.ts
```diff
@@ -83,18 +83,20 @@ Then<CustomWorld>("I should be able to see a flow", async function (this) {
   );
 });
 
-Then<CustomWorld>(
-  "I should be able to {string} the flow",
```
DafyddLlyr, line 87: Two issues I had to fix here - 
 - Async/await wasn't being handled correctly (thus allowing `undefined` to pass) - `hasSucceeded` is always truthy.
 - `updateFlow` changes the flow slug, and `deleteFlow` tries to get it by it's _original_ value - it will always fail and not try to delete anything. To resolve this I'm now tracking the ID.

jamdelion (approved):
Well done for getting to the bottom of this!

PR #4297, opened 2025-02-11
fix: initial sort stutter when loading Team page

RODO94:


editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -114,6 +108,26 @@ const Team: React.FC = () => {
     },
   ];
 
+  useEffect(() => {
+    const diffFlows =
+      searchedFlows?.filter((flow) => filteredFlows?.includes(flow)) || null;
+
+    if (matchingFlows === null) {
```
RODO94, line 115: On initial load, we need to sort flows from the search params, but after this, the `<SortControl/>` component should handle this

editor.planx.uk/src/ui/editor/Filter/Filter.tsx
```diff
@@ -140,6 +140,7 @@ export const Filters = <T extends object>({
 
     if (!values.filters) {
       parseStateFromURL();
+      setFilteredRecords(records);
```
RODO94, line 143: We require this to trigger the Effect to `setMatchingRecords` in the Team.tsx

editor.planx.uk/src/ui/editor/SortControl/SortControl.tsx
```diff
@@ -47,13 +48,17 @@ export const SortControl = <T extends object>({
   setRecords: React.Dispatch<React.SetStateAction<T[] | null>>;
   sortOptions: SortableFields<T>[];
 }) => {
-  const [selectedSort, setSelectedSort] = useState<SortableFields<T>>(
-    sortOptions[0],
-  );
-  const [sortDirection, setSortDirection] = useState<SortDirection>("desc");
+  const route = useCurrentRoute();
```
RODO94, line 51: We shouldn't rely on `sortOptions[0]` but rather pull in what is in the search params at the start

editor.planx.uk/src/ui/editor/SortControl/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+): {
+  sortObject: SortableFields<T>;
+  sortDirection: "asc" | "desc";
+} => {
+  const { sort, sortDirection: sortDirectionParam } = route.url.query;
+  const sortObject = sortOptions.find(
+    (option) => slugify(option.displayName) === sort,
+  );
+  const sortDirection =
+    sortDirectionParam === "asc" || sortDirectionParam === "desc"
```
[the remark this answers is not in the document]
DafyddLlyr: I'd consider zod for this - super useful for taking an `unknown` into a defined shape.

Alternatively something like `["asc", "desc"].includes(sortDirectionParam)` is a little more succinct and easier to read (in my opinion!).

editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -86,15 +87,9 @@ const Team: React.FC = () => {
     null,
   );
   const [shouldClearFilters, setShouldClearFilters] = useState<boolean>(false);
+  const [sortedFlows, setSortedFlows] = useState<FlowSummary[] | null>(null);
```
RODO94, line 90: In similar fashion to `filteredFlows` and `searchedFlows` - I have introduced a `sortedFlows` which can take the final `matchedFlows` and apply the necessary sort options.

This fixes the issue where sort was not running when the list was searched as `<SearchBox/>` will reset the order of the array by relevance, and it wouldn't trigger the sort to run again on the correct records. 

I did find it hard to achieve what we want without this, and without feeding the `SortControl` component something which may either be outdated (`filteredFlows` or `searchedFlows`) and something which it can't react to (`flows`). This made me settle with feeding it the `matchedFlows`, but this created the need to also give it something it could set, which was not `matchedFlows`. 

I'll take another look in the morning, make sure I've not missed something obvious, but this is giving me the desired result

RODO94: Had another look this morning and couldn't find another way which maintains the pattern of the Sort, Search, and Filter components taking in an array, and returning an altered array. It does feel repetitive, which is my main concern, but it does keep a consistency in how they all work, and should make it easy to follow or change in the future

editor.planx.uk/src/pages/Team/components/FlowCard.tsx
```diff
@@ -136,15 +136,6 @@ const FlowCard: React.FC<FlowCardProps> = ({
 }) => {
   const [deleting, setDeleting] = React.useState(false);
 
-  const handleDelete = () => {
```
RODO94, line 139: This came from a rebase I done, I guess due to us removing the `deleteFlow` mut

editor.planx.uk/src/pages/Team/components/FlowCard.tsx
```diff
@@ -136,15 +136,6 @@ const FlowCard: React.FC<FlowCardProps> = ({
...
-    useStore
-      .getState()
-      .deleteFlow(teamId, flow.slug)
-      .then(() => {
-        setDeleting(false);
-        refreshFlows();
-      });
-  };
```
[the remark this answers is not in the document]
RODO94: I rebased the chain right now, but it remains, if I revert the change, I get an error in my react tests for `deleteFlow()` not existing. I'll revert these file changes, and rebase again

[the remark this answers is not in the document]
RODO94: @DafyddLlyr this is because `FlowCard` does not yet appear in `main`, it was introduced in the feature branch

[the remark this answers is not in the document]
DafyddLlyr: Ah that makes sense - thanks for confirming 👍

PR #4189, opened 2025-01-23
feat: Updated team page with filters

ianjon3s:
## Main branch for merging team page work into

Testing Notes:

#### General

- As a “teamEditor” of a team I can navigate to this team see all of my services
- For each service I can:
  - Read the service title
  - See whether a service is online or offline
  - See whether a service is a submission service
  - See whether a service is a statutory service
  - See and operate the functional menu for each service
- If I’m not a “teamEditor” of a team I can do all of the above, with exception that the functional menu is not visible
- When I publish a service with a Send component, this service is given the submission tag
- When I publish a submission service with application.type in the data field of a component and a valid data value according to the ODP Schema, this service is given the statutory tag

#### Searching
- When viewing my team page I can locate the search input to:
  - Enter a search term, to filter down the list of displayed services
  - Use the close (X) button to remove my search term and reset results
- When a search term is present:
  - The list of services is displayed with valid and useful results
  - The services list subtitle shows me how many results I am viewing
  - If there are no results, I am shown the subtitle “No results”
  - I can use the “clear all filters” link to reset to default

#### Filtering
- I can locate the filters dropdown to:
  - Toggle filters off and on
- When one or more filters are enabled
  - The list of services is displayed with valid results (corresponding to selected filters)
  - I can see which filters are active with the addition of “chips” for each applied filter
  - I can remove an applied filter by using the close (X) button on a chip
  - The services list subtitle shows me how many results I am viewing
  - If there are no results, I am shown the subtitle “No results”
  - I can use the “clear all filters” link to reset to default
  - I can copy the page address/URL, to share a filtered view of services (open in a new tab to test)
  
#### Reordering/sorting
- By default, the list of services is ordered “Last updated” + “Newest first”
- I can use the dropdown menus to:
  - Change the parameter for ordering
  - Change the direction of ordering
- I can copy the page address/URL, to share an ordered view of services (open in a new tab to test)

PR #4298, opened 2025-02-11
fix: remove transparent background to Checkbox component

RODO94:
Recurring issue in the Filters work here: https://github.com/theopensystemslab/planx-new/pull/4189

Initial fix was to set the `background:"inherit"` but it didn't seem to solve the issue, maybe due to an issue on what it was inheriting.

Found the simplest solution was to set it to the colour of it's non-focused style. I also tried `unset` and other attributes, but nothing worked as well.


DafyddLlyr (approved):
Perfect - thank you!

PR #4300, opened 2025-02-11
fix: Clickable width and alignment of checkboxes

ianjon3s:
## What does this PR do?

Fixes a global (public & editor) issue where the clickable width of checkboxes extends beyond the text label.

Also addresses vertical alignment so that the first line of text is always aligned with the centre of the checkbox (rather than the whole text block).

Making both of these changes brings us inline with the GOV.UK checkbox style:
https://design-system.service.gov.uk/components/checkboxes/

Preview:
https://4300.planx.pizza/testing/checkboxes-variable-width/published?analytics=false


jessicamcinchak (approved):
Good spot, thank you!

DafyddLlyr (approved):

PR #4303, opened 2025-02-12
refactor: Readme page external portal switch to expand

ianjon3s:
## What does this PR?

Updates readme page show/hide external portals from using a `<Switch>` to use `<SimpleExpand>`.

Rationale:
We use the switch component as an in-form input to toggle functionality or settings, this is occasionally tied in with an interface change (for example when using expandable checklists). For interface only (show/hide) functions we use `<SimpleExpand>` or the accordion component.

Something I'm not sure about (cc @jamdelion ) — is the switch/toggle a performance consideration (only fetch portal list on "show"), which I'm overriding with this change, or is it a purely visual thing?

Preview:
https://4303.planx.pizza/barnet/apply-for-prior-approval/about

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/ReadMePage/ReadMePage.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/ReadMePage/ReadMePage.tsx
@@ -2,6 +2,7 @@ import Box from "@mui/material/Box";
 import Button from "@mui/material/Button";
 import Container from "@mui/material/Container";
 import Typography from "@mui/material/Typography";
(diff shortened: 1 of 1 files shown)


jamdelion (approved):
Thanks Ian, I've tried this out and works as expected.

As for your question about performance, the toggle was purely visual for showing the portals - we fetch the portals either way from the store when the component renders for the first time (see lines 33-51). 

By using the SimpleExpand, you've removed the need for this page to store state about whether the portals are showing or not, so line 57 `  const [showExternalPortals, setShowExternalPortals] = useState(false);` (and the import of `useState` from React) can also be removed. 😁

PR #4305, opened 2025-02-12
feat: hoist submissions log to team level

jamdelion:
## In this PR:

- Adjust submissions log query to get all submissions for a team
- Add flow name to `submissions_log_summary` view
- Pass props into the Submission log to filter on the front end for a particular flow if navigating to the log from a flow

🎟️ Trello ticket: https://trello.com/c/tuVo5aIF/3149-make-submission-log-at-the-team-level-filterable

## NOT in this PR:

- Making the submissions log filterable/searchable

## Screenshot

Screenshot 2025-02-12 at 17 44 00


editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/types.ts
```diff
@@ -0,0 +1,38 @@
...
+    | "Submitted"
+    | "Capturable"
+    | "Failed"
+    | "Cancelled"
+    | "Error"
+    | "Unknown";
+  retry: boolean;
+  response: Record<string, any>;
+  createdAt: string;
+  flowName: string;
```
jamdelion, line 28: `flowName` added here

editor.planx.uk/src/routes/submissionsLog.tsx
jamdelion, line 1: split out into `flowSubmissionsLog` and `teamSubmissionsLog`

editor.planx.uk/src/routes/teamSubmissionsLog.tsx
```diff
@@ -0,0 +1,38 @@
...
+          title: makeTitle([teamSlug, "submissions-log"].join("/")),
+          view: Submissions,
+        };
+      }),
+    ),
+    "/:flow": compose(
+      route(async (req) => {
+        const { team: teamSlug, flow: flowSlug } = req.params;
+
+        console.log("req.params", req.params);
```
RODO94, line 27: nit: console.log here

hasura.planx.uk/migrations/1739375815458_run_sql_migration/up.sql
```diff
@@ -0,0 +1,78 @@
...
+ SELECT ls.flow_id,
+    ae.session_id,
+    ae.event_id,
+    ae.event_type,
+    ae.status,
+    ae.response,
+    ae.created_at,
+    ae.retry,
+    f.team_id,
+    f.name AS flow_name
```
RODO94, line 73: @jamdelion is `flow_name` the change?

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/index.tsx
```diff
@@ -60,31 +26,52 @@ const Submissions: React.FC = () => {
...
+          available to be downloaded by team editors.`}
         </Typography>
       </SettingsSection>
       <SettingsSection>
-        <EventsLog submissions={submissions} loading={loading} error={error} />
+        <EventsLog
+          submissions={filteredSubmissions}
+          loading={loading}
+          error={error}
+          filterByFlow={Boolean(flowSlug)}
```
RODO94, line 73: `filterByFlow` is used here to determine whether we're showing submissions at Flow level or Team level?

RODO94 (commented):
It was working well for me on the pizza, I just had a few questions to make sure I was following what was happening.

Also, I couldn't run `sync-data` on the branch locally due to extra data on seeding flows, might just need a rebase to fix

editor.planx.uk/src/routes/team.tsx
```diff
@@ -90,8 +90,10 @@ const routes = compose(
 
     "/:flow/service": setFlowAndLazyLoad(() => import("./serviceSettings")),
 
-    "/:flow/submissions-log": setFlowAndLazyLoad(
-      () => import("./submissionsLog"),
+    "/submissions-log": lazy(() => import("./teamSubmissionsLog")),
+
+    "/submissions-log/:flow": setFlowAndLazyLoad(
+      () => import("./flowSubmissionsLog"),
```
DafyddLlyr, line 96: I think that if we make this the same route, but with an optional query param for a pre-set filter of flowSlug (or flowId) we might give ourselves a bit more flexibility here.

So rather than `/submissions-log/:flow` being a unique route, we just have a link at flow level that points to `/:teamSlug/submissions-log?flow=:flowSlug`.

This then take the user to a pre-filtered view of the same page, and they could remove this filter if they wish to do so. This might be a little simpler and have a lot of the conditionals we currently have on this component. This might just be something we need to do in future once we have the datagrid in place and might not be wholly practical just now. What do you think?

DafyddLlyr (approved):
One small question about implementation that we can discuss, fix forward, or leave as-is - just wanted to raise the suggestion.

Looking good - nice and simple 👍

hasura.planx.uk/migrations/1739375815458_run_sql_migration/up.sql
jamdelion: Yep added that as a new column in the view :)

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/index.tsx
jamdelion: Exactly - if there is a flowSlug in the query params, the table only shows submissions for that flow. Otherwise it shows all submissions for that team across all flows.

PR #4312, opened 2025-02-13
feat: Dedicated styles for locked editor graph and modal

ianjon3s:
## What does this PR do?

- Sets styles for non-editable graph + components (to be shared with templates)
- Configures editor modal to have `disabled` inputs when non-editable (only "Section" component done so far)

Preview:
https://4312.planx.pizza/barnet/apply-for-prior-approval

### DIFF
--- a/editor.planx.uk/src/@planx/components/Checklist/Editor/components/OptionsEditor.tsx
+++ b/editor.planx.uk/src/@planx/components/Checklist/Editor/components/OptionsEditor.tsx
@@ -2,7 +2,6 @@ import {
   BaseOptionsEditor,
   BaseOptionsEditorProps,
 } from "@planx/components/shared/BaseOptionsEditor";
-import { props } from "ramda";
 import React from "react";
 import SimpleMenu from "ui/editor/SimpleMenu";
 
--- a/editor.planx.uk/src/@planx/components/Checklist/Editor/components/OptionsEditor.tsx
+++ b/editor.planx.uk/src/@planx/components/Checklist/Editor/components/OptionsEditor.tsx
@@ -2,6 +2,7 @@ import {
   BaseOptionsEditor,
   BaseOptionsEditorProps,
 } from "@planx/components/shared/BaseOptionsEditor";
+import { props } from "ramda";
 import React from "react";
 import SimpleMenu from "ui/editor/SimpleMenu";
 
@@ -11,6 +12,7 @@ export type ChecklistOptionsEditorProps = BaseOptionsEditorProps & {
   groups?: Array<string>;
   onMoveToGroup?: (itemIndex: number, groupIndex: number) => void;
   showValueField?: boolean;
+  disabled?: boolean;
 };
 
 const ChecklistOptionsEditor: React.FC<ChecklistOptionsEditorProps> = ({
@@ -21,6 +23,7 @@ const ChecklistOptionsEditor: React.FC<ChecklistOptionsEditorProps> = ({
   groups,
   onMoveToGroup,
   index,
+  disabled,
 }) => {
   return (
     <BaseOptionsEditor
@@ -28,6 +31,7 @@ const ChecklistOptionsEditor: React.FC<ChecklistOptionsEditorProps> = ({
       schema={schema}
       onChange={onChange}
       showValueField={showValueField}
+      disabled={disabled}
     >
       {typeof index !== "undefined" && groups && onMoveToGroup && (
         <SimpleMenu
--- a/editor.planx.uk/src/@planx/components/InternalPortal/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/InternalPortal/Editor.tsx
@@ -36,7 +36,7 @@ const InternalPortalForm: React.FC<{
   flows = [],
   tags = [],
   notes = "",
-  disabled = false,
+  disabled,
 }) => {
   const formik = useFormik({
     initialValues: {
@@ -113,7 +113,11 @@ const InternalPortalForm: React.FC<{
           </ModalSectionContent>
         )}
       </ModalSection>
-      <ModalFooter formik={formik} showMoreInformation={false} disabled={disabled} />
+      <ModalFooter
+        formik={formik}
+        showMoreInformation={false}
+        disabled={disabled}
+      />
     </form>
   );
 };
--- a/editor.planx.uk/src/@planx/components/InternalPortal/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/InternalPortal/Editor.tsx
@@ -28,6 +28,7 @@ const InternalPortalForm: React.FC<{
   handleSubmit?: (val: any) => void;
   flows?: Array<Flow>;
   tags?: NodeTag[];
+  disabled?: boolean;
 }> = ({
   handleSubmit,
   text = "",
@@ -35,6 +36,7 @@ const InternalPortalForm: React.FC<{
   flows = [],
   tags = [],
   notes = "",
+  disabled = false,
 }) => {
   const formik = useFormik({
     initialValues: {
@@ -79,7 +81,7 @@ const InternalPortalForm: React.FC<{
               placeholder="Enter a portal name"
               rows={2}
               value={formik.values.text}
-              disabled={!!formik.values.flowId}
+              disabled={disabled || !!formik.values.flowId}
               id="portalFlowId"
             />
           </ErrorWrapper>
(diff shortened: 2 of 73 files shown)


editor.planx.uk/src/@planx/components/Section/Editor.tsx
```diff
@@ -14,6 +15,8 @@ import { parseSection, Section } from "./model";
 
 type Props = EditorProps<TYPES.Section, Section>;
 
+const teamSlug = window.location.pathname.split("/")[1];
```
jamdelion, line 18: I'd expect to get the team slug from the store rather than the pathname on the page (i.e. maintaining that the store is our source of truth which feeds into everything else, including the pathname). 

There are [examples in other files](https://github.com/theopensystemslab/planx-new/blob/b2df58e95769cf0f42f4c656d004d8c112c085e6/editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/EventsLog.tsx#L101) of getting both the team and the `canUserEditTeam` function from the store at once, e.g. 

```typescript
const [teamSlug, canUserEditTeam] = useStore((state) => [
    state.teamSlug,
    state.canUserEditTeam,
    ])
    
     canUserEditTeam(teamSlug)
    ...
    
  ```

**Comment by ianjon3s** [comment]
@DafyddLlyr @jamdelion , thanks both for your comments and guidance.

I've refactored this so that the `disabled` state is not controlled from `FormModal.tsx` based on user permission.

This is working for the Section and Find property components:
https://4312.planx.pizza/barnet/apply-for-prior-approval/nodes/_root/nodes/f9HnFqX403/edit
https://4312.planx.pizza/barnet/apply-for-prior-approval/nodes/_root/nodes/T1zEiqX403/edit

If this approach is the correct one I'll continue to work through the other components to correctly apply the `disabled` state.

I was able to apply the `disabled` state to the tag selector but this caused tests to fail, I'll need some assistance with this eventually as I'm not familiar with the logic required for multiple test scenarios.

File: editor.planx.uk/src/@planx/components/Section/Editor.tsx [hunk 2]
```diff
@@ -1,6 +1,7 @@
 import { ComponentType as TYPES } from "@opensystemslab/planx-core/types";
 import { EditorProps } from "@planx/components/shared/types";
 import { useFormik } from "formik";
+import { useStore } from "pages/FlowEditor/lib/store";
```
DafyddLlyr, line 4: nit: Unused import

editor.planx.uk/src/@planx/components/shared/types.tsx
```diff
@@ -6,6 +6,7 @@ export interface EditorProps<Type, Data> {
   id?: string;
   handleSubmit?: (data: { type: Type; data: Data }) => void;
   node?: any;
+  disabled?: boolean;
```
DafyddLlyr, line 9:
```suggestion
  disabled: boolean;
```

We'll eventually want to make this mandatory. Having it optional for now allows you to fix a component at a time however. Something to make required just before opening for review I think 👍

editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
```diff
@@ -143,8 +143,11 @@ const FormModal: React.FC<{
   ]);
   const handleClose = () => navigate(rootFlowPath(true));
 
-  // useStore.getState().getTeam().slug undefined here, use window instead
-  const teamSlug = window.location.pathname.split("/")[1];
+  const teamSlug = useStore.getState().getTeam().slug;
```
DafyddLlyr, line 146: Have you had any issues with this method? The comment this replaces suggests there might be an issue with using the store. Not sure of the original source of this, just flagging!

editor.planx.uk/src/theme.ts
```diff
@@ -580,6 +581,11 @@ const getThemeOptions = ({
                 opacity: 0,
               },
             },
+            "& .MuiSwitch-switchBase.Mui-disabled": {
+              "& + .MuiSwitch-track": {
```
DafyddLlyr, line 585: Are we able to import these from MUI - e.g. `switchClasses`?

editor.planx.uk/src/@planx/components/shared/types.tsx
DafyddLlyr: Just seen this comment applies a few times throughout this PR!

editor.planx.uk/src/ui/editor/RichTextInput/RichTextInput.tsx
```diff
@@ -40,7 +40,7 @@ import {
   linkSelectionError,
 } from "./validationHelpers";
 
-const RichTextInput: FC<Props> = (props) => {
+const RichTextInput: FC<Props & { disabled?: boolean }> = (props) => {
```
DafyddLlyr, line 43: I'd modify `Props` directly here as opposed to extending this just in-line.

DafyddLlyr (commented):
Looking great - this is exactly what I had in mind 🙌

editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
ianjon3s: I've not had any issues getting this from the store, I'll keep on eye on this thought.

editor.planx.uk/src/theme.ts
ianjon3s: Yes, I've updated to use the correctly imported classes. In order to get this to work I had to import + export from `<Switch>` as we're using our version of the component.

editor.planx.uk/src/ui/editor/RichTextInput/RichTextInput.tsx
ianjon3s: Have updated to edit `RichTextInput` props directly

editor.planx.uk/src/@planx/components/InternalPortal/Editor.tsx
```diff
@@ -28,13 +28,15 @@ const InternalPortalForm: React.FC<{
...
   tags?: NodeTag[];
+  disabled?: boolean;
 }> = ({
   handleSubmit,
   text = "",
   flowId = "",
   flows = [],
   tags = [],
   notes = "",
+  disabled = false,
```
DafyddLlyr, line 39: We shouldn't need to define a fallback value like this at a low level - we should be able to do it once at a high level where we run our permission check. This goes hand-in-hand with making `disabled` a mandatory field and not an optional one 👍

editor.planx.uk/src/theme.ts
```diff
@@ -22,6 +22,7 @@ import { gridClasses } from "@mui/x-data-grid";
 import type {} from "@mui/x-data-grid/themeAugmentation";
 import { TeamTheme } from "@opensystemslab/planx-core/types";
 import { getContrastTextColor } from "styleUtils";
+import { switchClasses } from "ui/shared/Switch";
```
DafyddLlyr, line 25: We should be able to import this directly and remove the export from out `Switch` by adding the following line - 

```suggestion
// eslint-disable-next-line no-restricted-imports
import { switchClasses } from "@mui/material/Switch";
```

editor.planx.uk/src/ui/shared/Switch.tsx
```diff
@@ -14,15 +17,19 @@ interface Props {
   name?: string;
   variant?: "editorPage" | "editorModal";
   capitalize?: boolean;
+  disabled?: boolean;
 }
 
+export { switchClasses };
```
DafyddLlyr, line 23: Please see previous comment.

DafyddLlyr (approved):
Worked through the code here - all looking good and as expected - mammoth task!

A few small issues to address - 

 - `disabled` shouldn't be an optional prop (remove `?`)
 - `disabled` shouldn't need a fallback value (as it will always be `true` or `false`)

Working through the UI now and will list any issues I hit (if any) before coming back for final approval ✅

DafyddLlyr (changes requested):
A few visual comments - 

 - Incorrect border colour on data fields in graph view 
 
 
image

- Clones - white BG and black dashed line

image

- Alt test "Edit" button on images is still enabled

image

- DateInput modal still has editable fields

image

ianjon3s:
Thanks so much for these @DafyddLlyr , I'll work through 👍

editor.planx.uk/src/@planx/components/Checklist/Editor/components/OptionsEditor.tsx
```diff
@@ -2,6 +2,7 @@ import {
   BaseOptionsEditor,
   BaseOptionsEditorProps,
 } from "@planx/components/shared/BaseOptionsEditor";
+import { props } from "ramda";
```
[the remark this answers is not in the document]
ianjon3s: Good catch, have removed

ianjon3s:
@DafyddLlyr I've tested this with the `all-components` flow with permission to edit true + false and all seems well. Just need a ✅ to get it across the line

DafyddLlyr (approved):

PR #4313, opened 2025-02-13
feat: add archiving option to services

RODO94:
## What does this PR do?

Trello ticket: https://trello.com/c/fhoa1wCa/3166-add-ability-to-archive-flows-instead-of-deleting-them

A user noted that it is quite scary to have a delete option on the flows, as they may hit it accidentally. To resolve this we have updated the messaging on the dialog to note that any deleted flows can be restored, and we've also changed the naming from `delete` to `archive`. 

With this, I have added a new mutation to the API with relevant testing. To `archive` a flow, we are setting `deleted_at` to the current `timestampz` 

---

I think there is future improvements here to break up the `StyledMenu` component in the Team page in the front end, and add a bit more test coverage for what you can do to a flow and what should be shown at different times in the page. This is an improvement for a separate PR given the work going on in the Team/index.tsx page.


editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -72,42 +67,6 @@ const LinkSubText = styled(Box)(({ theme }) => ({
   paddingTop: "0.5em",
 }));
 
-const Confirm = ({
```
RODO94, line 75: Confirm dialog has been moved to `ArchiveDialog`

DafyddLlyr (changes requested):
This solution is slightly over-engineered...!

For a simple mutation, I'd expect to see just a graphql mutation via the frontend, much like [the delete mutation which this was replacing](https://github.com/theopensystemslab/planx-new/pull/4234/files#diff-acaaa6455b303e1390c950ad473c46ecc82d3a63110ef08f8855ce381c9737e5L347-L364).

I'd reach for a REST API endpoint if - 
- There's custom business logic that needs to be called from multiple places
- We need to hide secrets (e.g. calling OS or GovPay)
- We need to call business logic from another application - not the frontend (e.g. Hasura)
- We need complex validation

For CRUD operations with our database, we should be using the GraphQL API - nice and simple!

Admittedly we have the slight quirk in that sometimes we have planx-core functions. Again though, I'd reach for these when we need to be calling something from multiple places or if it contains business logic which we don't want to repeat.

This is certainly something we need to clear up and be more consistent with in the repo - I've got the workings of a proposal for how we can simplify and standardise a lot of this. Something for an upcoming dev call once I have a bit more thought and some diagrams to talk through!

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -158,6 +158,7 @@ export interface FlowSummary {
 
 export interface EditorStore extends Store.Store {
   addNode: (node: any, relationships?: any) => void;
+  archiveFlow: (flowId: string, teamSlug: string) => Promise<any>;
```
DafyddLlyr, line 161: There's no need for an `any` here - we know the return type.

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -228,6 +229,28 @@ export const editorStore: StateCreator<
     send(ops);
   },
 
+  archiveFlow: (flowId, teamSlug) => {
+    const valid = get().canUserEditTeam(teamSlug);
+    if (!valid) {
+      alert(`You do not have permission to archive this flow from ${teamSlug}`);
+      return Promise.resolve();
+    }
```
DafyddLlyr, line 237: This validation is now happening multiple times - 
 - In the UI displaying the archive button
 - Here
 - `useTeamAuth` middleware
 - Hasura permissions
 
It's good to be safe but this is a bit much 😅 If permissions change we'll need to remember to keep all of these in sync.

editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -123,25 +82,43 @@ const FlowItem: React.FC<FlowItemProps> = ({
   teamSlug,
   refreshFlows,
 }) => {
+  const [isArchiving, setIsArchiving] = useState<boolean>(false);
```
DafyddLlyr, line 85: nit: `isArchiving` reads to me like an active process (i.e. the operation is running). What this really is referring to is if the dialog is open. Maybe something more explicit like `isArchiveDialogOpen` would be more appropriate here.

DafyddLlyr (changes requested):

DafyddLlyr (approved):
Perfect - thanks for making those changes!

PR #4332, opened 2025-02-17
feat: refine fuzzy search on SearchBox comp

RODO94:
## What does this PR do?

This PR is responding to feedback here: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1739788849390769

Where we want our search matching to match words not letters. To meet this need I went through the **Extended Search** examples in FuseJS: https://www.fusejs.io/examples.html#extended-search

In order to match only with a word when searching rather than letters, I have adjusted the way we process the search `pattern` by adding a ` ' ` to each word in the search term.

The space acts as an `AND` operator while the ` ' ` acts as an `include-match` search option which includes the whole word in the search matching


editor.planx.uk/src/ui/shared/SearchBox/utils.ts
DafyddLlyr, line 1: As opposed to modifying our input string, a better design here might be to control this behaviour via config on the `useSearch()` hook.

Ideally we want this hook to be an abstraction of all search logic, allowing us to maybe swap it out or re-implement it in future. By splitting up the responsibility for how search works this will get harder to understand and maintain.

DafyddLlyr (changes requested):

editor.planx.uk/src/ui/shared/SearchBox/utils.ts
RODO94: @DafyddLlyr added the change I think you mean, `word-match | letter-match` made sense to me based on what you are trying to search match with. Iterated a bit on `word-search` or `letter-search` but I felt matching made more sense than searching

RODO94:
[[EPIC] As an Editor I want to have a way to organise my flows, so that I can easily find what I'm looking for](https://trello.com/c/7av8lsVr/3160-epic-as-an-editor-i-want-to-have-a-way-to-organise-my-flows-so-that-i-can-easily-find-what-im-looking-for)

editor.planx.uk/src/hooks/useSearch.ts
```diff
@@ -4,6 +4,7 @@ import { useEffect, useMemo, useState } from "react";
 interface UseSearchProps<T extends object> {
   list: T[];
   keys: Array<FuseOptionKey<T>>;
+  searchType?: "word-match" | "letter-match"
```
DafyddLlyr, line 7:
```suggestion
  searchType?: "include-match" | "fuzzy-match"
```

nit: This is a bit more accurate I think.

editor.planx.uk/src/hooks/useSearch.ts
```diff
@@ -61,7 +63,20 @@ export const useSearch = <T extends object>({
...
+  // split pattern by words and remove empty spaces
+  const patternArray = pattern.split(" ").filter(Boolean)
+
+  // In FuseJS extended search the character: ' 
+  // is used for the include-match search option
+  // so it will include this whole word in matching
+  const formattedPattern = `'${patternArray.join(" '")}`
+
+  return formattedPattern
+}
```
DafyddLlyr, line 82: nit: I'd try to find a way to reduce comments like this - the cognitive load feels quite high for a simple function. I'm having to to read the js, match this to comments, and refer back to fusejs docs to really get a hold of what's happening here.

Something like this illustrates my point (there's a lot of room for improving variable names here! 😅) - 

```suggestion
const convertToIncludeMatchPattern = (fuzzyMatchPattern: string) => {
  const words = fuzzyMatchPattern.split(" ").filter(Boolean);
  // Docs: http://fuse.js/link-to-docs
  const includeCharacter = "'"
  const includeMatchPattern = words.map((word) =>`${includeCharacter}word`).join(" ");
  
  return includeMatchPattern;
}
```

This is untested code, and we're maybe drifting into personal styles / opinions here, but I think the above is more explicit and there's less room for drift between comments and code.

DafyddLlyr (approved):
A few things to take a look at, but no showstoppers!

PR #4333, opened 2025-02-17
feat(iac): Configure RDS parameter group

DafyddLlyr:
## What does this PR do?
- Creates an RDS parameter group
- Turns "force SSL" off, to enable the connection to ShareDB
- Applies parameter group to database

This will need to be manually ran as the data layer is not part of our CI.

Please see https://opensystemslab.slack.com/archives/C01E3AC0C03/p1739530888458399 and https://opensystemslab.slack.com/archives/C01E3AC0C03/p1739792174735719 for further context (OSL Slack).

### DIFF
--- a/infrastructure/data/index.ts
+++ b/infrastructure/data/index.ts
@@ -10,6 +10,20 @@ const DB_ROOT_USERNAME = "dbuser";
 const env = pulumi.getStack();
 const networking = new pulumi.StackReference(`planx/networking/${env}`);
 
+// ShareDB does not use SSL - from Postgres 16 onwards this is forced on my default
+// Create a parameter group which turns this setting off
+const parameterGroup = new aws.rds.ParameterGroup("parameterGroup", {
+  name: "postgres16-force-ssl-off",
+  family: "postgres16",
+  parameters: [
+    {
+      applyMethod: "immediate",
+      name: "rds.force_ssl",
+      value: "0",
+    },
+  ],
+});
+
 const db = new aws.rds.Instance("app", {
   engine: "postgres",
   // AWS restricts the maximum upgrade leap, see available versions:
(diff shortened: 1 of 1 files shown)


RODO94 (commented):
@DafyddLlyr a parameter group is a way for us to set custom parameters to our RDS for Postgres instance, it's default is 1 and we're switching to 0 to turn it off as set here: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.PostgreSQL.CommonDBATasks.Parameters.html#:~:text=rds.force_ssl,to%200%20(off).

This is ultimately needed because the ShareDB server needs to interact with our `operations` table, and Postgres wants an SSL connection which Sharedb isn't providing us?

jessicamcinchak (approved):
Thank you! Happy to screenshare on staging later to test this deploy if you need second eyes

jessicamcinchak:
[Update AWS RDS postgreSQL engine version](https://trello.com/c/qda9058V/3186-update-aws-rds-postgresql-engine-version)

infrastructure/data/index.ts
```diff
@@ -10,6 +10,20 @@ const DB_ROOT_USERNAME = "dbuser";
 const env = pulumi.getStack();
 const networking = new pulumi.StackReference(`planx/networking/${env}`);
 
+// ShareDB does not use SSL - from Postgres 16 onwards this is forced on my default
+// Create a parameter group which turns this setting off
+const parameterGroup = new aws.rds.ParameterGroup("parameterGroup", {
+  name: "postgres16-force-ssl-off",
+  family: "postgres16",
```
freemvmt, line 17: My only thought here is whether the 'default' parameter group prescribed by Pulumi may or may not have any differences from `postgres16`? Seems unlikely. I also wonder if there's any merit to doing this slightly differently, as per the example [here](https://www.pulumi.com/registry/packages/aws/api-docs/rds/instance/#rds-db2-usage)) (again, I think probably unnecessary, just thinking out loud).

Out of curiosity, you find the family `postgres16` by running the following command as per [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbparametergroup.html#aws-resource-rds-dbparametergroup-properties)?

```
aws rds describe-db-engine-versions --query "DBEngineVersions[].DBParameterGroupFamily" --engine postgres
```

freemvmt (approved):

PR #4334, opened 2025-02-17
feat: setup event trigger for new PlanX team -> new Metabase collection

zz-hh-aa:
# What does this PR do?
- Creates a new event trigger that listens for new rows added to `teams` table
- Changes the route (slug is now in the request body, not a URL parameter)

# Why?
- Most of the code has been written to automate the creation of collections (Metabase speak for 'folder'), we just need the plumbing in place / to know when to access the `/metabase/collection/` endpoint and what params to pass to it

# Questions
1. Will there be any instances of teams that shouldn't have a Metabase collection?
2. What should we do for all existing teams that don't have Metabase collections yet? Should they be created and updated manually? 
3. Do we need to set up retries in Hasura? 

I've tested this locally (adding new teams in `teams`) and new collections were successfully created in Metabase staging

### DIFF
--- a/api.planx.uk/modules/analytics/metabase/collection/types.ts
+++ b/api.planx.uk/modules/analytics/metabase/collection/types.ts
@@ -25,15 +25,13 @@ export type MetabaseCreateCollectionParams = {
  * the Metabase collection ID is for the "Council" collection
  * see https://github.com/theopensystemslab/planx-new/pull/4072#discussion_r1892631692
  **/
-// const COUNCILS_COLLECTION_ID = 58;
+const COUNCILS_COLLECTION_ID = 58;
 
 export const createTeamCollectionSchema = z.object({
-  params: z.object({
-    slug: z.string().min(1),
-  }),
   body: z.object({
+    slug: z.string().min(1),
     description: z.string().optional(),
-    parentId: z.number().optional(), //.default(COUNCILS_COLLECTION_ID),
+    parentId: z.number().optional().default(COUNCILS_COLLECTION_ID),
   }),
 });
 
--- a/api.planx.uk/modules/analytics/metabase/collection/types.ts
+++ b/api.planx.uk/modules/analytics/metabase/collection/types.ts
@@ -25,7 +25,7 @@ export type MetabaseCreateCollectionParams = {
  * the Metabase collection ID is for the "Council" collection
  * see https://github.com/theopensystemslab/planx-new/pull/4072#discussion_r1892631692
  **/
-const COUNCILS_COLLECTION_ID = 58;
+const COUNCILS_COLLECTION_ID = 78;
 
 export const createTeamCollectionSchema = z.object({
   body: z.object({
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -2619,6 +2619,26 @@
           - slug
         filter: {}
         check: null
+  event_triggers:
+    - name: setup_metabase_collection
+      definition:
+        enable_manual: false
+        insert:
+          columns: '*'
+      retry_conf:
+        interval_sec: 10
+        num_retries: 0
+        timeout_sec: 60
+      webhook: '{{HASURA_PLANX_API_URL}}/metabase/collection/'
+      headers:
+        - name: authorization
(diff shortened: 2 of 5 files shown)


hasura.planx.uk/metadata/tables.yaml
```diff
@@ -2619,6 +2619,26 @@
...
+        num_retries: 0
+        timeout_sec: 60
+      webhook: '{{HASURA_PLANX_API_URL}}/metabase/collection/'
+      headers:
+        - name: authorization
+          value_from_env: HASURA_PLANX_API_KEY
+      request_transform:
+        body:
+          action: transform
+          template: "{\r\n  \"description\": \"Collection for {{$body.event.data.new.name}}\",\r\n  \"parentId\": 58\r\n}"
```
DafyddLlyr, line 2639: We also need to pass in the slug variable here now it's not in the path.

I currently get this error when trying to add a team on the Pizza - 

```json
{
    "data": {
        "message": "{\"issues\":[{\"code\":\"invalid_type\",\"expected\":\"string\",\"received\":\"undefined\",\"path\":[\"body\",\"slug\"],\"message\":\"Required\"}],\"name\":\"ZodError\"}"
    },
    "type": "client_error",
    "version": "2"
}
```

api.planx.uk/modules/analytics/metabase/collection/types.ts
```diff
@@ -25,15 +25,14 @@ export type MetabaseCreateCollectionParams = {
...
  * see https://github.com/theopensystemslab/planx-new/pull/4072#discussion_r1892631692
  **/
-// const COUNCILS_COLLECTION_ID = 58;
+const COUNCILS_COLLECTION_ID = 58;
 
 export const createTeamCollectionSchema = z.object({
-  params: z.object({
-    slug: z.string().min(1),
-  }),
+  params: z.object({}),
```
DafyddLlyr, line 31: There are no params, we can just remove this line.

DafyddLlyr (changes requested):
Hit a few issues, please see comments.

hasura.planx.uk/metadata/tables.yaml
zz-hh-aa: I believe that's fixed here [a468247](https://github.com/theopensystemslab/planx-new/pull/4334/commits/a4682470ae8765ea2432bb9bd66e0402727fa6f3) and I created a team successfully on the Pizza. Though is there Hasura staging on Pizzas / am I able to check that the db side of things went through correctly?

DafyddLlyr: There's no Metabase on Pizzas, and often not locally. We could set this up, and manage a data sync so that all IDs etc are in line.

I don't really know if this is necessary, other options would be -
 - Hitting errors on Pizzas when accessing Metabase ID (not ideal as it could be noisy)
 - Checking the env (maybe with a middleware) and returning a 200 if there's no Metabase available
 
 This can be managed as a follow up task tbh - I'll have a look at this just now!

DafyddLlyr (approved):
All looking good! Getting the expected error on the Pizza now. Not able to properly test, but we can do so on staging and prod.

We may need to make the council collection ID and env var so it works on staging - alternatively we can cheat this one and manually increment it to match...! We could do this by setting the nextval() for the sequence or just my making and deleting a bunch of teams.

> 1. Will there be any instances of teams that shouldn't have a Metabase collection?
Probably in future! Creating teams will be a platform admin task so we can always add a "Setup Metabase?" checkbox to the form in future.

>2. What should we do for all existing teams that don't have Metabase collections yet? 
> Should they be created and updated manually?
Yes, we should create these on Metabase and backfill their IDs via Hasura.

> 3. Do we need to set up retries in Hasura?
I'd say not yet - if we find that the Metabase API is flakey we can setup retries, but right now a failure on first try would actually be good to know about. Metabase is stable (no downtime reported in a very long time) so I don't think we'll need retries just yet.

zz-hh-aa:
Thanks Daf!

I've now created enough folders to get the parent IDs to match in staging and production (updated ID in [d9b4784](https://github.com/theopensystemslab/planx-new/pull/4334/commits/d9b4784a91d208c2cfc44321934b4b4a0c24832b)) and I've also now manually backfilled team metabase IDs / made Metabase collections for the ones that didn't have any.

PR #4256, opened 2025-02-05
chore: rename "placeholder" tag to "customisation", bump planx-core

jessicamcinchak:
Follows on from #https://github.com/theopensystemslab/planx-core/pull/643 (also picking up two non-related schema PRs on `main`) 

**Key changes:**
- "Placeholder" now renamed "Customisation" per templates design call
- "Customisation" tag is _only_ visible on the graph to platformAdmins, never to anyone else (we will have a new visual design to indiccate editability that will not rely on the tag band)
  - Note the tag is still visible, and disabled, when opening the modal

| As a `platformAdmin`   | As any non-admin user |
| -------- | ------- |
| Screenshot from 2025-02-05 16-08-30 | Screenshot from 2025-02-05 16-09-32 |

PR #4335, opened 2025-02-17
fix: Add temporary guard to convert placeholder tags

DafyddLlyr:
Please see https://opensystemslab.slack.com/archives/C01E3AC0C03/p1739802552794749 for context (OSL Slack).

**What's the problem?**
When https://github.com/theopensystemslab/planx-new/pull/4256 was merged, we didn't consider that nodes may already have the "placeholder" tag applied. Currently, any nodes with this tag will fail to render in the Editor.

**What's the solution?**
The right solution is to migrate any flows which contain the legacy "placeholder" tag to use the new "customisation" tag. I've added a blue trello card for this here - https://trello.com/c/Cvwe9t9N/3241-migrate-placeholder-tags-to-customisation

In the meantime, a quick guard to convert manually will unblock any Editors.

### DIFF
--- a/editor.planx.uk/src/routes/flow.tsx
+++ b/editor.planx.uk/src/routes/flow.tsx
@@ -118,6 +118,12 @@ const editNode = validateNodeRoute(
       [key: string]: any;
     };
 
+    // Temp guard to handle non-migrated tags
+    // TODO: Migrate "placeholder" tags to "customisation"
+    node.data.tags = node.data.tags?.map((tag: string) =>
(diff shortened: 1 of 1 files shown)


RODO94 (approved):
Code does look good as a temporary solution, thanks for picking this up.

Happy to merge once CI passes

jessicamcinchak (approved):
Thanks for this! Apologies didn't spot this sooner, and specifically thought we would _not_ need a data migration for this / eg was non-breaking

PR #4336, opened 2025-02-17
feat: sort null to bottom in SortControl component

RODO94:
## What does this PR do?

This PR responds to feedback here: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1739788849390769

I have added a function to the `orderBy` iteratee section which mutates the original sort array by sorting all null values to the bottom. Currently, for date formats especially, these were being formatted to the 'newest' sort. 

The sort option `last updated` has now been changed to `last edited` as well.

### DIFF
--- a/editor.planx.uk/src/ui/editor/SortControl/SortControl.tsx
+++ b/editor.planx.uk/src/ui/editor/SortControl/SortControl.tsx
@@ -1,7 +1,7 @@
 import Box from "@mui/material/Box";
 import MenuItem from "@mui/material/MenuItem";
 import { styled } from "@mui/material/styles";
-import { orderBy } from "lodash";
+import { get, orderBy } from "lodash";
 import React, { useEffect, useMemo, useState } from "react";
 import { useCurrentRoute, useNavigation } from "react-navi";
 import { Paths } from "type-fest";
@@ -88,7 +88,7 @@ export const SortControl = <T extends object>({
 
   useEffect(() => {
     const { fieldName } = selectedSort;
-    const sortNewFlows = orderBy(records,fieldName, sortDirection).sort((a:T,b:T) => sortNullToBottom(a,b, fieldName))
+    const sortNewFlows = orderBy(records,[(flow) => get(flow, fieldName) || ""], sortDirection)
     setRecords(sortNewFlows);
   }, [selectedSort, sortDirection, records, setRecords]);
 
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/ui/editor/SortControl/SortControl.tsx
```diff
@@ -88,7 +88,7 @@ export const SortControl = <T extends object>({
 
   useEffect(() => {
     const { fieldName } = selectedSort;
-    const sortNewFlows = orderBy(records, fieldName, sortDirection);
+    const sortNewFlows = orderBy(records,fieldName, sortDirection).sort((a:T,b:T) => sortNullToBottom(a,b, fieldName))
```
jessicamcinchak, line 91: on quick search of "lodash sort nulls last", I find this one-line solution which is perhaps much simpler & easier to read - does this work? 
```
_.orderBy(myArray, ({ propertyName }) => propertyName || "", ['desc'])
```

editor.planx.uk/src/ui/editor/SortControl/utils.ts
```diff
@@ -3,6 +3,8 @@ import { slugify } from "utils";
 import { z } from "zod";
 
 import { SortableFields } from "./SortControl";
+import { get } from "lodash";
```
jessicamcinchak, line 6: nit: when using a single lodash function we typically favor `import get from "lodash/get"` to import just the required module and not the whole package

jessicamcinchak (commented):
Quick idea on this one that might be worth testing

editor.planx.uk/src/ui/editor/SortControl/SortControl.tsx
RODO94: Yeah this does work, although I wasn't sure what the clearest implementation was. If this is simpler and clearer for you, I'll switch them!

editor.planx.uk/src/ui/editor/SortControl/utils.ts
```diff
@@ -28,3 +30,16 @@ export const getSortParams = <T extends object>(
...
+  const aValue = get(a, fieldName);
+  const bValue = get(b, fieldName);
+  
+  if (!aValue && !bValue) return 0;
+  if (!aValue) return 1;
+  if (!bValue) return -1;
+  
+  // For non-null values, maintain their relative order
+  return 0;
+}
```
jessicamcinchak, line 45: Please remove all changes from this file if now solving in one-line above.

editor.planx.uk/src/ui/editor/SortControl/SortControl.tsx
jessicamcinchak: In your original implementation, I personally found it very confusing/redundant to call each: 
- Lodash's `orderBy`
- Native `sort`
- Custom function `sortNullToBottom`

jessicamcinchak (approved):
Thanks for changes here!

PR #4339, opened 2025-02-18
feat: add unit tests for Filters component

RODO94:
## What does this PR do?

This PR adds unit test coverage for the new `Filters` component as defined here: https://trello.com/c/YU56MRW8/3230-add-unit-tests-for-filter-component

Purpose is to ensure the component can receive an array called `records`, filter the records using `filterOptions` and set a new array with `setRecords`.

It does not delve into testing the inputs at their component level, for instance checking that checkboxes can be checked. I focused on ensuring Filters specific logic works as expected. 

With this in mind, a test in the Technical Plan made on the Trello ticket that I have not added related to displaying an array of records, but as I was structuring the tests I found this to sit outside the role of `Filters` and would be something the parent component would be responsible for. `Filters` should only receive the array and set a new filtered array when prompted.


RODO94:
[Add Unit tests for Filter Component](https://trello.com/c/YU56MRW8/3230-add-unit-tests-for-filter-component)

RODO94:
Closing and replacing with https://github.com/theopensystemslab/planx-new/pull/4351

PR #4340, opened 2025-02-18
feat: add `property.type.userProvided` to Hasura summary views

zz-hh-aa:
Co-authored with @christinelangston

# What does this PR do?
- Adds `property.type.userProvided` to `ALLOW_LIST` array to show in `analytics_summary` and `submission_services_summary`
- Creates Hasura migrations to show them in views and also in Metabase as read only

# Why?
We want to be able to know what user input property types are in analytics


api.planx.uk/modules/webhooks/service/analyzeSessions/operations.ts
```diff
@@ -34,6 +34,7 @@ const ALLOW_LIST = [
   "service.type",
   "usedFOIYNPP",
   "user.role",
+  "property.type.userProvided",
```
jessicamcinchak, line 37: nit: can we insert into the list in alphabetical order for readability?

hasura.planx.uk/migrations/1739880884117_add_property_type_user_provided_analytics/down.sql
```diff
@@ -0,0 +1,59 @@
...
+    ((al.allow_list_answers -> '_overrides'::text))::text AS overrides,
+    ((al.allow_list_answers -> 'rab.exitReason'::text))::text AS rab_exit_reason,
+    ((al.allow_list_answers -> 'service.type'::text))::text AS pre_app_service_type,
+    ((al.allow_list_answers -> 'application.information.harmful'::text))::text AS pre_app_harmful_info,
+    ((al.allow_list_answers -> 'application.information.sensitive'::text))::text AS pre_app_sensitive_info,
+    ((al.allow_list_answers -> 'application.type'::text) ->> 0) AS application_type,
+    (((al.allow_list_answers -> '_feedback'::text) ->> 'feedbackScore'::text))::integer AS feedback_score,
+    ((al.allow_list_answers -> 'applicant.researchOptIn'::text) ->> 0) AS applicant_research_opt_in,
+    al.has_clicked_save,
+    (((al.allow_list_answers -> 'property.type.userProvided'::text) -> 0))::text AS property_type_user_provided
```
jessicamcinchak, line 55: It's highly unlikely we'll ever have to run the `down` migration - but if we did I'd expect it to be the state of the view _before_ `property_type_user_provided` column was added ! 

(Same feedback applies to submissions view migrations too!)

hasura.planx.uk/migrations/1739880884117_add_property_type_user_provided_analytics/up.sql
```diff
@@ -0,0 +1,61 @@
...
+    ((al.allow_list_answers -> '_overrides'::text))::text AS overrides,
+    ((al.allow_list_answers -> 'rab.exitReason'::text))::text AS rab_exit_reason,
+    ((al.allow_list_answers -> 'service.type'::text))::text AS pre_app_service_type,
+    ((al.allow_list_answers -> 'application.information.harmful'::text))::text AS pre_app_harmful_info,
+    ((al.allow_list_answers -> 'application.information.sensitive'::text))::text AS pre_app_sensitive_info,
+    ((al.allow_list_answers -> 'application.type'::text) ->> 0) AS application_type,
+    (((al.allow_list_answers -> '_feedback'::text) ->> 'feedbackScore'::text))::integer AS feedback_score,
+    ((al.allow_list_answers -> 'applicant.researchOptIn'::text) ->> 0) AS applicant_research_opt_in,
+    al.has_clicked_save,
+    (((al.allow_list_answers -> 'property.type.userProvided'::text) -> 0))::text AS property_type_user_provided
```
jessicamcinchak, line 55: Because `property.type.userProvided` is collected via a TextInput component, it's value will be a plain string in the passport rather than an array of strings - so I don't think the ` -> 0` should be necessary here? 

(Same feedback applies to submissions view migrations too!)

DafyddLlyr (approved):

jessicamcinchak (commented):
Thanks both for picking up this change :star2: 

All the steps are covered correctly here, just a few minor comments !

hasura.planx.uk/migrations/1739880884117_add_property_type_user_provided_analytics/down.sql
DafyddLlyr: Ah missed this 🙈 Good catch!

PR #4344, opened 2025-02-18
feat: filter live services from drop-down menu in admin dashboard

jamdelion:
## In this PR

- Add the ability to filter columns whose values are an array by using a multi select drop-down modelled on MUI's autocomplete.
- If an array item is selected, display it in bold in the filtered rows.
- Be able to delete array items from the filter.
- Generally tidy up the array list by alphabetising, stripping white space etc 
- Refactor to `utils` file

Please thoroughly test the filtering of all the columns and check it behaves as expected (note the Article 4s column is an outstanding issue to come back to).

## Mild concerns 

~~There are a few types I need to come back to - substituted a lot of `any`s while I was hacking/prototyping.~~

I think I've got a bit too down in the weeds with this to properly see if it remains readable/easy to use the generic component. Hopefully fresh eyes/time will help, and implementing it in the next few features!


editor.planx.uk/src/ui/shared/DataTable/utils.tsx
```diff
@@ -0,0 +1,91 @@
...
+) => [
+  {
+    value: "contains",
+    getApplyFilterFn: (filterItem: Record<string, any>) => {
+      if (filteringOne) {
+        return getSingleFilterFn(filterItem);
+      }
+      return getDoubleFilterFn(filterItem);
+    },
+    InputComponent: filteringOne ? SingleInputFilter : DoubleInputFilter,
```
jessicamcinchak, line 35: Ideally, it seems like this would/should be select-one or select-many (rather than select-two). Out of curiousity, have you tried passing in our existing `SelectMultiple` component here?

editor.planx.uk/src/ui/shared/DataTable/mockTeams.tsx
```diff
@@ -38,7 +38,7 @@ export const mockTeams: AdminPanelData[] = [
     liveFlows: [
       "Notify completion of planning application",
       "Find out if you need planning permission",
-      "Apply for prior approval",
+      "apply for prior approval ",
```
jamdelion, line 41: To test the whitespace and casing processes

editor.planx.uk/src/ui/shared/DataTable/DataTable.tsx
```diff
@@ -1,74 +1,109 @@
...
         <DataGrid
           rows={rows}
           columns={dataColumns}
+          onFilterModelChange={handleFilterChange}
           getRowHeight={() => "auto"}
           getRowClassName={(params) =>
             params.indexRelativeToCurrentPage % 2 === 0 ? "even" : "odd"
           }
+          slots={{
+            toolbar: GridToolbar,
```
jamdelion, line 105: Adds the additional toolbar at the top of the table

editor.planx.uk/src/ui/shared/DataTable/components/MultipleOptionSelectFilter.tsx
jamdelion, line 1: Decided to make a separate component from the `SelectMultiple` since a lot of the props were superfluous, and the styling a bit different!

editor.planx.uk/src/ui/shared/DataTable/utils.tsx
jamdelion, line 1: Some of these moved across from `DataTable.tsx`

editor.planx.uk/src/ui/shared/DataTable/utils.tsx
```diff
@@ -0,0 +1,88 @@
...
+  [ColumnType.ARRAY]: (value: string[], filterValues?: string[]) => (
+    <Box component="ol" padding={0} margin={0} sx={{ listStyleType: "none" }}>
+      {value?.map((item: string, index: number) => (
+        <Typography py={0.4} variant="body2" key={index} component="li">
+          {filterValues?.includes(item) ? <strong>{item}</strong> : item}
+        </Typography>
+      ))}
+    </Box>
+  ),
+};
```
jessicamcinchak, line 77: nit: took me a minute to unpack what `componentRegistry` meant here - is something like `cellValueRegistry` or `valueOptionsRegistry` more specific maybe?

jessicamcinchak (commented):
Took a quick pass through the code - only a few very minor comments, otherwise looking great ! 

Happy to do final review & approval today once pizza is up :+1: 

Agree that if this is "working" now, then best way to test current re-useability-ness and spot any further prop tweaking is to just implement it again soon for another case :slightly_smiling_face:

editor.planx.uk/src/ui/shared/DataTable/utils.tsx
```diff
@@ -0,0 +1,88 @@
...
+  return (
+    Array.isArray(filterItem.value) &&
+    filterItem.value.length > 0 &&
+    filterItem.value.every((val) => val != null)
+  );
+};
+
+const containsItem = (item: string, value: Pick<GridFilterItem, "value">) => {
+  if (typeof value === "string") {
+    return item.toLowerCase().includes((value as string).toLowerCase());
```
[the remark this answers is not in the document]
jamdelion: Yeah this is a strange one - despite the type guard conditional, without the `as string` it errors with `Property 'toLowerCase' does not exist on type 'never'` 🤔

editor.planx.uk/src/ui/shared/DataTable/utils.tsx
jamdelion: Thanks, maybe something more like `columnCellComponentRegistry`?

jessicamcinchak:
Just checked pizza:
- :heavy_check_mark: "Team" & "Reference code" free text filters are working as expected
- :heavy_check_mark: "Live services" multiselect filters are working as expected
- :x: non-A4 true/false single select filters are throwing an error - full white screen & `TypeError: t.includes is not a function` console log
- :heavy_check_mark: "Sorting" is working as expected across all columns 
  - nit: should we disable sort for array-type columns like "Live services"; what does "sort" actually mean in that context?

PR #4346, opened 2025-02-18
dp/api revoke on logout

DafyddLlyr:
- **feat: Add `revoked_tokens` table**
- **test: Introspection**
- **chore: Split up modules/auth/service.ts**
- **feat: Store token in revoked_tokens table on logout**


DafyddLlyr:
[Invalidate JWT on logout](https://trello.com/c/GZIRbSNO/3217-invalidate-jwt-on-logout)

api.planx.uk/modules/auth/controller.ts
```diff
@@ -70,3 +73,30 @@ function setJWTSearchParams(returnTo: string, res: Response, req: Request) {
...
+export const logout: RequestHandler = async (_req, res, next) => {
+  const { jwt, sub: userId } = userContext.getStore()?.user || {};
+
+  if (!jwt) {
+    return next(
+      new ServerError({
+        message: `JWT missing from logout request, no token to revoke for user ${userId}`,
+      }),
+    );
+  }
```
DafyddLlyr, line 90: I'm pretty sure this is unreachable, but keeping it here for peace of mind, being explicit with the intentions of the code, and for type narrowing.

The `useLoginAuth()` middleware called as part of the route will reject if there's no JWT - tests prove this behaviour.

editor.planx.uk/src/components/Header/Header.tsx
```diff
@@ -476,6 +473,14 @@ const EditorToolbar: React.FC<{
     setOpen(!open);
   };
 
+  const logout = async () => {
```
DafyddLlyr, line 476: I think there's no issues with writing `revoked_token` records to production tables - the records just won't get used or checked yet.

api.planx.uk/modules/auth/routes.ts
```diff
@@ -7,12 +7,15 @@ export default (passport: Authenticator): Router => {
   const router = Router();
 
   router.get("/auth/login/failed", Controller.failedLogin);
+  router.post("/auth/logout", Middleware.useLoginAuth, Controller.logout);
```
jamdelion, line 10: very minor nit: maybe `useLoginAuth` is a bit of a funny name now it's used on a logout route? Maybe it could be renamed `useLoggedInUserAuth`?

api.planx.uk/modules/auth/service/jwt.ts
```diff
@@ -72,19 +58,27 @@ const getDefaultRoleForUser = (user: User): Role => {
...
+      throw new Error("Invalid JWT format");
+    }
+
+    const payload = decoded.payload;
+
+    if (typeof payload === "string" || !payload?.exp) {
+      throw new Error("Expiry date missing from JWT");
+    }
+
+    return new Date(payload.exp * 1000);
```
jamdelion, line 78: nit: could the `1000` be a descriptive const instead so it's clearer why it is 1000? I had to google what format jwt expiries are in to figure it out :)

api.planx.uk/modules/auth/routes.ts
DafyddLlyr: Good call - please see c9847979e

api.planx.uk/modules/auth/service/jwt.ts
DafyddLlyr: Thanks - yep - this Unix time in seconds vs milliseconds was in my head from working on this recently. Much better to be clear and explicit!

Please see [a91c97d](https://github.com/theopensystemslab/planx-new/pull/4346/commits/a91c97d00b9cf08225723ec8a60349a6bac5f7d2)

api.planx.uk/modules/auth/service/logout/logout.test.ts
```diff
@@ -0,0 +1,156 @@
...
+});
+
+test("invalid tokens will be rejected", async () => {
+  await supertest(app)
+    .post(ENDPOINT)
+    .set({ authorization: "Bearer INVALID JWT" })
+    .expect(401)
+    .then((res) => {
+      expect(res.body).toEqual({
+        error: "No authorization token was found",
```
jamdelion, line 28: Wondering if we want this error to be more specific about the validity of the token in this case. From a debugging point of view I guess we can still just inspect the headers though

api.planx.uk/modules/auth/service/logout/revokeToken.ts
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Generate a digest of the JWT and add the revoked_tokens blocklist
+ * Docs: https://cheatsheetseries.owasp.org/cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.html#no-built-in-token-revocation-by-the-user
+ */
+export const revokeToken = async (jwt: string): Promise<boolean> => {
+  const tokenDigest = createTokenDigest(jwt);
+
+  // Already revoked, no action needed
+  const isRevoked = await isTokenRevoked(tokenDigest);
+  if (isRevoked) return false;
```
jamdelion, line 17: For my own learning, interested in this boolean pattern even though when you call this function you don't use the returned value. I guess if I was writing this I'd have written it like this:

```typescript
if (isRevoked) return
...
await trackRevokedToken()
// no need to return
```

Is there any benefit to doing it your way instead? :)

DafyddLlyr: This was actually not needed in the end, I've just added [6674b54](https://github.com/theopensystemslab/planx-new/pull/4346/commits/6674b54a6f73cbd7cc14944a3983b3c0c21d0f3c) to tidy this up.

You're correct, I think that returning `void` is better if we don't need to use the return value.

My thought here was that I'd need to match the signature of `IsRevoked` from `express-jwt`

```ts
/**
 * A function to check if a token is revoked
 */
export type IsRevoked = (req: express.Request, token: jwt.Jwt | undefined) => boolean | Promise<boolean>;
```

Once I hit the implementation in https://github.com/theopensystemslab/planx-new/pull/4353 I realised this wasn't necessary and I've used [another function](https://github.com/theopensystemslab/planx-new/pull/4353/files#diff-30720519af4d416b4dc82e021ea3257520f791a0bc4d418888407a7a633a864aR139-R144) to match this callback's signature.

api.planx.uk/modules/auth/service/logout/logout.test.ts
DafyddLlyr: Good catch - this should be more specific! I wasn't hitting the issue as the dummy JWT I passed in wasn't being picked up by our `getToken()` callback. Once I correctly formatted it, `express-jwt` now throws a more specific error message.

Please see [5a70fd7](https://github.com/theopensystemslab/planx-new/pull/4346/commits/5a70fd7d66ab58085a05b79ef53648221f68b863)

PR #4351, opened 2025-02-19
feat: unit tests for Filters

RODO94:
## What does this PR do?

This PR adds unit test coverage for the new `Filters` component as defined here: https://trello.com/c/YU56MRW8/3230-add-unit-tests-for-filter-component

Purpose is to ensure the component can receive an array called `records`, filter the records using `filterOptions` and set a new array with `setRecords`.

It does not delve into testing the inputs at their component level, for instance checking that checkboxes can be checked. I focused on ensuring Filters specific logic works as expected. 

With this in mind, a test in the Technical Plan made on the Trello ticket that I have not added related to displaying an array of records, but as I was structuring the tests I found this to sit outside the role of `Filters` and would be something the parent component would be responsible for. `Filters` should only receive the array and set a new filtered array when prompted.

> [!IMPORTANT]
> Waiting for the feature branch to merge to main before adding tests

### DIFF
--- a/editor.planx.uk/src/ui/editor/Filter/tests/Filter.test.tsx
+++ b/editor.planx.uk/src/ui/editor/Filter/tests/Filter.test.tsx
@@ -12,11 +12,7 @@ import {
 import React from "react";
 import { screen } from "@testing-library/react";
 import { axe } from "vitest-axe";
-import {
-  addFilter,
-  expandFilterAccordion,
-  removeFilter,
-} from "./helpers";
+import { addFilter, expandFilterAccordion, removeFilter } from "./helpers";
 
 vi.mock("react-navi", () => ({
   useNavigation: () => ({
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/ui/editor/Filter/tests/Filter.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+  it("manages filter chips correctly when selecting filters", async () => {
+    const { user } = setupTestEnvironment();
+
+    await expandFilterAccordion(screen, user);
+
+    await user.click(screen.getByRole("checkbox", { name: "Online" }));
+    const onlineChip = screen.getByRole("button", { name: "Online" });
+
+    // when you select a filter, a clickable chip should appear for the option
+    expect(onlineChip).toBeVisible();
```
DafyddLlyr, line 85: We need to check it's not there first in order to prove our action made it appear.

editor.planx.uk/src/ui/editor/Filter/tests/helpers.ts
DafyddLlyr, line 1: We can drop an argument from all functions here by just importing `screen` and not passing it in.

`import { screen } from "@testing-library/react";`

DafyddLlyr (approved):
Two small suggestions!

Issue #2404, opened 2023-11-08
Standardise mocking of `CoreDomainClient`


DafyddLlyr:
Please see https://github.com/theopensystemslab/planx-new/pull/2356#discussion_r1381387150 for full context

It would be great to find a way of wrapping this repeated mocking logic in a helper function.

PR #4353, opened 2025-02-19
feat(api): Check if JWT is revoked before allowing access

DafyddLlyr:
## What does this PR do?
- Adds a check of the `revoked_tokens` table by using the `isRevoked` callback from `express-jwt`
- This ensures that any tokens which are revoked, fail validation
- Updates API tests to assume all tokens are not revoked by default - doing this once upstream where `queryMock` is instantiated means we don't have to add this in multiple places.
- API tests - 
  - adds a mock so that all JWTs are non-revoked by default
  - updates mocks and `describe()` blocks - please see comment on PR below 

## Not yet covered
Requests made directly to Hasura with an existing token (not proxied via API).

## To test...
 - Log in, make a copy of the token
 - Log out, check a record has been added to the `revoked_tokens` table
 - Try to re-use token (e.g. via API docs) - access is not allowed!

https://github.com/user-attachments/assets/c5b0762f-14f6-43c6-989b-67838880616d


DafyddLlyr:
[Invalidate JWT on logout](https://trello.com/c/GZIRbSNO/3217-invalidate-jwt-on-logout)

api.planx.uk/modules/admin/session/csv.test.ts
```diff
@@ -12,13 +13,18 @@ const mockGenerateCSVData = vi.fn().mockResolvedValue([
     metadata: {},
   },
 ]);
-vi.mock("@opensystemslab/planx-core", () => {
+
+vi.mock("@opensystemslab/planx-core", async (importOriginal) => {
```
DafyddLlyr, line 17: This pattern is repeated a few times here, grouped in b1d6037b5cabd11ae672d29fd88790625807f3e4

Now that JWT verification needs to make a round trip to the DB, this test needs to access a GraphQL client to do this (`$api.client.request`) which is not mocked. Using `importOriginal()` we can only mock the function we want to mock in this test suite and keep the rest of the original behaviour. In this case it's `this.export.csvData`.

We have an open issue to consolidate these mocks and to make them easier and simpler to manage - the pattern here could be DRY'ed up a but into a `setupPlanXCoreMock()` function called by vitest setup. I've not tackled this here in order to make process and keep things separate. The issue is here - https://github.com/theopensystemslab/planx-new/issues/2404

api.planx.uk/tests/setup/graphQL.ts
```diff
@@ -1,5 +1,19 @@
...
+  data: {
+    revokedToken: {
+      revokedAt: null,
+    },
+  },
+};
 
 beforeEach(() => {
   queryMock.setup(process.env.HASURA_GRAPHQL_URL!);
+  queryMock.mockQuery(mockIsTokenRevokedQuery);
```
jessicamcinchak, line 18: :+1:

jessicamcinchak (approved):
Thanks for super clear testing instructions here, all works as expected for me repeated over a couple times :raised_hands:

freemvmt (approved):
beautiful! such thorough work on the tests.

also I tested as you suggested and it works like a dream! ✨

PR #4356, opened 2025-02-20
chore: add template attributes to editor store

jessicamcinchak:
When we load a flow in the editor, we want to easily know if it's been templated from another flow, or if it's a source template itself for future changes to visual styles and available functionalities.

**Changes:**
- Adds two boolean methods to the editor store `isTemplate` and `isTemplatedFrom`
- Sets them accordingly when we navigate to / load an individual flow
- Ensures `flows.is_template` and `flows.templated_from` columns can be selected by all roles

https://trello.com/c/huXP4VBO/2859-implementation-make-sure-live-services-are-always-up-to-date-aka-flows-copied-from-templates-should-receive-updates

### DIFF
--- a/editor.planx.uk/src/routes/views/flowEditor.tsx
+++ b/editor.planx.uk/src/routes/views/flowEditor.tsx
@@ -81,6 +81,7 @@ export const flowEditorView = async (req: NaviRequest) => {
   const [flow] = req.params.flow.split(",");
   const { id, flowAnalyticsLink, isFlowPublished, isTemplate, templatedFrom } =
     await getFlowEditorData(flow, req.params.team);
+
   useStore.setState({
     id,
     flowAnalyticsLink,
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -482,6 +490,10 @@ export const editorStore: StateCreator<
 
   isFlowPublished: false,
 
+  isTemplate: false,
+
+  isTemplatedFrom: false,
```
jessicamcinchak, line 495: I'm assuming that in the _editor_, we only care about knowing this flow has been templated from another, not specifically _which one_ it's been templated from. That'll come into play later in API-side changes when we set up publishing triggers, etc. 

Any objections?

DafyddLlyr: It's a good question - it seems like it might be handy to have the ID - e.g. we need a title or to query something. I don't think we have a specific use case just now - no issues with a boolean 👍

DafyddLlyr (approved):

PR #4357, opened 2025-02-20
feat: remove statutory application type from tags and filters

RODO94:
Until we get feedback on how to define what a statutory application is, we should remove this from the work.

Current issues with this have been captured in this Slack thread: https://opensystemslab.slack.com/archives/C5Q59R3HB/p1739800315291939?thread_ts=1739788849.390769&cid=C5Q59R3HB

In short, we are incorrectly labelling some flows which are Discretionary services, but within their external portals are still assigning `application.type` to an ODP Schema approved value, which incorrectly triggers the `statutory` tag.


jessicamcinchak (approved):
Will merge this one into feature branch since you're away :heavy_check_mark:

PR #4359, opened 2025-02-20
feat: Create '*' role to denote all users can access route

DafyddLlyr:
## What does this PR do?
- Creates a new value for `Route.accessibleBy` - `*`
- Replaces the full list of roles with `*` - this will mean we don't have to maintain a full list each time we add a role
- Updates logic for `visibleRoutes` to handle `*`

image


ianjon3s (approved):
Working perfectly for me when set as non platformAdmin
✅ Can see resource links at the root level
✅ Can see readme page for a team of which I'm not an editor

PR #4362, opened 2025-02-20
feat: use data grid for submissions log

jamdelion:


editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
jamdelion, line 1: A few linting changes in this PR from merging from main - sorry

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/mockSubmissions.ts
jamdelion, line 1: Useful for trying this out locally without loads of actual submissions - just set submissions in `EventsLog` props to something else (e.g. `submission: bla`) and then add the line `const submissions = mockSubmissions` in `EventsLog.tsx`. This should also come in handy for upcoming tests.

jessicamcinchak (approved):
Thanks for this, all looks good! 

One minor comment related to more realistic mock generation, but feel free to address in a followup PR if necessary !

editor.planx.uk/src/ui/shared/DataTable/types.ts
```diff
@@ -5,9 +5,11 @@ import {
   GridTreeNodeWithRender,
 } from "@mui/x-data-grid";
 
-export const ColumnType = {
+export const ColumnFilterType = {
   BOOLEAN: "boolean",
-  ARRAY: "array",
+  ARRAY: "array", // when the column can be filtered from a range of known values
+  DATE: "date",
+  CUSTOM: "custom",
```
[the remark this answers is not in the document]
jamdelion: No, it does just mean that there is a custom implementation of either the filter (e.g because the cell's data isn't straightforward, such as with the JSON), or because we need to render something special in the cell. Happens to both be JSON related here!

PR #4363, opened 2025-02-21
fix: Filter submission by flowId

DafyddLlyr:
## What's the problem?
Certain flows fail to display submissions, e.g. https://editor.planx.dev/camden/submissions-log?flow=camden-apply-for-a-lawful-development-certificate

Others work fine, e.g. https://editor.planx.dev/tewkesbury/submissions-log?flow=report-a-breach

## What's the cause?
We're currently passing through the flow slug and trying to generate a flow name from this in order to filter. This isn't possible as `slugify()` is a destructive, one-way, process.

For example -
 - `"CAMDEN - Apply for a lawful development certificate"` would become `"camden-apply-for-a-lawful-development-certificate"`
 - `"special CharaCTERS1@@£!!"` becomes `"special-characters1"`

There's no way to reconstruct a flow name from a flow slug, so we need to compare two static values in order to filter.

## What's the solution?
 - Filters the `/${teamSlug}/submissions-log` route by flowId instead of flowSlug

### DIFF
--- a/editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
+++ b/editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
@@ -24,11 +24,12 @@ function EditorNavMenu() {
   const { navigate } = useNavigation();
   const { url } = useCurrentRoute();
   const isRouteLoading = useLoadingRoute();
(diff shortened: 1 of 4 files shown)


jessicamcinchak (approved):
Thanks for fix here! 

As a dev, actually quite happy to have `flowId` exposed for easier trouble-shooting all-around, don't think this will impact wider editor experience :+1:

PR #4366, opened 2025-02-21
feat: Output feature flag list in the editor

ianjon3s:
## What does this PR do?

Idea: show an output of active vs inactive/available feature flags in the editor.


ianjon3s:
Closing in favour of a more UI based solution (to be added to Trello new requests)

PR #4370, opened 2025-02-21
feat: Add env guard middleware

DafyddLlyr:
## What's the problem?
 - Metbase doesn't exist locally, or on Pizzas
 - It's going to get called via Hasura events (see #4334)
 - This will lead to a bunch of errors which we could avoid

## What are the alternatives?
- Running Metabase on pizzas (bad - more CI time for little gain)
- Running Metabase locally, syncing data (also not great - nice to optionally do it, but making it a requirement isn't ideal)
- Ignoring this, letting it fail (maybe!)

## What's the solution?
Very open to ideas here - the current implementation will stop Metabase routes from being accessible locally and on Pizzas.


zz-hh-aa:
Thanks for looking into this! 

One thing I've found useful is being able to hook up my local API to Metabase staging by putting staging vars in my local `env` files. I've been testing calls in development that way (currently using Insomnia). Not sure if this might be bad practice but being able to do this has been useful so far. 

Might there be a way to have the env guard on Pizzas but not necessarily locally?

jamdelion (approved):
Looks easy to reuse too 👍

PR #4371, opened 2025-02-24
fix: show submissions for all flows if no flowId

jamdelion:
A condition was missed off in the submissions filter that meant no submissions were showing if no `flowId` rather than all.

This probably would have been caught if we had test coverage in this area - I'll try to rectify that in the upcoming 'filterable submissions table' PRs.


jessicamcinchak (approved):
Thanks for quick fix!

PR #4378, opened 2025-02-25
feat: add `SearchBox` to Teams page

RODO94:
## What does this PR do?

This PR is for the ticket: https://trello.com/c/XOqpIQic/3013-search-teams

It introduces the `SearchBox` component to the `Teams` table which lists out all of the Team a user can edit and view. The search box only applies to teams a user can **view** not teams that they can **edit**.

To fit with the compact UI of the title, I have added a new prop to `SearchBox` so we can hide the label if it doesn't make contextual sense. To help communicate the use of the text box, I have thus added a `Search` icon from material ui. This also sits in the Service list `Team` page.

### DIFF
--- a/editor.planx.uk/src/pages/Teams.tsx
+++ b/editor.planx.uk/src/pages/Teams.tsx
@@ -14,7 +14,6 @@ import { slugify } from "utils";
 
 import { useStore } from "./FlowEditor/lib/store";
 
-
 interface TeamTheme {
   slug: string;
   primaryColour: string;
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/pages/Teams.tsx
```diff
@@ -56,15 +58,24 @@ const Teams: React.FC<Props> = ({ teams }) => {
     state.createTeam,
   ]);
 
-  const editableTeams: Team[] = [];
-  const viewOnlyTeams: Team[] = [];
+  const [searchedTeams, setSearchedTeams] = useState<Team[] | null>(null);
+  const [clearSearch, setClearSearch] = useState<boolean>(false);
+
+  const viewOnlyTeams = useMemo(
```
RODO94, line 64: This was causing infinite rendering when I passed it into the `SearchBox` component so I had to put it inside a `useMemo`

editor.planx.uk/src/pages/Teams.tsx
```diff
@@ -56,15 +58,24 @@ const Teams: React.FC<Props> = ({ teams }) => {
     state.createTeam,
   ]);
 
-  const editableTeams: Team[] = [];
-  const viewOnlyTeams: Team[] = [];
+  const [searchedTeams, setSearchedTeams] = useState<Team[] | null>(null);
```
RODO94, line 61: This state variable handles the changing list of Teams as you search.

PR #4381, opened 2025-02-25
style: MUI data grid

ianjon3s:
## What does this PR do?

Adds styles to MUI data grid:
- Updates border styles to fit with wider editor styles
- Filter toolbar uses `midGray` background, consistent with flow filters
- Extends our bordered focus styles to focused cells
- Extends our button focus styles to buttons
- Unifies icon sizes across buttons
- Aligns boolean cell content (tick/cross) to top, maintaining horizontal line with service title and other text content
- Bumps size of tick/cross icons to increase visibility
- Unifies selected cell highlight colour
- Adjusts padding of container element to maximise canvas area for table

**Before changes:**
image

**After changes:**
image


jessicamcinchak (approved):
Looks & feels a lot more like PlanX :raised_hands:

PR #4383, opened 2025-02-26
feat: add route for service feedback and refine route for submissions

RODO94:
## What does this PR do?

This PR relates to the ticket here: https://trello.com/c/mBRWXi4l/3246-make-feedback-log-at-the-flow-level-filterable

I have expanded the work of the feedback table sitting at the Team level to now sit at the Service or Flow level as well.

I've replicated a lot of the code and added an extra filter for the `service_slug` coming from the `req.params`.

Follow on work will be done to enable Filtering. There is a current dependency on the data grid migration work happening before filtering can be wired up.

As per a discussion had in a Dev Call, we decided to change the routing of the `submission` page at the flow level to not take the user back to the Team page with a flow in the query params, and rather keep them at the flow level and separate the routes.


RODO94:
[Add Feedback Log to Service Level](https://trello.com/c/mBRWXi4l/3246-add-feedback-log-to-service-level)

editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
```diff
@@ -150,6 +150,12 @@ function EditorNavMenu() {
       route: `/${teamSlug}/submissions-log?flow=${flowId}`,
       accessibleBy: ["platformAdmin", "teamEditor", "demoUser"],
     },
+    {
```
jamdelion, line 153: nit: at the team level the editor nav items have Feedback first, then Submissions second. Could we make it consistent between flow and team level in terms of the ordering?

editor.planx.uk/src/routes/serviceFeedback.tsx
jamdelion, line 1: This file has a lot in common with the existing `routes/feedback.tsx` file. If you have a look at how we did the submissions log hoisting, we used the same route but just with a query param on it. This meant we could have just one line in `routes/team.tsx` - `"/submissions-log": lazy(() => import("./submissionsLog")),`, instead of two which is what you're currently doing for feedback. 

I went through the exact same process, have a look at my commit/PR here for more info and Daf's comments around the same thing: https://github.com/theopensystemslab/planx-new/pull/4305/commits/05113b0c2dc2b62a4040699a061696073674e725

jamdelion (changes requested):
Works well as is but I think we can be more succinct! See my comments about query params. :)

editor.planx.uk/src/routes/serviceFeedback.tsx
RODO94: Aahh yeah I was wondering about this. did you discuss the confusion created from moving the user from flow level to teams level when you do?

https://github.com/user-attachments/assets/e72fb0be-d2f2-4fd5-91cd-bad81061fa3a

It feels like a confusing user journey for me, but have I missed something or a larger gain that we get?

jamdelion: You mean the "sending the user to team level"? Hmm maybe worth a wider discussion on design chat maybe?

RODO94: Yeah, like we make it available at flow level, but really it isn't, it just directs people back to the Team level. I'll do a Slack post

editor.planx.uk/src/routes/serviceFeedback.tsx
```diff
@@ -0,0 +1,63 @@
...
+      const isAuthorised = useStore.getState().canUserEditTeam(teamSlug);
+      if (!isAuthorised)
+        throw new NotFoundError(
+          `User does not have access to ${req.originalUrl}`,
+        );
+
+      const {
+        data: { feedback },
+      } = await client.query<{ feedback: Feedback[] }>({
+        query: gql`
```
jamdelion, line 30: I'm mindful that I'm currently changing this query on my PR, and this file duplicates the query across the two existing route files. Could we extract the query (and maybe some other stuff?) so that it's as easy as possible to change the shared logic without having to know about the multiple places to change it?

RODO94: Ahh yeah good shout, I guess it would need to be considering `flowSlug` as optional since it won't be a condition in the normal submission log? 

@jamdelion would this be a use case for Fragments? 

https://hasura.io/docs/3.0/graphql-api/queries/variables-aliases-fragments-directives/

jamdelion: I was thinking more like https://github.com/theopensystemslab/planx-new/blob/3a576db8dfb98b57a040dc5191eeaefb50706a7c/editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackLog/queries/getFeedbackById.ts#L4

RODO94: Would that remove the need to have different `where` conditions at the Team level and the Flow level?

Team level only needs `teamSlug` but Flow level needs `teamSlug & flowSlug`?

RODO94: I was thinking something like this:

```ts
fragment QueryFragment on feedback_summary {
  address
  createdAt: created_at
  feedbackScore: feedback_score
  flowName: service_name
  id: feedback_id
  nodeTitle: node_title
  nodeType: node_type
  type: feedback_type
  userComment: user_comment
  userContext: user_context
}

query MyQuery {
  feedback_summary(where: {team_slug: {_eq: "teamSlug"}, service_slug: {_eq: "serviceSlug"}}) {
    ...QueryFragment
  }
}

```

jessicamcinchak: As long as the parameter in the query definition is _not_ marked as required (eg no `!`), then GraphQL should simply ignore undefined `variables` passed in and not apply that `where` condition.

Fragments do not apply here because the response shape (fields returned) should be the same/consistent regardless of parameters passed in right?

Eg define query in a single place like: 
```
query GetFeedbackForFlow($teamSlug: String, $flowSlug: String) {
            feedback: feedback_summary(
              order_by: { created_at: desc }
              where: {
                team_slug: { _eq: $teamSlug }
                _and: { service_slug: { _eq: $flowSlug } }
              }
            ) {
              ...
            }
          }
```
Then pass `variables: { teamSlug, flowSlug }` or `variables: { teamSlug }` from respective routes !

editor.planx.uk/src/routes/team.tsx
```diff
@@ -90,6 +90,12 @@ const routes = compose(
 
     "/:flow/service": setFlowAndLazyLoad(() => import("./serviceSettings")),
 
+    "/:flow/feedback": setFlowAndLazyLoad(() => import("./serviceFeedback")),
+
+    "/:flow/submissions-log": setFlowAndLazyLoad(
+      () => import("./serviceSubmissionsLog"),
+    ),
```
jessicamcinchak, line 97: nit: I know we didn't catch this in a previous PR, but now that we've renamed menu item to one-word "submissions" should the route names match? Eg `/:flow/submissions` & `/submissions` and we drop "-log" ?

RODO94: Yeah agreed!

editor.planx.uk/src/routes/feedback.tsx
```diff
@@ -44,21 +45,14 @@ const feedbackRoutes = compose(
         data: { feedback },
       } = await client.query<{ feedback: Feedback[] }>({
         query: gql`
+          ${FEEDBACK_SUMMARY_FIELDS}
+
           query GetFeedbackForTeam($teamSlug: String!) {
             feedback: feedback_summary(
-              order_by: { created_at: asc }
+              order_by: { created_at: desc }
```
RODO94, line 52: Seen this was putting the oldest at the top rather than newest, so this change made sense. Have I missed something here?

jamdelion (approved):

PR #4387, opened 2025-02-26
fix: Submissions table container for height

ianjon3s:
## What does this PR do?

Ensures the submissions table is wrapped in the appropriate `<FixedHeightDashboardContainer>` so that it can inherit the correct height.

Preview:
https://4387.planx.pizza/testing/submissions-log


jamdelion (approved):
Works as expected now, and have fixed forward for the feedback log too, thanks!

PR #4388, opened 2025-02-26
fix: Make data grid cells respond to density

ianjon3s:
## What does this PR do?

Currently the cells of the data grid have a fixed padding, and do not respond to the "density" setting, this PR targets the density classes to set ascending levels of padding.

Density = compact:

Density = standard:

Density = comfortable:


RODO94 (approved):
Looks good to me and it works well on the pizza

PR #4390, opened 2025-02-26
feat: add Archeaological Priority Areas & how-to doc for new PD layers

jessicamcinchak:
Relies on https://github.com/theopensystemslab/planx-core/pull/663

See how-to doc added here for pizza testing notes!

### DIFF
--- a/api.planx.uk/package.json
+++ b/api.planx.uk/package.json
@@ -13,7 +13,7 @@
     "@airbrake/node": "^2.1.8",
     "@aws-sdk/client-s3": "^3.696.0",
     "@aws-sdk/s3-request-presigner": "^3.701.0",
-    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#30cdfa6",
+    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#18a1501",
     "@types/isomorphic-fetch": "^0.0.36",
     "adm-zip": "^0.5.10",
     "axios": "^1.7.4",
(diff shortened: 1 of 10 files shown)


jessicamcinchak:
[Add Archaeological Priority Areas to planning constraints queries & ODP Schema](https://trello.com/c/jhziZIKP/3239-add-archaeological-priority-areas-to-planning-constraints-queries-odp-schema)

ianjon3s (approved):
- ✅ APA selected by default in planning constraints
- ✅ Intersecting property in Barnet EN5 5UU shows APA applicable (updated in spreadsheet)
- ✅ Non-intersecting property shows as "does not apply"

One comment about the output (image below), there is a leading "p" character in the output from planning data, is this due to an error in our formatting?

image

jamdelion (approved):
Thanks for how-to doc! 

One question I had was: how easy is it to find an address that is affected by a particular constraint?

jessicamcinchak:
Thanks both ! 

- @ianjon3s that misformatted description unfortunately comes right from Planning Data's `name` property - we can raise to them to see if they can trim/tidy up, but ideally not something we want to get into the business of special-formatting on our end only
- @jamdelion Planning Data maintains a map of all constraints which is a handy tool for cross-checking addresses or I usually just google ! Eg in this case see https://www.planning.data.gov.uk/map/?dataset=archaeological-priority-area, then pick a POI from the basemap within the relevant zone (a landmark or random shop etc), then put its' name into google to get the postcode & address?

PR #4399, opened 2025-02-28
refactor: Remove feedback type icons from log

ianjon3s:
## What does this PR do?

Removes feedback type icons from feedback log.


jessicamcinchak (approved):
I think the less stylistic, more data-forward approach is right with the new data grid styles, thanks for removing icons throughout for consistency :+1:

PR #4402, opened 2025-02-28
feat: Set up filter options for feedback type

ianjon3s:
## What does this PR do?

Sets up a `feedbackFilterOptions` file in order to filter by feedback type in the feedback log.

### DIFF
--- a/editor.planx.uk/src/ui/shared/DataTable/types.ts
+++ b/editor.planx.uk/src/ui/shared/DataTable/types.ts
@@ -7,7 +7,7 @@ import {
 
 export const ColumnFilterType = {
   BOOLEAN: "boolean",
-  ARRAY: "array", // when the column can be filtered from a range of known values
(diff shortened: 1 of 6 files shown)


editor.planx.uk/src/ui/shared/DataTable/types.ts
```diff
@@ -7,7 +7,7 @@ import {
 
 export const ColumnFilterType = {
   BOOLEAN: "boolean",
-  ARRAY: "array", // when the column can be filtered from a range of known values
+  ARRAY: "singleSelect", // when the column can be filtered from a range of known values
```
DafyddLlyr, line 10: Docs: https://mui.com/x/react-data-grid/column-definition/#special-properties

ianjon3s:
Thanks for this @DafyddLlyr, all working as expected (including existing `ARRAY` filters).

DafyddLlyr (approved):

PR #4407, opened 2025-02-28
style: Refine data table column widths based on content

ianjon3s:
## What does this PR do?

- Makes first column "Service" text bold
- Refines column widths based on content to make space optimal (can be refined as we continue to use the logs)

**Feedback log before changes:**
image

**Feedback log after changes:**
image

**Submissions log before changes:**
image

**Submissions log after changes:**
image


jessicamcinchak (approved):
Bold left-most column is a nice compromise if it can't be "sticky" - helpful visual orientation :+1:

PR #4408, opened 2025-02-28
fix: Check for valid JSON before outputting in submissions log

ianjon3s:
## What does this PR do?

Possible temp fix to prevent errors in the submissions log caused by invalid JSON.


jessicamcinchak:
Thanks again for queueing this one up - but think we can close in favor of #4425 now !

PR #4411, opened 2025-02-28
[application] simplify process for passing correct DB URLs to images

freemvmt:
Postgres databases are accessed using a URL making use of the `postgres://` protocol, as per [this spec](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS) (note that [URLs are a subset of URIs](https://medium.com/tuanhdotnet/secrets-of-uris-urls-and-urns-understanding-their-differences-and-uses-b2b791af7323)).

We have one postgres server, hosted on AWS RDS. Connecting to the staging instance via pgAdmin, I can see the databases therein:

image

Metabase accesses the `metabase` database via a role called `metabase`, while Sharedb and Hasura access the default `postgres` database via the `dbuser` role. As such there are 2 different URLs that need to be baked into the respective images at build time.  We are currently storing these as Pulumi secrets called `db-url` and `md-db-connection-uri`.

This means that if we have to restore the db, and it's name changes as a result (as happened when I ran the test I describe in #4392), we have to change several Pulumi inputs in order to ensure that all constituent parts of the application layer are pointing at the new db.

However, these URLs are composed of predictable parts (see the new `helper.ts` file), and the only part that changes when the db name is cycled, is the AWS URI, which we store as a Pulumi secret under `db-host`.

Therefore, this PR proposes changes which will enable us to repair the application layer simply by rotating `db-host`.


DafyddLlyr (approved):
Thanks for the clear documentation and tidy up here 👍

PR #4416, opened 2025-03-03
feat: Close WS connections on JWT expiry

DafyddLlyr:
## What's the problem?
The WebSocket connection to ShareDB only verifies the JWT on initial connection, and not on every message. This means that an expired token can write to the operations table as long as it was valid when the WS connection was opened.

This was identified here - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1740754874553229 (OSL Slack). Once a token expires, `ReconnectingWebsocket` closes, and then fails to re-open a connection silently, without any feedback for users.

## What's the solution?
Check the token every minute, catching any errors in the frontend and then logging out the user.

We could do this on each message, but this is probably overkill?

## To test
- Change JWT expiry to 1 or 2 minutes locally
- Open a flow, wait for expiry, get logged out automatically

### DIFF
--- a/editor.planx.uk/src/lib/graphql.ts
+++ b/editor.planx.uk/src/lib/graphql.ts
@@ -73,6 +73,9 @@ const authWsLink = new WebSocketLink({
         },
       };
     },
+    connectionCallback: (error) => {
+      if (error) handleExpiredJWTErrors();
+    },
   },
 });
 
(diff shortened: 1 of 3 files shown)


jessicamcinchak (approved):
Thanks for getting to the bottom of this one, works as expected for me !

PR #4421, opened 2025-03-04
demo: sync analytics to pizza for testing data migrations

jessicamcinchak:
See https://opensystemslab.slack.com/archives/C01E3AC0C03/p1741016799482609

An alternative to running locally, but easier/less risky for experimentation than staging !


jessicamcinchak:
Closing this one - not necessary to property.type migration any longer

PR #4422, opened 2025-03-04
fix: Don't display `GetStarted` card in empty internal portals

DafyddLlyr:
This PR enhances the `showGetStarted` check to make sure that the `GetStarted` card is only displayed once at the flow root.

| Before | After |
|--------|--------|
| image | image |


ianjon3s (approved):
Tested at multiple internal portal levels and working as expected

PR #4423, opened 2025-03-04
fix: False positive ITP fee breakdown error

DafyddLlyr:
## What's the problem?
Airbrake is logging a fee breakdown error for each ITP session. However, as we already fallback to the `inviteToPayFeeBreakdown` passed into the component, the payee is presented with a valid and correct fee breakdown table.

Please see https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1740579357423249 (OSL Slack) for context.

## What's the solution?
As hooks cannot be called conditionally, this is currently written to check for the presence of the `inviteToPayFeeBreakdown` after the hook is called (and an Airbrake error has been called).

This PR checks the `inviteToPayFeeBreakdown` as part of the hook, meaning that the Airbrake error will not be triggered for ITP payments.

## To test
Please try to resume the following ITP payment (devops email address) - https://4423.planx.pizza/buckinghamshire/apply-for-a-lawful-development-certificate/pay?paymentRequestId=d8955063-1310-4ad8-9eed-7d86d27c9b5c

image

A fee breakdown is displayed, and an Airbrake error is not thrown.


jessicamcinchak (approved):
Perfect , thanks !

PR #4424, opened 2025-03-04
fix: Display timestamps correctly in 24hr format

DafyddLlyr:
## What does this PR do?
- Adds a helper function to standardise datetime outputs - `dateFormatter()`
- Applies this in feedback and submission tables

### DIFF
--- a/editor.planx.uk/src/ui/shared/DataTable/utils.tsx
+++ b/editor.planx.uk/src/ui/shared/DataTable/utils.tsx
@@ -5,6 +5,7 @@ import {
   GridValueOptionsParams,
   ValueOptions,
 } from "@mui/x-data-grid";
+import { format } from "date-fns";
 import capitalize from "lodash/capitalize";
 import React from "react";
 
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/ui/shared/DataTable/utils.tsx
```diff
@@ -111,3 +112,8 @@ export const getColumnFilterType = (columnType?: ColumnRenderType) => {
       return undefined;
   }
 };
+
+/**
+ * Format date times to a standard format, e.g 31/12/25 14:40:45
+ */
+export const dateFormatter = (value: string) => format(new Date(value), "dd/MM/yy HH:mm:ss");
```
DafyddLlyr, line 119: There is GDS guidance on date and time formatting, but I'd consider these rules we should apply in applicant-facing situations, not internally.

Elsewhere in PlanX I think we just display time as "13 hours ago" etc - if there are other instances we need to standardise please point them out.

https://www.gov.uk/guidance/style-guide/a-to-z#times

https://www.gov.uk/guidance/style-guide/a-to-z#dates

ianjon3s (approved):
Works as expected and the standardised format helper function is a useful addition

jessicamcinchak (approved):
Thanks for getting to bottom of this one!

PR #4425, opened 2025-03-04
fix: JSONB response error boundaries in submissions table

DafyddLlyr:
## What's the problem?
The submissions table throws an error when parsing values from `submissions_summary_log.response`. This causes the entire page to error. 

The fundamental issue here is that we're displaying HTTP responses which we do not control the format of. We could do some work to ensure a more consistent format along the way, but I'd argue there's a lot of value in keeping the raw HTTP response. We might find in future that this isn't that useful to display at this level and maybe just the status is good enough.

## What's the solution?
A few things implemented here!
 - Add error boundary at the cell level
 - Add error boundary at the DataGrid / page level
 - Added Storybook and mock data so we can test this manually more easily.

I spent a bit of time coming up with a solution for all of our response shapes without much luck. Grabbing a higher level (e.g. `response.data`) just gives a poorly formed output.

Using the following query on production I can see just a small number of responses that would trigger the `InvalidJSONError` component - 

```graphql
query GetAllNonSuccessNonPaymentLogs {
  submission_services_log(where: {_or: {status: {_neq: "Success"}, event_type: {_neq: "Pay"}}}) {
    response(path: "data.message")
  }
}
```


editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/Submissions.tsx
DafyddLlyr, line 1: Fixed imports here also.

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/components/FormattedResponse.tsx
```diff
@@ -14,20 +16,23 @@ const Response = styled(Box)(({ theme }) => ({
...
 
+const InvalidJSONError = () => <Typography>Error parsing response</Typography>
+
 export const FormattedResponse: React.FC<Submission> = (submission) => {
-  return submission.eventType === "Pay" ? (
-    <Response component="pre">
-      {JSON.stringify(submission.response, null, 2)}
-    </Response>
-  ) : (
+  const getResponse = ({ eventType, status, response }: Submission) => {
```
DafyddLlyr, line 22: Consolidated this logic to a single function instead of splitting it over multiple terneries within the template.

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/mockSubmissions.ts
DafyddLlyr, line 1: This wasn't being used (as far as I can tell) so I've just updated it with a few illustrative examples. Certainly some room here for expanding this. If there's some context for this file I'm missing I'm happy to roll back the changes here.

jessicamcinchak: :+1: Your changes seem fine here - previous context/comment on this here: https://github.com/theopensystemslab/planx-new/pull/4362#discussion_r1969739166

jessicamcinchak (approved):
Works as expected, thank you!

PR #4432, opened 2025-03-05
fix: Correct background colour for tooltips

ianjon3s:


DafyddLlyr (approved):
Thanks! Also just noticed this one.

PR #4433, opened 2025-03-05
fix: ensure `SetValue` "removeOne" operation preserves array types

jessicamcinchak:
See `property.type` bugs reported via Jonny https://opensystemslab.slack.com/archives/C01E3AC0C03/p1741172093172059 & https://opensystemslab.slack.com/archives/C01E3AC0C03/p1741171167791389

**What was happening:**
- SetValue appended a value => passport `{ "property.type": ["residential.house.detached"] }`
- SetValue removed one value (not an applicable value) => passport `{ "property.type": "residential.house.detached" }` 
  - !! Changed type from array to string !!
  - This type change both breaks future automations _and_ causes single letter, rather than first item, to be displayed in "About the property" summary table
- Question/Checklist could not be auto-answered because passport is now string, not array type

Simple fix & test cases cover all possible scenarios now!

This flow is isolated before & after example: https://editor.planx.uk/testing/set-value-subsequent-automations

### DIFF
--- a/editor.planx.uk/src/@planx/components/SetValue/utils.test.ts
+++ b/editor.planx.uk/src/@planx/components/SetValue/utils.test.ts
@@ -84,6 +84,7 @@ describe("calculateNewValues() helper function", () => {
         expected: ["bear", "dog", "monkey"],
       },
       { previous: ["bear", "dog", "lion"], expected: ["bear", "dog"] },
+      { previous: ["bear"], expected: ["bear"]},
     ])(
       "input of $previous sets passport value to be $expected",
       ({ previous, expected }) => {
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/SetValue/utils.test.ts
```diff
@@ -84,6 +84,7 @@ describe("calculateNewValues() helper function", () => {
         expected: ["bear", "dog", "monkey"],
       },
       { previous: ["bear", "dog", "lion"], expected: ["bear", "dog"] },
+      { previous: ["bear"], expected: ["bear"]},
```
jessicamcinchak, line 87: Previously, `["bear"]` was resulting in `"bear"` and falling under the same conditions as `"panda"` to `"panda"` above (because `previous` values get auto-formatted as an array)

DafyddLlyr (approved):
Thanks for the detective work here - PR comments and updated tests super clear 👍

PR #4440, opened 2025-03-06
feat: Filter admin panel teams by A4 status

DafyddLlyr:
## What's the problem?
 - MUI datagrid does not (and will not) support filter and sort options for asynchronous data - please see https://github.com/mui/mui-x/issues/5705
 - This means it's not possible to sort and filter on A4 status, as it's an additional API request (after the grid is rendered)

## What's the solution?
 - Tracks A4 status via the DB, not via an object in the API
   - Adds `team_settings.has_article4_schema` column
   - Updates `teams_summary` view
 - Reads new column in frontend, drop custom component for plain boolean datagrid column
 - Drops unused `/gis/:localAuthority/article4-schema` REST API route

https://github.com/user-attachments/assets/e5946e3f-34ea-4838-bee0-24ae99149ed2


DafyddLlyr:
[Global admin dashboard shows council/service status and integrations set up](https://trello.com/c/ZZxX4MAk/3162-global-admin-dashboard-shows-council-service-status-and-integrations-set-up)

api.planx.uk/modules/gis/service/digitalLand.ts
```diff
@@ -36,7 +36,10 @@ export interface LocalAuthorityMetadata {
   };
 }
 
-/** When a team publishes their granular Article 4 data, add them to this list. Key must match team slug */
+/**
+ * When a team publishes their granular Article 4 data, add them to this list. Key must match team slug
+ * The database column team_setting.has_article4_schema also needs to be updated via the Hasura console
+ */
```
jessicamcinchak, line 42: :+1:

api.planx.uk/modules/gis/routes.ts
```diff
@@ -1,12 +1,10 @@
 import { Router } from "express";
 import { locationSearch } from "./service/index.js";
-import { hasArticle4Schema } from "./service/article4Schema.js";
 import { classifiedRoadsSearch } from "./service/classifiedRoads.js";
 
 const router = Router();
 
 router.get("/gis/:localAuthority", locationSearch);
-router.get("/gis/:localAuthority/article4-schema", hasArticle4Schema);
```
jessicamcinchak, line 9: Would argue there's maybe a future, separate use-case of this endpoint for suggesting council-specific data field autocomplete values in the editor at some point (right now `property.constraints.planning` only suggests the schema-typed data fields, but there's no reason an editor modal couldn't concatenate granular team-specific values when applicable?)

All to say should we keep it around as a utility endpoint, or just re-introduce it in the future if we need?

jessicamcinchak (approved):
Pragmatic solution, thanks for sorting !

api.planx.uk/modules/gis/routes.ts
DafyddLlyr: Sorry - meant to come back to this comment!

Yep it's a very fair question. I'd say as it's currently just returning a `true`/`false` we can remove for now - but I can certainly see a future use case for getting this list as you suggest.

PR #4441, opened 2025-03-06
fix: Prevent checklists wrapping in constraint overrides

ianjon3s:
## What does this PR do?

(Reported by August in Slack)
In the modal for contesting constraints the checklist rows do not sit at 100% width this is due to them being placed inside a wrapping `<Grid>` but not being assigned as being `<Grid item>`s, introducing this fixes the issue.

**Before change:**
image

**After change:**
image


jessicamcinchak (approved):
Thanks !

PR #4442, opened 2025-03-06
feat: Load submissions response in modal

ianjon3s:
## What does this PR do?

Updates submissions data grid to load JSON responses in a modal. The modal is kept generic (`DataTableModal.tsx`) so that it can be used to display other fields where necessary.

**Testing:**
https://4442.planx.pizza/a-new-team/submissions

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/components/EventsLog.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/components/EventsLog.tsx
@@ -15,6 +15,7 @@ import {
 import { EventsLogProps, Submission } from "../types";
 import { DownloadSubmissionButton } from "./DownloadSubmissionButton";
 import { FormattedResponse } from "./FormattedResponse";
+import { OpenModalButton } from "./OpenModalButton";
 import { StatusChip } from "./StatusChip";
 import { SubmissionEvent } from "./SubmissionEvent";
 
@@ -86,7 +87,7 @@ const EventsLog: React.FC<EventsLogProps> = ({
       headerName: "Date",
       width: 125,
       columnOptions: {
-        valueFormatter: dateFormatter
+        valueFormatter: dateFormatter,
       },
       type: ColumnFilterType.DATE,
     },
@@ -94,9 +95,9 @@ const EventsLog: React.FC<EventsLogProps> = ({
     {
       field: "response",
       headerName: "Response",
-      width: 450,
+      width: 100,
       type: ColumnFilterType.CUSTOM,
-      customComponent: (params) => <FormattedResponse {...params.row} />,
+      customComponent: OpenModalButton,
       columnOptions: {
         sortable: false,
         filterOperators: [
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/components/EventsLog.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/components/EventsLog.tsx
@@ -14,8 +14,7 @@ import {
 } from "../submissionFilterOptions";
 import { EventsLogProps, Submission } from "../types";
 import { DownloadSubmissionButton } from "./DownloadSubmissionButton";
-import { FormattedResponse } from "./FormattedResponse";
-import { OpenModalButton } from "./OpenModalButton";
+import { OpenResponseButton } from "./OpenResponseButton";
 import { StatusChip } from "./StatusChip";
 import { SubmissionEvent } from "./SubmissionEvent";
 
@@ -91,15 +90,16 @@ const EventsLog: React.FC<EventsLogProps> = ({
       },
       type: ColumnFilterType.DATE,
     },
-    { field: "sessionId", headerName: "Session ID", width: 200 },
+    { field: "sessionId", headerName: "Session ID", width: 400 },
     {
       field: "response",
       headerName: "Response",
       width: 100,
       type: ColumnFilterType.CUSTOM,
-      customComponent: OpenModalButton,
+      customComponent: OpenResponseButton,
       columnOptions: {
         sortable: false,
+        cellClassName: "MuiDataGrid-cell--textCenter",
         filterOperators: [
           {
             value: "contains",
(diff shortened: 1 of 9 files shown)


ianjon3s:
[Make submission log at the team level (filterable)](https://trello.com/c/tuVo5aIF/3149-make-submission-log-at-the-team-level-filterable)

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/components/OpenResponseButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+    }
+  };
+
+  return (
+    <>
+      <Button variant="text" size="small" onClick={handleButtonClick}>
+        View
+      </Button>
+      <DataTableModal
+        title="JSON response"
```
jessicamcinchak, line 44: nit: I wonder if we can avoid using "JSON" language with end-users - perhaps just "Response" or "Respose for [sessionId]" ??

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/components/OpenResponseButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+      }
+      setResponse(fetchedData);
+    }
+  };
+
+  return (
+    <>
+      <Button variant="text" size="small" onClick={handleButtonClick}>
+        View
+      </Button>
```
jessicamcinchak, line 42: nit: we're using a text button for the modal, but an icon button for the adjacent "Download" column - should we perhaps standardise? 

Screenshot from 2025-03-07 14-29-01

Don't necessarily have a preference either way. Maybe slightly in favor of less text & more visual language using something like the `[Launch](https://mui.com/material-ui/material-icons/?srsltid=AfmBOor_h8WrjP1XiZBWFMl6CiQz69B5bodPXRz0_56rYmKlPRYNnb4A&query=open&selected=Launch)` icon for "view"?

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/components/FormattedResponse.tsx
```diff
@@ -1,41 +1,49 @@
...
-        <Response response={response} />
+        <ReactJson
+          src={parsedResponse}
+          theme="monokai"
+          collapsed={2}
+          displayDataTypes={false}
+          indentWidth={2}
+          style={{ background: "transparent" }}
+          enableClipboard={false}
+        />
```
jessicamcinchak, line 46: Nice to see this becoming a re-used "pattern" :heavy_plus_sign:

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/components/EventsLog.tsx
```diff
@@ -86,19 +86,20 @@ const EventsLog: React.FC<EventsLogProps> = ({
...
       },
       type: ColumnFilterType.DATE,
     },
-    { field: "sessionId", headerName: "Session ID", width: 200 },
+    { field: "sessionId", headerName: "Session ID", width: 400 },
     {
       field: "response",
       headerName: "Response",
-      width: 450,
+      width: 100,
```
jessicamcinchak, line 97: nit: I'm still find the empty white space to the right of the last column a bit annoying & feels like "missing data" ?? Is there a straightforward way for columns to fill the viewport width (if their combined set `width` are smaller?) ? 

Or, at minimum, is it possible to remove the right border from the final column header? So then it'd look more like Download is the final, full remaining width column ? 

Screenshot from 2025-03-07 14-29-08

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/components/OpenResponseButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+    if (eventType === "Pay") return response;
+    if (status === "Success") return response?.data?.body;
+
+    return response?.data?.message;
+  };
+
+  const handleButtonClick = () => {
+    setModalIsOpen(true);
+    if (!response) {
+      let fetchedData = getResponse(row);
```
jessicamcinchak, line 25: nit: `parsedData` would be a bit more clear here than `fetchedData` for me I think ?

jessicamcinchak (approved):
Looking really tidy ! A few notes to consider below, no show-stoppers !

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/components/OpenResponseButton.tsx
ianjon3s: Good to be consistent with non-information "actions", I've brought in the `<IconButton>`.

`Launch` icon has been fairly universally adopted as "open in new tab", so I've gone with [Preview](https://mui.com/material-ui/material-icons/?query=preview&selected=Preview) as it has a very literal eye-in-a-modal

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/components/OpenResponseButton.tsx
ianjon3s: Agreed, have updated

editor.planx.uk/src/pages/FlowEditor/components/Settings/Submissions/components/EventsLog.tsx
ianjon3s: There are options for expanding/responsive columns in the grid but I think we should avoid going the full-width expanding route, mostly due to the variance in screen sizes, things will get very stretched out on a 27" monitor.

An option would be to make the last column expand, but this would push the column menu to the right-hand-side of  the screen, and also only apply to a predefined final column, rather than considering any hidden columns (options are scoped to a particular column, rather than the rendered view at a given time).

I've hidden the border for the final column, seems to help to visually tie together the table a bit (the menu and column resize are still accessible in the header).

PR #4447, opened 2025-03-11
fix: Constraints override checklist structure

ianjon3s:
## What does this PR do?

Reported by August here: https://opensystemslab.slack.com/archives/C01E3AC0C03/p1741597807847749?thread_ts=1741255799.707409&cid=C01E3AC0C03

Currently the constraints override checklist is incorrectly wrapped in a `<label>` which is automatically associated with the first `<input>`, meaning the first input (checkbox) is actioned when clicking anywhere in the element (except other checklist items. Removing the wrapping label and unhiding the repeated `<legend>` rectifies this.

The `<textarea>` "Tell us why" was also not associated with its `<label>` (incorrectly associating with `name` rather than `id` as a ref), I've also fixed this by adding an `id` property.

**Testing:**
https://4447.planx.pizza/lambeth/constraints-test/published?analytics=false
Multiple TPO intersections at SE1 6HZ (Imperial War Museum)

### DIFF
--- a/editor.planx.uk/src/@planx/components/PlanningConstraints/Modal.tsx
+++ b/editor.planx.uk/src/@planx/components/PlanningConstraints/Modal.tsx
@@ -178,38 +178,39 @@ export const OverrideEntitiesModal = ({
           </Typography>
           <Divider sx={{ marginY: 2 }} />
           <Box marginBottom={2}>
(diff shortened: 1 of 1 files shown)


jessicamcinchak (approved):
Thanks for this fix !

PR #4454, opened 2025-03-12
Add filter test for platform admin panel

jamdelion:
It felt important to get test coverage for the filtering in the admin panel to avoid future regressions, but haven't managed to get it to work yet!


jessicamcinchak:
Chosing to close this one for now since it was never quite working & come back to in the future !

PR #4457, opened 2025-03-12
feat: migration for adding `is_trial_team` to teams table 

RODO94:
### What does this PR do? 

This PR adds a column to the `teams` table to control the access level of a team. It will be used to create a _**Trial Team**_ as specified in the Ticket here: https://trello.com/c/J7ipkEsU/3126-create-a-trial-account-in-the-editor

The PR adds the column and the necessary permissions to provide parity with the current database, while restricting update rights of `access_rights` to only `platformAdmins`.

The testing looks to cover:

- A teamEditor in a normal team can still update flow status 
- A teamEditor of a trial team cannot update the status of their flow
- A platformAdmin can still update the status of the trial team flows 

Once this PR is merged and on main, I will update the `seed-database` team file


hasura.planx.uk/migrations/1741771336243_alter_table_public_teams_add_column_access_rights/down.sql
```diff
@@ -0,0 +1,5 @@
+alter table "public"."teams" drop constraint "teams_access_rights_fkey";
+
+alter table "public"."teams" drop column "access_rights";
+
+DROP TABLE "public"."team_access_rights_enum";
```
RODO94, line 5: Dropping enum table comes at the end so it is not blocked by dropping a table which has dependents

hasura.planx.uk/metadata/tables.yaml
RODO94, line 1: I did wonder whether adding an enum table was the right approach here, but I felt it followed similar patterns with `flow.status` and `user.roles` and is something we will need to be accurate and static.

jessicamcinchak: I do not agree that an enum is a requirement here. Why introduce extra language like "full access" if we only have a singular outward concept of trial team or not trial team - would personally find `isTrialAccount` (boolean type; false by default) much more clear here

jessicamcinchak (changes requested):

e2e/tests/api-driven/src/demo-workspace/helper.ts
RODO94, line 1: There were shared methods and types with trial workspace, so I moved these to the global helpers to share

e2e/tests/api-driven/src/trial-workspace/helpers.ts
```diff
@@ -0,0 +1,163 @@
...
+            ) {
+              id
+            }
+          }
+        `,
+        {
+          name: teamData.name,
+          id: Number(teamData.id),
+          slug: teamData.slug,
+          is_trial_team: teamData.is_trial_team === "true" ? true : false,
```
RODO94, line 69: the `datatable` is of type `Record<string, string>` so I need to convert this to the right Boolean

PR #4464, opened 2025-03-14
chore: Add input validation to `templated_flow_edits` table

DafyddLlyr:
## What does this PR do?
 - Adds input validation for `INSERT` and `UPDATE` operations into the `templated_flow_edits` table
 - The `node_data` (JSONB) column could contain HTML which needs to be sanitised
 - This uses the existing `/webhooks/hasura/validate-input/jsonb/clean-html` endpoint
 
 Spotted whilst doing pre-pen test checks of newly added tables / columns / permissions!

### DIFF
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -2910,6 +2910,15 @@
           - updated_at
           - flow_id
           - id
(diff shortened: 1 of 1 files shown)


jessicamcinchak (approved):
Great spot thanks !

PR #4466, opened 2025-03-14
chore: Don't sync `team_integrations.power_automate_webhook_url`

DafyddLlyr:
## What?
Removes `power_automate_webhook_url` from data sync scripts

## Why?
In some specific testing cases, council may have a staging power automate URL. This has happened recently with Doncaster. If this becomes a recurring pattern we can address with with `_staging` and `_production` column as other data values are handled.

### DIFF
--- a/scripts/seed-database/container.sh
+++ b/scripts/seed-database/container.sh
@@ -48,7 +48,7 @@ done
 
 # Copy subset of team_integrations columns
 # Do not copy production values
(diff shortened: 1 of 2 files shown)


jessicamcinchak (commented):
Since this table isn't a `select *` when synced, also expecting to see a field removed from `container.sh` here ! 

Thanks for picking this up!

jessicamcinchak (approved):

PR #4474, opened 2025-03-17
fix: Setup internal DNS for Fargate (take 2)

DafyddLlyr:
Follow up to #4473 

This now more correctly mimics the details required via the console. This should allow Hasura to connect directly to the API Fargate service without exiting the VPC and getting blocked by Cloudflare.

Changes to network layer already manually run for staging stack via Pulumi CLI.

image


DafyddLlyr:
Force merged in order to test this.

PR #4445, opened 2025-03-10
feat: Auth proxy for Hasura

DafyddLlyr:
## What does this PR do?
- Updates Caddy to v2.9.1
- Adds check for auth, if present, redirects to REST API (see logic below)
- Adds a header to show validation status

### To test
It might be nice to add a little script / small test suite to cover this in future.

For the following commands, I get a 200 with the `x-auth-validated: true` header on the response -

```sh
curl -v -H "Authorization: Bearer ${VALID_TOKEN_HERE}" https://hasura.4445.planx.pizza
```
```sh
curl -v -b "jwt=${VALID_TOKEN_HERE}" https://hasura.4445.planx.pizza
```
```sh
curl -v "https://hasura.4445.planx.pizza?token=${VALID_TOKEN_HERE}"
```

For the following commands, I get a 401 with the `x-auth-validated: false` header on the response -

```sh
curl -v -H "Authorization: Bearer ${REVOKED_TOKEN_HERE}" https://hasura.4445.planx.pizza
```
```sh
curl -v -b "jwt=${REVOKED_TOKEN_HERE}" https://hasura.4445.planx.pizza
```
```sh
curl -v "https://hasura.4445.planx.pizza?token=${REVOKED_TOKEN_HERE}"
```

(In order to revoke a token, please log in and take a copy, then logout)

For the following commands, I get a 200 with the `x-auth-validated: skip` header on the response -

```sh
curl -v https://hasura.4445.planx.pizza
```
```sh
curl -v -b "notJWT=randomCookie" https://hasura.4445.planx.pizza
```
```sh
curl -v "https://hasura.4445.planx.pizza?notToken=randomQueryString"
```

```mermaid
flowchart TD
    A[Add security headers to response] --> B{Does request need auth?};
    B -- Yes --> C[Hit REST API validation endpoint];
    C -- 2xx --> D[Forward to Hasura];
    C -- 4xx, 5xx --> E[401 Unauthorized - expired or invalid token];
    B -- No --> D;
```

## To Do...
- [x] Tests for new endpoint - https://github.com/theopensystemslab/planx-new/pull/4469
- [x] Docs for new endpoint - https://github.com/theopensystemslab/planx-new/pull/4468

PR #4487, opened 2025-03-19
feat: Hasura auth proxy using Cloudflare snippets

DafyddLlyr:
## What does this PR do?
 - Documents our Cloudflare snippet setup which handles additional Hasura auth (checking for block-listed token)
 - This replaces the approach originally implemented in #4445 
 - Updates the JWT endpoint to also verify JWTs (required for upcoming ShareDB PR, makes more sense to check all in one place)
 
 For more details, please see included README.md

## To do
- Handle Hasura console
  - Requests made via the console will fail if you're logged out of PlanX
  - Checking for origin or referer headers is insecure as these headers can be spoofed (thus bypassing the revoked token check)
  - Is it an issue that we need to be logged into PlanX for this to work?
 - Add back the logic for `x-auth-validated` header as a breadcrumb

## Testing
The same applies here as did in #4445 (currently minus the headers). It breaks down to -
 - I can log in / log out of staging without issue
 - Once logged in, I can make requests to Hasura (e.g. via `curl`)
 - Once logged out, requests to Hasura will fail with a `401` as they're intercepted by the CF snippet
 - Public requests to Hasura work (e.g. published, public-facing, routes)

### DIFF
--- a/infrastructure/application/cloudflare/snippets/validate_jwt.js
+++ b/infrastructure/application/cloudflare/snippets/validate_jwt.js
@@ -8,7 +8,7 @@ export default {
       if (authHeaderJWT) return authHeaderJWT;
 
       const cookies = request.headers.get("cookie");
-      const cookieJWT = cookies?.match(/(?:^|;\s*)jwt=([^;]*)(?:;|$)/)?.[1];
+      const cookieJWT = cookies?.match(/(?:^|;\s*)jwt=([^;]+)(?:;|$)/)?.[1];
       if (cookieJWT) return cookieJWT;
     };
 
--- a/infrastructure/application/cloudflare/snippets/validate_jwt.js
+++ b/infrastructure/application/cloudflare/snippets/validate_jwt.js
@@ -1,0 +1,41 @@
+export default {
+  async fetch(request) {
+    const API_URL_EXT = "https://api.editor.planx.dev";
+
+    const getJWT = (request) => {
+      const authHeader = request.headers.get("authorization");
+      const authHeaderJWT = authHeader?.match(/^Bearer (\S+)$/)?.[1];
+      if (authHeaderJWT) return authHeaderJWT;
+
+      const cookies = request.headers.get("cookie");
+      const cookieJWT = cookies?.match(/(?:^|;\s*)jwt=([^;]*)(?:;|$)/)?.[1];
+      if (cookieJWT) return cookieJWT;
+    };
+
+    const validateJWT = async (request) => {
+      const response = await fetch(`${API_URL_EXT}/auth/validate-jwt`, {
+        method: "GET",
+        headers: request.headers
+      });
+
+      return response.ok
+    };
+
+    try {
+      const jwt = getJWT(request);
+
+      // Forward request directly to Hasura
+      // Non-JWT requests (admin and public) are validated internally
+      if (!jwt) return fetch(request);
+
+      // Requests with a JWT need additional validation via our REST API
+      // We check if the token is valid (signed by PlanX) or revoked (user has logged out)
+      const isValidToken = await validateJWT(request)
+      if (isValidToken) return fetch(request);
+
+      return new Response(`Invalid or revoked token`, { status: 401 });
+    } catch (error) {
+      return new Response("Error validating token: " + error.message, { status: 500 })
+    }
+  }
+};
--- a/api.planx.uk/modules/auth/controller.ts
+++ b/api.planx.uk/modules/auth/controller.ts
@@ -1,3 +1,4 @@
+import jwt from "jsonwebtoken";
 import type { CookieOptions, RequestHandler, Response } from "express";
 import type { Request } from "express-jwt";
 import {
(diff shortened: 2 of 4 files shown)


infrastructure/application/cloudflare/snippets/validate_jwt.js
DafyddLlyr, line 1: This is currently deployed and active on the staging environment.

image

infrastructure/application/cloudflare/snippets/validate_jwt.js
```diff
@@ -0,0 +1,41 @@
+export default {
+  async fetch(request) {
+    const API_URL_EXT = "https://api.editor.planx.dev";
+
+    const getJWT = (request) => {
+      const authHeader = request.headers.get("authorization");
+      const authHeaderJWT = authHeader?.match(/^Bearer (\S+)$/)?.[1];
```
freemvmt, line 7: `\S` is the negation of `\s` (i.e. not whitespace)? I think we are expecting a JWT token to be passed here, i.e. three URL-safe base 64 strings separated by dots? So we will definitely capture it but it would generously match other stuff too? so I wonder if we should be more specific, e.g. something like: 

```regex
^Bearer ((?:[A-Za-z0-9-_]+\.?){3})$
```

see [here](https://stackoverflow.com/questions/61802832/regex-to-match-jwt) for inspo. might be overkill tho!

infrastructure/application/cloudflare/snippets/validate_jwt.js
```diff
@@ -0,0 +1,41 @@
...
+      if (authHeaderJWT) return authHeaderJWT;
+
+      const cookies = request.headers.get("cookie");
+      const cookieJWT = cookies?.match(/(?:^|;\s*)jwt=([^;]*)(?:;|$)/)?.[1];
```
freemvmt, line 11:
```suggestion
      const cookieJWT = cookies?.match(/(?:^|;\s*)jwt=([^;]+)(?:;|$)/)?.[1];
```

maybe use `+` here instead since we need a non-zero length string?

freemvmt (approved):
Looks good to me, great solution ✨ 

Re the code itself, only some superfluous points about regex, which is just cause I'm a regex nerd and always have opinions, rather than cause any changes are actually needed 😅 

On a higher level, one thought - looking at the Cloudflare [docs](https://developers.cloudflare.com/rules/snippets/#limits), are we are confident we will always come in under the 5ms timeout?

image

e.g. if the Hasura request is pulling down an especially large flow, or the Hasura server is under greater load than usual and therefore just responding slowly, could we trip this limit 🤔

Obviously the Hasura proxy is not subject to this constraint, so it's something to consider that we wouldn't have to do in that context.

On that note, do you know why the Caddy approach was failing? I might have missed that in the Slack discussions somewhere, but am curious.

freemvmt:
Oh also the Mermaid flows in the README are superb! 🚀

jessicamcinchak (approved):
Thanks again for this piece of work! Repeated curl requests and working as expected :white_check_mark:

infrastructure/application/cloudflare/snippets/validate_jwt.js
DafyddLlyr: Yep this is actually more correct - the reason I didn't put this in place are (rightly or wrongly - discuss!)
 - It's being forwarded to an endpoint which is going to check the format anyway (any token not formatted `header.payload.signature` will get rejected, we don't need to check twice)
 - I'm slightly paranoid about the snippet timeout (the longest operation is very likely our DB query there though for revoked tokens!)

DafyddLlyr:
> On that note, do you know why the Caddy approach was failing? I might have missed that in the Slack discussions somewhere, but am curious.

I really tried to get to the bottom of this without much luck. My best understanding was that requests going from Hasura Proxy (Caddy) to Hasura were going via HTTPS (and thus via Cloudflare). CF was then blocking these requests (responses had 403 and CF headers) - I couldn't actually see these requests in the CF console despite having `cf-ray-id` headers attached or work out a way to configure a rule to ignore these requests that worked. 

I spent a good while trying to configure internal Fargate networking (ie internal HTTP within the AWS VPC using service discovery) but pretty soon was out of my depth here and feeling the time pressure to switch to another track. In the end this proved to be simpler and quicker. Wish I could understand though what the issue was here - I think there's scope to dig back into this one after the pen test 👍 

> are we are confident we will always come in under the 5ms timeout?

Not entirely! I think there are optimizations we can make on the `revoked_tokens` table (ejecting records over 24hrs old, better indexing) that I'm planning to pick up this week. If we start hitting timeouts a few options would be - 
 - A cache (Redis?) to replace the `revoked_tokens` table
 - Convert CF Snippet to CF worker

infrastructure/application/cloudflare/snippets/validate_jwt.js
freemvmt: @DafyddLlyr makes sense to me! regex's can indeed get quite hungry unless 🐛

freemvmt:
@DafyddLlyr thanks for the explanation. Sounds like you really tried a lot of angles. Maybe next week we can post-mortem the whole thing and dig back into it as you suggest.

Having the option to turn the snippet into a worker if we run into problems is a great card up our sleeve 👌🏼

PR #4488, opened 2025-03-19
feat(sharedb): Hand auth responsibility over to `/auth/validate-jwt` endpoint

DafyddLlyr:
## What does this PR do?
 - Removes responsibility on ShareDB to verify JWT, and hands this over to the REST API
 - This endpoint both validates the JWT and checks for block-listed tokens
 - It also now returns the decoded JWT as ShareDB needs the user ID to write operations to Postgres
 - Replaces https://github.com/theopensystemslab/planx-new/pull/4467

We could simplify things here further and rely entirely on the same CF snippet used for Hasura auth here, but that would mean that ShareDB would be unsecured locally and on Pizzas. Not really a risk, but maybe just a bad practice and it would make the gap between Pizzas and staging/prod a little too large? I think that keeping visibility of this auth step, as in this PR, seems the pragmatic approach.

### DIFF
--- a/api.planx.uk/modules/auth/controller.ts
+++ b/api.planx.uk/modules/auth/controller.ts
@@ -111,13 +111,12 @@ export const isJWTRevoked: RequestHandler = async (req, res) => {
   if (!token) return res.status(401).send();
 
   try {
-    jwt.verify(token, process.env.JWT_SECRET!);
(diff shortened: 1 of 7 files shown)


freemvmt (approved):
Looks great! Works for me on local and pizza ✨

PR #4453, opened 2025-03-12
fix: Table array filtering

ianjon3s:
## What does this PR do?

- Reintroduces custom filtering (multiselect) for `ARRAY` option in data table (removed here: https://github.com/theopensystemslab/planx-new/pull/4402)
- Makes "type" filter in the feedback log function correctly

### To test

Please check filters across the platform admin panel, submissions log and feedback log! This PR should fix the 'live services' filter that was broken by a previous iteration.

PR #4492, opened 2025-03-19
feat: Display feedback status as chips

DafyddLlyr:


editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackLog/FeedbackLog.tsx
```diff
@@ -17,12 +17,23 @@ import {
...
 export const FeedbackLog: React.FC<FeedbackLogProps> = ({ feedback }) => {
   const columns: ColumnConfig<Feedback>[] = [
+    {
+      field: "status",
+      headerName: "Status",
+      type: ColumnFilterType.ARRAY,
+      customComponent: StatusChip,
+      columnOptions: {
+        valueOptions: statusOptions,
+        filterable: false,
```
DafyddLlyr, line 34: This is temporarily set to `false` as I need to get my head around the changes here - https://github.com/theopensystemslab/planx-new/pull/4453

It should be the case that one a column is set to `singleSelect` it automatically gets treated as an enum but the above PR undoes this default behaviour. I'll dig into this a bit more tomorrow and may have a few follow up questions!

DafyddLlyr:
Marked as draft, will sort out conflicts here.

editor.planx.uk/src/pages/FlowEditor/components/Flow/FeedbackLog/FeedbackLog.tsx
DafyddLlyr: I've got an upcoming PR which will address this, leaving as non-filterable initially.

editor.planx.uk/src/pages/FlowEditor/components/FeedbackLog/components/StatusChip.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { Feedback, FeedbackStatus } from "routes/feedback";
+
+import { feedbackStatusText } from "../utils";
+
+const FEEDBACK_COLOURS: Record<FeedbackStatus, ChipProps["color"]> = {
+  unread: "success",
+  urgent: "error",
+  to_follow_up: "warning",
+  read: "default",
+};
```
ianjon3s, line 13: Chips work great as an initial indicator (I had imagined some kind of vertical stroke on the left to denote status but we can refine this at a later date if need be).

As we're using a sort of progressive system of status from unread -> read I'd imagine the green (success) colour to be most effect as "read", with a more ambiguous colour "`theme.palette.info.main`" as unread.

DafyddLlyr: Thanks - great suggestion 👌 

image

PR #4489, opened 2025-03-19
feat: Editable comments column in feedback table

ianjon3s:
## What does this PR do?

Sets up front-end of feedback data table to include an editable column for notes.

There is already a column for user comments, so I've title the editable column "Editor notes" to make a clear distinction.

Column position TBC, I'm imagining this is most useful as an always visible column so I've placed it near the start.

For testing:
https://4489.planx.pizza/testing/feedback-test/feedback

PR #4493, opened 2025-03-19
feat: Add `feedback.editor_notes` column

DafyddLlyr:
## What does this PR do?
 - Adds `feedback.editor_notes` column
 - Exposes via `feedback_summary` view
 - Configures permissions
 - To be wired up to UI in https://github.com/theopensystemslab/planx-new/pull/4489

## Questions
I'm working off the Figma here and unfortunately missed the last design call where this was discussed.

My initial instinct here would have been to have a new `feedback_editor_notes` table, with `user_id`, timestamps etc and a many-to-one relationship to `feedback`. Going off the designs it actually seems that for now a single `notes` column, without this associated metadata, is going to meet our requirements at least in the medium term. Is this a fair summary of the decisions we've made here?

### DIFF
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -302,12 +302,14 @@
     - role: platformAdmin
       permission:
         columns:
+          - editor_notes
           - id
         filter: {}
       comment: ""
     - role: teamEditor
       permission:
         columns:
+          - editor_notes
           - id
         filter:
           team:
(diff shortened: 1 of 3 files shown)


jessicamcinchak (approved):
Per your open questions - single column on main table is exactly right here :+1: 

Even though councils triage feedback collaboratively, we don't actually anticipate need for back-and-forth comments - or in those cases there's no issue with them living in the same cell and being manually initialed/signed. Or we could decide on the frontend to auto-append name & timestamp to any updated comment text before saving to db, and this implementation supports either route :ok_hand:

DafyddLlyr:
> Per your open questions - single column on main table is exactly right here 👍

Perfect thanks - thought this was the case but always good to check!

PR #4494, opened 2025-03-19
chore: adjust platform admin filter logic

jessicamcinchak:
**A couple minor tweaks here:** 
- We don't need to maintain `internalTeamNames` on the frontend, we can instead filter these out at the source in the `teams_summary` database view because it's only relied on by the admin panel
- "Live services" names are sorted alphabetically in the filter options select, but not in the actual row cells - updating `array_agg(f.name)` to `array_agg(f.name ORDER BY f.name ASC)` in the view returns tidier data directly so cells will naturally render sorted names too


editor.planx.uk/src/pages/PlatformAdminPanel/utils.ts
```diff
@@ -1,15 +1,13 @@
...
-export const internalTeamNames = [
-  "WikiHouse",
-  "PlanX",
-  "Open Systems Lab",
-  "Testing",
-  "Open Digital Planning",
-  "Environment Agency",
-  "Templates",
-];
+export const getFlowNamesForFilter = (data: AdminPanelData[]) => {
```
jessicamcinchak, line 4: This helper function was previously in a `.tsx` file (but it's not a component), I've simply moved it to `utils.ts` instead here

editor.planx.uk/src/pages/PlatformAdminPanel/utils.ts
```diff
@@ -1,15 +1,13 @@
+import { capitalize } from "lodash";
 import { AdminPanelData } from "types";
 
-export const internalTeamNames = [
-  "WikiHouse",
-  "PlanX",
-  "Open Systems Lab",
-  "Testing",
-  "Open Digital Planning",
-  "Environment Agency",
```
jessicamcinchak, line 9: This original list filtered out the EA, but I'm not omitting them in this update - they may actually launch services that we want to see in the admin panel !

DafyddLlyr (approved):

PR #4496, opened 2025-03-20
fix: Filterable single selects in data grid

DafyddLlyr:
## What's the problem?
Whilst working on #4492 I realised that the `singleSelect` behaviour wasn't working as expected. This is due to changes made in https://github.com/theopensystemslab/planx-new/pull/4453. This is currently using a "contains" instead of an equals match which lead to filter options with "read" being a substring of "unread".

> [!IMPORTANT]
> If there's context from the week I was away or #4453 that I'm missing (i.e. another example of broken filtering?) please let me know!

Here's my understanding of the data model, and how I've managed this in this PR - 

**Single select**
This is a native MUI datagrid column type. It's used for enums (strings) and changes filters to is/is not/is any of.

image

By setting a column to have a type of `singleSelect`, we get this behaviour out of the box.

**Array**
This is a custom data type which is PlanX specific. It's used for the list of live services in the admin panel as this is an list of services, contained within a single "parent" (the team). We want to treat this column as an enum (filter by discrete name, not using string matching - e.g. "contains", "does not contain" etc).

As it's a custom type, we need custom filters to manage the behaviour.

## What's the solution?
- Create custom filter for our custom `ARRAY` column type
- Revert actual enum columns to `SINGLE_SELECT`

The main benefits here are that all single selects will be easily filterable using standard MUI functionality, and will not need any custom filter options - this was the issue I hit in #4492. Out custom `ARRAY` column does require custom filter operators.

## To test
 - I can filter Admin Panel > Services as expected (same behaviour as staging) ([link](https://4496.planx.pizza/))
 - I can filter Submissions Log > Status as expected (same behaviour as staging) ([link](https://storybook.4496.planx.pizza/?path=/docs/editor-components-submissions-events-log--docs))
   - Additionally, "is not" and "is one of" are added back as options

### DIFF
--- a/editor.planx.uk/src/ui/shared/DataTable/components/MultipleOptionSelectFilter.tsx
+++ b/editor.planx.uk/src/ui/shared/DataTable/components/MultipleOptionSelectFilter.tsx
@@ -58,7 +58,6 @@ export function MultipleOptionSelectFilter<T>(props: Props<T>) {
           setChipData(value);
           return applyValue({ ...item, value });
         }}
-        value={item.value}
       />
     </FormControl>
   );
--- a/editor.planx.uk/src/ui/shared/DataTable/utils.tsx
+++ b/editor.planx.uk/src/ui/shared/DataTable/utils.tsx
@@ -32,7 +32,9 @@ export const containsItem = (
   return false;
 };
 
-export const createFilterOperator = (columnValueOptions: ValueOptions[]) => [
+export const createArrayFilterOperator = (
+  columnValueOptions: ValueOptions[],
+) => [
   {
     value: "is",
     getApplyFilterFn: (filterItem: GridFilterItem) => {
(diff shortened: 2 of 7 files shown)


editor.planx.uk/src/ui/shared/DataTable/components/MultipleOptionSelectFilter.tsx
```diff
@@ -58,7 +58,6 @@ export function MultipleOptionSelectFilter<T>(props: Props<T>) {
           setChipData(value);
           return applyValue({ ...item, value });
         }}
-        value={item.value}
```
DafyddLlyr, line 61: Removed as this was throwing the "controlled" / "uncontrolled" input react error.

editor.planx.uk/src/ui/shared/DataTable/utils.tsx
```diff
@@ -32,31 +32,24 @@ export const containsItem = (
...
       }
 
-      // return a function that is applied to all the rows in turn
       return (currentRowValue: string[]): boolean => {
-        if (!currentRowValue?.length) {
-          return false;
-        }
+        if (!currentRowValue?.length) return false;
 
-        return Array.isArray(currentRowValue)
```
DafyddLlyr, line 49: Additional conditionality is not needed here as it will always be an array.

jessicamcinchak (approved):
Thanks for sorting this, nice simplifications here !

PR #4497, opened 2025-03-20
feat: Add online status chip to header

ianjon3s:
## What does this PR do?

Trello ticket:
https://trello.com/c/Nm29gakJ/3152-clarify-and-simplify-publishing-and-turning-a-service-online-offline

Appends an online/offline chip to the breadcrumbs of a flow when in the flow view (including settings pages):

image

The chip can be clicked to reach the settings page, where the status can be updated.

**Testing:**
https://4497.planx.pizza/barnet/apply-for-prior-approval

### DIFF
--- a/editor.planx.uk/src/components/Header/Header.tsx
+++ b/editor.planx.uk/src/components/Header/Header.tsx
@@ -1,12 +1,11 @@
-import Edit from "@mui/icons-material/Edit";
 import KeyboardArrowDown from "@mui/icons-material/KeyboardArrowDown";
 import OpenInNewIcon from "@mui/icons-material/OpenInNew";
 import OpenInNewOffIcon from "@mui/icons-material/OpenInNewOff";
 import Person from "@mui/icons-material/Person";
-import Visibility from "@mui/icons-material/Visibility";
 import AppBar from "@mui/material/AppBar";
 import Avatar from "@mui/material/Avatar";
 import Box from "@mui/material/Box";
+import Button from "@mui/material/Button";
 import Chip from "@mui/material/Chip";
 import { grey } from "@mui/material/colors";
 import Container from "@mui/material/Container";
@@ -40,6 +39,8 @@ import {
   LINE_HEIGHT_BASE,
 } from "theme";
 import { ApplicationPath } from "types";
+import FlowTag from "ui/editor/FlowTag/FlowTag";
+import { FlowTagType, StatusVariant } from "ui/editor/FlowTag/types";
 import Reset from "ui/icons/Reset";
 
 import { useStore } from "../../pages/FlowEditor/lib/store";
@@ -55,7 +56,7 @@ const Root = styled(AppBar)(({ theme }) => ({
   color: theme.palette.common.white,
 }));
 
-const BreadcrumbsRoot = styled(Box)(() => ({
+const BreadcrumbsRoot = styled(Box)(({ theme }) => ({
   cursor: "pointer",
   fontSize: 20,
   display: "flex",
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/components/Header/Header.tsx
```diff
@@ -231,58 +235,70 @@ const Breadcrumbs: React.FC = () => {
     state.previewEnvironment === "standalone",
   ]);
 
+  const flowStatus = useStore.getState().status as StatusVariant | undefined;
+
+  console.log(flowStatus);
```
DafyddLlyr, line 240:
```suggestion
  const flowStatus = useStore(state => state.flowStatus);
```

Within React components we should be using the hook variant of our Zustand store (`useStore()` as a function). We only really need to reach for the non-hook version outside React components - helper functions or tests generally.

DafyddLlyr (changes requested):

editor.planx.uk/src/routes/views/flowEditor.tsx
```diff
@@ -9,6 +9,7 @@ import { useStore } from "../../pages/FlowEditor/lib/store";
 
 interface FlowEditorData {
   id: string;
+  status: string;
```
DafyddLlyr, line 12: This will be of type `FlowStatus`

editor.planx.uk/src/routes/views/flowEditor.tsx
```diff
@@ -18,6 +19,7 @@ interface FlowEditorData {
 interface GetFlowEditorData {
   flows: {
     id: string;
+    status: string;
```
DafyddLlyr, line 22: Same here 👍

DafyddLlyr (changes requested):

editor.planx.uk/src/routes/views/flowEditor.tsx
ianjon3s: type `FlowStatus` imported and working correctly 👍

editor.planx.uk/src/components/Header/Header.tsx
ianjon3s: Thanks for this, very useful to know, I had copied the `.getState()` function from elsewhere, have updated

editor.planx.uk/src/components/Header/Header.tsx
```diff
@@ -231,58 +233,70 @@ const Breadcrumbs: React.FC = () => {
     state.previewEnvironment === "standalone",
   ]);
 
+  const flowStatus = useStore((state) => state.status) as
```
DafyddLlyr, line 236:
```suggestion
  const flowStatus = useStore((state) => state.flowStatus)
```

This is saved in the store as `flowStatus` not `status` 👍

DafyddLlyr (changes requested):

editor.planx.uk/src/routes/views/flowEditor.tsx
```diff
@@ -79,11 +84,18 @@ export const getFlowEditorData = async (
  */
 export const flowEditorView = async (req: NaviRequest) => {
   const [flow] = req.params.flow.split(",");
-  const { id, flowAnalyticsLink, isFlowPublished, isTemplate, templatedFrom } =
-    await getFlowEditorData(flow, req.params.team);
+  const {
+    id,
+    status,
```
DafyddLlyr, line 89: This was the cause of the bug, we're assigning `status` to the store, but only `flowStatus` is the expected / known property.

I'll take a look to see if we can make our Zustand stores stricter so that assigning a variable like this will get caught by TypeScript.

DafyddLlyr (approved):

editor.planx.uk/src/components/Header/Header.tsx
DafyddLlyr: Please see comment below - fixed in [daa2b7a](https://github.com/theopensystemslab/planx-new/pull/4497/commits/daa2b7aeca44f1e0a3837df445679a0c28440134)

DafyddLlyr:
Merging in @ianjon3s' absence 🚀

ianjon3s:
Thanks for this @DafyddLlyr !

PR #4498, opened 2025-03-21
feat: Change feedback status

DafyddLlyr:
image

image


editor.planx.uk/src/pages/FlowEditor/components/FeedbackLog/components/ChangeStatusTool.tsx
```diff
@@ -0,0 +1,106 @@
...
+  paddingTop: theme.spacing(0.5),
+  paddingBottom: theme.spacing(0.5),
+  marginBottom: theme.spacing(1),
+  "&:last-child": {
+    marginBottom: 0,
+  },
+}));
+
+export const ChangeStatusTool: React.FC = () => {
+  const apiRef = useGridApiContext();
```
jessicamcinchak, line 32: Cool to see that data grid has built in / preconfigured context that we can hook into :heavy_plus_sign:

editor.planx.uk/src/pages/FlowEditor/components/FeedbackLog/queries/updateFeedbackStatus.ts
```diff
@@ -0,0 +1,36 @@
...
+      }
+    `,
+    variables: {
+      ids: selectedRowIds,
+      status,
+    },
+    update: (cache) => {
+      // As we're reading from feedback_summary but pushing an update to feedback, we need to manually invalidate Apollo's cache
+      cache.evict({ fieldName: "feedback_summary" });
+      cache.gc();
```
jessicamcinchak, line 33: :+1:

jessicamcinchak (approved):
Pizza working exactly as expected here, thanks so much for picking this one up !

editor.planx.uk/src/pages/FlowEditor/components/FeedbackLog/components/ChangeStatusTool.tsx
DafyddLlyr: Totally - this was a real nice realisation for me also - a whole load of built in functionality we can use out of the box.

editor.planx.uk/src/pages/FlowEditor/components/FeedbackLog/queries/updateFeedbackStatus.ts
DafyddLlyr: Ah neat this is a super good observation - as it's a subscription it's already automatically pulling in updates. This would work!

I think we need to have a bit of a re-think about how we handle local state in the frontend - there's increasingly more tedious edge cases like this we need to work around.

PR #4501, opened 2025-03-21
feat: Filter `FeedbackLog` by rating, node type/title, and help text

DafyddLlyr:
## What does this PR do?
 - Addresses outstanding "TODO" comments on `FeedbackLog`
 - Allows filtering by rating, node type, and help text (see code for more details on some caveats here)
 - Relies on https://github.com/theopensystemslab/planx-new/pull/4496

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/FeedbackLog/FeedbackLog.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/FeedbackLog/FeedbackLog.tsx
@@ -110,9 +110,8 @@ export const FeedbackLog: React.FC<FeedbackLogProps> = ({ feedback }) => {
       headerName: "Help text (more information)",
       width: 280,
       type: ColumnFilterType.CUSTOM,
-      customComponent: ExpandableHelpText,
+      customComponent: (params) => <ExpandableHelpText {...params} />,
       columnOptions: {
-        filterable: false, // TODO: make filterable
         valueFormatter: (params) => stripHTMLTags(params),
       },
     },
--- a/editor.planx.uk/src/pages/FlowEditor/components/FeedbackLog/FeedbackLog.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/FeedbackLog/FeedbackLog.tsx
@@ -96,13 +96,9 @@ export const FeedbackLog: React.FC<FeedbackLogProps> = ({ feedback }) => {
       field: "nodeType",
       headerName: "Where",
       width: 280,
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/pages/FlowEditor/components/FeedbackLog/FeedbackLog.tsx
```diff
@@ -73,9 +73,9 @@ export const FeedbackLog: React.FC<FeedbackLogProps> = ({ feedback }) => {
       field: "feedbackScore",
       headerName: "Rating",
       width: 125,
+      type: ColumnFilterType.SINGLE_SELECT,
       columnOptions: {
-        filterable: false, // TODO: make filterable
-        valueFormatter: (params) => EmojiRating[params],
+        valueOptions: EmojiRating,
```
DafyddLlyr, line 78: We could try to treat this as a "number" column (allowing filter operators like >=, <, <= etc).

However, I've opted against this as we're displaying a text status to the Editors here - it makes sense to treat these as either discrete text categories or numbers, not a mixture of both.

editor.planx.uk/src/pages/FlowEditor/components/FeedbackLog/FeedbackLog.tsx
```diff
@@ -96,13 +96,9 @@ export const FeedbackLog: React.FC<FeedbackLogProps> = ({ feedback }) => {
       field: "nodeType",
       headerName: "Where",
       width: 280,
-      type: ColumnFilterType.CUSTOM,
```
DafyddLlyr, line 99: This has been reverted to a default (string) column type allowing string filtering (contains, does not contain etc). We could change this to be a `SINGLE_SELECT` again (based on just node type, not node title). Again open to ideas on what's best - this seems to offer the most flexibility just now.

editor.planx.uk/src/pages/FlowEditor/components/FeedbackLog/FeedbackLog.tsx
```diff
@@ -114,9 +110,8 @@ export const FeedbackLog: React.FC<FeedbackLogProps> = ({ feedback }) => {
       headerName: "Help text (more information)",
       width: 280,
       type: ColumnFilterType.CUSTOM,
-      customComponent: ExpandableHelpText,
+      customComponent: (params) => <ExpandableHelpText {...params} />,
```
DafyddLlyr, line 113: Looks like just a type issue was stopping this being filterable.

PR #4458, opened 2025-03-12
feat: toggle flow copiable permissions

jamdelion:

PR #4511, opened 2025-03-24
feat: Allow Editor to toggle if new flows can be created from a copy of their service

DafyddLlyr:
## What does this PR do?
- Adds a new column - `flows.can_create_from_copy`
- Adds form section allowing this property to be toggled
- Based on #4458

Relies on https://github.com/theopensystemslab/planx-core/pull/689

## Context
This thread is a good starting point (OSL Slack) - https://opensystemslab.slack.com/archives/C5Q59R3HB/p1742818367005979

We've previously described this as a flow being "copiable" but this isn't quite right. "Copy" as an action is unchanged, and should remain unchanged. This is the action of duplicating a flow within a single team. This new property will (in future) control if a flow can be used as a starting point for new services.

I'm very open to alternative (better!) naming conventions here - only strong option I have is that `is_copiable` is both incorrect and a bit misleading.


editor.planx.uk/src/pages/FlowEditor/lib/store/settings.ts
```diff
@@ -62,6 +65,21 @@ export const settingsStore: StateCreator<
     return Boolean(result?.id);
   },
 
+  flowCanCreateFromCopy: undefined,
```
DafyddLlyr, line 68: This is another example (I think) of issues with state management / boilerplate code for a relatively simple feature.

I'd love to dive into this, but it seems like these properties should likely live in either the Apollo cache or route data, as opposed to a global Zustand store.

Again, not advocating that we need a change here now but it's worthwhile revisiting next time we have a few similar settings or fields to manage.

jessicamcinchak (approved):
All makes sense & works as expected :+1:

PR #4512, opened 2025-03-25
style: List manager add row style update

ianjon3s:
## What does this PR do?

Updates styling of the list manager "insert" function:

- Keep the "add" symbol visible but simplify to a plus symbol to reduce visual noise
- Instinctively it feels like the left-side drag column should be kept free of any other interactions as it interacts in a vertical sense, so move the add icon to the right-side, so it sits with remove

Testing:
https://4512.planx.pizza/testing/long-checklist


DafyddLlyr (approved):
Much much nicer! 👍

PR #4513, opened 2025-03-26
style: Menu icon alignment

ianjon3s:
## What does this PR do?

Quick one:

- Updates editor menu icon/text alignment to accommodate multiple lines of text
- Removes redundant imports

Before (left) vs after (right):
image


editor.planx.uk/src/components/EditorNavMenu/styles.tsx
DafyddLlyr, line 1: CI is passing but I don't quite understand how this is working without the imported icons!

DafyddLlyr (approved):

editor.planx.uk/src/components/EditorNavMenu/styles.tsx
DafyddLlyr: Ah I'm with you sorry - mixing my files up here 😅 Great thanks!

PR #4515, opened 2025-03-26
test(e2e): Mock requests to Planning data

DafyddLlyr:
## What does this PR do?
- Resolves issues with flakey / unreliable E2E tests faced on https://github.com/theopensystemslab/planx-new/pull/4312 and https://github.com/theopensystemslab/planx-new/pull/4415
- Mocks request to Planning Data - this was sometimes returning inconsistent results or timing out, which led to non-deterministic outcomes
  - We should be mocking this anyway instead of hitting their live services for our CI
  - This should also (marginally) speed up tests as the request takes a good few seconds each time

✅ Regression tests passing here - https://github.com/theopensystemslab/planx-new/actions/runs/14087259503


e2e/tests/ui-driven/src/helpers/geospatialChecks.ts
```diff
@@ -51,10 +51,22 @@ export const checkGeoJsonContent = async (
...
+  const expectedCoords = geoJson.geometry.coordinates[0];
+
+  coords
+    .flat(2)
+    .forEach((coord, index) =>
+      expect(
+        coord,
+        "map attribute matches expected mock attribute",
+      ).toBeCloseTo(expectedCoords.flat(2)[index]),
+    );
```
DafyddLlyr, line 69: This was failing as the two GeoJSONs didn't match. A few reasons this could happen - 
 - Bumps to map repo (OL version, proj4 version)
 - Bumps to Playwright

Something in the mapping of mouse x & y to coords gives an inexact / unreliable response - these coordinates are always very close but never quite exact.

Relevant XKCD - https://xkcd.com/2170/

e2e/tests/ui-driven/src/mocks/gisResponse.ts
```diff
@@ -17,7 +18,7 @@ export async function setupGISMockResponse(page: Page) {
 export function checkGISMockRequestUrl(url: string) {
   const splitUrl = url.split("/").pop()?.split("%2C");
   return (
-    !splitUrl?.includes("designated.conservationArea") &&
+    splitUrl?.includes("designated.conservationArea") &&
```
DafyddLlyr, line 21: Again, this change was required in order to get tests to pass but it seems like a fairly basic change (`true` → `false`).

`setupGISMockResponse()` is only used in a single context, and it must resolve to `true`.

Any ideas what's caused this change?

jessicamcinchak: Nothing immediate comes to mind here ! On the editor side, the component is setup to check this/all datasets, right ?

e2e/tests/ui-driven/src/mocks/gisResponse.ts
```diff
@@ -32,6 +33,18 @@ export async function setupRoadsMockResponse(page: Page) {
...
+    /https:\/\/www\.planning\.data\.gov\.uk\/entity\.geojson/,
+  );
+  await page.route(planningDataEndpoint, async (route) => {
+    await route.fulfill({
+      status: 200,
+      body: JSON.stringify(planningDataResponse),
+    });
+  });
+}
+
```
jessicamcinchak, line 47: Big thing we missed - thanks so much for getting to bottom of this !

jessicamcinchak (approved):

e2e/tests/ui-driven/src/mocks/gisResponse.ts
DafyddLlyr: Yep as far as I can tell - it makes a lot more sense that both these conditions would be `true`.

PR #4517, opened 2025-03-26
feat: Transitions for `ListManager` items

DafyddLlyr:
## What does this PR do?
- Wraps ListManager items in a `TransitionGroup` ([docs](https://mui.com/material-ui/transitions/?srsltid=AfmBOop3o4bTsmlAoTgIOOUVIe6M1j3v_MQCXpB1eE9J301tFhhmMvOO#transitiongroup))
- Adds unique keys to each item, and a `Collapse` transition (other ideas welcome!)

https://github.com/user-attachments/assets/9197ea70-089f-4849-8d22-afea78708c84

### DIFF
--- a/editor.planx.uk/package.json
+++ b/editor.planx.uk/package.json
@@ -43,6 +43,7 @@
     "@turf/bbox-polygon": "^7.0.0",
     "@turf/buffer": "^7.0.0",
     "@turf/helpers": "^7.0.0",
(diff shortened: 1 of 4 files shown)


ianjon3s (approved):
Working perfectly, only one small nit:

- As each list item is now wrapped in a `<Collapse>` element, the `:last-child` 20px bottom margin property is now being applied to each item, should be applied to parent element

PR #4525, opened 2025-03-28
fix: React prop error (`inaccurateConstraints`)

DafyddLlyr:
image


jessicamcinchak (approved):
Thanks for catching / fixing this one !

PR #4534, opened 2025-04-01
feat: Create "analyst" role with access to admin panel

DafyddLlyr:
## What does this PR do?
- Creates new `users.is_analyst` column
- Creates new `analyst` role in Hasura which can query the `teams_summary` (admin panel) view
- Updates types and test (relies on https://github.com/theopensystemslab/planx-core/pull/693)
- Updates admin panel query to set role based on user

## To do
 - [x] Update `users.sql` once PR merged to `production` to sync new column
   -  PR https://github.com/theopensystemslab/planx-new/pull/4537 queued up

## To test
 - Toggle your user on the Pizza to set `user.is_platform_admin = false` and `user.is_analyst = true`
 - Access admin panel

### DIFF
--- a/hasura.planx.uk/tests/utils.js
+++ b/hasura.planx.uk/tests/utils.js
@@ -94,6 +94,7 @@ const introspectAs = async (role, userId = undefined) => {
     platformAdmin: gqlWithRole("platformAdmin", userId),
     teamEditor: gqlWithRole("teamEditor", userId),
     api: gqlWithRole("api"),
+    analyst: gqlWithRole("analyst"),
   }[role];
   const INTROSPECTION_QUERY = `
     query IntrospectionQuery {
@@ -112,10 +113,10 @@ const introspectAs = async (role, userId = undefined) => {
   const { types } = response.data.__schema;
   const queries = types
     .find((x) => x.name === "query_root")
-    .fields.map((x) => x.name);
+    ?.fields.map((x) => x.name) || []
   const mutations = types
     .find((x) => x.name === "mutation_root")
-    .fields.map((x) => x.name);
+    ?.fields.map((x) => x.name) || []
 
   return {
     types,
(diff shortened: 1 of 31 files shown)


hasura.planx.uk/tests/utils.js
```diff
@@ -94,6 +94,7 @@ const introspectAs = async (role, userId = undefined) => {
     platformAdmin: gqlWithRole("platformAdmin", userId),
     teamEditor: gqlWithRole("teamEditor", userId),
     api: gqlWithRole("api"),
+    analyst: gqlWithRole("analyst"),
```
jessicamcinchak, line 97: nit: should this be `gqlWithRole("analyst", userId)` ?

jessicamcinchak (commented):
:bug: Spotted one issue on the pizza when I switch myself to an analyst in the database (not a platform admin):
- I can see "Admin panel" in the menu
- But I can't access the page - console is throwing `Uncaught ApolloError: Your requested role is not in allowed roles`

DafyddLlyr:
Thanks - I can recreate this on the Pizza but not locally.

Here's what's causing this (I'm pretty sure) -
 - Pizza auth is happening on staging, where the JWT is created (missing the new role)
 - https://api.4534.planx.pizza/docs/#/user/get_user_me is hit on the Pizza, which has the correct details (saved in store, used to check `EditorNavMenu`)
 - When we make the admin panel request, we're using the JWT returned at time of auth (staging token)
 
What I think this means is - let's test this on staging and see if we can still recreate the issue 😅 If it's working, great. If not, I'll roll back.

hasura.planx.uk/tests/utils.js
DafyddLlyr: Yep good catch - just updated.

jessicamcinchak (approved):
Makes sense re: staging auth - let's merge & test again :+1:

PR #4541, opened 2025-04-02
refactor: Simplify team page filters

ianjon3s:
## What does this PR do?

- Introduces a simpler filter block with no accordion
- Simplifies styling of filter cards
- Introduces a grid/row view switch to the team page UI, which is updated in the store
- Adds a `Source Template` filter option to the template filter (run `window.featureFlags.toggle("TEMPLATES")` to see)

**Testing:**
https://4541.planx.pizza/a-new-team?sort=last-edited&sortDirection=desc

### DIFF
--- a/editor.planx.uk/src/pages/Team/index.tsx
+++ b/editor.planx.uk/src/pages/Team/index.tsx
@@ -1,7 +1,12 @@
+
+import TableRowsIcon from "@mui/icons-material/TableRows";
+import ViewModuleIcon from "@mui/icons-material/ViewModule";
 import Box from "@mui/material/Box";
 import Button from "@mui/material/Button";
 import Container from "@mui/material/Container";
 import { styled } from "@mui/material/styles";
+import ToggleButton from "@mui/material/ToggleButton";
+import ToggleButtonGroup from "@mui/material/ToggleButtonGroup";
 import Typography from "@mui/material/Typography";
 import { isEmpty, orderBy } from "lodash";
 import React, { useCallback, useEffect, useState } from "react";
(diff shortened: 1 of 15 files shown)


editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -68,13 +94,31 @@ const Team: React.FC = () => {
 
   const route = useCurrentRoute();
 
+  const [showCardGrid, toggleCardGrid] = useStore((state) => [
+    state.showCardGrid,
+    state.toggleCardGrid,
+  ]);
```
DafyddLlyr, line 100: This doesn't need to be tracked in both React state and the Zustand store - doing both could well lead to bugs. If we just get the functions/values we need from Zustand we could call them directly.

```ts
const [{ id: teamId, slug }, canUserEditTeam, getFlows, isTrial, showCardGrid, toggleCardGrid] = useStore(
  (state) => [
    state.getTeam(),
    state.canUserEditTeam,
    state.getFlows,
    state.teamSettings?.isTrial,
    state.showCardGrid,
    state.toggleCardGrid,
  ],
  ```

File: editor.planx.uk/src/pages/Team/index.tsx [hunk 2]
```diff
@@ -68,13 +94,31 @@ const Team: React.FC = () => {
...
+    _event: React.MouseEvent<HTMLElement>,
+    newView: "grid" | "row",
+  ) => {
+    if (!newView) return;
+
+    const isGrid = newView === "grid";
+    if (isGrid !== showCardGrid) {
+      toggleCardGrid();
+    }
+  };
```
DafyddLlyr, line 114: I think we can skip this and just call `() => toggleCardGrid()` from the onClick of the component.

editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -183,21 +240,22 @@ const Team: React.FC = () => {
...
+                  <ViewModuleIcon />
+                </ToggleButton>
+                <ToggleButton value="row" disableRipple>
+                  <TableRowsIcon />
+                </ToggleButton>
+              </ToggleButtonGroup>
             </Box>
             {sortedFlows && (
-              <DashboardList>
+              <DashboardList viewType={viewType as "grid" | "row"}>
```
DafyddLlyr, line 258: nit: I don't think the casting is required here.

DafyddLlyr (changes requested):
Looking great and working as expected - much simpler and cleaner.

One design/structure suggestion - 

 - We might be best considering the state of the flow cards as a string union `"grid" | "row"` instead of using a boolean
 - It's easy to fall in [the boolean trap](https://gunargessner.com/boolean-trap), and refactoring if a third state is introduced becomes surprisingly hard
 - It also means we can define our type once and share it - it's currently repeated numerous times
 
 
 ```ts
 // editor.ts
 
export type FlowCardView = "grid" | "row";

 ... 

 flowCardView: FlowCardView = "grid"
 setFlowCardView: (view: FlowCardView) => void;
...

editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -19,23 +23,40 @@ import FlowCard, { Card, CardContent } from "./components/FlowCard";
...
+    display: viewType === "grid" ? "grid" : "flex",
+    flexDirection: viewType === "grid" ? "row" : "column",
+    gap: theme.spacing(2),
+    gridAutoRows: "1fr",
+    gridTemplateColumns: viewType === "grid" ? "repeat(1, 1fr)" : "none",
+    [theme.breakpoints.up("md")]: {
+      gridTemplateColumns: viewType === "grid" ? "repeat(2, 1fr)" : "none",
+    },
+    [theme.breakpoints.up("lg")]: {
+      gridTemplateColumns: viewType === "grid" ? "repeat(3, 1fr)" : "none",
```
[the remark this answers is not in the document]
ianjon3s: Have updated, simpler and more readable 👍

editor.planx.uk/src/pages/Team/index.tsx
ianjon3s: Updated as `<DashboardList viewType={flowCardView}>`

ianjon3s:
Thanks for this @DafyddLlyr, I've implemented the suggestions to improve and simplify the logic and use of store.

A timely reminder to avoid using boolean logic to set variants, as I've been unpicking a boolean state in this PR today: https://github.com/theopensystemslab/planx-new/pull/4749

DafyddLlyr (approved):
Great! Thanks for picking up all those suggestions.

PR #4542, opened 2025-04-02
feat(api): Improved error handling for files deleted by Scanii

DafyddLlyr:
## What does this PR do?
 Improves error handling - it's possible an S3 file could be deleted by Scanii and we need to account for this

Docs on narrowing these issues down can be found here - #4543

### DIFF
--- a/api.planx.uk/modules/file/controller.ts
+++ b/api.planx.uk/modules/file/controller.ts
@@ -140,7 +140,7 @@ export const privateDownloadController: DownloadController = async (
 
 export type DeleteController = ValidatedRequestHandler<
   typeof hostedFileSchema,
-  Record<string, never>
+  Record<string, never> | { error: string }
 >;
 
 export const publicDeleteController: DeleteController = async (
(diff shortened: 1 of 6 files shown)


api.planx.uk/modules/file/controller.ts
```diff
@@ -140,8 +152,8 @@ export const publicDeleteController: DeleteController = async (
   const filePath = buildFilePath(fileKey, fileName);
 
   try {
-    const { isPrivate } = await getFileFromS3(filePath);
-    if (isPrivate) throw Error("Bad request");
+    const file = await getFileFromS3(filePath);
+    if (!file || file.isPrivate) throw Error("Bad request");
```
freemvmt, line 156: is it useful to differentiate between cases here? e.g. if someone is trying to delete a public file, but it in fact does not exist because it was never uploaded in the first place, it might be misleading to throw and error which says `Failed to delete...`?

freemvmt (approved):
Brilliant 💾

api.planx.uk/modules/file/controller.ts
DafyddLlyr: Good spot, fixed in [c926c52](https://github.com/theopensystemslab/planx-new/pull/4542/commits/c926c5206ccd7bf8983ba876e1300dcf6f0d7119)

PR #4547, opened 2025-04-03
fix: `AddFlow` modal fixes

DafyddLlyr:
- Changes incorrect "Service URL" label to more accurate "Editor URL"
 - Fixes bug where the URL changes on submit due to `window.location` updating


ianjon3s (approved):
Working as expected!
✅ slug no longer appends/replicates on submit

PR #4549, opened 2025-04-03
feat: add calculate component Boolean validation check

augustlindemer:
Adds a validation check to the calculate component, ensuring the component is toggled to 'automate a future question or checklist' if the formula returns a boolean value.


DafyddLlyr (approved):
Perfect! 🏅

PR #4540, opened 2025-04-02
fix: restrict duplicate send to email and s3 submissions

jessicamcinchak:
See thread https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1743588468574189

PR #4556, opened 2025-04-04
fix: correct GraphQL type for `email_applications` table

jessicamcinchak:
Fixes https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1743778821764219

Introduced in https://github.com/theopensystemslab/planx-new/pull/4540 (`s3_applications` table legitimately has `text` type column for `session_id`, so this was only wrong in one of the two places introduced)


freemvmt (approved):
Nice catch!

PR #4560, opened 2025-04-07
fix: publish validation checks for schema values should "fail" not "warn"

jessicamcinchak:
We've had two recent cases of submissions failing on unsupported `FileType` or `ProjectType` enum values. The validation checks for these are currently set to a "Warning" status rather than "Fail", which makes them very easy to ignore.

Now that passport values auto-suggest in the editor, there's less barrier to making them full failures and easier to discover/fix to use supported values.


DafyddLlyr (approved):
Nice simple change!

PR #4564, opened 2025-04-08
feat: add node heading to extractNodeTitle

zz-hh-aa:
Tiny change so that we can track Confirmation node headings in analytics too. I tested this locally (previously `node_title` would have been `null` for Confirmation nodes but they're pulling through successfully now). 

Let me know if there's a better way to do this, or if there was a specific reason as to why this was omitted before? Or if there might be any tricky implications that I've overlooked..? See both after and before in logs screenshot!

image


jessicamcinchak (approved):
I think this was previoulsy overlooked, great catch - thank you !!

DafyddLlyr (approved):

PR #4568, opened 2025-04-08
fix: Confirmation banner link colour

ianjon3s:
## What does this PR do?

Issue reported by @augustlindemer in Slack (also reported by council officers):
https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1744024034664429

The background colour was being passed in to the confirmation dialog as an opacity of a solid colour using the alpha value `rgba(1, 99, 96, 0.1)` , I've updated it to be a solid colour (defined unused theme variable palette.success.light). Interestingly VSCode previews this as a dark colour based on the rgb values alone, could be that this was causing `getContrastTextColor` to do the same, albeit intermittently. Either way we should use solid colours wherever possible.

Also, the colour values (bg + text) were being passed into the confirmation dialog as `initialValues`, presumably because we had intended this component to be user-customisable. As it isn't this leaves us no way to change the colour after the component is created, so I've updated to hardcoded values pulling from the theme palette.

Reported service fixed:
https://4568.planx.pizza/templates/apply-for-pre-application-advice-work-in-progress-copy

image

### DIFF
--- a/editor.planx.uk/src/@planx/components/Confirmation/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/Confirmation/Editor.tsx
@@ -56,10 +56,6 @@ export default function ConfirmationEditor(props: Props) {
   const type = TYPES.Confirmation;
   const formik = useFormik({
     initialValues: {
-      color: props.node?.color || {
-        text: "#000",
-        background: "rgba(1, 99, 96, 0.1)",
-      },
       heading: props.node?.data?.heading || "Application sent",
       description:
         props.node?.data?.description ||
--- a/editor.planx.uk/src/@planx/components/Confirmation/Confirmation.stories.tsx
+++ b/editor.planx.uk/src/@planx/components/Confirmation/Confirmation.stories.tsx
@@ -19,7 +19,6 @@ export const Basic = {
     heading: "Application sent",
     description: `A payment receipt has been emailed to you. You will also 
     receive an email to confirm when your application has been received.`,
-    color: { background: "#EFF7EE", text: "#0B0C0C" },
     sessionId: "123-t3st-456",
     applicableDetails: {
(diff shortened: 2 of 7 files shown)


editor.planx.uk/src/@planx/components/Confirmation/Editor.tsx
```diff
@@ -56,10 +56,6 @@ export default function ConfirmationEditor(props: Props) {
   const type = TYPES.Confirmation;
   const formik = useFormik({
     initialValues: {
-      color: props.node?.color || {
```
DafyddLlyr, line 59: Nice!

It's actually a historic leftover I think - there's so editor field to configure this value. We should also update `model.ts` to remove this if confirmation is just using a hardcoded colour 👍

augustlindemer:
Nice one! Works perfectly on the pizza. And yes, to confirm, no customisation of the banner colour needed here (we use one confirmation screen per application, and light green just says successfully sent).

editor.planx.uk/src/@planx/components/Confirmation/Editor.tsx
ianjon3s: Thanks @DafyddLlyr, have removed 👍

ianjon3s:
Thanks for testing @augustlindemer !

DafyddLlyr:
@ianjon3s I can see that dropping the `color` prop lead to a few type issues here.

Once you remove them you can run `pnpm check` in the Editor directory to test all TypeScript issues. It looks like you'll just need to remove the colour prop in associated test and Storybook files.

ianjon3s:
Thanks for checking back in on this one @DafyddLlyr , I had confused the need for a prop override in the public file with the props pulling in from `<Banner>`. I've removed from model, tests and storybook and all is working as expected.

DafyddLlyr (approved):

PR #4569, opened 2025-04-09
fix: Fee breakdown table text contrast

ianjon3s:
## What does this PR do?

Quick one: spotted in Storybook, the fee breakdown table uses the system text colour `GrayText` which does not give adequate contrast for WCAG AA. I've updated to use `palette.text.secondary` which bumps the contrast to a passing level.

image

`GrayText` (left) vs `text.secondary` (right):

image

There are several uses of text colour `GrayText` also in the editor that I'll pick up as a separate PR.

### DIFF
--- a/editor.planx.uk/src/@planx/components/Pay/Public/FeeBreakdown/FeeBreakdown.tsx
+++ b/editor.planx.uk/src/@planx/components/Pay/Public/FeeBreakdown/FeeBreakdown.tsx
@@ -46,7 +46,10 @@ const Header: FeeBreakdownRow = ({ amount }) => (
         Amount {amount.payableVAT ? ` (excl VAT)` : ``}
       </TableCell>
       {amount.payableVAT ? (
-        <TableCell align="right" sx={{ color: "GrayText" }}>
+        <TableCell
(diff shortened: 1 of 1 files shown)


jessicamcinchak (commented):
Thanks for spotting! If this is an issue with `GrayText` throughout - could we actually just overwrite the default `GrayText` hex in our global theme to match `theme.palette.text.secondary` and save per-line changes here? 

For me, the mental modal of gray = "GrayText" is easier to remember than gray = "theme.palette.text.secondary"

ianjon3s:
@jessicamcinchak I don't think there is way to customise or override `GrayText` if used this way, it is part of the system colors that pull directly from the browser or OS (also depreciated in CSS2 I believe)

jessicamcinchak (approved):
Aha that makes sense, thanks - I'd confusing thought `GrayText` was a MUI thing :+1: (Also aware React tests are failing on my base branch, so welcome to merge this whenever!)

ianjon3s:
Yeah it's part of a browser/OS defined system colors spec that I also had to look up!

PR #4573, opened 2025-04-10
feat: Improved loading state for planning constraints

ianjon3s:
## What does this PR do?

On reviewing loading states with Alice from Nomensa, we identified a weakness in our loading state for planning constraints, in that it does not accurately describe what is in progress. This is an important one to get right as there is a goldmine of information here, and the option to skip ahead is always present.

Preview (animation):
loading

Testing:
https://4573.planx.pizza/lambeth/constraints-test/published


DafyddLlyr (commented):
Quick comment as I've looked at MUI Skeleton before - 

- It doesn't currently meet WCAG standards (see issue here- https://github.com/mui/material-ui/issues/26930)
- It looks like we need to add `aria-busy` to the skeleton (this is always a good resource! https://design.sis.gov.uk/components/skeleton/accessibility)
- We can infer the dimensions of the child component by using `Skeleton` as a wrapper - not actually sure how this will work when we're loading data but worth a look ([MUI docs](https://mui.com/material-ui/react-skeleton/?srsltid=AfmBOorEvKNBra7Nj3T9D4F2wEiYjJZdG-RAa03WXtW4ELn5OqjLMmXG#inferring-dimensions))

DafyddLlyr (approved):

PR #4577, opened 2025-04-10
feat: Create team modal

DafyddLlyr:
## What does this PR do?
- Opens a modal when creating a team
- Allows the `isTrial` property to be toggled
- All additional team settings etc can be configured in the existing team settings interface

image

### DIFF
--- a/editor.planx.uk/src/pages/Teams/AddTeamButton.tsx
+++ b/editor.planx.uk/src/pages/Teams/AddTeamButton.tsx
@@ -1,0 +1,157 @@
+import Box from "@mui/material/Box";
+import Button from "@mui/material/Button";
+import Dialog from "@mui/material/Dialog";
+import DialogActions from "@mui/material/DialogActions";
+import DialogContent from "@mui/material/DialogContent";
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/pages/Teams/AddTeamButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+                        "settings.isTrial",
+                        !values.settings.isTrial,
+                      )
+                    }
+                    label={"Trial account"}
+                  />
+                  <Typography variant="body2" mt={-2}>
+                    A trial account has limited access to PlanX functionality
+                    (e.g. turning services online). A team can be promoted to
+                    having full access via the settings panel.
```
jessicamcinchak, line 129: Nit: in the second sentence, maybe "Trial teams can be promoted..." rather than "A team..." for clarity ?

jessicamcinchak (approved):
Team creation works just as expected ! 

Just confirming that services created within my new "trial" team can still be toggled online - but I suspect that is coming in a follow-up PR?

PR #4588, opened 2025-04-16
fix: Trial account caching, toggling flows offline

DafyddLlyr:
## What does this PR do?
- Fixes cache issue with Zustand store so `teamSettings.isTrial` is correctly persisted
- Toggle all flows offline for a team when setting `team_settings.is_trial = true`
  - Relies on https://github.com/theopensystemslab/planx-core/pull/712

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/TrialAccountForm.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/TrialAccountForm.tsx
@@ -62,10 +62,13 @@ export default function TrialAccountForm({
             }
           />
           <SettingsDescription>
-            <p>Toggle this team between "trial account" and "full access"</p>
+            <p>Toggle this team between "trial account" and "full access".</p>
             <p>
               A trial account has limited access to PlanX features - they are not
               able to turn services online.
+            </p>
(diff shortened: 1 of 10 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/TrialAccountForm.tsx
```diff
@@ -62,11 +62,14 @@ export default function TrialAccountForm({
...
           <SettingsDescription>
-            <p>Toggle this team between "trial" access and full access</p>
+            <p>Toggle this team between "trial" access and full access.</p>
             <p>
               A trial team had limited access to PlanX features - they are not
               able to turn services online.
             </p>
+            <p>
+              Toggling this to a trial account will turn this team's flows offline.
+            </p>
```
DafyddLlyr, line 72: This can be tested by toggling a team to and from trial account status.

The query where `toggle_flows_offline: true` will return a number of `affected_rows` (flows toggled offline). This will not be present when a service is promoted back to full access.

**Toggling to trial account**
image

**Toggling back to full access**
image

Toggling a team back to full access will not restore the previous online state of flows, this will need to be manually handled.

jessicamcinchak (approved):

PR #4596, opened 2025-04-17
fix(sharedb): Improved error handling

DafyddLlyr:
image

## What does this PR do?
 - Addresses issues raised a few times recently where ShareDB is silently failing, notably in this thread - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1744706505311059(OSL Slack)
 - Fixes bug with ShareDB auth checking
 - Improves error messages and error handling in frontend
 - Ensure WS client is only instantiated when required. This currently this happens on all public services! (see video below)

https://github.com/user-attachments/assets/1d1046e7-af15-4524-b969-5e078d4ae17e

 
I'm not totally convinced that this resolves the issues we're facing here - if this was the issue faced by Jonny (silent ShareDB error due to invalid token) I would have still expected Apollo to recognise this and trigger a logout.

At a minimum this PR rules some issues out and should give better visibility to any remaining issues. When a session is expired, this is now handled by both Hasura / Apollo as well as ShareDB.

### DIFF
--- a/editor.planx.uk/src/lib/graphql.ts
+++ b/editor.planx.uk/src/lib/graphql.ts
@@ -158,11 +158,11 @@ const parseErrorTypeFromHasuraResponse = (message: string) => {
 };
 
 export const handleExpiredJWTErrors = () => {
-  toast.error("Session expired, redirecting to login page...", {
-    toastId: "jwt_expiry_error",
+  toast.error("[GraphQL error]: Session expired, redirecting to login page...", {
+    toastId: "hasura_jwt_expiry_error",
     hideProgressBar: false,
     progress: undefined,
-    autoClose: 2_000,
+    autoClose: 4_000,
     onClose: () => (window.location.href = "/logout"),
   });
 
--- a/sharedb.planx.uk/server.js
+++ b/sharedb.planx.uk/server.js
@@ -93,7 +93,7 @@ async function validateJWT(authToken) {
     return decodedToken;
   }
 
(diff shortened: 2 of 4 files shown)


editor.planx.uk/src/lib/graphql.ts
```diff
@@ -158,11 +158,11 @@ const parseErrorTypeFromHasuraResponse = (message: string) => {
...
 
 export const handleExpiredJWTErrors = () => {
-  toast.error("Session expired, redirecting to login page...", {
-    toastId: "jwt_expiry_error",
+  toast.error("[GraphQL]: Session expired, redirecting to login page...", {
+    toastId: "hasura_jwt_expiry_error",
     hideProgressBar: false,
     progress: undefined,
-    autoClose: 2_000,
+    autoClose: 4_000,
```
DafyddLlyr, line 165: I found 2 seconds a little too short, bumped up to 4 throughout.

sharedb.planx.uk/server.js
```diff
@@ -93,7 +93,7 @@ async function validateJWT(authToken) {
     return decodedToken;
   }
 
-  throw Error(`Invalid JWT. Please log in again. Error: ${response.body}`)
+  throw Error("Invalid JWT. Please log in again.");
```
DafyddLlyr, line 96: The API just returns an empty 401 here, there's no body to parse.

jessicamcinchak (approved):
Thanks for changes here - all look good ! Agree at minimum this is good fixes and helps narrow things down further, and we'll keep iterating if needed :crossed_fingers:

PR #4599, opened 2025-04-18
feat: checklist options allow and display descriptions

jessicamcinchak:
[This is one from the content team's wishlist](https://opensystemslab.slack.com/lists/T03T7JTFV/F08HZNBTA7R?record_id=Rec08HZNC1RGX) and also useful for updates to the Send component 

`description` has always been in the `Option` model/types, but never displayed to editors or on the public interface. 

Now, editors can add descriptions to all non-exclusive options, grouped or not. I am still _not_ exposing the "description" input for the exclusive "or" option, does this seem correct?

### DIFF
--- a/editor.planx.uk/src/ui/shared/ChecklistItem/ChecklistItem.tsx
+++ b/editor.planx.uk/src/ui/shared/ChecklistItem/ChecklistItem.tsx
@@ -25,6 +25,7 @@ const Label = styled(Typography)(({ theme }) => ({
 interface Props {
   id?: string;
   label: string;
+  description?: string;
   checked: boolean;
   onChange: (event?: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
   inputProps?: React.InputHTMLAttributes<HTMLInputElement>;
@@ -34,6 +35,7 @@ interface Props {
 
 export default function ChecklistItem({
   label,
+  description,
   onChange,
   checked,
   id,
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/ui/shared/ChecklistItem/ChecklistItem.tsx
```diff
@@ -51,9 +53,32 @@ export default function ChecklistItem({
...
+      ) : (
+        <Label
+          variant="body1"
+          className="label"
+          component={"label"}
+          htmlFor={id}
+        >
+          {label}
+        </Label>
+      )}
```
jessicamcinchak, line 81: I'm not 100% sure of the `htmlFor` markup here - as I understand this only the first/primary label is truly associated with the checkbox itself, and the description might be a bit adrift ? 

I looked at `DescriptionRadio` for reference and was surprised to see there wasn't any use of `htmlFor` there ! 

Let's make sure this new "description" variation is covered in our a11y test flow for upcoming audit :heavy_plus_sign:

editor.planx.uk/src/@planx/components/Checklist/Checklist.stories.tsx
```diff
@@ -94,6 +125,25 @@ export const WithImages = {
   },
 };
 
+export const ExclusiveOr = {
```
DafyddLlyr, line 128: Thanks for this as well!

editor.planx.uk/src/ui/shared/ChecklistItem/ChecklistItem.tsx
DafyddLlyr: So I think the way this should work is that the label needs an `aria-describedby` which points to the `id` property of the description element added here 👍 

But yep, this is a good one to include on the upcoming test!

DafyddLlyr (changes requested):
Looking great - one suggested a11y change that I think we need here.

jessicamcinchak:
[Checklist "options" should support descriptions, consistent with Questions](https://trello.com/c/UrGTF8pR/3289-checklist-options-should-support-descriptions-consistent-with-questions)

PR #4601, opened 2025-04-20
feat: add customisable disclaimer to the Review component

augustlindemer:
Adds a customisable disclaimer to the Review component. 

First commit adds a rich text input _Disclaimer_ section to the editor modal, with default content.

Second commit adds a somewhat sketchy approach using the standard library method 'find' to (inefficiently) search for the Review component by node type and retrieve its disclaimer. I can't see anything else using this approach, so presumably there are cleaner and more efficient ways of retrieving the data of the Review component.

Unfortunately any of my attempts to retrieve the disclaimer through props defaulted to the default content. I'd much appreciate suggestions!

### DIFF
--- a/editor.planx.uk/src/@planx/components/Review/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/Review/Public/index.tsx
@@ -20,7 +20,8 @@ function Component(props: PublicProps<Review>) {
   return (
     <Presentational
       title={props.title}
-      description={props.description || ""}
+      description={props.description}
+      disclaimer={props.disclaimer}
       breadcrumbs={breadcrumbs}
       flow={flow}
       passport={passport}
--- a/editor.planx.uk/src/@planx/components/Review/Public/index.tsx
+++ b/editor.planx.uk/src/@planx/components/Review/Public/index.tsx
@@ -3,6 +3,7 @@ import { useStore } from "pages/FlowEditor/lib/store";
 import React from "react";
 
 import type { Review } from "../model";
+import { DEFAULT_REVIEW_DISCLAIMER } from "../model";
 import Presentational from "./Presentational";
 
 export default Component;
@@ -21,7 +22,7 @@ function Component(props: PublicProps<Review>) {
     <Presentational
       title={props.title}
       description={props.description}
-      disclaimer={props.disclaimer}
+      disclaimer={props.disclaimer || DEFAULT_REVIEW_DISCLAIMER}
       breadcrumbs={breadcrumbs}
       flow={flow}
       passport={passport}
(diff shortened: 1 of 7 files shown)


augustlindemer:
[Make Review component 'Change' warning editable](https://trello.com/c/Dilh77dX/3284-make-review-component-change-warning-editable)

editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
```diff
@@ -331,19 +332,13 @@ function SummaryList(props: SummaryListProps) {
...
-              application
-            </li>
-          </ul>
-        </Typography>
+        <ReactMarkdownOrHtml
+          // sketchy way to find the Review node in the flow to get the disclaimer, breaks if more than one review node in flow
+          source={
+            Object.values(props.flow).find((node) => node.type === TYPES.Review)
+              ?.data?.disclaimer
+          }
```
jessicamcinchak, line 340: Aha welcome to React land :dizzy: this is where the concept of "props" come in ! https://react.dev/learn/passing-props-to-a-component

So now, the top-level public/preview `Component` in Review/Public/index.tsx` has access to `props.disclaimer` which you've defined in the editor & model.ts and, instead of `find`, we can pass this down the component tree to `SummaryList` (going through a `Presentational` component in the middle) ! 

So try something like :
1. `<Presentational disclaimer={props.disclaimer} ... />`
2. `<SummaryListsBySection disclaimer={props.disclaimer} ... />`
3. Then here `source={props.disclaimer}` :slightly_smiling_face:

editor.planx.uk/src/@planx/components/Review/Public/index.tsx
```diff
@@ -20,7 +20,8 @@ function Component(props: PublicProps<Review>) {
   return (
     <Presentational
       title={props.title}
-      description={props.description || ""}
```
augustlindemer, line 23: Redundant blank fallback removed, handled in model.

editor.planx.uk/src/@planx/components/Review/Public/index.tsx
```diff
@@ -20,7 +20,8 @@ function Component(props: PublicProps<Review>) {
...
+      description={props.description}
+      disclaimer={props.disclaimer}
```
jessicamcinchak, line 24: Here, I think we need to annoyingly & redundantly handle the fallback text _again_ as _existing_ Review components will not have a `disclaimer` prop currently! Eg `{props.disclaimer || DEFAULT_DISCLAIMER_TEXT}`. 

The fallback defined in `model.ts` correctly already handles Editor modals and _new_ public components, but not pre-existing public components! (Hence why we do content migrations in more complex scenarios to "add" props!)

Eg when I go through an existing LDC on the pizza _without_ updating the Review component and re-publishing, I get an empty modal: https://4601.planx.pizza/lambeth/apply-for-a-lawful-development-certificate/published?analytics=false&sessionId=04bb6cf3-80e7-4971-ae4f-b53d9ed2b589
Screenshot from 2025-04-24 13-32-50

augustlindemer: Thanks! With the edits existing review pages now correctly fetch the default disclaimer.
image

jessicamcinchak (approved):
Perfect, all working as expected - thanks again for picking this one up :slightly_smiling_face:

PR #4603, opened 2025-04-22
chore(api & hasura/tests): Bump vitest, specify vite version

DafyddLlyr:
Resolves the following long-running dependabot issues - 

 - https://github.com/theopensystemslab/planx-new/security/dependabot/195
 - https://github.com/theopensystemslab/planx-new/security/dependabot/196
 - https://github.com/theopensystemslab/planx-new/security/dependabot/199
 - https://github.com/theopensystemslab/planx-new/security/dependabot/200
 - https://github.com/theopensystemslab/planx-new/security/dependabot/202
 - https://github.com/theopensystemslab/planx-new/security/dependabot/203
 - https://github.com/theopensystemslab/planx-new/security/dependabot/205
 - https://github.com/theopensystemslab/planx-new/security/dependabot/206

I'd been waiting for a vitest version to included an updated version of vite, but it's still at 6.2.0.

It turns out it's not an issue to just specify a higher version to resolve this - https://github.com/vitest-dev/vitest/issues/3850

I'll update the Editor issues later today, this is a bit trickier as we're a few versions behind on vitest and need to update tests and mocks accordingly.

### DIFF
--- a/api.planx.uk/package.json
+++ b/api.planx.uk/package.json
@@ -102,9 +102,9 @@
     "@types/uuid": "^10.0.0",
     "@typescript-eslint/eslint-plugin": "^5.62.0",
     "@typescript-eslint/parser": "^5.62.0",
-    "@vitest/coverage-istanbul": "^3.0.7",
-    "@vitest/eslint-plugin": "^1.1.36",
-    "@vitest/ui": "^3.0.7",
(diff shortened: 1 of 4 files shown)


jessicamcinchak (approved):
Thanks for resolving all these !

PR #4611, opened 2025-04-24
fix: align list and page number input styling with number input component

augustlindemer:
This PR replaces the generic Box container in the NumberFieldInput for List and Page component schemas with InputRow styling to ensure consistency with the NumberInput component (in particular, units prop margin). 

Alternatively we could add additional styling to the Box container, not sure if this is a preferable solution @DafyddLlyr?
image

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/Schema/InputFields/NumberFieldInput.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Schema/InputFields/NumberFieldInput.tsx
@@ -1,8 +1,8 @@
-import Box from "@mui/material/Box";
 import type { NumberField } from "@planx/components/shared/Schema/model";
 import React from "react";
 import InputLabel from "ui/public/InputLabel";
 import Input from "ui/shared/Input/Input";
+import InputRow from "ui/shared/InputRow";
 import InputRowLabel from "ui/shared/InputRowLabel";
 
 import { DESCRIPTION_TEXT, ERROR_MESSAGE } from "../../constants";
@@ -19,7 +19,7 @@ export const NumberFieldInput: React.FC<Props<NumberField>> = (props) => {
       {data.description && (
         <FieldInputDescription description={data.description} />
       )}
-      <Box sx={{ display: "flex", alignItems: "baseline" }}>
+      <InputRow>
         <Input
           {...fieldProps}
           onChange={formik.handleChange}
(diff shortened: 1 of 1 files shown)


DafyddLlyr (approved):
So this is actually already wrapped by an `InputRow` here - 

https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/src/@planx/components/shared/Schema/SchemaFields.tsx#L26-L39

In order for the styles to be consistent we want this to match the hierarchy of [the existing `NumberInput` component](https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/src/@planx/components/NumberInput/Public.tsx#L19), which is `InputRow` > (`InputLabel` > `Input`) & `InputRowLabel`.

It looks like this doesn't match that pattern (`InputRowLabel` is within the `InputLabel`).

It'd suggest trying this -
 - Remove the wrapping `Box`
 - Move the `InputRowLabel` outside the `InputLabel` to match the structure of `NumberInput`
   - As React components can only return a single element, you'll need to reach for a fragment (`<></>`) here ([docs](https://react.dev/reference/react/Fragment)).
   
If this doesn't work (I think labels are handled a little differently in schema fields vs plain components), then this solution of multiple `InputRows` is a good one - just worth trying the alternative first I think...! 👍

augustlindemer:
Thanks!

I tried matching the NumberInput as follows:

```
export const NumberFieldInput: React.FC<Props<NumberField>> = (props) => {
  const fieldProps = getFieldProps(props);
  const { id, errorMessage, required, title } = fieldProps;

  return (
    <>
      <InputLabel label={title} htmlFor={id}>
        {data.description && (
          <FieldInputDescription description={data.description} />
        )}
        <Input
          {...fieldProps}
          required={required}
          bordered
          type="number"
          inputProps={{
            "aria-describedby": [
              data.description ? DESCRIPTION_TEXT : "",
              errorMessage ? `${ERROR_MESSAGE}-${id}` : "",
            ]
              .filter(Boolean)
              .join(" "),
          }}
        />
      </InputLabel>
      {data.units && <InputRowLabel>{data.units}</InputRowLabel>}
    </>
  );
};
```

but the labels start drifting off:

image

Something I missed in the above, or is this due to the label handling differences you mentioned?

DafyddLlyr:
Ahh yep, thanks for testing @augustlindemer.

The `description` field is causing the suffix to be vertically centered across both the label and input element here. Your current solution of the additional `InputRow` is great!

augustlindemer:
Grand, thank you!

PR #4616, opened 2025-04-25
chore: Bump API rate limits

DafyddLlyr:
## What's the problem?
Editors are hitting the API rate limit - as confirmed by AWS logs (see thread here on OSL Slack - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1745418917141539).

This is happening because we're now hitting the `/auth/validate-jwt` endpoint far more frequently.

## Quick short term fix
 - Bump API rate limits
 - Log error to be able to confirm (via dev console and AWS logs) if this is the issue if hit again

## Better medium term fixes (TBD - added to dev call agenda)
- More nuanced rate limiting (ignore ShareDB requests?)
- Rate limit via Cloudflare for more sophisticated rate limiting controls
- Implement more sophisticated auth pattern in ShareDB

### DIFF
--- a/api.planx.uk/rateLimit.ts
+++ b/api.planx.uk/rateLimit.ts
@@ -5,7 +5,7 @@ import rateLimit from "express-rate-limit";
 const apiLimiter = rateLimit({
   message: "[API limiter]: Too many requests, please try again",
   windowMs: 10 * 60 * 1000, // 10 minutes
-  max: 250,
+  max: 2500,
   standardHeaders: true,
   legacyHeaders: false,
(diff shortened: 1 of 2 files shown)


jessicamcinchak (approved):
Thanks for pinpointing this ! Question: where did the original 250 come from? Was this an audit recommendation/requirement at some point or our discretion?

DafyddLlyr:
> where did the original 250 come from?

It was just an arbitrary / random number - the audit recommended a rate limit, and I think I just chose a sensible sounding low number.

PR #4617, opened 2025-04-25
fix: dashboard automation 

zz-hh-aa:
Updates `api` role permissions to include `select` on `flows.analytics_link`. (While testing #4415 on staging I realised that the last step was failing because of this.) 

The RAB template has also been changed since this was written, so the new production ID is updated.


DafyddLlyr:
@zz-hh-aa Just merging this so I can include it in a prod deploy 👍

PR #4618, opened 2025-04-25
refactor: GOV.UK Notify templates

DafyddLlyr:
## What problem am I trying to solve?
Updating (and understanding) templates is surprisingly hard currently. This is largely because they've grown quite organically, both in the number of templates and the data included within them. 

Our currently implementation relies on quite hierarchical types where one extends another, and shared functions which fetch the data for many templates. This makes the relationships between templates, personalisations, and data fetchers quite hard to understand. It also means templates are more closely coupled than they need to be - updating just a single template is not that simple.

### Motivation
This has been a bit of a long-running pain point, and with [a ticket in this sprint](https://trello.com/c/IFD6ow3h/3278-can-we-add-more-info-in-the-overview-summary-of-send-to-email) to add fields to a template it seems as good a time as any to revisit this.

## What am I trying to achieve?
Ideally, I'd like something which is clear, intuitive, and can be easily referenced against Gov Notify. It should be possible for us to add/remove personalisations per-template without much fuss.

## Implementation
I've scrapped the previous hierarchical types, replacing with a file (and type) per-template. The file structure here mimics the folder structure in GovNotify to keep things in line. Templates are held in a registry which contains both runtime details (template ID) as well as the type information (`NotifyConfig<T>`).

Personalisations are listed per template, with some repetition, instead of attempting to tie these together.


api.planx.uk/lib/notify/index.test.ts
```diff
@@ -1,14 +1,19 @@
 import { sendEmail } from "./index.js";
 import { NotifyClient } from "notifications-node-client";
-import type { NotifyConfig } from "../../types.js";
+import type { TemplateRegistry } from "./templates/index.js";
 
 vi.mock("notifications-node-client");
 
 const TEST_EMAIL = "simulate-delivered@notifications.service.gov.uk";
-const mockConfig: NotifyConfig = {
+const mockConfig: TemplateRegistry["save"]["config"] = {
```
DafyddLlyr, line 8: Type checking now requires the correct type to be passed in here.

api.planx.uk/lib/notify/index.ts
```diff
@@ -65,10 +28,17 @@ const sendEmail = async (
...
+
+    if (template === "expiry") {
+      const expiryConfig = config as TemplateRegistry["expiry"]["config"];
+      await softDeleteSession(expiryConfig.personalisation.sessionId);
+    }
+
+    if (template === "save") {
+      const saveConfig = config as TemplateRegistry["save"]["config"];
+      returnValue.expiryDate = saveConfig.personalisation.expiryDate;
+    }
```
DafyddLlyr, line 40: Casting is necessary here, TS is unable to match and type narrow on the shared generic `T`.

Same problem as previously discussed here - https://github.com/theopensystemslab/planx-new/pull/4153#discussion_r1919809762

api.planx.uk/modules/pay/service/inviteToPay/sendConfirmationEmail.test.ts
```diff
@@ -60,12 +60,12 @@ describe("sendAgentAndPayeeConfirmationEmail", () => {
     });
 
     const expectedConfig = {
+      emailReplyToId: "123",
```
DafyddLlyr, line 63: 🐛 This PR highlighted this issue in a few places.

`emailReplyToId` should be part of the config, not a part of the personalisation ([docs](https://docs.notifications.service.gov.uk/node.html#send-an-email-method)). In some tests where we're actually calling the test instance of the notify client it also needs to be both a valid UUID, as well as a valid `emailReplyToId` from our account.

This explains why we've previously had issue with applicants replying to an incorrect email address.

api.planx.uk/lib/notify/index.ts
```diff
@@ -87,6 +57,8 @@ const sendEmail = async (
 };
 
 const getClientForTemplate = (template: Template) =>
-  template in privateEmailTemplates ? $api.client : $public.client;
+  templateRegistry[template].access === "private"
+    ? $api.client
+    : $public.client;
```
jessicamcinchak, line 62: `templateRegistry[template].access` feels much more sensible :+1:

api.planx.uk/lib/notify/templates/saveAndReturn/user-confirmation.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export type Config = NotifyConfig<{
+  serviceName: string;
+  sessionId: string;
+  address: string;
+  projectType: string;
+  helpEmail: string;
+  helpOpeningHours: string;
+  helpPhone: string;
+}>;
```
jessicamcinchak, line 20: Agree one-file-per-template with repeated types feels like a much easier-to-work-with solution here than previous types, the overlap is tedious!

Does feel like there's one small type that's maybe useful to extract though - something like: 
```ts
type EmailFooter = {
  helpEmail: string;
  helpOpeningHours: string;
  helpPhone: string;
}; 

export type Config = NotifyConfig<{EmailFooter & {...}>;
``` 

Just a thought ! Feels slightly useful in my mental modal to separate fields related to applicant journey and team-level settings if possible. Aware not even the footer is share among all templates though !

doc/how-to/how-to-manage-gov-notify-templates.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. **Update codebase**
+   - If adding or removing fields, update the code first since templates are live - changes will go into effect immediately
+   - Additional fields are ignored, so these can be added in advance without any adverse effect
+
+2. **Update GOV.UK Notify template**
+   - After code changes are deployed to production, update the template in the GOV.UK Notify platform
+
+> [!TIP]
+> If making substantial changes to a template, it may be preferable to create a new template so that the changes are synced correctly on deploy
```
jessicamcinchak, line 63: Thanks for doc :star2:

jessicamcinchak (approved):
Much better & long overdue, thanks for finally tackling this one !

api.planx.uk/lib/notify/templates/saveAndReturn/user-confirmation.ts
DafyddLlyr: Great shout 🙂  I'll add this now.

PR #4622, opened 2025-04-28
chore: review page and list schemas

augustlindemer:
This PR includes example page schemas for London Datahub uses.


jessicamcinchak (approved):
Looks good :star:

augustlindemer:
Unfortunately revealed an issue with page component checklist handling. Instead of capturing multiple values, it adds them as numbered list items.

We end up with

```
"proposal.utilities.one":"fireSuppression"
"proposal.utilities.two":"communityOwned"
```

instead of
```
"proposal.utilities":[
0:"fireSuppression"
1:"communityOwned"
]
```

Will look into it.

editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
```diff
@@ -639,14 +641,19 @@ function Page(props: ComponentProps) {
...
+
   return (
     <>
       <Box component="dt">{props.node.data.title}</Box>
       <Box
         component="dd"
         sx={{ gap: 2, display: "flex", flexDirection: "column" }}
       >
-        {fields.map((field) => (
+        {fieldsWithAnswers.map((field) => (
```
jessicamcinchak, line 656: Also included a quick fix to how Pages display on "Review" component here: 
- Checklist answers should be comma-separated (still a bit imperfect that Review displays data value rather than user-facing text - let's come back to that one separately!)
- Optional questions that were not answered should be omitted

**Before:**
Screenshot from 2025-04-29 17-34-51

**After:**
Screenshot from 2025-04-29 17-35-14

editor.planx.uk/src/@planx/components/Page/Public.tsx
```diff
@@ -19,14 +19,17 @@ function PageComponent(props: Props) {
...
     // Default data - used for "back", to maintain a copy of the data as generated by the component
-    const defaultPassportData = { [props.fn]: data.schemaData };
+    const defaultPassportData = makeData(props, values.schemaData)?.["data"];
 
     // Flattened data - used for access via automation, SetValue etc
-    const flattenedData = flatten({ [props.fn]: data.schemaData[0] });
+    const flattenedData = flatten(defaultPassportData, {
+      depth: 2,
+      omitIndexKeys: true,
+    });
```
jessicamcinchak, line 32: Two main changes / fixes here: 
- Updated these methods to be much more consistent with the List components `handleSubmit` which I think will be a bit easier to maintain/reconcile going forward
- Added an optional `omitIndexKeys` argument to `flatten` which allows us to output flattened, but not indexed passport key names (eg without `one`, `two` segments)
  - This still matches the previously expected data output in `.../Page/Public.test.tsx` but now correctly accounts for Checklist-type fields too

augustlindemer: Perfection. Thank you!

PR #4630, opened 2025-04-30
fix: format data values as text for `Page` components on `Review` page

jessicamcinchak:
A nicer fix / follow-up to #4622 ! 

We already have an existing helper function `formatSchemaDisplayValue` which is used for active > inactive `List` cards

**Latest**
Screenshot from 2025-04-30 09-14-47

### DIFF
--- a/editor.planx.uk/src/@planx/components/shared/Schema/InputFields/ChecklistFieldInput.tsx
+++ b/editor.planx.uk/src/@planx/components/shared/Schema/InputFields/ChecklistFieldInput.tsx
@@ -10,11 +10,8 @@ import { getFieldProps, Props } from ".";
 import { FieldInputDescription } from "./shared";
 
 export const ChecklistFieldInput: React.FC<Props<ChecklistField>> = (props) => {
-  const {
-    data: { title, description, options },
-    formik,
-  } = props;
-  const { id, errorMessage, name, value } = getFieldProps(props);
+  const { data, formik } = props;
+  const { id, errorMessage, name, value, title } = getFieldProps(props);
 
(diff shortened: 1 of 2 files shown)


augustlindemer:
The data formatting works nicely! 

There is a small regression here in that unanswered checklists are shown in the Review page again, previously addressed for all field types [here](https://github.com/theopensystemslab/planx-new/pull/4622/commits/fca99c71dbf748f94d95d060984c7f5029aa88c5#r2066897251).

Current staging:
image

Pizza:
image

editor.planx.uk/src/@planx/components/shared/Schema/InputFields/ChecklistFieldInput.tsx
```diff
@@ -10,11 +10,8 @@ import { getFieldProps, Props } from ".";
 import { FieldInputDescription } from "./shared";
 
 export const ChecklistFieldInput: React.FC<Props<ChecklistField>> = (props) => {
-  const {
-    data: { title, description, options },
-    formik,
-  } = props;
-  const { id, errorMessage, name, value } = getFieldProps(props);
+  const { data, formik } = props;
+  const { id, errorMessage, name, value, title } = getFieldProps(props);
```
jessicamcinchak, line 14: Previously, we were reading `title` directly from props instead of passing through `getFieldProps` which is responsible for appending `(optional)` onto field titles. Checklist titles now displaying correctly: 
Screenshot from 2025-04-30 10-27-06

augustlindemer (approved):
Excellent stuff, works!

PR #4631, opened 2025-04-30
feat(wip): show `platform` in `submission_services_summary`

zz-hh-aa:
I have to say I'm very unsure about this! (Hence draft & wip.) 

Because we've had requests for graphs to show breakdowns by device type, I tried to add `platform` to the allow list in a little workaround to show the device used for submission in `submission_services_summary`. 

Local testing of submissions is just showing me many `NULL` values which I'm also confused by:
image

Am I totally on the wrong track? Tips and advice greatly appreciated!

# What does this PR do?
- Creates a new migration to show `platform` in `submission_services_summary` table
- In the Send component, saves `platform` and passes it to `handleSubmit`
- Adds `platform` to allow list
- Includes `platform` in `createAnalytics`

### DIFF
--- a/editor.planx.uk/src/@planx/components/Send/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/Send/Public.tsx
@@ -2,6 +2,7 @@ import ErrorOutline from "@mui/icons-material/ErrorOutline";
 import Typography from "@mui/material/Typography";
 import { SendIntegration } from "@opensystemslab/planx-core/types";
 import axios, { AxiosResponse } from "axios";
+import Bowser from "bowser";
 import DelayedLoadingIndicator from "components/DelayedLoadingIndicator/DelayedLoadingIndicator";
 import { useStore } from "pages/FlowEditor/lib/store";
 import React, { useEffect } from "react";
@@ -107,7 +108,8 @@ const CreateSendEvents: React.FC<Props> = ({
       ]),
     );
 
-    props.handleSubmit && props?.handleSubmit({ data });
+    const platform = Bowser.parse(window.navigator.platform); // This is a weird workaround so that we can include platform in `allow_list_answers` in order to pull it through easily in the `submission_services_summary` table
+    props.handleSubmit && props?.handleSubmit({ data: { ...data, platform } });
   }, [loading, error, value, destinations, props]);
 
   // Throw errors so that they're caught by our error boundaries and Airbrake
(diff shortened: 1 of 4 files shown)


editor.planx.uk/src/@planx/components/Send/Public.tsx
```diff
@@ -107,7 +108,8 @@ const CreateSendEvents: React.FC<Props> = ({
       ]),
     );
 
-    props.handleSubmit && props?.handleSubmit({ data });
+    const platform = Bowser.parse(window.navigator.platform); // This is a weird workaround so that we can include platform in `allow_list_answers` in order to pull it through easily in the `submission_services_summary` table
+    props.handleSubmit && props?.handleSubmit({ data: { ...data, platform } });
```
jessicamcinchak, line 112: So a couple thoughts here: 
- Rather than only grab `platform` here, I wonder if we can directly mimic what analytics log do (eg save a lot of info about the referrer & user agent to the passport upfront), then parse out just the specific fields we need in the postgres view? This should have benefit of being more prepared in case any other adjacent M&E requests in future? 
- Also a nit picky suggestion about passport field names: it's nice to be able to quickly distinguish which passport fields are picked up by a user versus which are for internal tracking, navigation, analytics etc - we have an existing pattern like `{componentName}.action` which we use for tracking how users proceed through geo-components, maybe something like `{componentName}.analytics` could work similarly here? 
  - And would inherently remind us downstream of _where_ we captured this info per Daf's early point that we can't guarantee it's the same browser someone started on !

So maybe like: 
```tsx
... 
const userAgent = Bowser.parse(window.navigator.userAgent);
const referrer = document.referrer || null;

props.handleSubmit & props.handleSubmit({ 
  data: {
    ...data,
    "send.analytics.userAgent": userAgent,
    "send.analytics.referrer": referrer,
  }
});
```
Then hopefully can re-use these lines from `analytics_summary` ? :
```pg
(user_agent -> 'os' ->> 'name') :: text AS operating_system,
(user_agent -> 'browser' ->> 'name') :: text AS browser,
(user_agent -> 'platform' ->> 'type') :: text AS platform,
referrer,
```

(Untested code above! If you implement this and still getting mostly blanks, let me know & very happy to take a closer look !!)

jessicamcinchak (commented):
Thanks for picking this one up! Looks great and right on track, left a couple initial thoughts about what to capture in passport :slightly_smiling_face:

editor.planx.uk/src/@planx/components/Send/Public.tsx
zz-hh-aa: Thanks for the pointers Jess! 🙏 That makes a lot of sense. I've made the changes but am still getting `NULL`s, across fields unrelated to the changes I'm trying to make--please see #4643 (I started a new branch / PR for the changes because of local migrations confusion). 

As before, no rush at all!

PR #4643, opened 2025-05-01
feat: show `userAgent` and `referrer` in `submission_services_summary` through allow list

zz-hh-aa:
Thanks Jess for the pointers!!

This is a new version of #4631 because I got myself into a migrations pickle and had to open a new branch. Please see earlier conversation for context. 

# Local tests context
When I test locally, I'm still getting some unexpected `NULL`s in Hasura. Can't quite figure out where I'm going wrong!

Certain values populate correctly in `analytics_summary` (but not all, eg `flow_direction` is also `NULL`)
Screenshot 2025-05-01 155610

Same is happening in `submission_services_summary`; certain values populate but not all (eg the submission date, etc.)
Screenshot 2025-05-01 155654
Screenshot 2025-05-01 155717

# What does this PR do?
- Creates a new migration to show `userAgent` properties (`os`, `browser`, `platform`) as well as `referrer` in submission_services_summary table
- In the Send component, saves `userAgent` and `referrer` and passes them to `handleSubmit`

# Why?
M&E want to be able to break submissions down by platform.

### DIFF
--- a/editor.planx.uk/src/@planx/components/Send/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/Send/Public.tsx
@@ -2,6 +2,7 @@ import ErrorOutline from "@mui/icons-material/ErrorOutline";
 import Typography from "@mui/material/Typography";
 import { SendIntegration } from "@opensystemslab/planx-core/types";
 import axios, { AxiosResponse } from "axios";
+import Bowser from "bowser";
 import DelayedLoadingIndicator from "components/DelayedLoadingIndicator/DelayedLoadingIndicator";
 import { useStore } from "pages/FlowEditor/lib/store";
 import React, { useEffect } from "react";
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/Send/Public.tsx
```diff
@@ -107,7 +108,16 @@ const CreateSendEvents: React.FC<Props> = ({
...
+    const userAgent = Bowser.parse(window.navigator.userAgent); // This is a weird workaround so that we can include platform in `allow_list_answers` in order to pull it through easily in the `submission_services_summary` table
+    const referrer = document.referrer || null;
+    props.handleSubmit &&
+      props.handleSubmit({
+        data: {
+          ...data,
+          "send.analytics.userAgent": userAgent,
+          "send.analytics.referrer": referrer,
+        },
+      });
```
jessicamcinchak, line 120: Passport populating for me as-expected across Chrome, incognito Chrome, Firefox :heavy_plus_sign: 

`userAgent` is consistently recorded, `referrer` is spottier but this is what I'd expect with localhost testing where I don't always have a true "referrer" !

jessicamcinchak (approved):
This is all working as expected for me ! 

A couple guesses as to what might explain your local testing screenshots:
- `submitted_at` date in `submission_services_summary` is only going to be populated when the actual send event succeeds (different than just passing through the Send component in a flow) - in a local env, these events will only succeed if you have the integration configured (eg setup submission email in team settings if using send to email). And populating the allow-list column on `lowcal_sessions` is only triggered by a successful submission!
- less sure about the occassionally null `flow_direction` in analytics summary, but that seems unrelated to these changes & possibly a local env quirk? 

I'd say this is okay to merge & we can monitor on live services & fix-forward as needed!

editor.planx.uk/src/@planx/components/Send/Public.tsx
zz-hh-aa: Ah super! All sounds good, thanks very much for your help jess 💫

PR #4648, opened 2025-05-02
feat: update Metabase automation according to new structure

zz-hh-aa:
The big Metabase restructure this week has meant that most of the previous client was obsolete 😿 thankfully the new structure makes the code side of dashboard automation much more straightforward!

I've tested the changes locally and they're all working for me. 

# What does this PR do?
- Deletes most of old client code (eg client itself and Metabase collections stuff removed entirely)
- Removes the old Hasura event listener that would create a new Metabase collection
- Removes the `collection` route from the API
- Adapts some of the dashboard-related code according to new Metabase structure
- Updates tests accordingly

# Why?
Following our analytics chat, Metabase has been drastically restructured to make maintenance and change propagation easier. This means we are no longer duplicating dashboards ore creating collections for teams. All service dashboards are views of an existing template, just with the filters / params customised. Simpler for everyone involved...


DafyddLlyr (changes requested):
A few comments below on structure - I think constructing the URLs would be a bit better than calling string replacements.

Great PR though, always great to see deletions and simplifications. 🥇 

Here's a challenge / question though - can we take this further? Right now we're storing the URLs in the database meaning they're static - if we add a new filter or change a template dashboard ID we'd need to manually update a bunch of data fields. Now that the dashboard URL is derivable from the team name and service slug, could we just move this logic to the frontend and lose all API and database logic? The original motivation for having a db column was that each flow would have a corresponding unique Metabase dashboard ID we'd need to track but this is no longer the case.

I _think_ this would give us something simpler still and more flexible / maintainable (possibly something I'm not thinking about here).

So this would mean - 
 - Drop all API and Hasura code
 - Move logic in `api.planx.uk/modules/analytics/metabase/findDashboardPublicLink.ts` to the frontend
 - Generate URL when user clicks on the analytics link in the frontend
 
 What do you think?
 
On another note, I'd still support trying to find a way to use flowId as a filter for data continuity over the slug. Maybe segments in Metabase can achieve this, or maybe there's another method (using ID as a filter but then displaying the slug prominently?). Something we can iterate on (and it will be easier to do so using dynamic URLs than having to migrate and manage values already written to the DB).

zz-hh-aa:
Thanks for the feedback @DafyddLlyr! 

Suuuper interesting questions. I hadn't considered this way of running analytics at all--the one potential complication I can think of now is in the case of a council changing a flow's name and wanting the analytics dashboard to filter for both the current and past slug, as we discussed with Silvia the other week. This is an instance in which having a db column would still be necessary, I think. 

For now, I have worked your other requested changes into this PR. Links are now generated, rather than managed with string replace. 

I'm happy to explore frontend link generation as well as filtering on flow IDs. Could we perhaps merge this and #4649 for now, in order to separate concerns and also to bring the codebase more in line with the current Metabase structure? (eg once we merge these, I can delete the old templates without causing errors on flow creation, etc.)

api.planx.uk/modules/analytics/metabase/generateDashboardLink.ts
```diff
@@ -0,0 +1,91 @@
...
+  const baseDomain = environment === "production" ? "uk" : "dev";
+  const host = `https://metabase.editor.planx.${baseDomain}`;
+  const pathname = `/public/dashboard/${dashboardId}`;
+  const url = new URL(pathname, host);
+  const search = new URLSearchParams({
+    service: serviceSlug,
+    team: teamSlug,
+  }).toString();
+  url.search = search;
+  console.log({ url });
```
DafyddLlyr, line 88: A few leftover `console.log()`s here

PR #4649, opened 2025-05-02
feat: Metabase listens for `flows.status` in event trigger instead of insert

zz-hh-aa:
# What does this PR do?
- Drops `metabase_id` column from `teams` table
- Updates event listener to track updates to `flows.status` instead of insertions
- Checks if `status` = `online` before proceeding 

# Why? 
- Following big Metabase restructure, we no longer have team collection IDs to store in `teams.metabase_id`
- We also thought that only creating analytics links when services go live (rather than when they are created) would be tidier

Everything tested and working locally!


api.planx.uk/modules/analytics/metabase/service.ts
```diff
@@ -9,10 +9,15 @@ import { getTeamSlug } from "./getTeamSlug.js";
...
  */
 export async function createNewDashboardLink({
+  status,
   flowId,
   teamId,
   serviceSlug,
 }: CreateNewDashboardLinkParams): Promise<string | undefined> {
+  if (status !== "online") {
+    return;
+  }
```
DafyddLlyr, line 19: Do we now also need to check if there's an existing dashboard, or control this in some other way?

What happens if a flow goes offline → online → offline → online ?

zz-hh-aa: Good question! I figured since we're not actually creating a new dashboard (and the link that would be generated and saved would be the same), that it would be fine to run anyway--but if running the rest of the code is unnecessary, there's a more elegant way of handling this. 

Now I'm selecting `flows.analytics_link`, checking if `null` and adding that to the condition too, see [09fd9f8](https://github.com/theopensystemslab/planx-new/pull/4649/commits/09fd9f82ed6268bc89d1ea03c3f8317bb554bcfb)

DafyddLlyr: Yep great point, thanks for clarifying!

DafyddLlyr (commented):
Please see comments on https://github.com/theopensystemslab/planx-new/pull/4648

We could still achieve this by using dynamic URLs (not saving strings to the db, just deriving them at runtime) and then just disabling the button in the frontend for services which are not online 👍

api.planx.uk/modules/analytics/metabase/service.ts
```diff
@@ -9,34 +9,38 @@ import { getTeamSlug } from "./getTeamSlug.js";
...
 export async function createNewDashboardLink({
+  analyticsLink,
+  status,
   flowId,
   teamId,
   serviceSlug,
 }: CreateNewDashboardLinkParams): Promise<string | undefined> {
-  const environment =
-    process.env.APP_ENVIRONMENT === "production" ? "production" : "staging";
+  if (status === "online" && analyticsLink === null) {
```
DafyddLlyr, line 18: nit: Could we change this to a guard clause (inversing the condition) to save a level of indentation (and overhead) here?

```ts
// Before
if (condition) {
  doSomething()
} else { 
  return;
}

// After
if (!condition) return;

doSomething()
```

DafyddLlyr (approved):

PR #4651, opened 2025-05-06
feat: Chip and filter options for template flows

DafyddLlyr:
## What does this PR do?
 - Adds a "Template" chip
 - Adds a "Template" filter option
 - Both behind the `TEMPLATES` feature flag still

https://github.com/user-attachments/assets/5fc4a9a5-9335-4df8-946a-26001df14aaa

### DIFF
--- a/editor.planx.uk/src/pages/Team/helpers/sortAndFilterOptions.ts
+++ b/editor.planx.uk/src/pages/Team/helpers/sortAndFilterOptions.ts
@@ -1,3 +1,4 @@
+import { hasFeatureFlag } from "lib/featureFlags";
 import { FlowSummary } from "pages/FlowEditor/lib/store/editor";
 import { FilterOptions } from "ui/editor/Filter/Filter";
 import { SortableFields } from "ui/editor/SortControl/SortControl";
@@ -30,7 +31,12 @@ const checkFlowServiceType: FilterOptions<FlowSummary>["validationFn"] = (
   _value,
 ) => flow.publishedFlows[0]?.hasSendComponent;
 
-export const filterOptions: FilterOptions<FlowSummary>[] = [
+const checkFlowTemplatedFrom: FilterOptions<FlowSummary>["validationFn"] = (
+  flow,
+  _value,
+) => Boolean(flow.templatedFrom)
+
+const baseFilterOptions: FilterOptions<FlowSummary>[] = [
   {
     displayName: "Online status",
     optionKey: "status",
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/ui/editor/FlowTag/styles.ts
```diff
@@ -38,4 +38,8 @@ export const Root = styled(Box, {
   ...(tagType === FlowTagType.ServiceType && {
     backgroundColor: theme.palette.flowTag.serviceType,
   }),
+  // TODO: Is there a colour to communicate "template"?
+  ...(tagType === FlowTagType.Template && {
+    backgroundColor: theme.palette.flowTag.serviceType,
+  }),
```
DafyddLlyr, line 44: @ianjon3s I've just repeated an existing colour here, just flagging for your attention. No immediate response needed, we can pick things like this up in a final style pass on the templates feature.

editor.planx.uk/src/pages/Team/helpers/sortAndFilterOptions.ts
```diff
@@ -44,3 +50,14 @@ export const filterOptions: FilterOptions<FlowSummary>[] = [
     validationFn: checkFlowServiceType,
   },
 ];
+
+const templateFilterOption: FilterOptions<FlowSummary> =  {
+  displayName: "Template",
+  optionKey: "templatedFrom",
+  optionValue: ["template"],
+  validationFn: checkFlowTemplatedFrom,
+};
```
DafyddLlyr, line 59: I'm not sure just "Template" really works here without context, something we can think about!

jessicamcinchak: I do wonder if "Template" can be an option under "Service type" rather than it's own third category with same-name option?

DafyddLlyr: This was also my first thought, but these options are always exclusive.

https://github.com/user-attachments/assets/995b4fa2-a661-48e6-9d81-95cb5489ca7b

Not ideal and not a simple change after a little while looking at this. A lot of the filter code is pretty complex and I'd love to actually revisit some of this, especially compared to how easy and powerful it was to use the MUI datagrid API. I think there's also a much better general solution here using something like tanstack query (filtering server side, not client side).

jessicamcinchak (approved):
Great ! Since feature flagged, I'd say no harm merging this first go even if Ian is planning a re-style soon?

editor.planx.uk/src/ui/editor/FlowTag/styles.ts
DafyddLlyr: Merged for now @ianjon3s as this is feature flagged, we can come back and pick up this styling 👍

PR #4652, opened 2025-05-06
feat: Create source template

DafyddLlyr:
## What does this PR do?
 - Adds a "source template" toggle to the create flow modal (platform admins only)
 - Updates API to accept this new argument, refactors `createFlow()` to take a single options argument
 - Adds temporary "Source Template" tag to `FlowCard`
   - This will be replaced by a styled card, but serves as a visual cue for now and sorts out the data needed for this restyle
   - Please see https://github.com/theopensystemslab/planx-new/pull/4651#discussion_r2075513134

image

### DIFF
--- a/api.planx.uk/helpers.ts
+++ b/api.planx.uk/helpers.ts
@@ -125,7 +125,7 @@ const createFlow = async ({
         copied_from: copiedFrom,
         templated_from: templatedFrom,
         is_template: isTemplate,
-      }
+      },
     );
 
     await createAssociatedOperation(id);
--- a/api.planx.uk/helpers.ts
+++ b/api.planx.uk/helpers.ts
@@ -66,6 +66,58 @@ interface CreateFlowResponse {
   };
 }
 
+const createFlowMutation = gql`
+  mutation InsertFlow(
+    $team_id: Int!
+    $slug: String!
+    $name: String!
+    $data: jsonb = {}
+    $copied_from: uuid
+    $templated_from: uuid
+  ) {
+    flow: insert_flows_one(
+      object: {
+        team_id: $team_id
+        slug: $slug
+        name: $name
+        data: $data
+        version: 1
+        copied_from: $copied_from
+        templated_from: $templated_from
+      }
+    ) {
+      id
+    }
+  }
+`;
+
+const createSourceTemplateMutation = gql`
+  mutation InsertSourceTemplateFlow(
+    $team_id: Int!
+    $slug: String!
+    $name: String!
+    $data: jsonb = {}
+    $copied_from: uuid
+    $templated_from: uuid
+    $is_template: Boolean
+  ) {
+    flow: insert_flows_one(
+      object: {
+        team_id: $team_id
+        slug: $slug
+        name: $name
+        data: $data
+        version: 1
+        copied_from: $copied_from
+        templated_from: $templated_from
+        is_template: $is_template
+      }
+    ) {
+      id
+    }
+  }
+`;
+
 // Insert a new flow into the `flows` table
 const createFlow = async ({
   teamId,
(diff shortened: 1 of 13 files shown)


api.planx.uk/helpers.ts
```diff
@@ -66,55 +66,96 @@ interface CreateFlowResponse {
...
+  copiedFrom?: Flow["id"];
+  templatedFrom?: Flow["id"];
+}) => {
   const { client: $client } = getClient();
   const userId = userContext.getStore()?.user?.sub;
 
+  /** Only platformAdmins have write permissions for the is_template column */
+  const mutation = isTemplate
+    ? createSourceTemplateMutation
+    : createFlowMutation;
```
DafyddLlyr, line 145: This feels slightly awkward, not sure if there's a better pattern we could use here?

api.planx.uk/helpers.ts
```diff
@@ -66,55 +66,96 @@ interface CreateFlowResponse {
...
+    flow: insert_flows_one(
+      object: {
+        team_id: $team_id
+        slug: $slug
+        name: $name
+        data: $data
+        version: 1
+        copied_from: $copied_from
+        templated_from: $templated_from
+        is_template: $is_template
```
jessicamcinchak, line 113: If this is the `createSourceTemplateMutation`, does `is_template` actually have to be passed in as an argument or can we always default to `true`? 

Would this help address the pattern below at all? I'm thinking something like:
- `createFlow` still accepts an `isTemplate` argument
- But we keep a single mutation and simply set the correct `isTemplate` graphql param ourselves? (defaulting to false?)

Does that still work with Hasura permissions to restrict creation of source templates to platform admins?

editor.planx.uk/src/pages/Team/components/AddFlow/types.ts
```diff
@@ -1,3 +1,4 @@
+import { boolean } from "mathjs";
```
jessicamcinchak, line 1: Is the right import? I'd expect `import { number, object, string, boolean } from "yup";` ?

editor.planx.uk/src/pages/Team/components/FlowCard.tsx
```diff
@@ -145,6 +145,7 @@ const FlowCard: React.FC<FlowCardProps> = ({
 
   const isSubmissionService = flow.publishedFlows?.[0]?.hasSendComponent;
   const isTemplateService = Boolean(flow.templatedFrom);
+  const isSourceTemplateService = Boolean(flow.isTemplate);
```
jessicamcinchak, line 148: nit: `flow.isTemplate` is naturally type boolean, do we actually need to wrap it in `Boolean()` again here?

jessicamcinchak (approved):
Thanks for picking this one up - works as expected, a few minor comments !

api.planx.uk/helpers.ts
DafyddLlyr: > But we keep a single mutation and simply set the correct isTemplate graphql param ourselves? (defaulting to false?)

This was what I wanted to go for and tried first of all - non `platformAdmins` can't even write `false` to this column however, it needs to be a column they can't interact with. I tried a few GraphQL approaches with `@include` for this variable and didn't manage to get a working version without two different mutations.

The other approach here could be either - 
 - we allow anybody to write to this column (not ideal, but is it actually that risky?)
 - we allow anybody to write to the column, but then validate their permission after with a post-update check (would work, but a fair bit more complex)

editor.planx.uk/src/pages/Team/components/AddFlow/types.ts
DafyddLlyr: Good spot! I've auto imported the wrong thing here

api.planx.uk/helpers.ts
jessicamcinchak: Thanks for explanation ! I'd say either accept the risk of any role writing to column (we know that non-OSL authored templates are probably a near future feature anyways - eg RAB), or keep this approach for now (seems better than more complex post-update check?)

editor.planx.uk/src/pages/Team/components/FlowCard.tsx
DafyddLlyr: Good catch (again! 😅), looks like a heavy-handed `Cmd + Shift + D` on my end here!

api.planx.uk/helpers.ts
DafyddLlyr: Yep good shout on non-OSL templates, updated permissions and simplified mutation - much better!

PR #4653, opened 2025-05-06
fix: Headings in rich text editor

ianjon3s:
## Identified issue

Currently in the rich text editor we do the following:
- Automatically increment `<h1>` to `<h2>` and `<h2>` to `<h3>` for all instances of the rich text editor except for the Content component.

In addition to this we enforce the following rules:
- The level 1 heading must come first in the document.
- There cannot be more than one level 1 heading in the document.
- A level 1 heading must come before a level 2 heading.

In the cases where we convert `<h1>` to `<h2>` the first two of these rules do not apply, as a `<h2>` does not need to be first in the document and it can appear in multiple instances.

Additionally, allowing giving editors the ability to place (or at least expect to be able to place) a `<h1>` within in a document that already has a heading is bad practice (question descriptions being a prime example of this).

## Solution

I've taken the following steps to rectify this issue:
- Remove the automatic incrementation of heading levels.
- Configure the rich text editor to have default options for `<h2>` and `<h3>`, with a prop `allowH1` to enable `<h1>` (used in content component).
- Adjust the rule `A level 1 heading must come before a level 2 heading.` to only be applied if a `<h1>` is present.
- Introduce the rule `A level 2 heading must come before a level 3 heading.` if a `<h3>` is present.

As the headings were previously assigned there should be no impact on the current set up for headings (will test).

**Testing:**

Two content components, the first has no violations and the second has multiple:
https://4653.planx.pizza/testing/heading-testing


ianjon3s:
Closing as issue addressed by https://github.com/theopensystemslab/planx-new/pull/4672

PR #4654, opened 2025-05-07
fix: Use Vite env var, add eslint rule

DafyddLlyr:
## What does this PR do?
 - Uses `import.meta.env` to access env vars instead of `process.env` - this is required for Vite applications ([docs](https://vite.dev/guide/env-and-mode#env-variables-and-modes))
 - Adds eslint rule so we shouldn't make this mistake again!

### DIFF
--- a/editor.planx.uk/.eslintrc
+++ b/editor.planx.uk/.eslintrc
@@ -104,6 +104,13 @@
           "expect"
         ]
       }
+    ],
+    "no-restricted-syntax": [
+      "error",
+        {
+          "selector": "MemberExpression[object.name='process'][property.name='env']",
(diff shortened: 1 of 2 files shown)


editor.planx.uk/.eslintrc
```diff
@@ -104,6 +104,13 @@
           "expect"
         ]
       }
+    ],
+    "no-restricted-syntax": [
+      "error",
+        {
+          "selector": "MemberExpression[object.name='process'][property.name='env']",
+          "message": "Use import.meta.env instead of process.env in Vite applications"
+        }
```
DafyddLlyr, line 113: image

ianjon3s (approved):

PR #4656, opened 2025-05-07
feat: setup event trigger to update `templated_flow_edits`

jessicamcinchak:
Stubs out a new database event trigger that listens for updates on `flows.data`. If that flow is templated from another, it's going to diff the updated `data` against the source template's `data`, filter out differences to nodes tagged "customisation", then format and save them to `templated_flow_edits` table. This code doesn't actually do the logical second part yet, it's just the skeletal event trigger setup and request validation.

I think there might be benefit in merging just the skeletal part here early to monitor performance? This event trigger will be far more frequent than our other uses so far. Then I'll come back and fill in the logic soon.

**To test:** 
- Update a flow that is not templated from another, check the Hasura events log to confirm an event was triggered. It should have succeeded (200) with message `"Not a templated flow, skipping updates"`. 
- Update a flow that is templated from another, check the Hasura events log to confirm an event was triggered. It should have succeeded (200) with response `{ flowId, templatedFrom, data }`.

### DIFF
--- a/api.planx.uk/modules/webhooks/controller.ts
+++ b/api.planx.uk/modules/webhooks/controller.ts
@@ -1,21 +1,24 @@
+import isNull from "lodash/isNull.js";
 import { ServerError } from ".././../errors/index.js";
-import type { CreateSessionEventController } from "./service/lowcalSessionEvents/schema.js";
+import { analyzeSessions } from "./service/analyzeSessions/index.js";
 import {
   createSessionExpiryEvent,
   createSessionReminderEvent,
 } from "./service/lowcalSessionEvents/index.js";
-import type { SendSlackNotification } from "./service/sendNotification/types.js";
-import { sendSlackNotification } from "./service/sendNotification/index.js";
-import type { CreatePaymentEventController } from "./service/paymentRequestEvents/schema.js";
+import type { CreateSessionEventController } from "./service/lowcalSessionEvents/schema.js";
 import {
   createPaymentExpiryEvents,
   createPaymentInvitationEvents,
   createPaymentReminderEvents,
 } from "./service/paymentRequestEvents/index.js";
-import type { SanitiseApplicationData } from "./service/sanitiseApplicationData/types.js";
+import type { CreatePaymentEventController } from "./service/paymentRequestEvents/schema.js";
 import { sanitiseApplicationData } from "./service/sanitiseApplicationData/index.js";
+import type { SanitiseApplicationData } from "./service/sanitiseApplicationData/types.js";
+import { sendSlackNotification } from "./service/sendNotification/index.js";
+import type { SendSlackNotification } from "./service/sendNotification/types.js";
+import type { UpdateTemplatedFlowEditsController } from "./service/updateTemplatedFlowEdits/schema.js";
 import type { IsCleanJSONBController } from "./service/validateInput/schema.js";
-import { analyzeSessions } from "./service/analyzeSessions/index.js";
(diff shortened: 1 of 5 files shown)


api.planx.uk/modules/webhooks/controller.ts
```diff
@@ -188,3 +191,33 @@ export const isCleanJSONBController: IsCleanJSONBController = async (
...
+export const updateTemplatedFlowEditsController: UpdateTemplatedFlowEditsController =
+  async (_req, res, next) => {
+    const { flowId, templatedFrom, data } = res.locals.parsedReq.body.payload;
+
+    try {
+      if (isNull(templatedFrom)) {
+        return res.status(200).send({
+          message: `Not a templated flow, skipping updates (${flowId})`,
+        });
+      }
```
jessicamcinchak, line 204: Ideally, we'd be able to define a trigger that listens for updates on `flows.data` column only **`where flows.templated_from is not null`** - but Hasura doesn't support the second condition. 

So this means we'll have an event for _every_ flow update independent of whether or not it's a template.

At best, we can ensure that `templatedFrom` is directly passed through the request body so that we can exit early without making any additional requests.

api.planx.uk/modules/webhooks/service/updateTemplatedFlowEdits/index.ts
```diff
@@ -0,0 +1,14 @@
...
+  flowId: string,
+  templatedFrom: string,
+  data: any,
+) => {
+  // TODO
+  // 1. Fetch flows.data where id = templatedFrom
+  // 2. JSON diff against event `data`
+  // 3. Filter diff for nodes tagged 'customisation'
+  // 4. Transform diff to `templated_flow_edits` data shape
+  // 5. Update `templated_flow_edits`
```
jessicamcinchak, line 11: Intentional TODOs ! For now, this function simply returns the arguments passed to it.

jessicamcinchak:
[[Implementation] Make sure live services are always up to date (aka flows copied from Templates should receive updates)](https://trello.com/c/huXP4VBO/2859-implementation-make-sure-live-services-are-always-up-to-date-aka-flows-copied-from-templates-should-receive-updates)

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -1226,6 +1226,37 @@
...
         version: 2
+    - name: update_templated_flow_edits
+      definition:
+        enable_manual: false
+        update:
+          columns:
+            - data
+      retry_conf:
+        interval_sec: 10
+        num_retries: 0
```
jessicamcinchak, line 1237: Again intentionally setting 0 retries here - as this will be a very frequent trigger, think it's safe to assume that if one fails, updates _should_ simply be picked up by the next successful event because it'll always do a full data diff.

DafyddLlyr (approved):
Great! Good to get the simple scaffolding in place here.

Working as expected 👍

api.planx.uk/modules/webhooks/routes.ts
```diff
@@ -62,6 +64,12 @@ router.post(
   isCleanJSONBController,
 );
 
+router.post(
+  "/webhooks/hasura/update-templated-flow-edits",
+  validate(updateTemplatedFlowEditsEventSchema),
+  updateTemplatedFlowEditsController,
+);
```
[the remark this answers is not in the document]
jessicamcinchak: Yep good note - will pick up once final response shape is more confirmed :heavy_plus_sign:

PR #4663, opened 2025-05-08
feat: Rename flow dialog

DafyddLlyr:
## What does this PR do?
- Replaces prompt with a form and dialog on the "rename" action
- Improves error handling
- Trims flow names on creation

Issues raised here (OSL Slack) - https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1746617788647349 

image
image

### DIFF
--- a/editor.planx.uk/src/pages/Team/components/RenameDialog.tsx
+++ b/editor.planx.uk/src/pages/Team/components/RenameDialog.tsx
@@ -18,8 +18,8 @@ import { slugify } from "utils";
 import { object, string } from "yup";
 
 interface RenameFlow {
-  name: string;
-  slug: string;
+  flowName: string;
+  flowSlug: string;
 }
 
 interface Props {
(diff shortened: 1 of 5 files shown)


editor.planx.uk/src/pages/Team/components/RenameDialog.tsx
```diff
@@ -0,0 +1,163 @@
...
+    name: string,
+    slug: string,
+    id: string,
+  }
+}
+
+const validationSchema = object().shape({
+  slug: string().required("Slug is required"),
+  name: string().required("Name is required")
+});
```
jessicamcinchak, line 38: nit: do these schemas also need `.trim()` to trigger error on empty strings?

Can currently enter an empty string and click "Rename" and it just sort of freezes, no visible error
Screenshot from 2025-05-08 16-54-13

jessicamcinchak (approved):
Looks good! One small edge casey bug noted below

PR #4667, opened 2025-05-08
fix: Trim flow name before adding ' (copy)' suffix

DafyddLlyr:
Based on feedback by @augustlindemer - 

> _for any flow currently with a trailing whitespace, when the copy dialogue appends  (copy) it does so without trimming the whitespace first, so it creates a flow with two whitespaces before the (copy)._

| Before | After |
|--------|--------|
| image | image |


augustlindemer (approved):
nice and clean!

PR #4668, opened 2025-05-09
fix: handle incoming value = 0 correctly in Calculate, don't fallback to default value

jessicamcinchak:
See thread https://opensystemslab.slack.com/archives/C07F7215W7P/p1746708641958439?thread_ts=1746524633.032529&cid=C07F7215W7P

**To test:** 
- Compare same flow on this pizza to production flow https://editor.planx.uk/testing/calculator-defaults
- If you have 0 apples, you should need 0 containers
  - Production flow will incorrectly output 3 containers in passport when Calculate has default value of 10 and receives a valid incoming value of 0 apples (default should only be triggered when incoming apples are undefined)
- From what I can tell only impacting passport & `handleSubmit`, not the "Try" section of the editor modal

### DIFF
--- a/editor.planx.uk/src/@planx/components/Calculate/model.test.ts
+++ b/editor.planx.uk/src/@planx/components/Calculate/model.test.ts
@@ -82,6 +82,13 @@ describe("evaluate", () => {
     expect(expected).toEqual(actual);
   });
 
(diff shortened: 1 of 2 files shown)


augustlindemer (approved):
Works!

PR #4669, opened 2025-05-09
fix: Race condition between routes and `initUserStore()`

DafyddLlyr:
## What's the problem?
There's a race condition between routes loading (and their associated data fetches and renders) and the `UserStore` being initialised. We need the `UserStore` to be initialised for route guards to work, and as `react-navi` sits outside it react rendering hierarchy it doesn't re-render / update when Zustand stores do.

What's happening currently - 
 - User loads page
 - (Concurrently) Fetch user, load data (e.g. for `/admin-panel`)
 - If user is not returned yet, route guard fails

This behaviour is most commonly seen of refreshes and when direct linking to a nested route - simply as the home page does not have a route guard build into `react-navi`. The user request succeeds and is saved to the store before nested, guarded, routes are loaded.

## What's the solution?
The user store was being incorrectly initialised within a `View` component, which despite being hierarchical from a DOM point of view, is not awaited by `react-navi`.

By moving the call to `initUserStore()` into react-navi's context, we can access the `user` within the route for our route guards. This is then correctly tied to the react-navi lifecycle.

### To test...
- Navigate to `/admin-panel` from home
- Refresh (try it many times, should always resolve and not throw a `NotFound` error)
- Navigate directly to `/admin-panel` without first landing on home - https://4669.planx.pizza/admin-panel

**Note:** This does not yet resolve the issue outlined in [this `#planx-bugs` thread](https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1746711213940839) (OSL Slack) - this appears to be a very similar issue related to the `TeamStore`.

### DIFF
--- a/editor.planx.uk/src/routes/authenticated.tsx
+++ b/editor.planx.uk/src/routes/authenticated.tsx
@@ -1,4 +1,4 @@
-import type { Team, TeamSettings, TeamTheme } from "@opensystemslab/planx-core/types";
+import type { Team, TeamSettings, TeamTheme, User } from "@opensystemslab/planx-core/types";
 import gql from "graphql-tag";
 import {
   compose,
@@ -7,6 +7,7 @@ import {
   mount,
   NotFoundError,
   route,
+  withContext,
   withView,
 } from "navi";
 import { useStore } from "pages/FlowEditor/lib/store";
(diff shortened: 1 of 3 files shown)


DafyddLlyr:
`TeamStore` issue now addressed by https://github.com/theopensystemslab/planx-new/pull/4671

editor.planx.uk/src/routes/authenticated.tsx
```diff
@@ -26,9 +27,22 @@ export type TeamSummary =
...
   withView(authenticatedView),
 
+  /**
+   * Ensure userStore is initialised
+   * Required for route guards to avoid race conditions
+   */
+  withContext(async (): Promise<Context> => {
+    const user = await useStore.getState().initUserStore();
+    return { user }
+  }),
```
jessicamcinchak, line 44: :raised_hands: simple solution, very effective !

jessicamcinchak (approved):

editor.planx.uk/src/routes/authenticated.tsx
DafyddLlyr: I know....! 🤦‍♂️ Spent a long time to get these few lines!

PR #4672, opened 2025-05-09
refactor: Setup, validation and tests for rich text headings

ianjon3s:
# What does this PR do?

Address a number of issues and inconsistencies in the rich text editor with regards to headings.

[A first attempt at this](https://github.com/theopensystemslab/planx-new/pull/4653) was deemed unfeasible as it would require a database update to convert instances of `<h1>` to `<h2>` and `<h2>` to `<h3>` (for cases where we modify these levels on render).

This is a cleaner solution that will not require any database modifications.

**Testing:**
https://4672.planx.pizza/testing/content-validation

## Issue: Heading levels selected in the editor do not match the output

Currently for all cases except the Content component (where levels are not modified on render). If we're giving best practice / accessibility advice for setting heading levels (`h1` should be first in the document etc) to editors the we need to be consistent in the heading levels we allow them to author.

## Solution: Present options for heading levels that reflect the output

Setting a prop of `rootLevelContent` that can be set for any content that appears without a preceding `<h1>`, and if this prop isn't present give options for `<h2>` and `<h3>`.

**Root-level content (h1 + h2):**
image

**Non root-level content (h2 + h3):**
image

## Issue: Some hierarchy rules are applied in error

For cases where we convert heading levels, the following rules are given in error:
`There cannot be more than one level 1 heading in the document.`
`The level 1 heading must come first in the document.`

These are falsely being applied to instances of `<h2>` after we convert the levels.

## Solution: Refactor rules so that they apply for both root-level and non root-level content

Testing:
https://4672.planx.pizza/testing/content-validation/nodes/_root/nodes/TGckqOuWto/edit

Additional rule created for root-level content:
`The document must start with a level 1 heading (H1).`
Rather than stating a h1 must be first only if a h1 is present but not first, require that a h1 is both present and first in the doc.

## Issue: Error popover blocks editing of content

When the popover is present it makes editing of content difficult as it prevents clicking into the rich text window:
image

## Solution: Display error messages inline

This also allows for multiple messages to be display (not possible with popover):
image


editor.planx.uk/src/ui/editor/RichTextInput/components/InlineError.tsx
```diff
@@ -0,0 +1,18 @@
...
+  padding: theme.spacing(1),
+}));
+
+export const InlineError: FC<{ id: string; error: string }> = ({ id, error }) => {
+  return (
+    <Root id={id}>
+      <Typography variant="body1" color="error" fontWeight={FONT_WEIGHT_SEMI_BOLD}>{error}</Typography>
+    </Root>
+  );
+};
```
jessicamcinchak, line 18: nit: instead of a new `InlineError` component, can we re-use existing `ErrorWrapper` here (already in-use in many editor modals)? 

There's also some styling inconsistencies introduced here (should be automatically cleared up and irrelevant if can re-use `ErrorWrapper`):
- I'd expect the red left-hand border to extend _around_ the input and any error messages
- I'd expect errors to display _above_ the input rather than below it

jessicamcinchak (commented):
Overall seems like a much tidier & effective solution!! One initial comment, need to spend a bit more time on pizza before final review - will aim to soon!

editor.planx.uk/src/ui/editor/RichTextInput/components/InlineError.tsx
ianjon3s: Thanks for this @jessicamcinchak , agree that it's better to use the existing `ErrorWrapper` rather than replicate / imitate with a similar component.

Longer term I'll keep thinking on this, as I don't think `ErrorWrapper` is necessarily the ideal solution, due to:
- Causing layout shift in the text input while typing. Likely this would cause us to fair usibility testing if used in a public service
- We're not actually showing blocking errors in the same way that we usually use error messaging, rather validation warnings (as user can save despite errors being in place.

This raises a bigger question: we should consider enforcing these as blocking errors for the sake of continual a11y compliance.

editor.planx.uk/src/ui/editor/RichTextInput/validationHelpers.ts
```diff
@@ -100,39 +100,57 @@ export const linkSelectionError = (selectionHtml: string): string | null => {
...
         h2Index = index;
-        return;
+      }
+    } else {
+      if (level === 2) {
+        if (h2Index === -1 && h1Index === -1) {
+          h2Index = index;
+        }
+        if (h1Index === -1) {
+          errors.push("A level 2 heading (H2) must come before a level 3 heading (H3).");
```
jessicamcinchak, line 145: Unexpectedly seeing this H2 & H3 error when I only have H1 & H2 available: https://4672.planx.pizza/testing/content-validation/nodes/_root/nodes/TGckqOuWto/edit

Screenshot from 2025-05-13 15-27-11

jessicamcinchak (approved):
Think this is a really nice solution ! One unexpected error message noted below, but otherwise happy for this to merge :heavy_plus_sign: 

Can we close #4653 now as well?

editor.planx.uk/src/ui/editor/RichTextInput/validationHelpers.ts
jessicamcinchak: Perfect, thanks! :white_check_mark:

PR #4675, opened 2025-05-12
Production deploy

jessicamcinchak:
The planx-core bump & #4674 fix a buggy schema/submission edge case & should get to prod ! 

#4656 has been performing fine on staging all day (see staging Hasura event logs for "update_templated_flow_edits" trigger), so no concerns about going to prod.


ianjon3s (approved):
Sorry I'd not spotted this one waiting earlier.

When merged let's keep an eye on the watermarks on production!

PR #4677, opened 2025-05-13
feat: Watermark on production draft and preview routes

ianjon3s:
## Issue

The testing watermark is currently only visible on non-`.uk` platforms, it needs to be visible on `.uk` for `draft` and `preview` public urls.

## Solution

- Introduce a


editor.planx.uk/src/ui/shared/WatermarkBackground.tsx
```diff
@@ -65,7 +68,7 @@ const WatermarkBackground: React.FC<WatermarkBackgroundProps> = ({
         permanent content changes.
       </Box>
       <Wrapper>
-        <Background variant={variant} opacity={opacity} />
+        <Background variant={variant} opacity={opacity} forceVisibility={forceVisibility ?? false} />
```
jessicamcinchak, line 71: Since we're passing the default `= false` above on L57, I don't think we actually need `?? false` here?

jessicamcinchak (approved):
This looks good to me! Will make a mental note to immediately check after next prod deploy and can rollback easily if needed :heavy_plus_sign:

editor.planx.uk/src/ui/shared/WatermarkBackground.tsx
ianjon3s: Good spot, indeed we don't, I've removed 👍

PR #4678, opened 2025-05-14
feat: update `templated_flow_edits` when node properties in a templated flow are updated

jessicamcinchak:
**Key changes:** 
- [Hasura] Updates `templated_flow_edits` table schema and permissions
- [API] On `update_templated_flow_edits` event trigger, does actual diff logic to record which properties of which node ID have been updated
- [Editor] Introduces a "Customise" sidebar tab which acts as your to-do list of each node tagged "customisation" that you need to update. Currently styled to match tag on first load, then grey with a :heavy_check_mark: once edited - styles intentionally bare & rough here ! They're behind a feature flag and can be properly designed in a follow-up PR

Screenshot from 2025-05-20 15-39-04

**Testing:**
- Toggle on `window.featureFlags.toggle("TEMPLATES")` on this pizza
- Create a new flow from a template (there's aleady a demo "source template" available)
- Confirm the graph does not display the "hangers" (circles which allow you to drag or add new nodes)
- Confirm the sidebar defaults to "Customise" tab and displays each node tagged "customisation" in order of graph (top-down, left-right)
- Confirm that you can edit customisable nodes and see the sidebar items automatically get a green tick

**Next:**
- We'll use the `templated_flow_edits.data` structure on publish events to reconcile changes!


editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
```diff
@@ -220,12 +235,20 @@ const Sidebar: React.FC = React.memo(() => {
           </Header>
           <TabList>
             <Tabs onChange={handleChange} value={activeTab} aria-label="">
+              {hasFeatureFlag("TEMPLATES") && isTemplatedFrom && (
+                <StyledTab value="Customise" label="Customise" />
+              )}
```
jessicamcinchak, line 240: "Customise" language very much still to-review in a future design call, but safely feature-flagged for now

hasura.planx.uk/metadata/tables.yaml
```diff
@@ -3095,19 +3095,28 @@
       using:
         foreign_key_constraint_on: flow_id
   insert_permissions:
+    - role: api
```
jessicamcinchak, line 3098: Because we update `templated_flow_edits` via a Hasura event webhook, we're relying on `api` permissions here now rather than `teamEditor` or otherwise

hasura.planx.uk/migrations/1747216207505_alter_table_public_templated_flow_edits_add_unique_flow_id/up.sql
```diff
@@ -0,0 +1,4 @@
+alter table "public"."templated_flow_edits" drop constraint "templated_flow_edits_node_id_flow_id_key";
+alter table "public"."templated_flow_edits" add constraint "templated_flow_edits_flow_id_key" unique ("flow_id");
+alter table "public"."templated_flow_edits" drop column "node_id" cascade;
+alter table "public"."templated_flow_edits" rename column "node_data" to "data";
```
jessicamcinchak, line 4: Flip-flopped on the table structure again! Dropped column `node_id` so table is now unique based on `flow_id` alone and we just record a single `data` jsonb (this has benefit of capturing changes to Node `edges` as well as `data`!)

api.planx.uk/modules/webhooks/service/updateTemplatedFlowEdits/index.ts
```diff
@@ -1,14 +1,77 @@
...
+
+  // For each changed node property in the delta, extract its' current data state only
+  //   to store in templated_flow_edits (delta returns before and after state)
+  const templatedFlowEditsData: Record<NodeId, Partial<Node>> = {};
+  Object.entries(delta).forEach(([nodeId, nodeData]) => {
+    const updatedNodeData: Node["data"] = {};
+    const updatedKeys: string[] = nodeData?.data && Object.keys(nodeData.data);
+    updatedKeys?.forEach((updatedKey) => {
+      // TODO check how this handles removing properties eg 'description'
+      updatedNodeData[updatedKey] = data[nodeId]?.["data"]?.[updatedKey];
```
jessicamcinchak, line 33: Known TODO - let's come back to this one after we get more clarity and realistic examples of customisable nodes I think, unclear still how much of an issue/edge-case tracking deleted fields would actually be !

editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
```diff
@@ -274,7 +274,7 @@ const FormModal: React.FC<{
                   makeUnique(id, parent);
                   navigate(rootFlowPath(true));
                 }}
-                disabled={disabled}
+                disabled={disabled || (isTemplatedFrom && isCustomisableNode(node))}
```
jessicamcinchak, line 277: When a node is tagged "customisation", we want it to be editable and you need to be able to "update" it, but you shouldn't be able to "delete" or "make unique" - these extra conditions account for that

jessicamcinchak: Open question here I suppose if we're happy with these buttons being disabled, or we could also completely hide them on the second condition !

DafyddLlyr (approved):
Code looking great, no big comments below. Amazing how it's all fit into such a small PR 😅 

I've not tested the Pizza as I'm not able to toggle feature flags! Tried across browsers and also on `.dev`. I get the following error - 

image

I'm also not seeing the usual `🎏 2 feature flags enabled: XX` log when the site is loaded either.

This regression isn't introduced in this PR, I'm happy to take a look at this issue just now - any idea if anything here has changed recently?

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Customisations/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+      <List>
+        {sortedCustomisableNodeIds.map((nodeId) => (
+          <ListItem
+            key={nodeId}
+            sx={{
+              display: "list-item",
+              backgroundColor: (theme) =>
+                Object.keys(edits)?.includes(nodeId)
+                  ? theme.palette.background.paper
+                  : "#FAE1B7", // same color as "customisation" tag for now
```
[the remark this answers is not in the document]
jessicamcinchak: :bulb: thanks for highlighting this is already encoded in theme - will also fix hardcoded values elsewhere that I copied this from !

api.planx.uk/modules/webhooks/service/updateTemplatedFlowEdits/index.ts
```diff
@@ -1,14 +1,77 @@
+import type { Node, NodeId } from "@opensystemslab/planx-core/types";
+import { gql } from "graphql-request";
+import * as jsondiffpatch from "jsondiffpatch";
+import isEmpty from "lodash/isEmpty.js";
+import { $api } from "../../../../client/index.js";
+import { getFlowData } from "../../../../helpers.js";
+import type { Flow } from "../../../../types.js";
+
 export const updateTemplatedFlowEdits = async (
```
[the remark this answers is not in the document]
jessicamcinchak: Good shout! Moved the logic bit out to a `helpers.ts` so we can independently mock a delta & flow data and test the data transformation separate from full request. Tests are still `.todo` and will come back too soon - going to try to keep end-to-end momentum :heavy_plus_sign:

PR #4679, opened 2025-05-15
chore: add `flow_id` to `analytics_summary` and `submission_services_summary`

zz-hh-aa:
Generates migrations to add column `flow_id` to the above table views. 

This is because we want to experiment with filtering analytics dashboards on `flow_id` instead of `service_slug` for continuity (eg we would still show the same data in analytics regardless of whether or not the service name / slug changed). See conversation in other [analytics PR](https://github.com/theopensystemslab/planx-new/pull/4648#pullrequestreview-2828833546) and in [Slack](https://opensystemslab.slack.com/archives/C5Q59R3HB/p1746698531200869).


jessicamcinchak (approved):
Looks good :star2:

PR #4692, opened 2025-05-20
fix: Handle first publish of flows with no history

DafyddLlyr:
## What's the problem?
An error is thrown when an unpublished flow is first published.

image

## What's the cause?
The `GetHistory` query requires a `last_published_at` timestamp which does not exist on unpublished flows.

## What's the solution?
Conditionally return `null` from `getHistory()` if there's no `last_published_at` timestamp.

### DIFF
--- a/api.planx.uk/modules/flows/validate/validate.test.ts
+++ b/api.planx.uk/modules/flows/validate/validate.test.ts
@@ -54,7 +54,7 @@ beforeEach(() => {
     matchOnVariables: false,
     data: {
       flow: {
-        publishFlows: [
+        publishedFlows: [
           {
             createdAt: "2024-12-31",
           },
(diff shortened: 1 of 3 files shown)


api.planx.uk/modules/flows/validate/validate.test.ts
```diff
@@ -54,7 +54,7 @@ beforeEach(() => {
     matchOnVariables: false,
     data: {
       flow: {
-        publishFlows: [
+        publishedFlows: [
```
DafyddLlyr, line 57: Typo in mock queries was causing other tests to fail with the updated logic. This now correctly
 matched the `GetMostRecentPublishedFlowVersion` query.

https://github.com/theopensystemslab/planx-new/blob/main/api.planx.uk/helpers.ts#L225-L234

jessicamcinchak (approved):
Thanks for fix here ! 

A bit confused about context though in description - all _new_ flows are (or should be) published on creation, is this only re-creatable for flows that were created before that was changed and have never been published until now?

DafyddLlyr:
Yeah that's right, clumsy description here on my part.

[Here's the flow in question on prod](https://hasura.editor.planx.uk/console/data/default/schema/public/tables/flows/browse?filter=id%3B%24eq%3B73e38934-1afa-4db1-b76d-e109c06f03b1), with no associated published flow. The flow was created in May, and the auto-publishing [was added in February](https://github.com/theopensystemslab/planx-new/pull/4230). 🤔 

I can't spot an issue in the code (all requests go via API and include a publish step, including copy). There's a few others which match this pattern - 

```sql
SELECT f.*
FROM flows f
LEFT JOIN published_flows pf ON f.id = pf.flow_id
WHERE pf.flow_id IS NULL
ORDER BY f.created_at desc
```

PR #4695, opened 2025-05-21
fix: Exclude `forceVisibility` prop from the DOM

DafyddLlyr:
image


ianjon3s (approved):
Thanks for picking this up! I'd noticed in the console when debugging heading errors, was due to fix this week.

PR #4699, opened 2025-05-22
feat: add custom character limits to text input

augustlindemer:
This PR adds a custom character limit input to the text input component.

It removes the previous 'Default' option and defaults to 'Short'.

It adds a new 'Custom' option that displays a number input field in which the editor can define a maximum character count.

If 'Custom' is selected but no number is provided it defaults to 120 characters (same as default 'Short').

It refactors the previous condition for multiline inputs and character count from type match (long, extra long) to directly check the applicable text limit.

If text limit >120 (either through custom character limit or selection of long/extra long), it renders the text input as multiline with character count.

Note: I have not added new tests for the custom input.

### DIFF
--- a/editor.planx.uk/src/@planx/components/TextInput/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/TextInput/Editor.tsx
@@ -68,14 +68,14 @@ const TextInputComponent: React.FC<Props> = (props) => {
         </ModalSectionContent>
         <ModalSectionContent title="Input style">
           <FormControl component="fieldset">
-            <RadioGroup defaultValue="default" value={formik.values.type}>
+            <RadioGroup defaultValue="short" value={formik.values.type}>
               {[
-                { id: "default", title: "Default" },
                 { id: "short", title: "Short (max 120 characters)" },
                 { id: "long", title: "Long (max 250 characters)" },
                 { id: "extraLong", title: "Extra long (max 750 characters)" },
                 { id: "email", title: "Email" },
                 { id: "phone", title: "Phone" },
+                { id: "custom", title: "Custom" },
               ].map((type) => (
                 <BasicRadio
                   key={type.id}
@@ -88,6 +88,15 @@ const TextInputComponent: React.FC<Props> = (props) => {
                 />
               ))}
             </RadioGroup>
+            {formik.values.type === "custom" && (
+              <Input
+                placeholder="Maximum characters"
+                name="customLength"
+                value={formik.values.customLength}
+                onChange={formik.handleChange}
+                disabled={props.disabled}
+              />
+            )}
           </FormControl>
         </ModalSectionContent>
       </ModalSection>
--- a/editor.planx.uk/src/@planx/components/TextInput/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/TextInput/Editor.tsx
@@ -93,7 +93,14 @@ const TextInputComponent: React.FC<Props> = (props) => {
                 placeholder="Maximum characters"
                 name="customLength"
                 value={formik.values.customLength}
-                onChange={formik.handleChange}
+                onChange={(e) => {
+                  const value = e.target.value;
+                  if (value.startsWith("-")) {
+                    return;
+                  }
+                  formik.handleChange(e);
+                }}
+                type="number"
                 disabled={props.disabled}
               />
             )}
--- a/editor.planx.uk/src/@planx/components/TextInput/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/TextInput/Public.tsx
@@ -4,7 +4,7 @@ import { PublicProps } from "@planx/components/shared/types";
 import { useFormik } from "formik";
 import React from "react";
 import InputLabel from "ui/public/InputLabel";
-import { CharacterCounter, getTextLimit } from "ui/shared/CharacterCounter";
+import { CharacterCounter } from "ui/shared/CharacterCounter";
 import Input from "ui/shared/Input/Input";
 import InputRow from "ui/shared/InputRow";
 import { object } from "yup";
@@ -12,7 +12,7 @@ import { object } from "yup";
 import { DESCRIPTION_TEXT, ERROR_MESSAGE } from "../shared/constants";
 import { getPreviouslySubmittedData, makeData } from "../shared/utils";
 import type { TextInput } from "./model";
-import { TextInputType, textInputValidationSchema } from "./model";
+import { getTextLimit,textInputValidationSchema } from "./model";
 
 export type Props = PublicProps<TextInput>;
 
@@ -32,10 +32,8 @@ const TextInputComponent: React.FC<Props> = (props) => {
     }),
   });
 
-  const characterCountLimit = props.type && getTextLimit(props.type);
-  const displayCharacterCount = Boolean(
-    props.type !== TextInputType.Short && characterCountLimit,
-  );
+  const characterCountLimit = getTextLimit(props.type, props.customLength);
+  const displayCharacterCount = characterCountLimit > 120;
 
   return (
     <Card handleSubmit={formik.handleSubmit} isValid>
@@ -78,8 +76,8 @@ const TextInputComponent: React.FC<Props> = (props) => {
           />
           {displayCharacterCount && (
             <CharacterCounter
+              limit={characterCountLimit}
               count={formik.values.text.length}
-              textInputType={props.type || TextInputType.Long}
               error={Boolean(formik.errors.text)}
             />
           )}
--- a/editor.planx.uk/src/@planx/components/TextInput/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/TextInput/Public.tsx
@@ -12,7 +12,7 @@ import { object } from "yup";
 import { DESCRIPTION_TEXT, ERROR_MESSAGE } from "../shared/constants";
 import { getPreviouslySubmittedData, makeData } from "../shared/utils";
 import type { TextInput } from "./model";
-import { getTextLimit,textInputValidationSchema } from "./model";
+import { getTextLimit, textInputValidationSchema } from "./model";
 
 export type Props = PublicProps<TextInput>;
 
@@ -52,11 +52,9 @@ const TextInputComponent: React.FC<Props> = (props) => {
               else if (type === "phone") return "tel";
               return "text";
             })(props.type)}
-            multiline={props.type && ["long", "extraLong"].includes(props.type)}
+            multiline={getTextLimit(props.type, props.customLength) > 120}
             rows={
-              props.type && ["long", "extraLong"].includes(props.type)
-                ? 5
-                : undefined
+              getTextLimit(props.type, props.customLength) > 120 ? 5 : undefined
             }
             name="text"
             value={formik.values.text}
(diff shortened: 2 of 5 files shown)


editor.planx.uk/src/@planx/components/TextInput/Editor.tsx
```diff
@@ -88,6 +88,22 @@ const TextInputComponent: React.FC<Props> = (props) => {
...
+            {formik.values.type === "custom" && (
+              <Input
+                placeholder="Maximum characters"
+                name="customLength"
+                value={formik.values.customLength}
+                onChange={(e) => {
+                  const value = e.target.value;
+                  if (value.startsWith("-")) {
+                    return;
+                  }
```
jessicamcinchak, line 100: Currently, I can enter a negative number + press enter and the modal will simply close, then be blank when I re-open it which is a bit unexpected! 

Instead, I think we want this to be plain `onChange={formik.handleChange}` with an extra prop `errorMessage={formik.errors.customLength}`, and then add a validation schema entry to editor modal for error handling when this field changes?

See `NumberInput`'s `model.ts` & public-facing input for an example!

editor.planx.uk/src/@planx/components/TextInput/Public.tsx
```diff
@@ -32,10 +32,9 @@ const TextInputComponent: React.FC<Props> = (props) => {
     }),
   });
 
-  const characterCountLimit = props.type && getTextLimit(props.type);
-  const displayCharacterCount = Boolean(
-    props.type !== TextInputType.Short && characterCountLimit,
-  );
+  const characterCountLimit = getTextLimit(props.type, props.customLength);
+  const displayCharacterCount =
+    characterCountLimit > 120 && props.type !== "email";
```
jessicamcinchak, line 37: Much clearer :heavy_plus_sign:

editor.planx.uk/src/@planx/components/TextInput/Public.tsx
```diff
@@ -54,9 +53,13 @@ const TextInputComponent: React.FC<Props> = (props) => {
...
               return "text";
             })(props.type)}
-            multiline={props.type && ["long", "extraLong"].includes(props.type)}
+            multiline={
+              getTextLimit(props.type, props.customLength) > 120 &&
+              props.type !== "email"
+            }
             rows={
-              props.type && ["long", "extraLong"].includes(props.type)
+              getTextLimit(props.type, props.customLength) > 120 &&
```
jessicamcinchak, line 61: nit: can you reuse `characterCountLimit` definition above rather than repeating `getTextLimit(props.type, props.customLength)` here?

editor.planx.uk/src/@planx/components/TextInput/model.ts
```diff
@@ -41,25 +55,31 @@ export const textInputValidationSchema = ({
...
-          return "Enter an email address in the correct format, like name@example.com";
+          return `Enter an email address in the correct format, like name@example.com.`;
         }
-        return `Your answer must be ${TEXT_LIMITS[type]} characters or fewer.`;
+        return `Your answer must be ${limit} characters or fewer.`;
       })(),
       test: (value?: string) => {
         if (!value) return true;
         if (!type) return true;
+        if (!(value && value.length <= getTextLimit(type, customLength))) {
```
jessicamcinchak, line 72: nit: I think this can be: `(!(value && value.length <= limit))`

editor.planx.uk/src/pages/FlowEditor/ReadMePage/ReadMePage.tsx
```diff
@@ -164,7 +163,7 @@ export const ReadMePage: React.FC<ReadMePageProps> = ({
             />
             <CharacterCounter
               count={formik.values.serviceSummary.length}
-              textInputType={TextInputType.Short} // 120 characters
+              limit={120}
```
jessicamcinchak, line 166: Can we still dynamically reference the short input length here? Just in case we ever want to update 120 in future we don't have to remember to do it multiple places !

editor.planx.uk/src/ui/shared/CharacterCounter.tsx
```diff
@@ -1,32 +1,16 @@
 import Typography from "@mui/material/Typography";
 import { visuallyHidden } from "@mui/utils";
-import { TEXT_LIMITS, TextInputType } from "@planx/components/TextInput/model";
 import { debounce } from "lodash";
 import React, { useCallback, useEffect, useState } from "react";
 import { FONT_WEIGHT_SEMI_BOLD } from "theme";
 
 export type Props = {
-  textInputType: TextInputType;
+  limit: number;
```
jessicamcinchak, line 8: Retrospectively feels like a much better, more flexible approach to just pass in a `limit` here rather than type :heavy_plus_sign:

editor.planx.uk/src/@planx/components/TextInput/Editor.tsx
```diff
@@ -68,14 +68,14 @@ const TextInputComponent: React.FC<Props> = (props) => {
         </ModalSectionContent>
         <ModalSectionContent title="Input style">
           <FormControl component="fieldset">
-            <RadioGroup defaultValue="default" value={formik.values.type}>
+            <RadioGroup defaultValue="short" value={formik.values.type}>
               {[
-                { id: "default", title: "Default" },
```
jessicamcinchak, line 73: I'm not positive we can remove the `default` option here just yet because `type` is frustratingly an optional prop on node data - meaning existing "default" text input components won't know that they _should_ be "short" ! Let me do some more pizza/staging comparisons here before confirming this theory - just want to make sure we don't _break_ any existing components here !

jessicamcinchak (commented):
This is very cool ! Took a first pass and left a few initial comments & ideas below. 

Before final review & merge, let's also talk about how to update Storybook and a bit of test coverage here because you're really making quite substantial changes now :slightly_smiling_face:

editor.planx.uk/src/@planx/components/TextInput/Editor.tsx
augustlindemer: Important consideration, thanks!

I think generally it would be good to think about solutions for getting rid of the 'default' option. This is due to the existing default rendering a single line input with no character limit, which isn't really a type of input we'd want to support. Could this be one to find and replace any existing 'default' with 'short'? This may be rather rare, as far as I know we never use the default input in any of our actual service content. There may however be council led content that uses it.

editor.planx.uk/src/@planx/components/TextInput/Editor.tsx
```diff
@@ -88,6 +88,22 @@ const TextInputComponent: React.FC<Props> = (props) => {
...
+                  const value = e.target.value;
+                  if (value.startsWith("-")) {
+                    return;
+                  }
+                  formik.handleChange(e);
+                }}
+                type="number"
+                disabled={props.disabled}
+              />
+            )}
```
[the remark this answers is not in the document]
augustlindemer: Wrapped into its own box to space away from the radio button, same approach as used in expandable checklist to space out 'add new option' and 'add new group' buttons.

editor.planx.uk/src/@planx/components/TextInput/Editor.tsx
augustlindemer: Added an editorValidationSchema to the component, orienting myself on date input and number input validation approach. Now correctly calls for number input if Custom is selected but input is empty, and a character limit greater than 0 if the provided number is negative or 0. If we expand to allowing for regex too, we can condition this to require either a regex or character limit (as the latter may be superfluous if regex is provided).

editor.planx.uk/src/@planx/components/TextInput/model.ts
augustlindemer: moved limit const up in scope to do so

augustlindemer:
Noticed a bug: short text input is no longer character limit validated on submission

editor.planx.uk/src/pages/FlowEditor/ReadMePage/ReadMePage.tsx
augustlindemer: Character counter now gets character limit dynamically from TEXT_LIMITS.

I noticed that the existing code actually hardcoded the character limit in the error message despite dynamically setting the character counter:
```
serviceSummary: string().max(
        120,
        "Service description must be 120 characters or less",
```
        
This now too dynamically gets the limit:

```
serviceSummary: string().max(
        characterCountLimit,
        `Service description must be ${characterCountLimit} characters or less`,
```

editor.planx.uk/src/@planx/components/TextInput/Editor.tsx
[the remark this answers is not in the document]
ianjon3s: GOV design system have a simple pattern for nesting inside radios that we could use here:

image

Although we're appending rather than nesting structurally, it'd be a good change to make here to reenforce the relationship (and we should see about setting up proper nesting styles if it's something we'll use regularly):

image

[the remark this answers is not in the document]
ianjon3s: I did this in-browser but the styles you'll need to achieve this (after wrapping the `<Box>` in the `"custom" &&` logic):

```<Box sx={(theme) => ({ borderLeft: `4px solid ${theme.palette.border.main}`, padding: theme.spacing(1, 2), marginLeft: "13px", })}>```

**Comment by augustlindemer** [comment]
> Noticed a bug: short text input is no longer character limit validated on submission

Update: Looks to be an issue with 'Short' as a default. Values are initialised via parseTextInput, which I think means that upon node creation it has no data type to initalise and doesn't set it without any interaction.

```
export const parseTextInput = (
  data: Record<string, any> | undefined,
): TextInput => ({
  title: data?.title || "",
  description: data?.description,
  fn: data?.fn,
  type: data?.type,
  customLength: data?.customLength,
  ...parseBaseNodeData(data),
});
```

Adding a fallback to type short fixes this issue:

```
export const parseTextInput = (
  data: Record<string, any> | undefined,
): TextInput => ({
  title: data?.title || "",
  description: data?.description,
  fn: data?.fn,
  type: data?.type ?? TextInputType.Short,
  customLength: data?.customLength,
  ...parseBaseNodeData(data),
});
```

Is this an appropriate solution? Alternative is to change the initialisation in the Editor.tsx itself:

```
    initialValues: {
      ...parseTextInput(props.node?.data),
      type: TextInputType.Short,
    },
```

File: editor.planx.uk/src/@planx/components/TextInput/model.ts [hunk 9]
```diff
@@ -76,6 +115,7 @@ export const parseTextInput = (
   title: data?.title || "",
   description: data?.description,
   fn: data?.fn,
-  type: data?.type,
+  type: data?.type ?? TextInputType.Short,
```
jessicamcinchak, line 118: :heavy_check_mark: great solution to handling existing components with  now-retired "default"/undefined type !

jessicamcinchak (approved):
Apologies it's taken me a bit to get back to this one - latest modal styling looks great! Confident it's handling all cases correctly for existing & new text input components! 

I'd say let's merge this one (still may be nice to pick up storybook additions and a few fresh unit tests in a follow-up PR, but happy for these changes to merge now as-is!)

PR #4702, opened 2025-05-23
fix: Env based admin panel query

DafyddLlyr:
## What's the problem?
- The query for the `teams_summary` view coalesces staging and production values, when there can be a significant time gap between these values being populated
- This leads to the production admin panel showing the incorrect value, when only staging is popualted
- More detail here (OSL Slack) - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1747909379331239

## What's the solution?
 - Per-env columns in the `teams_summary` view (to match the data source of `team_integrations`)
 - Conditionally query different columns, based on environment
 - Still map these columns to single, simpler, `AdminPanelData` type within the Editor

### DIFF
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -3027,21 +3027,24 @@
     - role: analyst
       permission:
         columns:
-          - action_colour
-          - article_4s_enabled
-          - bops_submission_url
-          - favicon
-          - govnotify_personalisation
-          - govpay_enabled
-          - homepage
-          - id
-          - is_trial
-          - link_colour
           - live_flows
+          - article_4s_enabled
+          - govpay_enabled_production
+          - govpay_enabled_staging
+          - is_trial
+          - planning_data_enabled
+          - power_automate_enabled_production
+          - power_automate_enabled_staging
+          - id
+          - govnotify_personalisation
+          - action_colour
+          - bops_submission_url_production
+          - bops_submission_url_staging
+          - favicon
+          - homepage
+          - link_colour
           - logo
           - name
-          - planning_data_enabled
-          - power_automate_enabled
           - primary_colour
           - reference_code
           - send_to_email_address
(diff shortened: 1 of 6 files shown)


hasura.planx.uk/metadata/tables.yaml
```diff
@@ -3027,21 +3027,24 @@
...
+          - planning_data_enabled
+          - power_automate_enabled_production
+          - power_automate_enabled_staging
+          - id
           - govnotify_personalisation
-          - govpay_enabled
+          - action_colour
+          - bops_submission_url_production
+          - bops_submission_url_staging
+          - favicon
```
DafyddLlyr, line 3043: Something that took me way too long to work out and lead to me destroying, recreating, and rebasing my local env a bunch of times - if you destroy the only table where a role is used (`teams_summary` and `analyst`), then that role is also destroyed 🙃

hasura.planx.uk/migrations/1747989699392_run_sql_migration/up.sql
```diff
@@ -0,0 +1,42 @@
...
+    (
+        (ti.staging_file_api_key IS NOT NULL) AND
+        (ti.staging_power_automate_api_key IS NOT NULL) AND
+        (ti.power_automate_webhook_url IS NOT NULL)
+    )::BOOLEAN AS power_automate_enabled_staging,
+    (
+        (ti.production_file_api_key IS NOT NULL) AND
+        (ti.production_power_automate_api_key IS NOT NULL) AND
+        (ti.power_automate_webhook_url IS NOT NULL)
+    )::BOOLEAN AS power_automate_enabled_production,
```
DafyddLlyr, line 32: These are now columns per-env, with a conditional query in the frontend which means we can correctly display the values for a given environment.

hasura.planx.uk/metadata/tables.yaml
jessicamcinchak: Not expected behavior !!

jessicamcinchak (approved):
Thanks for getting to bottom of this one !

PR #4703, opened 2025-05-23
feat: add ward to property info query on `FindProperty`

jessicamcinchak:
Planning Data now publishes [wards](https://www.planning.data.gov.uk/dataset/ward) which means we can use the address point to query the ward name and add it to the passport as `property.ward`. Can merge this anytime and will capture in next schema release.

Screenshot from 2025-05-23 11-30-42

https://trello.com/c/OkSHmSQh/3295-auto-collect-ward-boundary-data


DafyddLlyr (approved):
Nice and simple! 💪

PR #4705, opened 2025-05-23
fix: Autocomplete icon alignment

ianjon3s:
## What does this PR do?

Fixes autocomplete icon across all instances of autocomplete. I'm unsure what has caused this recent styling regression, but removing all instances of `top: "unset",` on the icon element fixes the issue.

**Preview (before vs after for each):**

image

image

image

image


DafyddLlyr (approved):

PR #4707, opened 2025-05-27
feat(lps): Integrate `PostcodeSearch` Lit webcomponent

DafyddLlyr:
A few questions here...!
 - I think there was an a11y issue with the `<postcode-search/>` - is that right?
 - Styling - how easy will it be to restyle (inject classes) here?

image


jessicamcinchak (approved):
No a11y issue as far as I remember, just that we never setup styling override variables for this one so haven't properly integrated it yet !

Thanks for thinking of this too :slightly_smiling_face:

DafyddLlyr:
Ah great - much simpler then! 🤞

PR #4711, opened 2025-05-27
feat: add new responsive question and checklist component

augustlindemer:
PR as of this moment cannot render editor modal due to FormModal error.

> Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
> 
> Check the render method of `FormModal`.

Suggestions welcome!


editor.planx.uk/src/@planx/components/ResponsiveQuestion/Public.test.tsx
DafyddLlyr, line 1: Our testing library will complain (and fail) if it finds empty test files.

You can either add a placeholder `test.todo()`, or delete these files for now.

image

DafyddLlyr (approved):
One small thing!

PR #4715, opened 2025-05-28
feat: migrate flows from localStorage to IndexedDB

freemvmt:
Relates to [this Trello ticket](https://trello.com/c/vaNBw2LS). Extensive scoping notes can be found [here](https://www.notion.so/opensystemslab/Scoping-Client-side-storage-1c935d469ad180a7a5e5fd5e1653079b).

This PR adds a new file, `local.idb.ts` which mirrors `local.ts` (and therefore also `local.new.ts`) by supplying `get`, `set` and `clear` methods for storing flow sessions in [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API), as opposed to [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage).

As it stands, the previous logic remains in place as a fallback (for example, if the logic for migrating all flows across the first time a client connects with the new logic in place).

## Testing

Note that this PR currently contains lots of debug logs to assist with understanding, testing and debugging this new feature (i.e. every `console.info` log, and various `# TMP` sections -  I'll scrub all this stuff once folks are satisfied), so you should get good feedback from the dev console.

🍕 The [pizza](https://4715.planx.pizza/opensystemslab/fee-calculator/published?analytics=false) should _just work_, as usual, but with flows immediately stored in IndexedDB instead of localStorage.

🐈‍⬛  But to test the migration logic, which is crucial, we have to use a local setup, in the following fashion:

1. Change tmp constant `USE_LOCALSTORAGE_OVER_IDB_TMP` in `Questions.tsx` to `true`
2. Starting with a clean browser/application state, run the stack and navigate to a couple of published flows ([e.g.](http://localhost:3000/opensystemslab/fee-calculator/published?analytics=false))
3. Exit the browser/editor, flip the constant from (1) back to `false`, and revisit one of the published flows
4. Follow the logs in the console, and check the console/application state

You should see something like the below:

Screenshot from 2025-05-29 16-52-44

Screenshot from 2025-05-29 16-53-52


editor.planx.uk/src/lib/local.idb.ts
```diff
@@ -0,0 +1,161 @@
...
+    if (flowIds.length === 0) {
+      console.debug("No flows in localStorage to migrate");
+      return 0;
+    }
+
+    console.log(
+      `Migrating ${flowIds.length} flow(s) from localStorage to IndexedDB`,
+    );
+    const store = tx.objectStore(DB_STORE_FLOW_SESSIONS);
+    const failed_migrations = new Set<string>();
```
DafyddLlyr, line 129: nit: 
```suggestion
    const failedMigrations = new Set<string>();
```

🐍  😉

DafyddLlyr (approved):
Thanks for the detailed write up and testing notes 🏅 

Ran through this a few times locally and getting the expected results.

Code was super clear and idiomatic which was great to see, very easy to review and follow 👍

editor.planx.uk/src/lib/local.idb.ts
freemvmt: hehe nice catch, that's all the py I've been writing for the [application project](https://mnist.commune.london/) I did coming thru ;)

freemvmt:
@DafyddLlyr thanks for review here! Great that it works for you as expected 😁 

I'll leave up for @jessicamcinchak to test too, and then take some time later next week to merge it in 🧜🏼

PR #4719, opened 2025-05-29
Revert "feat: setup database event to update templated flows on source publish"

jessicamcinchak:
Reverts theopensystemslab/planx-new#4710

See Airbrake error https://planx.airbrake.io/projects/329753/groups/4078317535395680408?tab=notice-detail - 1.4k occurances on staging since yesterday.

Guessing this is the culprit of why Hasura staging has been spinning up & down overnight ! 
Screenshot from 2025-05-29 08-19-05


jessicamcinchak:
Realising that because this event is triggered on `published_flows` INSERT - it was called repeatedly _for every flow_ ~2am when our nightly staging data sync ran :see_no_evil: 

Need to rethink how to approach this one!

PR #4720, opened 2025-05-29
chore: add `flow_id` to `feedback_summary` table view

zz-hh-aa:
Forgot to include `feedback_summary` view in #4679 

These changes are:
>...because we want to experiment with filtering analytics dashboards on flow_id instead of service_slug for continuity (eg we would still show the same data in analytics regardless of whether or not the service name / slug changed). See conversation in other https://github.com/theopensystemslab/planx-new/pull/4648#pullrequestreview-2828833546 and in [Slack](https://opensystemslab.slack.com/archives/C5Q59R3HB/p1746698531200869).

### DIFF
--- a/hasura.planx.uk/metadata/tables.yaml
+++ b/hasura.planx.uk/metadata/tables.yaml
@@ -366,6 +366,15 @@
           remote_table:
             name: teams
             schema: public
+    - name: flow
+      using:
+        manual_configuration:
+          column_mapping:
+            flow_id: flow_id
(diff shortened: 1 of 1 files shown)


DafyddLlyr (commented):
Two quick questions!
 
- Do we also need to set up `flow_id` as a relationship? I would say _probably_ even if we don't have an immediate use case
 - Do we need to grant permissions for this additional column?

DafyddLlyr (approved):

PR #4732, opened 2025-05-30
feat(lps): Configure Fathom analytics (production only)

DafyddLlyr:


localplanning.services/src/components/Analytics.astro
```diff
@@ -0,0 +1,13 @@
+---
+const isProduction = import.meta.env.MODE === "production";
```
DafyddLlyr, line 2: This is read from the build script - this will only be true when `pnpm build --mode production` is called.

Docs: https://docs.astro.build/en/reference/programmatic-reference/#mode

PR #4736, opened 2025-06-01
chore: remove unused `isStatutoryApplicationType`

jessicamcinchak:
Our original implementation of this doesn't work for a number of reasons - mainly that checking flows for `application.type` data values doesn't work because "scoping" flows like `permitteddevelopment` & `london-data-hub` will alway have forking questions to triage many various types _and_ be nested in guidance services like FOI too. 

This previous implementation led to _many_ more flows than intended being tagged "isStatutoryApplicationType" and therefore we haven't referenced it in frontend editor filters, etc.

Any future implementation will likely involve a manual categorisation via flow settings - so I think it's best to remove this unused code for a fresh start and to limit complexity and responsibilities of the API's `/publish` endpoint !


jessicamcinchak:
Merging this one as E2E are passing now and all working locally :heavy_plus_sign:

PR #4737, opened 2025-06-01
fix: `lastPublished` and `lastPublisher` queries should use `flows_by_pk`

jessicamcinchak:
Minor tidy up - `flows_by_pk` is more performant than quering all `flows` with a `where` condition!


DafyddLlyr (approved):
Good catch 👍

PR #4745, opened 2025-06-01
feat: create one-off scheduled events to update templated flows when source flow is published

jessicamcinchak:
Replaces #4710 

On publish of a source flow, queue up "one-off" scheduled events to then subsequently update each templated flow. 

Update & "customisation" reconciliation logic still outstanding, this PR covers event queue only - no harm in this merging first!

Test locally by publishing a source flow and checking the Hasura "events" tab.


api.planx.uk/modules/flows/publish/controller.ts
```diff
@@ -3,18 +3,24 @@ import type { ValidatedRequestHandler } from "../../../shared/middleware/validat
...
   params: z.object({
     flowId: z.string(),
   }),
   query: z.object({
-    summary: z.string().optional(),
+    summary: z.string(),
+    templatedFlowIds: z
+      .string()
+      .optional()
+      .transform((z) => z?.split(",")),
```
DafyddLlyr, line 23: Neat!

PR #4750, opened 2025-06-03
feat: when a source template is published, merge source data with templated flow edits and update templated flows data

jessicamcinchak:
**Testing:**
- Create a source template on this pizza
- Create a templated flow from that source. Customise some nodes.
- Update and publish the source template
- Refresh your templated flow (still need to figure out subscription/live reload here bc not going through ShareDB) and ensure that your customisations are prioritised/persisted but otherwise you match the source data. There should be a single new comment in "History" explaining the update

### DIFF
--- a/api.planx.uk/modules/flows/updateTemplatedFlow/service.ts
+++ b/api.planx.uk/modules/flows/updateTemplatedFlow/service.ts
@@ -47,13 +47,13 @@ export const updateTemplatedFlow = async (
     );
   const edits = templatedFlowEditsResponse.edits[0].data;
 
-  // Apply templated flow edits on top of source data using Lodash's deep merge
+  // Apply templated flow edits on top of source data using Lodash's deep merge (order of args matters!)
   const data = merge(sourceData, edits);
 
   // Set merged data as `flows.data` for templatedFlowId
   const updateFlowResponse = await $api.client.request<UpdateTemplatedFlowData>(
     gql`
-      mutation Insert($data: jsonb!, $id: uuid!) {
+      mutation UpdateTemplatedFlowData($data: jsonb!, $id: uuid!) {
         flow: update_flows_by_pk(
           pk_columns: { id: $id }
           _set: { data: $data }


api.planx.uk/modules/flows/updateTemplatedFlow/service.ts
```diff
@@ -0,0 +1,97 @@
...
+        }
+      `,
+      {
+        flow_id: templatedFlowId,
+      },
+    );
+  const edits = templatedFlowEditsResponse.edits[0].data;
+
+  // Apply templated flow edits on top of source data using Lodash's deep merge
+  const data = merge(sourceData, edits);
```
jessicamcinchak, line 51: This is satifyingly simple and feels reassuring that using a jsonb column type for `templated_flow_edits.data` is the right call ! 

Lodash's `merge` is doing exactly what I expect here so far - and if we find out we need more flexibilty then we can look into `mergeWith` instead which offers a `customizer` param ([docs](https://lodash.com/docs/4.17.15#mergeWith))

api.planx.uk/modules/flows/updateTemplatedFlow/service.ts
```diff
@@ -0,0 +1,97 @@
...
+          id
+        }
+      }
+    `,
+    {
+      flow_id: templatedFlowId,
+      comment: `Updated based on source template publish with summary "${summary}"`,
+      actor_id: sourcePublisher,
+    },
+  );
```
jessicamcinchak, line 91: This will look like 
Screenshot from 2025-06-06 12-20-05

Maybe want some specific "type" attached to these comments in future, but will be able to identify via string "starts with" match for unique styling or similar in the short term!

api.planx.uk/modules/flows/updateTemplatedFlow/service.ts
```diff
@@ -0,0 +1,97 @@
...
+  comment: {
+    id: number;
+  };
+}
+
+export const updateTemplatedFlow = async (
+  sourceFlowId: string,
+  templatedFlowId: string,
+) => {
+  const { data: sourceData, publishedFlows } = await getFlowData(sourceFlowId);
```
jessicamcinchak, line 30: `getFlowData` here is a bit tricky - if you publish your source template, then immediately edit it, by the time the `updateTemplatedFlow` event runs (staggered by 10 secs), it could _technically_ be fetching an unpublished edit or two. 

I don't think this is particularly high risk right now, but important to be clear about. This all comes back to our flattened versus unflattened data structures and not having an easy way to do hook into ShareDB via the API, which also related to initial templated flow creation. Will come back to these once initial e2e functionality !

api.planx.uk/modules/flows/updateTemplatedFlow/controller.ts
```diff
@@ -28,8 +30,11 @@ export const updateTemplatedFlowController: UpdateTemplatedFlowController =
       res.locals.parsedReq.body.payload;
 
     try {
+      const response = await updateTemplatedFlow(sourceFlowId, templatedFlowId);
+
       res.status(200).send({
-        message: `Successfully queued up event to update ${templatedFlowId}`,
+        message: `Successfully updated templated flow on source publish (source ID ${sourceFlowId}, templated flow ID ${templatedFlowId})`,
+        data: response,
```
jessicamcinchak, line 37: TODO - we don't actually need to pass `summary` into this controller anymore, can easily include in later fetch in `service.ts`

PR #4752, opened 2025-06-04
feat: post a Slack notification when a council updates their Gov UK Notify reply-to email

jessicamcinchak:
Gov UK Notify does not have an API to automatically configure reply-to email addresses.

Councils can self-serve setting this info in the editor's team settings, but it won't correctly propagate to Notify templates leading to errors like this one which use the fallback `devops-notify@` email - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1748973868076949

There's not a good alternative to just handling the configuration manually, but this at least adds a quick Slack notification so we know when it's been updated !

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/ContactForm.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/ContactForm.tsx
@@ -1,3 +1,4 @@
+import axios from "axios";
 import { useFormik } from "formik";
 import { useStore } from "pages/FlowEditor/lib/store";
 import React, { ChangeEvent } from "react";
@@ -9,6 +10,12 @@ import { SettingsForm } from "../shared/SettingsForm";
 import { FormProps } from ".";
 
 export default function ContactForm({ formikConfig, onSuccess }: FormProps) {
+  const [teamSlug, token, helpEmail] = useStore((state) => [
+    state.teamSlug,
+    state.jwt,
+    state.teamSettings.helpEmail,
+  ]);
+
   const formSchema = Yup.object().shape({
     helpEmail: Yup.string()
       .email(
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/ContactForm.tsx
```diff
@@ -44,14 +60,40 @@ export default function ContactForm({ formikConfig, onSuccess }: FormProps) {
...
+      "templates",
+      "testing",
+      "wikihouse",
+    ];
+    if (skipTeamSlugs.includes(teamSlug)) return;
+
+    const message = `:e-mail: *${teamSlug}* updated their Gov UK Notify reply-to email to *${newEmail}*; ensure this is configured in Gov UK Notify dashboard.`;
+
+    return axios.post(
+      `${import.meta.env.VITE_APP_API_URL}/send-slack-notification`,
```
jessicamcinchak, line 77: A minor limitation of this endpoint is that is currently supports a single webhook URL / channel. 

Therefore we'll be logging this to #planx-notifications - this should only ever be council email addresses, never personal users details so I think that's fine for a shared ODP Channel, but in the future we may want to consider supporting a simple split between #planx-notifications & #planx-notifications-internal ?

editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/ContactForm.tsx
```diff
@@ -44,14 +60,40 @@ export default function ContactForm({ formikConfig, onSuccess }: FormProps) {
...
+  const sendEmailChangeSlackNotification = async (newEmail: string) => {
+    const skipTeamSlugs = [
+      "open-digital-planning",
+      "opensystemslab",
+      "planx",
+      "templates",
+      "testing",
+      "wikihouse",
+    ];
+    if (skipTeamSlugs.includes(teamSlug)) return;
```
jessicamcinchak, line 72: `skipTeamSlugs` is another classic example (like analytics, LPS, etc) where we'd benefit from a more clear condition !

editor.planx.uk/src/pages/FlowEditor/components/Settings/GeneralSettings/ContactForm.tsx
DafyddLlyr: Thanks for pointing this out, I could easily see this API endpoint taking a channel argument in future. I think this is great for now though.

DafyddLlyr (approved):

PR #4757, opened 2025-06-05
[WIP] use custom vultr caddy to serve wildcard SSL cert on pizzas

freemvmt:
Related [Trello ticket](https://trello.com/c/aiwKQ8lb).

PR to reduce certs requested from Lets Encrypt (we're starting to hit our 50/week cap) by using a wildcard.

That is, it swaps out `caddy-gen` for plain `caddy`, the latest version of which (`2.10`) will only requests one certificate for all subdomains referenced by a wildcard address (e.g. `*.4757.planx.pizza`) - see [changelog](https://github.com/caddyserver/caddy/releases/tag/v2.10.0) - so if we can make this work, it should save us 6 certs per pizza.

This is mostly a copycat of #[4748](https://github.com/theopensystemslab/planx-new/pull/4748). Props to @DafyddLlyr for most of the legwork on this 🚀 

Took some iterating, but in the process I discovered a nice command to keep an eye on the TXT records on the DNS server (which provide the _challenge_ from Lets Encrypt):

```
watch -n2 'dig +short TXT _acme-challenge.4757.planx.pizza'
```

Relatedly, this PR adds a script to delete superfluous TXT records after the fact. Run it like:

```
./scripts/vultr/delete-acme-txt.sh 4757
```

NB. I note that to SSH into the Vultr server (like `ssh root@4757.planx.pizza`) I was still having to use the pw from the Vultr portal. This might be just because I don't know the value of `SSH_PRIVATE_KEY` though? (which I assume is just the private key for the `GHA Key` I can see in the Vultr SSH panel)

NBB. The caddy image with Vultr dns plugin takes _ages_ to build, so we might want to upload the image somewhere (e.g. AWS ECR) and pull it down each time.

**Useful refs**:

- LetsEncrypt docs on the [DNS-01 challenge](https://letsencrypt.org/docs/challenge-types/#dns-01-challenge)
- Caddy's docs on their [automatic cert provision](https://caddyserver.com/docs/automatic-https)
- The [Vultr DNS plugin](https://github.com/caddy-dns/vultr) we bake into custom caddy build via `xcaddy` (we're dealing with a [pkg in beta](https://github.com/caddy-dns/vultr/issues/8#issuecomment-2927256068) here)
- Note that there might be [other ways](https://github.com/bsorahan/certbot-dns-vultr) to handle the DNS challenge (certbot)
- A [useful discussion on the propagation knobs](https://caddy.community/t/timeout-waiting-for-record-to-fully-propagate/22696/3)
- Solution to [increase buffer size](https://github.com/nextcloud/all-in-one/discussions/1970#discussion-4848534) (and related [quic-go docs](https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes)) - as a response to seeing the log `failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 7168 kiB, got: 416 kiB)` early in caddy spin-up

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -28,6 +28,7 @@ AWS_SECRET_KEY=👻
 
 # Vultr API key for pizzas to handle DNS challenge
 VULTR_API_KEY=👻
+VULTR_SSH_KEY_IDS=👻
 
 # User-uploaded S3 files require token-based access (token is used internally and shared with BOPS; local and pull request environments use the staging token)
 FILE_API_KEY=👻
--- a/.env.example
+++ b/.env.example
@@ -26,6 +26,9 @@ AWS_S3_BUCKET=👻
 AWS_ACCESS_KEY=👻
 AWS_SECRET_KEY=👻
 
+# Vultr API key for pizzas to handle DNS challenge
+VULTR_API_KEY=👻
+
 # User-uploaded S3 files require token-based access (token is used internally and shared with BOPS; local and pull request environments use the staging token)
 FILE_API_KEY=👻
 FILE_API_KEY_NEXUS=👻
--- a/ci/caddy/Caddyfile
+++ b/ci/caddy/Caddyfile
@@ -1,19 +1,21 @@
 # initial block determines global options
 {
 	email {$TLS_EMAIL}
+
 	# dev / debugging options (comment out for main / in version control)
 	debug
 	acme_ca https://acme-staging-v02.api.letsencrypt.org/directory
 }
 
-# TRY: separating out the wildcard and root domain?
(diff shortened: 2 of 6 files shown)


freemvmt:
This is not solved, but I think we've made a lot of headway and excluded some possibilities.

My current hypothesis is that the Vultr DNS module might not be keeping up its end of the bargain (esp as it's currently undergoing some development to reflect Caddy2.10/libdns1.0). I've [posted about this at length on the issue](https://github.com/caddy-dns/vultr/issues/8#issuecomment-2949531926).

PR #4760, opened 2025-06-05
feat: Support SVG file uploads

DafyddLlyr:
- Adds the ability to upload SVGs via frontend and API
- More explicit test for filetypes

### DIFF
--- a/api.planx.uk/modules/file/file.test.ts
+++ b/api.planx.uk/modules/file/file.test.ts
@@ -151,7 +151,7 @@ describe("File upload", () => {
         });
     });
 
-    it("should upload file", async () => {
+    it("should upload JPG", async () => {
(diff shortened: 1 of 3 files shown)


ianjon3s (approved):
SVG uploads are not working for component images 👍

Not directly related to this update, but possibly something that needs addressing separately: PDF uploads are not working for (for example) question component images.

DafyddLlyr:
> Not directly related to this update, but possibly something that needs addressing separately: PDF uploads are not working for (for example) question component images.

This is intentional - in these instances we only allow images to be uploaded. 

https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/src/ui/shared/PublicFileUploadButton.tsx#L15-L17

PR #4763, opened 2025-06-05
feat: automate analytics link creation in frontend

zz-hh-aa:
# What does this PR do?
- Moves all Metabase automation to the frontend
    - Deletes everything in `api/` and assorted Hasura data (permissions, `flows.analytics_link` column)
    - Generates link based on `flows.status`

## Testing
To test this, create a new flow (or go to an existing offline flow) in the editor. Toggle to `online` and check analytics link (after refreshing!)--this should now be active and take you to an empty Metabase staging dashboard (with a flow ID filter correctly set). 

# Why?
Since we didn't actually need to store Metabase values in the database anymore, the automation didn't need to be in `api/`.

Pleasing bonus: generating links in the frontend solves a maintenance problem involving keeping track of which services should have analytics links and when they were updated. If template dashboards change in the future, we don't have to update the db column per-service, we can just change how the link is generated 🎆 Thanks @DafyddLlyr for the suggestion! 

# Questions
1. Is all of the code I've added in the right place? (`flowEditor.tsx`)
2. Are we okay with services losing analytics links if their services go back offline at some point? Obviously no data or dashboard is lost, but since link generation is only 
___

Potential multiple birds with one stone here. Merging this could not only close the ["Move analytics to frontend"](https://trello.com/c/dz09HipW/3306-discovery-move-analytics-automation-to-frontend-no-planx-db) card, but also the ["Handle changing slugs"](https://trello.com/c/dz09HipW/3306-discovery-move-analytics-automation-to-frontend-no-planx-db) card. We could also close #4735.


editor.planx.uk/src/routes/views/flowEditor.tsx
```diff
@@ -52,6 +52,46 @@ interface GetFlowEditorData {
...
+const RAB = {
+  id: DASHBOARD_PUBLIC_IDS[environment].RAB,
+  slugs: ["report-a-planning-breach"],
+};
+const submission = {
+  id: DASHBOARD_PUBLIC_IDS[environment].submission,
+  slugs: [
+    "apply-for-planning-permission",
+    "apply-for-a-lawful-development-certificate",
+    "pre-application",
```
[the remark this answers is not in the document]
zz-hh-aa: Makes a lot of sense, thanks :) moved to `utils` and refactored as functions.

editor.planx.uk/src/pages/FlowEditor/lib/analytics/utils.ts
```diff
@@ -190,3 +190,50 @@ export function getAllowListAnswers(
...
+});
+export const getSubmission = (environment: Environment) => ({
+  id: DASHBOARD_PUBLIC_IDS[environment].submission,
+  slugs: [
+    "apply-for-planning-permission",
+    "apply-for-a-lawful-development-certificate",
+    "pre-application",
+  ],
+});
+export const getDiscretionary = (environment: Environment) => ({
```
DafyddLlyr, line 231:
```suggestion
});

export const getRAB = (environment: Environment) => ({
  id: DASHBOARD_PUBLIC_IDS[environment].RAB,
  slugs: ["report-a-planning-breach"],
});

export const getSubmission = (environment: Environment) => ({
  id: DASHBOARD_PUBLIC_IDS[environment].submission,
  slugs: [
    "apply-for-planning-permission",
    "apply-for-a-lawful-development-certificate",
    "pre-application",
  ],
});

export const getDiscretionary = (environment: Environment) => ({
```

nit: A bit of whitespace here really helps with readability.

editor.planx.uk/src/routes/views/flowEditor.tsx
```diff
@@ -121,13 +130,23 @@ export const flowEditorView = async (req: NaviRequest) => {
...
+  let flowAnalyticsLink: string | null = null;
+
+  if (flowStatus === "online") {
+    flowAnalyticsLink =
+      generateAnalyticsLink({
+        environment,
+        flowId: id,
+        flowSlug: flow,
+      }) ?? null;
+  }
```
DafyddLlyr, line 148:
```suggestion
const flowAnalyticsLink = generateAnalyticsLink({
  environment,
  flowId: id,
  flowSlug: flow,
  flowStatus,
});
```

I'd suggest grouping all this logic together. Currently, we have a flow status check and fallback to `null` outside the `generateAnalyticsLink()` function.

If we needed to call this again from somewhere, we'd need to synchronise this logic.

editor.planx.uk/src/routes/views/flowEditor.tsx
```diff
@@ -144,3 +163,45 @@ export const flowEditorView = async (req: NaviRequest) => {
...
+
+export const generateAnalyticsLink = ({
+  environment,
+  flowId,
+  flowSlug,
+}: {
+  environment: Environment;
+  flowId: string;
+  flowSlug: string;
+}): string | undefined => {
```
DafyddLlyr, line 182: I guess the `?? null` above is required as the signature of this returns `undefined`.

You can update the type of `flowAnalyticsLink` in the store to be `string | undefined` which might simplify things here 👍

editor.planx.uk/src/routes/views/flowEditor.tsx
```diff
@@ -144,3 +163,45 @@ export const flowEditorView = async (req: NaviRequest) => {
...
 };
+
+const includedServices = [
+  getFOIYNPP(environment),
+  getRAB(environment),
+  getSubmission(environment),
+  getDiscretionary(environment),
+];
+
+export const generateAnalyticsLink = ({
```
DafyddLlyr, line 174: I'd move these to `analytics/utils.ts` as well - they're analytics specific and aren't related to the routing and data fetching logic this file is responsible for.

editor.planx.uk/src/routes/views/flowEditor.tsx
```diff
@@ -144,3 +163,45 @@ export const flowEditorView = async (req: NaviRequest) => {
...
+  flowSlug,
+}: {
+  environment: Environment;
+  flowId: string;
+  flowSlug: string;
+}): string | undefined => {
+  let dashboardId: string | undefined;
+
+  for (const service of includedServices) {
+    const found = service.slugs.some((slug) => flowSlug.includes(slug));
```
DafyddLlyr, line 186: Should this be an `includes` check or a `===` check?

editor.planx.uk/src/routes/views/flowEditor.tsx
```diff
@@ -144,3 +163,45 @@ export const flowEditorView = async (req: NaviRequest) => {
...
+  flowId,
+  flowSlug,
+}: {
+  environment: Environment;
+  flowId: string;
+  flowSlug: string;
+}): string | undefined => {
+  let dashboardId: string | undefined;
+
+  for (const service of includedServices) {
```
DafyddLlyr, line 185: This is a custom implementation of `Array.prototype.find()` - it's probably easier / cleaner to reach for this built-in function.

```ts
const service = includedServices.find(service => 
  service.slugs.some(slug => flowSlug.includes(slug))
);
  
if (!service) return undefined;
```

Alternatively, with a callback - 

```ts
const hasMatchingSlug = (service) => 
  service.slugs.some(slug => flowSlug.includes(slug));
  
const service = includedServices.find(hasMatchingSlug);

if (!service) return undefined;
```

DafyddLlyr (commented):
A few small comments / suggestions.

Love to see a PR like this though with more red than green - a good sign! 👍

editor.planx.uk/src/routes/views/flowEditor.tsx
zz-hh-aa: I previously intended it as an `includes` (so that `pre-application` would catch `tree-pre-application-advice` and `find-out-if` would catch `council-find-out-if`) but I suppose what you were [saying about allow lists and good patterns](https://opensystemslab.slack.com/archives/C5Q59R3HB/p1749121313223269?thread_ts=1749120611.984609&cid=C5Q59R3HB) is applicable here. Now updated allow list and check for `flowSlug === slug` in [3e3e1d4](https://github.com/theopensystemslab/planx-new/pull/4763/commits/3e3e1d4385a46d66e646daffc8a9ffcac62783f1)

DafyddLlyr (approved):

PR #4768, opened 2025-06-06
feat: add new Result component customisation

augustlindemer:
This PR adds new customisation to the Result component. It adds:

- Rich text descriptions
- No result title and description customistion
- Reset to start toggles for each individual result
- 'Result - ' prefix to Result component node texts

It also removes the list of responses if the result is 'No result', due to their inapplicability to a no result page.


editor.planx.uk/src/@planx/components/Result/Editor.tsx
```diff
@@ -30,6 +32,7 @@ const FlagEditor: React.FC<{
   existingOverrides?: FlagDisplayText;
   onChange: (newValues: any) => any;
   disabled?: boolean;
+  children?: React.ReactNode;
```
augustlindemer, line 35: Making the switches children makes for better visual association between switch and flag

editor.planx.uk/src/@planx/components/Result/Editor.tsx
```diff
@@ -84,7 +89,20 @@ const ResultComponent: React.FC<Props> = (props) => {
     validate: () => {},
   });
 
-  const allFlagsForSet = flags[formik.values.flagSet];
+  const allFlagsForSet = [
+    ...flags[formik.values.flagSet],
+    {
```
augustlindemer, line 94: Creates a flag set specific no result mock flag that allows us to treat noResult overrides separately for each flag set

editor.planx.uk/src/@planx/components/Result/Editor.tsx
```diff
@@ -121,22 +139,34 @@ const ResultComponent: React.FC<Props> = (props) => {
             </Typography>
             <Box mt={2}>
               {allFlagsForSet.map((flag) => {
+                const override = formik.values.overrides?.[flag.value] || {};
```
augustlindemer, line 142: Added const since we need this for both FlagEditor and Switch.

editor.planx.uk/src/@planx/components/Result/Public/index.tsx
```diff
@@ -83,26 +87,41 @@ export const Presentational: React.FC<PresentationalProps> = ({
   reasonsTitle = "",
   responses,
   disclaimer,
+  resetButton = false,
+  resetPreview,
 }) => {
   const visibleResponses = responses.filter((r) => !r.hidden);
   const hiddenResponses = responses.filter((r) => r.hidden);
 
+  const { trackEvent } = useAnalyticsTracking();
```
augustlindemer, line 96: Same reason as tracking on Notice - users reaching dead ends on result pages is important to know about.

editor.planx.uk/src/@planx/components/Result/Public/index.tsx
```diff
@@ -141,6 +160,17 @@ export const Presentational: React.FC<PresentationalProps> = ({
             </Box>
           </WarningContainer>
         )}
+        {resetButton && (
+          <Button
+            variant="contained"
+            size="large"
+            type="submit"
+            onClick={handleResultResetClick}
+            sx={contentFlowSpacing}
```
augustlindemer, line 169: Ensures that there’s enough margin on top between reset button and Result disclaimer box.

editor.planx.uk/src/@planx/components/Result/Public/index.tsx
```diff
@@ -159,6 +189,8 @@ const ResultComponent: React.FC<Props> = (props) => {
   const data = resultData(flagSet, props.overrides);
 
   const { flag, responses, displayText } = data[flagSet];
+  const override =
+    flag?.value && props.overrides ? props.overrides[flag.value] || {} : {};
```
augustlindemer, line 193: Struggled with override interaction between different flag switches. This solution comes straight out of GPT-4, but it works.

augustlindemer:
Test flow: https://4768.planx.pizza/testing/flags-example

editor.planx.uk/src/@planx/components/Result/Editor.tsx
```diff
@@ -84,7 +89,20 @@ const ResultComponent: React.FC<Props> = (props) => {
     validate: () => {},
   });
 
-  const allFlagsForSet = flags[formik.values.flagSet];
+  const allFlagsForSet = [
+    ...flags[formik.values.flagSet],
+    {
+      value: flags[formik.values.flagSet]?.[0]?.value?.replace(
+        /[^.]+$/,
+        "noResult",
```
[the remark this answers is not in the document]
augustlindemer: Looking at the option to hardcode no result flags as part of flags.ts - what is your thinking on how to keep these synthetic 'non flags' and make sure they aren't available as flags (to set to response options)? Are you thinking filtering them out of the flags available for FlagSelect? For how these would be 'picked up' as the lowest order flag in preview.ts I'm thinking to reference them (as defined in flags.ts) as the fallback if no flag is found.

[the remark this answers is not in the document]
augustlindemer: That should work. We'd need to do the same for the filter component I suppose, since that too shouldn't show 'No result' as a flag next to 'No flag result'. I wonder however if doing this doesn't produce some confusion around what the No result "flag" actually is, which is not actually a collected flag but a type of result? An argument for a more complex but more contained mock flag? Arguably the issue is just how the Result component expects a 'flag', but perhaps something to clean up after the move towards editor flagsets?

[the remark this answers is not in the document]
jessicamcinchak: We can always keep iterating on this & clean up later, happy to go in favor of whatever you think most pragmatic decision is for now (appreciate you're on holiday next week and would be nice to get this merged & in before then :slightly_smiling_face:)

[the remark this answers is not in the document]
augustlindemer: I unfortunately also had to rebase again because for some reason the analytics_link PR broke my local branch!

editor.planx.uk/src/@planx/components/Result/Editor.tsx
```diff
@@ -89,19 +90,9 @@ const ResultComponent: React.FC<Props> = (props) => {
...
-        /[^.]+$/,
-        "noResult",
-      ),
-      text: "No result",
-      category: formik.values.flagSet,
-      bgColor: "#EEEEEE",
-      color: "#000000",
-      description: "",
-    },
+  const allFlagsForSet: FlagWithValue[] = [
```
augustlindemer, line 93: I didn't refactor the `FlagWithValue` stuff nor PlanX core's `undefined` default for `value`, but since everything should now have a `value` (no result previously was `undefined` but should never be any longer) we can probably clean this up.

PR #4770, opened 2025-06-10
feat: in source templates, render a templates configuration section in all editor modals

jessicamcinchak:
Relies on https://github.com/theopensystemslab/planx-core/pull/754

We no longer want to rely on "Customisation" tags to denote nodes which can be edited in templated flows. 

Instead, we need more granularity: 
- If edits should be allowed to this modal at all
- If yes, then require "instructions" and specify if editing is required or optional

Follows a very similar pattern as introducing node "tags" [here](https://github.com/theopensystemslab/planx-new/pull/3762), but checks `isTemplate` status within the `ModalFooter` component directly to save changing every individual `@planx/component` editor file !

**Source template modal examples:**
Screenshot from 2025-06-10 11-34-54
Screenshot from 2025-06-10 11-35-05

 **Future scope:**
- This PR ONLY introduces the _new_ source modal functionality & fields so that editors can start populating ASAP
- I'll add graph styling for `isTemplatedNode` in a future PR
- I'll expose the instructions & required/optional as read-only content at the top of modals in templated flow nodes in a future PR
- I'll expose the instructions & required/optional in the sidebar in templated flows in a future PR
- I'll remove "customisation" tag logic in a future PR

### DIFF
--- a/api.planx.uk/package.json
+++ b/api.planx.uk/package.json
@@ -13,7 +13,7 @@
     "@airbrake/node": "^2.1.8",
     "@aws-sdk/client-s3": "^3.696.0",
     "@aws-sdk/s3-request-presigner": "^3.701.0",
-    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#c936d77",
+    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#fca70dc",
     "@types/isomorphic-fetch": "^0.0.36",
     "adm-zip": "^0.5.10",
     "axios": "^1.8.3",
--- a/api.planx.uk/pnpm-lock.yaml
+++ b/api.planx.uk/pnpm-lock.yaml
@@ -25,8 +25,8 @@ importers:
         specifier: ^3.701.0
         version: 3.812.0
       '@opensystemslab/planx-core':
-        specifier: git+https://github.com/theopensystemslab/planx-core#c936d77
-        version: https://codeload.github.com/theopensystemslab/planx-core/tar.gz/c936d77(@types/react@19.1.4)
+        specifier: git+https://github.com/theopensystemslab/planx-core#fca70dc
+        version: https://codeload.github.com/theopensystemslab/planx-core/tar.gz/fca70dc(@types/react@19.1.4)
       '@types/isomorphic-fetch':
         specifier: ^0.0.36
         version: 0.0.36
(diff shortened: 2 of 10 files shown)


editor.planx.uk/src/ui/editor/ModalFooter.tsx
```diff
@@ -20,32 +22,47 @@ export const ModalFooter = <T extends BaseNodeData>({
...
+        <TemplatedNodeConfiguration
+          formik={formik}
+          isTemplatedNode={formik.values.isTemplatedNode}
+          templatedNodeInstructions={formik.values.templatedNodeInstructions}
+          areTemplatedNodeInstructionsRequired={
+            formik.values.areTemplatedNodeInstructionsRequired
+          }
+          disabled={disabled}
+        />
+      )}
```
jessicamcinchak, line 65: This is a main change: anytime the `ModalFooter` is rendered within the editor modal of a source template flow, it'll automatically have another final section !

editor.planx.uk/src/ui/editor/TemplatedNodeConfiguration.tsx
```diff
@@ -0,0 +1,84 @@
...
+export const TemplatedNodeConfiguration = <T extends BaseNodeData>({
+  formik,
+  isTemplatedNode,
+  templatedNodeInstructions,
+  areTemplatedNodeInstructionsRequired,
+  disabled,
+}: TemplatedNodeConfigurationProps<T>) => {
+  return (
+    <ModalSection>
+      <ModalSectionContent title="Templates" Icon={EditNoteIcon}>
```
jessicamcinchak, line 30: Do we want to lean into the star icon here or happy with this `EditNoteIcon`?

jessicamcinchak: Swapped for :star: based on feedback in this thread https://opensystemslab.slack.com/archives/C07F7215W7P/p1749479047827879

ianjon3s (approved):
Works as expected on pizza and locally 👍

One issue I can see: the `Templates` section is visible in `Question` components in all states (non-template flows)

jessicamcinchak:
Thanks @ianjon3s ! Great spot on Questions - now correctly updated to be conditional. I'll merge this to `main` once CI passes and start exposing read-only instructions in _templated flows_ next - any styling easy to come back to & adjust in smaller doses from here out now (eg if we decide we want checkbox for edits required instead of second switch)

PR #4775, opened 2025-06-10
feat: display editor instructions in modals within templated flows

jessicamcinchak:
Templated flows are made up of editable and disabled nodes. Editable nodes (`isTemplatedNode === true`) will always have instructions that are required or optional. We want to display them clearly at the top of the modal.

**Key changes:**
- Adds a new `TemplatedNodesInstruction` component which conditionally renders based templated node and template flow conditions (eg should _not_ render in source templates or in disabled nodes within templated flows)
- Adds this component to _every_ `@planx/component` editor modal (lots of file changes here)

Screenshot from 2025-06-12 15-22-03

### DIFF
--- a/editor.planx.uk/src/ui/editor/TemplatedNodeInstructions.tsx
+++ b/editor.planx.uk/src/ui/editor/TemplatedNodeInstructions.tsx
@@ -21,7 +21,7 @@ export const TemplatedNodeInstructions = ({
   return (
     <ModalSection sectionBackgroundColor={theme.palette.template.main}>
       <ModalSectionContent
-        title={`Instructions ${areTemplatedNodeInstructionsRequired ? `(required)` : `(optional)`}`}
+        title={`Customise ${areTemplatedNodeInstructionsRequired ? `(required)` : `(optional)`}`}
         Icon={StarIcon}
       >
         <Typography variant="body2">{templatedNodeInstructions}</Typography>


api.planx.uk/modules/flows/updateTemplatedFlow/service.ts
```diff
@@ -45,7 +45,7 @@ export const updateTemplatedFlow = async (
         flow_id: templatedFlowId,
       },
     );
-  const edits = templatedFlowEditsResponse.edits[0].data;
+  const edits = templatedFlowEditsResponse.edits?.[0]?.data || {};
```
jessicamcinchak, line 48: Sneaking in a reconciliation edge case bug fix here!

editor.planx.uk/src/ui/editor/ModalSection.tsx
```diff
@@ -2,13 +2,33 @@ import Box from "@mui/material/Box";
...
+    "& + .modalSection": {
+      borderTop: `1px solid ${theme.palette.border.main}`,
+    },
+  }),
+);
 
-export default function ModalSection(props: PropsWithChildren) {
-  return <Root className="modalSection">{props.children}</Root>;
+export default function ModalSection(
+  props: PropsWithChildren & { sectionBackgroundColor?: string },
```
jessicamcinchak, line 20: `ModalSections` now accept an optional `sectionBackgroundColor` prop

editor.planx.uk/src/ui/editor/TemplatedNodeInstructions.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  return (
+    <ModalSection sectionBackgroundColor="#e6d6ff">
+      <ModalSectionContent
+        title={`Instructions ${areTemplatedNodeInstructionsRequired ? `(required)` : `(optional)`}`}
+        Icon={StarIcon}
+      >
+        <Typography variant="body2">{templatedNodeInstructions}</Typography>
+      </ModalSectionContent>
+    </ModalSection>
```
jessicamcinchak, line 26: I thought about using an existing `WarningContainer` here, but ultimately preferred ability to keep title and icon aligned using `ModalSectionContent` - think it's a bit tidier and more prominent.

Title of this section is another interesting one - I also thought of: 
- "Templated node instructions" - but essentially I think we want purple to communicate "templated" for us, so we can just say "Instructions"
- "Instructions*" or "Instructions" - but decided to spell out "(required)" and "(optional)" instead as we don't use the `*` elsewhere in input labels etc

editor.planx.uk/src/ui/editor/ModalSection.tsx
ianjon3s: Smart!

ianjon3s (approved):

editor.planx.uk/src/ui/editor/TemplatedNodeInstructions.tsx
jessicamcinchak: Updated - good shout!

PR #4779, opened 2025-06-11
feat: add default privacy notice content and validation

augustlindemer:
This PR adds default privacy notice content to flows. It also adds service status validation requiring the policy page to be enabled to set services online and preventing the policy page to be disabled for services that are online (both are only enforced on production). It also removes the external PlanX website link from the feedback form and instead links to `/pages/privacy` in the same way the footer links do.


jessicamcinchak (approved):
A few minor comments, no show stoppers - thanks for picking this one up :tada: 

Demo here was very helpful, linking for easy future reference :bookmark: https://opensystemslab.slack.com/archives/C01E3AC0C03/p1749669345476089?thread_ts=1749554294.700519&cid=C01E3AC0C03

PR #4782, opened 2025-06-12
chore: delete metabase setup for new team or service docs

zz-hh-aa:
After #4763 merging, Metabase setup for new team / service docs were totally obsolete. I deleted the file entirely, but do let me know if it's worth keeping it and having a quick explainer of how `analytics_link` generation is created...


jessicamcinchak (approved):
Happy to :scissors:, thanks for cleaning up !

PR #4786, opened 2025-06-13
feat: add new Result component customisations

augustlindemer:
This PR adds new customisation to the Result component. It adds:

- Rich text descriptions
- No result title and description customistion
- Reset to start toggles for each individual result
- 'Result - ' prefix to Result component node texts

It also removes the list of responses if the result is 'No result', due to their inapplicability to a no result page.

Original working PR with comments [here](https://github.com/theopensystemslab/planx-new/pull/4768).

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Node.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Node.tsx
@@ -128,7 +128,9 @@ const Node: React.FC<any> = (props) => {
       return (
         <Question
           {...allProps}
-          text={`Result - ${node?.data?.flagSet ?? "Result"}`}
+          text={
+            node?.data?.flagSet ? `Result - ${node.data.flagSet}` : `Result`
+          }
         />
       );
     case TYPES.Review:


editor.planx.uk/src/@planx/components/Result/Editor.tsx
```diff
@@ -51,9 +55,10 @@ const FlagEditor: React.FC<{
             disabled={props.disabled}
           />
         </InputLabel>
-        <InputLabel label="Description">
-          <Input
-            multiline
+        <InputLabel label="Description" htmlFor="description">
+          <RichTextInput
+            name="description"
+            id="description"
```
jessicamcinchak, line 61: :+1: appreciated a11y fix

editor.planx.uk/src/@planx/components/Result/Editor.tsx
```diff
@@ -85,7 +91,10 @@ const ResultComponent: React.FC<Props> = (props) => {
     validate: () => {},
   });
 
-  const allFlagsForSet = flags[formik.values.flagSet];
+  const allFlagsForSet: FlagWithValue[] = [
+    ...(flags[formik.values.flagSet] ?? []),
+    getNoResultFlag(formik.values.flagSet) as FlagWithValue,
+  ];
```
jessicamcinchak, line 97: Helper function feels like nice & tidy solution here :raised_hands:

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Node.tsx
```diff
@@ -128,7 +128,7 @@ const Node: React.FC<any> = (props) => {
       return (
         <Question
           {...allProps}
-          text={[node?.data?.flagSet ?? "Result"].join("")}
+          text={`Result - ${node?.data?.flagSet ?? "Result"}`}
```
jessicamcinchak, line 131: nit: this could result in `Result - Result` currently, instead I think we want something like this? 
```
text={node?.data?.flagSet ? `Result - ${node.data.flagSet}` : `Result`}
```

PR #4787, opened 2025-06-13
feat: persist `user_role` in `analytics_summary`

zz-hh-aa:
# What does this PR do?
Persists `user_role` values to subsequent analytics logs for the same `analytics_id`. 

# Why?
Now that we filter various graphs by `user_role` in analytics, it would be useful to persist values to subsequent analytics logs, to avoid the need for SQL aggregation. Performing aggregation in Metabase adds complexities to queries and lots of dashboards are loading slowly--best to avoid where possible!


jessicamcinchak (approved):
New join looks good here, thanks! Hope this helps ease metabase filtering :crossed_fingers:

PR #4785, opened 2025-06-13
feat: Set up styling for template cards in graph

ianjon3s:
## What does this PR do?

Sets up basic styling for template/custom cards in the graph ("Required" only):

image

PR #4788, opened 2025-06-16
feat: templated node graph and card display logic

jessicamcinchak:
Builds on Ian's styles in #4785 and swaps hard-coded "Required" for correct logic.

Screenshot from 2025-06-16 12-29-52


editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Checklist.tsx
```diff
@@ -24,8 +25,7 @@ type Props = {
   type: TYPES;
   [key: string]: any;
   wasVisited?: boolean;
-} & NodeTags &
-  TemplatedNodeData;
```
jessicamcinchak, line 28: `props.data` _already_ captures tags and templated node data fields - we don't need to define them _again_ at the outer level here !

PR #4789, opened 2025-06-16
chore: when creating a service from template, only select from "online" options

jessicamcinchak:
Does exactly what it says - rather than immediately show all source templates in options dropdown when creating a new service "from template", only show online source templates !


editor.planx.uk/src/pages/Team/components/AddFlow/CreateFromTemplateFormSection.tsx
```diff
@@ -19,8 +19,10 @@ export const CreateFromTemplateFormSection: React.FC = () => {
   const { values, setFieldValue } = useFormikContext<CreateFlow>();
 
   const { data } = useQuery<{ templates: TemplateOption[] }>(gql`
-    query GetTemplates {
-      templates: flows(where: { is_template: { _eq: true } }) {
+    query GetOnlineSourceTemplates {
+      templates: flows(
+        where: { is_template: { _eq: true }, status: { _eq: online } }
```
jessicamcinchak, line 24: `online` is not wrapped in double quotes here because it's a graphql-recognized enum value, not string !

PR #4799, opened 2025-06-17
fix(a11y): Skip link (bypass blocks)

DafyddLlyr:
## What does this PR do?
 - Resolves https://trello.com/c/N2PDqhOg/3308-bypass-blocks-a-p14
 - Refactors out the repeated `<main/>` components into a single reusable component
 - Switches skip link from a plain anchor and href to a React component in order to call `e.preventDefault()`

I've not actually worked out what the regression was the broke this, as it's previously passed an a11y audit last year. Possibly something related to routing or structure I'm missing here?

https://github.com/user-attachments/assets/7f30c689-c622-4681-b1fb-ef4337f9197b

## To test....
- First click of `tab` opens skip link
- Link can be used via keyboard or mouse
- Scrolls to the `main` element
- Shifts focus to the `main` element
  - The next `tab` press will take you to the first navigable element within the `main` component


ianjon3s (approved):
Working perfectly and tidy introduction of simple reusable `<Main>` component

PR #4803, opened 2025-06-18
feat: LPA masthead, content update

ianjon3s:
## What does this PR do?

- Updates homepage copy
- Introduces LPA branded masthead

### DIFF
--- a/localplanning.services/src/components/LpaMasthead.astro
+++ b/localplanning.services/src/components/LpaMasthead.astro
@@ -1,0 +1,34 @@
+---
+import Container from "@components/Container.astro";
+
+interface Props {
+  title: string;
+  lpaName: string;
+  backgroundColor?: string;
+  primaryColor?: string;
+  logo?: string;
+}
+
+const { title, lpaName, backgroundColor = "bg-bg-main", primaryColor, logo } = Astro.props;
+---
+
+<section 
+  class={`clamp-[py,5,15] ${primaryColor ? '' : backgroundColor}`}
+  style={primaryColor ? `background-color: ${primaryColor}` : undefined}
+>
+  <Container class="text-center">
+    <div class="flex flex-col items-start justify-start gap-6">
+      <img
+        src={logo}
+        alt={`${lpaName} logo`}
+        class="clamp-[w,28,40]"
+      />
+      <div>
+        <h1 class="text-heading-lg text-white m-0">
+          {title.replace('{lpaName}', lpaName)}
+        </h1>
+      </div>
+    </div>
+    
+  </Container> 
+</section>
--- a/localplanning.services/src/lib/lpa-api.ts
+++ b/localplanning.services/src/lib/lpa-api.ts
@@ -10,6 +10,10 @@ export interface Service {
 interface LPA {
   name: string;
   slug: string;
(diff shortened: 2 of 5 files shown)


localplanning.services/src/lib/lpa-api.ts
```diff
@@ -10,6 +10,10 @@ export interface Service {
 interface LPA {
   name: string;
   slug: string;
+  theme: {
+    logo: string;
+    primary_colour: string;
```
DafyddLlyr, line 15: We should use camelCase throughout and handle casting at the query level.

```suggestion
    primaryColour: string;
```

localplanning.services/src/lib/lpa-api.ts
```diff
@@ -25,6 +29,10 @@ export async function fetchAllLPAs(): Promise<LPA[]> {
             lpas: teams(order_by: { name: asc }) {
               name
               slug
+              theme {
+                logo
+                primary_colour
```
DafyddLlyr, line 34:
```suggestion
                primaryColour: primary_colour
```

localplanning.services/src/components/LpaMasthead.astro
```diff
@@ -0,0 +1,39 @@
+---
+import Container from "@components/Container.astro";
+import { Image } from "astro:assets";
+
+interface Props {
+  title: string;
+  lpaName: string;
+  backgroundColor?: string;
+  primaryColor?: string;
```
DafyddLlyr, line 9: It's safe to consider `primaryColor` as required props as we set defaults when generating a new team in PlanX - each team will always have a value.

localplanning.services/src/components/LpaMasthead.astro
```diff
@@ -0,0 +1,39 @@
+---
+import Container from "@components/Container.astro";
+import { Image } from "astro:assets";
+
+interface Props {
+  title: string;
+  lpaName: string;
+  backgroundColor?: string;
```
DafyddLlyr, line 8: `backgroundColor` is never passed in as a prop, and will always just be `bg-bg-main`. Maybe we can just simplify this on line 17?

localplanning.services/src/components/LpaMasthead.astro
```diff
@@ -0,0 +1,39 @@
...
+  logo?: string;
+}
+
+const { title, lpaName, backgroundColor = "bg-bg-main", primaryColor, logo } = Astro.props;
+---
+
+<section 
+  class={`clamp-[py,5,15] ${primaryColor ? '' : backgroundColor}`}
+  style={primaryColor ? `background-color: ${primaryColor}` : undefined}
+>
```
DafyddLlyr, line 19:
```suggestion
<section 
  class={"clamp-[py,5,15]"}
  style={`background-color: ${primaryColor}` }
>
```

DafyddLlyr (changes requested):
A few small tidy ups / improvements noted above 👍

ianjon3s:
@DafyddLlyr, you got there before I came in to reply. Thanks for all these! 

I've also renamed the `primaryColour` prop to `backgroundColour` within the LPA masthead as it better describes the usage within the component

PR #4804, opened 2025-06-18
feat(a11y): Keyboard access for `MapAndLabel` "Copy feature" select

DafyddLlyr:
## What does this PR do?
 - Fixes "Copy feature" dropdown to be keyboard accessible
 - Adds explicit "Copy" button to control action
 - Resolves https://trello.com/c/64ZlcVF7/3312-keyboard-a-aaa-p-23-last-prio-bc-mapandlabel

## To test...
Try using the `MapAndLabel` component in Storybook via the keyboard - https://storybook.4804.planx.pizza/?path=/docs/planx-components-mapandlabel--docs

https://github.com/user-attachments/assets/3511bee1-5e17-4346-842a-ced937ef9c0b

### DIFF
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/CopyFeature.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/CopyFeature.tsx
@@ -71,6 +71,7 @@ export const CopyFeature: React.FC<Props> = ({
         </InputLabel>
       </Box>
       <Button
+        data-testid="copyButton"
         type="submit"
         disableRipple
         sx={{ alignSelf: "flex-end", ml: 2, mb: 0.5 }}
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/CopyFeature.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/CopyFeature.tsx
@@ -1,6 +1,8 @@
 import Box from "@mui/material/Box";
+import Button from "@mui/material/Button";
 import MenuItem from "@mui/material/MenuItem";
 import { visuallyHidden } from "@mui/utils";
+import { useFormik } from "formik";
 import { Feature } from "geojson";
 import React from "react";
 import SelectInput from "ui/editor/SelectInput/SelectInput";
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/@planx/components/MapAndLabel/Public/CopyFeature.tsx
```diff
@@ -27,39 +29,56 @@ export const CopyFeature: React.FC<Props> = ({
...
   return (
-    <Box>
-      <InputLabel label="Copy from" id={`select-${destinationIndex}`}>
-        <SelectInput
-          disabled={isDisabled}
-          aria-describedby="copy-feature-description"
-          bordered
-          required
-          title={"Copy from"}
-          value={""}
```
DafyddLlyr, line 39: This line was the cause of the keyboard access problem. As there's no value, the MUI select doesn't automatically grant focus to the child correctly. Once I added a value here, keyboard access worked as expected.

I expanded this to use Formik as it soon became apparent that we'd need to handle change/blur/updating value etc and this actually saves a fair amount of boilerplate.

editor.planx.uk/src/@planx/components/MapAndLabel/Public/CopyFeature.tsx
```diff
@@ -27,39 +29,56 @@ export const CopyFeature: React.FC<Props> = ({
...
+      </Box>
+      <Button
+        data-testid="copyButton"
+        type="submit"
+        disableRipple
+        sx={{ alignSelf: "flex-end", ml: 2, mb: 0.5 }}
+        disabled={isDisabled}
+      >
+        Copy
+      </Button>
```
DafyddLlyr, line 81: I'm not a huge fan of this button, in terms of layout and complexity added to an already dense number of interactions. The report states the following - 

> _It is also noted that the form updates automatically as an option is selected. If a user
accidentally makes a selection, this will overwrite the details they have already entered. It is
suggested that users should be able to save details for a tree as a template before being
able to copy those details._

And later - 

> _Navigating through the options of a select component in this way changes the value of the control as a
user browses each option. Therefore, if the form submits as the value is changed, users would inadvertently overwrite the details in the form simply by browsing through the options._

Very open to alternate ideas here, though I do agree it's worth making this an explicit action which triggers on a "submit" action as opposed to a "change". This seemed better than an alert/modal("Are you sure you want to overwrite XXX?").

editor.planx.uk/src/@planx/components/MapAndLabel/test/public.test.tsx
```diff
@@ -386,8 +386,10 @@ describe("copy feature select", () => {
     await user.click(copyInput);
 
     const listItemTwo = getByRole("option", { name: "Tree 2" });
+    const copyButton = getByTestId("copyButton");
```
DafyddLlyr, line 389: Spent a while trying to grab this via `getByRole("button", { name: "Copy" });` but RTL wasn't finding it.

PR #4805, opened 2025-06-18
feat: when publishing a templated flow, validate that every required templated node has been edited

jessicamcinchak:


DafyddLlyr:
> _On failure, can we further pinpoint which required templated nodes have not been updated yet and link you to them - eg may be in nested templated flow_

A few approaches here! If we return the failing nodeId we can try to link to it (even though we're scanning a published flow I think in reality most of these nodes will be within the current flow). We could also just have a link which takes you out of the publish modal and directly to the "Customise" panel where the offending nodes would be already highlighted?

PR #4810, opened 2025-06-19
feat: "Visibility" flow settings should show a static message for source templates and templated flows

jessicamcinchak:
Follows on from #4774 

If a flow is a source template or templated flow, replace the usual "Visibility" settings toggle with a static message: 
Screenshot from 2025-06-19 10-06-48


DafyddLlyr (approved):
Nice! We should also restrict this via Hasura permissions - I think we can do "an update" check to make sure it's not a template where `can_create_from_copy` is true.

PR #4812, opened 2025-06-19
feat: Handle rich text errors at the form validation schema level

DafyddLlyr:
## What does this PR do?
Hoists validation for rich text fields from a visual-only (ie bypass-able) implementation on the field component, up to the validation schema level (handled by yup and formik). This ensures that the a11y rules we wish to enforce within `RichTextInput`s now must be adhered to within the Editor modal in order to submit and save.

This contributes to solving the following issue within the DAC a11y report, as going forwards it would not be possible to save an image to PlanX without valid `alt` text - https://trello.com/c/REGig63B/3309-non-text-content-a-p16

> [!IMPORTANT]
> This PR strips the visual error check from all `RichTextInput` components, but does not yet update all forms to implement the `richText()` validation schema. Until this is completed, Editors will be able to add non-compliant rich text without visual warnings.
> 
> I'll open another PR shortly to implement this.

## Testing notes
I've implemented the `richText()` validation schema on the "About this service" page as a proof of concept. If you navigate here on the Pizza you should be able to trigger the warnings, now tied the form submission, which will block you from saving invalid content (see below).

https://github.com/user-attachments/assets/1d6bb0cd-2294-4559-ab3e-f473ca46a491

## Next steps...
 - Replace `string()` with `richText()` where fields use the `RichTextInput` component


editor.planx.uk/src/lib/yupExtensions.test.ts
```diff
@@ -0,0 +1,167 @@
...
+        richText().validate(
+          '<p><a href="https://www.planx.uk">Try clicking here!</a></p>',
+        ),
+      ).rejects.toThrow(
+        "Links must be set over text that accurately describes what the link is for. Avoid generic language such as 'click here'.",
+      );
+    });
+  });
+
+  describe("content hierarchy (headings)", () => {
```
DafyddLlyr, line 64: @ianjon3s Would appreciate an eye on these tests - do they correctly cover the behaviour we expect to see?

One issue I've spotted here is that for `rootLevelContent`, we can still have `h1+h3+h2` which is incorrect but not covered by the current implementation.

editor.planx.uk/src/ui/editor/RichTextInput/validationHelpers.ts
```diff
@@ -76,57 +77,97 @@ export const getLegislationLinkError = (
...
   const errors: string[] = [];
-  const topLevelNodes = doc.content || [];
+  if (!content) return;
 
   switch (variant) {
     case "rootLevelContent":
-      validateRootLevelContent(topLevelNodes, errors);
-      break;
-    case "nestedContent":
-    case "paragraphContent":
```
DafyddLlyr, line 119: @ianjon3s I _think_ this is a bug which I've fixed - the intention here is that `"nestedContent"` and `"paragraphContent"` are covered by the `validateDefault()` check right?

Currently, they hit the `break` and are never checked.

jessicamcinchak (commented):
Nice simplifications here, appreciate the extra test coverage! 

Unless time sensitive, I'll leave the final :white_check_mark: to Ian since a few open questions here still !

editor.planx.uk/src/ui/editor/RichTextInput/validationHelpers.ts
ianjon3s: Indeed, the intention is that both `nestedContent` & `paragraphContent` don't get checked for heading hierarchy errors.

Here's the breakdown of current usage:

---

### `rootLevelContent`
used on a "blank" content page, heading options `h1` + `h2`
example usage: content component
- `h1` must be the first element
-  only one `h1` may be used
- `h2` cannot be used before `h1` (could remove this as the violation is covered by rule one, `h1` first is the only possibility here)

---

### `default`
used in instances where a `h1` is already present, heading options `h2` + `h3`
example usage: component descriptions
- `h3` cannot be used before `h2`

---

### `nestedContent`
used in instances where a `h1` and `h2` are already present, heading options `h2` + `h3`
example usage: help text
- no heading validation rules, a `h2` or `h3` can be used first and in any order

---

### `paragraphContent`
used in examples where we don't want headings to be used, but do want to allow bulleted lists
example usage: section descriptions
- no heading validation rules, no headings can be authored

DafyddLlyr: Thanks so much @ianjon3s - super helpful! 

I'll make this change, and add a few tests and comments to make sure this is captured in the codebase.

ianjon3s (approved):
All validation rules implemented as expected 👍

PR #4814, opened 2025-06-19
feat: when creating a templated flow, copy all "readme" fields of the source template

jessicamcinchak:

Quick one! 

**Testing:**
- Create a source template
- Add a service summary, description and limitations & turn it online 
- Create a templated flow from that source
- Check that the service readme info has come through too

Screenshot from 2025-06-19 13-25-46


DafyddLlyr (approved):
Should we also copy limitations and description?

PR #4818, opened 2025-06-19
fix: error text message improvements

nichgalzin:
This pull request includes an update to the `ErrorWrapper` component in `ErrorWrapper.tsx` to enhance accessibility and improve error handling.

### Updates to `ErrorWrapper` component:

* Modified the `role` attribute to be be on the ```ErrorText``` component instead of the wrapper container. This is to prevent the role prop being passed to the input container and causing alert behavior from screen readers when a user type in the input field.  
* Added an `aria-label` attribute that dynamically provides an accessible label when there is an error, improving screen reader support.
* Updated the `ErrorText` content to prepend "Error: " to the error message when an error exists, and display an empty string otherwise, providing clearer context for users.


nichgalzin:
I realise there are failing test and am working to resolve.

DafyddLlyr (commented):
@nichgalzin A quick look at the error logs here and I think the following issue is being hit in a few places - 

RTL's `getByText()` matches strings exactly (it has `exact = true` as a default I believe). With the `ErrorWrapper` modified with the "Error: " prefix it's likely a bunch of these locators are failing. 

If this is the issue, a pragmatic fix would be to convert the locators to use regular expressions which are a bit more flexible in instances like this. Alternatively you can set `{ exact: false }` to match on substrings. I'd guess there's a similar issue happening with Playwright.

```ts
const myError = getByText(/my error message/);
```

We already use this pattern often, e.g. - 

https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/src/@planx/components/List/Public/tests/index.listBuilding.test.tsx#L265-L294

editor.planx.uk/src/ui/shared/ErrorWrapper.tsx
```diff
@@ -52,9 +52,13 @@ export default function ErrorWrapper({
...
-    <Root error={error} role={role} data-testid="error-wrapper">
-      <ErrorText id={inputId} data-testid={inputId} variant="body1">
-        {error}
+    <Root
+      error={error}
+      data-testid="error-wrapper"
+      aria-label={error ? "error message" : undefined}
+    >
+      <ErrorText id={inputId} data-testid={inputId} variant="body1" role={role}>
+        {error ? `Error: ${error}` : ""}
```
jessicamcinchak, line 61: :+1:

jessicamcinchak (approved):
Works exactly as expected ! Very clear & easy to follow even though touches a lot of files :star2:

editor.planx.uk/src/ui/shared/ErrorWrapper.tsx
nichgalzin: Thanks Jess!

PR #4821, opened 2025-06-20
feat: disable reordering, adding, and deleting options in templated node Editor modals

jessicamcinchak:
When editing templated node "options", we want to allow customising the existing inputs but disable reordering, adding, and deleting options. Since it's a template, the shape of the graph should remain locked to everynoe who's not a platform admin.

Applies to:
- Question
- Checklist (when "expandable", also includes disabling ability to add or remove groups)
- NextSteps
- TaskList
- FileUploadAndLabel

Does not apply to `ListManager` implementations in:
- Pay metadata

### DIFF
--- a/editor.planx.uk/src/@planx/components/Question/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/Question/Editor.tsx
@@ -210,6 +210,7 @@ export const Question: React.FC<Props> = (props) => {
                 currentOptionVals,
               ),
             }}
+            isTemplatedNode={props.node.data?.isTemplatedNode}
           />
         </ModalSectionContent>
       </ModalSection>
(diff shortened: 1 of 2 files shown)


editor.planx.uk/src/ui/editor/ListManager/ListManager.stories.tsx
```diff
@@ -60,4 +60,24 @@ export const Basic = () => {
...
+      Editor={Editor}
+      newValue={() => ({
+        name: "",
+        color: "#fff",
+      })}
+      isTemplatedNode={true}
+    />
+  );
+};
+
```
jessicamcinchak, line 82: In the future, it'll be much more ideal to introduce more editor tests here per component with user roles mocked - but Storybook is a nice interim / MVP option for demonstrating how we expect the `ListManager` to render within a fully-editable versus templated node:
- Re-order handle hidden
- Bin deletion icon hidden
- Insert item between existing items `+` button hidden
- `add new` button hidden
Screenshot from 2025-06-24 13-31-29

DafyddLlyr (approved):

DafyddLlyr:
Approved! ✅ 

Agree that these comments/refactors are very much a fix-forward issue that don't need to hold this up.

PR #4826, opened 2025-06-20
chore: Refactor `MoreInformation` to have access to entire Formik instance

DafyddLlyr:
Required for #4812, I've split this up as it's an independent change and that PR is growing a bit too much.

Currently, `MoreInformation` takes a number of props, but in reality we can just pass in the entire Formik instance. The alternative would be the also need to pass in additional error information (per field) in order to display these in the `RichTextInput`'s associated `ErrorWrapper`.

Drilling the Formik instance down through props like this is an anti-pattern I'd ideally rather avoid. One solution could be to use the `useFormikContext()` hook consistently. I've opted not to go for this as - 
 - it would require a much wider refactor, touching pretty much all Editor modal components
 - if we're looking to switch from Formik, it's ok to have an imperfect pattern in place in the short/medium term

### DIFF
--- a/editor.planx.uk/src/@planx/components/Notice/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/Notice/Editor.tsx
@@ -30,7 +30,7 @@ export interface NoticeEditorProps {
 }
 
 const NoticeEditor: React.FC<NoticeEditorProps> = ({ formik, disabled }) => {
-  const { values, handleChange } = formik;
+  const { values, handleChange, setFieldValue } = formik;
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/ui/editor/MoreInformation/MoreInformation.tsx
```diff
@@ -25,8 +25,9 @@ export const MoreInformation = ({
               multiline
               name="info"
               id="info"
-              value={info}
-              onChange={changeField}
+              value={formik.values.info}
+              // errorMessage={getIn(formik.errors, "info")}
```
DafyddLlyr, line 29: Comments intentionally left in this file to illustrate the aim of this PR once combined with #4812 

Once the validation is the responsibility of the parent form, we need to pass these errors down into the `RichTextInput` as opposed to bubbling them up from the `RichTextInput` to it's parent `ErrorWrapper`.

editor.planx.uk/src/@planx/components/Notice/Editor.tsx
```diff
@@ -46,116 +47,96 @@ const NoticeEditor: React.FC<NoticeEditorProps> = (props) => {
...
   return (
     <form onSubmit={formik.handleSubmit} id="modal">
-      <NoticeEditor
-        value={formik.values.notice}
-        onChange={(notice) => {
-          formik.setFieldValue("notice", notice);
-        }}
-        disabled={props.disabled}
-      />
+      <NoticeEditor formik={formik} disabled={props.disabled} />
```
DafyddLlyr, line 139: Again, we're passing the full `formik` instance along here so it's available to `MoreInformation`.

PR #4831, opened 2025-06-23
feat: Apply validation schema within Editor modal (1/3)

DafyddLlyr:
## What does this PR do?
 - Builds on https://github.com/theopensystemslab/planx-new/pull/4812
 - Applies validation for `RichTextInputs`, at the formik / Editor modal level, for the following components 
   - `AddressInput`
   - `Calculate`
   - `Checklist`
   - `Confirmation`
   - `ContactInput`
   - `Content`
   - `DateInput`
   - `DrawBoundary`
   - `ExternalPortal`
   - `Feedback`
   - `FileUpload`
- Updates docs, e2e tests, and non-a11y compliant default content
- Refactors the `richText()` validator to remove manual handling of the `required` field - we can just chain `.required()` which is much simpler (see updated tests).

Regression tests passing here - https://github.com/theopensystemslab/planx-new/actions/runs/15877428905 ✅


editor.planx.uk/src/@planx/components/AddressInput/model.ts
```diff
@@ -31,3 +36,12 @@ export const parseAddressInput = (
 
 export const formatAsSingleLineAddress = (address: Address) =>
   Object.values(address).filter(Boolean).join(", ");
+
+export const validationSchema: SchemaOf<AddressInput> =
+  baseNodeDataValidationSchema.concat(
+    object({
+      fn: string().nullable().required(),
```
jessicamcinchak, line 43: nit: because `fn` has always been required (just browser enforced before) for this component, can we use `string().required()` here without `nullable()` ?

editor.planx.uk/src/@planx/components/shared/index.ts
```diff
@@ -32,6 +35,33 @@ export const parseBaseNodeData = (
...
+      string()
+        .oneOf([...NODE_TAGS], "Invalid tag")
+        .required("Tag cannot be empty"),
+    )
+    .optional()
+    .default([]),
+
+  // TemplatedNodeData
+  isTemplatedNode: boolean().optional(),
+  templatedNodeInstructions: string().optional(),
```
jessicamcinchak, line 54: Something I didn't quite figure out when introducing these is how to have a tighter type - as props each technically has to be introduced optional for existing flow content to work, but ideally we want `templatedNodeInstructions` & `areTemplatedNodeInstructionsRequired` to be required anytime `isTemplatedNode` is `true`. Would be nice to pick up in future :bookmark:

jessicamcinchak: Also - I have to admit - I don't totally understand when it's necessary to use `.optional()` versus when optional is simply implied by omission of `.required()` !

jessicamcinchak (approved):
Appreciate this one - much simpler, more consistent direction of travel for maintaining and adding new editor components :rocket:

editor.planx.uk/src/@planx/components/shared/index.ts
DafyddLlyr: > _Also - I have to admit - I don't totally understand when it's necessary to use `.optional()` versus when optional is simply implied by omission of `.required()` !_

This is just me being inconsistent sorry! In Yup everything is `.optional()` by default and adding it functionally changes nothing. You have to explicitly set `.required()` when needed.

PR #4797, opened 2025-06-17
feat(lps): Basic layout for 'Your applications'

DafyddLlyr:
## What does this PR do?
 - Adds the "Your applications" page
 - Makes basic form which submits to hit PlanX API (response not possible currently, see below)
 
image

## Why doesn't the existing API endpoint work?
Currently the PlanX `/resume-application` endpoint requires a `teamSlug`. This is used for - 
 - Email personalisation
 - Filtering list of applications to include in the "resume" email content

For LPS, we don't want this restriction (single team) in place. The email template should send a single magic link which takes the user back to LPS, where the full list of applications is displayed.

### Can we just modify the existing endpoint?
Yes, we could do this. However, I think new endpoints (grouped within an `/lps` route/module) are a cleaner and safer option in terms of maintainability going forward. Whilst the "resume" options within PlanX and LPS are fundamentally similar we don't necessarily want them closely coupled.

## Next steps...
- [ ] Create `/lps` module in API
  - [ ] POST `lps/resume` endpoint
  - [ ] GET `lps/applications?token=abc123&email=test@example.com`
- [ ] Create `magic_links` table (containing expiry dates and unique UUIDs)
- [ ] Write to LPS to hit new endpoints
- [ ] Display applications (cards / data grid / list.... TBD!)

## User journey
Once finalised, I'll make sure an updated and accurate version of this remains within the repo in a readme, and/or on Notion

```mermaid
sequenceDiagram
    participant User
    participant LPS
    participant PlanX_API as PlanX API
    participant PlanX_Frontend as PlanX Frontend

    User->>LPS: Enter email
    LPS->>PlanX_API: POST `lps/resume`
    PlanX_API->>PlanX_API: Trigger email via GovNotify
    PlanX_API->>User: Send email with link
    
    User->>LPS: Click magic link
    LPS->>PlanX_API: GET `lps/applications?token=abc123&email=test@example.com`
    PlanX_API->>PlanX_API: Validate details
    
    alt Error
        PlanX_API->>LPS: Reject

    else Success
        PlanX_API->>LPS: Return list of applications
    end

    LPS->>PlanX_Frontend: Resume single application
    PlanX_Frontend->>PlanX_API: /validate-session API call
    PlanX_API->>PlanX_Frontend: Success
 ```

PR #4833, opened 2025-06-24
feat: Create table for `lps_magic_links`

DafyddLlyr:
## What does this PR do?
 - Adds basic structure for a `magic_links` table
 - Basic introspection tests

It's likely / possible that some of the design here will change as we reach implementation and I'll be sure to come back with some proper test coverage (API, or maybe e2e/api) once the pieces are in place.

Supports the plan outlined here - https://github.com/theopensystemslab/planx-new/pull/4797

```mermaid
sequenceDiagram
    participant User
    participant LPS
    participant PlanX_API as PlanX API
    participant PlanX_Frontend as PlanX Frontend

    User->>LPS: Enter email
    LPS->>PlanX_API: POST `lps/login`
    PlanX_API->>PlanX_API: Trigger email via GovNotify
    PlanX_API->>User: Send email with link
    
    User->>LPS: Click magic link
    LPS->>PlanX_API: GET `lps/applications?token=abc123&email=test@example.com`
    PlanX_API->>PlanX_API: Validate details
    
    alt Error
        PlanX_API->>LPS: Reject

    else Success
        PlanX_API->>LPS: Return list of applications
    end

    LPS->>PlanX_Frontend: Resume single application
    PlanX_Frontend->>PlanX_API: /validate-session API call
    PlanX_API->>PlanX_Frontend: Success
 ```

### DIFF
--- a/hasura.planx.uk/migrations/default/1750770623762_create_table_public_magic_link_tokens/up.sql
+++ b/hasura.planx.uk/migrations/default/1750770623762_create_table_public_magic_link_tokens/up.sql
@@ -1,0 +1,8 @@
+CREATE TABLE "public"."lps_magic_links" (
+  "token" uuid NOT NULL DEFAULT gen_random_uuid(),
+  "email" text NOT NULL,
+  "created_at" timestamptz NOT NULL DEFAULT now(),
+  "used_at" timestamptz,
+  "operation" text NOT NULL,
+  PRIMARY KEY ("token")
+);
--- a/hasura.planx.uk/metadata/databases/default/tables/public_lps_magic_links.yaml
+++ b/hasura.planx.uk/metadata/databases/default/tables/public_lps_magic_links.yaml
@@ -1,0 +1,37 @@
+table:
+  name: lps_magic_links
+  schema: public
+insert_permissions:
+  - role: api
+    permission:
+      check: {}
+      columns:
+        - email
+        - operation
+        - created_at
+        - used_at
+        - token
+    comment: ""
+select_permissions:
+  - role: api
+    permission:
+      columns:
+        - email
+        - operation
+        - created_at
+        - used_at
+        - token
+      filter: {}
+    comment: ""
+update_permissions:
+  - role: api
+    permission:
+      columns:
+        - email
+        - operation
+        - created_at
+        - used_at
+        - token
+      filter: {}
+      check: {}
+    comment: ""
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: hasura.planx.uk/metadata/databases/default/tables/public_magic_links.yaml [hunk 1]
```diff
@@ -0,0 +1,37 @@
+table:
+  name: magic_links
```
jessicamcinchak, line 2: nit: should this specifically be `lps_magic_links` ? Not to be confused with pre-existing resume or payment magic links which because not token-based don't need to be stored in same way?

hasura.planx.uk/migrations/default/1750770623762_create_table_public_magic_link_tokens/up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE "public"."magic_links" (
+  "token" uuid NOT NULL DEFAULT gen_random_uuid(),
+  "email" text NOT NULL,
+  "created_at" timestamptz NOT NULL DEFAULT now(),
+  "used_at" timestamptz,
+  "operation" text NOT NULL,
+  PRIMARY KEY ("token")
+);
+
+CREATE EXTENSION IF NOT EXISTS pgcrypto;
```
jessicamcinchak, line 10: nit: is this extension actually in use here? can we remove?

jessicamcinchak (approved):
A couple minor comments but looks good ! 

One note on diagram in readme: I think it's been a common source of confusion (at least for me!) to hear "resume" language to describe the initial request on LPS - I wonder if "access" or even "login" would be more clear here? 

So instead of enter email then POST to `lps/resume` it'd be POST to `lps/access` (or `lps/login`) to trigger the Notify magic link email? 

That then frees up "resume" language to continue meaning "resume a single session using a council service-specific magic link". Just a thought on framing / clarity of external language before we get too far down the rabbit hole of code naming !

hasura.planx.uk/metadata/databases/default/tables/public_magic_links.yaml
DafyddLlyr: Good shout, updated!

DafyddLlyr:
> _One note on diagram in readme: I think it's been a common source of confusion (at least for me!) to hear "resume" language to describe the initial request on LPS - I wonder if "access" or even "login" would be more clear here?_
>
> _So instead of enter email then POST to lps/resume it'd be POST to lps/access (or lps/login) to trigger the Notify magic link email?_
> 
> _That then frees up "resume" language to continue meaning "resume a single session using a council service-specific magic link". Just a thought on framing / clarity of external language before we get too far down the rabbit hole of code naming !_

Thanks for raising this, also be struggling with this one! We've framed it as "resume" as it's the same as the PlanX "resume" email, but I agree it's not clear or meaningful within this context.

"Login" feels appropriate and clear for me, and I think is the term users would consider for this action 👍 

I've updated the diagram in the PR above for posterity, and I'll apply this elsewhere also.

PR #4835, opened 2025-06-25
fix: ensure that templated `InternalPortals` can be edited

jessicamcinchak:

**Context:**
- The graph `Hanger` is hidden on read-only flows, including templated flows. This is correct behavior for a templated flow in every instance _expect_ when there's an internal portal that has been tagged `isTemplatedNode`. In this case, editors should have full flexibility to add, remove, and re-order nodes within this internal portal which requires a visible `Hanger`
- The `FormModal` for templated nodes should hide the "delete" & "make unique" buttons and only allow "updates" when a node is tagged `isTemplatedNode`. This is correct behavior for a templated flow in every instance _expect_ modals within an internal portal that has been tagged `isTemplatedNode`. In this case, editors should see a fully enabled form modal with all buttons.

**Changes:**
- Correctly displays graph `Hanger` in internal portals tagged `isTemplatedNode`
- When clicking a `Hanger` or existing node within a templated internal portal, displays an _enabled_ `FormModal` with "delete" & "make unique" buttons visible (even though child nodes will not be marked `isTemplatedNode` themselves)

Screenshot from 2025-06-26 11-51-22 Screenshot from 2025-06-26 11-51-34

**What this doesn't cover yet:**
- I've done logic-only here & the greyed out "locked" background style is now out of sync for templated internal portals - @ianjon3s could you help with this? (as a fix-forward PR is totally fine!)
- [Also please see August's latest feedback / suggestion around styling of this](https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1750877038481739?thread_ts=1750424102.836949&cid=C04DZ1NBUMR): 
  > I however noticed that template content in an editable portal that isn't additionally toggled as editable looks locked (greyed out) despite being editable (as anything in an internal portal). Could be that it inherits the background? If this isn't just a visual glitch, this makes me wonder if hybrid locked existing content alongside editable hangers is something that we could have for free (not something I think we originally planned for, but if possible by default we should take that).
- Tests ! Visibility of hangers & ability to add, reorder, delete nodes is a good case for future e2e coverage - noted on Trello, but not picked up here yet !


editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Customisations/CustomisationCard.tsx
```diff
@@ -34,6 +34,7 @@ export const CustomisationCard: React.FC<Props> = ({
         ComponentType.Checklist,
         ComponentType.ResponsiveQuestion,
         ComponentType.ResponsiveChecklist,
+        ComponentType.InternalPortal,
```
jessicamcinchak, line 37: This line addresses an adjacent bug report https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1750877038481739?thread_ts=1750424102.836949&cid=C04DZ1NBUMR: 
> There's also an issue with how internal portal Allow edit toggles are treated in the 'Customise' tab. They appear as customisation of the portal itself (i.e. the name/reference), rather than customisations within that portal.

Now, we'll mark an internal portal as "complete" if any children have been added to it (in addition to original behavior of if it's been directly edited/renamed etc)

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Hanger.tsx
```diff
@@ -30,18 +34,36 @@ const buildHref = (before: any, parent: any) => {
...
-    state.pasteNode,
-    state.isTemplatedFrom,
-  ]);
+  const [moveNode, pasteNode, isTemplatedFrom, flow, orderedFlow] = useStore(
+    (state) => [
+      state.moveNode,
+      state.pasteNode,
+      state.isTemplatedFrom,
+      state.flow,
+      state.orderedFlow,
```
DafyddLlyr, line 43: `orderedFlow` in `undefined` by default - I'm currently populating it for the "Customise" and "Search" panels with a snippet like this - 

```ts

  useEffect(() => {
    if (!orderedFlow) setOrderedFlow();
  }, [orderedFlow, setOrderedFlow]);
```

The reasons for this are - 
 - it's potentially expensive / slow to calculate
 - it's not always needed
 - when a flow changes (e.g. nodes added/removed/updated), the `orderedFlow` is not automatically updated. This means that if we generate it at "late" as possible, it's more likely to be accurate.
 
I _think_ this is working as anybody hitting this path has already landed on the "Customise" side panel (and thus generated an `orderedFlow`).

It might be best to be a bit more defensive here with the code and add a clause like - 

```ts
  const [moveNode, pasteNode, isTemplatedFrom, flow, orderedFlow, getOrderedFlow] = useStore(
    (state) => [
      state.moveNode,
      state.pasteNode,
      state.isTemplatedFrom,
      state.flow,
      state.orderedFlow,
      state.getOrderedFlow(),
]

if (!orderedFlow) getOrderedFlow();

// Use orderedFlow - I _think_ Zustand will populate it?
```

If Zustand doesn't update the variable in place (not sure if it will!) we get either refetch from the store, or get `setOrderedFlow()` to return `orderedFlow`. The reason I hadn't already just moved this logic to the store was I was finding it was getting called a lot more often if I just wrapped the return in `getOrderedFlow()` - a `useEffect()` was much more stable.

```ts
  setOrderedFlow: () => {
    // Return early if already populated
    let orderedFlow = get().orderedFlow;
    if (orderedFlow) return orderedFlow;

    // Calculate for the first time
    const flow = get().flow as FlowGraph;
    const orderedFlow = sortFlow(flow);
    set({ orderedFlow });
    return orderedFlow;
  },
```

☝️ When I tried this instead of the `useEffect()`, `sortFlow()` was still being called multiple times within the "Search" panel.

DafyddLlyr (approved):
One comment on `orderedFlow`, but super clear and straightforward as always - great stuff!

PR #4836, opened 2025-06-25
feat(lps): Set up directory and basic accordion

ianjon3s:
## What does this PR do?

- Sets up HTML structure and basic styling for LPS directory page
- Uses generic JS structure for accessible accordion (can be changed to a plugin component if we need to)

https://localplanning.4836.planx.pizza/directory/

image

### DIFF
--- a/localplanning.services/src/components/Accordion.astro
+++ b/localplanning.services/src/components/Accordion.astro
@@ -8,10 +8,10 @@ interface Props {
 const { title } = Astro.props as Props;
 ---
 
-<details class="w-full max-w-screen-lg bg-bg-light rounded overflow-hidden [&[open]__.accordion-icon]:rotate-180 [&[open]__summary]:bg-gray-200">
-  <summary class="list-none flex items-center justify-between gap-4 clamp-[p,4,6] cursor-pointer hover:bg-gray-200 text-heading-sm text-left m-0">
+<details class="w-full max-w-screen-lg bg-bg-light rounded overflow-hidden group">
+  <summary class="list-none flex items-center justify-between gap-4 clamp-[p,4,6] cursor-pointer hover:bg-gray-200 group-open:bg-gray-200 text-heading-sm text-left m-0">
     {title}
-    <Icon class="accordion-icon clamp-[w,8,10] h-auto" name="expand" />
+    <Icon class="clamp-[w,8,10] h-auto group-open:rotate-180" name="expand" />
   </summary>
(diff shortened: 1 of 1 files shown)


localplanning.services/src/components/Accordion.astro
```diff
@@ -0,0 +1,116 @@
...
+    aria-labelledby={`${idTitle}-accordion-button`}
+    class="accordion__content hidden max-h-0 overflow-hidden transition-all duration-300 ease-in-out"
+  >
+    <section class="clamp-[py,4,6] clamp-[px,4,6] flex flex-col gap-6 border-t border-gray-300">
+      <slot />
+    </section>
+  </div>
+</div>
+
+<script>
```
DafyddLlyr, line 39: I think you can get all this functionality out of the box with `details` and `summary` native HTML elements which is [well supported](https://caniuse.com/mdn-html_elements_details_name).

The hardest part is probably rotating the icon, but I think there would be a way using Tailwind to manage this - have a look at [Open/closed state docs](https://tailwindcss.com/docs/hover-focus-and-other-states#openclosed-state).

The benefit of this would be - 
 - Less JS (quicker loading times)
 - More accessible
 - Simpler!

DafyddLlyr (changes requested):
Quick question - is this possible without JS? 🪄

localplanning.services/src/components/Accordion.astro
ianjon3s: One question I have: I added the Astro Icon package and it's added 700+ lines of code, is this considered normal? (asking as I don't usually add packages)

DafyddLlyr (approved):
Great - much simpler!

localplanning.services/src/components/Accordion.astro
```diff
@@ -0,0 +1,19 @@
...
+---
+
+<details class="w-full max-w-screen-lg bg-bg-light rounded overflow-hidden [&[open]__.accordion-icon]:rotate-180 [&[open]__summary]:bg-gray-200">
+  <summary class="list-none flex items-center justify-between gap-4 clamp-[p,4,6] cursor-pointer hover:bg-gray-200 text-heading-sm text-left m-0">
+    {title}
+    <Icon class="accordion-icon clamp-[w,8,10] h-auto" name="expand" />
+  </summary>
+  <section class="clamp-[p,4,6] flex flex-col gap-6 border-t border-gray-300">
+    <slot />
+  </section>
```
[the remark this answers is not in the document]
ianjon3s: I think the transition is a nice-to-have, and we can leave for now

PR #4840, opened 2025-06-25
chore: Drop IndexDB migration logging

DafyddLlyr:


jessicamcinchak (approved):
Thanks for remembering these !

PR #4853, opened 2025-06-28
fix: ensure internal and external portal modals set all `initialValues`

jessicamcinchak:
Fixes source template bug reported here https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1750877038481739?thread_ts=1750424102.836949&cid=C04DZ1NBUMR:
> I also noticed an issue with the Allow edits toggle on portals: Internal portal nodes within a source template always open with the Allow edits toggle off within the Editor modal. They are irrespective of that correctly saved as toggled, but their instructions disappear and in order to toggle them off you'd have to double toggle them. External portal nodes within a source template do not retain their Allow edits toggle at all, and can't be toggled on or off.

Quick one - simply missed these `initialValues` in the big PR that introduced across all component types!

### DIFF
--- a/editor.planx.uk/src/@planx/components/ExternalPortal/Editor.test.tsx
+++ b/editor.planx.uk/src/@planx/components/ExternalPortal/Editor.test.tsx
@@ -39,6 +39,9 @@ test("adding an external portal", async () => {
         flowId: "b",
         tags: [],
         notes: "",
+        isTemplatedNode: false,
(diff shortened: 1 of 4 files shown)


jessicamcinchak:
Don't know what's up with Vultr here - have destroyed/re-created multiple times to no avail.

DafyddLlyr (approved):

PR #4856, opened 2025-06-30
test(e2e): Coverage for the `/lps/applications` endpoint

DafyddLlyr:
Coverage for functionality added in #4852

Regression tests running here https://github.com/theopensystemslab/planx-new/actions/runs/15980425980 🚀

### DIFF
--- a/e2e/tests/api-driven/src/lps/steps.ts
+++ b/e2e/tests/api-driven/src/lps/steps.ts
@@ -1,0 +1,86 @@
+import { strict as assert } from "node:assert";
+import { Given, When, Then, Before, After, World } from "@cucumber/cucumber";
+import {
+  cleanup,
+  getApplications,
+  getLatestMagicLink,
+  login,
+  setup,
+} from "./helpers.js";
+
+export class CustomWorld extends World {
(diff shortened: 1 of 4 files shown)


e2e/tests/api-driven/src/lps/steps.ts
```diff
@@ -0,0 +1,110 @@
...
+  async function () {
+    this.applications = await getApplications("wrongEmail@example.com", this.token);
+  }
+);
+
+When<CustomWorld>(
+  "the expiry time has passed",
+  async function () {
+    // Sleep over the 3s expiry length in test envs
+    await new Promise((resolve) => setTimeout(resolve, 3_500));
```
DafyddLlyr, line 60: The tricky bit here is that we can't easily use fake timers as we're within an E2E context and don't have a test framework (we're just using `node:assert`).

I spent a while trying to configure this for a significantly shorter interval (~100ms) but couldn't quite get it to work consistently - we need to test it both expired and unexpired.

I figure a 3.5 second sleep isn't critical within the regression test suite, but clearly this won't be a sustainable pattern if we need to repeat something similar many times.

PR #4858, opened 2025-07-01
fix(ci): Re-install git incase of original create script failure

DafyddLlyr:


jessicamcinchak (approved):
Thanks for this one !

PR #4860, opened 2025-07-01
feat: Display applications in LPS

DafyddLlyr:


localplanning.services/src/components/applications/hooks/useFetchApplications.tsx
DafyddLlyr, line 1: I've reached for Tanstack Query here as - 
 - It's a good sandbox to try it out
 - I'd initially did this manually with just `fetch()` but you very soon start writing boilerplate to implement `isLoading`, `error` and handling caches/retries etc. A query / state management library solves this for us.

PR #4883, opened 2025-07-02
fix(e2e): Fix flakey geospatial tests, bump Playwright version

DafyddLlyr:
## What does this PR do?
- Updates Playwright
- Fixes flakey geospatial tests (please see comments on code below)

Regression tests passing against branch here - https://github.com/theopensystemslab/planx-new/actions/runs/16025538764 ✅

### DIFF
--- a/e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
+++ b/e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
@@ -5,7 +5,6 @@ import {
   tearDownTestContext,
 } from "./helpers/context.js";
 import { getTeamPage } from "./helpers/getPage.js";
-import { createAuthenticatedSession } from "./helpers/globalHelpers.js";
 import { answerQuestion, clickContinue } from "./helpers/userActions.js";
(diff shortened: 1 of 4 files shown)


e2e/tests/ui-driven/src/create-flow-with-geospatial.spec.ts
```diff
@@ -61,8 +60,10 @@ test.describe("Flow creation, publish and preview", () => {
     await tearDownTestContext();
   });
 
-  test("Create a flow", async ({ browser }) => {
-    test.setTimeout(70_000);
+  test("Create, publish, and preview a flow with geospatial components", async ({
+    browser,
+  }) => {
+    test.setTimeout(140_000);
```
DafyddLlyr, line 66: This is now one large test instead of two.

Previously, the following pattern was happening -
 - Test 1 fails
 - Test 2 fails (because it depends on the flow created by test 1)
 - Test 1 re-runs, and passes, it's just flakey
 - Report shows that Test 1 passed, Test 2 failed, but without meaningful logs or errors as nothing failed per-se - the test just wasn't set up correctly.

PR #4887, opened 2025-07-03
feat: Validation and type improvements for `InternalPortal`

DafyddLlyr:
image

### DIFF
--- a/editor.planx.uk/src/routes/flow.tsx
+++ b/editor.planx.uk/src/routes/flow.tsx
@@ -72,6 +72,8 @@ const newNode = route(async (req) => {
 
   const extraProps = {} as any;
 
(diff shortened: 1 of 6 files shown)


editor.planx.uk/src/routes/flow.tsx
```diff
@@ -72,6 +72,8 @@ const newNode = route(async (req) => {
 
   const extraProps = {} as any;
 
+  // Pass in list of relevant flows for portals
+  // This makes testing and mocking simpler
```
DafyddLlyr, line 76: Added a note here as this is a pattern only used here, it's an alternative to the `Presentational<T>` we use elsewhere.

We should probably just stick to one method and tidy this up, but this is outside the scope of this PR.

PR #4888, opened 2025-07-03
feat: Validation for `List`

DafyddLlyr:
image


editor.planx.uk/src/@planx/components/List/model.ts
```diff
@@ -22,7 +23,11 @@ export const parseContent = (data: Record<string, any> | undefined): List => ({
   ...parseBaseNodeData(data),
 });
 
-export const validationSchema = object({
+export const validationSchema = baseNodeDataValidationSchema.concat(object({
```
DafyddLlyr, line 26: I've skipped assigning the type `SchemaOf<List>` here as that would require a full validation schema for all fields etc.

jessicamcinchak (approved):

PR #4890, opened 2025-07-03
feat: Validation for NextSteps

DafyddLlyr:
image


editor.planx.uk/src/@planx/components/NextSteps/model.ts
```diff
@@ -22,3 +26,17 @@ export const parseNextSteps = (
...
+  object({
+    title: string().required(),
+    description: richText(),
+    steps: array(
+      object({
+        title: string().required(),
+        description: string().required(),
+        url: string().url(),
+      })
+    ).min(1),
```
DafyddLlyr, line 40: I've not yet tackled trying to pass these errors down into the `ListManager` (which is why I've also skipped `FileUpload` and `FileUploadAndLabel`)

jessicamcinchak (approved):

PR #4901, opened 2025-07-07
feat: Convert `PostcodeSearch` to a React component

DafyddLlyr:
## What?
Converts `PostcodeSearch.astro` to `PostcodeSearch.tsx`

## Why?
I think there's going to be an increasing level of complexity here (even if it is just hitting a PlanX API and handling the response), and having a React component will make it simpler to deal with this logic. Also laying the groundwork a little for Jimmy next week - I think dealing with plain React will be simpler than managing Astro/VanillaJS here.


localplanning.services/src/layouts/Layout.astro
```diff
@@ -16,6 +16,8 @@ import Analytics from "@components/Analytics.astro";
     <meta name="generator" content={Astro.generator} />
     <title>localplanning.services</title>
     <Analytics/>
+    <script type="module" is:inline src="/src/scripts/registerWebcomponents.js"
+    ></script>
```
jessicamcinchak, line 20: :+1:

jessicamcinchak (approved):
Great + straight forward! Thanks for this one, apologies for missing last week!

PR #4905, opened 2025-07-07
feat: PropertyInformation validation

DafyddLlyr:


jessicamcinchak (approved):

editor.planx.uk/src/@planx/components/PropertyInformation/model.ts
```diff
@@ -16,3 +19,9 @@ export const parseContent = (
 });
 
 const defaultDescription = "<p>This is the information we currently have about the property.</p><p>The blue line shows the <strong>outline</strong> of the property (known as the title boundary). If this looks incorrect, go back a step and <strong>check you have selected the correct address</strong>.</p><p>We use this outline to create the site boundary where the project will take place. If your project covers a different area, you can change or redraw the site boundary on the next page.</p>";
+
+export const validationSchema = baseNodeDataValidationSchema.concat(object({
+  title: string().required(),
+  description: richText().required(),
```
jessicamcinchak, line 25: nit: I understand we have a default `description` for this component, but I'm not sure I agree it's actually `required`? Eg if I want to remove the default, I should be able to save with `title` only?

PR #4909, opened 2025-07-08
feat: Validation for Result

DafyddLlyr:
image

### DIFF
--- a/editor.planx.uk/src/@planx/components/Result/Editor.tsx
+++ b/editor.planx.uk/src/@planx/components/Result/Editor.tsx
@@ -3,10 +3,11 @@ import Typography from "@mui/material/Typography";
 import {
   ComponentType as TYPES,
   Flag,
+  FlagSet,
   flatFlags,
   getNoResultFlag,
 } from "@opensystemslab/planx-core/types";
-import { useFormik } from "formik";
+import { Form, Formik, useFormikContext } from "formik";
 import groupBy from "lodash/groupBy";
 import React from "react";
 import { ModalFooter } from "ui/editor/ModalFooter";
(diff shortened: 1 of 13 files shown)


editor.planx.uk/src/@planx/components/Result/Editor.tsx
```diff
@@ -79,106 +83,110 @@ const FlagEditor: React.FC<{
...
-                  {flagSet}
-                </option>
-              ))}
-            </select>
-          </InputRow>
+      }}
+      validationSchema={validationSchema}
+      validateOnBlur={false}
+      validateOnChange={false}
+    >
```
DafyddLlyr, line 105: Replacing the `useFormik()` hook with the `<Formik>` context component allows us to call `useFormikContext<Result>` in the child (without prop drilling the errors).

I've avoided other components that use the `ListManager` so far - I think this is the pattern I'll need to repeat there.

editor.planx.uk/src/@planx/components/Result/model.ts
```diff
@@ -34,3 +37,27 @@ export interface PresentationalProps {
...
+export const validationSchema = baseNodeDataValidationSchema.concat(object({
+  flagSet: mixed().oneOf([
+    "Planning permission", 
+    "Works to listed buildings", 
+    "Works to trees & hedges", 
+    "Demolition in a conservation area", 
+    "Planning policy", 
+    "Community infrastructure levy", 
+    "Material change of use",
+  ]),
```
jessicamcinchak, line 60: nit: is there a way to read this list from planx-core so we don't have to remember to update two places if adding a new flagset now?

editor.planx.uk/src/@planx/components/Result/Editor.tsx
```diff
@@ -79,106 +83,110 @@ const FlagEditor: React.FC<{
...
+      {(formik) => (
+        <Form name="modal" id="modal">
```
jessicamcinchak, line 107: Looks like `onSubmit` has been dropped when moving from `<form />` to `<Form />` - just confirming that's intentional?

editor.planx.uk/src/@planx/components/Result/model.ts
DafyddLlyr: Good shout - fixed in d5b45607f038eab47d3af1a11712b798552963b6

PR #4910, opened 2025-07-08
feat(lps): Universal input styles

ianjon3s:
## What does this PR do?

- Introduces custom properties (CSS variables) to control input styling across Astro, React and GOV.UK web components
- Uses the GOV.UK postcode lookup web component style as a base, adding custom properties where needed
- Works across Astro and React components with the class `input--text`

Using custom properties allows us to reach web components through the shadow DOM, the compromise is that we can't use Tailwind classes for custom properties, however it does mean we can have a universally defined set of styles for inputs.

Still to do in follow up work: standardise label styles

**Web component:**

image

image

**Astro component:**

image

image

React component:

image

image

### DIFF
--- a/localplanning.services/src/components/applications/LoginForm.tsx
+++ b/localplanning.services/src/components/applications/LoginForm.tsx
@@ -21,7 +21,6 @@ const LoginForm: React.FC = () => {
           Enter your email address to log in
         </label>
         <input
-          className="input--text"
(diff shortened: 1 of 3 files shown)


ianjon3s:
Great shout @DafyddLlyr , I've updated to target the native HTML elements

PR #4913, opened 2025-07-08
feat(lps): Content for about page

ianjon3s:


DafyddLlyr (approved):
Great! 🥇 Good result for both maintainability and expansion in future if we need a new page here 👍

PR #4915, opened 2025-07-08
feat: Rich text validation for FileUploadAndLabel

DafyddLlyr:


editor.planx.uk/src/@planx/components/FileUploadAndLabel/schema.ts
```diff
@@ -65,18 +45,18 @@ export const ruleSchema: SchemaOf<Rule> = object({
 
 export const fileTypeSchema: SchemaOf<FileType> = object({
   name: string().required(),
-  fn: string().required(),
+  fn: string().required("Required field"),
```
DafyddLlyr, line 48: Without a more generic message, the error on deeply nested fields (for filetypes) becomes `filetype[3].fn is a required field`.

PR #4916, opened 2025-07-08
chore: Drop redundant legistlation.gov.uk link check

DafyddLlyr:


jessicamcinchak (approved):
Satisfying :scissors: 

Thanks for working through all these, feels much more standardised and noticibly less lag in modals now that `validateOnBlur` & `validateOnChange` are consistently "off" :star2:

PR #4918, opened 2025-07-09
fix: Errors for grouped checklist sections

DafyddLlyr:


editor.planx.uk/src/@planx/components/Checklist/Editor/components/GroupedOptions.tsx
```diff
@@ -51,7 +52,7 @@ export const GroupedOptions = ({
             <Box display="flex" pb={1}>
               <InputRow>
                 <Input
-                  errorMessage={`formik.errors[${groupIndex}].title`}
```
DafyddLlyr, line 54: Problems with this original implementation - 
 - It's a string not a variable!
 - It's the wrong path

PR #4922, opened 2025-07-09
feat(lps): Size variant for masthead

ianjon3s:
## What does this PR do?

Adds a size variant to the masthead.

**Default size:**
https://localplanning.4922.planx.pizza/about/
image

**Large size:**
https://localplanning.4922.planx.pizza/
image

### DIFF
--- a/localplanning.services/src/components/Masthead.astro
+++ b/localplanning.services/src/components/Masthead.astro
@@ -3,17 +3,17 @@ import Container from "@components/Container.astro";
 
 interface Props {
   title: string;
-  variant?: "default" | "large";
+  size?: "default" | "large";
 }
 
 const {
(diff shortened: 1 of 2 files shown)


localplanning.services/src/components/Masthead.astro
```diff
@@ -1,11 +1,34 @@
...
+  title: string;
+  size?: "default" | "large";
+}
+
+const {
+  title,
+  size = "default",
+} = Astro.props;
+
+const baseClasses = " bg-bg-main";
```
DafyddLlyr, line 14:
```suggestion
const baseClasses = "bg-bg-main";
```

DafyddLlyr (approved):

PR #4926, opened 2025-07-11
dp/verify email page

DafyddLlyr:
- **feat: Editor**
- **feat: API**


freemvmt (approved):
I didn't inspect the code closely since you already approved in #4000, but I did test it on the 4926 pizza as per the instructions in that PR, and it works like a dream! ✨

PR #4927, opened 2025-07-11
feat(lps): Handle magic link error states

DafyddLlyr:
## What does this PR do?
 - Adds an Express middleware function (`validateMagicLinkStatus()`) to the `/lps/applictions` endpoint responsible for handling error states for magic links
 - Updates LPS frontend to handle these error responses

### Motivation
Originally, this endpoint was just returning an empty response if no matching applications were found - regardless of the cause. It was pretty apparent that this was a confusing experience - users need to know if it's an expired, invalid, or inactive magic link which they've used.

### DIFF
--- a/api.planx.uk/modules/lps/index.test.ts
+++ b/api.planx.uk/modules/lps/index.test.ts
@@ -101,58 +101,7 @@ describe("logging into LPS applications", () => {
   });
 });
 
-describe("fetching applications", () => {
-  const mockLowcalSession: Omit<Application, "id"> = {
-    createdAt: "createdAtTime",
-    addressLine: null,
-    addressTitle: null,
-    service: {
-      name: "Service Name",
-      slug: "service-slug",
-      team: {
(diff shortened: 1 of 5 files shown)


localplanning.services/src/components/applications/ApplicationsAuth.tsx
```diff
@@ -0,0 +1,14 @@
+import { ApplicationsList } from "./ApplicationsList";
+import { useSearchParams } from "./hooks/useSearchParams";
+import LoginForm from "./LoginForm";
+
+export const ApplicationsAuth: React.FC = () => {
+  const { token, email } = useSearchParams();
+  const hasUsedMagicLink = Boolean(token && email);
```
DafyddLlyr, line 7: I've moved this `hasUsedMagicLink` check up a level as handling the permutations of this combined with the various error and success states within `ApplicationsList` was getting quite complex 🍝

PR #4932, opened 2025-07-14
feat: add `teamAdmin` role

jessicamcinchak:
Relies on https://github.com/theopensystemslab/planx-core/pull/782

Foundation for managing service charges & your PlanX software subscription in the editor!

Apologies in advance this one is so many file changes, but many are very formulaic & automatically generated. 

**Here's a summary of changes:**
- **Hasura**
  - Migration to add `teamAdmin` to `public.user_roles` enum table
  - Adds `teamAdmin` permissions to _every_ table in the `public` schema. I created these baseline permissions by "copying" existing `teamEditor` role - then we'll expand in the future as we go!
  - Adds introspection test for `teamAdmin` role to _every_ table, again illustrating parity with `teamEditor` for now
- **Editor**
  - Updates all editor navigation route `accessibleBy` permissions to account for new role
  - Stubs out a basic "Subscription" page within the team menu only accessible by `teamAdmin` (easy visual proof these changes work!)
- **API & E2E**:
  - Basic `TeamRole` type updates only where necessary

**To test:**
- On this pizza's Hasura console - go to `users` and switch your platform admin status off
- Add yourself to `team_members` with role `teamAdmin` (any team)
- Log into this pizza's editor and confirm that for this team in the settings menu, you can see "Subscription"

### DIFF
--- a/api.planx.uk/package.json
+++ b/api.planx.uk/package.json
@@ -13,7 +13,7 @@
     "@airbrake/node": "^2.1.8",
     "@aws-sdk/client-s3": "^3.696.0",
     "@aws-sdk/s3-request-presigner": "^3.701.0",
-    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#a2dc2a7",
+    "@opensystemslab/planx-core": "git+https://github.com/theopensystemslab/planx-core#ad790bc",
     "@types/isomorphic-fetch": "^0.0.36",
     "adm-zip": "^0.5.10",
     "axios": "^1.8.3",
--- a/api.planx.uk/pnpm-lock.yaml
+++ b/api.planx.uk/pnpm-lock.yaml
@@ -25,8 +25,8 @@ importers:
         specifier: ^3.701.0
         version: 3.842.0
       '@opensystemslab/planx-core':
-        specifier: git+https://github.com/theopensystemslab/planx-core#a2dc2a7
-        version: https://codeload.github.com/theopensystemslab/planx-core/tar.gz/a2dc2a7(@types/react@19.1.8)
+        specifier: git+https://github.com/theopensystemslab/planx-core#ad790bc
+        version: https://codeload.github.com/theopensystemslab/planx-core/tar.gz/ad790bc(@types/react@19.1.8)
       '@types/isomorphic-fetch':
         specifier: ^0.0.36
         version: 0.0.36
@@ -1011,8 +1011,8 @@ packages:
     resolution: {integrity: sha512-oGB+UxlgWcgQkgwo8GcEGwemoTFt3FIO9ababBmaGwXIoBKZ+GTy0pP185beGg7Llih/NSHSV2XAs1lnznocSg==}
     engines: {node: '>= 8'}
 
-  '@opensystemslab/planx-core@https://codeload.github.com/theopensystemslab/planx-core/tar.gz/a2dc2a7':
-    resolution: {tarball: https://codeload.github.com/theopensystemslab/planx-core/tar.gz/a2dc2a7}
+  '@opensystemslab/planx-core@https://codeload.github.com/theopensystemslab/planx-core/tar.gz/ad790bc':
+    resolution: {tarball: https://codeload.github.com/theopensystemslab/planx-core/tar.gz/ad790bc}
     version: 1.0.0
 
   '@paralleldrive/cuid2@2.2.2':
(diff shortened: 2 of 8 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
```diff
@@ -53,10 +53,11 @@ export const MembersTable = ({
   const [initialValues, setInitialValues] = useState<TeamMember | undefined>();
 
   const roleLabels: Record<Role, string> = {
-    platformAdmin: "Admin",
+    platformAdmin: "Platform admin",
+    teamAdmin: "Team admin",
     teamEditor: "Editor",
     teamViewer: "Viewer",
-    demoUser: "Demo User",
+    demoUser: "Demo user",
```
jessicamcinchak, line 60: `Demo User` to `Demo user` is a minor unrelated change for better consistency with other Planx roles and copy edit styles !

editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
```diff
@@ -99,31 +100,37 @@ function EditorNavMenu() {
...
       Icon: GroupIcon,
       route: `/${teamSlug}/members`,
-      accessibleBy: ["platformAdmin", "teamEditor"],
+      accessibleBy: ["platformAdmin", "teamAdmin", "teamEditor"],
+    },
+    {
+      title: "Subscription",
+      Icon: CurrencyPoundIcon,
+      route: `/${teamSlug}/subscription`,
+      accessibleBy: ["platformAdmin", "teamAdmin"],
```
jessicamcinchak, line 121: The `Subscription` page simply says "Coming soon" for now and I think `accessibleBy: ["platformAdmin", "teamAdmin"]` is basically as good as feature flagged - any objections?

DafyddLlyr: Yep this is great 👍

hasura.planx.uk/metadata/databases/default/tables/public_feedback.yaml
```diff
@@ -55,6 +55,20 @@ select_permissions:
         - id
       filter: {}
     comment: ""
+  - role: teamAdmin
```
DafyddLlyr, line 58: Ideally, we wouldn't have to double up and maintain permissions like this - a user should be able to have both the `teamAdmin` and `teamEditor` role which would grant them the same set of privileges as the `teamAdmin` described here.

This would mean that `teamAdmin` would grant a small set of advanced permissions, without having to duplicate (and maintain) the same permission set as `teamEditor`.

Hasura does allow us to use [inherited roles](https://hasura.io/docs/2.0/auth/authorization/inherited-roles/), but I'm not sure how cascading permissions work here (what happens the child permission set when the parent updates?).

editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
```diff
@@ -99,31 +100,37 @@ function EditorNavMenu() {
...
     },
     {
       title: "Feedback",
       Icon: RateReviewIcon,
       route: `/${teamSlug}/feedback`,
-      accessibleBy: ["platformAdmin", "teamEditor", "demoUser"],
+      accessibleBy: ["platformAdmin", "teamAdmin", "teamEditor", "demoUser"],
```
DafyddLlyr, line 127: We might need to change the way that `getUserRoleForCurrentTeam()` works to return an array now that we have multiple team-based roles - most (all?) `teamAdmins` will already have `teamEditor` permissions.

editor.planx.uk/src/routes/subscription.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { makeTitle } from "./utils";
+
+const subscriptionRoutes = compose(
+  withData((req) => ({
+    mountpath: req.mountpath,
+  })),
+
+  mount({
+    "/": compose(
```
DafyddLlyr, line 13: We need to implement an auth guard here to stop non-teamAdmin users navigating directly to this page.

A non-teamAdmin user can access the following - 

Image

editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
DafyddLlyr: I've actually just also spotted a bug in the current implementation which also means we should return an array.

🐛 If a user is both an `analyst` and a `teamAdmin`, they cannot view non-analyst links (such as `/subscription`.

hasura.planx.uk/metadata/databases/default/tables/public_feedback.yaml
DafyddLlyr: Thanks for this, and thanks for reading up on those docs - I just know inherited roles exist but haven't actually tried implementing it or dug very deeply. I agree that based on the fact they're always additive (which we can't guarantee we'll stick to) and it must be 2+ roles (not an extension of a single role) then it's certainly not the right option here 👍 

> _What is the benefit of multiple roles per user per team?_
I think this is they key question here for us! I think there are technical benefits which we'd want to leverage (smaller scope of roles, easier to test, easier to maintain etc), but there is a product-based question on how we want to present and manage this. Presenting multiple roles within the user interface is tricky and might not fit Editors' mental model (they'll more likely expect that users have a "type" as opposed to multiple roles). 

One way to handle this would be that if a user is created as a teamAdmin, we'd just call them an admin in the interface but assign them both the `teamAdmin` and `teamEditor` roles behind the scenes. This would give them the same set of permissions granted in this PR, but without the overhead of having to keep the permission sets in sync.

PR #4940, opened 2025-07-15
chore: improve unoccupied buildings list item legibility

augustlindemer:
Improves UnoccupiedCIL list item legibility from 'Building' to more specific 'Temporary or unoccupied building' in the CIL section.


jessicamcinchak (approved):
(Vultr is flakely lately - once other jobs complete, try the "re-run failed jobs" action on CI & should work without issue on second try!)

PR #4943, opened 2025-07-15
fix: Checkbox clickable area

ianjon3s:
## What does this PR do?

- Fixes clickable area of checkboxes so that the full bordered box is clickable

Before changes:
https://editor.planx.dev/testing/checkboxes/preview

After changes:
https://4943.planx.pizza/testing/checkboxes/preview

### Also fixes

- Clickable area of checkboxes labels with descriptions (only the label text should be clickable), to match https://design-system.service.gov.uk/components/checkboxes/error/

- Alignment of labels for checkboxes with images (before vs after):

image


DafyddLlyr (approved):
Great to see it's just a CSS fix - as we've now changed the element responsible for the interaction it's likely our styles were still geared towards the previous implementation.

Nice to also spot and fix the alignment issue! 🏅

PR #4944, opened 2025-07-15
feat: add `fee_breakdown` to `payment_status` audit table

jessicamcinchak:
Follows on from #4938

Everytime someone reaches, proceeds through, or fails at GOV.UK Pay we insert an audit log (applies to both self-pay and invite to pay). These audits are retained for all-time. 

We can add a `fee_breakdown` snapshot here which will allow us to sum the `serviceCharge` overtime for applicable payments. Nothing in the fee breakdown should constitute personally identifiable data so can continue to be retained indefinitely - which is a requirment for forthcoming "Subscription"/billing page. 

Still leaving analytics ALLOW_LIST approach as-is, as I suspect we'll still want a global view of this on OSL dashboards at some point?

### DIFF
--- a/api.planx.uk/modules/pay/helpers.ts
+++ b/api.planx.uk/modules/pay/helpers.ts
@@ -1,9 +1,9 @@
 import { getFeeBreakdown } from "@opensystemslab/planx-core";
-import type { FeeBreakdown } from "@opensystemslab/planx-core/types";
+import type { FeeBreakdown, Session } from "@opensystemslab/planx-core/types";
 import { gql } from "graphql-request";
+
 import airbrake from "../../airbrake.js";
 import { $api } from "../../client/index.js";
-import { getSessionData } from "../send/email/service.js";
 
 /**
  * Gracefully handle GovPay errors
@@ -41,11 +41,16 @@ export async function logPaymentStatus({
   } else {
     try {
       // get fee breakdown for this session
-      const sessionData = await getSessionData(sessionId);
-      const passportData = sessionData?.passport?.data;
+      const passportData = await getPassportData(sessionId);
+      if (!passportData) {
+        reportError({
+          error:
+            "Could not log the payment status due to missing passport data",
+          context: { sessionId, flowId, teamSlug },
+        });
+      }
 
-      let feeBreakdown: FeeBreakdown | undefined;
-      if (passportData) feeBreakdown = getFeeBreakdown(passportData);
+      const feeBreakdown = getFeeBreakdown(passportData);
 
       // log payment status response
       await insertPaymentStatus({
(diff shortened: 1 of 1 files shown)


api.planx.uk/modules/pay/helpers.ts
```diff
@@ -36,15 +39,23 @@ export async function logPaymentStatus({
...
     });
   } else {
-    // log payment status response
     try {
+      // get fee breakdown for this session
+      const sessionData = await getSessionData(sessionId);
+      const passportData = sessionData?.passport?.data;
+
+      let feeBreakdown: FeeBreakdown | undefined;
+      if (passportData) feeBreakdown = getFeeBreakdown(passportData);
```
DafyddLlyr, line 48: nit: There's no reason we should get to this stage without `passportData` - maybe a better way to type guard this would be to throw an error?

api.planx.uk/modules/pay/helpers.ts
```diff
@@ -1,6 +1,9 @@
+import { getFeeBreakdown } from "@opensystemslab/planx-core";
+import type { FeeBreakdown } from "@opensystemslab/planx-core/types";
 import { gql } from "graphql-request";
 import airbrake from "../../airbrake.js";
 import { $api } from "../../client/index.js";
+import { getSessionData } from "../send/email/service.js";
```
DafyddLlyr, line 6: We don't really have a good pattern in the API for sharing these common queries (there's a few bundles up in `helpers.ts` but I think that's it).

It's probably worth moving this to a shared location if it's being used across modules - somebody working in `/send` could update this function without considering the impact of this on other sections of the codebase.

editor.planx.uk/src/@planx/components/SetFee/model.ts
```diff
@@ -33,7 +33,7 @@ export const validationSchema: SchemaOf<SetFee> = baseNodeDataValidationSchema.c
   object({
     applyCalculatedVAT: boolean().required(),
     fastTrackFeeAmount: number()
-      .positive()
+      .min(0)
```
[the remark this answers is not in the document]
DafyddLlyr: Thanks for fixing this!

DafyddLlyr (approved):

api.planx.uk/modules/pay/helpers.ts
jessicamcinchak: Yep good shout - updated :heavy_check_mark:

api.planx.uk/modules/pay/helpers.ts
jessicamcinchak: Fair point - instead of moving to a shared spot, actually just defined a fresh `getPassportData` query direct in this file (it's a good excuse to use graphql's `path` !)

PR #4946, opened 2025-07-15
feat(lps): Split up search pages, add nav link

DafyddLlyr:
## What does this PR do?
Replaces a single generic search page with `/search/apply`, `/search/notify` and `/search/guidance`, each with custom content


localplanning.services/src/content/actions/index.ts
```diff
@@ -1,22 +1,21 @@
+export const ACTIONS = ["apply", "notify", "guidance"] as const;
 
+export type Action = typeof ACTIONS[number];
 
-export const cards = [
-  {
+export const actions: Record<Action, { title: string, description: string }> = {
```
DafyddLlyr, line 5: I'm reading from this content block for the `OptionsCard` and the title/description on the search pages.

It might be the case we need to expand on this slightly and have different content in these contexts but it felt right for now to share this instead of duplicating it.

ianjon3s: Works great!

ianjon3s (approved):
Works perfectly, good to have proper subdirectories for search indexing too

PR #4951, opened 2025-07-16
test(e2e): Coverage for LPS magic link error statuses

DafyddLlyr:


api.planx.uk/modules/saveAndReturn/service/validateSession.ts
```diff
@@ -34,7 +34,11 @@ export async function validateSession(
   const flowId = fetchedSession.flow_id!;
 
   // if a user has paid, skip reconciliation
-  const userHasPaid = sessionData?.govUkPayment?.state?.status === "created";
+  // Docs: https://docs.payments.service.gov.uk/api_reference/#payment-status-meanings
+  const paymentStartedStatuses = ["created", "submitted", "success"];
+  const userStatus = sessionData?.govUkPayment?.state?.status;
+  const userHasPaid = userStatus && paymentStartedStatuses.includes(userStatus);
+
```
DafyddLlyr, line 41: This is a genuine bug which as been fixed - prior to including  the `"submitted"` and `"success"` status here, the user was incorrectly shown a reconciliation page when returning from GovPay.

jessicamcinchak (approved):
Thanks for getting to bottom of these !

PR #4954, opened 2025-07-17
feat: review service charges in the editor

jessicamcinchak:
See https://trello.com/c/ppRd8Uoh/3203-show-total-amount-of-service-charges-paid-directly-in-the-editor

**Changes:**
- Adds a `service_charges_summary` database view
- Builds out a team-level "Subscription" page which only `teamEditors` can access

**Todo in a follow-up PR:**
- Hook up actual CSV download to UI link ! Omitting here as this has grown very big and will be easier to pickup separately in isolation. No one is actually live with service charges yet, so no immediate user impact


DafyddLlyr (commented):
Appreciate this is a draft and styles are WIP!

A few very minor style comments - 
- We should add padding to the rows so the text isn't touching the edges of the rows (particularly noticeable against the gray backgrounds)

Image

- Only the final "total" role should be in bold, not the total for each row

Image

- We should use the existing `Caret` component for the accordions

editor.planx.uk/src/pages/FlowEditor/components/Subscription/Subscription.tsx
```diff
@@ -0,0 +1,25 @@
...
+        Subscription
+      </Typography>
+      <Typography variant="body1">
+        Details about your Plan✕ software subscription.
+      </Typography>
+    </SettingsSection>
+    <ServiceCharges serviceCharges={serviceCharges} />
+    {/* <Contract /> */}
+    <Discount serviceCharges={serviceCharges} />
+    {/* <SSLRenewal /> */}
```
jessicamcinchak, line 23: I've stubbed out `Contract` and `SSL` components, but they aren't "active" yet until we decide how to integrate data points from our Notion CRM into the editor - let's leave these for a follow-up PR !

editor.planx.uk/src/pages/FlowEditor/components/Subscription/components/Discount.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const amountToGo = DISCOUNT_THRESHOLD - rawAmountCollected;
+  const numberSubmissionsToGo = amountToGo / DEFAULT_SERVICE_CHARGE_AMOUNT;
+
+  const amountExceeded = rawAmountCollected - DISCOUNT_THRESHOLD;
+
+  return (
+    <Box sx={{ width: "100%" }}>
+      <Gauge
+        height={160}
+        value={percentOfThreshold > 100 ? 100 : percentOfThreshold}
```
jessicamcinchak, line 60: In the case of discount eligibility, we want to ensure that the bottom of the gauge is still level (rather than the right side starting to curve down & overflow) so the value should never exceed 100 ! Eg 
Screenshot from 2025-07-22 11-32-38

editor.planx.uk/src/pages/FlowEditor/components/Subscription/components/Discount.tsx
```diff
@@ -0,0 +1,88 @@
...
+    </SettingsSection>
+  );
+};
+
+const DiscountProgress = ({ serviceCharges }: SubscriptionProps) => {
+  const rawAmountCollected = sumBy(serviceCharges, "amount");
+  const percentOfThreshold = (rawAmountCollected / DISCOUNT_THRESHOLD) * 100;
+
+  const amountToGo = DISCOUNT_THRESHOLD - rawAmountCollected;
+  const numberSubmissionsToGo = amountToGo / DEFAULT_SERVICE_CHARGE_AMOUNT;
```
jessicamcinchak, line 52: Currently, all teams will use the `DEFAULT_SERVICE_CHARGE_AMOUNT` - but in the future we may need to adjust this to account for different amounts per team.

editor.planx.uk/src/pages/FlowEditor/components/Subscription/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+export const getUKFiscalYear = (fyQuarter: number, calendarYear: number) => {
+  if (fyQuarter === 4) return calendarYear - 1;
+  return calendarYear;
+};
+
+// Eg 2025 => "FY 2025/26"
+export const formatUKFiscalYear = (fiscalYear: number): string => {
+  const abbreviatedYear = Number(fiscalYear.toString().slice(-2));
+  return `FY ${fiscalYear}/${abbreviatedYear + 1}`;
+};
```
jessicamcinchak, line 35: I've tried to handle the majority of the "quarter", "fiscal year" categorisation in the PSQL view itself (as this will also ensure availability in the CSV download) but there's a still a few formatting utils we need to handle and slightly duplicate here - particularly because the "Subscription" page takes a _current visit date_ (not exclusively past payment date) and needs to get current fiscal year quarter and year from it

hasura.planx.uk/migrations/default/1753098512705_update_service_charges_view_fiscal_year/up.sql
```diff
@@ -0,0 +1,26 @@
...
+      to_char(ps.created_at, 'Month') as month_text,
+      case 
+        when extract(quarter from ps.created_at) = 1 then 4
+        else extract(quarter from ps.created_at) - 1
+      end as fiscal_year_quarter,
+      case
+        -- UK Fiscal Year starts April 1, so Jan - March of current year are "last" fiscal year
+        when extract(month from ps.created_at) <= 3 then extract(year from ps.created_at) - 1
+        else extract(year from ps.created_at)
+      end as fiscal_year
```
jessicamcinchak, line 21: TODO - double-check if "FY 2025-26" is correctly categorised as "2025" (because most months of this fiscal year start/fall within this year) or if it should actually be "2026" (because this is when the fiscal year ends)

editor.planx.uk/src/pages/FlowEditor/components/Subscription/components/Contract.tsx
```diff
@@ -0,0 +1,16 @@
...
+export const Contract = () => (
+  <SettingsSection background>
+    <Typography variant="h3" component="h4" gutterBottom>
+      Contract
+    </Typography>
+    <Typography variant="body2" color="text.secondary">
+      Terms and key dates of your software contract. OSL invoices annually for
+      the fixed subscription cost of Plan✕.
+    </Typography>
+    <Typography variant="body1">[Coming soon]</Typography>
```
jessicamcinchak, line 14: Intentionally "Coming soon" to illustrate direction of travel - but we ultimately need to revisit this one in a follow-up PR and decide how to integration Notion CRM data points like amount, start and end date !

editor.planx.uk/src/pages/FlowEditor/components/Subscription/components/ServiceCharges.tsx
```diff
@@ -0,0 +1,357 @@
...
+    </>
+  );
+};
+
+const ThisQuarterServiceChargeCard = ({
+  serviceCharges,
+}: SubscriptionProps) => {
+  // "This quarter" is relative to when you access this page
+  const thisQuarter = getUKFiscalYearQuarter(getQuarter(new Date()));
+  const thisFiscalYear = getUKFiscalYear(thisQuarter, new Date().getFullYear());
```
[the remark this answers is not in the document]
jessicamcinchak: I separately rely on `thisQuarter` (number) for filtering throughout, hence why not bundled !

DafyddLlyr (approved):
Had a look through the rebased changes, previously approved the code changes here.

First thing I want to do once the monorepo setup is in place is to finally sort out Husky and prettier so we can put these linting issues to bed...!

PR #4957, opened 2025-07-17
fix: ensure Confirmation retains initial base node data

jessicamcinchak:
See https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1752674680628479

PR #4961, opened 2025-07-17
Production deploy

jessicamcinchak:
Would be great to get #4957 on prod for source template editing, but aware we still have outstanding `payment_status` regression tests failing on staging - so let's sort those (or confirm we're confident it won't impact production data and is an issue with bad mocks) before going forward !!

### DIFF
--- a/docker-compose.pizza.yml
+++ b/docker-compose.pizza.yml
@@ -54,7 +54,7 @@ services:
     image: pierrezemb/gostatic
     volumes:
       - ./localplanning.services/dist:/srv/http
-    entrypoint: "/goStatic -fallback /index.html"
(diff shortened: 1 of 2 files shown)


DafyddLlyr (approved):
@jessicamcinchak I've now confirmed that the remaining regression tests are failing here as `document.referrer` is not defined in the Playwright context - 

https://github.com/theopensystemslab/planx-new/blob/fb015a6b4d1489356c9171dcfa5a20200e0e13db/editor.planx.uk/src/pages/Preview/ResumePage.tsx#L186-L199

Happy for this deploy to go ahead and I'll get a fix in place for this shortly 👍

PR #4966, opened 2025-07-18
style(lps): Refine accordion and card style

ianjon3s:
## What does this PR do?

- Optimises size and spacing of accordions and cards so that they can be easily navigated
- Removes auto scroll-to-accordion on LPA page load, as doing so jumps over the information about which LPA is being viewed

https://localplanning.4966.planx.pizza/buckinghamshire/
https://localplanning.4966.planx.pizza/directory/

### DIFF
--- a/localplanning.services/src/components/Button.astro
+++ b/localplanning.services/src/components/Button.astro
@@ -4,6 +4,7 @@ interface Props {
   variant?: "primary" | "secondary";
   size?: "small" | "medium" | "large";
   external?: boolean;
+  class?: string;
 }
 
 const {
@@ -11,6 +12,7 @@ const {
   variant = "primary",
   size = "medium",
   external = false,
+  class: className = "",
 } = Astro.props;
 
 const variantClasses = {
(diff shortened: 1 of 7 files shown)


localplanning.services/src/components/Button.astro
```diff
@@ -26,8 +28,7 @@ const sizeClasses = {
 
 const externalClasses = external ? "after:content-['_↗']" : "";
 
-
-const classes = `button ${variantClasses[variant]} ${sizeClasses[size]} ${externalClasses}`;
+const classes = `button ${variantClasses[variant]} ${sizeClasses[size]} ${externalClasses} ${className}`.trim();
```
DafyddLlyr, line 31: Astro also accepts a `class:list` directive ([docs](https://docs.astro.build/en/reference/directives-reference/#classlist)) which we could use here to avoid worrying about string templating.

```ts
import type { AstroBuiltinAttributes } from "astro";

const classes: AstroBuiltinAttributes["class:list"] = [
  "button",
  variantClasses[variant],
  sizeClasses[size],
  externalClasses,
  className,
];
```

DafyddLlyr (approved):

localplanning.services/src/components/Button.astro
ianjon3s: Thanks for this @DafyddLlyr , much cleaner than manually stringing!

PR #4978, opened 2025-07-23
test: Improve coverage for checklist validation (editor)

DafyddLlyr:
## What does this PR do?
Ensures that the current behaviour is covered by tests - does not update or change existing behaviour.

### Motivation
[This ticket](https://trello.com/c/BWS2DpVZ/3315-nested-checklists-a-p29) requires that we have validation in place to manage unique labels within the checklist component. Prior to doing this I want to ensure the current functionality has adequate coverage.

The workflow will go something like this -
 - add test coverage for existing functionality (this PR)
 - migrate validation checks from the `test` property, into a single validation schema (https://github.com/theopensystemslab/planx-new/pull/4980)
 - add additional validation for unique label rules (upcoming)

_Screenshot from coverage report_

image

### DIFF
--- a/editor.planx.uk/package.json
+++ b/editor.planx.uk/package.json
@@ -126,6 +126,7 @@
     "@types/testing-library__jest-dom": "^5.14.9",
     "@typescript-eslint/eslint-plugin": "^5.62.0",
     "@typescript-eslint/parser": "^5.58.0",
+    "@vitest/coverage-v8": "3.2.4",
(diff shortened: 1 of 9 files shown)


ianjon3s (approved):
I've tested through all failing test criterium and was able to create each one by reverse engineering the error (with the exception of creating multiple "exclusive" options). I'd suggest waiting for a member of the content team to review in light of informed real-world usage, but as they are thin on the ground this week I'm sure we can fix-forward with their input.

PR #4989, opened 2025-07-24
dp/unique labels tests

DafyddLlyr:
- **test: Improve coverage for checklist validation (editor)**
- **fix: Add missing index prop**
- **feat: Migrate yup tests to validation schema**
- **chore: Drop old `test` validation rules**
- **test: Scenarios without data values**
- **test: Grouped checklist coverage**

### DIFF
--- a/editor.planx.uk/src/@planx/components/Checklist/model.test.ts
+++ b/editor.planx.uk/src/@planx/components/Checklist/model.test.ts
@@ -1,4 +1,4 @@
-import { checklistInputValidationSchema } from "./model";
+import { checklistInputValidationSchema, validationSchema } from "./model";
 
 describe("Checklist - validation", () => {
   describe("optional checklist fields in schema", () => {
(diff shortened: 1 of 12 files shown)


editor.planx.uk/src/@planx/components/Checklist/model.test.ts
```diff
@@ -194,23 +194,194 @@ describe("Editor validation", () => {
...
+              ],
+            },
+          ],
+        });
+
+        expect(result).toBeTruthy();
+      });
     });
 
     describe("options with data values", () => {
```
jessicamcinchak, line 306: nit: can we add another mock the realistic valid example from the Slack thread here which actually repeats `val` with unique labels? All these mocks also have unique data values, which is easy to misread a a requirement here IMO ! 

Eg 
```
options: [
    {
      id: "yes",
      data: { text: "Yes", val: "true" },
    },
    {
      id: "no",
      data: { text: "No", val: "false" },
    },
    {
      id: "unsure",
      data: { text: "I don't know", val: "false" },
    },
  ],
```

jessicamcinchak (approved):
Thanks for this one! 

Bigger picture question: do we need to also enforce unique option labels (not data values) on `Questions`, even though only `Checklists` were flagged in audit?

editor.planx.uk/src/@planx/components/Checklist/model.test.ts
jessicamcinchak: Similarly, would like to see a mock that confirms label white-space is trimmed - eg ` text: "Option A"` and `text: "Option A "` should _not_ be unique!

DafyddLlyr:
Thanks for the detailed review - always appreciate a close look at the unit test coverage 👍 

> _Bigger picture question: do we need to also enforce unique option labels (not data values) on `Questions`, even though only `Checklists` were flagged in audit?_

Yep - this is on my radar and will be picked up in another PR (questions + the simplified checklists used for `Schema` components).

PR #4994, opened 2025-07-25
chore: Disable continue button whilst PlanningConstraints is in loading state

DafyddLlyr:
image

Ticket - https://trello.com/c/ThfLZgxH/3375-disable-continue-button-while-planning-constraints-is-loading


ianjon3s (approved):
Works well. My initial thought was that we should be consistent with the address lookup and append the loading state to sit beside the button. However this is a distinctly different state in that we are indicating that content will be loaded in / there will be content shift, which I believe the skeleton does a good job of prompting.

DafyddLlyr:
Yep I also wondered about this - especially considering the two components appear back-to-back so it's more noticeable.

I think the pragmatic thing to do here is to revisit loading states holistically across the app as recently discussed and then make both of these components consistent.

PR #4995, opened 2025-07-25
feat(lps): Static LPA map

ianjon3s:
## What does this PR do?

Adds an example static LPA map to the homepage mashead


DafyddLlyr (approved):
Looking great 👌 Looking forward to feedback on this element

PR #5000, opened 2025-07-25
chore: Disable autocomplete for common password managers

DafyddLlyr:
## What does this PR do?
1. Adds specific overrides to our `Input` element to not allow browsers (or browser extensions such as 1Password) to autocomplete fields based on their own criteria.

1Password (used internally by OSL) ignored the `autocomplete` property, and aggressively matches inputs based on the `name` property - even with substrings (e.g. `newAddressTitle` is picked up as an autocomplete field).

2. Adds correct `autcomplete` values for our Address and ContactInput components
These `autocomplete` properties now adhere to the standards of the HTML spec ([docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#variations)).

Most browsers / extensions are probably already reading this correctly based on the existing `name` property (Chrome and Firefox already autocomplete for me as expected here prior to this change).

Motivation (OSL Slack) - https://opensystemslab.slack.com/archives/C5Q59R3HB/p1753452019980339?thread_ts=1753446778.442799&cid=C5Q59R3HB

### DIFF
--- a/editor.planx.uk/src/@planx/components/TextInput/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/TextInput/Public.tsx
@@ -48,6 +48,11 @@ const TextInputComponent: React.FC<Props> = (props) => {
       <InputRow>
         <InputLabel label={props.title} hidden htmlFor={props.id}>
           <Input
+            autoComplete={((type) => {
+              if (type === "email") return "email";
+              else if (type === "phone") return "tel";
+              return "off";
+            })(props.type)}
             type={((type) => {
               if (type === "email") return "email";
               else if (type === "phone") return "tel";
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/ui/shared/Input/Input.tsx
```diff
@@ -107,6 +107,12 @@ export default forwardRef((props: Props, ref): FCReturn => {
           "aria-label": ariaLabel,
           "aria-describedby": ariaDescribedBy,
           "aria-labelledby": ariaLabelledBy,
+          // Disable autofill using common browser password managers, unless an autocomplete value is provided
+          "data-1p-ignore": !props.autoComplete,
+          "data-op-ignore": !props.autoComplete,
+          "data-lpignore": !props.autoComplete,
+          "data-form-type": props.autoComplete || "other",
+          "data-bwignore": !props.autoComplete,
```
DafyddLlyr, line 115: Sources - 
 - https://developer.1password.com/docs/web/compatible-website-design/#ignore-offers-to-save-or-fill-specific-fields
 - https://www.stefanjudis.com/snippets/turn-off-password-managers/

PR #5014, opened 2025-07-28
feat: Add `FileUploadField` to `List` and `Page` components

DafyddLlyr:
## What does this PR do?
 - Adds a `FileUploadField` to the List and Page components, which allows users to upload files
 - Handles mapping passport values to `PASSPORT_REQUESTED_FILES_KEY` and the relevant root `FileType` in the passport on submit


editor.planx.uk/src/@planx/components/List/Public/tests/index.listBuilding.test.tsx
```diff
@@ -11,6 +12,21 @@ import { fillInResponse } from "./testUtils";
 
 Element.prototype.scrollIntoView = vi.fn();
 
+const mocks = vi.hoisted(() => {
```
DafyddLlyr, line 15: I spent a while trying to move these repeated mocks of file uploads to either a helper function or just into `testUtils.ts` directly but couldn't quite get it to work. I think the issue is the order in which imports and mocks are hoisted.

This means there's a bit of repetition each place we need to use `fillInResponse()` - if anybody knows ways to avoid this and directly define a mock which will be correctly hoisted please let me know.

editor.planx.uk/src/@planx/components/shared/PrivateFileUpload/Dropzone.tsx
```diff
@@ -80,6 +85,8 @@ export const Dropzone: React.FC<Props> = ({
 }) => {
   const MAX_UPLOAD_SIZE_MB = 30;
 
+  const isWithinListCard = Boolean(useOptionalListContext());
```
DafyddLlyr, line 88: This feels like a bit of a hacky solution to work out if the component is within a List or a Page but it works! 

A good alternate solution here would be CSS container queries but these aren't supported in our version of MUI.

augustlindemer:
I had some trouble testing this. Neither page nor list allow me to select the 'Zoo' schema (due to the removal for e2e testing?), so I couldn't re-create a new node to test the file upload. The Zoo schema in the existing testing flow (https://5014.planx.pizza/testing/file-upload-list-test) does not have a FileType matching fn (`photos`). The new (but removed) Zoo schema in these commits does as far as I can see (`photographs.existing`). Unfortunately I can't test the latter.

jessicamcinchak:
Okay one final comment after pizza testing ! 

Review page looks good, but did notice one glitch in `application.json` "files" generation - we pick up _two_ values when really we just want one: 
Screenshot from 2025-08-26 11-58-29

Available to download here https://5014.planx.pizza/testing/zoo-test/submissions

DafyddLlyr:
@jessicamcinchak Thanks for revisiting this one - meant to put a message here when I added that last commit.

editor.planx.uk/src/@planx/components/List/Public/tests/index.payloadGeneration.test.tsx
```diff
@@ -24,15 +41,47 @@ describe("Payload generation", () => {
...
 
       expect(handleSubmit).toHaveBeenCalled();
-      expect(handleSubmit.mock.calls[0][0]).toMatchObject(mockZooPayload);
+      const result: typeof mockZooPayload = handleSubmit.mock.calls[0][0];
+      expect(result).toMatchObject(mockZooPayload);
+
+      // All FileUploadField values are merged (2 files for each of the 2 list items)
+      expect(result.data["photographs.existing"]).toHaveLength(4);
+
+      // Schema "FileTypes" are extracted to the root level of the passport
```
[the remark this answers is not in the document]
DafyddLlyr: Good comment - all file types are now partitioned out (see [a6af686](https://github.com/theopensystemslab/planx-new/pull/5014/commits/a6af68614cecb560bea8c00ee0172dce99be225e)).

This means each file will be moved to the root, and will not be listed within a flattened passport variable. The original (nested) data is always kept intact within the breadcrumbs allowing "back" to work.

PR #5025, opened 2025-07-30
fix: correctly set aria attribute to input rather then label

nichgalzin:
Moves aria describe by when a description is present to the checkbox input instead of the label element.


DafyddLlyr:
@nichgalzin Merging this to tidy up the PR list 👍

PR #5033, opened 2025-07-31
refactor(a11y): Sections in line with GOV.UK task list

ianjon3s:


editor.planx.uk/src/@planx/components/Section/Public.tsx
```diff
@@ -153,71 +153,112 @@ export function SectionsOverviewList({
...
-            {getTag(sectionStatuses[sectionId], sectionNode.data.title)}
-          </SectionState>
-        </SectionRow>
-      ))}
+      {Object.entries(sectionNodes).map(([sectionId, sectionNode]) => {
+        const status = sectionStatuses[sectionId];
+        const clickHandler = getClickHandler(sectionId, status);
+        const isClickable = clickHandler !== undefined;
+        const statusId = `section-${sectionId}-status`;
+        const hintId = `section-${sectionId}-hint`;
```
DafyddLlyr, line 191: Generally I'd take this as a sign that we need a new component here. 

If we need to define these variables for the logic to work, it's worth defining them as their own component instead of increasing in-line complexity.

editor.planx.uk/src/@planx/components/Section/Public.tsx
```diff
@@ -153,71 +153,112 @@ export function SectionsOverviewList({
...
+          >
+            <SectionContent>
+              <SectionDescription>
+                {isClickable ? (
+                  <SectionTitleLink variant="subtitle1" component="h4">
+                    <Link
+                      href="#"
+                      onClick={(e) => {
+                        e.preventDefault();
+                        clickHandler?.();
```
DafyddLlyr, line 220: I _think_ this is redundant. As this is within a clickable element, and we're not calling `e.stopPropogation()`, the event will bubble up and just trigger the click handler there.

editor.planx.uk/src/@planx/components/Section/Public.tsx
```diff
@@ -153,71 +153,112 @@ export function SectionsOverviewList({
...
+      status === SectionStatusEnum.NeedsUpdated ||
+      status === SectionStatusEnum.ReadyToStart ||
+      status === SectionStatusEnum.ReadyToContinue
+    ) {
+      return () => nextQuestion();
+    }
+    return undefined;
+  };
+
+  const getTag = (section: SectionStatusEnum, sectionTitle: string) => {
```
DafyddLlyr, line 170:
```suggestion
  const getTag = (section: SectionStatusEnum) => {
```

nit: Unused param.

editor.planx.uk/src/@planx/components/Section/Public.tsx
```diff
@@ -153,71 +153,112 @@ export function SectionsOverviewList({
...
+            </SectionContent>
+          </SectionRowWrapper>
+        );
+      })}
     </DescriptionList>
   );
 }
 
-const Table = styled("dl")(({ theme }) => ({
+const Table = styled("ul")(({ theme }) => ({
```
DafyddLlyr, line 259: It's probably worth renaming some of these components - `Table` is not that meaningful now it's not a `dl` (arguably it wasn't a great name to begin with!). The same applies the the `DescriptionList` component.

editor.planx.uk/src/@planx/components/Section/Public.tsx
DafyddLlyr: This probably simplifies the check above on the `SectionRowWrapper` as well.

DafyddLlyr (changes requested):
Looking great! A few minor comments, but no show stoppers 👍

editor.planx.uk/src/@planx/components/Section/Public.tsx
```diff
@@ -153,71 +272,30 @@ export function SectionsOverviewList({
...
+            key={sectionId}
+            sectionId={sectionId}
+            sectionNode={sectionNode}
+            status={status}
+            showChange={showChange}
+            changeFirstAnswerInSection={changeFirstAnswerInSection}
+            nextQuestion={nextQuestion}
+          />
+        );
+      })}
```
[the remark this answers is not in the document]
ianjon3s: Have updated, thanks for your help with this one 🏁

PR #5039, opened 2025-08-01
feat: Refactor feedback submission form to show in drawer, not footer

DafyddLlyr:
## What does this PR do?
 - Refactors `Feedback` component (in footer) to use a Drawer, not to expand in place
 - Sorts `theme.components` ([28052aa](https://github.com/theopensystemslab/planx-new/pull/5039/commits/28052aa83deb0c01d64e8ff305593dd5c38d535e))
 - Moves shared `Drawer` styling to the MUI theme instead of either repetition or a shared component.

## Motivation
This was raised in the accessibility audit, with regards to how focus is handled when this element scrolls in and out of view on open/close.

## Demo
Storybook - https://storybook.5039.planx.pizza/?path=/docs/design-system-molecules-feedback--docs

### DIFF
--- a/editor.planx.uk/src/components/Feedback/index.test.tsx
+++ b/editor.planx.uk/src/components/Feedback/index.test.tsx
@@ -26,9 +26,6 @@ vi.mock("lib/feedback", () => ({
   insertFeedbackMutation: vi.fn(),
 }));
 
-const scrollIntoViewMock = vi.fn();
-window.Element.prototype.scrollIntoView = scrollIntoViewMock;
-
 describe("Feedback component triage journey", () => {
   beforeEach(() => {
     vi.clearAllMocks();
@@ -40,15 +37,12 @@ describe("Feedback component triage journey", () => {
     expect(getByText("PUBLIC BETA")).toBeInTheDocument();
     expect(getByText("Report an issue with this page")).toBeInTheDocument();
     expect(getByText("feedback")).toBeInTheDocument();
-    expect(scrollIntoViewMock).not.toBeCalled();
   });
 
   test("Selecting 'feedback' scrolls triage into view", async () => {
     const { getByText, getByRole, user } = setup(<Feedback />);
 
     await user.click(getByText("feedback"));
-
-    expect(scrollIntoViewMock).toBeCalledTimes(1);
 
     await waitFor(() => {
       expect(getByRole("button", { name: "Issue" })).toBeInTheDocument();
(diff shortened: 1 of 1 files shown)


editor.planx.uk/src/components/Feedback/index.tsx
```diff
@@ -38,52 +39,33 @@ import {
...
         break;
       default:
+        setIsDrawerOpen(true);
         setCurrentFeedbackView(event);
         break;
     }
   }
 
   async function handleFeedbackFormSubmit(values: UserFeedback) {
+    if (!currentFeedbackView) throw Error("Cannot submit feedback in current view");
```
DafyddLlyr, line 67: This is just type narrowing, it's not possible to call this function whilst `currentFeedbackView` is `null`.

editor.planx.uk/src/components/Feedback/index.tsx
```diff
@@ -138,14 +120,7 @@ const Feedback: React.FC = () => {
   }
 
   function ReportAnIssueTopBar(): FCReturn {
```
DafyddLlyr, line 122: The order of the conditions here have been reversed but the logic is still consistent - now we only show the "back" button if they've previously been on "triage", as opposed to not showing it if they were on "banner" (now removed as a view).

editor.planx.uk/src/ui/public/FeedbackOption.tsx
```diff
@@ -24,7 +24,6 @@ const Root = styled(Button, {
   justifyContent: "flex-start",
   ...(showArrow && {
     width: "100%",
-    maxWidth: "460px",
```
DafyddLlyr, line 27: This is the only style change made here - as this is just shy of the 600px drawer (minus padding) it looked a bit odd and out of line. 

It now shows as full width - 

Image

editor.planx.uk/src/themeOverrides.d.ts
```diff
@@ -149,3 +149,10 @@ declare module "@mui/material/Typography" {
     data: true;
   }
 }
+
+declare module "@mui/material/Drawer" {
+  interface DrawerProps {
+    // a11y: Make aria-label required for all instances of Drawer
+    "aria-label": string;
+  }
+}
```
jessicamcinchak, line 158: Smart !

jessicamcinchak (approved):

ianjon3s:
Coming in with late feedback here @DafyddLlyr , looks and works superb! There's a couple of follow up tasks I've spotted:
- ESC key to close the drawer, similar to help text drawer
- Returning focus to the correct place on the page when the drawer is closed (which I believe was the original feedback)

PR #5043, opened 2025-08-01
refactor(a11y): Use checkboxes for upload and label

ianjon3s:
## What does this PR do?

- Refactors Upload + label component to use checkboxes rather than an autocomplete combobox

**Demo:**
https://5043.planx.pizza/a-new-team/upload-label/preview

### DIFF
--- a/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/SelectMultipleFileTypes.tsx
+++ b/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/SelectMultipleFileTypes.tsx
@@ -1,14 +1,12 @@
-import {
-  AutocompleteChangeReason,
-  AutocompleteProps,
-} from "@mui/material/Autocomplete";
 import Box from "@mui/material/Box";
+import FormControl from "@mui/material/FormControl";
+import FormLabel from "@mui/material/FormLabel";
+import Typography from "@mui/material/Typography";
 import { visuallyHidden } from "@mui/utils";
 import capitalize from "lodash/capitalize";
-import React, { forwardRef, PropsWithChildren, useMemo } from "react";
-import { RenderGroupHeaderBlock } from "ui/shared/Autocomplete/components/RenderGroupHeaderBlock";
-import { RenderOptionCheckbox } from "ui/shared/Autocomplete/components/RenderOptionCheckbox";
-import { SelectMultiple } from "ui/shared/SelectMultiple";
+import React, { useMemo } from "react";
+import { FONT_WEIGHT_SEMI_BOLD } from "theme";
+import ChecklistItem from "ui/shared/ChecklistItem/ChecklistItem";
 
 import { FileUploadSlot } from "../../FileUpload/model";
 import {
@@ -19,7 +17,7 @@ import {
   UserFile,
 } from "../model";
 
-interface SelectMultipleProps {
+interface ChecklistProps {
   uploadedFile: FileUploadSlot;
   fileList: FileList;
   setFileList: (value: React.SetStateAction<FileList>) => void;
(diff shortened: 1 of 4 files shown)


DafyddLlyr:
Regression tests running here - https://github.com/theopensystemslab/planx-new/actions/runs/16684132569

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/SelectMultipleFileTypes.tsx
```diff
@@ -102,73 +54,99 @@ export const SelectMultipleFileTypes = (props: SelectMultipleProps) => {
...
+      sx={(theme) => ({
+        width: "100%",
+        backgroundColor: theme.palette.background.default,
+        borderColor: theme.palette.border.main,
+        borderStyle: "solid",
+        borderWidth: "1px",
+        borderTopColor: theme.palette.border.light,
+        padding: theme.spacing(2),
+      })}
+    >
```
DafyddLlyr, line 109: This is probably complex enough to justify it being its own styled component (`Root`?).

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/SelectMultipleFileTypes.tsx
```diff
@@ -102,73 +54,99 @@ export const SelectMultipleFileTypes = (props: SelectMultipleProps) => {
...
+            }}
+          >
+            {capitalize(category)} information
+          </FormLabel>
+
+          <Box>
+            {categoryOptions.map((option) => (
+              <ChecklistItem
+                key={`${category}-${option.name}`}
+                id={`${sanitizeId(uploadedFile.id)}-${sanitizeId(category)}-${sanitizeId(option.name)}`}
```
DafyddLlyr, line 141:
```suggestion
                id={sanitizeId(`${uploadedFile.id}-${category}-${option.name}`)}
```

I think this would output the same value and is a bit more readable?

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/SelectMultipleFileTypes.tsx
```diff
@@ -102,73 +54,99 @@ export const SelectMultipleFileTypes = (props: SelectMultipleProps) => {
...
-    switch (reason) {
-      case "selectOption": {
-        const updatedTags = selectedTags.filter(
-          (tag) => !initialTags?.includes(tag),
-        );
+      if (newCheckedState) {
+        // Add the tag
         const updatedFileList = addOrAppendSlots(
-          updatedTags,
+          [option.name],
```
DafyddLlyr, line 82: This is very much a follow up / fix forward issue, but wanted to raise it here.

The `Autocomplete` bundled many tags within a single update which required this more complex event handler. However, now that we're using a checkbox per-tag, the event handling is much simpler - it's always a one-to-one relationship.

This means that we don't need to wrap into an array here, and our helper functions can be simplified to handle just a single tag at a time.

For example, `addOrAppendSlots()` (accepting `string[]`) would become `addOrAppendSlot()` (accepting just `string`). The same applies for `removeSlots()`.

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/index.test.tsx
```diff
@@ -396,26 +395,23 @@ describe("Adding tags and syncing state", () => {
...
-    expect(roofPlanOption).toBeNull();
-
-    // Open dropdown
-    await user.click(autocompleteInputs[0]);
-    roofPlanOption = getByRole("option", { name: /Roof plan/ });
-    expect(roofPlanOption).toBeVisible();
+    // Find the "Roof plan" checkbox using data-testid which includes the file ID
+    // The pattern is: ${fileId}-${category}-${optionName}
+    const roofPlanCheckbox =
+      within(fileTaggingModal).getByTestId(/.*-.*-Roof-plan$/);
```
DafyddLlyr, line 410: I think we should be able to get these by label


```suggestion
    const roofPlanCheckbox =
      within(fileTaggingModal).getByLabelText("Roof plan");
```

DafyddLlyr (approved):
Love it 👏 

Very quick turnaround, and a solid pragmatic option as discussed on Slack 👍 

A number of small comments below, all non-functional - happy for this to merge as-is and for follow up fixes to be put through later.

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/index.test.tsx
```diff
@@ -396,26 +395,23 @@ describe("Adding tags and syncing state", () => {
...
+    expect(roofPlanCheckbox).toBeInTheDocument();
```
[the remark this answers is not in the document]
ianjon3s: Thanks for this, learning more about tests by modifying and updating, this is really useful to know. I've removed the unnecessary checks (now using `getByLabelText` as suggested in another comment)

editor.planx.uk/src/@planx/components/shared/PrivateFileUpload/UploadedFileCard.tsx
```diff
@@ -105,7 +106,7 @@ export const UploadedFileCard: React.FC<Props> = ({
       }
     >
       <>
-        <FileCard>
+        <FileCard sx={{ ...sx }}>
```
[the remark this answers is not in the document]
ianjon3s: Yep, makes sense, have updated

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/SelectMultipleFileTypes.tsx
ianjon3s: Agreed, I've created a styled `<Root>` component

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/index.test.tsx
ianjon3s: Yep, label text works and is semantically cleaner

editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/SelectMultipleFileTypes.tsx
ianjon3s: I had suspected as much, I took the most direct route to re-render the tag list as quickly as possible with the expectation that there were more wins to be had in terms of simplicity. I'll review again once this is across the line.

PR #5049, opened 2025-08-01
feat: `MapAndLabel` - responsive tabs and reflow

DafyddLlyr:
## What does this PR do?
- Applies a basic responsive design to the `MapAndLabel` component
- Splits the `PreviewBrowser` off into it's own component, and forces a mobile style

https://github.com/user-attachments/assets/4278add0-f214-4b8c-a652-66822afb9243

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/PreviewBrowser/index.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/PreviewBrowser/index.tsx
@@ -1,0 +1,51 @@
+
+import Button from "@mui/material/Button";
+import { createTheme, styled, ThemeProvider, useTheme } from "@mui/material/styles";
+import { useStore } from "pages/FlowEditor/lib/store";
+import React from "react";
+import Reset from "ui/icons/Reset";
+
+import Questions from "../../../../Preview/Questions";
+
+const ResetToggle = styled(Button)(({ theme }) => ({
+  position: "absolute",
+  top: 0,
+  right: theme.spacing(3),
+  padding: theme.spacing(1, 1, 1, 0),
+  textDecorationStyle: "solid",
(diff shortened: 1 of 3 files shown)


ianjon3s (approved):
Works well on the pizza!

editor.planx.uk/src/hooks/useFormErrorFocus.ts
DafyddLlyr, line 1: Spotted a typo in the file name here.

editor.planx.uk/src/@planx/components/MapAndLabel/test/public.test.tsx
```diff
@@ -171,8 +171,6 @@ describe("validation and error handling", () => {
     await clickContinue(user);
 
     await checkErrorMessagesPopulated();
-
-    expect(tabOne).toHaveStyle("border-left: 5px solid #D4351C");
```
DafyddLlyr, line 175: This was failing - presumably as JSDom isn't handling media queries (which the style is now based on).

editor.planx.uk/src/pages/FlowEditor/components/Sidebar/PreviewBrowser/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+        lg: 9999,
+        xl: 9999,
+        formWrap: 9999,
+        contentWrap: 9999,
+      },
+    },
+  });
+
+  return (
+    <ThemeProvider theme={mobileTheme}>
```
DafyddLlyr, line 40: A more robust solution here is probably a same-origin iframe - this could also solve our issues of the preview browser not going past the currently available nodes.

As it's same-origin, we could read and write the same local storage (for breadcrumbs etc).

A few good ways to solve this one!

PR #5051, opened 2025-08-01
fix(a11y): Improvements to feedback drawer

DafyddLlyr:
Follow up to #5039, responding to [the following comment](https://github.com/theopensystemslab/planx-new/pull/5039#issuecomment-3144940766) - 

> Coming in with late feedback here @DafyddLlyr , looks and works superb! There's a couple of follow up tasks I've spotted:
>
> - ESC key to close the drawer, similar to help text drawer
> - Returning focus to the correct place on the page when the drawer is closed (which I believe was the original feedback)

https://github.com/user-attachments/assets/fdea3689-a282-4a3a-b75e-a6ccc81dfc05

### DIFF
--- a/editor.planx.uk/src/components/Feedback/FeedbackPhaseBanner.test.tsx
+++ b/editor.planx.uk/src/components/Feedback/FeedbackPhaseBanner.test.tsx
@@ -1,4 +1,4 @@
-import React, { createRef } from "react";
+import React from "react";
 import { setup } from "testUtils";
 import { vi } from "vitest";
 import { axe } from "vitest-axe";
@@ -18,8 +18,6 @@ describe("FeedbackPhaseBanner presentation and functionality", () => {
       <FeedbackPhaseBanner
         handleFeedbackClick={handleFeedbackClick}
         handleReportAnIssueClick={handleReportAnIssueClick}
-        feedbackButtonRef={createRef()}
-        reportIssueButtonRef={createRef()}
       />,
     );
 
@@ -33,8 +31,6 @@ describe("FeedbackPhaseBanner presentation and functionality", () => {
       <FeedbackPhaseBanner
         handleFeedbackClick={handleFeedbackClick}
         handleReportAnIssueClick={handleReportAnIssueClick}
-        feedbackButtonRef={createRef()}
-        reportIssueButtonRef={createRef()}
       />,
     );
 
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/components/Feedback/index.tsx
```diff
@@ -355,10 +360,11 @@ const Feedback: React.FC = () => {
...
   return (
     <Box>
-      <FeedbackPhaseBannerView />
-      <Drawer 
+      <FeedbackPhaseBannerView handleFeedbackViewClick={handleFeedbackViewClick}/>
+      <Drawer
         aria-label="Feedback triage and submission form"
-        open={isDrawerOpen} 
+        open={isDrawerOpen}
+        onClose={closeDrawer}
```
DafyddLlyr, line 367: Passing a handler here for `onClose` ensures that the modal works as expected when there's a click away or the `esc` button is hit.

editor.planx.uk/src/components/Feedback/index.tsx
```diff
@@ -37,6 +36,23 @@ import {
   UserFeedback,
 } from "./types";
 
+const FeedbackPhaseBannerView: React.FC<{
```
DafyddLlyr, line 39: By default, the MUI Drawer retains focus on the initiating element and returns it on close ([docs](https://v5.mui.com/material-ui/react-drawer/#temporary-drawer)).

This wasn't happening here as the main `Feedback` component is a complex component which holds within in it many child components. As a result, on re-render (e.g. change of `FeedbackView`) we lose a stable ref to the button which opened the drawer and focus is lost.

By splitting this out, we are able to maintain a stable ref and the default behaviour kicks back in - resolving this issue.

A wider piece of work here would be to properly split this up and handle the state but is outside the scope of this PR.

editor.planx.uk/src/components/Feedback/FeedbackPhaseBanner.test.tsx
```diff
@@ -55,8 +55,8 @@ describe("FeedbackPhaseBanner accessibility", () => {
   test("should have no accessibility violations", async () => {
     const { container } = setup(
       <FeedbackPhaseBanner
-        handleFeedbackClick={() => {}}
-        handleReportAnIssueClick={() => {}}
+        handleFeedbackClick={vi.fn()}
+        handleReportAnIssueClick={vi.fn()}
```
DafyddLlyr, line 59: Image

jessicamcinchak (approved):

ianjon3s (approved):
Working perfectly for me as well 👍

PR #5054, opened 2025-08-05
feat: Scroll to file card on "change" click

DafyddLlyr:
## What does this PR do?
 - Scrolls to a selected file when the "change" button is hit in file upload and label

## Motivation
 - Raised as usability issue in the accessibility audit
 - More useful than ever now that there's a lot more vertical space within the modal

## Questions
 - Should we keep the modal title in view at all times?

https://github.com/user-attachments/assets/09053238-9dc0-40c1-add1-a306a2144656


ianjon3s (approved):
Brilliant, I have tested with both long and short file lists and this is working perfectly now. Card is scrolled into view and the modal scroll behaviour is as expected.

DafyddLlyr:
Perfect - thanks for catching and testing. I've added a tag on the PR to indicate that this can be merged post-audit.

DafyddLlyr:
Perfect - thanks for catching and reviewing!

PR #5058, opened 2025-08-06
feat: Show progress indicator to users

DafyddLlyr:
## What does this PR do?
- Refactor: Move `NavBar` to its own file and out of `Header.tsx`
- Adds a `<ProgressBar/>` component to the section nav bar
  - Styles by @ianjon3s - https://github.com/theopensystemslab/planx-new/pull/5060
- Progress is measured by section weight, which is then converted to a percentage for the MUI `LinearProgress` component
  - The current section is shown as a "buffer" to indicate the relative size
  - Weights are scaled exponentially - in future if we find we need more categories or a different scale we can easily amend this
 - Persists the section navbar when section nodes are reached for a more consistent experience. Previously I believe this was hidden as we didn't have a solution for how to count a section node as being the start of the next section. See proposed solution here - 

https://github.com/theopensystemslab/planx-new/blob/5703844f5024a68221e808f41f0429298df31587/editor.planx.uk/src/pages/FlowEditor/lib/store/navigation.ts#L91-L94

https://github.com/user-attachments/assets/435a3e0c-dade-47fe-92f8-1d45fc3e467d

### DIFF
--- a/editor.planx.uk/src/components/AnalyticsDisabled/AnalyticsDisabledBanner.tsx
+++ b/editor.planx.uk/src/components/AnalyticsDisabled/AnalyticsDisabledBanner.tsx
@@ -36,7 +36,7 @@ const AnalyticsDisabledBanner: React.FC = () => {
       {isAnalyticsDisabled() && showAnalyticsWarning && (
         <AnalyticsWarning>
           <Container
-            maxWidth={false}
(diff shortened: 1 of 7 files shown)


ianjon3s (approved):
Tested on a real service and all is working as expected

PR #5059, opened 2025-08-06
fix: display flow name when archiving a flow

jessicamcinchak:
I'm in the process of cleaning up old services in the "Testing" team and finding it quite unhelpful how generic the existing "Archive" dialog is - so suggesting a few quick sytle updates here! 

**Changes:**
- Clearly display name of flow you're about to archive
- "Warning" button style for "Archive" action
- Update "remove it from PlanX" language as this isn't really technically accurate!

**Before:**
Screenshot from 2025-08-06 13-17-25

**After:**
Screenshot from 2025-08-06 13-17-09


DafyddLlyr (approved):
Much improved 👍

PR #5070, opened 2025-08-08
chore: hard delete all archived flows in "Testing" team

jessicamcinchak:
Final step in clean up here https://trello.com/c/OQhyVSjp/3372-templates-remove-customisation-tags & here https://opensystemslab.slack.com/archives/C07F7215W7P/p1754493778238819

Adds all Hasura-"suggested" relationships to `flows` table first so that deletion naturally cascades correctly.

**Pizza testing:**
- Manually confirm the editor still displays the same 56 flows as production 
Screenshot from 2025-08-08 10-15-35

- Run the following snippets in the Hasura console's `Raw SQL` window and confirm each return exactly 0
```sql
SELECT COUNT(*) FROM flows
WHERE team_id = 5 -- Testing
AND deleted_at IS NOT NULL;

SELECT COUNT(*) 
FROM operations -- sub in 'published_flows', 'lowcal_sessions', flow_status_history', 'analytics' etc
WHERE flow_id NOT IN (
   SELECT id
   FROM flows
);
```

### DIFF
--- a/hasura.planx.uk/metadata/databases/default/tables/public_flows.yaml
+++ b/hasura.planx.uk/metadata/databases/default/tables/public_flows.yaml
@@ -25,6 +25,13 @@ object_relationships:
           name: templated_flow_edits
           schema: public
 array_relationships:
+  - name: analytics
+    using:
+      foreign_key_constraint_on:
+        column: flow_id
+        table:
+          name: analytics
+          schema: public
   - name: comments
     using:
       manual_configuration:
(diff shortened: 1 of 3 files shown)


DafyddLlyr (commented):
Thanks for this and the testing notes 👍 

I can see 56 flows in the testing team, but the following snippet still returns 262 flows - 

```sql
SELECT COUNT(*) FROM flows
WHERE team_id = 5 -- Testing
AND deleted_at IS NOT NULL;
```

jessicamcinchak:
Aha thanks @DafyddLlyr :bulb:

Was just coming back to this one. Will merge to staging now :slightly_smiling_face:

PR #5074, opened 2025-08-09
fix(a11y): Focus order (`DAC_Focus_Order_01`)

DafyddLlyr:
## What's the problem?
From latest a11y audit, issue `DAC_Focus_Order_01` (page 19) - 

> _When the pages load, focus is not taken to the start of the page. This may be navigationally disorientating for users reliant on the use of the keyboard alone to navigate. This issue is throughout._

Trello ticket - https://trello.com/c/tMNFxDG2/3393-a-focus-order-p19

## Solution
There’s not much detail on this one, and focus seems to be behaving correctly. Here’s what they may possibly be referring to -

- [Building a robust frontend using progressive enhancement](https://www.gov.uk/service-manual/technology/using-progressive-enhancement#single-page-applications)

- [Other keyboard focus considerations — Web Accessibility Guide — NZ Government](https://govtnz.github.io/web-a11y-guidance/ka/accessible-ux-best-practices/keyboard-a11y/keyboard-focus/other-keyboard-focus-considerations.html?utm_source=chatgpt.com#focus-in-single-page-applications-spas)

As there’s no route change when the “Continue” button is pressed there’s no indicator other than the title change. When a user navigates the service going from card to card, the focus seems to go back to body on each card change (tested with `document.activeElement`).

I've solved this by making the `CardHeader` take initial focus - the `h1` (`title`) element. Where components do not use this (`Notice`, `Content`) I've implemented a custom check.

## Questions
I’m not totally sure on the above so it’s worth another pair of eyes here and maybe a question back to DAC to confirm.

 - Does this actually identify and fix the described issue?
 - Would it actually be sufficient to (more simply) make the `<MainContent/>` element have this behaviour? I think this would not work as this component is stable on navigation to the next card
 - Are there components I've missed?

### DIFF
--- a/editor.planx.uk/src/@planx/components/Content/Public.tsx
+++ b/editor.planx.uk/src/@planx/components/Content/Public.tsx
@@ -7,7 +7,7 @@ import type { Content } from "@planx/components/Content/model";
 import Card from "@planx/components/shared/Preview/Card";
 import { PublicProps } from "@planx/components/shared/types";
 import { useAnalyticsTracking } from "pages/FlowEditor/lib/analytics/provider";
-import React from "react";
+import React, { useEffect } from "react";
 import { getContrastTextColor } from "styleUtils";
 import { emptyContent } from "ui/editor/RichTextInput/utils";
 import ReactMarkdownOrHtml from "ui/shared/ReactMarkdownOrHtml/ReactMarkdownOrHtml";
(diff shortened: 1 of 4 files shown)


jessicamcinchak:
a11y results are in and can safely close this one now in favor of #5079 :heavy_plus_sign:

PR #5075, opened 2025-08-09
fix(a11y): Associate `DateInput` header with form group label

DafyddLlyr:
image

Structure now matches suggestion in audit report. No change is required for the `DateInputField` on pages as this already uses a `fieldset` and `legend`.

Trello ticket: https://trello.com/c/Zc3wCtRm/3392-a-form-groups-p17


ianjon3s (approved):
Tested locally and structure matches that suggested by report.

Input group is correctly associated with title using `aria-labelledby`, and `aria-describedby` conditionally appears if either description text or error is present.

PR #5020, opened 2025-07-29
refactor: chages how lable is associated with number input

nichgalzin:
This pull request refactors the `NumberInputComponent` in `Public.tsx` to improve accessibility and code maintainability. The changes focus on dynamically generating unique IDs for input elements, simplifying `aria-describedby` logic, and restructuring how labels are handled.

### Accessibility Improvements:
* Introduced `useId` to dynamically generate unique IDs for input and label elements, ensuring better accessibility and avoiding potential ID conflicts. (`[editor.planx.uk/src/@planx/components/NumberInput/Public.tsxL6-R22](diffhunk://#diff-a05895331276d30c6db914f0148d5fec10a8abe017bf510dae0eef1833c5b974L6-R22)`)
* Updated the `aria-describedby` property to simplify its logic and ensure it references the correct error message or label dynamically. (`[editor.planx.uk/src/@planx/components/NumberInput/Public.tsxL63-R76](diffhunk://#diff-a05895331276d30c6db914f0148d5fec10a8abe017bf510dae0eef1833c5b974L63-R76)`)
* Added the `aria-labelledby` property to associate the input with its label for improved screen reader compatibility. (`[editor.planx.uk/src/@planx/components/NumberInput/Public.tsxL63-R76](diffhunk://#diff-a05895331276d30c6db914f0148d5fec10a8abe017bf510dae0eef1833c5b974L63-R76)`)

### Code Restructuring:
* Removed the `InputLabel` wrapper around the `Input` component and replaced it with a more direct association using dynamically generated IDs. (`[editor.planx.uk/src/@planx/components/NumberInput/Public.tsxL53](diffhunk://#diff-a05895331276d30c6db914f0148d5fec10a8abe017bf510dae0eef1833c5b974L53)`)
* Updated the `InputRowLabel` to include `inputProps` with the dynamically generated label ID, aligning it with the new accessibility structure. (`[editor.planx.uk/src/@planx/components/NumberInput/Public.tsxL63-R76](diffhunk://#diff-a05895331276d30c6db914f0148d5fec10a8abe017bf510dae0eef1833c5b974L63-R76)`)

PR #5076, opened 2025-08-10
fix(a11y):  Number input labels (`DAC_Labels_not_Programmatically_Associated_01`)

DafyddLlyr:
This is aiming to resolve a new "A" issue flagged by the audit - `DAC_Labels_not_Programmatically_Associated_01`

> _The label provided to enable users to identify the purpose of the edit field is not fully programmatically associated with the component. Screen reader users encountering the field ‘out of context’, for example, when filtering the page by element type, will encounter the component as ‘units’. Users would have to switch navigational methods, and search through the surrounding page content in order to identify the purpose of the control._

Trello ticket: https://trello.com/c/7MidNreB/3390-a-labels-not-programmatically-associated-1-p13

I believe this was a regression introduced here - https://github.com/theopensystemslab/planx-new/pull/5020

I'm proposing that we just revert this usability fix in order to resolve the new "A" issue. Ideally we'd circle back here and also address the usability issue also.

### DIFF
--- a/e2e/tests/ui-driven/src/helpers/userActions.ts
+++ b/e2e/tests/ui-driven/src/helpers/userActions.ts
@@ -244,8 +244,8 @@ export async function answerNumberInput(
     continueToNext: boolean;
   },
 ) {
-  await expect(page.locator("h1", { hasText: expectedQuestion })).toBeVisible();
-  await page.locator("div input[type='number']").fill(answer.toString());
+  await expect(page.locator("p", { hasText: expectedQuestion })).toBeVisible();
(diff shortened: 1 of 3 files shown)


ianjon3s (approved):
Agree on a revert being the best way forward here.

Our setup needs a broader review here, both to fix the usabiliy issue and clarify our use of labels. We're using a `<label>` for the suffix on number inputs, whereas we should move towards the GDS example where the suffix is not a `<label>` element (their styling is also neat):
https://design-system.service.gov.uk/components/text-input/input-suffix/

PR #5085, opened 2025-08-18
chore: rename portal component types to `Folder` and `NestedFlow`

jessicamcinchak:
Relies on https://github.com/theopensystemslab/planx-core/pull/794

PR #5087, opened 2025-08-19
chore: rename portals and services (UI only)

jessicamcinchak:
An alternative, lighter touch approach to #5085 

One for the dev call - what's our team tolerance for code names being a bit out of sync with UI? This is a deeply engrained one which the original PR just barely starts to unravel! If our main priority is clear communication to editors, does this alone accomplish that?


editor.planx.uk/src/@planx/components/shared/icons.tsx
```diff
@@ -50,7 +50,7 @@ export const ICONS: {
   [TYPES.Filter]: FilterAltOutlined,
   [TYPES.FindProperty]: SearchOutlined,
   [TYPES.Flow]: undefined,
-  [TYPES.InternalPortal]: DoorFrontOutlined,
+  [TYPES.InternalPortal]: Folder,
```
DafyddLlyr, line 53: Good catch!

DafyddLlyr (commented):
Looking at all 3 PRs I'd favour this approach - I think the drift between public facing labels and code-base variable naming is acceptable given the level of complexity involved in a migration here 👍

PR #5088, opened 2025-08-19
feat: Ensure options have unique labels (question component)

DafyddLlyr:


editor.planx.uk/src/@planx/components/Question/Editor.tsx
```diff
@@ -102,18 +102,10 @@ export const Question: React.FC<Props> = (props) => {
 
   const focusRef = useRef<HTMLInputElement | null>(null);
 
-  // horrible hack to remove focus from Rich Text Editor
-  useEffect(() => {
-    setTimeout(() => {
-      (document.activeElement as any).blur();
-      focusRef.current?.focus();
-    }, 50);
-  }, []);
```
DafyddLlyr, line 111: I believe this is a historic issue which isn't required. We don't do this in any other components and I can't see any behavioural differences with this removed.

PR #5092, opened 2025-08-19
fix: ensure that "emptied out" templated folders are not overwritten on publish

jessicamcinchak:


api.planx.uk/modules/webhooks/service/updateTemplatedFlowEdits/helpers.ts
```diff
@@ -19,11 +23,13 @@ export const transformDeltaToTemplatedFlowEditsData = (
...
 
-    const updatedEdges = nodeData?.edges;
+    // Always preserve a templated folders' "edges", even if it's been "emptied" out
+    const updatedEdges =
+      nodeData?.edges || nodeData?.type === ComponentType.InternalPortal;
 
     templatedFlowEditsData[nodeId] = {
       ...(Object.keys(updatedNodeData).length > 0 && { data: updatedNodeData }),
-      ...(updatedEdges && { edges: data[nodeId]?.["edges"] }),
+      ...(updatedEdges && { edges: data[nodeId]?.["edges"] || [] }),
```
jessicamcinchak, line 32: Previous behavior would have been to omit the `edges` property all together if the node was edited to remove it's edges, but we want to explicitly store the empty array `[]` for correct reconciliation later

api.planx.uk/modules/webhooks/service/updateTemplatedFlowEdits/helpers.ts
```diff
@@ -19,11 +23,13 @@ export const transformDeltaToTemplatedFlowEditsData = (
       updatedNodeData[updatedKey] = data[nodeId]?.["data"]?.[updatedKey];
     });
 
-    const updatedEdges = nodeData?.edges;
+    // Always preserve a templated folders' "edges", even if it's been "emptied" out
+    const updatedEdges =
+      nodeData?.edges || nodeData?.type === ComponentType.InternalPortal;
```
jessicamcinchak, line 28: I _believe_ `ComponentType.InternalPortal` is the only edge case we need to explicitly account for here because it's valid for it to be completely empty. Upload and Label `rules` and Question or Checklist `options` cannot be "saved" if empty arrays.

Any others I might be forgetting?

api.planx.uk/modules/flows/updateTemplatedFlow/service.ts
```diff
@@ -95,3 +97,32 @@ export const updateTemplatedFlow = async (
...
+  // Check for orphaned nodes IDS (deleted edges) and remove them to ensure graph displays correctly
+  const validEdges = Object.values(mergedData).flatMap(
+    (nodeData) => nodeData.edges,
+  );
+  const orphanedNodeIds = Object.keys(mergedData).filter(
+    (nodeId) => nodeId !== "_root" && !validEdges.includes(nodeId),
+  );
+  orphanedNodeIds.forEach((nodeId) => {
+    delete mergedData[nodeId];
+  });
```
jessicamcinchak, line 125: The `mergeWith` above correctly reconciles arrays which means that our _navigation_ store methods will work correctly. We need to do this additional step of ensuring the overall flow graph doesn't have any now-orphaned nodes lingering to prevent them from displaying in the graph and "Search" or "Customise" sidebars.

api.planx.uk/modules/flows/updateTemplatedFlow/customMerge.test.ts
```diff
@@ -0,0 +1,289 @@
...
+            moreInformation: {
+              info: "<p>It shows where the work is happening</p>",
+            },
+          },
+        ],
+      },
+    },
+  };
+
+  expect(customMerge(sourceTemplate, templatedFlowEdits)).toEqual({
```
jessicamcinchak, line 255: By introducing a `customMerge` we now get added benefit of easier to unit test - we can essentially mock mini source flows and their edits and see what the reconciliation produces

DafyddLlyr (approved):
Great - super clear, and nice unit test coverage! ✅

PR #5094, opened 2025-08-20
feat: Save progress to flow data

DafyddLlyr:
## What does this PR do?
Adds `Progress` object to user's session data.

## Motivation
Required to show progress in LPS. There's no need to read this data within PlanX directly - we can always just calculate it via the store as we do currently.

image

### DIFF
--- a/editor.planx.uk/src/components/Header/Sections/ProgressBar.tsx
+++ b/editor.planx.uk/src/components/Header/Sections/ProgressBar.tsx
@@ -27,9 +27,13 @@ const Root = styled(LinearProgress)(({ theme }) => ({
 }));
 
 export const ProgressBar: React.FC = () => {
-  const { completed, current } = useStore((state) =>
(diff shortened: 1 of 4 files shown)


editor.planx.uk/src/components/Header/Sections/ProgressBar.tsx
```diff
@@ -27,10 +27,14 @@ const Root = styled(LinearProgress)(({ theme }) => ({
 }));
 
 export const ProgressBar: React.FC = () => {
-  const { completed, current } = useStore((state) =>
+  const progress = useStore((state) =>
     state.getSectionProgress(),
   );
 
+  if (!progress) return;
```
DafyddLlyr, line 34: This is just a type guard - the parent `SectionNavBar` will already be hidden if the flow does not use sections.

jessicamcinchak (approved):
Great !

PR #5111, opened 2025-08-20
Revert "Revert "chore: unpersist `user_role` in `analytics_summary` [skip pizza]""

DafyddLlyr:
Reverts theopensystemslab/planx-new#5097

This correctly re-instates #5080, and removes the `user_role` table


freemvmt (approved):
💯

PR #5116, opened 2025-08-21
feat: create materialized view of analytics_summary

zz-hh-aa:
Testing out materialized views for Metabase performance issue. Since this is creating a _new_ view and isn't touching the existing `analytics_summary` , am I right in that we can merge this to prod, test out if the new view improves performance or not, and--if we decide it's useful--then we set a Hasura cron trigger up?


DafyddLlyr (approved):
Great! I'll queue up a prod deploy just now so we can test this 🧪

PR #5117, opened 2025-08-21
add script for seeding local db with lowcal_sessions data

freemvmt:
as per title


scripts/seed-database/write/lowcal_sessions.sql
```diff
@@ -0,0 +1,63 @@
...
+  updated_at,
+  email,
+  flow_id,
+  deleted_at,
+  submitted_at,
+  has_user_saved,
+  sanitised_at,
+  locked_at,
+  allow_list_answers
+FROM
```
jessicamcinchak, line 47: Totally understand the utility of this, but it's one we've avoided in the past so that we don't encourage pulling/keeping user data on our personal machines - I wonder if we could sync just a subset of columns here (notably omitting `email` and full `data`) and still get same local utility? I think something like `SELECT null as email, '{}' as data...` would work and preserve expected table schema?

`alllow_list_answers` is non-personally-identifiable but will still include application type, project types, proposal description and more for any record that has not null `submitted_at` which I'm guessing is what you're most after?

PR #5124, opened 2025-08-26
chore: add materialized view to yaml

zz-hh-aa:
The materialized view created in #5116 wasn't showing up. Here's a one-liner that should track the change (as discussed [here](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1756224725828469)).


freemvmt (approved):
Nice catch ⚾

PR #5125, opened 2025-08-26
reduce metabase role statement timeout to 15mins

freemvmt:


DafyddLlyr (approved):

freemvmt:
> Not sure this is quite right - integration tests are failing as vitest is not found.

yeah not sure why that is - gonna have a look today :)

PR #5127, opened 2025-08-27
chore: retire flow document templates

jessicamcinchak:


hasura.planx.uk/migrations/default/1756281219229_drop_tables_flow_document_templates/down.sql
```diff
@@ -0,0 +1,4 @@
+-- Could not auto-generate a down migration.
+-- Please write an appropriate down migration for the SQL below:
+-- DROP TABLE document_template CASCADE;
+-- DROP TABLE flow_document_templates CASCADE;
```
jessicamcinchak, line 4: Intentionally omitting a `down` migration here because:
- we have zero intention of ever re-creating these tables, 
- and in the future event that we do need to re-introduce a `.doc` file in the zips, I would _not_ manage that via a db column but rather based on session-data logic like other document types

DafyddLlyr (approved):

PR #5130, opened 2025-08-27
Production deploy

DafyddLlyr:


freemvmt (approved):
Looks guuurd - let's keep an eye when you deploy this since Hasura rollouts have been playing up?

PR #5139, opened 2025-08-28
chore: Optimisations for progress bar rendering

DafyddLlyr:
This is a companion to #5141 - just a few small improvements I noticed when digging deeper into optimising the `record()` function.

I'm not convinced this makes a real noticeable difference, but it is slightly less computationally heavy (and results in fewer re-renders).

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/lib/store/navigation.ts
+++ b/editor.planx.uk/src/pages/FlowEditor/lib/store/navigation.ts
@@ -29,7 +29,8 @@ export interface NavigationStore {
   filterFlowByType: (type: TYPES) => Store.Flow;
   getSortedBreadcrumbsBySection: () => Store.Breadcrumbs[];
   getSectionForNode: (nodeId: string) => SectionNode;
-  getSectionProgress: () => Progress | undefined;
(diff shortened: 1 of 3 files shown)


editor.planx.uk/src/pages/FlowEditor/lib/store/navigation.ts
```diff
@@ -97,16 +100,22 @@ export const navigationStore: StateCreator<
...
+
     // No sections in breadcrumbs, first section values already set in store
     if (!mostRecentSectionId) return;
 
     const currentSectionTitle = sectionNodes[mostRecentSectionId].data.title;
     const currentSectionIndex = sectionIds.indexOf(mostRecentSectionId) + 1;
+    const sectionProgress = _calculateSectionProgress(currentSectionIndex);
 
-    set({ currentSectionTitle, currentSectionIndex });
+    set({ currentSectionTitle, currentSectionIndex, sectionProgress });
```
DafyddLlyr, line 118: Before `set()` was being called multiple times resulting in re-renders - now it's all batched as a single update.

jessicamcinchak (approved):

PR #5142, opened 2025-08-29
feat: Optimise `removeOrphansFromBreadcrumbs()`

DafyddLlyr:


editor.planx.uk/src/pages/FlowEditor/lib/__tests__/mocks/removeOrphansFlow.json
DafyddLlyr, line 1: I'm not very keen on committing this to `main` really - I find having these large flow JSONs can really slow down local development and we end up with brittle tests.

I'm happy enough for this file and associated test to live as artifacts on this branch (I'll add a comment on existing unit tests pointing here). I'll then remove them before this PR progresses to `main`.

PR #5147, opened 2025-08-29
chore: Fix React errors on `PlanningConstraint` component

DafyddLlyr:
Two small issues I spotted when working on #5142 

image


jessicamcinchak (approved):
Thanks for these !

PR #5150, opened 2025-08-30
fix: Sanitise feedback allow list answers

DafyddLlyr:
## What's the problem?
On production, the Hasura migration from https://github.com/theopensystemslab/planx-new/pull/5116 is failing. This leaves Hasura in the unhealthy "inconsistent metadata" state, leading to rollbacks on AWS Fargate. This has no impact on the GraphQL API as the migration is adding the unused `analytics_summary_materialized` view.

The error is not reproducible locally or on staging - the same migration succeeds without issue on these environments.

## What's the cause?
When the materialised view is being created on production, it's running for the first time with real data. When this happen, it fails at the following line - 

https://github.com/theopensystemslab/planx-new/blob/3cdbf9636ca44bc7f6f3902b81a3ad59c4bd1a32/hasura.planx.uk/migrations/default/1755790603644_create_materialized_analytics_summary/up.sql#L55

There are records where `feedback_score` is `""` - this can't be case to an integer, and the query fails. This means that the migration fails, and Hasura rolls back.

## Why don't we have this issue with the `analytics_summary` view?
It turns out we do! The materialised view is just effectively doing a `SELECT *` and hitting inconsistent data immediately. The issue can be reproduced with the existing view with the following query - 

```sql
SELECT * FROM analytics_summary WHERE feedback_score IS NOT NULL;
```

This likely explains some of the timeouts and inconsistent issues we've seen with queries to this view within Metabase - if we hit a record without a populated `feedback_score` value the query would have failed.

## What's the solution?
There's a few steps to this as I see it - 

- Stop writing incorrect / invalid data to the `feedback` table (this PR)
- Merge to production
- Update allow list data to fix records where `""` has been inserted as either a `feedbackScore` or `userComment`
  - I'll pick this up as a task once this PR has been reviewed and approved
- Deploy to production again and the migration should succeed (if there's no other issues with the view!)

### DIFF
--- a/hasura.planx.uk/migrations/default/1756799907578_set_fk_public_feedback_flow_id/down.sql
+++ b/hasura.planx.uk/migrations/default/1756799907578_set_fk_public_feedback_flow_id/down.sql
@@ -1,0 +1,5 @@
+alter table "public"."feedback" drop constraint "feedback_flow_id_fkey",
+  add constraint "feedback_flow_id_fkey"
+  foreign key ("flow_id")
+  references "public"."flows"
+  ("id") on update restrict on delete restrict;
--- a/hasura.planx.uk/migrations/default/1756799907578_set_fk_public_feedback_flow_id/up.sql
+++ b/hasura.planx.uk/migrations/default/1756799907578_set_fk_public_feedback_flow_id/up.sql
@@ -1,0 +1,5 @@
+alter table "public"."feedback" drop constraint "feedback_flow_id_fkey",
+  add constraint "feedback_flow_id_fkey"
+  foreign key ("flow_id")
+  references "public"."flows"
+  ("id") on update cascade on delete cascade;


editor.planx.uk/src/pages/FlowEditor/lib/analytics/utils.ts
```diff
@@ -131,22 +133,34 @@ export function getAnswers(
...
 export function getData(
   breadcrumb: Store.UserData,
-): Partial<Record<AllowListKey, any>> | undefined {
+): Partial<Record<AllowListKey, unknown>> | undefined {
   const dataSetByNode = breadcrumb.data;
   if (!dataSetByNode) return;
 
   const filteredEntries = Object.entries(dataSetByNode)
+    .map(sanitiseAllowListValues)
     .filter(([key, value]) => isAllowListKey(key) && Boolean(value))
```
DafyddLlyr, line 160: 🐛 Bug! Here's the reason we're saving empty feedback values as allow list answers.

`Boolean(value)` is too simplistic when checking that allow list values are populated. In our feedback example the current object is always truthy (which passes this check).

```ts
{ 
  feedbackScore: "",
  userComment: "",
}
```

Another approach here would be to have typed allow list keys _and_ values, but this is a more complex task without a properly typed graph (we'll just have to resort to casting or parsing via Zod to make it actually work currently).

editor.planx.uk/src/@planx/components/Feedback/model.ts
```diff
@@ -16,9 +16,10 @@ export interface Feedback extends BaseNodeData {
   disclaimer?: string;
   feedbackRequired: boolean;
 }
+
 export interface FormProps {
-  feedbackScore: 1 | 2 | 3 | 4 | 5;
-  userComment: string;
+  feedbackScore?: 1 | 2 | 3 | 4 | 5;
+  userComment?: string;
 }
```
DafyddLlyr, line 23: These properties are optional within the form (unless `feedbackRequired` is set).

editor.planx.uk/src/@planx/components/Feedback/Public/Public.tsx
```diff
@@ -36,22 +36,25 @@ export const PASSPORT_FEEDBACK_KEY = "_feedback";
...
       },
     );
-    props.handleSubmit?.(makeData(props, values, PASSPORT_FEEDBACK_KEY));
-    if (!submitFeedbackResult) {
-      return;
-    }
+  }
+
+  const handleSubmitFeedback = async (values: FormProps) => {
+    if (values.feedbackScore || values.userComment) await logFeedback(values);
```
DafyddLlyr, line 54: Is adding this guard clause right...? Was there a business case for always logging all feedback here that I'm forgetting (e.g. always capturing user context or other metadata)?

editor.planx.uk/src/lib/feedback.ts
```diff
@@ -95,7 +95,11 @@ export async function insertFeedbackMutation(data: {
         }
       }
     `,
-    variables: data,
+    variables: {
+      ...data,
+      feedbackScore: data.feedbackScore || null,
+      userComment: data.userComment || null,
+    },
```
DafyddLlyr, line 102: 🐛 Bug! We're currently passing an empty string into this function is a user does not select a `feedbackScore`.

This can't be written to this column (nullable integer) so we end up throwing an error here and not capturing the feedback at all.

https://github.com/user-attachments/assets/7abada20-4e3d-479b-aca0-7b952107b1a3

PR #5179, opened 2025-09-04
feat: Toggle within PlanX allowing users to list services on LPS

DafyddLlyr:
## What does this PR do?
Adds to a toggle allowing Editors to control if their service will be listed on LPS, alongside basic validation.

Notably, we still have checks in place on LPS for slugs, so this isn't yet entirely configurable here. The next step would have to be a way within PlanX to categorise a service.

```ts
const NOTIFY_SERVICE_SLUGS = [
  "report-a-planning-breach",
  "camden-report-a-planning-breach",
];
```

image

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings/LPS/LPSListingSection.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings/LPS/LPSListingSection.tsx
@@ -85,7 +85,7 @@ export const LPSListing = () => {
           </p>
           <p>
             Listing your service requires a description and summary. These can
-            we provided on{" "}
+            be provided on{" "}
             <Link href="../about">the "About this flow" page</Link>.
           </p>
         </SettingsDescription>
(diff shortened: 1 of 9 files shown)


editor.planx.uk/src/pages/FlowEditor/components/Settings/ServiceSettings/LPS/LPSListingSection.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  return (
+    <Box component="form" onSubmit={form.handleSubmit} mb={2}>
+      <SettingsSection>
+        <Typography variant="h2" component="h3" gutterBottom>
+          Listings
+        </Typography>
+        <Typography variant="body1">
+          Manage how this service is listed and indexed
+        </Typography>
```
DafyddLlyr, line 61: Very open to ideas here when it comes to the copy and how we present this to users.

localplanning.services/src/lib/lpa-api/query.ts
DafyddLlyr, line 1: Intentionally commented out here. As a query is made to _staging_ as part of the Pizza CI due to sequencing, we need to wait for this PR to be merged before we can test on a Pizza.

PR #5072, opened 2025-08-08
fix(a11y): Associate label to `SelectInput` (`DAC_Labels_not_Programmatically_Associated_02`)

DafyddLlyr:
Trello: https://trello.com/c/FgwFBbwM/3391-a-labels-not-programmatically-associated-2-p16

## What's the solution?
 - Remove redundant label
 - Correctly associate label and select

PR #5181, opened 2025-09-04
fix: MapAndLabel 'copy feature' selection

DafyddLlyr:
## What's the problem?
When using the "copy feature" select in the `MapAndLabel` component, it's only possible to select the first item. The `onChange()` handler is not working as expected.

Reported here (video and write up) by @augustlindemer - https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1756986067330309

## What's the cause?
This regression was introduced here - https://github.com/theopensystemslab/planx-new/pull/5072

When we updated the `name` property to meet a11y requirements, we missed the fact that this is also used by formik internally for it's build in `handleChange` function.

## Why wasn't this caught by unit tests?
We actually have pretty good coverage here, but the example used for "copy feature" relied on the first option of the select being clicked - this didn't actually trigger a failing call to `handleChange()`. I've now updated the tests to account for this.

## What's the solution?
Use a custom function for `handleChange()` which does not rely on the `name` property directly. This now meets both accessibility and functional requirements.

### DIFF
--- a/editor.planx.uk/src/@planx/components/MapAndLabel/Public/CopyFeature.tsx
+++ b/editor.planx.uk/src/@planx/components/MapAndLabel/Public/CopyFeature.tsx
@@ -52,7 +52,9 @@ export const CopyFeature: React.FC<Props> = ({
             required
             title={"Copy from"}
             value={formik.values.sourceLabel}
-            onChange={formik.handleChange}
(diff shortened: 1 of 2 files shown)


augustlindemer (approved):
Works!

jessicamcinchak (approved):

PR #5192, opened 2025-09-05
fix: Map validated slots to payload

DafyddLlyr:
## What's the problem?
Users are reporting that some applications do not have all associated files in the the payload zip. When searching S3 we can see that all files were successfully uploaded, but the breadcrumbs/passport do not reflect this. Files in the passport have a status of "uploading" despite a validation check for exactly this.

## What's the case?
If a file is still in an "uploading" status when it's tagged, a snapshot is taken at that current moment which is mapped to the payload, not just a reference. On submit, we validate that all files are uploaded, but not that the generated payload has been updated with the most recent status of each file (which includes the generated URL).

## What's the solution?
When the payload is build, use the ID which was associated when the file was labelled to map the validated file slots. This ensures that the final payload is composed of a validated file list, and validated file slots.


jessicamcinchak (approved):
Thanks for getting to bottom of this one!!

PR #5194, opened 2025-09-08
feat: Don't prompt for additional email when redirecting from LPS

DafyddLlyr:
This improves the login and resume user journey for LPS (details below) - this was flagged as a pain point during UR.

I'm of the opinion that this does not introduce an additional element of risk as the user is already authenticated by proving they have access to their inbox.

**Before**
- User has to enter email address on LPS
- Navigate to their inbox
- Click magic link, navigate to LPS
- Find application in list, navigate to PlanX
- Re-enter email address
- Proceed with application

**Now**
- User has to enter email address on LPS
- Navigate to their inbox
- Click magic link, navigate to LPS
- Find application in list, navigate to PlanX
- ~Re-enter email address~ ❌ 
- Proceed with application

https://github.com/user-attachments/assets/e8ec4c1e-a053-430e-9a1a-f010dc310507

### DIFF
--- a/localplanning.services/src/lib/utils.ts
+++ b/localplanning.services/src/lib/utils.ts
@@ -15,3 +15,4 @@ export const getServiceURL = ({
   domain
     ? `https://${domain}/${serviceSlug}`
     : `${PUBLIC_PLANX_EDITOR_URL}/${lpaSlug}/${serviceSlug}/published`;
+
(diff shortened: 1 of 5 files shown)


jessicamcinchak (approved):

localplanning.services/src/lib/utils.ts
DafyddLlyr, line 1: Intentional whitespace change to trigger a build of LPS for this Pizza so that I can test this outside localhost.

PR #5208, opened 2025-09-09
fix: Allow null origin for map tile requests

DafyddLlyr:
## What does this PR do?
 - Moves CORS setup to it's own `cors.ts` file to tidy up `server.ts` a little
 - Allows `"null"` as a falsy origin

Context for this change can be found here - https://github.com/theopensystemslab/planx-core/pull/812

We can't actually test this on a Pizza due to a known issue with fetching map tiles which is separate to this CORS issue - we'll need to merge to `main` to test if this works with a local HTML file.


api.planx.uk/cors.ts
```diff
@@ -0,0 +1,33 @@
+import cors, { type CorsOptions } from "cors";
+
+const checkAllowedOrigins: CorsOptions["origin"] = (origin, callback) => {
+  if (!origin || origin === "null") return callback(null, true);
```
DafyddLlyr, line 4: This line is the only change here - when making requests from a local HTML file, origin is set to the truthy `"null"`;

jessicamcinchak (approved):

PR #5219, opened 2025-09-10
fix: when validating that all required templated nodes have been updated on publish, fetch templated node edits from all nested templated flows too

jessicamcinchak:
Reported via South Gloc - see thread here: https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1757411128487319

### DIFF
--- a/api.planx.uk/helpers.ts
+++ b/api.planx.uk/helpers.ts
@@ -17,6 +17,7 @@ export interface GetFlowDataResponse {
   limitations: string | null;
   team_id: number;
   team: { slug: string };
+  templatedFrom: string | null;
   publishedFlows:
     | {
         data: Flow["data"];
@@ -46,6 +47,7 @@ const getFlowData = async (id: string): Promise<GetFlowDataResponse> => {
           team {
             slug
           }
+          templatedFrom: templated_from
           publishedFlows: published_flows(
             limit: 1
             order_by: { created_at: desc }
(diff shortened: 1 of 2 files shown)


api.planx.uk/helpers.ts
```diff
@@ -418,6 +420,7 @@ const dataMerged = async (
         data: {
           text: `${team.slug}/${slug}`,
           flattenedFromExternalPortal: true,
+          templatedFrom: response.templatedFrom,
```
jessicamcinchak, line 423: When flattening flows, all 'external portals'/'nested flows' get re-categorised as 'internal portals'/'folders'. 

We already have a prop `flattenedFromExternalPortal` which lets us quickly determine it's origin, and we can similar add another prop `templatedFrom` to distinguish if the 'external portal'/'nested flow' is a templated flow itself.

api.planx.uk/modules/flows/validate/service/templatedNodes.ts
```diff
@@ -12,33 +12,51 @@ const validateTemplatedNodes = async (
...
+    nestedTemplatedFlowIds.map((nestedFlowId) =>
+      getTemplatedFlowEdits(nestedFlowId),
+    ),
+  ).then((values) => {
+    return values.flatMap((value) => Object.keys(value.edits?.data || {}));
+  });
+
+  const allEditedTemplatedNodeIds = Object.keys(
+    primaryTemplatedFlowEdits.edits?.data || {},
+  ).concat(editedNestedTemplatedNodeIds);
```
jessicamcinchak, line 44: This is the most important change ! 

When we get `allRequiredTemplatedNodeIds` we iterate over the _flattened_ structure, which naturally picks up required templated nodes within nested templated flows.

But when calculating `allEditedTemplatedNodes`, we were previously only fetching _this_ flows 'edits'/'customisations'. We need to fetch the edits for this flow _and each nested templated flow_ before comparing if every required templated node has been edited :heavy_plus_sign:

ianjon3s (approved):
Validation passes on pizza, vs failing on staging 👍

PR #5221, opened 2025-09-10
fix: Ensure branched components display template wrapper

ianjon3s:
## Issue

[Reported by Jonny in Slack.
](https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1757512373915809)

When viewing a source template, components placed on branches do not currently show the correct wrapper/label if `allow edits` = true.

## Solution

Ensure that `showTemplatedNodeStatus` is correctly passed down every time child nodes are rendered.


jessicamcinchak (commented):
Pizza works for me, thanks for quick ID on this one !

PR #5224, opened 2025-09-11
feat: write hasSections to published_flows and add column to table

zz-hh-aa:
Following discussion on measuring sections impact on user experience, here's a PR that:
- Creates a migration to add a `has_sections` column to `published_flows`
- Writes `hasSections` to the db when `publishFlow()` runs

This time with updated permissions!

Next step is to create another migration which will add this column to `analytics_summary` so we can use this in analytics. (This might become obsolete soon with alternative analytics views but I think this makes sense for now.)

### DIFF
--- a/hasura.planx.uk/metadata/databases/default/tables/public_published_flows.yaml
+++ b/hasura.planx.uk/metadata/databases/default/tables/public_published_flows.yaml
@@ -16,6 +16,7 @@ insert_permissions:
         - created_at
         - data
         - flow_id
+        - has_sections
         - has_send_component
         - id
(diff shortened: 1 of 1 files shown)


DafyddLlyr:
E2E tests here are failing as the `GetFlowDiff` query is calling the tracked Postgres function `diff_latest_published_flow()`. This function needs to return an entire `published_flows` record and this PR changes the shape of that record.

If you update this function, you should then see these tests passing 👍 

Docs: https://hasura.io/docs/2.0/schema/postgres/custom-functions/#pg-supported-sql-functions

PR #5250, opened 2025-09-16
fix `templated_flow_update_requests` migration script

freemvmt:
When building the stack on local (including just on latest from `main`), my Hasura console was reporting inconsistent metadata, like so:

Screenshot From 2025-09-16 17-10-09

I raised this on Slack [here](https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1758024700227139).

After some poking around I eventually noticed that the `up.sql` file for the relevant migration (introduced in #5175) does not explicitly specify the schema to which the view should be applied. The first commit here fixes that.

The second commit is just a minor addendum to ensure that all migrations concerning views are idempotent (i.e. any instance of `CREATE VIEW` is replaced by `CREATE OR REPLACE VIEW`), so will not fail if the view has already been instantiated. This is just good practice but could be discarded if we like. It's mostly a hangover from an initial theory I had as to the issue at hand.

## Testing

Try building the stack from this branch and head to `http://localhost:7100/console/data/default/schema/public` to check that no error is reported and `templated_flow_update_requests` is visible in the list.

### DIFF
--- a/hasura.planx.uk/migrations/default/1730730955039_application_type_as_text_in_view/up.sql
+++ b/hasura.planx.uk/migrations/default/1730730955039_application_type_as_text_in_view/up.sql
@@ -1,6 +1,6 @@
 DROP VIEW IF EXISTS "public"."submission_services_summary";
 
-CREATE VIEW "public"."submission_services_summary" AS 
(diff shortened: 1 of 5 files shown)


DafyddLlyr (approved):
Nice! Thanks for getting to the bottom of this one 👌

PR #5251, opened 2025-09-16
fix(lps): Conditionally called `files` build option on staging and production

DafyddLlyr:
- Generate sitemaps by defining a `site` value in the Astro config
- Only configures the required `file` build on AWS environments, which fixes routing on Pizza environments (served using goStatic)

### DIFF
--- a/localplanning.services/astro.config.mjs
+++ b/localplanning.services/astro.config.mjs
@@ -1,19 +1,21 @@
-// @ts-check
 import { defineConfig, envField, fontProviders } from "astro/config";
 import react from "@astrojs/react";
 import tailwindcss from "@tailwindcss/vite";
-
+import { loadEnv } from "vite";
 import icon from "astro-icon";
+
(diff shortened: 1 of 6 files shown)


localplanning.services/astro.config.mjs
```diff
@@ -1,19 +1,21 @@
...
+import { loadEnv } from "vite";
 import icon from "astro-icon";
 
+const { PUBLIC_LPS_URL, MODE } = loadEnv(process.env.NODE_ENV, process.cwd(), "");
+
 // https://astro.build/config
 export default defineConfig({
   integrations: [react(), icon()],
-  site: "https://localplanning.services",
+  site: PUBLIC_LPS_URL,
```
DafyddLlyr, line 12: Setting this will now generate sitemaps, which bumps up our Lighthouse score a fair bit 💹 

Docs: https://docs.astro.build/en/reference/configuration-reference/#site

jessicamcinchak (approved):

PR #5256, opened 2025-09-17
feat: add describe documents schema

augustlindemer:
This PR adds a 'Describe Documents' list schema that captures file uploads with user generated descriptions.

Request originally raised [here for the general enquiries service](https://opendigitalplanning.slack.com/archives/C01AD3YPZ24/p1758099144732219), but the schema is generic enough to be used in any case where we want a user generated description to one or multiple files.


jessicamcinchak (approved):
Sweet & simple !

PR #5260, opened 2025-09-17
feat: auto-answer input component types if you've seen a previous one of the same type and data field

jessicamcinchak:
See https://opensystemslab.slack.com/archives/C07F7215W7P/p1757580329766889

Trello https://trello.com/c/ZEL48rda/3143-accepting-discrete-number-dates-in-the-user-interface

**Testing:**
- Use flow `planx-university/201-automations-with-inputs` on this pizza and production for comparison
- Expected behavior annotated directly in the flow branches, and specific node patterns are also translated to unit tests here


editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -90,6 +100,7 @@ export interface PreviewStore extends Store.Store {
   saveToEmail?: string;
   overrideAnswer: (fn: string) => void;
   requestedFiles: () => FileList;
+  autoAnswerableInputs: (id: NodeId) => Value | undefined;
```
jessicamcinchak, line 103: `autoAnswerableOptions` and `autoAnswerableFlag` output a node ID which serves as the "answer ID" we can record in the breadcrumb of the auto-answered node, but in the case of input component types we want the breadcrumb to capture the actual user-entered value. 

The `Value` type works well as a bit of a "catch all" here, alternatively we could do a union based on the specific supported component type models like `Address | Contact | string | number` ? Thoughts/preferences?

editor.planx.uk/src/@planx/components/ContactInput/model.ts
```diff
@@ -17,40 +18,42 @@ export type Contact = {
...
-    });
-  },
-});
+      return context.createError({
+        path: "firstName",
+        message:
+          "'Test Test' is not a valid name - please submit test applications via the staging environment",
+      });
+    },
+  });
```
jessicamcinchak, line 50: Minor formatting and renaming adjustments here only to make more consistent with `AddressInput`, no functional change!

PR #5261, opened 2025-09-17
dp/lps wire up delete

DafyddLlyr:
- **feat: Update env vars to handle Hasura REST API**
- **feat: Wire up delete application API endpoint**


ianjon3s:
Agree on the need for a transition here, I'll look to bring something simple in.

Testing on the pizza, I created two draft applications for Barnet, and the delete button takes no effect.

.env.example
DafyddLlyr, line 1: The env files on S3 have also been updated to reflect this change.

DafyddLlyr:
@ianjon3s Finally fixed the CORS issues here - now re-open for review 👌

ianjon3s (approved):
Tested 2x save & return on the pizza and delete is working as expected. I'll add a card to add some level of animation so that it is clearer visually what is happening on reorganisation layout shift

PR #5262, opened 2025-09-17
fix(lps): Handle site and mode env vars in Astro config

DafyddLlyr:
A follow up to https://github.com/theopensystemslab/planx-new/pull/5251 as this approach didn't quite work as planned.

This PR now - 
 - Passes in a `--site` flag at runtime for the build, so that we don't have to work around env vars in the config file
 - Parses `mode` from the runtime build command, so that we can reliably build in `"files"` or `"folder"` mode based on this
 
 This should result in - 
  - Proper sitemaps
  - Correct routing in dev, pizza (docker) and staging/prod (AWS Cloudfront)


ianjon3s (approved):
Routing working correctly across pizza without need for `.html`

PR #5271, opened 2025-09-18
feat: add 'Relevant demolition' to fee breakdown exemptions

augustlindemer:
'Relevant demolition' is a fee exemption for various application types exempting the application from any planning fee. When applicable, we want to show this exemption in the fee breakdown.

Unsure if this single line addition covers all our bases. If not, please flag additional req (tests, mocks etc.)!


jessicamcinchak (approved):
Great, thanks for this one!

PR #5272, opened 2025-09-18
chore(lps): Only fetch services which have explicitly listed for LPS

DafyddLlyr:


localplanning.services/src/pages/directory.astro
```diff
@@ -32,6 +37,9 @@ const lpas = await fetchAllLPAs();
               services={lpa.guidanceServices}
               lpa={lpa}
             />
+            {!hasListedServices(lpa) && (
+              <p>{lpa.name} doesn't yet have any services listed.</p>
+            )}
```
DafyddLlyr, line 42: Very open to ideas for better copy here..!

jessicamcinchak (approved):

PR #5273, opened 2025-09-18
dp/consume download token

DafyddLlyr:
- **refactor: Split up types file**
- **feat: Add session_id column to lps_magic_links table This will allow us to restrict download tokens to a single application at a time**
- **test: Basic coverage for generate token endpoint**
- **feat: Generate download token endpoint**
- **docs: Swagger docs**
- **feat: Ensure only valid sessions can generate download tokens**
- **feat: Add `lps/download/html` endpoint**
- **test: Basic coverae for validation, middleware, and controller**


api.planx.uk/shared/middleware/validate.ts
```diff
@@ -19,6 +19,7 @@ export const validate =
   ) => {
     try {
       const parsedReq = schema.parse({
+        headers: req.headers,
```
DafyddLlyr, line 22: This allows us to also validate `req.headers` with Zod schemas.

PR #5195, opened 2025-09-08
feat: Create Hasura REST endpoint to delete session

DafyddLlyr:
## What does this PR do?
This PR creates a Hasura REST endpoint allowing users to delete their LPS applications, which will be called from the "Delete" button on application cards. The UI is not yet in place for this feature.

image

This uses the same public permissions as the current "reset" button uses within PlanX when `clearLocalFlow()` is called - https://github.com/theopensystemslab/planx-new/blob/ae460ebeeba3ca526ba23f564e48f0a492ffb54f/editor.planx.uk/src/lib/lowcalStorage.ts#L36-L54

To test - 
 - Create a lowcal session record on the Pizza
 - Hit `https://api.5195.planx.pizza/api/rest/lps/application/:applicationId` with the required params

Example cURL - 
```sh
curl -X DELETE 'https://hasura.5195.planx.pizza/api/rest/lps/application/<SESSION_ID>' \
-H 'x-hasura-lowcal-session-id: <SESSION_ID>' \
-H 'x-hasura-lowcal-email: <EMAIL>'
```

Examply Bruno `.bru` file - 
```bru
meta {
  name: Delete LPS application
  type: http
  seq: 1
}

delete {
  url: https://hasura.5195.planx.pizza/api/rest/lps/application/:applicationId
  body: none
  auth: inherit
}

params:path {
  applicationId: <SESSION_ID>
}

headers {
  x-hasura-lowcal-session-id: <SESSION_ID>
  x-hasura-lowcal-email: <EMAIL>
}

settings {
  encodeUrl: true
}
```

PR #5274, opened 2025-09-18
Revert "feat: Create Hasura REST endpoint to delete session"

DafyddLlyr:
Reverts theopensystemslab/planx-new#5195

Unfortunately this won't be a viable option for our requirements - I'm hitting issues with the implementation in https://github.com/theopensystemslab/planx-new/pull/5261

It turns out that the generated endpoint only supports the selected method, in this case `DELETE`. It does not support an `OPTIONS` request which is required by the browser when we're using custom headers (e.g. `x-hasura-lowcal-session-id` or `x-hasura-lowcal-email`).

```sh
curl https://hasura.5261.planx.pizza/api/rest/lps/application \
  -X OPTIONS
{"error":"Method OPTIONS not supported.","path":"$","code":"bad-request"}%   
```

I can't find anything in the docs stating this, so I'll open a GH issue on the Hasura repo to flag this.

The plan of action here is just to revert this, and in #5261 I'll make a GraphQL request for the same operation 👍

### DIFF
--- a/hasura.planx.uk/metadata/query_collections.yaml
+++ b/hasura.planx.uk/metadata/query_collections.yaml
@@ -10,10 +10,3 @@
               usersLast30Days: users_last_30_days
             }
           }
-      - name: LPS - Delete application
-        query: |
-          mutation SoftDeleteLowcalSessionLPS ($applicationId: uuid!) {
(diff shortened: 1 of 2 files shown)


jessicamcinchak (approved):
Bummer!

PR #5277, opened 2025-09-19
feat(lps): Wire up download HTML endpoint, display data within LPS

DafyddLlyr:
## What does this PR do?
 - Final part of the process laid out here - https://github.com/theopensystemslab/planx-new/pull/5273
 - Creates a custom React hook to wrap the `/lps/download/token` and `/lps/download/html` endpoints
 - Display this within the LPS Asto application as `application/:lpa` (branded by team)

https://github.com/user-attachments/assets/728670bb-d6c1-49e8-8e8c-4f22c4ba3388

### DIFF
--- a/api.planx.uk/modules/lps/service/getApplications/index.ts
+++ b/api.planx.uk/modules/lps/service/getApplications/index.ts
@@ -79,6 +79,7 @@ const mapSharedFields = (raw: Application) => ({
   },
   team: {
     name: raw.service.team.name,
+    slug: raw.service.team.slug,
   },
   address: raw.addressLine || raw.addressTitle,
 });
--- a/localplanning.services/src/components/applications/ApplicationViewer.tsx
+++ b/localplanning.services/src/components/applications/ApplicationViewer.tsx
@@ -1,0 +1,23 @@
+import { useHtmlDownload } from "./hooks/useHTMLDownload";
+
+export const ApplicationViewer = () => {
+  const applicationId = new URLSearchParams(window.location.search).get("applicationId")
+  const { htmlContent, loading, error, fetchHtml } = useHtmlDownload(applicationId);
+
+  if (loading) return <p>Loading application...</p>
+
+  if (error) {
+    return (
+      <div>
+        <h2>Failed to load application</h2>
+        <p>Error: {error}</p>
+        <button onClick={fetchHtml} className="button button--primary button--medium button-focus-style">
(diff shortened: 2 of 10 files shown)


localplanning.services/src/components/applications/ApplicationViewer.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <div className="error-state">
+          <h2>Failed to load application</h2>
+          <p>Error: {error}</p>
+          <button onClick={fetchHtml} className="retry-btn">
+            Retry
+          </button>
+        </div>
+      </div>
+    );
+  }
```
DafyddLlyr, line 30: This is pretty basic just now @ianjon3s - we don't really have a pattern here yet for loading and error states. Very open to ideas!

localplanning.services/src/components/applications/ApplicationViewer.tsx
```diff
@@ -0,0 +1,43 @@
+import { useHtmlDownload } from "./hooks/useHTMLDownload";
+
+export const ApplicationViewer = () => {
+  const applicationId = new URLSearchParams(window.location.search).get("applicationId")
```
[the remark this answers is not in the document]
DafyddLlyr: I've actually implemented this here - https://github.com/theopensystemslab/planx-new/pull/5285

Having this information persist in the URL is unsafe and creates a permalink to application data which we want to avoid. The above PR to pointing to `main` - I'll wait for this to be approved and merged, before I rebase this one and open it for review.

api.planx.uk/modules/lps/service/getApplications/index.ts
```diff
@@ -79,6 +79,7 @@ const mapSharedFields = (raw: Application) => ({
   },
   team: {
     name: raw.service.team.name,
+    slug: raw.service.team.slug,
```
DafyddLlyr, line 82: Returning this means that we can populate the team-branded applications page.

DafyddLlyr:
@ianjon3s I think the issue is likely how submissions are set up for you locally - I'd guess the session locally shows as `submitted_at: null` as the submission has maybe failed.

Thanks for the read-over and approval, I'll merge to `main` just now if you're ok to run a test there for me (I'll also do the same).

nichgalzin (commented):
See this was merged as I was looking at it 😅. But had no notes!

PR #5280, opened 2025-09-19
chore(api): Sanitise HTML before returning

DafyddLlyr:
A dose of healthy paranoia! As we're generating the HTML it _should_ be fine, but better safe than sorry as the underlying data is originally user-provided.


DafyddLlyr:
Waiting on #5273 to be reviewed and approved before I merge this one.

PR #5281, opened 2025-09-19
chore(lps): Style directory fallback message to match `[lpa].astro` fallback message

DafyddLlyr:
Quick follow up to https://github.com/theopensystemslab/planx-new/pull/5272 to improve content and style, this should now more closely match the other fallback content we have.

**`directory.asto`**
image

**`[lpa].astro`**
image


ianjon3s (approved):
Nice catch

PR #5285, opened 2025-09-20
feat(lps): Use nanostore for session state, not search params

DafyddLlyr:


localplanning.services/src/pages/[lpa].astro
```diff
@@ -93,21 +93,27 @@ const getFallbackMessage = (serviceType: string) =>
...
     if (!action) return;
 
-    const detailsEl = document.getElementById(`${action}-accordion`);
+    const detailsEl = document.getElementById(
+      `${action}-accordion`
+    ) as HTMLDetailsElement | null;
     if (!detailsEl) return;
 
     detailsEl.open = true;
+    detailsEl.scrollIntoView({ behavior: "smooth" });
```
Comment (DafyddLlyr) [inline localplanning.services/src/pages/[lpa].astro:113]: I've added this back in as we're no longer hitting race conditions related to the route. We can decide to keep or remove, I just wanted to test that this now works as expected.

https://github.com/user-attachments/assets/591e0430-6e0f-4f3e-a8e2-6fcb91a46f43

I _think_ we decided to remove this scroll as it was cropping out the team masthead on navigation, is that right?

Comment (ianjon3s) [reply localplanning.services/src/pages/[lpa].astro:113]: Indeed, we removed in favour of reducing the height of the accordions so that the open state is visible on first load, and keeping the LPA details / branding in view.

Comment (DafyddLlyr) [reply localplanning.services/src/pages/[lpa].astro:113]: Ah that's right - thanks - I'll remove this again so that we keep branding in line 👌

PR #5298, opened 2025-09-25
feat: Copy/paste across flows

DafyddLlyr:
## To do...
 - [ ] Tests
 
 ---

## What does this PR do?
 - Adds a context menu on right click, using [the MUI menu component](https://mui.com/material-ui/react-menu/#context-menu), instead of a hidden cloning feature
 - Adds the ability to copy/paste within a flow
 - Adds the ability to copy/paste across flows

### Context menu
My original plan here (which I lost a fair amount of time on!) was to have a `ClipboardStore` which would separate this logic out from the existing store and be responsible for -
 - Opening/closing the context menu
 - Managing relationships (required for copy/paste operations)
 - Defining the functions called by the menu (copy/paste) and in future maybe delete, make unique etc.

This didn't work for a few reasons, and I needed to just combine this into existing stores. The main limitation is that I needed access to the `doc` (ShareDB) document in order to paste (create) new nodes which I couldn't access without a shared store. To me this points at the need to rationalise some of this store / shared store / store slice code as it's very easy for them to balloon where a more modular approach could be a cleaner design. A nice outcome I was hoping for here would have been to leverage the existing `persist()` middleware layer to handle read/writes from localStorage.

So in the end we have - 
 - Clipboard state within the `EditorUIStore`
 - Node operations within the `EditorStore`

In terms of visual design I've kept this in line with our current context menus (default MUI style basically). This PR is already larger than planned so improvements here can be a follow up. This could include switching this from a context menu (right-click) to a plain menu (left-click, with options like "Add component..." etc.

### Copy / Paste
This ended up being a fair bit trickier than anticipated...! 

Copying and pasting a single node was very simple - copy the data to clipboard, insert a new node with this data. Next up comes child edges (essential for questions, checklists etc) - this also wasn't too bad and also fit within the structure expected by our `addNode()` function - 

```ts
// Shallow copy - just a node and it's immediate children

  copyNode(id) {
    const { flow } = get();
    const node = flow[id];
    const children: Store.Node[] = node.edges?.map((id) => flow[id]) || [];

    const payload = JSON.stringify({ node, children });
    localStorage.setItem("copiedNode", payload);
  },
```

Deep copying (to match how clone works) requires the following steps -
 - Keep copy of original node and all descendants
 - Generate new IDs for each node within the new flow
 - Map new IDs to old IDs (required to maintain relationships)
 - Structure list of nodes to graph structure
 - Call `addNode()`
 
 As with any graph / store function I've tended towards over-explaining, and over-commenting, just to make sure this is intelligible in future.

### DIFF
--- a/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/ContextMenu.tsx
+++ b/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/ContextMenu.tsx
@@ -53,13 +53,16 @@ export const ContextMenu: React.FC = () => {
   ]);
 
   const handleCopy = () => {
-    // TODO better types for `Relationships`?
-    copyNode(self!);
+    if (!self) return alert("Unable to copy, missing value for relationship 'self' (nodeId)")
+
+    copyNode(self);
     closeMenu();
   };
 
   const handleClone = () => {
-    cloneNode(self!);
+    if (!self) return alert("Unable to clone, missing value for relationship 'self' (nodeId)")
+
+    cloneNode(self);
     closeMenu();
   };
 


editor.planx.uk/src/@planx/graph/index.ts
```diff
@@ -173,8 +174,7 @@ const _add = (
     before = undefined,
   }: { children?: Child[]; parent: string; before?: string },
 ) => {
-  if (draft[id]) throw new Error("id exists");
```
DafyddLlyr, line 176: Removing this is required when pasting a collection of nodes which include clones.

My understanding of this is that it's just a guard clause against ID clashes - if we did have a clash it would overwrite the previous node.

Our `uniqueId()` function for clones (line 28 above) has an astronomically small chance of clashing - there's ~839 quadrillion possible outputs 👀

editor.planx.uk/src/pages/FlowEditor/components/Flow/components/ContextMenu.tsx
DafyddLlyr, line 1: This is largely drawn from here - https://mui.com/material-ui/react-menu/#context-menu

DafyddLlyr:
Opened for review, still some tests to come though... 🧪

ianjon3s (approved):
Approving from a functional perspective. Tested on the pizza, working perfectly for both copy and clone, pasting within and between flows.

One edge case that should warrant a notification/warning: cloning between flows rightly defaults to copy, but there isn't a notification as to why a clone isn't possible.

DafyddLlyr:
> _One edge case that should warrant a notification/warning: cloning between flows rightly defaults to copy, but there isn't a notification as to why a clone isn't possible._

I'll take a look at this, I had intended that paste would be disabled in this instance as it's not a valid operation. Thanks for spotting!

editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -574,15 +683,69 @@ export const editorStore: StateCreator<
...
       if (id) {
         const [, ops] = clone(id, { toParent, toBefore })(get().flow);
         send(ops);
       }
-    } catch (err: any) {
-      alert(err.message);
+    } catch (err) {
+      alert((err as Error).message);
+    }
+  },
```
jessicamcinchak, line 696: Question: If trying to "clone" _across_ flows, could we default to "copy" behaviour rather than disable "paste" option all together as now? 

From the current context menu, I'm not sure it's so clear which action is available _in this flow and across flows_ (copy) versus _in this flow only_ (clone) - and maybe rather than figuring out a way to explain that we could just have a smarter fallback behaviour?

PR #5121, opened 2025-08-26
feat: use ODP Schema `Enforcement` schema for all RAB submissions

jessicamcinchak:
Relies on https://github.com/theopensystemslab/planx-core/pull/800

For each council live with RAB, here's a spot-check of recent submissions for each:
| Council       | Session ID |  Validates | Using Power Automate |
|---------------|-----------|-----------|-----------|
| Barnet        | `40eb43a6-375a-4e32-90b3-f6d21c1b4745`|  :x: <br />`application.type` undefined | **Yes** | 
| Bucks         |`cc753f0a-5844-4e29-9218-4e4498999c58`|  :x: <br />`property.type` error | No |
| Camden        |`3514c951-399c-42bf-894a-c4c24c69ba26`| :white_check_mark: | No |
| Doncaster     |`6fcc291b-7673-4af7-939a-0819fc58079d`| :white_check_mark: | **Yes** |
| Epsom & Ewell |`14eb9604-d7f8-434f-9d4f-b31b8bf46987`| :white_check_mark: | No |
| Gateshead     |`40caeaf3-4d79-4905-96cb-d9e5861f0df0`| :x: <br />`application.type` undefined | No |
| Medway        |`679ee9a5-5945-4fba-ab06-8961b5099b4b`| :white_check_mark: | No | 
| Southwark     |`6f4f840e-ecfa-4144-8bc3-db9f10d820fd`| :x: <br />`property.type` error | **Yes, but already failing** |
| Tewkesbury    |`920da347-2c49-48d4-9fdd-e64f91076e30`| :x: <br />`application.type` undefined | **Yes** | 
| West Berks    | `ae233267-ebb0-49a6-865b-de0389ab9dda`| :x: <br />`application.type` undefined | No | 

"Validates" indicates if the existing production submission validates successfully against the new `Enforcement` schema. For each session ID above, we can use the existing admin endpoints to generate before/after payloads:
- "Before" = `api.editor.planx.uk/admin/session/:sessionId/digital-planning-application?skipValidation=true`
- "After" = `api.5121.planx.pizza/admin/session/:sessionId/digital-planning-application`

PR #5310, opened 2025-09-29
Production deploy

DafyddLlyr:
This includes #5212 which is scheduled to be deployed at 11am, 7th October.

If we need to go for a deploy prior to this, I'll revert this from `main`, deploy to prod, then re-merge. As #5121 is being relied on for testing I'll do my best to avoid this if possible, and run this process after-hours if required.


jessicamcinchak (approved):
Approving ahead of tomorrow morning - sorry again that enforcement schema rollout has led to such a long queue here !

PR #5316, opened 2025-09-30
fix: use pizza failover override compose config to build caddy when pull fails

freemvmt:


DafyddLlyr:
Merged so that I can rebase other PRs from this `main` once this is in place 👍

PR #5317, opened 2025-10-01
feat: create smaller materialized analytics views

zz-hh-aa:
This is a new version of #5225 after encountering a migrations tangle. 

`analytics_aggregated_mini` now has a more meaningful name (`analytics_journeys_aggregated`). I couldn't add comments to the new views because they aren't tables. 

Original PR description below:

---
After dev call today I've tested (and amended) the queries using the read only db role through pgAdmin. Happy with these data shapes--I think the next step is merging this and #5211 so that we can test the different shapes in Metabase (and eventually get rid of the ones we won't use, and set up cron jobs to update the views). 

This is possibly an alternative to #5211 (which creates one aggregated, materialised analytics view to satisfy most analytics needs), creating multiple smaller views. This creates 5 different materialized views.

`analytics_aggregated_mini`
A version of `analytics_aggregated_materialized` as created in #5211 but without half of the columns (as the rest have been split into other views). 

`analytics_sessions`
A summary of sessions. Shows `analytics_type`, `number_logs` (number of logs per session), `application_type`, `analytics_created_at`, `platform`, `operating_system`, `referrer` etc. 

`analytics_action_nodes`
This is the one I'm least sure about. I don't know that the shape is intuitive--should they be split into separate views? It's for slightly more granular queries to do with specific nodes: either action on `drawBoundary` or where users clicked for help or encountered errors. Shows results when `input_error` > 0, `has_clicked_help` = `true`, `has_clicked_save` = `true` OR `drawBoundary.action` is not null. `node_type` and `node_title` are columns. 

`analytics_exits`
Shows if `is_user_exit` = true (so whether Beacon API counted it as an exit or not), as well as last node type & title that the user encountered. Also contains `rab.exitReason`. 

`analytics_journey_results`
Basically for anything that has flags. One row per result (so one analytics ID may have multiple results)


DafyddLlyr (approved):
> I couldn't add comments to the new views because they aren't tables.

🤦 TIL! They really should remove that button!

PR #5332, opened 2025-10-01
fix(lps): Click handler for viewer link

DafyddLlyr:
## What's the problem?
ApplicationId persists on back navigation, which doesn't allow user to return to list of applications via the browser "back" button.

## What's the solution?
Application ID should only be set on click, and we should aggressively clear it when navigating back.


localplanning.services/src/components/applications/ApplicationCard.tsx
```diff
@@ -133,11 +133,11 @@ const DeleteButton: React.FC<Application> = ({ id }) => {
 }
 
 const ViewApplicationButton: React.FC<Application> = (application) => {
-  $applicationId.set(application.id)
```
DafyddLlyr, line 136: 🤦‍♂️

PR #5340, opened 2025-10-02
chore: planx_website_stats use tables not views

zz-hh-aa:
When altering `analytics_summary` in #5339 I realised there is a view that is `select`ing from another view, which as we've discussed at a dev call before is a dependency we'd like to avoid. 

This rewrites the view to pull from the source tables that `analytics_summary` and `submission_services_summary` pull from (`analytics` and `lowcal_sessions`)


DafyddLlyr (approved):
Great spot - thanks for tidying up after me! 😅

PR #5346, opened 2025-10-02
fix(lps): `ClientRouter` back navigation issue

DafyddLlyr:


ianjon3s (approved):
Confirming back navigation is now working as expected

PR #5347, opened 2025-10-02
feat(lps): Improve display of HTML application data

DafyddLlyr:


localplanning.services/src/components/applications/ApplicationViewer.module.css
```diff
@@ -0,0 +1,28 @@
+/* Overrides to style injected HTML */
+.viewer-overrides {
+  :global h1 {
```
DafyddLlyr, line 3: `:global()` is required here as we're trying to style pre-compiled HTML here, without additional scoped classes a CSS module would normally add.

ianjon3s (approved):

PR #5355, opened 2025-10-03
refactor(lps): Use `@tanstack/react-query` for custom `useHTMLDownload()` hook

DafyddLlyr:
## What does this PR do?
 - Replaces React primitive management and boilerplate (`useState()`, `useEffect()`) with `@tanstack/react-query`

## Motivations
- More clear and declarative state management and data fetching
- Good practice / upskilling for integration of `@tanstack/react-query` into the Editor
- Better performance and reliability server-side caching attempted in https://github.com/theopensystemslab/planx-new/pull/5333

## To test...
(On staging) Hit "View application" multiple times for the same application ID - slow re-load, repeated server fetch (and generation!) of the same data

(On Pizza) Hit "View application" multiple times for the same application ID - quick re-load, no external HTTPS calls, data re-served from `@tanstack/react-query` local cache.

### DIFF
--- a/api.planx.uk/modules/lps/controller.ts
+++ b/api.planx.uk/modules/lps/controller.ts
@@ -66,13 +66,7 @@ export const downloadHTMLController: DownloadHTML = async (_req, res, next) => {
   const { sessionId, email } = res.locals.parsedReq.body;
   try {
     const html = await generateHTML(sessionId, email);
-    return (
-      res
-        // Cache for 2hrs - submitted applications can't change so the data will not become stale
-        .setHeader("Cache-Control", "private, max-age=7200")
-        .header("Content-type", "text/html")
-        .send(html)
-    );
+    return res.header("Content-type", "text/html").send(html);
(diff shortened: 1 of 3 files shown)


api.planx.uk/modules/lps/controller.ts
```diff
@@ -66,13 +66,7 @@ export const downloadHTMLController: DownloadHTML = async (_req, res, next) => {
...
   try {
     const html = await generateHTML(sessionId, email);
-    return (
-      res
-        // Cache for 2hrs - submitted applications can't change so the data will not become stale
-        .setHeader("Cache-Control", "private, max-age=7200")
-        .header("Content-type", "text/html")
-        .send(html)
-    );
+    return res.header("Content-type", "text/html").send(html);
```
DafyddLlyr, line 69: Solving this client-side is both safer and more performant, given that we also need to control the token generation process.

jessicamcinchak (approved):
Cool !

PR #5356, opened 2025-10-06
feat: add 'To review' tab to sidebar

augustlindemer:
This PR adds a 'To review' tab to the sidebar that lists all nodes in the current flow tagged 'To review'. It displays them together with any 'Internal notes' on the node.

The tab helps editors to more easily find and navigate to nodes in need of reviewing. Internal notes are displayed because these are currently used by editors to provide context on what is to be reviewed.

The tab does not display on flows that are _templated_.

To be revised during future development of a wider in-editor notification and user tagging system.


DafyddLlyr (changes requested):
Looking great - all working as expected!

A few comments below on how to take this to the next step and get it production ready 🚀

augustlindemer:
Thanks for the detailed comments @DafyddLlyr ! Please let me know if I misunderstood any of the suggestions.

Also left a comment re useMemo dependencies that I'm not sure about!

PR #5366, opened 2025-10-08
fix(1/4): recreate mv `analytics_exits`

jessicamcinchak:
See #5364

Reverting #5359 one at a time to ensure no issues running on prod


jessicamcinchak:
@zz-hh-aa no problem - continues to be really hard to test these before they reach prod based on data persistance! A couple [notes about optimisations](https://github.com/theopensystemslab/planx-new/pull/5364) here if curious - basically nothing wrong with any of your original query logic, just hit a timeout when trying to create view for the first time and I think the swap from `lateral join` to `with` should be a bit more performant!

PR #5437, opened 2025-10-14
feat: Use singleton pattern for ShareDB

DafyddLlyr:
## What does this PR do?
- Implements a singleton pattern for ShareDB - allowing only one open WS connection, and one flow subscription at any one time
- Fixes persistent `.split() is not a function` error on flow navigation
- Creates loading states for flows
- (Mostly) resolves flow centering issues
- Adds loading skeleton for flow

Image

I've co-located most of the context with the code here, so won't repeat..!

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/index.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/index.tsx
@@ -16,7 +16,12 @@ export enum FlowLayout {
   LEFT_RIGHT = "left-right",
 }
 
-const Flow = ({ lockedFlow, showTemplatedNodeStatus }: any) => {
+interface Props {
+  lockedFlow: boolean;
+  showTemplatedNodeStatus: boolean;
+}
+
+const Flow: React.FC<Props> = ({ lockedFlow, showTemplatedNodeStatus }) => {
   const { url } = useCurrentRoute();
   const flowPath = url.pathname.split("/")[2];
   const [_flow, ...breadcrumbIds] = flowPath.split(",");
@@ -29,7 +34,7 @@ const Flow = ({ lockedFlow, showTemplatedNodeStatus }: any) => {
     state.flowLayout,
   ]);
 
-  const breadcrumbs = breadcrumbIds.map((id: any) => ({
+  const breadcrumbs = breadcrumbIds.map((id) => ({
     id,
     ...getNode(id),
     href: `${window.location.pathname.split(id)[0]}${id}`,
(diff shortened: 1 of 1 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/index.tsx
```diff
@@ -16,18 +16,20 @@ export enum FlowLayout {
...
 
-const Flow = ({
-  breadcrumbs = [],
-  lockedFlow,
-  showTemplatedNodeStatus,
-}: any) => {
+const Flow = ({ lockedFlow, showTemplatedNodeStatus }: any) => {
+  const { url } = useCurrentRoute();
+  const flowPath = url.pathname.split("/")[2];
+  const [_flow, ...breadcrumbIds] = flowPath.split(",");
```
DafyddLlyr: This fixes the persistent `Cannot read properties of undefined (reading 'split')` error we see (especially in dev mode).

Here's the solution - 
 - Only render this when a flow is actually loaded
 - Pass responsibility down to the correct component
 - use `useCurrentRoute()` hook instead of `window` call

Image

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -386,6 +389,13 @@ export const editorStore: StateCreator<
     );
   },
 
+  disconnectFromFlow: () => {
+    console.debug("[ShareDB] Disconnecting from flow:", doc?.id);
+    // Clear local store cache of flow data
+    set({ flow: {} });
+    doc.destroy();
+  },
```
DafyddLlyr: **Why are we nullifying a flow to the truthy `{}`?**

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/shared.ts
```diff
@@ -117,14 +115,6 @@ export const sharedStore: StateCreator<
     ).has(id);
   },
 
-  setFlowSlug(flowSlug) {
-    set({ flowSlug });
-  },
-
-  setFlowName(flowName) {
-    set({ flowName });
-  },
```
DafyddLlyr: I've dropped these are they were only being used in one place and don't offer much more than a plain `set()` does. Eventually, these values can just be held in the Apollo cache (or route data) as opposed to within the store.

apps/editor.planx.uk/src/routes/views/flowEditor.tsx
```diff
@@ -139,10 +139,6 @@ export const flowEditorView = async (req: NaviRequest) => {
     template,
   });
 
-  // Templated flows require access to an ordered flow
-  // Set this once upstream as it's an expensive operation
-  if (templatedFrom) useStore.getState().setOrderedFlow();
```
DafyddLlyr: Having this upstream as part of the route/view/layout means we don't have loaded flow data from ShareDB. I've now moved this to run once the flow data has been fetched.

PR #5454, opened 2025-10-15
chore: remove users from Templates team

jessicamcinchak:
Following templates migration and the pre-existing "make copiable by others" feature, users no longer need access to the Templates team.

**Changes:**
- Deletes all current team memberships to "Templates"
- Drops database trigger which automatically adds all new users to "Templates" directly (PR #2291)
- Drops corresponding E2E API tests

**To test:** 
- Query `team_members` table and ensure no rows where `team_id = 29`
- Add a new user to any team and ensure they are NOT automatically added to templates (eg above query still true)

**Next steps:**
- The Templates team is still going to stick around for awhile while content team archives and migrates flow content
- Once the actual team is ready to be deleted, they'll be further code remnants we can similarly clean up :broom:

### DIFF
--- a/apps/hasura.planx.uk/migrations/default/1760563920793_remove_templates_team_members/up.sql
+++ b/apps/hasura.planx.uk/migrations/default/1760563920793_remove_templates_team_members/up.sql
@@ -1,6 +1,6 @@
-DELETE FROM "public"."team_members"
+DELETE FROM team_members
 WHERE team_id = 29;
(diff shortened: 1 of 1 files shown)


DafyddLlyr:
Possibly unrelated, but I'm having issues on a few other Pizzas - maybe an issue with Vultr somewhere? Nothing on their status page thought 🤔

PR #5455, opened 2025-10-16
feat: add cut and paste (move) operation to context menu

augustlindemer:
This PR adds 'Cut' as a third operation to the context menu.

Cut stores a `parent` and `rootId` in `cutNode`.

When selecting Paste with a `cutNode` in local storage, it executes a move operation on the respective node. This was chosen over copy & delete in order to maintain the ids of the so-moved nodes.

Similar to clone, Paste is only enabled if the Cut node is in the current flow.

Cut uses MUI's [ContentCut](https://mui.com/material-ui/material-icons/?query=cut&selected=ContentCut) icon, open to alternatives @ianjon3s!

This PR also enforces that only one value (`copiedNode`, `clonedNodeId`, or `cutNode`) gets stored in local storage at any point in time.

Note: Went through a dozen iterations on how to best handle this, please feel free to suggest a completely different approach to this one!

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
@@ -30,7 +30,6 @@ import {
   ContextMenuPosition,
   ContextMenuSource,
 } from "pages/FlowEditor/components/Flow/components/ContextMenu";
-import { NewFlow } from "pages/Team/components/AddFlow/types";
 import type { StateCreator } from "zustand";
 import { persist } from "zustand/middleware";
 
@@ -216,6 +215,8 @@ interface CopiedPayload {
 interface CutPayload {
   rootId: string;
   parent: string;
+}
+
 export interface Template {
   id: string;
   team: {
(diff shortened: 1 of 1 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -695,6 +742,27 @@ export const editorStore: StateCreator<
...
   },
 
+  pasteCutNode(toParent, toBefore) {
+    const cutString = localStorage.getItem("cutNode");
+    if (!cutString) return;
+
+    try {
+      const { rootId, parent }: CutPayload = JSON.parse(cutString);
+      if (rootId === toBefore && parent === toParent) {
+        throw new Error("Cannot move before itself");
```
augustlindemer: Band-aid because the existing error 'toBefore does not exist in toParent' is cryptic ([see here](https://github.com/theopensystemslab/planx-new/blob/af68095d8a6d24dde84b472c06da8d2f6c38ef88/apps/editor.planx.uk/src/%40planx/graph/index.ts#L317)).

Unfortunate this is inconsistent. A node can't move before itself, but it can move after itself. Ideally we want consistency between the two (either disallow both, or allow both as full operations). 

Ideas?

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -695,6 +742,27 @@ export const editorStore: StateCreator<
     }
   },
 
+  pasteCutNode(toParent, toBefore) {
```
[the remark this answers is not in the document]
DafyddLlyr: This is a much more flexible pattern - agreed!

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -695,6 +742,27 @@ export const editorStore: StateCreator<
...
+      const [, ops] = move(rootId, parent, {
+        toParent,
+        toBefore,
+      })(get().flow);
+      send(ops);
+    } catch (err) {
+      alert((err as Error).message);
+    } finally {
+      localStorage.removeItem("cutNode");
+    }
```
DafyddLlyr: I think this probably works for us, but generally users can cut + paste, paste, paste multiple times in other contexts.

I'm happy to keep this as-is and wait for feedback 👍

augustlindemer: I went with Windows' single paste pattern for files here, but agreed that if we get respective feedback we can instead of clearing cutNode create a new cut reference to the (then) new parent upon paste, which allows for additional subsequent moves.

PR #5459, opened 2025-10-16
feat: Improved list view for team page

ianjon3s:
## What does this PR do?

Introduces an improved list view for flows on team pages, based on latest editor designs.

- Dedicated table component for list view
- Uses updated menu for `FlowCard`, unified into `FlowMenu` for use across both card and table
- Creates `useFlowMetadata` hook for unified data loading
- Creates `useFlowDates` to simplify showing of last edited/published dates
- Creates unified `FlowTemplateIndicator` to reduce repetition in components
- Creates `TruncatedText` utility component to responsively trim text with CSS

### DIFF
--- a/apps/editor.planx.uk/src/pages/Team/components/FlowCard/index.tsx
+++ b/apps/editor.planx.uk/src/pages/Team/components/FlowCard/index.tsx
@@ -70,7 +70,6 @@ const FlowCard: React.FC<Props> = ({ flow, refreshFlows }) => {
               isSourceTemplate={isSourceTemplate}
               isTemplatedFlow={isTemplatedFlow}
               teamName={flow.template?.team.name}
-              variant="banner"
             />
           </CardBanner>
         )}
--- a/apps/editor.planx.uk/src/pages/Team/components/FlowTable/index.tsx
+++ b/apps/editor.planx.uk/src/pages/Team/components/FlowTable/index.tsx
@@ -103,7 +103,6 @@ const FlowTableRow: React.FC<FlowTableRowProps> = ({
                 isSourceTemplate={isSourceTemplate}
                 isTemplatedFlow={isTemplatedFlow}
                 teamName={flow.template?.team.name}
-                variant="inline"
               />
             </Box>
           )}


apps/editor.planx.uk/src/pages/Team/components/FlowDialogs.tsx
DafyddLlyr: We've both ended up refactoring this component this week...! 🚌 🚌 https://github.com/theopensystemslab/planx-new/pull/5466

Have a look at that PR and let me know what you think in terms of merge order. 

The approach I've taken is to group the menu and dialogs together as they always need each other, as well as just using a single state variable to track the open modal.

apps/editor.planx.uk/src/pages/Team/components/hooks/useFlowActions.ts
DafyddLlyr: See previous comment - these actions are intrinsically linked to the menu and dialogs, and never get called outside this context (I believe!). A hook is a great way of wrapping up these actions.

I'm not certain that the computed values really line up here though as a flow "action".

What I think this highlights is that we need to know certain information about a flow to display a grid or table (such as `isSubmissionService`, `isSourceTemplate` etc). A few options that might be better than sharing here via a hook - 
 - Transforming flow data once we've fetched it
 - Util functions which can return computed values
 - Another, simpler, hook for the computed values
 - Just repeating! It's only in two places so really not that bad

What I'm aiming to avoid is somebody needing to know it an totally different context if a flow is a template, reaching for this hook, and importing a bunch of dialog actions etc which aren't needed and couldn't run.

apps/editor.planx.uk/src/pages/Team/components/FlowTable.tsx
```diff
@@ -0,0 +1,289 @@
...
+        </TableRow>
+      </StyledTableHead>
+      <TableBody>
+        {flows.map((flow) => (
+          <FlowTableRow
+            key={flow.slug}
+            flow={flow}
+            teamSlug={teamSlug}
+            refreshFlows={refreshFlows}
+            showPublished={showPublished}
```
DafyddLlyr: nit: Rather than prop drill these values in, we can just call `useFlowSortDisplay()` within the child.

apps/editor.planx.uk/src/pages/Team/components/FlowTable.tsx
```diff
@@ -0,0 +1,289 @@
...
+import { useFlowSortDisplay } from "./hooks/useFlowSortDisplay";
+
+const StyledTable = styled(Table)(({ theme }) => ({
+  marginTop: theme.spacing(2),
+  zIndex: 1,
+  position: "relative",
+  border: `1px solid ${theme.palette.border.main}`,
+  borderBottom: 0,
+  borderCollapse: "separate",
+  "& .MuiTableCell-root": {
```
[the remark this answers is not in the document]
ianjon3s: `tableCellClasses` employed 👍

apps/editor.planx.uk/src/pages/Team/components/hooks/useFlowMetadata.ts
DafyddLlyr: A much better solution. All this is encapsulated in one single place now 👍

DafyddLlyr (approved):
Big improvements overall - great!

One minor issue is that there's a flicker as the formatted date is not re-rendered on the same cycle as the card title. Not a showstopper, but might we worth trying to timebox a solution for - 

https://github.com/user-attachments/assets/e677bcb6-80df-4daa-87eb-fb12f9bdce24

PR #5466, opened 2025-10-19
refactor: Move flow

DafyddLlyr:
This one ended up a bit more involved than expected! 😅 

 - Splits up `FlowCard`
 - Refactors "move" operation to `@tanstack/react-query`
 - Creates modal for "move" instead of just dialogs and alerts

image

### DIFF
--- a/apps/editor.planx.uk/src/api/client.ts
+++ b/apps/editor.planx.uk/src/api/client.ts
@@ -20,7 +20,7 @@ apiClient.interceptors.request.use(
     }
     return config;
   },
-  (error) => Promise.reject(error)
+  (error) => Promise.reject(error),
 );
 
 /**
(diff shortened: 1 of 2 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -533,6 +532,9 @@ export const editorStore: StateCreator<
       variables: {
         teamId,
       },
+      // Flows are modified via REST API requests, not via the Apollo client
+      // Always get an up to date list when showing the page
+      fetchPolicy: "network-only",
```
DafyddLlyr: There are many better ways of handling this, such as manual invalidations on these operations.

Parked for now as this PR is already too big!

jessicamcinchak (approved):

PR #5469, opened 2025-10-20
refactor: Use `@tanstack/react-query` for "Save & Return" endpoints

DafyddLlyr:
## What does this PR do?
 - Moves API requests to the `/api` layer
 - Uses `useMutation()` throughout instead of direct calls to `axios` - no more manual `Status` tracking
 - Splits up `ResumePage` into `ValidateSession` and `SendResumeEmail`

## Context
Part of implementing https://github.com/theopensystemslab/planx-new/blob/main/doc/architecture/decisions/0011-standardised-data-fetching-and-state-management.md which is being tracked here - https://github.com/orgs/theopensystemslab/projects/7/views/1


apps/editor.planx.uk/src/pages/Preview/Resume/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Component which handles the "Resume" page used for Save & Return
+ * The user can access this page via four "paths"
+ * 1. Directly via PlanX, user enters email to trigger "dashboard" email with resume magic links
+ * 2. Magic link in email with a sessionId, user enters email to continue application
+ * 3. Redirect back from GovPay - sessionId and email come from query params
+ * 4. Redirect from localplanning.services - sessionId and email come from query params
+ */
+const ResumePage: React.FC = () => {
```
DafyddLlyr: **Why split this up?**

Whilst re-working on this component it was clear that by now it was doing too much. The fact that there were two separate mutations, with associated statuses, felt like a clear indication that this could be made simpler and clearer by breaking these out.

jessicamcinchak (approved):
These continue to be very satisfying, thanks for working through them !

PR #5502, opened 2025-10-22
chore: display `PlanningConstraints` and `PropertyInformation` summaries on `Review` page

jessicamcinchak:
These components do _not_ currently have a Review page representation, which also means users don't have an easy way to navigate "back" to "change" them from the Review step.

This adds a simple representation for each with "change" navigation.


jessicamcinchak:
Closing in favor of #5636 - lots of messy linter changes on rebase to `main` here !

PR #5503, opened 2025-10-22
docs: Update SSL docs

DafyddLlyr:
## What does this PR do?
 - Updates renewal and creation docs for subdomains
 - Updates script after pairing with @zz-hh-aa to handle upserts


infrastructure/application/index.ts
```diff
@@ -658,7 +669,7 @@ export = async () => {
     function createCustomDomain({
       domain,
       name,
-      certificateLocation = "pulumiConfig"
+      certificateLocation = "secretsManager"
```
DafyddLlyr: When updating the docs it seemed to make more sense to consider this as the default going forward. It's easier as a step to remove from Pulumi config, and update the `secretLocation` here also.

PR #5508, opened 2025-10-22
feat: switch to debian postgis image and install pg_cron

zz-hh-aa:
Switches postgis container to debian, not alpine, and installs `pg_cron` (so that we can use cron jobs to update the materialized views created in #5366 #5376 and #5384)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ services:
     restart: unless-stopped
     ports:
       - "${PG_PORT}:5432"
-    command: "postgres -c shared_preload_libraries=pg_cron -c cron.database_name=${PG_DATABASE}"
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -1,11 +1,13 @@
...
   postgres:
     container_name: postgres
-    image: postgis/postgis:16-3.5-alpine
+    build:
+      context: ./apps/postgres
+      dockerfile: Dockerfile
     restart: unless-stopped
     ports:
       - "${PG_PORT}:5432"
-    command: "postgres -c listen_addresses='*'"
```
freemvmt: we don't need the `listen addresses ` config anymore?

freemvmt (approved):
🚀  Looks good to me! Only one query :)

PR #5525, opened 2025-10-26
feat: Use apiClient for file upload endpoints

DafyddLlyr:
## What does this PR do?
 - Updates private and public file uploads to use shared `apiClient`
 - Does not implement `@tanstack/react-query` as we're already handling requests via `useDropzone()`
 - Updates mocks and tests
   - A few other migration PRs are failing due to this Axios mocking issue, so this PR will need to be merged first.

### DIFF
--- a/apps/editor.planx.uk/src/api/fileUpload/__mocks__/requests.ts
+++ b/apps/editor.planx.uk/src/api/fileUpload/__mocks__/requests.ts
@@ -1,0 +1,33 @@
+import { vi } from "vitest";
+
+/**
+ * Shared mock for default file upload behaviour
+ * Recreating this with MSW is challenging due to the onProgress event used by Axios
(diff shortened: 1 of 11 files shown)


apps/editor.planx.uk/src/api/fileUpload/__mocks__/requests.ts
DafyddLlyr: I spent a good while trying to get `msw` to work and trigger the `onProgress` event without much luck.

I'm sure it's possible via [@mswjs/interceptors](https://github.com/mswjs/interceptors) ([docs](https://mswjs.io/docs/recipes/xmlhttprequest-progress-events)), but I couldn't quite get it working.

A single shared mock here though works well and cuts down on a lot of the noise across multiple test files.

PR #5529, opened 2025-10-27
fix(lps): Improve page titles for home and search page

DafyddLlyr:
Currently, both the home and search pages have titles of `"Find local planning services | Find local planning services"`

image

image

### DIFF
--- a/apps/localplanning.services/src/pages/index.astro
+++ b/apps/localplanning.services/src/pages/index.astro
@@ -11,7 +11,7 @@ import lpaMap from "/public/planx-lpa-map.svg";
 const title = "Find local planning services";
 ---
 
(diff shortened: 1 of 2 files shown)


apps/localplanning.services/src/pages/index.astro
```diff
@@ -11,7 +11,7 @@ import lpaMap from "/public/planx-lpa-map.svg";
 const title = "Find local planning services";
 ---
 
-<Layout title={title}>
+<Layout>
```
DafyddLlyr: Alternatively...

```suggestion
<Layout title="Home">
```

jessicamcinchak (approved):

PR #5533, opened 2025-10-27
fix: type error when publishing source templates with children

jessicamcinchak:
See https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1761583046898099

### DIFF
--- a/apps/api.planx.uk/modules/flows/publish/controller.ts
+++ b/apps/api.planx.uk/modules/flows/publish/controller.ts
@@ -17,10 +17,7 @@ export const publishFlowSchema = z.object({
   }),
   query: z.object({
     summary: z.string(),
-    templatedFlowIds: z
-      .string()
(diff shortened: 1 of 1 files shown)


DafyddLlyr (commented):
I guess this regression was introduced here sorry! https://github.com/theopensystemslab/planx-new/pull/5397/files#diff-04d586bae3e108b34e96575176791478d4ecfef5ea942012c48e52aa37a64c21R15-R31

I also recently hit this with PlanningData needing params as `dataset=first&dataset=second` etc. The solution was the paramSerialiser - 

https://github.com/theopensystemslab/planx-new/pull/5524/files#diff-246e843173044a862f7d315457253a78aadf9eb364b2f67867c68731b2211115R22-R24

Another solution avoiding this odd param behaviour might be to make summary and templateFlowIds part of the body of the request.

PR #5536, opened 2025-10-28
chore: Replace hasura-cli

DafyddLlyr:
Replaces https://www.npmjs.com/package/hasura-cli with https://www.npmjs.com/package/@kirillvakalov/hasura-cli

This is a pretty new package in terms of adoption, but the current `hasura-cli` is seemingly abandoned. I reckon we give it a try for a while and see how we get on.

Just as a sense check I'd appreciate if another dev could please - 
 - Pull branch `dp/alternate-hasura-cli` locally
 - `cd apps/hasura.plan.uk && pnpm i`
 - `pnpm start`
 - See Hasura start locally on port 9695 without issue, still creates migration scripts if you make a random local change

Moving away from `hasura-cli` will fix - 
 - https://github.com/theopensystemslab/planx-new/security/dependabot/291
 - https://github.com/theopensystemslab/planx-new/security/dependabot/292
 - https://github.com/theopensystemslab/planx-new/security/dependabot/293

### DIFF
--- a/apps/hasura.planx.uk/package.json
+++ b/apps/hasura.planx.uk/package.json
@@ -1,6 +1,6 @@
 {
   "devDependencies": {
-    "hasura-cli": "^2.36.1"
+    "@kirillvakalov/hasura-cli": "^2.48.1"
   },
   "scripts": {
     "start": "hasura console --envfile ./../../.env"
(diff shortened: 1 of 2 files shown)


zz-hh-aa (approved):
Thanks for the clear testing notes! Can confirm after checking out branch, installing updated dependencies in `apps/hasura.planx.uk`, Hasura console starts and up / down migrations generate as expected when a random change is made.

PR #5538, opened 2025-10-28
feat: Style data fields within FileUploadAndLabel rules

DafyddLlyr:
## What does this PR do?
 - Converts plain inputs to `<DataFieldAutocomplete/>` fields for file upload rules which use data values
 - Fixes the background color to be consistent across all data fields
 - Fixes spacing for uploaded images next to the file's `moreInformation` field
 - Test fixes for the `ProgressEvent` polyfill

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/shared/DataFieldAutocomplete.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/shared/DataFieldAutocomplete.tsx
@@ -70,7 +70,6 @@ export const DataFieldAutocomplete: React.FC<Props> = (props) => {
     <InputRow>
       <ErrorWrapper error={props.errorMessage}>
         <AutocompleteInput
-          sx={{ background: "#f0f0f0" }}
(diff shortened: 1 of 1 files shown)


apps/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Editor.tsx
```diff
@@ -191,32 +193,30 @@ function FileTypeEditor(props: ListManagerEditorProps<FileType>) {
...
             }
-            placeholder="Data field"
             disabled={props.disabled}
           />
           <Operator>Equals</Operator>
-          <Input
+          <DataFieldAutocomplete
             required
-            name="val"
+            schema={["recommended", "required"]}
```
DafyddLlyr: @jessicamcinchak Really appreciate this bit of design 👍 Very nice to be able to suggest any values, but not constrain the user!

PR #5541, opened 2025-10-28
fix: Configure `DOCKER_DEFAULT_PLATFORM` for Apple silicone chips (arm64)

DafyddLlyr:
## What does this PR do?
- Simplified docker process for devs using Apple silicone (ARM64)
  - Just set `DOCKER_DEFAULT_PLATFORM` once in root `.env` (see `.env.example`)
  - No additional docker-compose file required, no additional scripts
  - Specifies in `docker-compose.yaml` which images require `linux/amd64` emulation, this will overwrite `DOCKER_DEFAULT_PLATFORM` allowing all containers to build without error
 - Uses `postgis/postgis:16-3.5` instead of `postgis/postgis:16-3.5-alpine` across the repo for consistency

## Testing
I had `DOCKER_DEFAULT_PLATFORM` set locally as an env var which has allowed me to avoid most of the complexity here so far. I've now unset this, destroyed all docker images locally, and re-ran `pnpm start` multiple times without issue to test this process.

In order to progress this I'd like at least one other dev using an arm64 chip (Apple) and one using anything else to please try the following -

 - Stop all local docker containers - `docker stop $(docker ps -q)`
 - Start from a totally clean slate - `docker system prune -a --volumes`
 - Run `pnpm start` in the repo - all containers and service start as expected ✅ 
 - Run `pnpm sync-data` in the repo - data synced without issue ✅


docker-compose.yml
```diff
@@ -1,5 +1,6 @@
 services:
   postgres:
+    platform: linux/amd64
```
DafyddLlyr: This image doesn't currently support ARM64 - https://github.com/postgis/docker-postgis/issues/216

Though the container seemed to be spinning up without issue when using `DOCKER_DEFAULT_PLATFORM=linux/arm64`, it was leading to Hasura migrations hanging and ultimately failing.

By specifying `linux/amd64` here, we can force emulation and a healthy build.

PR #5546, opened 2025-10-29
feat: set up cron jobs to refresh mvs

zz-hh-aa:
Sets up cron jobs to refresh materialized views nightly around 3am (spaced out by 5 min). 

I'm wondering how we'd check that these are set up correctly afterwards--checking AWS console?


DafyddLlyr:
> _I'm wondering how we'd check that these are set up correctly afterwards--checking AWS console?_

When working on cron jobs locally I've set the time to 1min, so they trigger frequently and then checked via the methods listed here - https://github.com/citusdata/pg_cron?tab=readme-ov-file#viewing-job-run-details

```sql
select * from cron.job_run_details order by start_time desc limit 5;
```

However, I noticed yesterday that there's a cron schema - maybe we should just get Hasura to track this? Seems like this would be a nice solution in terms of discoverability and observability. I've not tried this out yet, but please have a go and if you think it's helpful please open a PR 😄 

image

It should just be a case of hitting "track" on these two tables when Hasura is in dev mode locally (running on port 9695 listening for migrations).

PR #5548, opened 2025-10-29
feat(lps): Improve application card status styling

ianjon3s:
## What does this PR do?

- Refactors LPS application cards so that the progress bar is no longer the prominent visual marker of status
- Adds a new component "StatusBadge" that displays the three statuses (draft, awaiting payment, submitted)

image

### DIFF
--- a/apps/localplanning.services/src/components/applications/ApplicationCard.tsx
+++ b/apps/localplanning.services/src/components/applications/ApplicationCard.tsx
@@ -124,35 +124,22 @@ const ActionButtons: React.FC<Application> = (application) => {
   );
 };
 
-export const ApplicationCard: React.FC<Application> = (application) => {
-  const cardBgClass = (() => {
-    switch (application.status) {
-      case "awaitingPayment":
-        return "bg-red-100/30";
-      case "submitted":
-        return "bg-green-100/30";
-      default:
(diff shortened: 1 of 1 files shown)


apps/localplanning.services/src/components/applications/ApplicationCard.tsx
```diff
@@ -3,84 +3,22 @@ import type { Application } from "./hooks/useFetchApplications";
...
-  const progressColour = (() => {
-    switch (application.status) {
+        return formatDate(application.updatedAt);
       case "submitted":
-        return "bg-green-600";
-      case "awaitingPayment":
-        return "bg-red-900";
-      default:
-        return "bg-black";
+        return formatDate(application.submittedAt!);
```
DafyddLlyr:
```suggestion
        return formatDate(application.submittedAt);
```

Type narrowing via the `switch/case` should make this a `SubmittedApplication` so the `!` shouldn't be needed.

Were you hitting type issues here?

DafyddLlyr (approved):
Great solution! Shame to lose the progress bar, but I totally agree that this is a better general solution and visually much more impactful.

I'll open a PR up based on this which removes the `progress` property from the LPS model and query - no need to fetch this data if we're not using it 👍

PR #5555, opened 2025-10-31
feat: Extract `<RuleBuilder/>` from FileUploadAndLabel

DafyddLlyr:
## What does this PR do?
 - Extracts UI for FileUploadAndLabel rules to a new `<RuleBuilder/>` component
 - Extracts types and utils for the rule builder
 - Implements the `<RuleBuilder/>` within FileUploadAndLabel

## Motivation
I'm (finally!) starting to turn my attention to the `ResponsiveQuestion` and `ResponsiveChecklist` components. The first step was [to improve types](https://github.com/theopensystemslab/planx-new/pull/5553), and next up I'm aiming to share the current logic for building `Rules`. When we initially opened this ticket I thought that there wasn't enough here that was shareable but I think we could use this (with a limited set of `conditions`) to construct our rules.

Currently, these apply to `FileType`s, the intention is to apply them to `Options` for the new components.

I'll open a Slack thread on this, but I'm interested in revisiting the enhancements vs new component decision again, and going back over what lead us in this direction. It seems that wiring this into the current component might be both simpler than thought (with these two refactors in place), and a better Editor experience (I can image "convert this Checklist a ResponsiveChecklist"  being a pretty frequent task).

### DIFF
--- a/apps/editor.planx.uk/src/ui/editor/SelectInput/SelectInput.tsx
+++ b/apps/editor.planx.uk/src/ui/editor/SelectInput/SelectInput.tsx
@@ -41,11 +41,14 @@ const Root = styled(Select)(({ theme }) => ({
     alignItems: "center",
     justifyContent: "flex-start",
   },
-  [`&.${classes.icon}`]: {
-    right: theme.spacing(1),
-    color: "rgba(0,0,0,0.25)",
+  // Match caret in MultipleSelect component
+  [`& .${classes.icon}`]: {
+    right: theme.spacing(0.85),
+    bottom: theme.spacing(0.5),
+    color: theme.palette.primary.main,
+    fontSize: 35,
   },
-  [`&.${classes.menuPaper}`]: {
+  [`& .${classes.menuPaper}`]: {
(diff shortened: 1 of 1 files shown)


apps/editor.planx.uk/src/ui/editor/SelectInput/SelectInput.tsx
```diff
@@ -41,11 +41,14 @@ const Root = styled(Select)(({ theme }) => ({
...
   },
-  [`&.${classes.icon}`]: {
-    right: theme.spacing(1),
-    color: "rgba(0,0,0,0.25)",
+  // Match caret in MultipleSelect component
+  [`& .${classes.icon}`]: {
+    right: theme.spacing(0.85),
+    bottom: theme.spacing(0.5),
+    color: theme.palette.primary.main,
+    fontSize: 35,
```
DafyddLlyr: Noticed a small visual issue with the icon here when working on integrating this in Questions / Checklists - 

**Before**
Image

**After**

Image

PR #5579, opened 2025-11-01
feat(editor): Add optional rules to `Question` component

DafyddLlyr:
## What does this PR do?
Adds optional rules to the the `Question` component. Editors can set rules, and then the public interface will apply these at runtime based on the user's responses so far.

Originally, we had planned there to be separate `ResponsiveQuestion` and `ResponsiveChecklist` components, but I'm hoping that this PR lets us validate that assumption (either way!) - is it actually simpler to have this integrated into the current components? I think there's some valid UI questions around how this appears within the modal to be discussed, but I'm most interested in working out if we need two additional components before getting much further here.

### Changes
 - Model: adds optional `rule` property to `Option`
 - Model: simplifies the `Response` type
 - UI: Adds `<RuleBuilder/>` to the `<BaseOptionsEditor/>` component
   - This allows us to add rules to both the `Checklist` and `Question` model, but this is just implemented within `Question` currently

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/shared/RuleBuilder/utils.ts
+++ b/apps/editor.planx.uk/src/@planx/components/shared/RuleBuilder/utils.ts
@@ -1,9 +1,9 @@
 import { Store } from "pages/FlowEditor/lib/store";
 
-import { Condition, Operator, Rule } from "./types";
+import { Condition, Rule } from "./types";
 
 export const isRuleMet = (passport: Store.Passport, rule: Rule): boolean => {
-  // TODO: Tests
+  // Simple (non-conditional) rules always evaluate to true
   if (rule.condition === Condition.AlwaysRequired) return true;
   if (rule.condition === Condition.AlwaysRecommended) return true;
   if (rule.condition === Condition.NotRequired) return true;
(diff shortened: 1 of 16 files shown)


apps/editor.planx.uk/src/@planx/components/shared/RuleBuilder/utils.ts
```diff
@@ -1,11 +1,13 @@
 import { Store } from "pages/FlowEditor/lib/store";
 
-import { Condition, ConditionalRule, Operator, Rule } from "./types";
+import { Condition, Rule } from "./types";
+
+export const isRuleMet = (passport: Store.Passport, rule: Rule): boolean => {
+  // Simple (non-conditional) rules always evaluate to true
+  if (rule.condition === Condition.AlwaysRequired) return true;
+  if (rule.condition === Condition.AlwaysRecommended) return true;
+  if (rule.condition === Condition.NotRequired) return true;
```
DafyddLlyr: Slightly verbose, but checking each item allows TS to correctly type-narrow. Using something like `simpleConditions.includes(rule.condition)` won't work here to type-narrow.

PR #5589, opened 2025-11-04
refactor: Simplify select input

ianjon3s:
## What does this PR do?

- Moves `SelectInput` from `ui/editor` to `ui/shared/` directory as it's now being used in the public interface for Map + label
- Removes redundant styles from `SelectInput` (paper styles are handled in the theme)
- Creates a `size` prop to be used for smaller selects, namely flow filters and Map + label copy feature. This also reduces the need for inline styles

Before (size: **default):**
image

**After (size: small):**
image


ianjon3s:
No worries @DafyddLlyr, it was a good chance to pull a bunch of inputs inline

PR #5608, opened 2025-11-05
fix: Display grouped options in the editor modal

DafyddLlyr:
Currently, grouped options do not display in the editor due to a regression introduced recently as part of the `Question` and `ResponsiveQuestion` refactors.

We should get this to prod tomorrow morning, after the data sanitation has completed tonight.

I'll get test coverage up shortly!

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/Checklist/Editor/Editor.test.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/Checklist/Editor/Editor.test.tsx
@@ -10,7 +10,7 @@ import { HTML5Backend } from "react-dnd-html5-backend";
 import { setup } from "testUtils";
 import { vi } from "vitest";
 
-import { Checklist } from "../model";
+import { Checklist, Group } from "../model";
 import { ChecklistEditor } from "./Editor";
 
 const { getState } = useStore;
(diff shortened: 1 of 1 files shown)


apps/editor.planx.uk/src/@planx/components/Checklist/Editor/Editor.tsx
```diff
@@ -34,7 +34,11 @@ export const ChecklistEditor: React.FC<Props> = (props) => {
   const type = TYPES.Checklist;
 
   const formik = useFormik<ChecklistWithOptions>({
-    initialValues: parseChecklist({ ...props.node?.data, options: props?.options }),
+    initialValues: parseChecklist({
+      ...props.node?.data,
+      options: props?.options,
+      groupedOptions: props.groupedOptions,
```
jessicamcinchak: nit: I think this needs to be `props?.groupedOptions`? Or is it actually always guaranteed to be defined now (eg `[]` if "flat" options without groups rather than `undefined`)?

DafyddLlyr: Good spot - fixed on the rebase to make it clearer.

PR #5613, opened 2025-11-06
[api] handle session data aberrations gracefully during sanitisation

freemvmt:
After merging #5605 and letting the `sanitise_application_data` cronjob run over night, we saw that the `deleteApplicationFiles` operation failed, reporting `Error: Cannot read properties of undefined (reading 'data')`.

Because the same records were sanitised of their `data` field directly afterwards (by `sanitiseLowcalSessions`), we cannot see the malformed session data that prompted this error, but with this PR we introduce a guard to ensure that such an aberration doesn't cause the whole operation to fail.


DafyddLlyr (approved):
Brill - thanks for the extra coverage and debug logs here.

PR #5622, opened 2025-11-07
chore: populate flow_integrations table

zz-hh-aa:
We [reverted](https://github.com/theopensystemslab/planx-new/pull/5618) #5600, which successfully ran two migrations (setting `submission_integrations.email_id` as `gen_random_uuid()`, and then populating `submission_integrations`) before failing on the third (populating `flow_integrations`). 

This should now run successfully because the migration uses `SELECT DISTINCT` instead of just `SELECT`. 

❓  Is it fine to just run this when we don't have a history of the previous migrations in `apps/hasura.planx.uk/migrations` (because we reverted the PR instead of running the `down` migrations)? Is there a better way to handle this?

### DIFF
--- a/apps/hasura.planx.uk/migrations/default/1762504347289_populate_flow_integrations/down.sql
+++ b/apps/hasura.planx.uk/migrations/default/1762504347289_populate_flow_integrations/down.sql
@@ -1,0 +1,6 @@
+DELETE FROM flow_integrations
+WHERE flow_id IN (
+    SELECT pf.flow_id 
+    FROM published_flows pf
+    WHERE pf.has_send_component = true
+);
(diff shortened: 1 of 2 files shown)


apps/hasura.planx.uk/metadata/databases/default/tables/public_lowcal_sessions.yaml
```diff
@@ -158,7 +158,7 @@ event_triggers:
       method: POST
       query_params: {}
       template_engine: Kriti
-      url: "{{$base_url}}/send-email/confirmation"
+      url: '{{$base_url}}/send-email/confirmation'
```
zz-hh-aa: Linter caught these again, originally merged in previous PR

PR #5631, opened 2025-11-07
fix: skip validation when generating HTML for discretionary services

jmcbroom:
- check the application type and skip validation conditionally
- bump planx-core commit hash


jessicamcinchak (approved):
Similar to yesterday, let's merge to staging without Vultr then retest some of the same links ! Thanks for fixes here !

PR #5635, opened 2025-11-07
fix(build): Bump planx-core

DafyddLlyr:


jessicamcinchak (approved):
Thanks for :pizza: fixes here !

PR #5646, opened 2025-11-11
chore: update gitignore [skip pizza]

zz-hh-aa:
To ensure that any keys created in root are also ignored. 

I've tested by creating various files (`*.key`, `*.csr` etc.) in root and other directories too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,19 +30,19 @@ __pycache__
 .ruff_cache/
 
 # Ignore certificate files
-**/*.chain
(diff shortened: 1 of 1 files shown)


zz-hh-aa:
Oops thanks @DafyddLlyr, that's done in [4faccc7](https://github.com/theopensystemslab/planx-new/pull/5646/commits/4faccc7a335a3ea702ac4171e75d97ad7d8e5aff)

PR #5649, opened 2025-11-11
chore(lps): Add 'Check if...' to the guidance service allow list

DafyddLlyr:
This will allow https://editor.planx.dev/gloucester/check-if-you-need-planning-permission to be listed.

The allow list method is pretty tedious / brittle, I think we probably need to revisit this one!


ianjon3s (approved):
Agreed, `check` vs `find-out` is a common distinction and somewhat maintainable, however `...-energy-efficiency` is a complete outlier and we absolutely need a better solution to include these

PR #5659, opened 2025-11-12
feat: Update `SettingsForm` pattern

DafyddLlyr:
## What does this PR do?
This PR proposes a new pattern for out settings forms within PlanX - making this Apollo first, without relying on planx-core data fetching methods and the Zustand `SettingsStore` as a persistent layer.

### Timing / motivation
It's not ideal to be looking at another refactor here, but the aim here is to align this with the redesign of the settings pages and the ongoing work to improve data-fetching. As we're going from one pattern to another, I'm hoping this is a fairly smooth and past process. The two example component here were very easy - it's mostly a copy/paste exercise.

### Server state in both Apollo cache and Zustand store
Settings data lives in Zustand even though it's server state, not UI state. This means we're duplicating Apollo's cache and manually synchronising between them. We've seen this manifest as having to [manually invalidate caches on navigation](https://github.com/theopensystemslab/planx-new/blob/main/apps/editor.planx.uk/src/routes/views/draft.tsx#L41-L43) or manage [complex optimistic update patterns](https://github.com/theopensystemslab/planx-new/blob/main/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/lib/optimisticallyUpdateMembersTable.tsx#L22-L34).

Further reading - [[ADR]: Standardised data fetching and state management](https://github.com/theopensystemslab/planx-new/blob/90297a934dbd39f9b5659873d64cd538f778d0c6/doc/architecture/decisions/0011-standardised-data-fetching-and-state-management.md)

### Excessive boilerplate
Every form recreates the same ~100 lines of code, split across the current [`SettingsForm`](https://github.com/theopensystemslab/planx-new/blob/ebdf4a59d474b148ec8a0ef9f2310b0f137cfe94/apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/shared/SettingsForm.tsx) and each implementation.
- Formik initialisation
- Query/mutation handling via planx-core
- Loading states
- Submission logic
- Toast notifications
- Save/Reset button state

### Inconsistent patterns
Some forms use store methods, some fetch directly, some do both. It's not clear where data lives or how to fetch, update, or invalidate it.

## Solution
This PR creates the `SettingsFormContainer` component - very much inspired by the existing `SettingsForm`, but with more responsibilities. This is a generic wrapper that handles all the form responsibilities (data fetching, mutations, form state, basic UI structure, notifications).

Individual forms can then just provide
- Their specific GraphQL query/mutation
- Data transformation logic
- Validation schema
- Internal layout

## Key changes
**Component-level data fetching**
Instead of fetching all settings data at the route level, each component can now fetch its own data

**Apollo cache as single source of truth**
No more manual synchronisation between store and Apollo. Mutations automatically update the cache, forms re-render with fresh data.

## Next steps...
Roughly here's where I hope we go with this next - 

**Step 1 (this PR)** - Introduce the pattern with Contact and Visibility forms as examples
**Step 2** - Migrate the remaining forms, swap out `/new-settings` for `/settings`
**Step 3** - Remove settings from Zustand store

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/FlowSettings/Visibility/schema.ts
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/FlowSettings/Visibility/schema.ts
@@ -3,5 +3,5 @@ import { boolean, object, type SchemaOf } from "yup";
 import { VisibilityFormValues } from "./types";
 
 export const validationSchema: SchemaOf<VisibilityFormValues> = object({
-  canCreateFromCopy: boolean().required(),
+  canCreateFromCopy: boolean().required().default(false),
 });
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/Contact/schema.ts
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/Contact/schema.ts
@@ -5,12 +5,14 @@ export const validationSchema = object().shape({
     .email(
       "Enter an email address in the correct format, like example@email.com",
     )
-    .required("Enter a contact email address"),
-  helpPhone: string().required("Enter a phone number"),
-  helpOpeningHours: string().required("Enter your opening hours"),
+    .required("Enter a contact email address")
+    .default(""),
+  helpPhone: string().required("Enter a phone number").default(""),
+  helpOpeningHours: string().required("Enter your opening hours").default(""),
   homepage: string()
     .url(
       "Enter a homepage URL in the correct format, like https://www.localauthority.gov.uk/",
     )
-    .required("Enter a homepage"),
+    .required("Enter a homepage")
+    .default(""),
 });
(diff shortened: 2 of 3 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/Contact/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+            <Input
+              name="homepage"
+              onChange={(e: ChangeEvent<HTMLInputElement>) =>
+                formik.setFieldValue("homepage", e.target.value)
+              }
+              value={formik.values.homepage}
+              errorMessage={formik.errors.homepage}
+              id="homepage"
+            />
+          </InputLabel>
```
DafyddLlyr: In future we could look at a "builder" pattern here, where you pass an array of fields to the component and they're rendered in a standard way - 

```ts
export const ContactForm = createSettingsForm({
  query: GET_TEAM_SETTINGS,
  mutation: UPDATE_TEAM_SETTINGS,
  legend: "Contact information",
  fields: [
    { name: "helpEmail", type: "email", label: "Reply-to email" },
    { name: "helpPhone", type: "tel", label: "Phone number" },
    // ...
  ],
});
```

I've not gone for this here as - 
 - it's a bigger refactor, and will slow down the completion of this task
 - it's not really needed (?) - we don't have that many forms, and there's a decent level of novelty between fields
 - it's easily reachable as a next step if we find the number of pages starts to grow

PR #5679, opened 2025-11-13
feat: read flow submission email from db

zz-hh-aa:
# What does this PR do?
- Updates API `select` permissions on the `flow_integrations` and `submission_integrations` tables
- Sets `submission_integrations.email_id` to `gen_random_uuid()` (we did this before but reverted that PR)
- Creates functions to get the flow ID, email ID and then submission email address

# Why? 
This is the next step in the enabling multiple submission email work. I think next is ensuring that records are written to the DB when a new flow is created. 

# Testing
- On a Pizza, set a submission email in team settings for a team (use an actual email you can access)
- Create a bare bones submission service, set to online + publish it (I would create a new submission service, because if you use an existing one with `application.type` the submission may fail with random data because it doesn't satisfy that application type's schema)
- Submit an application
- Check the submission email address and download the application
- API Docker logs should successfully log { submission_email: 'your@address.here' }

### DIFF
--- a/apps/api.planx.uk/modules/send/downloadApplicationFiles/index.ts
+++ b/apps/api.planx.uk/modules/send/downloadApplicationFiles/index.ts
@@ -6,11 +6,10 @@ import {
   getTeamEmailSettings,
   getFlowSubmissionEmail,
 } from "../email/service.js";
-import type { UUID } from "node:crypto";
 
 export async function downloadApplicationFiles(
   req: Request<
-    { sessionId?: UUID },
+    { sessionId?: string },
     unknown,
     unknown,
     { email?: string; localAuthority?: string }
@@ -49,7 +48,7 @@ export async function downloadApplicationFiles(
 
     // Get flow ID, in order to access flow submission email
     const flowId = await getFlowId(sessionId);
-    if (!flowId || flowId === undefined) {
+    if (!flowId) {
       return next({
         status: 400,
         message: "Failed to find flow ID for this sessionId",
@@ -58,7 +57,7 @@ export async function downloadApplicationFiles(
 
     // Get the flow submission email, which will run parallel to getTeamEmailSettings for now
     const submissionEmail = await getFlowSubmissionEmail(flowId);
-    if (!submissionEmail.submissionEmail) {
+    if (!submissionEmail) {
       return next({
         status: 400,
         message: "Failed to retrieve submission email for this flow",
--- a/apps/api.planx.uk/modules/send/email/service.ts
+++ b/apps/api.planx.uk/modules/send/email/service.ts
@@ -84,7 +84,7 @@ export async function getFlowSubmissionEmail(flowId: string) {
       flowId,
     },
   );
-  return response?.flowIntegrations[0]?.submissionIntegration;
+  return response?.flowIntegrations[0]?.submissionIntegration.submissionEmail;
 }
 
 interface GetSessionData {


apps/api.planx.uk/modules/send/downloadApplicationFiles/index.ts
```diff
@@ -41,6 +42,13 @@ export async function downloadApplicationFiles(
       });
     }
 
+    // Get flow ID, in order to access flow submission email
+    const flowId = await getFlowId(sessionId);
```
zz-hh-aa: Instead of creating two functions here, would it make more sense to set up a relationship between `lowcal_sessions` and `flow_integrations` (so that a session can immediately be linked back to an `email_id` via the `flow_id`) instead of doing this in two steps?

apps/api.planx.uk/modules/send/downloadApplicationFiles/index.ts
```diff
@@ -41,6 +47,13 @@ export async function downloadApplicationFiles(
...
+
+    // Get the flow submission email, which will run parallel to getTeamEmailSettings for now
+    const submissionEmail = await getFlowSubmissionEmail(flowId);
+    console.log(submissionEmail);
```
DafyddLlyr: So eventually this will swap out the clause at lines 32-39 here. We're currently checking the singular email from team settings, but we want to be checking the emails from `flow_integrations`.

Right now, these are synchronised, and there's no way to change them (break them).

It's probably worth just running this check now directly - it will have the same effect and we won't need to come back to fix this.

apps/api.planx.uk/modules/send/email/service.ts
```diff
@@ -35,6 +36,57 @@ export async function getTeamEmailSettings(localAuthority: string) {
...
+          flow_id
+        }
+      }
+    `,
+    {
+      session_id: sessionId,
+    },
+  );
+  return response?.lowcal_sessions[0].flow_id;
+}
```
DafyddLlyr: We should always keep out internal types in camelCase, and we can [destructure](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring) (convert) as part of the GQL query.

```suggestion
interface GetFlowId {
  lowcalSessions: {
    flowId: UUID;
  }[];
}

export async function getFlowId(sessionId: UUID) {
  const response = await $api.client.request<GetFlowId>(
    gql`
      query GetFlowId($session_id: uuid!) {
        lowcalSessions: lowcal_sessions(where: { id: { _eq: $session_id } }) {
          flowId: flow_id
        }
      }
    `,
    {
      session_id: sessionId,
    },
  );
  return response?.lowcalSessions[0].flowId;
}
```

apps/api.planx.uk/modules/send/downloadApplicationFiles/index.ts
DafyddLlyr: Good observation!

The DB should probably hold all relationships so that we have a proper understanding of how data models operate, and maximum flexibility. 

In this instance though I'd argue that we still want a two-step process.
 - There's no real cost as this scale of an extra round-trip to the DB
 - The email validation could fail - we don't want to over fetch at the validation stage. There's a risk we could leak data the user doesn't have access to. Arguably this email validation step should be a separate auth middleware.

apps/api.planx.uk/modules/send/downloadApplicationFiles/index.ts
DafyddLlyr: This will also allow us to start writing tests immediately here...!

If we set up mock queries to handle the new queries added in this PR, we can then test the various scenarios here.

I'm not sure if we've talked through API tests before (if we have, it's certainly been a while). We can either catch up on a call, or I can leave you async notes on how to set this up.

DafyddLlyr (changes requested):

apps/api.planx.uk/modules/send/downloadApplicationFiles/index.ts
DafyddLlyr: Here's the current tests for this endpoint for reference - https://github.com/theopensystemslab/planx-new/blob/oz/read-flow-submission-email/apps/api.planx.uk/modules/send/email/index.test.ts#L254-L326

You can see query mocks for the `GetTeamEmailSettings` table - what we'd be looking for here is setting up mocks to cover your queries.

DafyddLlyr:
@zz-hh-aa If you run `pnpm lint:fix` in the API directory it will resolve the issues CI is flagging 👍

DafyddLlyr (approved):

PR #5683, opened 2025-11-13
fix: Error message for checklist `fn` field

DafyddLlyr:
image


apps/editor.planx.uk/src/@planx/components/shared/BaseChecklist/model.ts
```diff
@@ -271,14 +280,14 @@ const uniqueGroupTitlesTest: TestConfig<AnyChecklist> = {
...
-  }
+  },
 };
 
 export const baseChecklistValidationSchema =
   baseNodeDataValidationSchema.concat(
     object({
       description: richText(),
-      fn: string(),
+      fn: string().nullable(),
```
DafyddLlyr: This is the only change, the rest is just linting!

PR #5684, opened 2025-11-13
[application] revise logic for notifying Slack on ECS deployment rollbacks

freemvmt:
Relates to [this ticket](https://trello.com/c/DJHQh8rf). Builds off #4245 which attempted to address same issue.

Will need to test this because some of the strings subject to AWS specs are partial guesswork / pulled from a mosaic of docs, but I think it's either correct or very close.

This PR doesn't represent a complete solution to the task given because it doesn't address this requirement: 'If one service fails, how should others be handled? All need to rollback to stay in sync'.

We make use of AWS EventBridge instead of creating a metric alarm, because it's a more 'direct' solution. I relied a lot on the exemplar `SERVICE_DEPLOYMENT_FAILED` event given [here](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs_service_deployment_events.html).

Applies only to Hasura for the moment, which is the only ECS service we're running which has circuit breakers applied.

Probably doesn't need doing here but we may want to bump Pulumi to [6.x](https://www.pulumi.com/registry/packages/aws/how-to-guides/6-0-migration/) and then [7.x](https://www.pulumi.com/registry/packages/aws/how-to-guides/7-0-migration/) at some point. That's ignoring the `awsx` package, which might be [slightly more work](https://github.com/pulumi/pulumi-awsx?tab=readme-ov-file#migration-from-0x-to-10) to get beyond `0.x`.

### DIFF
--- a/.github/workflows/push-main.yml
+++ b/.github/workflows/push-main.yml
@@ -122,7 +122,7 @@ jobs:
 
   preview:
     name: Pulumi Up
-    needs: 
+    needs:
       - build_planx_frontend
       - build_localplanning_services
     runs-on: ubuntu-22.04
@@ -151,7 +151,7 @@ jobs:
         with:
           name: localplanning_services_build
           path: ./${{ env.LOCALPLANNING_SERVICES_DIRECTORY }}/dist
-      - uses: pulumi/actions@v5
+      - uses: pulumi/actions@v6
         with:
           command: up
           stack-name: staging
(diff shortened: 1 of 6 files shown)


DafyddLlyr (approved):
Looking good! 👌 

re: Pulumi packages for updating - very much on board with this, been a bit of an issue in the past with docs that we were so far behind. I'd love to bring them up to date and then uncomment everything in `dependabot.yml` which refers to these.

freemvmt:
> re: Pulumi packages for updating - very much on board with this, been a bit of an issue in the past with docs that we were so far behind. I'd love to bring them up to date and then uncomment everything in `dependabot.yml` which refers to these.

Fab yes, I made a [ticket to this effect](https://trello.com/c/nt68eDW9).

PR #5685, opened 2025-11-13
feat: Migrate boundary form to new settings pattern

DafyddLlyr:
## What does this PR do?
- Migrates `BoundaryForm` to the new setting form pattern
- Split ups child elements such as PreviewMap
- Move PlanningData request and types to the `lib/planningData` layer
- Handles the validation check with planning data and the fetch of the boundary as an async step within `getMutationVariables()`

https://github.com/user-attachments/assets/1cca9511-4e57-46cf-8e1a-05e42b76177a


ianjon3s (approved):
Works great, layout will be fixed when merging in

PR #5694, opened 2025-11-14
feat(lps): Add Build timestamp to footer

DafyddLlyr:
image

### DIFF
--- a/apps/localplanning.services/src/components/Footer.astro
+++ b/apps/localplanning.services/src/components/Footer.astro
@@ -55,7 +55,7 @@ const lastUpdatedAt = new Date(buildDateTime).toLocaleString("en-GB", {
           }
         </div>
       </div>
-      <div class="flex justify-between gap-10">
+      <div class="flex flex-col sm:flex-row justify-between gap-8">
         <p class="text-body-md">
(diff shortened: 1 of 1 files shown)


apps/localplanning.services/src/components/Footer.astro
```diff
@@ -43,15 +55,19 @@ const legalPages = await getCollection("legal");
...
-      
-      <p class="text-body-md">
-        All content is available under the <a
-          class="paragraph-link paragraph-link--external"
-          href="https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/"
-          target="_blank"
-          rel="noopener noreferrer nofollow">Open Government License v3</a
-        >, except where otherwise stated
-      </p>
+      <div class="flex justify-between gap-10">
```
ianjon3s: nit can we change this to the following, as the new column gets very narrow at mobile sizes (also keeping the gap consistent with the rest of the footer layout at smaller sizes):
`flex flex-col sm:flex-row justify-between gap-8`

DafyddLlyr: Great spot - thank you!

PR #5695, opened 2025-11-14
feat: Add LPS url to Save page

DafyddLlyr:
image


apps/editor.planx.uk/src/hooks/useLPS.ts
DafyddLlyr: I made this a hook as we'll need this in the `EditorNavMenu` for https://trello.com/c/iM05XxH2/3491-add-link-to-local-planning-services-within-an-lpas-service-area-on-the-editor (@augustlindemer) and probably the settings pages also.

PR #5696, opened 2025-11-14
feat: Environment toggle

DafyddLlyr:
## What does this PR do?
Adds a selector to the header (for logged in users), allowing them to toggle environments

https://github.com/user-attachments/assets/361b2e93-4f47-43ce-92bd-3060540c93b6

### DIFF
--- a/apps/editor.planx.uk/src/components/EditorNavMenu/components/EnvironmentSelect.tsx
+++ b/apps/editor.planx.uk/src/components/EditorNavMenu/components/EnvironmentSelect.tsx
@@ -13,6 +13,7 @@ import Typography from "@mui/material/Typography";
 import ButtonBase from "@planx/components/shared/Buttons/ButtonBase";
 import React, { useState } from "react";
 import { useLocation } from "react-use";
+import { FONT_WEIGHT_SEMI_BOLD } from "theme";
 
 export interface Environment {
   name: string;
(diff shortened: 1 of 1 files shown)


apps/editor.planx.uk/src/components/EditorNavMenu/components/EnvironmentSelect.tsx
```diff
@@ -0,0 +1,193 @@
...
+          >
+            <CloseIcon />
+          </IconButton>
+        </StyledDialogTitle>
+        <Stack p={1} bgcolor={(theme) => theme.palette.background.dark} gap={1}>
+          {environments.map((env) => (
+            <StyledCard key={env.name} selected={env.name === currentEnv}>
+              <CardActionArea
+                LinkComponent={"a"}
+                href={env.url + pathname}
```
DafyddLlyr: Unless already logged in to the destination env, this pathname won't get preserved due to this known bug - https://trello.com/c/047oaBJh/2962-fix-redirects-after-auth

apps/editor.planx.uk/src/components/EditorNavMenu/components/EnvironmentSelect.tsx
```diff
@@ -0,0 +1,193 @@
...
+  const { pathname } = useLocation();
+
+  const handleOpen = () => setOpen(true);
+  const handleClose = () => setOpen(false);
+
+  return (
+    <>
+      <StyledButtonBase onClick={handleOpen} selected={false}>
+        {currentEnv}
+        <UnfoldMoreIcon sx={{ ml: 0.5 }} />
```
ianjon3s: Looking at this in the browser, I think we should clarify the relationship between the icon and the environment title by removing the 5px margin from the icon and adding to the button itself. Creates separation from the PlanX link and tightens the env toggle button title & icon.

Before vs after:
image

PR #5697, opened 2025-11-17
feat: Use `@tanstack/react-query` for `PlanningConstraints` component

DafyddLlyr:
## What does this PR do?
 - Removes the `swr` dependency
 - Replaces two `useSWR` calls within `PlanningConstraints` to two `useQuery` calls, wrapped in custom hooks
 - Updates tests to use `msw` instead of `swr`

Task - https://github.com/orgs/theopensystemslab/projects/7?pane=issue&itemId=134253816

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/PlanningConstraints/Public/index.test.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/PlanningConstraints/Public/index.test.tsx
@@ -1,10 +1,10 @@
 import { act } from "@testing-library/react";
 import ErrorFallback from "components/Error/ErrorFallback";
+import { http, HttpResponse } from "msw";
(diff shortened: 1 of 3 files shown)


apps/editor.planx.uk/src/lib/api/gis/requests.ts
```diff
@@ -0,0 +1,39 @@
...
+  const { data, request } = await apiClient.get(`/gis/${teamSlug}`, {
+    params: {
+      geom,
+      vals: dataValues.join(","),
+      ...params,
+    },
+  });
+
+  // Return full URL for auditing purposes
+  return { ...data, url: request.responseURL };
```
DafyddLlyr: We could (should?) consider returning the `url` from the API as it's always required for auditing, but we can also get it reliably this way.

PR #5705, opened 2025-11-17
fix: Only add `delay()` to specific loading state tests

DafyddLlyr:
Follow up to https://github.com/theopensystemslab/planx-new/pull/5583

Passing on multiple retries now on CI.

I still think the issue is the one outlined in the PR above - there's a race condition in the teardown on `/Send/Public.test.tsx`. I've still not working out which test it is which was causing the issue, but by restricting the use of the asynchronous `delay()` function to only where it's needed (testing the loading states) it seems to have resolved the issue 🤞


jessicamcinchak (approved):
:star2: :crossed_fingers:

PR #5707, opened 2025-11-18
feat: allow councils to retrieve submissions via FME workbench (prototype)

jessicamcinchak:
See https://trello.com/c/2Jclhgcj/3436-explore-fme-workbench-alternative-to-integrations-submission-workflow-with-camden

Relies on https://github.com/theopensystemslab/planx-core/pull/871

**Changes:**
- [Editor] Adds a new "Send" destination "FME Workbench" and associated event creation
- [API] Adjusts _existing_ `/upload-submission/:localAuthority` endpoint to accept and parse an optional query param `?notify=false`. When set to `=false`, we'll still upload the application to S3 if it hasn't already been and write an audit table entry with it's URL, but we'll skip the Power Automate notification step. 

**Open questions:**
- The "Submissions" page already displays "Sent to AWS S3" without any specific mention of Power Automate, so despite being a second option in the editor modal, both of these will appear the same in the table (but with slightly unique response shapes). Does this seem okay or too confusing? They should be rarely used together to "dual-process". 

**Next steps:**
- Create a new endpoint to `GET` a basic view of the `s3_applications` table which FME Workbench (or any tool) can make authenticated requests to return a list of recent submissions and their file URLs

### DIFF
--- a/apps/api.planx.uk/modules/send/s3/index.ts
+++ b/apps/api.planx.uk/modules/send/s3/index.ts
@@ -17,13 +17,12 @@ interface CreateS3Application {
   };
 }
 
-const sendToS3: SendIntegrationController = async (req, res, next) => {
+const sendToS3: SendIntegrationController = async (_req, res, next) => {
   const {
     payload: { sessionId },
   } = res.locals.parsedReq.body;
   const localAuthority = res.locals.parsedReq.params.localAuthority;
-  const sendPowerAutomateNotification =
-    req.query?.notify === "false" ? false : true;
+  const sendPowerAutomateNotification = res.locals.parsedReq.query.notify;
 
   const env =
     process.env.APP_ENVIRONMENT === "production" ? "production" : "staging";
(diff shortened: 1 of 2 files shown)


apps/editor.planx.uk/src/@planx/components/Send/Editor.tsx
```diff
@@ -146,6 +148,22 @@ const SendComponent: React.FC<Props> = (props) => {
...
+                  checked={formik.values.destinations.includes("fme")}
+                  onChange={() => toggleSwitch("fme")}
+                  label="Retrieve using FME Workbench"
+                  disabled={props.disabled}
+                />
+              </InputRow>
+              <Typography variant="body2">
+                Retrieve submissions using FME Workbench to download to your
+                local network. This option requires you to setup a workflow
+                which polls for new submissions on a schedule of your choice.
```
jessicamcinchak: Copy suggestions welcome as always!

PR #5709, opened 2025-11-18
chore: update bold copy formatting to GDS guidance [skip pizza]

augustlindemer:
GDS style guidance on bold/strong text formatting suggests:

> Only use bold to indicate interface elements in text that are explicitly telling the user what to do

&

> Do not use bold in other situations, for example to emphasise text.

Guidance also suggests:

> Don’t use “click” when talking about user interfaces because not all users click. Use “select”.

This PR removes emphasising bold from default copy and a few remaining uses of 'click'. Retains 'click or tap' for interactions that don't have distinct elements that can be selected.


jessicamcinchak (approved):
Thanks for GDS alignment & tidy ups here !

PR #5714, opened 2025-11-19
feat: add Local Planning Services link to team menu

augustlindemer:
Adds a link to the council's LPS page to their team page menu.

Uses availability of a local authority reference code to estimate whether a team is a council or not so as to disable the link for non-council teams (same as the Planning Data link).


jessicamcinchak (approved):
Simple & effective !

PR #5724, opened 2025-11-20
feat: Add 'resubmit' button on PlatformAdmins on the submissions page

DafyddLlyr:
## What does this PR do?
 - Adds a (long overdue!) "Resubmit" button to the submissions page
 - The button is only available to PlatformAdmins
   - Although it's a public endpoint (`/create-send-events`) this seems right as a resubmission usually goes hand-in-hand with some form of data fix

image


apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/ResubmitButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+        body: {
+          sessionId: params.row.sessionId,
+        },
+      },
+    };
+
+    mutate({ sessionId: params.row.sessionId, ...payload });
+  };
+
+  const isDisabled = params.row.status === "Success" || isPending || isSuccess;
```
jessicamcinchak: :+1:

PR #5726, opened 2025-11-20
feat: settings page for multiple submission emails

zz-hh-aa:
If this is behind a feature flag, is it fine to merge even as partial as it is? 

Eventually we want all CRUD operations but for now this only reads.

To test:
1. Go to `/{team}/new-settings/integrations`
2. Run `window.featureFlags.toggle("TEAM_SUBMISSION_INTEGRATIONS")` and refresh
3. Check console for team ID number
4. Go to Hasura and add a new record in Hasura `submission_integrations` table, using the same team ID (you'll have to paste in a random UUID for now)
5. Refresh the new settings page and the newly submission email should be read

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/IntegrationSettings/SubmissionEmails/index.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/IntegrationSettings/SubmissionEmails/index.tsx
@@ -1,0 +1,76 @@
+import Typography from "@mui/material/Typography";
+import { useStore } from "pages/FlowEditor/lib/store";
+import React, { ChangeEvent } from "react";
+import InputLabel from "ui/editor/InputLabel";
+import Input from "ui/shared/Input/Input";
+
+import SettingsFormContainer from "../../../shared/SettingsForm";
+import {
+  GET_TEAM_SUBMISSION_INTEGRATIONS,
+  UPDATE_TEAM_SUBMISSION_INTEGRATIONS,
+} from "./queries";
+import { defaultValues, validationSchema } from "./schema";
+import {
+  GetTeamSubmissionIntegrationsData,
+  SubmissionEmailFormValues,
+  UpdateTeamSubmissionIntegrationsVariables,
+} from "./types";
+
+// Parallel feature to SubmissionEmail (singular)
+export const SubmissionEmails: React.FC = () => {
+  const teamId = useStore((state) => state.teamId);
+  console.log(teamId, typeof teamId);
+  console.log("Query Variables:", { team_id: teamId });
+
+  return (
+    <SettingsFormContainer<
+      GetTeamSubmissionIntegrationsData,
+      SubmissionEmailFormValues,
+      UpdateTeamSubmissionIntegrationsVariables
+    >
+      query={GET_TEAM_SUBMISSION_INTEGRATIONS}
+      defaultValues={defaultValues}
+      queryVariables={{ team_id: teamId }}
+      mutation={UPDATE_TEAM_SUBMISSION_INTEGRATIONS}
+      getInitialValues={({ submissionIntegrations }) =>
+        submissionIntegrations?.[0] || defaultValues
+      }
+      getMutationVariables={(values) => ({
+        teamId,
+        submissionEmail: values.submissionEmail,
+        defaultEmail: values.defaultEmail,
+      })}
+      validationSchema={validationSchema}
+      legend="Submission information"
+      description={
+        <>
+          <Typography variant="body2">
+            Enter multiple email addresses to receive applications. You can
+            assign one email address per service. A default address must be
+            selected, which will be used as a fallback for all applications.
+          </Typography>
+        </>
+      }
+    >
+      {({ formik }) => (
+        <InputLabel label="Submission email" htmlFor="submissionEmail">
+          <Input
+            name="submissionEmail"
+            onChange={(e: ChangeEvent<HTMLInputElement>) =>
+              formik.setFieldValue("submissionEmail", e.target.value)
+            }
+            value={formik.values.submissionEmail ?? ""}
+            errorMessage={
+              typeof formik.errors.submissionEmail === "string"
+                ? formik.errors.submissionEmail
+                : undefined
+            }
+            id="submissionEmail"
+          />
+        </InputLabel>
+      )}
+    </SettingsFormContainer>
+  );
+};
+
+export default SubmissionEmails;
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/IntegrationSettings/SubmissionEmails/index.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/IntegrationSettings/SubmissionEmails/index.tsx
@@ -19,6 +19,7 @@ import {
 // Parallel feature to SubmissionEmail (singular)
 export const SubmissionEmails: React.FC = () => {
   const teamId = useStore((state) => state.teamId);
+  console.log({ teamId });
 
   return (
     <SettingsFormContainer<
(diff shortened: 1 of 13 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/IntegrationSettings/SubmissionEmails/schema.ts
```diff
@@ -0,0 +1,16 @@
+import { boolean, number,object, string } from "yup";
+
+import { SubmissionEmailFormValues } from "./types";
+
+export const validationSchema = object().shape({
+  teamId: number().required(),
```
DafyddLlyr: We don't actually need this to be part of the form - there's no corresponding field for it so formik validation can never pass.

We do need it for the mutation variables, so we can pass it directly in via `getMutationVariables()` (which you're already doing).

apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/IntegrationSettings/SubmissionEmails/types.ts
```diff
@@ -0,0 +1,16 @@
...
+  submissionIntegrations: {
+    submissionEmail: string;
+    defaultEmail: boolean;
+  }[];
+}
+
+export interface UpdateTeamSubmissionIntegrationsVariables {
+  submissionEmail: string;
+  defaultEmail: boolean;
+}
```
DafyddLlyr:
```suggestion
export interface UpdateTeamSubmissionIntegrationsVariables {
  teamId: number;
  submissionEmail: string;
  defaultEmail: boolean;
}
```

We also need to pass this value in to the mutation. It's a bit tricky atm because there's no hard link between GQL queries and their associated types but we'll resolve this (in the medium-ish?) term via codegen.

apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/IntegrationSettings/SubmissionEmails/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+            name="submissionEmail"
+            onChange={(e: ChangeEvent<HTMLInputElement>) =>
+              formik.setFieldValue("submissionEmail", e.target.value)
+            }
+            value={formik.values.submissionEmail ?? ""}
+            errorMessage={
+              typeof formik.errors.submissionEmail === "string"
+                ? formik.errors.submissionEmail
+                : undefined
+            }
```
DafyddLlyr: This is an awkward formik things as `errors` can have many types.

There's ` getIn()` helper we reach for to make this a bit neater - 
```suggestion
            errorMessage={getIn(formik.errors, "submissionEmail")}
```

apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/IntegrationSettings/SubmissionEmails/queries.ts
```diff
@@ -0,0 +1,34 @@
...
+`;
+
+export const UPDATE_TEAM_SUBMISSION_INTEGRATIONS = gql`
+  mutation UpdateSubmissionIntegration {
+    update_submission_integrations_by_pk(pk_columns: { email_id: $email_id }) {
+      submissionEmail: submission_email
+      teamId: team_id
+    }
+  }
+`;
```
DafyddLlyr: When we have the UI figured out, we probably want an "upset" pattern which would allow an insert or update to be accepted.

Docs: https://hasura.io/docs/2.0/mutations/postgres/upsert/#update-selected-columns-on-conflict

apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/IntegrationSettings/SubmissionEmails/queries.ts
```diff
@@ -0,0 +1,34 @@
...
+      returning {
+        emailId: email_id
+      }
+    }
+  }
+`;
+
+export const UPDATE_TEAM_SUBMISSION_INTEGRATIONS = gql`
+  mutation UpdateSubmissionIntegration {
+    update_submission_integrations_by_pk(pk_columns: { email_id: $email_id }) {
```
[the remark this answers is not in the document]
DafyddLlyr: Image

DafyddLlyr (changes requested):
This is working as expected to read values from the DB! ✅ I think this is actually _really_ close to handling write operations as well with a few small changes.

A number of small comments below, generally related to types and queries.

> _you'll have to paste in a random UUID for now_

We should apply the `default: gen_random_uuid()` pattern here in Hasura, and probably rename the column to just `id` instead of `email_id` (sorry for missing both of these earlier). If we retain `email_id` we'll need to keep a configure a cache ID for it within the GraphQL client - it's simpler to just use `id` where possible.

Docs: https://www.apollographql.com/docs/react/caching/cache-configuration#customizing-cache-ids

> _If this is behind a feature flag, is it fine to merge even as partial as it is?_

Yep absolutely! We'll get a number of small hidden PRs merged here 👍

zz-hh-aa:
Thanks Daf!

> We should apply the `default: gen_random_uuid()` pattern here in Hasura, 

Done in [74c9813](https://github.com/theopensystemslab/planx-new/pull/5726/commits/74c9813292e1d1a3ed9751e7c598c65fe5bcef79)

> and probably rename the column to just `id` instead of `email_id` (sorry for missing both of these earlier). If we retain `email_id` we'll need to keep a configure a cache ID for it within the GraphQL client - it's simpler to just use `id` where possible.

Done in [ebbedc2](https://github.com/theopensystemslab/planx-new/pull/5726/commits/ebbedc2fdf64abaf511265e82ae85fdb4d32ae85)

apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/IntegrationSettings/SubmissionEmails/queries.ts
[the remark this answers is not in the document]
zz-hh-aa: Working + tested fix here in [44da60c](https://github.com/theopensystemslab/planx-new/pull/5726/commits/44da60c1d2b4a5ea8f7e4c8d86002c79f6bd05f6)

apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/IntegrationSettings/SubmissionEmails/types.ts
DafyddLlyr: Please see comment below!

apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/IntegrationSettings/SubmissionEmails/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+export const SubmissionEmails: React.FC = () => {
+  const teamId = useStore((state) => state.teamId);
+  console.log({ teamId });
+
+  return (
+    <SettingsFormContainer<
+      GetTeamSubmissionIntegrationsData,
+      SubmissionEmailFormValues,
+      UpdateTeamSubmissionIntegrationsVariables
+    >
```
DafyddLlyr:
```suggestion
    <SettingsFormContainer<
      GetTeamSubmissionIntegrationsData,
      UpdateTeamSubmissionIntegrationsVariables
      SubmissionEmailFormValues,
    >
```

Hard to spot, but these are in the wrong order - it needs to match the following type - 

```ts
interface SettingsFormContainerProps<
  TData,
  TVariables,
  TFormValues extends FormikValues,
> {
...
}
```

DafyddLlyr (changes requested):
One small fix required then this is good to go!

DafyddLlyr (approved):

PR #5734, opened 2025-11-21
chore: allow Upload & Label, Next Steps, and Task List options to be customised [skip pizza]

augustlindemer:
Allows customisation of options in customisable Upload & Label, Next Steps, and Task List components in templated copies.


jessicamcinchak (approved):
Nice & simple :+1: 

(Would appreciate a pizza for any editor-impacting changes like this in future - assume only skipped here because SSL issue? Happy to double-check staging once merged here!)

augustlindemer:
> Nice & simple 👍
> 
> (Would appreciate a pizza for any editor-impacting changes like this in future - assume only skipped here because SSL issue? Happy to double-check staging once merged here!)

Thanks, will do Pizzas for these going forward!

PR #5741, opened 2025-11-23
feat(editor): Use API layer functions for public Pay component

DafyddLlyr:
## What does this PR do?
- Updates public `Pay` component to call the API via the `/api/client` instead of just calling `axios` directly

**Not in this PR**
I've not made the broader refactor to convert this component to use Tanstack Query. 

I've had a pass at this and it soon drifts into a wider refactor. Currently the state machine which runs this component (managed by `useReducer()`) is tracking client, server, and UI state. An implementation of this new method would hand over UI (loading, error, success) to Tanstack, as well as server state (GovUkPayment), and probably introduce a custom `useGovPay()` hook to handle the business logic.

This is a non-trivial task, with a fair amount of risk as it touches the payment process, so I think it's worth taking the time and doing this properly (if we want to do it at all), instead of just picking up this migration as quick tasks between other tickets as I've been doing.

For now, I'm advocating we leave this as is because -
 - What we have is robust
 - High risk
 - High time for migration (probably a day or two's work considering additional tests?)
 - Calling the API layer gets us enough benefit and consistency to be valuable

Regression tests passing here - https://github.com/theopensystemslab/planx-new/actions/runs/19628532279 ✅

### DIFF
--- a/e2e/tests/ui-driven/src/invite-to-pay/nominee.spec.ts
+++ b/e2e/tests/ui-driven/src/invite-to-pay/nominee.spec.ts
@@ -48,7 +48,9 @@ test.describe("Nominee journey @regression", async () => {
     const { paymentRequest, sessionId } = await setupPaymentRequest(request);
     await navigateToPaymentRequestPage(paymentRequest, page);
 
-    await expect(page.getByRole("heading", { name: "Pay" })).toBeVisible();
+    await expect(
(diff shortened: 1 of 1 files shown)


jessicamcinchak (approved):
Thanks for clear write up of trade-offs here! Agree this is pragmatic solution for now :heavy_plus_sign:

PR #5742, opened 2025-11-23
feat: Migrate `/:flow/about` page to new settings format

DafyddLlyr:
## What does this PR do?
 - Removes the "About" menu option
 - Nests the form fields within the new settings pages

## Questions
1. Previously, the "About" page was accessible for teamViewers - do we want to keep this functionality? Is there a better or alternative way of presenting this information for them (cc @ianjon3s)

2. Does showing the nested flows in this context really make sense?

3. We've got a mapping of one field to one `<SettingsFormSection/>` - is it better to just group these with a single "Save" button?

image


ianjon3s (approved):
Works great, few small comments:
- We should keep `visibility` as the first settings tab, it's the settings item most likely to be used often (`about` is more set-and-forget), and it's linked from the flow `online` / `offline` chip in breadcrumbs
- The link to LPS does not current open in a new tab, despite warning that it will
- Let's think of a better way to display flow information to non-team-editors. A locked form is not an ideal way to do this

PR #5743, opened 2025-11-24
feat: add simple `/submissions/:localAuthority` endpoint which can be polled by an FME script or similar

jessicamcinchak:
Follows on from #5707

### DIFF
--- a/apps/api.planx.uk/modules/send/fme/index.ts
+++ b/apps/api.planx.uk/modules/send/fme/index.ts
@@ -1,29 +1,32 @@
 import { subDays } from "date-fns";
-import type { NextFunction, Request, Response } from "express";
 import { gql } from "graphql-request";
+import { z } from "zod";
 import { $api } from "../../../client/index.js";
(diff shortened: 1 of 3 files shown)


apps/api.planx.uk/modules/send/fme/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  res: Response,
+  next: NextFunction,
+) {
+  const localAuthority = req.params?.localAuthority;
+  if (!localAuthority) {
+    return next({
+      status: 400,
+      message: "Missing values required to access submissions",
+    });
+  }
```
DafyddLlyr: I think it's worth setting up a simple validation schema for this. Right now it's just one param but we're pretty sure filtering will be an upcoming requirement here.

jessicamcinchak: Yep very fair, validation schema now added :heavy_plus_sign:

PR #5747, opened 2025-11-24
feat: Show first online date in admin panel

DafyddLlyr:
## What does this PR do?
- Adds a new computed field `flows.first_online_at`
- Updates `teams_summary` to read new column
- Updates `teams_summary` to use a lateral join instead of a left join + group by
- Updates UI to include new column with dates

We're jumping through hoops in a few places here to show "child" flow pseudo-rows within the admin panel which is a table for team records. Our version of MUI Data Grid [doesn't support row grouping](https://mui.com/x/react-data-grid/row-grouping/) - we might find that if we want more features here, and more details about flows, it might be time for another flow-specific table.

image

Trello ticket - https://trello.com/c/3RoUazCP/3490-update-lps-listings-on-schedule-or-via-webhook

### DIFF
--- a/apps/editor.planx.uk/src/ui/shared/DataTable/DataTable.stories.tsx
+++ b/apps/editor.planx.uk/src/ui/shared/DataTable/DataTable.stories.tsx
@@ -1,23 +1,37 @@
-// @ts-nocheck
+import Box from "@mui/material/Box";
 import { Meta, StoryObj } from "@storybook/react";
-import { getFlowNamesForFilter } from "pages/PlatformAdminPanel/utils";
+import {
+  formatDate,
+  getFlowNamesForFilter,
+} from "pages/PlatformAdminPanel/utils";
+import React from "react";
+import type { AdminPanelData, LiveFlow } from "types";
 
 import { DataTable } from "./DataTable";
 import { mockTeams } from "./mockTeams";
-import { ColumnFilterType } from "./types";
+import { ColumnFilterType, type DataGridProps } from "./types";
+
+type TableStoryProps = DataGridProps<AdminPanelData>;
 
 const meta = {
   title: "Design System/Atoms/Data table",
   component: DataTable,
-} satisfies Meta<typeof DataTable>;
-
-type Story = StoryObj<typeof meta>;
+  decorators: [
+    (Story) => (
+      <Box sx={{ height: "500px" }}>
+        <Story />
+      </Box>
+    ),
+  ],
+} satisfies Meta<TableStoryProps>;
 
 export default meta;
 
-const liveFlowValueOptions = getFlowNamesForFilter(mockTeams);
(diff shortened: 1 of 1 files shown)


apps/editor.planx.uk/src/ui/shared/DataTable/DataTable.stories.tsx
DafyddLlyr: Realised that this was broken, must have been out of action for a while - we might want to consider adding a `@storybook/test-runner` action to CI at some point (maybe to regression tests?).

apps/editor.planx.uk/src/pages/PlatformAdminPanel/PlatformAdminPanel.tsx
```diff
@@ -36,14 +36,26 @@ export const PlatformAdminPanel = () => {
       field: "referenceCode",
       headerName: "Reference code",
     },
-
     {
-      field: "liveFlows",
+      field: "liveFlowsNames" as keyof AdminPanelData,
```
DafyddLlyr: Old column - list of flows

apps/editor.planx.uk/src/pages/PlatformAdminPanel/PlatformAdminPanel.tsx
```diff
@@ -36,14 +36,26 @@ export const PlatformAdminPanel = () => {
...
       columnOptions: {
-        valueOptions: liveFlowValueOptions,
+        valueGetter: (_value: LiveFlow[], row: AdminPanelData) =>
+          row.liveFlows?.map(({ name }) => name),
+        valueOptions: liveFlowNameValueOptions,
+        sortable: false,
+      },
+    },
+    {
+      field: "liveFlowsDates" as keyof AdminPanelData,
```
DafyddLlyr: New column - list of corresponding dates

apps/editor.planx.uk/src/types.ts
```diff
@@ -87,6 +87,11 @@ export interface SectionNode extends Store.Node {
   };
 }
 
+export interface LiveFlow {
+  name: string;
+  firstOnlineAt: string;
+}
```
DafyddLlyr: **Why return the data in this format if it has to be split up again for the table?**

As the admin panel table view team records, and not flow records, we need to unpack this into two columns. This could have been a little simpler if the `teams_summary` table did this natively - just added an extra column of `first_online_at` dates.

I think it's worth keeping these grouped together as it -
- Gives us more flexibility if we want to use this view for a different purpose
- It's less brittle - if any sorting is added to one column, it's easy to forget for the other
- The split is an abstraction of the UI, not the data structure

ianjon3s (approved):
Renders as expected, Storybook version is fixed as well

PR #5728, opened 2025-11-20
chore(editor): Remove old settings routes and components

DafyddLlyr:
## What does this PR do?
- Removes the current components residing at `/:flow/settings` and `/:team/general-settings`
- Replaces them with the new settings components
- Updates naming conventions on folders and components to reduce ["smurf naming"](https://blog.codinghorror.com/new-programming-jargon/)
  - e.g. `/Team/TeamSettings/TeamAdvancedSettings` just becomes `Team/Settings/Advanced`
- Removes unused properties and functions from the Zustand `settingsStore` 
- `DesignSettings` now lives within an `/OldSettings` folder - we can pick up a migration of this in a future sprint

Git has struggled with reconciliation on a few files here as they've been renamed - there's no new content, just removing and renaming.

PR #5751, opened 2025-11-24
feat: write multiple submission emails from new team settings

zz-hh-aa:
# Testing
1. Visit https://5751.planx.pizza/{TEAM-SLUG}/settings/integrations
2. Run `window.featureFlags.toggle("TEAM_SUBMISSION_INTEGRATIONS")` in the console and refresh
4. Add new email/s or update existing ones (changing address or toggling 'Default')
5. Check Hasura for table updates

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Integrations/SubmissionEmails/index.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Integrations/SubmissionEmails/index.tsx
@@ -68,11 +68,7 @@ export const SubmissionEmails: React.FC = () => {
             Submission Emails
           </Typography>
           <RadioGroup
-            value={
-              formik.values.submissionIntegrations.findIndex(
-                (emailObj) => emailObj.defaultEmail,
-              ) || 0
-            }
+            value={getDefaultRadioIndex(formik.values.submissionIntegrations)}
             onChange={(e) => {
               const selectedIndex = parseInt(
                 (e.target as HTMLInputElement).value,
@@ -106,6 +102,15 @@ export const SubmissionEmails: React.FC = () => {
       )}
     </SettingsFormContainer>
   );
+};
+
+const getDefaultRadioIndex = (
+  submissionIntegrations: SubmissionEmailInput[],
+) => {
+  const defaultIndex = submissionIntegrations.findIndex(
+    (emailObj) => emailObj.defaultEmail,
+  );
+  return defaultIndex !== -1 ? defaultIndex : 0;
 };
 
 const EmailsEditor: React.FC<EditorProps<SubmissionEmailInput>> = (props) => {


apps/editor.planx.uk/src/pages/FlowEditor/components/NewSettings/TeamSettings/IntegrationSettings/SubmissionEmails/index.tsx
```diff
@@ -17,10 +24,20 @@ import {
...
+    deleteSubmissionIntegration,
+    { loading: deleteLoading, error: deleteError },
+  ] = useMutation(DELETE_TEAM_SUBMISSION_INTEGRATIONS, {
+    refetchQueries: [
+      { query: GET_TEAM_SUBMISSION_INTEGRATIONS, variables: { teamId } },
+    ],
+    onError: (error) => {
+      console.error("Delete error:", error);
+    },
+  });
```
DafyddLlyr: Here's how destructive operations generally work in the Editor (we should write this down somewhere...!)

**Happy path**
- User removes item from list (e.g. via delete button)
- User hits save
- List updated

**Unhappy path**
- User removes item from list (e.g. via delete button)
- User changes mind (closes modal, hits cancel, closes tab)
- List not updated - item is retained

An example of this can be found in the question or checklist components, or the Global Settings page. This also matched how items are added - they aren't persisted until a user hits "Save".

This should simplify things as we just need to handle one operation here, when "Save" is it. The action for "Reset changes" then becomes clear - this just resets the form to its original state (already handled by `SettingsFormContainer`).

DafyddLlyr: So, actions here...!
 - We can remove this mutation (though thanks for using tanstack..! 😄)
- On delete, we update the form state, not server state (`ListManager` will do this for you)
- On save, many items can be updated

DafyddLlyr (changes requested):
A good start and some tricky UI challenged here for sure.

I'd recommend reaching for the `<ListMangager/>` component here - 
 - it's an established pattern within the Editor
 - it will handle the add/remove/update logic
 - it will allow us to simplify things here

I think once you make this change, it should be much simpler 👌

DafyddLlyr:
Just a heads up that I've just merged https://github.com/theopensystemslab/planx-new/pull/5728 which updates the file paths of these components - a rebase should resolve this 👌

zz-hh-aa:
I've gotten stuck again! I'm not sure if I overcomplicated things again... It feels like my current implementation is very close to working, `getMutationVariables` logs as expected, eg:
`{
    "teamId": 1,
    "emails": [
        {
            "submissionEmail": "asdf@fla.a",
            "defaultEmail": false
        }
    ]
}`

But even though my `teamId` logs the expected value, I'm stuck on a GQL error `Message: expecting a value for non-nullable variable: "teamId", Location: undefined, Path: undefined`

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Integrations/SubmissionEmails/schema.ts
```diff
@@ -1,15 +1,27 @@
...
+        .required("Submission email is required"),
+      defaultEmail: boolean().required(),
+      teamId: number().required(),
+    }),
+  ).required("At least one new email must be added."),
+  saved: array(
+    object().shape({
+      submissionEmail: string().email().required(),
+      defaultEmail: boolean().required(),
+      teamId: number().required(),
```
zz-hh-aa: Saved values validation seemed weird to include, but I needed to include it to get validation to pass. Should I be altering `SettingsForm` so that it expects a different shape instead?

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Integrations/SubmissionEmails/types.ts
```diff
@@ -1,16 +1,24 @@
...
+export interface SubmissionEmail {
   submissionEmail: string;
   defaultEmail: boolean;
+  teamId: number;
+}
+
+export interface SubmissionEmailMutation {
+  submission_email: string;
+  default_email: boolean;
+  team_id: number;
```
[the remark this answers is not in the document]
DafyddLlyr: Yeah there's no real ways around this when working with these `*_input` GraphQL objects - they need to be snake_cased as this is what the DB requires.

So we just have to cheat ;) 

We can reach for `SnakeCaseProperties<T>` from [`type-fest`](https://github.com/sindresorhus/type-fest) for a nice way to type these as camel_case in our interfaces, but still enforce that Hasura requires snakeCase

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Integrations/SubmissionEmails/index.tsx
```diff
@@ -1,71 +1,170 @@
...
+                formik.setFieldValue("saved", savedEmails);
+                formik.setFieldValue("input", newEmails);
+              }}
+              newValue={() => ({
+                submissionEmail: "",
+                defaultEmail: false,
+                teamId: teamId,
+              })}
+              Editor={EmailsEditor}
+              maxItems={10}
```
DafyddLlyr: I think this is just a copy/paste issue - there should be no restriction on the number of values here.

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Integrations/SubmissionEmails/types.ts
```diff
@@ -1,16 +1,24 @@
...
+}
+
+export interface SubmissionEmailFormValues {
+  input: SubmissionEmail[];
+  saved: SubmissionEmail[];
 }
```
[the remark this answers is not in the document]
zz-hh-aa: Absolutely (one of) the source(s) of confusion 😅 --don't we need to distinguish the two so that we know what exists in the db already and what needs to be passed as mutation variables? 

Or, I guess if it's an upsert operation, we pass everything and if the variables exist exactly as they are, nothing will happen?

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Integrations/SubmissionEmails/index.tsx
```diff
@@ -1,71 +1,143 @@
...
+            />
+          </RadioGroup>
+        </>
       )}
-    </SettingsFormContainer>
+    />
+  );
+};
+
+const EmailsEditor: React.FC<EditorProps<SubmissionEmailInput>> = (props) => {
```
DafyddLlyr: The radio button element is not vertically aligned, meaning it's not in-line with the input and delete button.

Image

zz-hh-aa: Have done this in `InputRow` with `alignItems: "center"` and checked that it hasn't affected instances elsewhere, eg in `SetFee`. Looks like it rarely has multiple children anyway!

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Integrations/SubmissionEmails/index.tsx
```diff
@@ -1,71 +1,143 @@
...
+        formik: FormikProps<SubmissionEmailValues>;
+      }) => (
+        <>
+          <Typography variant="h6" style={{ marginBottom: "1rem" }}>
+            Submission Emails
+          </Typography>
+          <RadioGroup
+            value={formik.values.submissionIntegrations.findIndex(
+              (emailObj) => emailObj.defaultEmail,
+            )}
```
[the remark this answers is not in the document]
zz-hh-aa: Done in [1344012](https://github.com/theopensystemslab/planx-new/pull/5751/commits/1344012a3dcad4ba716f3d5449376a7821aae741)

DafyddLlyr (approved):
Two minor issues, but good to go apart from that...! 👍

PR #5754, opened 2025-11-25
feat: display flow name and summary on first page of submission services

jessicamcinchak:
See https://trello.com/c/X4uY49nT/3444-show-service-title-and-summary-on-the-service-first-page-email-input

**Before:** 
Screenshot from 2025-11-24 19-50-03

**After:**
Screenshot from 2025-11-25 08-37-27

**Next:**
- Add a default first page for non-submission services !

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Flow/About/components/Summary.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Flow/About/components/Summary.tsx
@@ -35,7 +35,7 @@ const Summary: React.FC = () => {
           Summarise what the service does.
           <br />
           <br />
-          This wording is used on the{" "}
+          This is used on the entry page of online, published services and the{" "}
           <Link href={lpsURL} target="_blank" rel="noopener noreferrer">
             Find Local Planning Services (opens in a new tab)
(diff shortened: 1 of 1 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/lib/store/settings.ts
```diff
@@ -45,6 +46,8 @@ export const settingsStore: StateCreator<
 
   flowStatus: undefined,
 
+  flowSummary: undefined,
```
DafyddLlyr: non-blocking - Here's the alternative pattern I was outlining on the call yesterday - 

- Keep the additional property on on the fetch, and updated type as you have done
- Don't add flowSummary to the store - it's already in the Apollo cache
- Add a query where you need it to fetch flowId and flowSummary using Apollo's `useQuery()`
  - A network request won't fire
  - Apollo will match on flowId and return the flow summary

DafyddLlyr: This is for sure non-blocking as there's some more work to be done here around the graphql client first (type policies) as well as making some common utilities (`useTeam()`, `useFlow()` etc) that I want to set up.

PR #5773, opened 2025-11-26
feat: add a default entry page to non-submission services with flow name and summary

jessicamcinchak:
Follows on from #5754


jessicamcinchak:
Closing in favor of #5818

PR #5775, opened 2025-11-26
feat: Optimise `sortIdsDepthFirst()`

DafyddLlyr:


DafyddLlyr:
No real saving here - this isn't the cause of the issue.

PR #5782, opened 2025-11-26
feat: add a property.type fallback and conditional render

augustlindemer:
We have had repeated issues with submissions failing due to missing `property.type` data. This has occurred in fully council controlled services (Southwark's call for sites), discretionary services (Doncaster's report a breach), and statutory services (Barnet's amend a planning application). This is partially due to content dependencies with Property information components toggled to enable overwrites with no corresponding property-type content to handle it (strictly editor error) and partially due to services being wholly uninterested in property types (no editor error, but a mismatch between content and product requirements).

This PR adds a fallback in having the FindProperty component set an `unclassified` `property.type` value if we have no OS returned `property.type`. 

This PR also refactors the `showPropertyTypeOverride` switch to handle cases in which property types are irrelevant. It changes the 'allow overwrite' toggle to a 'support property type' toggle, which filters out displaying the property type to the user at all. It more clearly communicate the switch as one for supporting or not supporting property type specification and calls out the Property types flow dependency explicitly.


jessicamcinchak (changes requested):
Simple, smart solution ! But one question/suggestion below about implementation to consider please - happy to chat through if I'm missing anything about why to do at this stage instead

augustlindemer:
As of now, this PR does not address the Review component at all. Is it a concern that property type would be listed on the Review page with a 'Change' link irrespective of ability to do so specifically? This is identical to current behaviour and has to my knowledge never been raised as an issue -- **however**, there is a small change in that it could now be the first time a user is shown this. Looking at the Review component it is not clear to me how we could conditionally exempt the property type from the Review page.

apps/editor.planx.uk/src/@planx/components/PropertyInformation/Editor.tsx
```diff
@@ -76,10 +77,16 @@ function PropertyInformationComponent(props: Props) {
...
+              label="Support property type"
               disabled={props.disabled}
             />
           </InputRow>
+          <Typography variant="body2" sx={{ mt: 1 }}>
+            When enabled, this component will display the property type with a
+            'change' link to overwrite it. To support the property type, the{" "}
+            <strong>Open Systems Lab - Property types</strong> flow must be
+            nested immediately before this component.
+          </Typography>
```
jessicamcinchak: Very clear :heavy_plus_sign:

jessicamcinchak (approved):
Super! Thanks for changes here and chatting through this one yesterday ! 

Let's get this piece merged, then I can show you how to open a quick follow-up to adjust Review page to also hide "Property type" row if "Property type not supported" is toggled !

PR #5799, opened 2025-12-01
chore(perf): Improve lookup of component types when publishing

DafyddLlyr:
Small room for improvement spotted here - https://github.com/theopensystemslab/planx-new/pull/5733#discussion_r2576469938

Currently, we're iterating over the entire published flow multiple times in order to get component type metadata. By generating a set of unique component types once, as can use this as a source of truth to derive `hasSendComponent`, `hasPayComponent` and `hasSections`.

This should (marginally) improve the performance of the publish operation, as well as putting in a place a pattern next time we need to extract more metadata.


jessicamcinchak (approved):
Great !

PR #5818, opened 2025-12-02
feat: add a default entry page to non-submission services with flow name and summary

jessicamcinchak:
Follows on from #5754 

I still need to update / add tests here !

**Testing instructions:**
- Open the `/published` link of any online, non-submission service - it should now display a new default entry page with `flowName` and `flowSummary`
- Click "Continue" from the entry page and fill out a couple questions
- Confirm you can close your tab, re-open the URL, and you haven't lost your progress (eg should _not_ see the entry page again)
- Confirm you can "Restart" in the upper right to clear your session, and you start back over at the entry page

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/lib/store/shared.ts
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/lib/store/shared.ts
@@ -91,6 +91,7 @@ export const sharedStore: StateCreator<
       changedNode: undefined,
       saveToEmail: "",
       currentCard: null,
+      hasAcknowledgedSingleSessionEntry: false,
     });
 
     removeSessionIdSearchParam();
--- a/apps/editor.planx.uk/src/pages/Preview/SingleSession.tsx
+++ b/apps/editor.planx.uk/src/pages/Preview/SingleSession.tsx
@@ -1,0 +1,80 @@
+import Box from "@mui/material/Box";
+import Typography from "@mui/material/Typography";
+import Card from "@planx/components/shared/Preview/Card";
+import { CardHeader } from "@planx/components/shared/Preview/CardHeader/CardHeader";
+import { getLocalFlow } from "lib/local";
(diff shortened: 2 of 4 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/lib/store/shared.ts
```diff
@@ -91,6 +91,7 @@ export const sharedStore: StateCreator<
       changedNode: undefined,
       saveToEmail: "",
       currentCard: null,
+      hasAcknowledgedSingleSessionEntry: false,
```
jessicamcinchak: On "restart", we nullify the `saveToEmail` for submission services and now set `hasAcknowledgedSingleSessionEntry` to false - both of these help control visibility of the default entry page for their respective service "types".

Is this still a reasonable responsibility of the `shared` store?

apps/editor.planx.uk/src/pages/Preview/SingleSession.tsx
```diff
@@ -0,0 +1,80 @@
...
+      ) : (
+        <Main>
+          <Card handleSubmit={() => setHasAcknowledgedSingleSessionEntry()}>
+            <CardHeader title={flowName} description={flowSummary}></CardHeader>
+            <Box
+              mb={(theme) => theme.spacing(1)}
+              mt={(theme) => theme.spacing(0.75)}
+            >
+              <Typography variant="subtitle1" component="div">
+                {`You can leave and come back to this service anytime using the same link, device, and browser.`}
```
jessicamcinchak: Copy suggestions welcome here! 

Here's an example of before (using Content node) and after:
Screenshot from 2025-12-01 20-23-54
Screenshot from 2025-12-01 20-23-38

I'm guessing councils won't want to do away with the "this is guidance-only" disclaimer, nor does it make sense to necessarily add to the "summary" - is there a generic way we can include on this page that is less "planning" specific and would work across all non-submission services?

PR #5828, opened 2025-12-03
feat: redirect users to default team

jmcbroom:
Redirects users to the default team upon login.

- add column migration
- populate column migration
- add redirect to `/` route based on the existence of the `defaultTeamId`
- assign `defaultTeamId` during `createAndAddUserToTeam`
- updates many tests and mock users along the way to include `defaultTeamId`

### DIFF
--- a/apps/editor.planx.uk/src/routeTree.gen.ts
+++ b/apps/editor.planx.uk/src/routeTree.gen.ts
@@ -602,6 +602,7 @@ const AuthenticatedAppTeamFlowFlowEditorNodesParentNodesIdEditBeforeRoute =
   )
 
 export interface FileRoutesByFullPath {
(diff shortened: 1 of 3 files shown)


apps/hasura.planx.uk/migrations/default/1764739906434_assign_default_team_id/up.sql
```diff
@@ -0,0 +1,116 @@
...
+    final_to_update as (
+        select * from unioned
+        union all
+        select * from assigned_by_email
+    )
+    
+update users
+set default_team_id = final_to_update.team_id
+from final_to_update
+where users.id = final_to_update.id;
```
jessicamcinchak: Pizza `users.default_team_id` values look good :heavy_check_mark:

PR #5833, opened 2025-12-03
fix(hasura): Filter out delete session events from submission log

DafyddLlyr:
## What's the problem?
Delete events are being added to the `submissions_services_log`. This is happening how as it's 28 days since https://github.com/theopensystemslab/planx-new/pull/5284 hit production.

image

## What's the solution?
Filter out delete events, in the same way we filter out email events.

### DIFF
--- a/apps/hasura.planx.uk/migrations/default/1764758889205_filter_out_delete_events/up.sql
+++ b/apps/hasura.planx.uk/migrations/default/1764758889205_filter_out_delete_events/up.sql
@@ -1,0 +1,82 @@
+CREATE OR REPLACE VIEW "public"."submission_services_log" AS 
+ WITH payments AS (
+         SELECT ps.session_id,
+            ps.payment_id AS event_id,
+            'Pay'::text AS event_type,
+            initcap(ps.status) AS status,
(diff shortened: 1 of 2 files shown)


apps/hasura.planx.uk/migrations/default/1764758889205_filter_out_delete_events/up.sql
```diff
@@ -0,0 +1,82 @@
...
+            (seil.response)::jsonb AS response,
+            seil.created_at,
+            (EXISTS ( SELECT 1
+                   FROM retries r
+                  WHERE (r.id = seil.id))) AS retry
+            FROM (hdb_catalog.hdb_scheduled_events se
+               LEFT JOIN hdb_catalog.hdb_scheduled_event_invocation_logs seil ON ((seil.event_id = se.id)))
+            WHERE (
+               ((se.webhook_conf)::text !~~ '%email/%'::text)
+               AND ((se.webhook_conf)::text !~~ '%delete-session/%'::text)
```
DafyddLlyr: This is the new condition we're checking -  the rest stays all the same.

PR #5834, opened 2025-12-03
feat: View application (via `<SummaryList />`)

DafyddLlyr:
## What does this PR do?
 - Allows applicants to view a summary of their application, and then print to PDF if they wish to do so
 - Replaces current `.csv` download method

## Context
Originally, the plan here was to use the HTML download endpoint - the same as LPS and the submissions page. I got this up and running in https://github.com/theopensystemslab/planx-new/pull/5829 but quite quickly hit a number of issues (outlined below). I've changed this across to now use the existing `<SummaryList/>` component.

|  | HTML | `<SummaryList/>` component |
|--------|--------|--------|
| Will work natively for non-S&R sessions? | ❌ | ✅ |
| Immediate display (no API request) | ❌ | ✅ |
| Matches current applicant view | ❌ | ✅ |
| Simpler refactor once we have a unified component | ❌ | ✅ |
| Meaningful information hierarchy | ✅ | ❌ | 
| Meet a11y requirements | ❓ | ✅ |

It's certainly not a perfect fit, but hopefully a decent placeholder until we can get a proper shared component.

image

image

Example PDF - [Your form - PlanX.pdf](https://github.com/user-attachments/files/24142487/Your.form.-.PlanX.pdf)

### DIFF
--- a/apps/editor.planx.uk/src/pages/Preview/ApplicationViewer.tsx
+++ b/apps/editor.planx.uk/src/pages/Preview/ApplicationViewer.tsx
@@ -5,24 +5,29 @@ import SummaryListsBySections from "@planx/components/shared/Preview/SummaryList
 import { PrintButton } from "components/PrintButton";
 import { useStore } from "pages/FlowEditor/lib/store";
 import React from "react";
+import { useLocation } from "react-use";
 import ApplicationSummary from "ui/public/ApplicationSummary";
 
 import StatusPage from "./StatusPage";
 
 const NoContentPage = () => (
-  <StatusPage
-    bannerHeading="Your form"
-    buttonText="Start a new form"
-  >
-    <Typography variant="body2">
-      No answers to display
-    </Typography>
+  <StatusPage bannerHeading="Your form" buttonText="Start a new form">
+    <Typography variant="body2">No answers to display</Typography>
   </StatusPage>
-)
+);
 
 const ApplicationViewer: React.FC = () => {
-  const [flow, passport, breadcrumbs, govUkPayment] = useStore(state => [state.flow, state.computePassport(), state.breadcrumbs, state.govUkPayment]);
-  const hasPaymentInfo = Boolean(govUkPayment);
+  const [flow, passport, breadcrumbs, govUkPayment] = useStore((state) => [
+    state.flow,
+    state.computePassport(),
+    state.breadcrumbs,
+    state.govUkPayment,
+  ]);
+  const { pathname } = useLocation();
+
+  const hasPaid = Boolean(govUkPayment);
+  const hasSentInviteToPay = pathname?.match("pay/invite");
+  const showFeeBreakdown = hasPaid || hasSentInviteToPay;
 
   // No answered questions - saving on first question
   const noBreadcrumbs = Object.keys(breadcrumbs).length === 0;
@@ -31,7 +36,7 @@ const ApplicationViewer: React.FC = () => {
   return (
     <StatusPage bannerHeading="Your form" bannerText="Review your answers">
       <Typography variant="h2">Overview</Typography>
-      <ApplicationSummary/>
+      <ApplicationSummary />
       <Typography variant="h2">Your responses</Typography>
       <SummaryListsBySections
         changeAnswer={() => null}
@@ -41,15 +46,15 @@ const ApplicationViewer: React.FC = () => {
         breadcrumbs={breadcrumbs}
         sectionComponent="h2"
       />
-      {hasPaymentInfo &&
+      {showFeeBreakdown && (
         <Box mb={4}>
           <Typography variant="h2">Fee breakdown</Typography>
           <FeeBreakdown />
         </Box>
-      }
+      )}
       <PrintButton />
     </StatusPage>
   );
-}
+};
 
 export default ApplicationViewer;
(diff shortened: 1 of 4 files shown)


apps/editor.planx.uk/src/pages/Preview/ApplicationViewer.tsx
```diff
@@ -0,0 +1,51 @@
...
+const ApplicationViewer: React.FC = () => {
+  const [flow, passport, breadcrumbs, govUkPayment] = useStore(state => [state.flow, state.computePassport(), state.breadcrumbs, state.govUkPayment]);
+  const hasPaymentInfo = Boolean(govUkPayment);
+
+  // No answered questions - saving on first question
+  const noBreadcrumbs = Object.keys(breadcrumbs).length === 0;
+  if (noBreadcrumbs) return <NoContentPage />;
+
+  return (
+    <StatusPage bannerHeading="Your form" bannerText="Review your answers">
```
jessicamcinchak: Aware there is currently no clear record of your sessionId in the `SummaryList` below, should we add it to the header here?

apps/editor.planx.uk/src/pages/Preview/ApplicationViewer.tsx
```diff
@@ -0,0 +1,51 @@
...
+  >
+    <Typography variant="body2">
+      No answers to display
+    </Typography>
+  </StatusPage>
+)
+
+const ApplicationViewer: React.FC = () => {
+  const [flow, passport, breadcrumbs, govUkPayment] = useStore(state => [state.flow, state.computePassport(), state.breadcrumbs, state.govUkPayment]);
+  const hasPaymentInfo = Boolean(govUkPayment);
```
jessicamcinchak: nit: I think this will fail to correctly catch ITP which does _not_ set `govUkPayment` key? 

Instead I think we just want to check breadcrumbs for presence of `Pay` component type (in _any_ mode including info-only)?

apps/editor.planx.uk/src/ui/public/ViewApplicationLink.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const ViewApplicationLink: React.FC = () => (
+  <Link
+    sx={{ display: "block", mt: 1, textAlign: "right" }}
+    component={ReactNaviLink}
+    href="view-application"
+    prefetch={false}
+  >
+    <Typography variant="body2">
+      View form
```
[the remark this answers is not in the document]
DafyddLlyr: I think "Print form" is clearer and more meaningful, copy updated ✅ 

Implementing `"(opens in a new tab)"` is more complex as this relies on the state (`sessionId` + `emailReplyToId`) which just lives in memory. Opening in a new tab currently will give you a "not found" error due to these missing values.

We don't want to use these variables as URL params as this makes a permalink to the session data which is risky.

I think the best solution for this would be -
 - implement Zustand store as localStorage/sessionStorage (shareable across tabs) - not quite sure of the wider implications of this
 - implement access via tokens
 
Neither of these are particularly quick or easy so my inclination would be to keep this within the same window for now.

The issue of course is that a user might try to "open in a new tab" which will not work. We can convert this to a button? (cc @ianjon3s)

apps/editor.planx.uk/src/pages/Preview/ApplicationViewer.tsx
```diff
@@ -0,0 +1,51 @@
...
+        passport={passport}
+        breadcrumbs={breadcrumbs}
+        sectionComponent="h2"
+      />
+      {hasPaymentInfo &&
+        <Box mb={4}>
+          <Typography variant="h2">Fee breakdown</Typography>
+          <FeeBreakdown />
+        </Box>
+      }
```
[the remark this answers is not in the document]
DafyddLlyr: Please see above comment, added via the `<ApplicationSummary/>` component

DafyddLlyr:
@jessicamcinchak I think all the above comments have been addressed and this is now ready for a re-review 🙂

apps/editor.planx.uk/src/@planx/components/Confirmation/Confirmation.stories.tsx
```diff
@@ -47,8 +36,29 @@ export const Basic = {
...
+              title: '45, Greenfield Road, London SE22 7FF',
+            },
+            'application.type': 'ldc.proposed',
+          },
+        }),
+      });
+
+      return <Story />;
+    },
+  ],
```
jessicamcinchak: :+1:

apps/editor.planx.uk/src/ui/public/ApplicationSummary.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+// TODO - Retire in favour of ODP Schema application type descriptions or fallback to flowName
+function getWorkStatus(passport: Store.Passport): string | undefined {
+  switch (passport?.data?.["application.type"]?.toString()) {
+    case "ldc.existing":
+      return "existing";
+    case "ldc.proposed":
+      return "proposed";
+  }
+}
```
jessicamcinchak: nit: I think we _can_ replace this now using planx-core's `getValidSchemaValues("ApplicationType")` ?? 

Or do we have another exported helper which returns both value _and_ description? Not a blocker, but a nice cleanup if we can!

jessicamcinchak (approved):
Great! Sorry it's taken me a bit to come back to this one

PR #5838, opened 2025-12-03
fix(hasura): Filter out delete events from `submission_services_log` (take 2)

DafyddLlyr:
Follow up and fix for #5833

### DIFF
--- a/apps/hasura.planx.uk/migrations/default/1764778977255_v2_filter_delete_events/up.sql
+++ b/apps/hasura.planx.uk/migrations/default/1764778977255_v2_filter_delete_events/up.sql
@@ -1,0 +1,158 @@
+CREATE
+OR REPLACE VIEW "public"."submission_services_log" AS WITH payments AS (
+  SELECT
+    ps.session_id,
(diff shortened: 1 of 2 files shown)


apps/hasura.planx.uk/migrations/default/1764778977255_v2_filter_delete_events/up.sql
```diff
@@ -0,0 +1,158 @@
...
+      hdb_catalog.hdb_scheduled_events se
+      LEFT JOIN hdb_catalog.hdb_scheduled_event_invocation_logs seil ON ((seil.event_id = se.id))
+    )
+  WHERE
+    (
+      -- Exclude email events
+      ((se.webhook_conf) :: text !~~ '%/send-email%' :: text)
+      AND (
+      -- Exclude delete events
+        (se.webhook_conf) :: text !~~ '%/delete-session%' :: text
```
DafyddLlyr: This is the only change - before it was `'%delete-session/%'`

This now matches the `hdb_catalog.hdb_scheduled_events` table - 

Image

PR #5841, opened 2025-12-03
fix: ensure "change" link is also hidden on Review page for otherwise hidden rows

jessicamcinchak:
Fixing a bug spotted post-merge for #5837 

The individual `SummaryList` "components" can't returned `undefined`, as that only hides the first two columns and the "change" link lingers in the third still. Instead, we want to move these visibility conditions up into the existing `isValidComponent` check !

**Before (property type is only hidden in 2/3 columns in row 3):**
image

**After (property type correctly fully hidden):**
Screenshot from 2025-12-03 14-41-56

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
@@ -180,18 +180,38 @@ function SummaryListsBySections(props: SummaryListsBySectionsProps) {
     const Component = node.type && presentationalComponents[node.type];
     const isPresentationalComponent = Boolean(Component);
     const isAutoAnswered = userData.auto;
-    const isInfoOnlyMode =
-      node.type === TYPES.FileUploadAndLabel &&
-      props.flow[nodeId].data?.hideDropZone;
 
-    return !isAutoAnswered && isPresentationalComponent && !isInfoOnlyMode;
+    // If FileUploadAndLabel was presented for info-only without a dropzone,
+    //  then omit like Content and Notice component types
+    const isInfoOnlyMode =
+      node.type === TYPES.FileUploadAndLabel && node.data?.hideDropZone;
+
+    // If we didn't show `Property type` on the PropertyInformation card,
+    //   then don't show on Review page either (still in passport for schema validation)
+    const propertyTypeIsNotSupported =
+      node.type === TYPES.PropertyInformation &&
+      !node.data?.showPropertyTypeOverride;
+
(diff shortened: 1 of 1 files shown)


apps/editor.planx.uk/src/@planx/components/shared/Preview/SummaryList.tsx
```diff
@@ -180,18 +180,38 @@ function SummaryListsBySections(props: SummaryListsBySectionsProps) {
...
+    const propertyTypeIsNotSupported =
+      node.type === TYPES.PropertyInformation &&
+      !node.data?.showPropertyTypeOverride;
 
-    return !isAutoAnswered && isPresentationalComponent && !isInfoOnlyMode;
+    // If we weren't able to fetch external data, omit PlanningConstraints component from Review
+    //   Any constraints will instead have been manually prompted (and already captured on Review) as questions or checklists
+    const failedTofetchConstraints =
+      node.type === TYPES.PlanningConstraints &&
+      !props.passport?.data?.["_constraints"];
```
jessicamcinchak: Adding both of these as "double negatives" was very unintuitive and honestly lost a silly amount of time to it and still don't fully understand? 

All I know is that this snippet filtered out _both_ the valid `FindProperty` node _and_ my invalid `PropertyInformation` node:
```
const propertyTypeSupported = node.type === TYPES.PropertyInformation && node.data?.showPropertyTypeOverride;

return !isAutoAnswered && isPresentationComponent && !isInfoOnlyMode && propertyTypeSupported; 
```

What am I missing??

augustlindemer: From what I can see, the way `propertyTypeSupported` requires `node.type === TYPES.PropertyInformation`, means that

> return !isAutoAnswered && isPresentationComponent && !isInfoOnlyMode && propertyTypeSupported; 

by default filters out anything that isn't `PropertyInformation` (because it will always be `!propertyTypeSupported`).

Making it `propertyTypeIsNotSupported` means we only return false if it is `TYPES.PropertyInformation` without `showPropertyTypeOverride`.

In the positive, `&& propertyTypeSupported` gatekeeps _everything_ from being a valid components except overwrite toggled `PropertyInformation` nodes. In the negative, `propertyTypeIsNotSupported` gatekeeps only nodes that are `PropertyInformation` without the toggle (as it should).

jessicamcinchak: Appreciate extra eyes & sense check on this, your explanation makes perfect sense !

PR #5706, opened 2025-11-17
feat(api): Define API contract for enhance project descriptions feature

DafyddLlyr:
## What does this PR do?
- Defines API for `ai/project-description` feature, as agreed with @ianjon3s and @freemvmt 
- Sets up boilerplate for API endpoint, basic input validation schema, swagger docs
- Sets up API requests in frontend + types

## Not in this PR
 - Any functionality!
 - Tests (or test setup)
 - Guardrails - input content validation (e.g. jailbreaking attempts, off-topic requests, abuse, etc...)

## Next steps...
 - @ianjon3s to wire up frontend
 - @freemvmt to continue work on model

PR #5848, opened 2025-12-04
[api] enhance project descriptions with external AI model

freemvmt:
Adds the business logic to the API for the contract defined in #5706.

This PR adds logic such that the `ai/project-description/enhance` endpoint uses Vercel's AI SDK to prompt Gemini 2.5 Pro (for now) to enhance the proposal description submitted in the body of the request.

I note here that I'm not entirely convinced by AI SDK and the attendant AI Gateway - I think a [LangChain](https://docs.langchain.com/oss/javascript/langchain/overview) / [OpenRouter](https://openrouter.ai/docs/guides/community/frameworks-and-integrations-overview) set up could be more flexible, but this was potentially easier to implement when we're trying to get an MVP out, and is sufficient for our current use case. More thoughts on this scattered through [this Slack thread](https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1764948993437619?thread_ts=1764928037.193329&cid=C088K9ZL8EA).

## TODO

- [x] Add some middleware to defend against prompt injection attacks. I've been doing some reading and have collected some resources [here](https://www.notion.so/opensystemslab/Planning-Proposal-Description-Generator-24035d469ad180469bf6f834aa5ea69b?source=copy_link#2bf35d469ad18072ac27f6b5010b4a58).
- [x] Emit audit trail to db, including original and enhanced descriptions.
- [ ] Think about how to validate the system prompt and model choice.
- [x] Think about how to test all this!
- [x] Add Pulumi config for staging/prod

### DIFF
--- a/apps/api.planx.uk/vitest.config.ts
+++ b/apps/api.planx.uk/vitest.config.ts
@@ -17,7 +17,7 @@ export default defineConfig({
       reporter: ["lcov", "html", "text-summary"],
       thresholds: {
         statements: 77.31,
-        branches: 60.14,
+        branches: 60.36,
         functions: 77.11,
         lines: 77.65,
         autoUpdate: true,


apps/api.planx.uk/modules/ai/projectDescription/service.ts
```diff
@@ -1,21 +1,98 @@
...
+If the description can be improved, return your amended version, with the status $${GatewayStatus.ENHANCED}.`;
 
 export const enhanceProjectDescription = async (
-  _original: string,
-): Promise<Result> => {
-  // return { ok: false, error: "INVALID_DESCRIPTION" }
-  // return { ok: false, error: "SERVICE_UNAVAILABLE" }
-  return { ok: true, value: "Enhanced!" };
+  original_description: string,
+  model_name: string = DEFAULT_MODEL,
```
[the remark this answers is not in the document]
freemvmt: the python is in my bones ☠️

apps/api.planx.uk/modules/ai/projectDescription/service.ts
```diff
@@ -1,21 +1,98 @@
...
+} from "./types.js";
 
-type Result = Success | Failure;
+const DEFAULT_MODEL = "google/gemini-2.5-pro";
+const SYSTEM_PROMPT_ENHANCE = `Act as a trained planning officer at a local council in the UK.
+You will be prompted with a description of a planning application which is to be submitted to a local council.
+Please improve it such that it has the best chance of being accepted and validated, without adding unneccessary detail.
+If the description does not seem to be related to a planning application, respond with the status $${GatewayStatus.INVALID}.
+If the description is already good enough, return the original without amendment, with the status $${GatewayStatus.NO_CHANGE}.
+If the description can be improved, return your amended version, with the status $${GatewayStatus.ENHANCED}.`;
```
DafyddLlyr: I'd be very interested in exposing this as an `.md` file.

This would mean a few things - 
 - Easier iteration for non-GH OSL users
 - We can more easily reference and link to it
 - Separation of concerns - splitting out business logic and the prompt more explicitly
 - More structure, more easily - headings, bullet points, etc. This can all get done now, but a little less easily
 - Sets up a simpler pattern if we want a future shared location (e.g. a `/packages/prompts` within the monorepo)

In all fairness a few of these feel slightly YAGNI, but it seems better to set up a good pattern now.

apps/hasura.planx.uk/migrations/default/1768329328302_create_table_public_ai_gateway_exchange_log/up.sql
```diff
@@ -0,0 +1,44 @@
...
+  "prompt" text NOT NULL,
+  "response" text,
+  "gateway_status" text,
+  "guardrail_tripped" boolean NOT NULL DEFAULT FALSE,
+  "guardrail_reason" text,
+  "guardrail_message" text,
+  "token_usage" integer,
+  "cost_usd" float8,
+  "vercel_generation_id" text,
+  "response_time_ms" integer,
```
DafyddLlyr: Nice...!

apps/api.planx.uk/server.ts
```diff
@@ -108,7 +108,7 @@ const authRoutes = await getAuthRoutes(passport);
 
 // Setup API routes
 app.use(adminRoutes);
-app.use("/ai", aiRoutes);
+app.use(aiRoutes);
```
DafyddLlyr: Thanks for sorting this - I recall it being an issue with the AI rate limiter being applied correctly.

Testing via API docs I can see this is working as expected ✅ 

Image

apps/api.planx.uk/modules/ai/routes.ts
```diff
@@ -1,16 +1,29 @@
...
+import { sanitiseInput } from "./middleware/sanitiseInput.js";
 
 const router = Router();
-router.use(aiLimiter);
 
+// we apply a stricter rate limit to the below /ai routes
+router.use("/ai", aiLimiter);
+
+// we implement some basic deterministic 'guardrails' before sending user input to (third-party) LLM
+// XXX: if we find these to be insufficient, consider a solution based on AWS Bedrock (or similar)
```
DafyddLlyr: We should set (via trello or Slack) an audit date to ensure we check logs on a semi-regular basis during the initial roll-out.

apps/api.planx.uk/modules/ai/projectDescription/system.md
DafyddLlyr: This is great!

apps/api.planx.uk/modules/ai/projectDescription/service.ts
```diff
@@ -1,21 +1,111 @@
...
+    }
+
+    const prompt = `<user_input>${original_description}</user_input>`;
+    const res = await generateObject({
+      model: result.model,
+      output: "object",
+      schema: z.object({
+        enhancedDescription: z.string().trim().max(250),
+        status: z.enum([...GATEWAY_SUCCESS_STATUSES, GATEWAY_STATUS.INVALID]),
+      }),
```
DafyddLlyr: nit: Maybe this could be co-located with other schemas in `types.ts`?

apps/api.planx.uk/modules/ai/index.test.ts
DafyddLlyr: Always appreciate good test coverage! 🙌 

A suggestion re:maintainability - it's likely the middlewares here will be shared by further LLM integration endpoints so we don't really need to tie the their tests to closely to the "project description" implementation. I think co-locating a set of `redactPii.test.ts`, `sanitiseInput.test.ts` etc alongside the middlewares themselves will serve us better as a method of ensuring that any additions or changes get test coverage also. In this case they will of course will still have to be hitting the project description endpoint ultimately.

I'd also take a look at coverage here - we don't currently enforce this via CI, but this PR is brining coverage down below the current threshold - 

Image

Image

It looks like it might be a case of adding coverage for logging (only mocking the actual LLM call).

DafyddLlyr (changes requested):
This is great - clear, robust, and well-structured 👍 

There's a few minor comments below to consider, in particular the test coverage which should be a quick fix all being well.

apps/api.planx.uk/modules/ai/middleware/validateModel.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { API_ERROR_STATUS } from "../types.js";
+
+// currently we just accept a fixed list of recent Google Gemini models
+export const ACCEPTED_MODEL_IDS = [
+  "google/gemini-2.5-pro",
+  "google/gemini-2.5-flash",
+  "google/gemini-3-pro-preview",
+  "google/gemini-3-flash",
+];
```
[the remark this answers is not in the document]
freemvmt: Great question! It's up for grabs. For now I thought this was a safer move, because there a couple risks with opening up model choice to [anything in Vercel's AI Gateway](https://vercel.com/open-systems-labs-projects/~/ai-gateway/model-list?sortField=outputCost):
1. Someone could make requests to extremely expensive models (e.g. `openai/gpt-5.2-pro`) and rinse through our budget super fast, even with stricter rate limiting
2. Not all of the models claim 'zero data retention', which I think is preferable for us? (Google does)
3. Other stuff I haven't thought of yet?

Of course, this also adds an extra bit of maintenance which we'd maybe like to avoid. An argument in favour of a home rolled gateway like LiteLLM perhaps 🤔

apps/api.planx.uk/modules/ai/routes.ts
DafyddLlyr: Sounds great!

PR #5849, opened 2025-12-05
feat: add "Guidance disclaimer" to flow settings

jessicamcinchak:
A follow-up requirement to proceeding with #5818 per `#planx-design` thread here https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1764680130281489?thread_ts=1764033551.816899&cid=C04DZ1NBUMR

**Changes:**
- Renames route from `/settings/legal-disclaimer` to `/settings/disclaimers`
- Renames tab from `Legal disclaimer` to `Disclaimers`
- Renames `Legal disclaimer` to `Result disclaimer` (UI-only, _not_ planning to migrate any database JSONB property names here)
- Adds `Guidance disclaimer` which is "on" by default with defualt text

Screenshot from 2025-12-04 20-02-53


apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Flow/Disclaimers/index.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+import GuidanceDisclaimer from "./components/Guidance";
+import ResultDisclaimer from "./components/Result";
+
+const Disclaimers: React.FC = () => (
+  <>
+    <ResultDisclaimer />
+    <GuidanceDisclaimer />
```
jessicamcinchak: In the future, this a good candidate for "conditional settings" - eg hide for submission services. But leaving always-visible for now!

PR #5868, opened 2025-12-08
fix: Visual regression of `ListManager` on global settings page

DafyddLlyr:
Small visual regression - `ListManager` icons (drag handle and delete) are outside the wrapper.

| Before | After |
|--------|--------|
| image | image |


ianjon3s:
Thanks for picking this up @DafyddLlyr

PR #5869, opened 2025-12-09
chore: include notification status in Submissions log S3 upload events

jessicamcinchak:
Quick one! 

Camden is officially testing the new "FME Workbench" send destination (here https://editor.planx.dev/camden/camden-apply-for-a-lawful-development-certificate/submissions), which currently shows up in the Submissions log as "Upload to S3" - same as the "MS Power Automate" send destination. 

It's intentional that these destinations do basically the same thing under the hood and councils aren't limited to specific tools, but it's helpful for debugging to know if a notification was sent or not, which is currently only possible by expanding the response to check for status 202 (notification sent and webhook received) or 204 (no content/notification).

Now, we'll display "Upload to AWS S3" or "Upload to AWS S3 (no notification)" as the "Event" name.

### DIFF
--- a/apps/hasura.planx.uk/migrations/default/1765238556684_update_s3_upload_event_with_notification/up.sql
+++ b/apps/hasura.planx.uk/migrations/default/1765238556684_update_s3_upload_event_with_notification/up.sql
@@ -1,0 +1,79 @@
+CREATE OR REPLACE VIEW "public"."submission_services_log" AS 
+ WITH payments AS (
+         SELECT ps.session_id,
+            ps.payment_id AS event_id,
+            'Pay'::text AS event_type,
+            initcap(ps.status) AS status,
+            jsonb_build_object('status', ps.status, 'description', pse.comment, 'govuk_pay_reference', ps.payment_id, 'govuk_pay_metadata', ps.gov_pay_metadata) AS response,
+            ps.created_at,
+            false AS retry
+           FROM (payment_status ps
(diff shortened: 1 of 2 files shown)


apps/hasura.planx.uk/migrations/default/1765238556684_update_s3_upload_event_with_notification/up.sql
```diff
@@ -0,0 +1,79 @@
...
+                  WHERE (se.tries > 1)
+                  GROUP BY seil.event_id))
+        ), submissions AS (
+         SELECT ((((seil.request -> 'payload'::text) -> 'payload'::text) ->> 'sessionId'::text))::uuid AS session_id,
+            se.id AS event_id,
+                CASE
+                    WHEN ((se.webhook_conf)::text ~~ '%bops%'::text) THEN 'Submit to BOPS'::text
+                    WHEN ((se.webhook_conf)::text ~~ '%uniform%'::text) THEN 'Submit to Uniform'::text
+                    WHEN ((se.webhook_conf)::text ~~ '%email-submission%'::text) THEN 'Send to email'::text
+                    WHEN ((se.webhook_conf) :: text ~~ '%?notify=false%' :: text) THEN 'Upload to AWS S3 (no notification)' :: text -- eg FME Workbench
```
jessicamcinchak: This extra `WHEN` condition is the only change!

PR #5871, opened 2025-12-09
feat: Warn against losing editor modal changes (formikRef)

ianjon3s:
## What does this PR do?

Second attempt at a solution to this, based on guidance by @DafyddLlyr here:
https://github.com/theopensystemslab/planx-new/pull/5851

As part of https://trello.com/c/2xsHU7tv/3448-optimise-editing-modal we have the following task:

> When trying to close without saving, editors will be prompted with a reminder.

This PR uses `Ref` to notify parent forms of a change against initial values, with a new hook `useFormikWithRef` to keep component changes light.

As proof-of-concept I've applied the changes to the Text input component.

Testing:
https://5871.planx.pizza/testing/close-modal-warning

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
@@ -165,15 +165,18 @@ const FormModal: React.FC<{
     return obj;
   };
 
+  const isDirty = (formik: FormikProps<any>): boolean => {
+    return !isEqual(
+      normalizeFormValues(formik.values),
+      normalizeFormValues(formik.initialValues),
+    );
+  };
+
   const hasUnsavedChanges = () => {
-    if (!formikRef.current) return false;
+    const formik = formikRef.current;
+    if (!formik) return false;
 
-    const { values, initialValues } = formikRef.current;
-
-    const normalizedValues = normalizeFormValues(values);
-    const normalizedInitialValues = normalizeFormValues(initialValues);
-
-    return !isEqual(normalizedValues, normalizedInitialValues);
+    return isDirty(formik);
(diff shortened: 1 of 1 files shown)


apps/editor.planx.uk/docs/adding-a-new-component.md
```diff
@@ -72,6 +73,8 @@ function SetValueComponent(props: Props) {
     },
   });
 
+  useFormikWithRef(formik, props.formikRef);
```
ianjon3s: Nice single line (+ import) solution for components

apps/editor.planx.uk/src/@planx/components/shared/useFormikWithRef.ts
```diff
@@ -0,0 +1,17 @@
...
+export function useFormikWithRef<Values>(
+  formik: FormikProps<Values>,
+  formikRef?: MutableRefObject<FormikProps<Values> | null>,
+): void {
+  useEffect(() => {
+    if (formikRef) {
+      formikRef.current = formik;
+    }
+  }, [formik, formikRef]);
+}
```
DafyddLlyr:
```suggestion
export function useFormikWithRef<T extends FormikValues>(
  formikConfig: FormikConfig<T>,
  formikRef?: MutableRefObject<FormikProps<T> | null>,
) {
  const formik = useFormik<T>({
    validateOnChange: false,
    validateOnBlur: false,
    ...formikConfig,
  });

  useEffect(() => {
    if (formikRef) {
      formikRef.current = formik;
    }
  });

  return formik;
}
```

I think something like this, which acts as a wrapper for the `useFormik()` hook will be a bit more comprehensive. We can now also store shared config here like `validateOnChange` and `validateOnBlur`.

PR #5884, opened 2025-12-11
refactor: Editor team nav menu as root menu

ianjon3s:
## What does this PR do?

Carried out in preparation for logging in directly to primary team page:
https://trello.com/c/NOMTFszI/3447-log-into-primary-team-page

- Refactors team level nav menu to account for it being the entry/main page for a team
- Updates routes so that Notion pages can be accessed at the team level
- Updates styles to complement new format
- Adds an "external" icon to external links
- Adds an optional "new" prop to add a chip

**Before changes:**
image

**After changes:**
image

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/GetStarted.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/GetStarted.tsx
@@ -3,6 +3,7 @@ import Box from "@mui/material/Box";
 import { styled } from "@mui/material/styles";
 import Typography from "@mui/material/Typography";
 import { Link } from "@tanstack/react-router";
+import { useParams } from "@tanstack/react-router";
 import React from "react";
 
 const Root = styled(Box)(({ theme }) => ({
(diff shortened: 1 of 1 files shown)


apps/editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
```diff
@@ -220,44 +257,103 @@ function EditorNavMenu() {
...
+
+  // Count total visible routes across all sections
+  const totalVisibleRoutes = visibleSections.reduce(
+    (count, section) => count + section.routes.length,
+    0,
+  );
 
   // Hide menu if the user does not have a selection of items
-  if (visibleRoutes.length < 2) return null;
+  if (totalVisibleRoutes < 2) return null;
```
DafyddLlyr: Is this logic actually still applicable?

Currently, isn't the case that - 
 - Only platform admin will see anything on the main menu (global settings, admin panel, select a team)
 - There will always be >2 routes on the team menu

Maybe we can simplify this with a check which is essentially "don't show the root menu if user is not platformAdmin or analyst"?

apps/editor.planx.uk/src/components/EditorNavMenu/styles.tsx
```diff
@@ -45,30 +45,46 @@ export const MenuButton = styled(IconButton, {
...
+    boxShadow: "0 1px 1.5px 0 rgba(0, 0, 0, 0.2)",
   },
   ...(isActive && {
     background: theme.palette.common.white,
     color: theme.palette.text.primary,
-    border: `1px solid ${theme.palette.border.main}`,
-    "&:hover": {
-      borderColor: theme.palette.border.main,
-    },
+    boxShadow: "0 1px 1.5px 0 rgba(0, 0, 0, 0.2)",
```
DafyddLlyr: This `isActive` condition doesn't work for both flows and team settings.

Image

Image

DafyddLlyr (changes requested):
A few minor comments below, no showstoppers. Actually very nice looking - big improvement to have the grouping I think 👍 

I'm very aware we've made life pretty challenging for @nichgalzin with many recent route changes - leading to a bunch of rebases.

For now, I'm inclined to hold this (and the "redirect" PR) back in order for the routing PR to proceed. I'll cross-check with Nich on progress at the start of next week and start a thread on `#planx-dev` to come up with a plan here.

apps/editor.planx.uk/src/components/EditorNavMenu/types.ts
```diff
@@ -9,8 +9,15 @@ export interface Route {
   route: string;
   accessibleBy: Role[] | AllUsers;
   disabled?: boolean;
+  isNew?: boolean;
```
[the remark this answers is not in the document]
DafyddLlyr: Yep lets run with this then until we need something more comprehensive 👍

DafyddLlyr:
@ian I think we can now rebase and merge this one...!

PR #5894, opened 2025-12-12
fix: Editor menu active states

ianjon3s:
## What does this PR do?

- Fixes editor navigation menu active states by taking into consideration:
  - Team page with filter/sort URL params
  - Nested settings pages
  
  
Currently failing on:
- Team pages (`/barnet` etc)
- Team settings pages
- Flow settings pages


DafyddLlyr (approved):
Ah very good catch - nice to fix directly on `main` 🙌

PR #5899, opened 2025-12-12
fix(ci): Add working-directory to `pnpm i` step

DafyddLlyr:
Next issue - missing a `working-directory`...!


jessicamcinchak (approved):
went for a walk between this one and the last :sweat_smile: sorry for slow feedback loop! thanks for working through these !

PR #5900, opened 2025-12-13
feat(editor): Setup `EnhancedTextInput` component

DafyddLlyr:
## What does this PR do?
 - Standard boilerplate for new `EnhancedTextInput` component (relies on https://github.com/theopensystemslab/planx-core/pull/895)
 - Introduces new "toggle" method of changing component type, instead of standard dropdown
 - Basic and rough public interface
 - Sets up an `ENHANCED_TEXTINPUT` feature flag

https://github.com/user-attachments/assets/ccf0371f-f06d-4133-96b0-243b7f4fddd7

## What's not in this PR?
1. I've not put much thought into the generated passport values, or analytics events, that this component will generate. A good topic for future discussion. I've put in placeholders like `userAction`, `original`, `suggested`, but not really explored how this will map to the actual passport.

2. `SummaryList` view for the component

3. Payload mapping for this component

4. Tests - to follow once we have everything in place

### One user-facing component, mapping to two real components
This is the first instance where there's a divergence between what an Editor considers to be a component, and what's actually a component.

After much discussion on show & tells and design calls, it was decided that we won't implement an "Agent" component (and AI specific component which needs to be placed in the graph), but instead integrate this directly into an existing component. There's a few reasons we would like this to be a separate component within the codebase, and not just an extension of the `TextInput` type - 

- It's not a true extension - we don't envision an email + AI enhancement, or a phone + AI enhancement in future
- It will generate a pretty different breadcrumb, and mixing these up would be complex
- It's simpler to expand the list of types and data structures, than to add fields and handle migrations (despite the added UI complexity)

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Editor.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Editor.tsx
@@ -19,8 +19,8 @@ import InputRow from "ui/shared/InputRow";
 import { DataFieldAutocomplete } from "../shared/DataFieldAutocomplete";
 import { ICONS } from "../shared/icons";
 import type { EditorProps } from "../shared/types";
-import { parseEnhancedTextInput, TASKS, validationSchema } from "./model";
-import { type EnhancedTextInput } from "./types";
+import { parseEnhancedTextInput, taskDefaults, TASKS, validationSchema } from "./model";
+import { type EnhancedTextInput, type Task } from "./types";
 
 type Props = EditorProps<ComponentType.EnhancedTextInput, EnhancedTextInput>;
 
@@ -83,7 +83,7 @@ const EnhancedTextInputComponent = (props: Props) => {
                   required
                   value={formik.values.fn}
                   onChange={(value) => formik.setFieldValue("fn", value)}
-                  disabled={props.disabled}
+                  disabled
                   errorMessage={formik.errors.fn}
                 />
               </InputRow>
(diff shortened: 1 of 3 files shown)


apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+            title={props.title}
+            description={props.description}
+            info={props.info}
+            policyRef={props.policyRef}
+            howMeasured={props.howMeasured}
+          />
+          <TaskComponent {...props} />
+        </Card>
+      )}
+    </Formik>
```
DafyddLlyr: I'm currently working on how we manage a multi-step / multi-card layout here as per the latest Figma designs.

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
```diff
@@ -119,7 +159,7 @@ const FormModal: React.FC<{
   extraProps?: any;
 }> = ({ type, handleDelete, Component, id, before, parent, extraProps }) => {
   const { navigate } = useNavigation();
-  const formikRef = useRef<FormikProps<any> | null>(null);
+  const formikRef = useRef<FormikProps<BaseNodeData & unknown> | null>(null);
```
DafyddLlyr: I missed that this was an `any` when reviewing https://github.com/theopensystemslab/planx-new/pull/5871

apps/editor.planx.uk/src/pages/FlowEditor/data/types.ts
```diff
@@ -46,5 +47,10 @@ export const fromSlug = (slug: string): TYPES | undefined => {
   const entry = Object.entries(SLUGS).find(
     ([_typeString, currentSlug]) => currentSlug === slug,
   );
+
+  // TextInput and EnhancedTextInput share a value in order to keep 
+  // the Select component consistent
+  if (entry?.[1] === "enhanced-text-input") return TYPES.TextInput;
```
DafyddLlyr: This is the hacky sort of thing I was hoping to avoid by having a totally separate component, but it doesn't feel _too_ bad currently...?

When we redesign the component selection we'll need to keep in mind that these are no longer 1:1 mappings.

PR #5923, opened 2025-12-16
fix: Editor overflow width and toggle position

ianjon3s:
## What does this PR do?

Fixes issue highlighted by @DafyddLlyr in Slack:
https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1765904899672269

- Fixing the graph editor max-width for non `teamEditors` (nav menu not present) caused a regression for `teamEditors` (nav menu present

To test:
1. View an active (graph larger than viewport) flow, content should not overflow the viewport width
2. Do the same with a flow that you don't have editor permissions for


DafyddLlyr (approved):
Perfect - thank you!

PR #5927, opened 2025-12-16
fix(editor): Consistent alignment on filters

DafyddLlyr:
| Before | After |
|--------|--------|
| image | image |


ianjon3s (approved):
Thanks for spotting this one, love the attention to detail.

PR #5932, opened 2025-12-18
feat: check if service charge is enabled on publish and add filter to flows list

jessicamcinchak:
Very similar to recent addition of "is fee carrying", but now explicitly checking if service charges are enabled (which should be a subset of fee-carrying, but not "every").

See this thread for recent context https://opensystemslab.slack.com/archives/C5Q59R3HB/p1765882440851399 & Trello ticket https://trello.com/c/6gs0QwfL/3520-add-a-filter-for-services-that-carry-service-charges

With this new column on publish, we'll also know from which date the service charge "went live to the public" (via join to `flows.status` where "online") which will support the next step of a "platform admin view of service charges". 

**Next steps / follow-up PRs:**
- Migration to populate existing/latest `published_flows.service_charge_enabled`
- Add new column to nightly sync script (once on prod)


apps/hasura.planx.uk/migrations/default/1766017701068_add_service_charge_col_diff_function/up.sql
DafyddLlyr: Well remembered..!

DafyddLlyr (approved):
Great!

PR #5933, opened 2025-12-18
feat: Add timestamp to and sequence to breadcrumbs

DafyddLlyr:
## What does this PR do?
 - Adds a `createdAt` property to breadcrumbs (`UserData`)
 - Adds a `seq` property to breadcrumbs

### Context
This is a simple addition to breadcrumb generation, which will not be referred to / implemented until at least 56 days have passed from when this gets merges to `production`. This will act as a failsafe to ensure **all** breadcrumbs have these new properties, whilst avoiding a complex data migration of existing sessions.

**Why both a sequence and timestamp?**
Whilst testing it became apparent that it's possible for two breadcrumbs to end up with the same `createdAt` - hence the need for fake timers. Whilst I've not been able to trigger this directly within the application, it's certainly possible that this could happen with auto-answered nodes. Relying purely on a timestamp will result in an unstable sorting method.

Adding a sequence serves as a more robust canonical source of the order of these objects - we can easily and reliably sort on this. When changing answer to navigating back, the sequence is always overwritten as a user cannot "jump" through the service.

It seems worthwhile keeping the `createdAt` as it's still valuable for debugging submissions. Discuss!

**Navigating via "back"**
```mermaid
sequenceDiagram
    autonumber
    participant U as User
    participant B as Breadcrumbs
    
    U->>B: Answer Q1, Q2, Q3
    Note over B: [ 1, 2, 3 ]
    U->>B: Click "Back"
    Note over B: [ 1, 2 ] <br/> 3 is dropped
    U->>B: Answer Q3 again
    Note over B: [ 1, 2, 3 ] <br/> Seq 3 is restored
```

**Navigating via "change"**
```mermaid
sequenceDiagram
    autonumber
    participant U as User
    participant B as Breadcrumbs

    Note over B: Current: [ 1, 2, 3 ]
    U->>B: Change Q2
    Note over B: [ 1 ] <br/> 2 & 3 are dropped
    U->>B: Submit new Q2
    Note over B: [ 1, 2* ] <br/> New data added to Seq 2
    U->>B: Re-answer Q3
    Note over B: [ 1, 2*, 3 ] <br/> Seq 3 added, data restored from cache
```

### Next steps...
I'll write these up in a doc / ticket / ADR in more detail, but I think these are the steps we would want to take next - 
- Update types in `planx-core`
- Update breadcrumb sorting functions to use `seq` instead of sorting by flow
  - Test old/new results for both order matching, and performance
  
Outside the functional tasks here, it's worth taking a broader look at data types here more generally at the same time - should this be a `Record<NodeId, UserData>` or is there a better format? Should we wrap it up in class so we can easily express it with suitable methods when needed?

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/preview/hasPaid.test.ts
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/preview/hasPaid.test.ts
@@ -32,8 +32,17 @@ test("hasPaid is updated if a Pay component has been recorded", () => {
 
   record("c", {});
   expect(getState().breadcrumbs).toEqual({
-    a: { answers: ["c"], auto: false },
-    c: { auto: false },
+    a: {
+      answers: ["c"], 
+      auto: false, 
+      createdAt: expect.any(String),
+      seq: expect.any(Number)
+    },
+    c: {
+      auto: false, 
+      createdAt: expect.any(String),
+      seq: expect.any(Number)
+    },
   });
 
   expect(hasPaid()).toBe(true);
--- a/apps/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/preview/record.test.ts
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/preview/record.test.ts
@@ -58,8 +58,18 @@ test("record(id, undefined) clears up breadcrumbs", () => {
   record("a", { answers: ["c"] });
   record("d", { answers: ["e", "f"] });
   expect(getState().breadcrumbs).toEqual({
-    a: { answers: ["c"], auto: false },
-    d: { answers: ["e", "f"], auto: false },
+    a: {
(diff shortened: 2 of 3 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/lib/store/index.ts
```diff
@@ -30,7 +30,10 @@ export declare namespace Store {
     data?: Record<string, any>;
     auto?: boolean;
     override?: Record<string, any>;
+    createdAt?: string;
+    seq?: number;
   };
+  // TODO: Wrap UserData in Required<T>
```
DafyddLlyr: Intentional TODO I'll look at in future (outside the scope of this PR).

`Breadcrumbs` will always contain a `createdAt` and `seq`, but `UserData` is used more broadly across the application (it exists as a partial object, before these fields are added).

Rather than loosely type these as optional fields, it's probably better to express them as they would be interacted with within PlanX.

PR #5936, opened 2025-12-18
chore(editor): Update labels for `<RuleBuilder/>` within responsive options

DafyddLlyr:
## What does this PR do?
 - Updates the labels for the `<RuleBuilder/>` used in "responsive" components. This was implemented in https://github.com/theopensystemslab/planx-new/pull/5641 and used in Storybook, but the never applied to the actual components 🙃 
 - Fixed an indexing issue with Storybook - `ResponsiveChecklist` and `ResponsiveQuestion` are not displaying on staging and prod, but certainly were when they were written.


apps/editor.planx.uk/.storybook/main.ts
```diff
@@ -1,5 +1,8 @@
 module.exports = {
-  stories: ["../src/**/*.stories.@(js|jsx|ts|tsx)"],
+  stories: [
+    "../src/**/*.stories.@(js|jsx|ts|tsx)",
+    "../src/@planx/**/*.stories.@(js|jsx|ts|tsx)"
+  ],
```
DafyddLlyr: These two globs should be identical, but I need to add this to get the other stories to show 🤷‍♂️

PR #5941, opened 2025-12-19
feat: Output images in grouped checklists

ianjon3s:
## What does this PR do?

- Updates expandable checklists to output images
- Also adds image output to "exclusive" option

**Testing:**
https://5941.planx.pizza/a-new-team/checklists/preview

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/shared/BaseChecklist/Public/components/ExclusiveChecklistItem.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/shared/BaseChecklist/Public/components/ExclusiveChecklistItem.tsx
@@ -3,6 +3,7 @@ import Grid from "@mui/material/Grid";
 import Typography from "@mui/material/Typography";
 import { visuallyHidden } from "@mui/utils";
 import { Option } from "@planx/components/Option/model";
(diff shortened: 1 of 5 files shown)


DafyddLlyr (changes requested):
This does work as expected, but is a little bit of an anti-pattern I think. We're prop-drilling `hasImages` into multiple components when we could reach for `getLayout()` as we always have the associated `option` or `groupedOptions`. There's a potential here for bugs as we have two (possibly diverging) sources of truth - how should a component render where `hasImages={true}` but there are no `img` properties, or vice-versa?

Our current layouts consist of - 

https://github.com/theopensystemslab/planx-new/blob/893bdf6c9893def9489c39be827506c1870bb1ee/apps/editor.planx.uk/src/%40planx/components/shared/BaseChecklist/model.ts#L26-L30

This PR would mean that this list is no longer a discrete description of the possible layouts - we can have grouped layouts with and without images.

I'd recommend taking a look at a solution which introduces a new layout, and updates the logic of `getLayout()` to reflect this.

ianjon3s:
Thanks for the feedback @DafyddLlyr, I have introduced a new layout `GroupedImages`, ensuring `getLayout()` works from a single source of truth for layout options

PR #5943, opened 2025-12-19
chore: update latest records and sync `published_flows.service_charge_enabled`

jessicamcinchak:
Follows on from #5932


apps/hasura.planx.uk/migrations/default/1766154999152_update_published_flows_service_charge_enabled/up.sql
```diff
@@ -0,0 +1,8 @@
+UPDATE published_flows pf
+SET service_charge_enabled = jsonb_path_exists(data, '$.* ? (@.data.applyServiceCharge == true)')
+FROM (
+    SELECT DISTINCT ON (flow_id) id
+    FROM published_flows
+    ORDER BY flow_id, created_at DESC
+) AS latest
+WHERE pf.id = latest.id;
```
jessicamcinchak: Locally, this snippet returns 0 records "before" `UPDATE` and 15 records "after" for me :heavy_check_mark:  
```
select count(id) from published_flows where service_charge_enabled = true;
```

PR #5945, opened 2025-12-29
feat: show analytics at the team level

zz-hh-aa:
This PR checks if a team has any `online` services and, if so, renders analytics in `teamLayoutRoutes`.
___

**Local testing**
- Change `production` to `development` in line 42 `apps/editor.planx.uk/src/pages/FlowEditor/lib/store/settings.ts` and analytics should be active on the team page for teams that have _any_ online service


apps/editor.planx.uk/src/pages/FlowEditor/lib/store/settings.ts
```diff
@@ -122,7 +127,7 @@ export const settingsStore: StateCreator<
 
     const analyticsLink =
       environment === "production" && dashboardId
-        ? generateAnalyticsLink({
+        ? generateFlowAnalyticsLink({
```
zz-hh-aa: Renamed this since we have two types of analytics links now.

PR #5961, opened 2026-01-05
fix(ci/e2e): Improve Hasura startup reliability (+ logging)

DafyddLlyr:
## What's the problem?
Many PRs fail at the integrations test, e2e test, or Vultr create steps. This is happening as when `start-containers-for-tests.sh` runs, Hasura fails as unhealthy. On a retry this will sometimes pass, and sometimes fail. This issue can be recreated locally by running the following from the root dir - 

```bash
# Clean slate
pnpm destroy

# Run test script
pnpm tests
```

Locally, this gives me the same inconsistent results - sometimes it spins up without issues, sometimes it fails.

## What's the cause?
From the logs it appears to be an issue with migrations timing out - we now have more and more migrations to run, and it's possible that the GHA runner / Vultr instance is struggling to complete these within the given time. This isn't an issue on staging and prod as we're always just running the few most recent migrations at a time, not all at once when building the app from scratch.

## What's the solution?
 - Bump timeouts for startup and healthcheck
 - Save logs locally on failure (currently only API logs are retained - this makes debugging Hasura failures a frustrating timing-based action when the container is spinning up)
   - Logs can also be accessed via Pizzas now
   - We could work to expose these as artifacts in GHA for failed integration or e2e test steps if it seems worthwhile
 - Add explicit `public` prefixes to views - the postgres container on `main` shows that this is a problem on startup - possibly causing further issues with timeouts / re-runs? We've seen this before as an intermittent problem - something to bare in mind for future PR reviews of Hasura migrations.
 
## To test...
I can now locally run `pnpm destroy` and `pnpm tests` back-to-back multiple times without hitting health issues with Hasura. Hopefully the same will be true on this PR...!

Regression and integreation tests all passing on this branch ✅ https://github.com/theopensystemslab/planx-new/actions/runs/20785199932/job/59692664136

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -73,7 +73,7 @@ services:
     restart: unless-stopped
     environment:
       HASURA_GRAPHQL_MIGRATIONS_SERVER_TIMEOUT: 90
-      HASURA_GRAPHQL_NO_OF_RETRIES: 20
+      HASURA_GRAPHQL_NO_OF_RETRIES: 15
       HASURA_GRAPHQL_ADMIN_SECRET: ${HASURA_GRAPHQL_ADMIN_SECRET}
       HASURA_GRAPHQL_CORS_DOMAIN: http://localhost:${API_PORT}, http://localhost:${HASURA_CONSOLE_PORT}, http://localhost:3000, http://api:${API_PORT}, http://localhost:${HASURA_PROXY_PORT}/v1/graphql, ${LPS_URL_EXT}, https://*.planx.uk, https://*.planx.dev, https://*.planx.pizza, https://*.gov.uk, https://planx-website.webflow.io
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -63,14 +63,17 @@ services:
...
+      timeout: 5s
+      retries: 15
+      start_period: 60s
     volumes:
       - "./apps/hasura.planx.uk/metadata:/hasura-metadata"
       - "./apps/hasura.planx.uk/migrations:/hasura-migrations"
     restart: unless-stopped
     environment:
+      HASURA_GRAPHQL_MIGRATIONS_SERVER_TIMEOUT: 90
+      HASURA_GRAPHQL_NO_OF_RETRIES: 20
```
jessicamcinchak: You've bumped `retries: 15` above, but `20` here - what exactly is the difference? Should they stay in sync?

PR #5964, opened 2026-01-06
fix(editor): Add Node events polyfill for ShareDB

DafyddLlyr:
## What's the problem?
The frontend build succeeds, but the app fails with the following error - 

image

This is present in both the built code, as well as when running in dev mode. 

## What's the cause?
In dev mode the error points to ShareDB, specifically this issue - https://github.com/share/sharedb/pull/598#issuecomment-1451557704

```
browser-external:events:9 Module "events" has been externalized for browser compatibility. Cannot access "events.EventEmitter" in client code. See http://vitejs.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details.
```

The underlying issue here is that ShareDB is relying on Node polyfills being provided (see https://github.com/share/sharedb/issues/255).

## What's the solution?
[The ShareDB docs](https://github.com/curran/sharedb/blob/36513407b84e195fd81e93109ac4e707b1230cf5/examples/counter-json1-vite/main.js#L3) suggest using [sharedb-client-browser](https://www.npmjs.com/package/sharedb-client-browser), but I'm keen to not add another dependency here.

Instead, we're handling the polyfills directly ourselves via [vite-plugin-node-polyfills](https://www.npmjs.com/package/vite-plugin-node-polyfills).

Longer term, this might be another reason to consider alternatives to ShareDB (spike already on roadmap).

### DIFF
--- a/apps/editor.planx.uk/package.json
+++ b/apps/editor.planx.uk/package.json
@@ -152,6 +152,7 @@
     "stream-browserify": "^3.0.0",
     "tsconfig-paths-webpack-plugin": "^4.0.1",
     "typescript": "^5.6.2",
+    "vite-plugin-node-polyfills": "^0.24.0",
     "vite-tsconfig-paths": "^5.1.4",
     "vitest": "^3.1.2",
     "vitest-axe": "1.0.0-pre.3"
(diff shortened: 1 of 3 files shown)


zz-hh-aa (approved):
Working locally--functional approval from me here! Thanks for quick fix :)

PR #5971, opened 2026-01-07
fix: update `email_events_log` view to correctly capture all emails

jessicamcinchak:
A few fixes following on from this help-issues report: https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1767015199631409

Our prior view definition (used for internal auditing/trouble-shooting only) was failing to account for:
- Emails can be triggered via scheduled events (most email types) _and_ database triggers (submission confirmation emails). This requires querying distinct tables in the `hdb_catalog` schema
- Invite to pay emails do _not_ have a `sessionId` in the payload or comment, but rather a `paymentRequestId` which needs to be handled/joined differently 

Should now work correctly! Once on prod, we should see 3 rows returned for original session (`9ad0dcce-f4de-4693-99f4-2576a4b0b396`) reported by Nora (invitation to pay, invitation to pay agent, and submission confirmation - all "successful" on our side - rather than 0 rows returned by current prod view definition).

### DIFF
--- a/apps/hasura.planx.uk/migrations/default/1767804023822_drop_view_public_email_events_log/down.sql
+++ b/apps/hasura.planx.uk/migrations/default/1767804023822_drop_view_public_email_events_log/down.sql
@@ -1,0 +1,53 @@
+-- OLD SQL does not properly account for ITP (joins on payment request id, not session id) and confirmation emails (db triggered, not scheduled)
+CREATE OR REPLACE VIEW public.email_events_log AS 
+WITH retries AS (
+   SELECT hdb_scheduled_event_invocation_logs.id
+   FROM hdb_catalog.hdb_scheduled_event_invocation_logs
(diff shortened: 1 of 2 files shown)


apps/hasura.planx.uk/migrations/default/1767804023822_drop_view_public_email_events_log/down.sql
```diff
@@ -0,0 +1,53 @@
+-- OLD SQL does not properly account for ITP (joins on payment request id, not session id) and confirmation emails (db triggered, not scheduled)
+CREATE OR REPLACE VIEW public.email_events_log AS 
+WITH retries AS (
```
jessicamcinchak: I've omitted this "retries" logic (and the whole join to `_invocation_logs`) from the new view as I don't think we need this complexity? We can still always manually check for certain cases if necessary!

PR #5969, opened 2026-01-07
feat: manage multiple submission emails with table

zz-hh-aa:
This PR replaces `ListManager` in the multiple submission emails settings form with a table, and also adds delete functionality to the UI. 

The previous implementation was proving to be quite complicated, since we were trying to shoehorn new use cases into components that weren't built for them. Using a table, like in the 'team members' section, simplifies the UI and the code.

# Testing
- Visit https://5969.planx.pizza/{TEAM-SLUG}/settings/integrations
- Run `window.featureFlags.toggle("TEAM_SUBMISSION_INTEGRATIONS")` in the console and refresh
- Try creating, updating and deleting emails. You can check the Hasura `submission_integrations` table to confirm that things are reading / writing as expected

# Questions
1. This is a first proof-of-concept version, with a lot of code copy-pasted from `apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx`. Is there enough similarity to extract this into a generic component that can be customised for both `MembersTable` and `EmailsTable` usages? Or are they different enough that having separate, similar components work?

The modals and columns are obviously different, but the biggest difference is the default email radio selector + logic. 

2. I think the pattern for ensuring a user does not delete the default email is a little awkward. Any suggestions here? 

https://github.com/user-attachments/assets/29f27f9b-d5a2-4a62-a341-6afdc439f9f3

PR #5977, opened 2026-01-09
feat: manage multiple submissions from the send component

zz-hh-aa:
This is the first part of managing multiple submission emails from the Send component, featuring a UI for selecting from the list of available integrations. (Writing a new one from Send will come next). 

# Testing
1. Pick a team and go to Hasura to manually add new emails to the `submission_integrations` table, with the relevant team ID. (This needs to be manual here because #5969 hasn't merged yet). 
2. In the editor, run `window.featureFlags.toggle("MULTIPLE_SUBMISSION_SEND_COMPONENT")`
3. Go to a Send component and try toggling 'Send to Email' on and off, changing the selected email, etc. When 'Send to Email' is toggled on, the default email for that team should be selected. 

# Questions
1. I tried to refactor this using the patterns shown [here](https://github.com/theopensystemslab/planx-new/blob/main/doc/guides/data-fetching-and-state-management.md) (separating queries, hooks and types) but am not sure I've done everything right, especially since this shares some code with the corresponding Settings page (see #5969 again). 
2. A strange bug appears when creating a new Send component in a flow that doesn't have one yet. It flickers for just a second and I haven't been able to figure out where it's coming from. Thoughts? 
Screenshot 2026-01-09 173422

It flickers around 4 seconds in:
https://github.com/user-attachments/assets/8431c6bc-5c74-4c5b-8f36-8943753579e5


apps/editor.planx.uk/src/@planx/components/Send/Editor.tsx
```diff
@@ -40,11 +50,67 @@ const SendComponent: React.FC<Props> = (props) => {
...
+            flowId: id,
+            emailId: defaultEmail.id,
+            teamId: teamId,
+          },
+        });
+      }
+    };
+
+    populateDefaultEmail();
+  }, [emailId, defaultEmail, id, teamId, insertFlowIntegration]);
```
DafyddLlyr: What's this doing - it looks like it's a workaround for a data migration maybe?

apps/editor.planx.uk/src/@planx/components/Send/Editor.tsx
```diff
@@ -40,11 +50,67 @@ const SendComponent: React.FC<Props> = (props) => {
...
+
+  useEffect(() => {
+    if (!formik.values.submissionEmail && defaultEmail) {
+      formik.setFieldValue("submissionEmail", defaultEmail.id);
+    }
+  }, [formik, defaultEmail]);
```
DafyddLlyr: I _think_ this will work exactly the same without the `useEffect()` wrapper 👍

apps/editor.planx.uk/src/@planx/components/Send/Editor.tsx
```diff
@@ -40,11 +50,67 @@ const SendComponent: React.FC<Props> = (props) => {
...
+  const handleChangeEmail = async (newEmailId: string) => {
+    if (emailId) {
+      await updateFlowIntegration({
+        variables: {
+          flowId: id,
+          emailId: newEmailId,
+        },
+      });
+    }
+  };
```
DafyddLlyr: I'd have a think about when this should trigger.

Triggering on "change" event means a user could change it, close the modal without saving, and expect that nothing has happened. The warning on close also does not revert this field currently.

The simpler approach here is to probably just trigger this on save 👍

DafyddLlyr (changes requested):
This is great! A few small questions / comments but this is working as expected here 👌

apps/editor.planx.uk/src/@planx/components/Send/Editor.tsx
DafyddLlyr: 1. Yes, I think this is simpler and more robust. Send components can come and go from flows, and tracking this will get complex pretty fast.
2. Yep, resolving in another PR is appropriate
3. Deletion is complex, and I don't think we've worked this through perfectly yet
  - If I delete a `flow_integrations` record (or just set `email_id = null`) then any _published_ submissions may fail immediately
  - We probably want to have a `published_flows.submission_email_id` so that there's a persistent record attached to a published flow at any given time
  - Let's talk this one through if unclear...!
4. I'm not sure if a trigger is necessary, a SQL unique constraint can cover many columns

DafyddLlyr (approved):
Great! Happy to address the next steps as new PRs.

PR #5978, opened 2026-01-09
chore: bump `@opensystemslab/map` to improve sorting of `address-autocomplete` options

jessicamcinchak:
See https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1767821240370329

Regression tests passed against this branch :white_check_mark: https://github.com/theopensystemslab/planx-new/actions/runs/20854511977


DafyddLlyr (approved):
Could you also please bump the version used in `apps/localplanning.services/package.json` so that we can resolve https://github.com/theopensystemslab/planx-new/security/dependabot/370 ?

PR #5999, opened 2026-01-16
feat: ensure that the first email per-team is always default = true

zz-hh-aa:
Tiny PR for the last outstanding task from #5969 (see [discussion on managing defaults](https://github.com/theopensystemslab/planx-new/pull/5969#discussion_r2694477555)). 

I did this in the frontend because the logic was simple & the UI seemed clear, but let me know if it would be more suited to another DB trigger?


DafyddLlyr (approved):
Great!

PR #6010, opened 2026-01-21
feat: search for an address to position the map when proposing a new address

jessicamcinchak:


jessicamcinchak:
Very outdated, closing in favor of #6455

PR #6015, opened 2026-01-21
fix: Custom text input length

ianjon3s:
## What does this PR do?

- Addresses 2 issues with the custom text input length:

### Layout issue

- As the custom text input appears outside of the `<fieldset>`, the layout is incorrect

### Structural issue

- The custom input was appended to the options, rather than inserted. This means that it's placement does not follow the logical option structure, and needed to be placed at the end

### Before changes:
https://editor.planx.dev/testing/text-input/nodes/_root/nodes/kFyXF2D0su/edit

image

### After changes:
https://6015.planx.pizza/testing/text/nodes/_root/nodes/pJQA8xBVj2/edit

image


DafyddLlyr (approved):
Great change - thank you!

PR #6017, opened 2026-01-21
feat: one flow_integration per-flow

zz-hh-aa:
This is part of the multiple submission email work, following on from #5977. 

# What does this PR do?
- Migration to update the `submission_integrations` table (eg for any teams that have updated their submission email since the last migration in early November). Since there will be max one per-team, they are all set as `default = true`. 
- Migration ensuring that `flow_integrations.team_id` is not nullable, since every flow should be associated with a team. 
- Migration to update the `flow_integrations` table, creating one entry per-flow. Where a team has a submission email, that `email_id` will show. Where a team does not, `email_id = null`. 
- Update `createFlow()` to ensure that `flow_integration` records are created with flows
- Create a relationship between `flows` and `flow_integrations` in Hasura
- Cascade flow deletions to `flow_integrations` too

# Why? 
Every flow [should have an entry](https://github.com/theopensystemslab/planx-new/pull/5977#discussion_r2686876725) in `flow_integrations` by default. This simplifies how we manage the `flow_integrations` table and follows other patterns (eg one `team` to `team_setting`). 

I did this in this order so that the `flow_integrations` table would be as up to date as possible, and so we won't need to do another migration to create records in `flow_integrations`. Updating the db to match this model also seemed like a prerequisite to finishing up the piece of work involving writing new emails from the send component.

We still need to work out deletion and add submission email records to `published_flows` so we can keep track of where a service's submissions are going at any given time.

### DIFF
--- a/apps/api.planx.uk/helpers.ts
+++ b/apps/api.planx.uk/helpers.ts
@@ -68,12 +68,6 @@ const getFlowData = async (id: string): Promise<GetFlowDataResponse> => {
   return flow;
 };
 
-interface CreateFlowResponse {
-  flow: {
-    id: Flow["id"];
-  };
-}
-
 const getDefaultEmail = async (teamId: number): Promise<string | null> => {
   const { client: $client } = getClient();
 
--- a/apps/api.planx.uk/helpers.ts
+++ b/apps/api.planx.uk/helpers.ts
@@ -74,6 +74,30 @@ interface CreateFlowResponse {
   };
 }
 
+const getDefaultEmail = async (teamId: number): Promise<string | null> => {
+  const { client: $client } = getClient();
+
+  const { submissionIntegrations } = await $client.request<{
+    submissionIntegrations: { id: string }[];
+  }>(
+    gql`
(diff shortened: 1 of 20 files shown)


apps/api.planx.uk/helpers.ts
```diff
@@ -151,6 +151,55 @@ const createFlow = async ({
       },
     );
 
+    // Fetch the default email id from submission_integrations table
+    const { submission_integrations } = await $client.request<{
+      submission_integrations: { id: string }[];
+    }>(
+      gql`
+        query GetDefaultEmail($team_id: Int!) {
```
zz-hh-aa: I followed the existing pattern used elsewhere in this file but given that we are [trying to move away from manual GQL queries](https://github.com/theopensystemslab/planx-new/blob/main/doc/architecture/decisions/0011-standardised-data-fetching-and-state-management.md), is it better to mix patterns in this file...?

DafyddLlyr (changes requested):
One thing to take a look at.

DafyddLlyr (approved):

PR #6038, opened 2026-01-29
Production deploy

DafyddLlyr:


zz-hh-aa:
I checked `analytics_planx_flows` on staging and the view works.

I also tested #6036 and #6017 (creating new flows for teams both with `submission_integrations` and without) and all seems to work as expected!

PR #6039, opened 2026-01-29
feat: write new emails from the Send component

zz-hh-aa:
# What does this PR do?
- Handles writing new emails from the Send component (previously, `flow_integrations` were created when flows were created, but the data model has changed--see #6017--so some hooks are deleted)
- New `useInsertSubmissionIntegration` hook to write new emails to the db
- Send component functionality: conditionally rendered input that writes new email to db
    - Validation (checking for duplicates, empty input, invalid email)
- Placeholder link to Settings page where multiple emails can be managed

# Why?
We want users to be able to write new emails from the Send component without having to navigate all the way back up to Settings and then return. 

# Questions
1. There's quite a bit more complexity here, is there a better way I can refactor?


DafyddLlyr (changes requested):
Hey there - apologies again for the delay in getting back to this PR - taken far longer than it should have sorry.

Functionally, this all seems to work as expected - I've been testing it out locally and not hit any issues 👍 

You asked in the PR description about the level of complexity here, and you're right - there's a lot going on. I see the main issue as being that we're intermingling the lifecycle for the submission emails and send component form submission in a way which leads to some crossed wires. This manifests as - 

 - Some data in formik (such as `newEmail?: string`), some data in plain React state (`isNewEmailSelected`)
 - Additional validation context needed for formik
 - Combined submission (`handleUpdateEmail` nested within `onSubmit`) - what happens if one succeeds and the other fails, for example?

Here's what I'd recommend implementing - 

 - **Make the email addition/selection immediate** - when a user selects or adds an email, update the database right away (similar to how team settings work). Show loading/success/error states so the user knows it's saved.
 - **The Send component form just stores a reference** - the form only needs to save which email ID to use (via `submissionEmailId`). This gets saved when the user clicks "Confirm".

This separation means - 
 - Each piece has clear responsibilities
 - Errors are easier to handle and show to users
 - The form submission becomes simpler
 - Users get immediate feedback when changing email config

Very happy to talk this through at dev call later or pair on a solution here 👍

PR #6041, opened 2026-01-29
spike(editor): Polymorphic routing for custom subdomains

DafyddLlyr:
## What does this PR do?
This is a spike which demonstrates we can achieve the following goals as part of the Tanstack Router migration - 
 - All teams can access their services via the standard routing pattern
 - Teams, with custom subdomains configured, can access their services via the subdomain routing pattern
 - This can be done in a type-safe method
 - This can be done whilst maintaining file-based routing
 
The changes demonstrated in this PR achieves all of the above via the following method - 
 - Split `_public` routes into two pathless layout routes (`_editor.planx.uk` and `_customSubdomains`)
 - Define a single shared loader and validation step which ensures all child routes will always have a `teamSlug` and `flowSlug`, regardless of their access pattern

As a neat bonus (and validation of our choice to change to Tanstack), we can now also drop the `PREVIEW_ONLY_DOMAINS` check - the asynchronous loader function allows us to make a DB query prior to initiating any routes. Please [see here for further context](https://github.com/theopensystemslab/planx-new/blob/nich/tanstack-router-migration/apps/editor.planx.uk/src/utils/routeUtils/utils.ts#L40-L54) on this problem.

Please note - this is just a proof-of-concept. I'm not proposing we merge this PR, I'll close once reviewed and tested and then I'll branch from `main` in order to implement this.

### Proposed route file structure
```
src/
└── routes/
    ├── _authenticated/               <-- Routes for authenticated users (flow editor)
    └── _public/                      <-- Routes for public users (public services)
        ├── -loader.ts                <-- Shared: Handle loader and validation logic once
        │
        ├── _editor[.]planx[.]uk/     <-- Path A: Standard URLs (planx.uk)
        │   └── $team/$flow/
        │       ├── route.tsx         <-- Gateway A: Load team details via $team param
        │       └── $sessionId/
        │           └── hello.tsx     <-- Example child page, with additional params
        │
        └── _customSubdomain/         <-- Path B: Custom subomains (planningservices.my-council.gov.uk)
            └── $flow/
               ├── route.tsx          <-- Gateway B: Load team details via custom subdomain DB lookup
               └── $sessionId/
                   └── hello.tsx      <-- Example child page, with additional params
```

## Migration plan
- [x] Proof of concept - will this work?
- [x] Close this PR, merge https://github.com/theopensystemslab/planx-new/pull/5591
- [ ] Move files to new structure, share loaders and other utils
- [ ] Simplify use of `validateTeamRoute()`
- [ ] Simplify (remove?) use of `useFlowRootPath()`
- [ ] Implement route switching in `<CustomLink/>`
- [ ] Remove `PREVIEW_ONLY_DOMAINS` frontend setup from our onboarding process
- [ ] Documentation / `README.md` explaining routes

### DIFF
--- a/apps/editor.planx.uk/src/routes/_public/-loader.ts
+++ b/apps/editor.planx.uk/src/routes/_public/-loader.ts
@@ -1,0 +1,87 @@
+import { notFound } from "@tanstack/react-router";
+import gql from "graphql-tag";
+import { client } from "lib/graphql";
+
+type LoaderArgs = {
+  params: {
+    /** All _public/ routes will have a $flow param */
+    flow: string;
+    /** $team param is either present, or can be derived from url.hostname (custom subdomain) */
+    team?: string;
+  };
+};
+
+/**
+ * All children of _public using this loader will have a $flowSlug and $teamSlug
+ */
+export type PublicContext = {
+  teamSlug: string;
+  flowSlug: string;
+};
+
+/**
+ * Prevents accessing a different team than the one associated with the custom domain.
+ * e.g. Custom domain is for Southwark but URL is looking for Lambeth
+ * e.g. https://planningservices.southwark.gov.uk/lambeth/some-flow
+ */
+export const validateTeamRoute = async (teamParam?: string) => {
+  const externalTeamName = await getTeamFromDomain(window.location.hostname);
+  if (teamParam && externalTeamName && externalTeamName !== teamParam) {
+    throw notFound();
+  }
+
+  return externalTeamName;
+};
+
+const QUERY_GET_TEAM_BY_DOMAIN = gql`
+  query GetTeamByDomain($domain: String!) {
+    teams(limit: 1, where: { domain: { _eq: $domain } }) {
+      slug
+      id
+    }
+  }
+`;
+
+export const getTeamFromDomain = async (
+  domain: string,
+): Promise<string | undefined> => {
+  const {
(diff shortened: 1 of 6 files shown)


apps/editor.planx.uk/src/routes/_public/-loader.ts
DafyddLlyr: As this is a proof-of-concept, I've just copied these utilities across here. When working on this for real, I'll tidy these up.

jessicamcinchak (commented):
Looked through all files here, but have _not_ tested locally yet - sorry! Hit larger issues here like missing AI-related env vars and ended up in a docker destroy/rebuild loop failing to pick up changes.

Now that we're reading from `teams.domain`, I _thought_ I'd be able to actually test on the  pizza by:
- Setting `teams.domain` to '6041.planx.pizza` in this pizza's Hasura console
- But https://6041.planx.pizza/doncaster/pre-application-advice/published does _not_ resolve to https://6041.planx.pizza/pre-application-adivce as I thought it might (it tacks on sort params like a team page, but does not throw any type of team-not-found error)
- Is this just a 'known' not possible? Should it be throwing an error or falling back to a 404?

nichgalzin (commented):
I think this all looks really clear, and works well. And visually in the file structure tells the story well of how the routes are separated. No more notes I can think of really!

PR #6043, opened 2026-01-30
fix(editor): Handle expired tokens and infinite redirect

DafyddLlyr:
## What's the problem?
Once a user's JWT has expired on https://github.com/theopensystemslab/planx-new/pull/5591, they are sent into a logout "doom loop". The page keeps refreshing without completing the logout operation.

https://github.com/user-attachments/assets/db2710cb-3f25-488d-975b-d769523a87a8

## What's the cause?
Right now the following is happening - 
 - A GraphQL request is made
 - Auth fails due to expired JWT
 - Apollo catches this, throws the toast
 - Toast sets route via `window.location.href`, not using router
 - Full page refresh triggered (white screen, all assets re-loading)
 - User lands on /logout, redirected to /login, redirected to /logout... 🌀 
   - This is ultimately happening as toast keeps getting re-triggered

## What's the solution?
I've put in place a few fixes here which resolve this.

1. Don't force redirect to `/login` on error

2. Handle all logout action on the `/logout` route
Currently this behaves differently if forced to logout, or if you logout manually. Keeping this logic all in one place is consistent and simpler.

3. Use `router.navigate()` instead of setting `window.location.href`
This means that we avoid a full refresh, meaning logging out it now much faster - no flash of white, and a bunch of requests for assets.

Locally you can complete the following actions - 

**1. Manual logout**
 - Login
 - Logout via menu
   - This is fast/instant, no flash of white page and full reload
   - JWT cookie is cleared (check dev tools > Application > Cookies > http://localhost:3000
 - Confirm you're actually logged out - if you try navigating to e.g. http://localhost:3000/admin-panel you'll just be sent back to the login page

**2. Forced logout**
 - Login
 - Wait 1 minute for JWT to expire
 - Try to make a request (e.g. load a team)
 - Expired session toast appears
   - Takes full 4 seconds for progress bar to complete, does not stop halfway like the video above
   - You're redirected to /logout
   - This is fast/instant, no flash of white page and full reload
   - JWT cookie is cleared (check dev tools > Application > Cookies > http://localhost:3000 
 - Confirm you're actually logged out - if you try navigating to e.g. http://localhost:3000/admin-panel you'll just be sent back to the login page

---

E2E tests here are failing in line with the parent branch (please see https://github.com/theopensystemslab/planx-new/actions/runs/21476653809/job/62185246192?pr=5591)

### DIFF
--- a/apps/editor.planx.uk/src/index.tsx
+++ b/apps/editor.planx.uk/src/index.tsx
@@ -7,23 +7,19 @@ import CssBaseline from "@mui/material/CssBaseline";
 import { StyledEngineProvider, ThemeProvider } from "@mui/material/styles";
 import { MyMap } from "@opensystemslab/map";
 import { QueryClientProvider } from "@tanstack/react-query";
-import {
-  createRouter,
-  RouterProvider as TanStackRouterProvider,
-} from "@tanstack/react-router";
+import { RouterProvider as TanStackRouterProvider } from "@tanstack/react-router";
 import { ToastContextProvider } from "contexts/ToastContext";
-import { getCookie, setCookie } from "lib/cookie";
 import { initFeatureFlags } from "lib/featureFlags";
 import { queryClient } from "lib/queryClient";
 import { AnalyticsProvider } from "pages/FlowEditor/lib/analytics/provider";
 import React, { Suspense, useEffect } from "react";
 import { createRoot } from "react-dom/client";
 import { ToastContainer } from "react-toastify";
+import { router } from "router";
 
 // init airbrake before everything else
 import * as airbrake from "./airbrake";
 import { client } from "./lib/graphql";
-import { routeTree } from "./routeTree.gen";
 import { defaultTheme } from "./theme";
 
 if (import.meta.env.VITE_APP_ENV !== "production") {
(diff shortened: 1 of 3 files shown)


apps/editor.planx.uk/src/routes/(auth)/logout.tsx
```diff
@@ -32,11 +23,8 @@ export const Route = createFileRoute("/(auth)/logout")({
     }
 
     // Clear cookies
-    const cookieString = `auth=; jwt=; expires=Thu, 01 Jan 1970 00:00:00 UTC; path=/;`;
```
DafyddLlyr: 🐛 It turns out that this doesn't actually work, upon logout the `jwt` cookie still persists (try it on pizza/staging/production).

> You can also assign to this property a string of the form "key=value", specifying the cookie to set/update. **Note that you can only set/update a single cookie at a time using this method.**

Source: https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie

There's no real risk here as the token should be expired or revoked, and we also rely on the `auth` token, but it's better to update this now I've spotted it and to ensure that the intended removal is carried out.

PR #6044, opened 2026-01-30
feat: Use radio pattern with recommended tag for AI descriptions

ianjon3s:
## What does this PR do?

- Updates AI enhanced project description output to use radio groups, with `recommended` prop to highlight AI generated result

image

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/Tasks/ProjectDescription.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/Tasks/ProjectDescription.tsx
@@ -26,28 +26,22 @@ import { CharacterCounter } from "ui/shared/CharacterCounter";
 import ErrorWrapper from "ui/shared/ErrorWrapper";
 import Input from "ui/shared/Input/Input";
 import InputRow from "ui/shared/InputRow";
+import ProgressiveLoading from "ui/shared/ProgressiveLoading";
 import ReactMarkdownOrHtml from "ui/shared/ReactMarkdownOrHtml/ReactMarkdownOrHtml";
 
 import { HOW_DOES_THIS_WORK } from "../../content";
 import type { EnhancedTextInputForTask } from "../../types";
+import type { TaskAction } from "../../types";
 import type { FormValues } from "../types";
 import ErrorCard from "./ErrorCard";
-import ProgressiveLoading from "ui/shared/ProgressiveLoading";
-import type { TaskAction } from "../../types"; 
-
 import {
-  StyledFormLabel,
-  RecommendedTag,
   QuoteDescription,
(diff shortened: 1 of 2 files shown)


apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/Tasks/ProjectDescription.tsx
```diff
@@ -35,29 +38,116 @@ import LoadingSkeleton from "./LoadingSkeleton";
 
 type Props = PublicProps<EnhancedTextInputForTask<"projectDescription">>;
 
-const Card = styled(Box)(({ theme }) => ({
+type DescriptionOption = "suggested" | "original" | "custom";
```
DafyddLlyr: This looks like a re-definition of `TaskAction` - it's probably simpler to retain a single type for these options.

apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/Tasks/ProjectDescription.tsx
```diff
@@ -35,29 +38,116 @@ import LoadingSkeleton from "./LoadingSkeleton";
...
     state.path,
   ]);
-
-  const { values, handleChange, errors, setFieldValue, setValues } =
+  const { values, errors, setFieldValue, setValues } =
     useFormikContext<FormValues>();
   const [open, setOpen] = useState(false);
+  const [selectedOption, setSelectedOption] =
+    useState<DescriptionOption | null>(null);
+  const [customDescription, setCustomDescription] = useState("");
```
DafyddLlyr: We're now managing state in two ways here - via `useState()` and via formik - I'd recommend sticking to a single source of truth, which should be the form state (formik) in this instance.

I think this will both simplify things a lot here, but it will reduce the chance of bugs if these two states get out of sync.

apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/Tasks/LoadingSkeleton.tsx
```diff
@@ -1,55 +1,102 @@
...
+  "Analysing your project description",
+  "Reviewing structure and tone",
+  "Generating suggested improvements",
+];
+
+const LoadingSkeleton: React.FC = () => {
+  const [visibleStages, setVisibleStages] = useState(1);
+
+  useEffect(() => {
+    if (visibleStages < LOADING_STAGES.length) {
```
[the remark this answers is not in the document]
ianjon3s: This works, just had to update
`isVisible = index < visibleStages;`
to
`isVisible = index <= visibleStages;`
to encompass all of the stages

DafyddLlyr (approved):

PR #6059, opened 2026-02-04
feat: Polymorphic routing for custom subdomains

DafyddLlyr:
## What does this PR do?

This PR aims to solve two problems -

1. Polymorphic routing for `_public/_customDomain` and `_public/_planXDomain`
2. The routing clash between `_authenticated/$team` and `_public/_customDomain/$flow`

I would have loved to have solved these independently in smaller PRs, but this proved quite challenging as they're quite closely linked solutions. Apologies for this PR being larger and more complex that I'd hoped - the diff here is a bit misleading as the majority of the changes come from the codegen file `routeTree.gen.ts`.

This PR is an implementation of the spike https://github.com/theopensystemslab/planx-new/pull/6041, which might be helpful in terms of context which I won't repeat here.

### Problem
Routing has to work consistently, regardless of if a user is on our domain or a custom subdomain. For example, `https://editor.planx.uk/doncaster/published/pages/privacy` has to serve the same content as `https://planningservices.doncaster.uk/pages/privacy` (`$team` is derived from the domain, and the `/published` "mode" is removed).

Please see https://github.com/theopensystemslab/planx-new/pull/6041 for further details.

### Solution
Routing is solved by splitting our `_public` routes into [two pathless layout routes](https://tanstack.com/router/v1/docs/framework/react/routing/file-naming-conventions) to differentiate them - `_customDomain` and `_planXDomain`. These share a single loader which is responsible for ensuring that any child components will always have a consistent `$team` and `$flow` param available via context. These values can then be used by any child `loader` or `beforeLoad` functions. There's certainly still some more scope for rationalising this but I'm keen to not grow this PR any more.

As a result of these split routes, we need to make sure that we always use relative links within these areas of the codebase. Once we're nested at `_customDomain/$flow` and `_planXDomain/$team/$flow/preview`, all other routes are consistent, e.g. `pages/$page`, `view-application` and `pay` routes. In order to manage this consistently I've added the `usePublicRouteContext()` hook which will allow us to always get the current location (as a `from` value) so that we can just link relatively from there. I've highlighted an example of this [in the files diff on this PR here](https://github.com/theopensystemslab/planx-new/pull/6059/changes#r2773909982).

### Problem
Both `_authenticated/$team` and `_public/_customDomain/$flow` ultimately resolve to `__root/$someParam`. As this is a clash, Tanstack Router isn't able to correctly evaluate which route to navigate to, leaving a route stuck in "pending". These a full write up of this issue [here on OSL Slack](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1770224801176889) which I won't repeat..!

### Solution
I've added a level of specificity to all `_authenticated` routes by nesting a level deeper within an `/app` path. This ensures there's no routing clash, and a clearer internal distinction between public and authenticated routes. This change will break any hardcoded links we have in our Notion documentation or elsewhere, as well as any bookmarks which Editors may have in their browsers. This is somewhat mitigated by the need to login when navigating to any of these routes, but worth flagging as a regression here.

#### Naming
`/app` isn't particularly meaningful - I was looking for a short and pretty meaningless param that would be fairly discrete. The character `#` cannot be used, the following error is thrown - `Error: Disallowed character "#" found in square brackets in route path "_authenticated/[#]/$team/design".`

Very open to alternate options here 👍

### DIFF
--- a/apps/editor.planx.uk/src/hooks/usePublicRouteContext.ts
+++ b/apps/editor.planx.uk/src/hooks/usePublicRouteContext.ts
@@ -5,6 +5,7 @@ type PublicRoutePattern =
   | "/$team/$flow/preview"
   | "/$team/$flow/draft"
   | "/$team/$flow/published"
+  | "/$team/$flow"
   | "/$flow";
 
 /**
@@ -32,6 +33,7 @@ export const usePublicRouteContext = (): PublicRoutePattern => {
       if (routeId.includes("/preview")) return "/$team/$flow/preview";
       if (routeId.includes("/draft")) return "/$team/$flow/draft";
       if (routeId.includes("/published")) return "/$team/$flow/published";
+      if (routeId.includes("/pay")) return "/$team/$flow";
     }
 
     throw notFound();
--- a/apps/editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
@@ -92,8 +92,9 @@ const InviteToPayForm: React.FC<InviteToPayFormProps> = ({
   const from = usePublicRouteContext();
 
   const redirectToConfirmationPage = (paymentRequestId: string) => {
+    // TODO: variable on custom subdomains?
     navigate({
-      to: "pay/invite",
+      to: "../pay/invite",
       from,
       search: { paymentRequestId },
     });
--- a/apps/editor.planx.uk/src/pages/Preview/ApplicationViewer.tsx
+++ b/apps/editor.planx.uk/src/pages/Preview/ApplicationViewer.tsx
@@ -26,6 +26,7 @@ const ApplicationViewer: React.FC = () => {
   const { pathname } = useLocation();
 
   const hasPaid = Boolean(govUkPayment);
+  // TODO: check this logic
   const hasSentInviteToPay = pathname?.match("pay/invite");
   const showFeeBreakdown = hasPaid || hasSentInviteToPay;
 
--- a/apps/editor.planx.uk/src/pages/Preview/Resume/ValidateSession.tsx
+++ b/apps/editor.planx.uk/src/pages/Preview/Resume/ValidateSession.tsx
@@ -61,7 +61,7 @@ export const LockedSession: React.FC<{
 
   return (
     <StatusPage
-      bannerHeading="Sorry, you can't make changes to this application"
+      bannerHeading="Sorry, you can't make changes to this form"
       additionalOption="startNewApplication"
     >
       {paymentRequest && (
@@ -73,7 +73,7 @@ export const LockedSession: React.FC<{
           ) to pay and changes might affect the fee.
           <br />
           <br />
-          You can{" "}
+          You can {/* TODO: Check / fix this link */}
           <CustomLink
             to="pay"
             from={from}
@@ -131,7 +131,6 @@ const ValidateSession: React.FC<{
   >({
     mutationFn: validateSession,
     onSuccess: (data) => resumeSession(data.reconciledSessionData),
-    onError: console.debug,
   });
 
   const handleSubmit = useCallback(
--- a/apps/editor.planx.uk/src/pages/layout/OfflineLayout.tsx
+++ b/apps/editor.planx.uk/src/pages/layout/OfflineLayout.tsx
@@ -6,9 +6,11 @@ const OfflineLayout = ({ children }: PropsWithChildren) => {
   const isFlowOnline = useStore.getState().flowStatus === "online";
   const searchParams = new URLSearchParams(window.location.search);
   const isUserResuming = Boolean(searchParams.get("sessionId"));
+  // TODO: Pay should be handled at a different level so we can just fallback to PaymentNotFound page
+  const isUserPaying = Boolean(searchParams.get("paymentRequestId"));
 
   // Allow users to complete Save & Return journeys, even if a flow is offline
-  const isFlowAccessible = isFlowOnline || isUserResuming;
+  const isFlowAccessible = isFlowOnline || isUserResuming || isUserPaying;
 
   return isFlowAccessible ? children : <OfflinePage />;
 };
--- a/apps/editor.planx.uk/src/utils/routeUtils/payQueries.ts
+++ b/apps/editor.planx.uk/src/utils/routeUtils/payQueries.ts
@@ -38,6 +38,7 @@ export const getPaymentRequest = async (
         retentionPeriod: getRetentionPeriod(),
       },
       context: {
+        role: "public",
         headers: {
           "x-hasura-payment-request-id": paymentRequestId,
         },
--- a/e2e/tests/ui-driven/src/invite-to-pay/nominee.spec.ts
+++ b/e2e/tests/ui-driven/src/invite-to-pay/nominee.spec.ts
@@ -51,6 +51,7 @@ test.describe("Nominee journey @regression", async () => {
     await expect(
       page.getByRole("heading", { name: "Pay", exact: true }),
     ).toBeVisible();
+    // TODO(a11y): Have we lost the `<main/>` structure?
     await expect(
       page.locator("#main-content").getByText("Invite to pay test"),
     ).toBeVisible();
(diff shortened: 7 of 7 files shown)


apps/editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
```diff
@@ -68,41 +68,42 @@ function EditorNavMenu() {
     }
   };
 
+  // TODO: How can we make this component type-safe?
```
jessicamcinchak: 👍

apps/editor.planx.uk/src/hooks/usePublicRouteContext.ts
```diff
@@ -0,0 +1,41 @@
+import { notFound, useMatches } from "@tanstack/react-router";
+import { useMemo } from "react";
+
+type PublicRoutePattern =
+  | "/$team/$flow/preview"
+  | "/$team/$flow/draft"
+  | "/$team/$flow/published"
+  | "/$team/$flow"
```
jessicamcinchak: In which scenario is `"/$team/$flow"` ever a public route? 

If cases of invite to pay (?), can this be more specific like `"/$team/$flow/pay"` to not be confused with authenticated editor graph route?

apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
```diff
@@ -171,54 +174,61 @@ const Sidebar: React.FC = React.memo(() => {
...
                   color="inherit"
+                  preload={false}
                 >
                   <OpenInNewIcon />
-                </Link>
+                </CustomLink>
               </Tooltip>
 
               {isFlowPublished ? (
+                // TODO: Tooltip + CustomLink doesn't work
```
jessicamcinchak: nit: another possible TODO here: on "Open published flow", conditionally link to `"/$flow"` _if_ a custom domain is configured _and_ environment = production?

apps/editor.planx.uk/src/routes/_public/_customDomain/$flow/pay/invite/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+    };
+  },
+  component: InviteToPayComponent,
+});
+
+function InviteToPayComponent() {
+  const data = Route.useLoaderData() as { paymentRequest: PaymentRequest };
+
+  return <InviteToPay {...data.paymentRequest} />;
+}
```
jessicamcinchak: Continues to be satisfying reading these very tidy, slick new route defintions ➕

apps/editor.planx.uk/src/routes/_public/_customDomain/$flow/pay/view-application.tsx
```diff
@@ -2,7 +2,7 @@ import { createFileRoute } from "@tanstack/react-router";
 import ApplicationViewer from "pages/Preview/ApplicationViewer";
 
 export const Route = createFileRoute(
-  "/_public/$team/$flow/published/view-application",
+  "/_public/_customDomain/$flow/pay/view-application",
```
jessicamcinchak: question: is this actually a correct "replace" here from `/published` to `/pay` ? Should it simply be `"/_public/_customDomain/$flow/view-application"` ?

apps/editor.planx.uk/src/routes/_public/_customDomain/$flow/route.tsx
```diff
@@ -0,0 +1,27 @@
...
+import React from "react";
+import {
+  createPublicRouteBeforeLoad,
+  createPublicRouteErrorComponent,
+  createPublicRouteHead,
+  PublicRouteLayout,
+  publicRouteSearchSchemas,
+} from "utils/routeUtils/publicRouteHelpers";
+
+// TODO: How much of this can live once at a higher level?
```
jessicamcinchak: 👍

apps/editor.planx.uk/src/routes/_public/_planXDomain/$team/$flow/draft/index.tsx
```diff
@@ -2,7 +2,9 @@ import { createFileRoute } from "@tanstack/react-router";
 import Questions from "pages/Preview/Questions";
 import React from "react";
 
-export const Route = createFileRoute("/_public/$team/$flow/draft/")({
+export const Route = createFileRoute(
+  "/_public/_planXDomain/$team/$flow/draft/",
```
jessicamcinchak: nit: trailing `/`in `draft/` (might mess with future find+replaces!)

apps/editor.planx.uk/src/routes/_public/_planXDomain/$team/$flow/pay/invite/index.tsx
```diff
@@ -6,22 +6,22 @@ import React from "react";
...
 import { z } from "zod";
 
-// Search schema for invite route
 const inviteSearchSchema = z.object({
   paymentRequestId: z.string().uuid(),
 });
 
-export const Route = createFileRoute("/_public/$team/$flow/pay/invite/")({
+export const Route = createFileRoute(
+  "/_public/_planXDomain/$team/$flow/pay/invite/",
```
jessicamcinchak: nit: another trailing `/` here

apps/editor.planx.uk/src/routes/_public/-loader.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  // Type-safety only - resolver should not reach here without a $team param
+  if (!params.team) throw notFound();
+
+  return {
+    team: params.team,
+    flow: params.flow,
+  };
+};
```
jessicamcinchak: 🌟

apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/index.tsx
DafyddLlyr: Yep, agreed - nice way of integrating custom subdomains a little more! 

Tickets added for both here - 
 - https://github.com/orgs/theopensystemslab/projects/8/views/1?pane=issue&itemId=155538647
 - https://github.com/orgs/theopensystemslab/projects/8/views/1?pane=issue&itemId=155539003

apps/editor.planx.uk/src/routes/_public/_customDomain/$flow/pay/view-application.tsx
DafyddLlyr: This is correct, it's just a weird git diff I think - the url structure matches the file path. 

I would love a simpler solution here but currently `/view-application` always lives at a level deeper than you are, repeated in many places (this is the same as `production` currently).

The reason for this is that there's always an additional `$mode` param on `_planXDomain` routes, and it's linked to from many places, so we can't consistently place it at `../view-application`. I do think there's room to improve and simplify this though - I've added a ticket to the board to take another look at the location of this page - https://github.com/orgs/theopensystemslab/projects/8/views/1?pane=issue&itemId=155547989

apps/editor.planx.uk/src/routes/_public/_planXDomain/$team/$flow/draft/index.tsx
DafyddLlyr: This is correct as it's [an index route](https://tanstack.com/router/v1/docs/framework/react/routing/routing-concepts#index-routes) (it's also auto-generated file and self-correcting).



https://github.com/user-attachments/assets/e6da831f-7992-406a-88bf-9989ffba3088

apps/editor.planx.uk/src/routes/_public/_planXDomain/$team/$flow/pay/invite/index.tsx
DafyddLlyr: Please see above.

apps/editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
DafyddLlyr: Ticket here for this one - https://github.com/orgs/theopensystemslab/projects/8/views/1?pane=issue&itemId=155550308

PR #6077, opened 2026-02-09
feat: add a 'limit to one file' switch to file upload

augustlindemer:
This PR adds a switch to the File Upload component to allow editors to restrict the number of files to one. Default is kept to multiple files, so no content migration needed.


jessicamcinchak (approved):
Works perfectly!

DafyddLlyr (commented):
Thanks for picking this one up 👍

PR #6238, opened 2026-02-28
address dependabot alerts

freemvmt:
Vercel's `serve` package, which we use in the `ui-driven` E2E tests, introduces [three](https://github.com/theopensystemslab/planx-new/security/dependabot/522) [high](https://github.com/theopensystemslab/planx-new/security/dependabot/550) [severity](https://github.com/theopensystemslab/planx-new/security/dependabot/549) issues due to the `minimatch`, and one [medium](https://github.com/theopensystemslab/planx-new/security/dependabot/515) due to `ajv` (both transitive dependencies).

Here we override these transitive dependencies to ensure we have the latest security fixes.

Any other outstanding [dependabot alerts](https://github.com/theopensystemslab/planx-new/security/dependabot) are currently `low` severity ✅ 

image

### DIFF
--- a/e2e/tests/ui-driven/package.json
+++ b/e2e/tests/ui-driven/package.json
@@ -32,8 +32,9 @@
       "@opensystemslab/planx-core"
     ],
     "overrides": {
-      "on-headers@<1.1.0": ">=1.1.0",
-      "minimatch": "^3.1.5"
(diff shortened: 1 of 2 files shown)


jessicamcinchak (approved):
Good spot, thanks !

PR #6251, opened 2026-03-03
feat: Implement recent flows list using store

ianjon3s:
## What does this PR do?

Implements `RecentFlows` component using a Zustand store to collect a list of flows traversed in a session.

- Recent flows list be hidden when directly accessing a flow from the team page or direct URL
- The list should populate a single flow when navigating into a nested flow
- The list should update the displayed "back to" link when navigating into subsequent nested flows
- When expanded the list should display a full trail back to the origin flow
- The list should respect native back/forward navigation in/out of nested flows

### Simplified flow for testing

https://6251.planx.pizza/app/testing/nested-flows

### Real world flow for testing

https://6251.planx.pizza/app/barnet/find-out-if-you-need-planning-permission
- Navigate into the flow `opensystemslab / permitteddevelopment`
- This flow contains several more nested flows, including `change-the-use-of-a-property-in-the-future` which has subsequent nested flows

### DIFF
--- a/apps/editor.planx.uk/src/components/Breadcrumbs.tsx
+++ b/apps/editor.planx.uk/src/components/Breadcrumbs.tsx
@@ -60,6 +60,7 @@ const Breadcrumbs: React.FC<BreadcrumbsProps> = ({
               }}
               {...(isStandalone && { target: "_blank" })}
               variant="body1"
+              preload={false}
             >
               {params.team}
             </BreadcrumbsLink>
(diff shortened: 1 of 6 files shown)


DafyddLlyr (changes requested):
It's taken me a while to get back to this one sorry!

It's a tricky one for sure - and more complex than expected to track the URL and then filter and display the required data.

I'd recommend trying to split this up - track the URLs, and then a pure function for the business logic (with tests) that can take this raw data and output the required list of flows. As discussed at dev call, I'm also very happy to have a crack at this one - just let me know!

PR #6305, opened 2026-03-10
chore: Remove old download application endpoint and UI

DafyddLlyr:
## What does this PR do?
 - Removes the `/download-application-files/{sessionId}` API endpoint
 - Replaces the `"/_public/_planXDomain/$team/$flow/$sessionId/download-application"` route with a simple error message

## Next steps...
Trigger the creation of a token, and associated email, when this page is hit.

## Sequencing
Ideally this will never make it to `production` - it's not a great user-facing solution. We can safely merge to `main` but I'd be very keen to get a more robust solution in place before it gets further.


jessicamcinchak (approved):
Approving for temporary merge to `main` !

PR #6310, opened 2026-03-11
feat: enable submission email deletion

zz-hh-aa:
Re-adds the functionality from #6056 but with the updated data model, as discussed at dev call this week. 

Testing
- Add emails to the Team Settings -> Integrations -> Emails Table
- Delete non-default email
- Go to a submission service for that team and select one of your new emails as the send to email destination in the Send component
- Try to delete the email, you should see a link to the flow in which the email is used and a disabled submit button

Screenshot 2026-03-11 121832
Screenshot 2026-03-11 121802
Screenshot 2026-03-11 121836

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Integrations/SubmissionEmails/EmailsTable.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Integrations/SubmissionEmails/EmailsTable.tsx
@@ -18,7 +18,11 @@ import { StyledTableRow } from "../../../../Team/styles";
 import { EmailsUpsertModal } from "./EmailsUpsertModal";
 import { GET_TEAM_SUBMISSION_INTEGRATIONS } from "./queries";
 import { RemoveEmailModal } from "./RemoveEmailModal";
-import { GetSubmissionEmails, SubmissionEmailInput } from "./types";
+import {
+  GetSubmissionEmails,
+  SubmissionEmailInput,
+  SubmissionEmailWithFlows,
+} from "./types";
 
 const TableRowButton = styled(Button)(({ theme }) => ({
   textDecoration: "underline",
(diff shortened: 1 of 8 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Integrations/SubmissionEmails/RemoveEmailModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+      {
+        variables: { emailId },
+      },
+    );
+
+  const flowIds = flowIdsData?.flowIds.map((flow) => flow.flowId) || [];
+
+  const { data: flowsData, error: flowsError } = useQuery<GetFlows>(GET_FLOWS, {
+    variables: { emailId, flowIds },
+  });
```
DafyddLlyr: We can drop these two waterfalled queries entirely by just enhancing `GetTeamSubmissionIntegrations` to return the flows associated with each integration.

Right now we need to wait for `GET_FLOW_IDS_BY_SUBMISSION_INTEGRATION` to complete, before we can kick off `GET_FLOWS` to get the data we actually need.

```ts
export const GET_TEAM_SUBMISSION_INTEGRATIONS = gql`
  query GetTeamSubmissionIntegrations($teamId: Int!) {
    submissionIntegrations: submission_integrations(
      where: { team_id: { _eq: $teamId } }
      order_by: { submission_email: asc }
    ) {
      teamId: team_id
      id
      submissionEmail: submission_email
      defaultEmail: default_email
      # Query the flows directly via relationship
      flows {
        id
        slug
        name
      }
    }
  }
`;
```

This will just need an additional array relationship set up - 

Image


The main benefit of this would just be simplification and cleaner code - fewer hooks and sequencing, and we should be able to define how we want the data in the query (removing the `usedFlows` map below). This modal wouldn't need to do any data fetching at all.

DafyddLlyr (changes requested):
It's worth taking a look at the setup of the queries and relationships here - we can probably simplify this a lot.

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Integrations/SubmissionEmails/RemoveEmailModal.tsx
zz-hh-aa: Thanks! This took me on a slight refactor journey but I think the types are cleaner for it... most of this done in [50d02a7](https://github.com/theopensystemslab/planx-new/pull/6310/commits/50d02a7c265b123f980c3b30cd7da53958fff090) with some tidying up after

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Integrations/SubmissionEmails/EmailsTable.tsx
```diff
@@ -45,41 +57,60 @@ const StyledTableContainer = styled(TableContainer)(({ theme }) => ({
...
-  const { data, loading } = useQuery<GetSubmissionEmails>(
+  const { data, loading, refetch } = useQuery<GetSubmissionEmails>(
     GET_TEAM_SUBMISSION_INTEGRATIONS,
     {
       variables: { teamId },
     },
   );
 
-  const emails = data?.submissionIntegrations;
+  const submissionIntegrations = data?.submissionIntegrations.map(
```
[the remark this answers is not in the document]
DafyddLlyr: I think the renaming is a good shout - let's keep thing simple for now, and properly evaluate and expand when / if we hit that question.

DafyddLlyr (changes requested):
Big improvement! One further comment on simplification here.

DafyddLlyr (approved):
Fab - much simpler and clearer 🚀

PR #6329, opened 2026-03-15
feat: basic UI for in-editor notifications

jessicamcinchak:
## Editor notifications UI

Initial work carried out by @jessicamcinchak :
- A very draft UI proposal for the functional changes introduced in #6320 & #6330 
- All feature-flagged !
- See #planx-design thread here: https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1773620847075079

Picked up by @ianjon3s :

- Introduce `NotificationCard` component
- Introduce `NoficationPanel` to summarise notifications
- New hook `useNotificationsCount` to get number of active notifications
- New hook `useRecentNotifications` to triage active and resolved notifications
- New hook `useResolveNotification` to resolve notifications

**To test:**

Toggle feature flag
```
```

https://6329.planx.pizza/app/barnet/notifications


DafyddLlyr:
@ianjon3s A follow up to your question at standup - integration test are currently correctly failing here - we're stating that teamEditors cannot query the notifications table, but this PR (see the Hasura YAML permission changes) do grant them access here.

In order to resolve this, you'll need to update the tests to correctly state the new permission model - both `platformAdmin`s and `teamEditor`s can query this table.

You can run these locally (they're super fast) by.. 

```sh
cd apps/hasura.planx.uk/tests
pnpm test
```

ianjon3s:
Thanks for this @DafyddLlyr, will work through updating the tests

apps/editor.planx.uk/src/components/EditorNavMenu/components/NavMenuButton.tsx
```diff
@@ -1,14 +1,15 @@
...
-import { MenuButton, MenuTitle, StyledChip } from "../styles";
+import { BadgeChip, MenuButton, MenuTitle, StyledChip } from "../styles";
 import { Route } from "../types";
 
 export interface NavMenuButtonProps {
   title: string;
   Icon: Route["Icon"];
   disabled?: boolean;
   isNew?: boolean;
+  badge?: number;
```
DafyddLlyr: nit: The prop name and type don't really align that well. Maybe `badgeLabel?: string` or `badgeCount?: number`?

apps/editor.planx.uk/src/hooks/data/useNotificationsCount.ts
DafyddLlyr, line 1: This is great 👍

apps/editor.planx.uk/src/components/EditorNavMenu/components/NavMenuItem.tsx
```diff
@@ -31,7 +32,10 @@ const NavMenuItem = ({
       onClick={onClick}
       sx={{ padding: "8px" }}
     >
-      <Icon />
+      <Box sx={{ display: "flex", flexDirection: "column", alignItems: "center", gap: 0.25 }}>
+        <Icon />
+        {badge ? <BadgeChip label={badge} color="info" sx={{ marginLeft: 0 }} /> : null}
+      </Box>
```
DafyddLlyr: nit: There's a slight layout shift caused by the async loading of the notification count here.

https://github.com/user-attachments/assets/b542af7c-62e8-4eb4-99d3-469e35c5bd60

apps/editor.planx.uk/src/hooks/data/useRecentNotifications.ts
```diff
@@ -0,0 +1,68 @@
...
+      order_by: { created_at: desc }
+    ) {
+      id
+      flow {
+        name
+        slug
+      }
+      team {
+        name
+        slug
```
DafyddLlyr: We should always be returning ids values for GQL queries - this allows the Apollo cache to find items instead of forcing a network request. The same applies for the `resolved` section below.

Docs: https://www.apollographql.com/docs/react/caching/overview

apps/editor.planx.uk/src/hooks/data/useRecentNotifications.ts
```diff
@@ -0,0 +1,68 @@
...
+      }
+      type
+      createdAt: created_at
+      resolvedAt: resolved_at
+    }
```
DafyddLlyr, line 45: Consider the use of a GQL fragment here to keep these two data types in-sync.

Docs: https://www.apollographql.com/docs/react/data/fragments

apps/editor.planx.uk/src/hooks/data/useResolveNotification.ts
DafyddLlyr, line 1: It looks like this hook is never used?

apps/editor.planx.uk/src/lib/featureFlags.ts
```diff
@@ -1,5 +1,5 @@
 // add/edit/remove feature flags in array below
-const AVAILABLE_FEATURE_FLAGS = [] as const;
+const AVAILABLE_FEATURE_FLAGS = ["ARCHIVE_VIEW", "NOTIFICATIONS"] as const;
```
DafyddLlyr: Regression from a rebase. The `ARCHIVE_VIEW` flag was removed here - https://github.com/theopensystemslab/planx-new/pull/6549

apps/editor.planx.uk/src/pages/FlowEditor/components/Notifications/types.ts
```diff
@@ -0,0 +1,22 @@
...
+  resolvedAt: string | null; // nullable timestamp
+}
+
+export interface NotificationProps {
+  notifications: Notification[];
+}
+
+export interface NotificationCardProps {
+  notification: Notification;
+}
```
DafyddLlyr: This is the second `export interface NotificationCardProps` in this PR which could lead to a few issues...
 - Auto-import confusion within IDEs - small fixable issue, but annoying DX problem
 - Declaration merging - this is a much worse, and more invisible problem. Interfaces in TS are open by default, so rather than clashing they merge. Example from docs here - 

```ts
interface Box {
  height: number;
  width: number;
}
interface Box {
  scale: number;
}
let box: Box = { height: 5, width: 6, scale: 10 };
```

Source: https://www.typescriptlang.org/docs/handbook/declaration-merging.html

There's no bug / immediate risk here as this is written but it's the type of issue that hard to spot and debug later - it's better to keep unique names now.

apps/editor.planx.uk/src/pages/FlowEditor/components/Notifications/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (seenFlows.has(notification.flow.slug)) {
+      superseded.push(notification);
+    } else {
+      seenFlows.add(notification.flow.slug);
+      current.push(notification);
+    }
+  }
+
+  return { current, superseded };
+};
```
DafyddLlyr, line 25: This is a fantastic candidate for coverage with unit tests.

apps/editor.planx.uk/src/routes/_authenticated/app/$team/notifications.tsx
```diff
@@ -0,0 +1,56 @@
+import { createFileRoute, notFound } from "@tanstack/react-router";
+import gql from "graphql-tag";
+import { client } from "lib/graphql";
+import { Notifications } from "pages/FlowEditor/components/Notifications/Notifications";
+import { Notification } from "pages/FlowEditor/components/Notifications/types";
+import { useStore } from "pages/FlowEditor/lib/store";
+
+export const Route = createFileRoute("/_authenticated/app/$team/notifications")(
```
DafyddLlyr, line 8: It's been a while since the last discussion we had on the routing setup here and I'm trying to remember exactly what we settled on here.

I _think_ we agreed to move this to `/_authenticated/app/notifications`, and have minimal "team styling" on the notifications.

apps/editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
DafyddLlyr, line 1: Please see previous comment on routing, this could then be moved to be co-located with the `AccountMenu`

DafyddLlyr (changes requested):
A few minor nits and comments, but some outstanding issues around routing we need to consider (I think!).

Currently behind a feature flag which is great, but there's a few requirements here we'll need to follow up with - Storybook coverage and testing. We've been caught out a few times recently by inadequate test coverage on Editor features so I'm quite keen to ensure we have decent coverage here.

apps/editor.planx.uk/src/components/EditorNavMenu/components/NavMenuButton.tsx
ianjon3s: Agreed, I've updated to `badgeCount`

apps/editor.planx.uk/src/components/EditorNavMenu/components/NavMenuItem.tsx
ianjon3s: The menu item for notifications now has a min-height set, which prevents the jump when the badge loads in

apps/editor.planx.uk/src/hooks/data/useRecentNotifications.ts
ianjon3s: Updated to return id for both

apps/editor.planx.uk/src/hooks/data/useResolveNotification.ts
ianjon3s: Indeed, I was prototyping the ability to resolve manually, but as we're doing programatic for now, I've removed

apps/editor.planx.uk/src/hooks/data/useRecentNotifications.ts
ianjon3s: Thanks for this, makes sense to avoid repetition

apps/editor.planx.uk/src/pages/FlowEditor/components/Notifications/types.ts
ianjon3s: It turns out this wasn't referenced anywhere so I've removed from the types file

apps/editor.planx.uk/src/pages/FlowEditor/components/Notifications/utils.ts
ianjon3s: I've added some basic unit tests, with some AI assistance

apps/editor.planx.uk/src/routes/_authenticated/app/$team/notifications.tsx
ianjon3s: For now we agreed to scope each set of notifications to a given team

DafyddLlyr: Thanks for clarifying - my mistake...! 😅

apps/editor.planx.uk/src/pages/FlowEditor/components/Notifications/utils.ts
DafyddLlyr: Perfect!

DafyddLlyr (approved):

ianjon3s:
Thanks for your help with this one @DafyddLlyr 🙏

PR #6355, opened 2026-03-23
feat(wip): allow editors to select which set of Gov Notify templates to use across a service

jessicamcinchak:


jessicamcinchak:
No longer needed. See latest work here https://trello.com/c/IOY0ZHEC/3121-add-a-second-govnotify-template-to-make-language-for-discretionary-services-less-application-planning-focused

PR #6374, opened 2026-03-26
feat: Assign users to existing team

DafyddLlyr:
## What does this PR do?
Allows existing users to be assigned as members of a team.

## Refactors
- I've split the user modals into separate Add / Edit / Remove modals - in reality there's not that much which is shared. The shared fields have been split out into `/Fields`
- Doing this allows us to compose the contents of the `AddUser` modal which now is a multi-step process
  - First, read the email to check if they're already a user
  - If yes, just add a `team_members` record
  - If no, insert both a `users` and `team_members` record
- This check / fetch / insert logic is wrapped up in the `useAddUserModal()` hook


jessicamcinchak (approved):
All looks good pending rebase! Apologies it's taken me a bit to get to this one!

PR #6382, opened 2026-03-30
feat: Analyst user management

DafyddLlyr:
## What does this PR do?
- Adds a "User management" page for users with the `platformAdmin` role
- This has a `<MembersTable/>` which lists users with the `analyst` role
- Analysts can be added, edited, and removed


apps/editor.planx.uk/src/pages/Users/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <>
+          <MembersTable
+            members={activeAnalysts}
+            showEditMemberButton
+            showRemoveMemberButton
+          />
+          <Box m={1}>
+           <AddButton onClick={toggleAddAnalystModal}>Add a new analyst</AddButton>
+          </Box>
+          {showAddAnalystModal && <AddAnalystModal onClose={toggleAddAnalystModal}/>}
```
DafyddLlyr, line 44: **Why add this and not use the `showAddMemberButton` on `<MembersTable/>`?**

This is a bit of a workaround / shortcut to be honest. 

The insert mutation on the members table is adding a record to the `users` table, and another to the `team_members` table. Here, we just want a `users` record with `is_analyst = true`.

As this is a one-off, and `platformAdmin` only feature, I thought it was simpler to add a custom mutation here rather than add another set of conditions to the existing mutations on the `MembersTable` (checking type of request, conditionally adding `team_members` etc).

jessicamcinchak: Thanks for explanation, very fair / pragmatic approach !

jessicamcinchak (approved):

PR #6387, opened 2026-04-01
chore: rename `flows.deleted_at` to `archived_at`

zz-hh-aa:
This is the first step in creating an archived flows view. It:
- renames `flows.deleted_at` to `flows.archived_at` since we wanted to migrate the data to the new column anyway
    - cascades permissions
- updates references to `flows.deleted_at` with the new column name and more accurately renames a few functions to `archiveFlow`
    - updates tests accordingly

I'll create a new, nullable `flows.deleted_at` column in another PR that creates delete functionality too.

# Questions
❓ I've just noticed that on the `flows` table we have custom row permission checks for `archived_at = null`. Does it still make sense for me to remove the permissions here and recreate them for `deleted_at` when I add that column in?


DafyddLlyr (approved):
Great! Just renaming as a first step is great as it solves the migration problem ✅ 

> ❓ _I've just noticed that on the flows table we have custom row permission checks for archived_at = null. Does it still make sense for me to remove the permissions here and recreate them for deleted_at when I add that column in?_

(If I'm understanding correctly!) Yes - we should be robust with the permissions and correctly apply them to `deleted_at` also. We set permissions which mean that users cannot update flows which are set to `archived_at _neq null` - this rule will also apply to `deleted_at` flows. Two main reasons for this - 
 - It's possible in future, somebody will remove the "archived" state - leaving the "deleted" state broken (allowing updates)
 - Authenticated users can interact with PlanX via the API, not just the UI. We need to ensure that a user's JWT cannot invoke mutations which we do not want - setting permissions correctly blocks this via all methods, not just the user interface (i.e. even if not possible to update a deleted flow via the UI, a user could update one via the GQL API if we do not set permissions).

PR #6403, opened 2026-04-03
refactor: Add a second step for modifying project descriptions

ianjon3s:
## Options for accepting or modifying descriptions in second step

- First step prompts for a user to choose between suggested or original
- User is advised `You will be able to modify either of the selected descriptions at the next step.`
- Second step allows user to modify selected description

image

image

image

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/index.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/index.tsx
@@ -18,6 +18,9 @@ const taskComponents: TaskComponentMap = {
 const EnhancedTextInputComponent = (props: Props) => {
   const previous = getPreviouslySubmittedData(props);
   const [step, setStep] = useState<"input" | "task">("input");
+  const [taskSubStep, setTaskSubStep] = useState<"selection" | "modification">(
+    "selection",
+  );
   const isRunningTask = useIsFetching({ queryKey: [props.task] });
 
   const initialValues: FormValues = previous
@@ -50,8 +53,10 @@ const EnhancedTextInputComponent = (props: Props) => {
 
     if (step === "input") return setStep("task");
 
-    if (step === "task")
+    if (step === "task") {
+      if (taskSubStep === "selection") return setTaskSubStep("modification");
       props.handleSubmit?.({ data: makeBreadcrumb(props.fn, values) });
+    }
   };
 
   const TaskComponent = taskComponents[props.task];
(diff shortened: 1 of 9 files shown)


DafyddLlyr (changes requested):
The nesting of these steps and subSteps leads to quite a lot of complexity in the code.

We should consider them to be all "one level" - the UI pattern we settle on for the `ProjectDescription` task will be the one we apply to _all_ `EnhancedTextInput` tasks in future. By nesting this `"modification"` step within `ProjectDescription` we're setting ourselves up to repeat this section of code each time we add tasks.

Shared state is already handled via the Formik wrapper which exists at the `EnhancedTextInputComponent` level.

Happy to talk through this approach or give more detail if you'd like. Also happy to pick this up myself as a series of commits on this branch - please just let me know 👍 

```tsx
const [step, setStep] = useState<"input" | "selection" | "modification">("input");

...

<Formik<FormValues>>
  ...
  <Card>
    ...
    // Shared for all tasks
    {step === "input" && <InitialUserInput {...props} />}
    {step === "modification" && <ModifyUserInput {...props} />}

    // Unique per-task
    {step === "task" && <TaskComponent {...props} />}
    ...
  </Card>
</Formik>
```

ianjon3s:
Thanks for this @DafyddLlyr , I have now separated the modify step into a separate component `ModifyUserInput.tsx` which simplifies the logic

apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/index.tsx
```diff
@@ -91,7 +113,11 @@ const EnhancedTextInputComponent = (props: Props) => {
               />
             )}
             {step === "input" && <InitialUserInput {...props} />}
-            {step === "task" && <TaskComponent {...props} />}
+            {step === "modification" && <ModifyUserInput {...props} />}
+            {step === "selection" && (
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              <TaskComponent {...(props as any)} />
+            )}
```
DafyddLlyr:
```suggestion
            {step === "selection" && <TaskComponent {...props} />}
```

I think this should work - I can't see any type issues here.

apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/index.tsx
```diff
@@ -70,8 +77,23 @@ const EnhancedTextInputComponent = (props: Props) => {
...
+
+        const handleBack = (() => {
+          if (step === "modification") return () => setStep("selection");
+          if (step === "selection" && !isRunningTask)
+            return () => {
+              if (values.status === "success") {
+                setFieldValue("userInput", values.original);
+              }
+              setStep("input");
+            };
```
DafyddLlyr, line 95: nit: This might be worth an explanation - `// Repopulate field with user's original input"

DafyddLlyr (approved):
Appreciate the additional tests here and for picking up the changes - this is now much clearer and more expandable.

One small type nit and a suggestion for adding a comment 👍

ianjon3s:
Thanks for this @DafyddLlyr, I've added your suggested changes. I'll hold off merging this one until we're in a better place with the React upgrade

DafyddLlyr (approved):
Looking great - thank you! Let's get a prod deploy queued up here also.

PR #6415, opened 2026-04-07
fix: clear auth cookies on failed login to editor

joecarver:
Fixes a couple of small issues I noticed when getting set up for the first time, and adjusts README accordingly.

- when trying to log in to the editor as a user who exists, but does not have proper permissions set, I was getting bounced back to the /login page but kept in a 'logged in' state, as the auth cookies were still set. This broken state made it impossible to properly sign in after permissions were correctly added
    - a more complete fix might be to add a 'permission denied' screen - this would cover the state accurately i.e. authenticated but not authorized. This PR just makes that state impossible to reach, which i think is fine for now
- achieved this by redirecting through the `/logout` route which already handles clearing of cookies. I tidied up the cookie clearing in this route to match what I was seeing also
- added a note to the README regarding permissions required for new dev users

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ planx-new is a monorepo containing our full application stack. Here's a quick su
 5. Move into the hasura directory `cd ../apps/hasura.planx.uk` and install dependencies `pnpm i`.
 
 6. Open [Hasura's](https://hasura.io/) web console (`cd apps/hasura.planx.uk` then `pnpm start`) and check that your Google email address is in the `users` table, if not then add it. This will eventually allow you to authenticate into the application as an admin.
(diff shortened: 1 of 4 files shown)


DafyddLlyr (changes requested):
Thanks for picking this up! I'm still hitting this issue on this Pizza - 

Image

It looks like it may be because we're returning on error on the API `/logout` route ([see code here](https://github.com/theopensystemslab/planx-new/blob/main/apps/api.planx.uk/modules/auth/routes.ts#L12)). It may be an issue with the order of operations - we should hit this endpoint before clearing the local cookies.

joecarver:
closing as unecessary, due to larger refactor in #6450

PR #6418, opened 2026-04-07
feat: Add 'Open in Editor' button to /draft links

augustlindemer:
This PR adds a button to /draft links that opens a link to the node of the active card in the PlanX Editor of its respective environment.

Currently uses a Floating Action Button, the first one in the codebase as far as I can see? Idea was that it's nicely detached, highly visible, and the sole action that will take you out of the /draft tab. Happy to replace with a different component, as it may compete too much with the 'Continue' button as the primary action on the screen.

### DIFF
--- a/apps/editor.planx.uk/src/components/OpenInEditorButton.tsx
+++ b/apps/editor.planx.uk/src/components/OpenInEditorButton.tsx
@@ -12,6 +12,7 @@ const StyledFab = styled(Fab)(() => ({
   zIndex: 9999,
   backgroundColor: "#2c2c2c",
   color: "#fff",
+  border: "1px solid #ffffff",
   "&:hover": {
     backgroundColor: "#444",
   },
@@ -26,16 +27,11 @@ const StyledFab = styled(Fab)(() => ({
  * On click, opens the corresponding node in the PlanX Editor in a new tab.
  */
 const OpenInEditorButton: React.FC = () => {
-  const getEditorURLForCurrentCard = useStore(
-    (state) => state.getEditorURLForCurrentCard,
-  );
-
-  const isDraft = window.location.pathname.includes("/draft");
-  if (!isDraft) return null;
+  const currentCardURL = useStore((state) => state.currentCardURL);
 
   const handleClick = () => {
     try {
-      const url = getEditorURLForCurrentCard();
+      const url = currentCardURL();
       if (url) {
         window.open(url, "_blank");
       }
--- a/apps/editor.planx.uk/src/components/OpenInEditorButton.tsx
+++ b/apps/editor.planx.uk/src/components/OpenInEditorButton.tsx
@@ -1,0 +1,62 @@
+import OpenInNewIcon from "@mui/icons-material/OpenInNew";
+import Fab from "@mui/material/Fab";
+import { styled } from "@mui/material/styles";
+import Tooltip from "@mui/material/Tooltip";
+import { useStore } from "pages/FlowEditor/lib/store";
+import React from "react";
+
+const StyledFab = styled(Fab)(() => ({
+  position: "fixed",
+  bottom: 20,
+  left: 20,
+  zIndex: 9999,
+  backgroundColor: "#2c2c2c",
+  color: "#fff",
+  "&:hover": {
+    backgroundColor: "#444",
+  },
+  textTransform: "none",
+  fontFamily: "sans-serif",
+  fontSize: "0.8125rem",
+  gap: 6,
+}));
+
+/**
+ * Renders an "Open in Editor" button when the current URL is a /draft preview.
+ * On click, opens the corresponding node in the PlanX Editor in a new tab.
+ */
+const OpenInEditorButton: React.FC = () => {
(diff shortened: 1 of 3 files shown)


joecarver (commented):
Nice, seems like a v helpful feature and is working as expected from what I can tell. Left a couple of small suggestions but I may be wrong, please correct me if so!

apps/editor.planx.uk/src/pages/Preview/Questions.tsx
```diff
@@ -257,6 +258,7 @@ const Questions = ({ previewEnvironment }: QuestionsProps) => {
           </ErrorBoundary>
         </Main>
       )}
+      <OpenInEditorButton />
```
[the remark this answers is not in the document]
joecarver: Oh yeah that's a great point. All good then!

apps/editor.planx.uk/src/components/OpenInEditorButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+  zIndex: 9999,
+  backgroundColor: "#2c2c2c",
+  color: "#fff",
+  "&:hover": {
+    backgroundColor: "#444",
+  },
+  textTransform: "none",
+  fontFamily: "sans-serif",
+  fontSize: "0.8125rem",
+  gap: 6,
```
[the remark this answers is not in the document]
augustlindemer: Not sure there's any concern about covering license copy on a /draft link, since not a public facing environment. I like floating because it means I don't have to scroll up or down on large content pages to get to the editor (especially since what I'm looking for may be a particular part of the content on the page, so worst case I'd have to tab back and scroll back to content to compare editor/front). If FAB generally is okayed I'll add a 1px solid border as suggested!

apps/editor.planx.uk/src/components/OpenInEditorButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+ * Renders an "Open in Editor" button when the current URL is a /draft preview.
+ * On click, opens the corresponding node in the PlanX Editor in a new tab.
+ */
+const OpenInEditorButton: React.FC = () => {
+  const getEditorURLForCurrentCard = useStore(
+    (state) => state.getEditorURLForCurrentCard,
+  );
+
+  const isDraft = window.location.pathname.includes("/draft");
+  if (!isDraft) return null;
```
jessicamcinchak: nit: I would personally expect the `OpenInEditorButton` component to always return a button, not a button _or null_ !

Is this `isDraft` logic perhaps better moved to `Questions` and then we could conditionally render like `{isDraft && <OpenInEditorButton />}` ?

Also, this would technically match a flow name that starts with the word draft like "/draft-test-flow/published";  `endsWith` would be safer choice than `includes` here?

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -117,6 +119,7 @@ export interface PreviewStore extends Store.Store {
   autoAnswerableFlag: (filterId: NodeId) => NodeId | undefined;
   hasAcknowledgedWarning: boolean;
   setHasAcknowledgedWarning: () => void;
+  getEditorURLForCurrentCard: () => string | null;
```
jessicamcinchak: nit: don't have particularly strong naming preferences, but I do wonder if something like `currentCardURL` or `currentCardEditorURL` would be simpler here (similar to existing `currentURL` in store) ?

Thinking about this from two angles: 
- When I'm accessing the store in the console via `window.api.getState()...` and it's auto-suggesting, it'd be nice for this to be grouped/suggested alongside `currentCard` ? 
- It's not always the case, but I typically think of `get...` store functions as requiring an argument or param, but that's not the case here as it reads all info from internal state. So maybe a name just describing the _output_ is more appropriate?

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -863,6 +866,67 @@ export const previewStore: StateCreator<
...
+
+    const cleanParentId = getCleanId(parent);
+    const cleanGrandparentId = getCleanId(grandparent);
+
+    const nodePath =
+      node?.type === TYPES.Answer
+        ? `nodes/${cleanGrandparentId}/nodes/${cleanParentId}/edit#${node.id}`
+        : `nodes/${cleanParentId}/nodes/${node?.id}/edit`;
+
+    return `${window.location.origin}/app/${basePath}${portalPath}/${nodePath}`;
```
jessicamcinchak, line 928: Link logic all working exactly as expected on pizza for me - navigating direct to something like osl/property-types while testing council/apply-for-planning-permission feels like proper magic 🌟

jessicamcinchak (approved):
This is one of those PRs that makes me feel like the future one-day dream editor _is here_ ! 🔮 🚀 

Functionally all works for me, a few minor comments but no show stoppers! Since this is `/draft` links only for now, very happy for any styles to be refined overtime 🙂

PR #6421, opened 2026-04-08
feat: user pinned flows - editor frontend 

joecarver:
### What's in this PR?
Editor frontend changes to display and manage pinned flows

### Design & implementation details / decisions
- used a star icon for the pin/unpin icon
- little loading animation whilst the mutation is being applied
- moved the grid/table toggle into the filters & sort section 
- pinned flows are not included in the regular flow list
- graphql mutation returns the full modified flow, and the frontend partially updates its `flows` state - this way you avoid refetching and rendering all the flows, but you also ensure that the browser reflects DB state directly rather than tracking its own state

### DIFF
--- a/apps/editor.planx.uk/src/pages/Team/components/FlowCard/queries.ts
+++ b/apps/editor.planx.uk/src/pages/Team/components/FlowCard/queries.ts
@@ -1,4 +1,6 @@
-import { gql } from "@apollo/client";
+import { useMutation } from "@apollo/client";
+import gql from "graphql-tag";
+import { FlowSummary } from "pages/FlowEditor/lib/store/editor";
 
 // this is copied from the getFlows query in FlowEditor/lib/store/editor.ts - might be worth re-using the fragment there as well?
 const FLOW_SUMMARY_FIELDS = gql`
@@ -63,3 +65,30 @@ export const UNPIN_FLOW = gql`
     }
   }
 `;
+
+interface PinFlowMutation {
+  insert_user_pinned_flows_one: { flow: FlowSummary };
+}
+
+interface UnpinFlowMutation {
+  delete_user_pinned_flows: { returning: { flow: FlowSummary }[] };
+}
+
+interface PinFlowVars {
+  flowId: string;
+  userId: number;
+}
+
+interface UnpinFlowVars {
+  flowId: string;
+}
+
+export const usePinFlow = (variables: PinFlowVars) =>
+  useMutation<PinFlowMutation, PinFlowVars>(PIN_FLOW, {
+    variables,
+  });
+
+export const useUnpinFlow = (variables: UnpinFlowVars) =>
+  useMutation<UnpinFlowMutation, UnpinFlowVars>(UNPIN_FLOW, {
+    variables,
+  });
--- a/apps/editor.planx.uk/src/pages/Team/components/FlowCard/queries.ts
+++ b/apps/editor.planx.uk/src/pages/Team/components/FlowCard/queries.ts
@@ -1,0 +1,65 @@
+import { gql } from "@apollo/client";
+
+// this is copied from the getFlows query in FlowEditor/lib/store/editor.ts - might be worth re-using the fragment there as well?
+const FLOW_SUMMARY_FIELDS = gql`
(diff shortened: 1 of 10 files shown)


apps/editor.planx.uk/src/pages/Team/components/FlowPinButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+export const FlowPinButton = ({
+  flowId,
+  isPinnedByCurrentUser,
+  updateFlow,
+}: Props) => {
+  const user = useStore((state) => state.user);
+
+  const [pinFlow, { loading: isPinLoading }] = useMutation<{
+    insert_user_pinned_flows_one: { flow: FlowSummary };
+  }>(PIN_FLOW);
```
zz-hh-aa: Nice! I believe the pattern we want here is to create a hook and consumer it in the component (see https://github.com/theopensystemslab/planx-new/blob/main/doc/guides/data-fetching-and-state-management.md) but I'll let Jess or Daf correct me if that's unnecessary here

apps/editor.planx.uk/src/pages/Team/components/FlowCard/queries.ts
```diff
@@ -0,0 +1,65 @@
+import { gql } from "@apollo/client";
+
+// this is copied from the getFlows query in FlowEditor/lib/store/editor.ts - might be worth re-using the fragment there as well?
```
zz-hh-aa, line 5: gql fragments are new to me! Just commenting cause I'd like to follow discussion here, maybe of use for archived flows queries too... 👀

zz-hh-aa (commented):
Pushpin looks good and works as expected locally :) 

I touched `Team/index` yesterday and again [today](https://github.com/theopensystemslab/planx-new/pull/6434), sorry for inconvenient timing--so guess you'll have to rebase (and I'll do the same after with the rest of the archive view stuff).

I'll leave a green tick to the more experienced eyes of Daf or Jess! 

One other question: not sure if it's worth looking at making the gql requests `network-only` to stop pinned flows flashing across teams? With the caveat that I'm not sure how often non-OSL editors will be toggling across teams with different pinned flows... just a thought
https://github.com/user-attachments/assets/15ad075a-ca0e-43eb-979c-0e241b6cbbc6

apps/editor.planx.uk/src/pages/Team/components/FlowPinButton.tsx
joecarver: good spot! done https://github.com/theopensystemslab/planx-new/pull/6421/commits/a036a5a7a0bec1052c265b29ef7d0552f18139f1

apps/editor.planx.uk/src/pages/Team/components/FlowCard/queries.ts
joecarver: yeah I spotted a couple of places places where we could potentially have some re-usable fragments... keen to hear others thoughts on this before making any further refactors

DafyddLlyr (approved):
Happy for this to merge once rebases are resolved!

Great piece of work - code is clear and well laid out, and the functionality meets the requirements. The animation on "pin" is really nice 👌 

nit: On the Pizza (which may be out of date - latest CI not ran) there's layout shift caused by the "showing X pinned flows" header not being the same height as the other headers. I wasn't able to quite identify where in the diff this is introduced (looks like it's the same component). Small detail but resolving it would be nice!

https://github.com/user-attachments/assets/5fb68e6d-0811-423a-a27e-142762951a18

apps/editor.planx.uk/src/pages/Team/components/FlowCard/queries.ts
DafyddLlyr: Yeah this is something we should be using more frequently for shared types - there are a few examples here and there in the codebase.

Ultimately the plan is to remove direct GQL queries into a library, which will generate types for codegen ✨ 

Please see https://github.com/theopensystemslab/planx-new/blob/main/doc/architecture/decisions/0011-standardised-data-fetching-and-state-management.md#phase-3-graphql-client

PR #6434, opened 2026-04-09
refactor: Teams page into separate components

zz-hh-aa:
Next step in creating archived flows view. Refactors some components from `Team/index.tsx` into separate components for reuse across `Flows` view and `Archive` view.


DafyddLlyr (approved):
Nice!

PR #6435, opened 2026-04-09
feat: archived flows view

zz-hh-aa:
Still behind a feature flag!

# What does this PR do?
- Create a new `Archive` component to show archived flows
- New `showDetails` prop in components to conditionally render flow info (eg we want to show status and service type for live flows, but not for archived ones)
- Hook `TeamLayout` component up to `Flows` and `Archive` components to conditionally render view depending on `flowView` state

# Testing
- Go to the Pizza and run `window.featureFlags.toggle("ARCHIVE_VIEW")`
- Navigate to the Teams page and toggle between Archive and Flows view (features missing include Actions tooltip on archived flows, removing links to archived flows, etc.)
Screenshot 2026-04-09 160653

Screenshot 2026-04-09 160659

### DIFF
--- a/apps/editor.planx.uk/src/pages/Team/helpers/useGetArchivedFlows.tsx
+++ b/apps/editor.planx.uk/src/pages/Team/helpers/useGetArchivedFlows.tsx
@@ -1,6 +1,10 @@
 import { useQuery } from "@apollo/client";
 
-import { GET_ARCHIVED_FLOWS, GetArchivedFlowsQuery, GetArchivedFlowsVars } from "../queries";
+import {
+  GET_ARCHIVED_FLOWS,
+  GetArchivedFlowsQuery,
+  GetArchivedFlowsVars,
+} from "../queries";
 
 export const useArchivedFlows = (teamId: number) =>
   useQuery<GetArchivedFlowsQuery, GetArchivedFlowsVars>(GET_ARCHIVED_FLOWS, {
--- a/apps/editor.planx.uk/src/pages/Team/helpers/useGetArchivedFlows.tsx
+++ b/apps/editor.planx.uk/src/pages/Team/helpers/useGetArchivedFlows.tsx
@@ -6,7 +6,7 @@ import {
   GetArchivedFlowsVars,
 } from "../queries";
 
-export const useArchivedFlows = (teamId: number) =>
+export const useGetArchivedFlows = (teamId: number) =>
   useQuery<GetArchivedFlowsQuery, GetArchivedFlowsVars>(GET_ARCHIVED_FLOWS, {
     variables: { teamId },
   });
--- a/apps/editor.planx.uk/src/pages/Team/index.tsx
+++ b/apps/editor.planx.uk/src/pages/Team/index.tsx
@@ -1,8 +1,7 @@
 import Box from "@mui/material/Box";
 import Container from "@mui/material/Container";
-import { styled } from "@mui/material/styles";
 import Typography from "@mui/material/Typography";
-import { useNavigate, useSearch } from "@tanstack/react-router";
+import { useSearch } from "@tanstack/react-router";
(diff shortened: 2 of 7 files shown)


apps/editor.planx.uk/src/pages/Team/queries.ts
```diff
@@ -31,11 +31,23 @@ export const GET_ARCHIVED_FLOWS = gql`
           name
         }
       }
+      publishedFlows: published_flows(
```
zz-hh-aa, line 34: The `useArchivedFlows` query wasn't hooked up to anything before when I merged #6425 so just had to fix a few things here. 

I thought about creating an `ArchivedFlowSummary` type in addition to `FlowSummary` (so that we wouldn't have to fetch unnecessary info, like the last published flows record) but this led to some downstream type complexities--figured I'd keep it simpler for now.

apps/editor.planx.uk/src/pages/Team/helpers/useGetArchivedFlows.tsx
DafyddLlyr, line 1: nit: Mismatch between filename and hook name

apps/editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -68,14 +55,16 @@ const Team: React.FC<TeamProps> = ({ flows: initialFlows }) => {
   ]);
 
   const [flows, setFlows] = useState<FlowSummary[] | null>(initialFlows);
+  const [flowView, setFlowView] = useState<FlowView>("flows");
+  const { data: archivedFlowsData } = useArchivedFlows(teamId);
```
DafyddLlyr: Over time, this list could grow to become quite substantial for a team - they're potentially fetching data they don't need to.

Is there a reason we need to fetch it here and not just in the `<Archive/>` component? If we did this we'd only request the data when needed.

If we do split this page into `/explore`, `/flows` and `/archive` pages in future we could also add a `preload` on the tab. We can also do this `onMouseOver` of the tab just now if we wanted to I guess - this is probably slightly overkill just now as it's a fairly quick request?

apps/editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -68,14 +55,16 @@ const Team: React.FC<TeamProps> = ({ flows: initialFlows }) => {
...
+  const archivedFlows = archivedFlowsData?.flows ?? null;
```
DafyddLlyr, line 60: nit: Rather than just falling back to `null` we should handle possible `error` and `loading` states. As with the above comment we could do this in the `<Archive/>` component.

DafyddLlyr (approved):
One small non-blocking comment - looking good!

This might be a known issue and for another PR but right now there's a success message on archive, but no UI updates - it remains within the current list and does not appear in the list of archived flows. It doesn't happen automatically as there's a filter on `archived_on` (and I don't think this field is included anyway). We'll need to refetch or cache-evict to get the UI to automatically update.

zz-hh-aa:
> This might be a known issue and for another PR but right now there's a success message on archive, but no UI updates - it remains within the current list and does not appear in the list of archived flows. It doesn't happen automatically as there's a filter on `archived_on` (and I don't think this field is included anyway). We'll need to refetch or cache-evict to get the UI to automatically update.

Thanks Daf, yep I will handle this next along with the decision on where exactly to fetch the archived flows in the first place. Since this is approved + feature flagged I will merge this for now!

re:
> [Is there a reason we need to fetch it here and not just in the <Archive/> component? If we did this we'd only request the data when needed.](https://github.com/theopensystemslab/planx-new/pull/6435#discussion_r3063045168)

I put this here since toggling between Flows and Archive was slow, but maybe that's not a particularly common use case anyway. I'll figure out fetching next :)

PR #6442, opened 2026-04-09
chore(editor): Drop `UPLOAD_LABEL_REBUILD` feature flag from public interface 

DafyddLlyr:
## What does this PR do?
- Removes check for `UPLOAD_LABEL_REBUILD` feature flag when rendering the public interface of the `FileUploadAndLabel` component - only display the "new" accordion based view, and not the modal based view
- Removes files and tests only associated with modal view

## Motivation
In order to proceed with the refactors of the `FileUploadAndLabel` component, as part of the broader goal of implementing the "drawing numbers" feature, I would next be working on updating the `Dropzone<T>` component. Currently, this would need to be done in two places (modal + accordion) - this is unnecessary if we're in the process of removing the modal view.

When we [were discussing feature flagging this](https://github.com/theopensystemslab/planx-new/pull/6216#pullrequestreview-3944660800), I didn't quite click that we could block the `drawingNumber` functionality by just hiding the toggle in the Editor. This is still in place, please see code here - 

https://github.com/theopensystemslab/planx-new/blob/babf73929fe2546a53937ff1dc3a8db8473347b9/apps/editor.planx.uk/src/%40planx/components/FileUploadAndLabel/Editor.tsx#L104-L120


apps/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/index.tsx
```diff
@@ -1,16 +1,284 @@
...
+  [theme.breakpoints.up("md")]: {
+    marginTop: "-1em",
+  },
+}));
 
 export default function Component(props: Props) {
-  if (hasFeatureFlag("UPLOAD_LABEL_REBUILD")) {
-    return <FileUploadAndLabelNew {...props} />;
-  }
-  return <FileUploadAndLabel {...props} />;
```
DafyddLlyr, line 15: Dropping this conditional switch is the only logical change in this PR - the rest is files which have been deleted and then renamed. The diff here has picked these up as file changes instead of deletes + renames.

jessicamcinchak (approved):

PR #6445, opened 2026-04-10
chore: rename `submission_integrations` table to `submission_emails` 

zz-hh-aa:
An outstanding task from the multiple submission email work. Renaming the table to be more accurate and intuitive, and columns to `submission_emails.address` (from `submission_email`), `submission_emails.is_default` (from `default_email`) to be less verbose / repetitive. 

# What does this PR do?
- Makes the DB changes
- Updates the Hasura event trigger that enforces one default email per-team
- Renames everything across the codebase

# Testing
I tested all parts of the codebase that will interact with the `submission_emails` table:
- Submission emails table in TeamSettings
- Updating the email destination from the Send component
- End-to-end submission


DafyddLlyr (changes requested):
Thanks...! Tedious but important work to catch up on this - we'll be grateful that we've sorted this in future 👍 

Only requested change would be to please run `hasura migrate squash` to logically group these migrations together ([docs](https://hasura.io/docs/2.0/hasura-cli/commands/hasura_migrate_squash/)).

zz-hh-aa:
Thanks Daf, yeah lesson learned it's better to rename earlier than later when you've got loads of references to the thing scattered across the codebase...

Ended up squashing manually after timeboxing CLI troubleshooting, but good to know about this for the future!

DafyddLlyr (approved):

PR #6446, opened 2026-04-10
chore: fetch archived flows in Archive component

zz-hh-aa:
This PR:
- Moves archived flows fetch into Archive component to speed up initial load time (as per [Daf's suggestion](https://github.com/theopensystemslab/planx-new/pull/6435#pullrequestreview-4088298478))
- Ensures that Flows UI updates on archive
- Uses `cache-and-network` fetch policy on archived flows to speed up load / ensure list updates with newly archived flows too

### DIFF
--- a/apps/editor.planx.uk/src/pages/Team/components/FlowMenu.tsx
+++ b/apps/editor.planx.uk/src/pages/Team/components/FlowMenu.tsx
@@ -56,17 +56,19 @@ const FlowMenu: React.FC<Props> = ({
     refreshFlows();
   };
 
-  const handleArchive = async () => {
-    try {
-      await archiveFlow(flow);
-      setOpenDialog(null);
-      toast.success("Archived flow");
-    } catch (error) {
-      toast.error(
-        "We are unable to archive this flow, refesh and try again or contact an admin",
-      );
-    }
-  };
+const handleArchive = async () => {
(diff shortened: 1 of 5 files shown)


apps/editor.planx.uk/src/pages/Team/components/FlowMenu.tsx
```diff
@@ -56,17 +56,19 @@ const FlowMenu: React.FC<Props> = ({
...
+    refreshFlows();
+    toast.success("Archived flow");
+  } catch (error) {
+    toast.error(
+      "We are unable to archive this flow, refesh and try again or contact an admin",
+    );
+  } finally {
+    setOpenDialog(null);
+  }
+};
```
DafyddLlyr, line 71: A nice outcome would be to leave this better than we found it and remove `archiveFlow()` from the Zustand store and convert this to a `useMutation()` hook 📈

DafyddLlyr (approved):

apps/editor.planx.uk/src/pages/Team/components/FlowMenu.tsx
zz-hh-aa: Working on my `useUnarchiveFlow` hook right now so will include as part of that PR :)

PR #6447, opened 2026-04-10
feat: disable links to archived flows

zz-hh-aa:
Removes flow links from Archive tab table view. They were already disabled for FlowCard Archive view. 

Before:

https://github.com/user-attachments/assets/f9a57285-a181-4a1c-927b-64e0a73d5519

After:

https://github.com/user-attachments/assets/eab49297-7e90-4125-a93b-818b09d1004a

### DIFF
--- a/apps/editor.planx.uk/src/pages/Team/components/FlowCard/index.tsx
+++ b/apps/editor.planx.uk/src/pages/Team/components/FlowCard/index.tsx
@@ -15,7 +15,7 @@ import {
   Card,
   CardBanner,
   CardContent,
-  DashboardLink,
+  FlowCardLink,
   LinkSubText,
 } from "./styles";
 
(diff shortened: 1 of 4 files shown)


apps/editor.planx.uk/src/pages/Team/components/FlowCard/index.tsx
```diff
@@ -15,7 +15,7 @@ import {
   Card,
   CardBanner,
   CardContent,
-  DashboardLink,
+  FlowCardLink,
```
zz-hh-aa, line 18: Renamed to distinguish `FlowCardLink` and `FlowRowLink` components

apps/editor.planx.uk/src/pages/Team/components/FlowTable/index.tsx
```diff
@@ -101,26 +96,8 @@ const FlowTableRow: React.FC<FlowTableRowProps> = ({
 
   const { displayTimeAgo, displayActor } = useFlowDates(flow);
 
-  const handleRowClick = (e: React.MouseEvent) => {
```
zz-hh-aa, line 104: We had two bits of link functionality here--I have opted for following the pattern that `FlowCard` was using (using a stylised version of `CustomLink` on top of flow info)

DafyddLlyr (approved):

PR #6450, opened 2026-04-13
fix(editor): Simplify auth process by centralising within an `AuthStore`

DafyddLlyr:
## What's the problem?
Auth state within the app is split across multiple sources truth. We're currently tracking and managing this in the following locations - 

- `UserStore` — we hold `user` and `jwt` properties which where initialised via `initUserStore()`
- `/_authenticated` route — this initUserStore() as its auth guard, via a domain validation loader
- `/logout` route — this directly called the API request `logout()`, cleared cookies, disconnected ShareDB, reset Apollo, then redirected 
- `hasJWT()` - helper function passed invoked as router context - the output of this is unused, but triggering it sets cookies and clears URL

This makes following (and fixing!) the auth lifecycle pretty challenging. The following two issues are the motivating factor behind this PR, and the catalyst to rationalise this code just now.

### Bug: logout produced a white screen after the TanStack Router migration
The logout route was protected by useLoggedInUserAuth middleware, which rejected expired tokens with a redirect. Expired tokens are a normal logout scenario — blocking them meant logout was broken for any user whose session had already expired.

### Bug: `httpOnly` JWT cookie not cleared client-side on logout
The JWT `httpOnly` cookie set by the API on login was never explicitly cleared on logout. We cleared auth and localStorage, but the `httpOnly` cookie persisted, leaving a revoked token in the browser. This isn't a security issue per-se as the partner `auth` cookie was removed (blocking the app from authenticating), and the JWT is block-listed on the DB, but it's better that we're removing this and following out the intent in the code.

## How did we get here?
Slowly I guess..! I think this has honestly accumulated over many PRs - originally this mostly lived within `App.tsx` as `hasJWT()` and as we've needed various features (expiry, logout, auth guards, JWT for new clients) it's spread and gotten a bit messy. The recent refactor to Tanstack Router has fragmented this even further.

## What's the solution?
All auth lifecycle logic has been moved into a single `authStore` Zustand slice, merged into the main store alongside the existing slices.

We now use a single `authStatus` variable (`"idle" | "loading" | "authenticated" | "unauthenticated"`) which can be referenced at any time - we don't need to infer this from the presence of a JWT, `User`, or failed `/user/me` request.

This should solve the issues listed above - auth state is clear and centralised, and the above bugs are resolved.

## Prior art
 - https://github.com/theopensystemslab/planx-new/pull/6448
 - https://github.com/theopensystemslab/planx-new/pull/6415

### DIFF
--- a/apps/api.planx.uk/modules/auth/routes.ts
+++ b/apps/api.planx.uk/modules/auth/routes.ts
@@ -7,11 +7,7 @@ export default (passport: Authenticator): Router => {
   const router = Router();
 
   router.get("/auth/login/failed", Controller.failedLogin);
-  router.post(
-    "/auth/logout",
-    Middleware.useLoggedInUserAuth,
-    Controller.logout,
-  );
+  router.post("/auth/logout", Controller.logout);
 
   router.get("/auth/google", Middleware.getGoogleAuthHandler(passport));
   router.get(
--- a/apps/editor.planx.uk/src/pages/FlowEditor/lib/store/auth.ts
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/lib/store/auth.ts
@@ -1,0 +1,94 @@
+import { CoreDomainClient } from "@opensystemslab/planx-core";
+import type { User } from "@opensystemslab/planx-core/types";
+import { getUser, logout } from "lib/api/auth/requests";
+import { clearCookie, setCookie } from "lib/cookie";
+import { client } from "lib/graphql";
+import { disconnectShareDB } from "pages/FlowEditor/lib/sharedb";
+import { type StateCreator } from "zustand";
+
+import type { SharedStore } from "./shared";
+
+export interface AuthStore {
+  authStatus: "idle" | "loading" | "authenticated" | "unauthenticated";
+  user: User | null;
+  jwt: string | null;
+
+  initAuthStore: () => Promise<void>;
+  logout: () => Promise<void>;
+}
+
+export const authStore: StateCreator<
+  AuthStore & SharedStore,
+  [],
+  [],
+  AuthStore
+> = (set, get) => ({
+  authStatus: "idle",
+  user: null,
+  jwt: null,
+
+  initAuthStore: async () => {
+    if (get().authStatus === "authenticated") return;
+    if (get().authStatus === "loading") return;
+
+    set({ authStatus: "loading" });
+
+    // On local and Pizza environments, JWT is stored as a URL param due to restrictions on
+    // cross-domain cookies (we auth via planx.dev). On staging and production these cookies 
+    // are set via response headers from the API.
+    const url = new URL(window.location.href);
+    const jwtParam = url.searchParams.get("jwt");
+
+    if (jwtParam) {
+      setCookie("jwt", jwtParam);
+      setCookie("auth", JSON.stringify({ loggedIn: true }));
+
+      url.searchParams.delete("jwt");
+      window.history.replaceState({}, document.title, url.pathname + url.search);
+    }
+
+    try {
+      const result = await getUser();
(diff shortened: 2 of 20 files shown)


apps/api.planx.uk/modules/auth/routes.ts
```diff
@@ -7,11 +7,7 @@ export default (passport: Authenticator): Router => {
   const router = Router();
 
   router.get("/auth/login/failed", Controller.failedLogin);
-  router.post(
-    "/auth/logout",
-    Middleware.useLoggedInUserAuth,
```
DafyddLlyr, line 12: We don't need an auth guard here - the API is still hit when you navigate to the `/logout` page in the frontend, where the user may not have a valid token (removed or expired).

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/user.ts
```diff
@@ -13,31 +11,17 @@ export const getDisplayRole = (user: User): string => {
...
-  jwt?: string;
-
-  setUser: (user: User & { jwt: string }) => void;
   canUserEditTeam: (teamSlug: Team["slug"]) => boolean;
-  initUserStore: () => Promise<User>;
   getUserRoleForCurrentTeam: () => Role | undefined;
   getUserRole: () => string | undefined;
 }
 
 export const userStore: StateCreator<
```
DafyddLlyr, line 19: We could probably remove this store and replace with hooks or components which handle the logic here - that would probably be simpler to manage and test. No need for this change as part of this PR though.

apps/editor.planx.uk/src/routes/(auth)/logout.tsx
```diff
@@ -1,35 +1,9 @@
...
-
-    // Clear cookies
-    clearCookie("auth");
-    clearCookie("jwt");
-
-    // Clear localStorage
-    localStorage.removeItem("jwt");
-
-    // Must throw redirect to trigger navigation
+    await useStore.getState().logout();
```
DafyddLlyr, line 6: **Why keep a `/logout` route?**
We could just invoke the `logout()` method `onClick()` of the "Logout" button but invoking this on a particular route is very handy for debugging locally and for council-facing issues - asking s a user to navigate here is much easier than having them clear cookies in their browser.

apps/editor.planx.uk/src/routes/_authenticated/app/route.tsx
DafyddLlyr, line 1: Auth logic now removed from here - it's all handled upstream once.

jessicamcinchak (approved):
All working as expected for me on pizza ! Thanks for this solution, agree this solution is much easier to reason with / will be nicer to maintain going forward !

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/auth.ts
```diff
@@ -0,0 +1,94 @@
...
+    await logout()
+
+    // Clean up client connections
+    disconnectShareDB();
+    await client.resetStore();
+
+    // Clear all client-side auth tokens
+    clearCookie("auth");
+    clearCookie("jwt");
+    localStorage.removeItem("jwt");
```
joecarver, line 83: question: is this line necessary, do we ever actually set the jwt in the localStorage? in my PR I was removing this line because I wasn't seeing it in the browser storage, and I had a search through the code and couldn't see anywhere it was being set.

apps/editor.planx.uk/src/routes/(auth)/logout.tsx
joecarver: also better for accessibility to keep the HTML semantic :~)

PR #6455, opened 2026-04-13
feat: search for an address to position the map when proposing a new address

jessicamcinchak:
See https://trello.com/c/70MKi9UT/2872-add-a-search-bar-to-my-site-does-not-have-an-address-map-interaction


apps/editor.planx.uk/src/lib/gis.ts
```diff
@@ -44,6 +44,52 @@ export const useGeoJSONChange: UseGeoJSONChange = (mapId, callback) => {
...
+      geocodeAutocomplete?.removeEventListener(
+        "addressSelection",
+        geocodeAutocompleteSearchHandler,
+      );
+    };
+  }, [callback, inputId]);
+
+  return [selectedAddress, setSelectedAddress];
+};
+
```
jessicamcinchak: This hook comes via suggestion here: https://opensystemslab.slack.com/archives/C01E3AC0C03/p1769090181944129?thread_ts=1769088559.434609&cid=C01E3AC0C03

But it's still un-used as I have yet to figure out where to successfully call it here !! In the other example, `MapAndLabel` is using React 'Context' which is not implemented here in `FindProperty`, and I hope can be avoided as this is _hopefully_ a simpler case?

apps/editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
```diff
@@ -142,7 +216,9 @@ export default function PlotNewAddress(props: PlotNewAddressProps): FCReturn {
               id="plot-new-address-map"
               ariaLabelOlFixedOverlay="An interactive map for providing your site location"
               data-testid="map-web-component"
-              zoom={10}
+              zoom={searchedPoint ? 18 : 10}
+              latitude={searchedPoint && searchedPoint.LPI.LAT}
+              longitude={searchedPoint && searchedPoint.LPI.LNG}
```
jessicamcinchak, line 213: These props demonstrate what I _expect_ to be rendered/re-rendered when `setSearchedPoint` is called ! 

On initial load of the component, there will _never_ be a `searchedPoint` so the map will load zoomed-out based on bbox. When I search and select an address from the dropdown, I expect the map to zoom into that point. `searchedPoint` appears to be setting and updating in state, but the map web component is failing to re-render.

DafyddLlyr (commented):
Ok, I've finally caught up here and I think I know what's going on - it looks like a Lit lifecycle issue 👍 

Right now we just one one way data transfer between the Map and React.
- From Map to React (Drawing made on map → `geoJsonChange` event fires → React state updates)
- From React to Map (pass in _initial_ data - boundary, center, existing features)

Notably the map is only updating based on this initial data - used in `firstUpdated()`. When a `@property` like `latitude` updates the Lit `render()` function is re-ran but nothing else happens. The internal state is reacting to the props change, but we're never forwarding that change onwards to the OL map.

Here's what I think the solution is - 
 - add an `updated()` lifecycle rule to `MyMap`
 - check for changes to lat/long
 - re-center map

This will give us a new method of data flow - from React to the Map, but not just on the initial load.

Docs: https://lit.dev/docs/components/lifecycle/#updated

apps/editor.planx.uk/src/lib/gis.ts
DafyddLlyr: Yep I think we can keep this without context - it's just abstracting away the logic from the component to here.

The implementation in the component will be something like...

```ts
const handleGeocodeChange = (event: GeocodeChangeEvent) => { ... }

const [searchedPoint, setSearchedPoint] = useGeocodeChange("geocode-autocomplete", handleGeocodeChange);
  ```

File: apps/editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx [hunk 3]
```diff
@@ -128,6 +184,15 @@ export default function PlotNewAddress(props: PlotNewAddressProps): FCReturn {
   return (
     <>
       <FullWidthWrapper>
+        <ErrorWrapper error={searchValidationError} id="geocode-autocomplete">
+          {/* @ts-ignore */}
+          <geocode-autocomplete
+            id="geocode-autocomplete"
+            label="Search for an address to position the map"
+            osProxyEndpoint={`${import.meta.env.VITE_APP_API_URL}/proxy/ordnance-survey`}
+            onaddressSelection={geocodeAutocompleteSearchHandler}
```
jessicamcinchak, line 193: Update: While React 19 rebase simplifies this alot, unfortunately still bumping up against same Lit `update` lifecycle I _think_ ! We can search for an address, listen for the selection event, set and validate it in component state, but the map ultimately continues to fail to recognise it's changed and re-render ! 

So going to keep on with WIP map repo approach here instead https://github.com/theopensystemslab/map/pull/618

jessicamcinchak:
Closing in favor of https://github.com/theopensystemslab/map/pull/618 ! Have made a "composite web component mode" with map + search as opposed to trying to hack two-way data flow and Lit "update" life cycles as outlined above.

PR #6457, opened 2026-04-14
refactor(api): Move 'welcome' template from codebase to Resend

DafyddLlyr:
## What does this PR do?
Currently, the "welcome" template lives as a HTML string within a `.ts` file within the codebase. If we need to make any changes, we need to get it through the local → pizza → staging → production pipeline.

This PR moves it to be a Resend-managed template - we can now update via the Resend web GUI (as we do with Notify).

## Motivation
I'm picking this up before https://github.com/theopensystemslab/planx-new/pull/6458 which adds a new template. This sets up some of the groundwork around types etc to make these easier to manage. This largely copies the current Notify `TemplateRegistry` pattern we already have in place.

### DIFF
--- a/apps/api.planx.uk/lib/resend/index.ts
+++ b/apps/api.planx.uk/lib/resend/index.ts
@@ -1,36 +1,23 @@
 import { Resend } from "resend";
-import {
-  templateRegistry,
-  type ResendTemplate,
-  type TemplateParams,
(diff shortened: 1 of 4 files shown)


joecarver (approved):
All looks good to me and certainly makes sense to move templates out of code in order to make them simpler to change.

PR #6458, opened 2026-04-14
feat(api): Migrate "submit" email from Notify to Resend

DafyddLlyr:
## What does this PR do?
This PR implements https://trello.com/c/mj94nv3p/3583-migrate-submission-emails-to-resend

This moves the "submit" template from Notify across to Resend, relying on https://github.com/theopensystemslab/planx-new/pull/6457

**Tests**
I think this is failing based on an old API key in GHA - taking a look now! Tests passing locally.


apps/api.planx.uk/modules/send/email/index.test.ts
```diff
@@ -78,10 +78,8 @@ describe(`sending an application by email to a planning office`, () => {
         teamSlug: "southwark",
         recipient: "planning.office.example@council.gov.uk",
         request: {
-          personalisation: {
-            serviceName: "Apply for something",
-            downloadLink: "https://api.editor.planx.uk/test/123",
-          },
```
DafyddLlyr, line 84: Note: the structure of this audit request is now changed, it's no longer wrapped in the `personalisation` object. I don't think this is a problem...?

apps/api.planx.uk/modules/send/email/index.ts
```diff
@@ -137,18 +131,15 @@ const getSubmitEmailConfig = async ({
...
-        serviceName: flowName,
-        sessionId,
-        applicantEmail: email,
-        downloadLink,
-        address: addressLine || "Address not submitted",
-        projectType: projectType || "Project type not submitted",
-        applicantName,
-        fee: fee || "N/A",
-      },
-      emailReplyToId: teamSettings.emailReplyToId,
```
joecarver, line 151: does it matter that we're no longer supporting custom Reply-To addresses for these emails?

apps/api.planx.uk/modules/send/email/index.test.ts
jessicamcinchak: Agree don't think this is an issue / relied on anywhere 👍

apps/api.planx.uk/modules/send/email/index.ts
jessicamcinchak: Good question! 

Since "submit" emails are used to notify councils of incoming submissions from PlanX, it's actually an erorr/inconsistency that we were using _council-specific_ reply-to email addresses before ! (Eg council would reply-to themselves about a submission from us ♻️)

Applicant-facing templates will _always_ keep `emailReplyToId` on the other hand.

jessicamcinchak (approved):
All looks good to me, thanks for this one !

PR #6467, opened 2026-04-14
feat: unarchive flow functionality

zz-hh-aa:
Still feature flagged!

In sum, creates unarchive functionality in the `Archive` view:

https://github.com/user-attachments/assets/d8acf83d-1b8c-424d-a546-a8ee2fceac6a

More specifically: 
- Creates an `ARCHIVE_FLOW` mutation and `useArchiveFlow` hook / removing the `archiveFlow` method from the Zustand store
- Creates a `GET_FLOWS` mutation too to run as part of mutation `refetchQueries` (question about this below) 
- Refactors `FlowMenu` into `ActiveFlowMenu` and `ArchivedFlowMenu` (the components have some similarity but enough distinction that I decided to keep them separate instead of having a generic component)
- `ArchivedFlowMenu` has an 'Unarchive' menu item (forthcoming PR to add 'Delete')
- Updates styling for actions column in table view

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
@@ -267,9 +267,6 @@ export interface EditorStore extends Store.Store {
   cutNode: (id: NodeId, parent: NodeId) => void;
   getCutNode: () => CutPayload | null;
   addNode: (node: any, relationships?: Relationships) => void;
-  archiveFlow: (
-    flow: FlowSummary,
-  ) => Promise<{ id: string; name: string } | void>;
   connect: (src: NodeId, tgt: NodeId, object?: any) => void;
   connectToFlow: (id: NodeId) => Promise<void>;
   disconnectFromFlow: () => void;
@@ -338,34 +335,6 @@ export const editorStore: StateCreator<
       { children, parent, before },
     )(get().flow);
     send(ops);
-  },
-
-  archiveFlow: async ({ id, slug }) => {
-    try {
-      const { data } = await client.mutate<{
-        flow: { id: string; name: string };
-      }>({
-        mutation: gql`
-          mutation updateFlow($id: uuid!, $slug: String!) {
-            flow: update_flows_by_pk(
-              pk_columns: { id: $id }
-              _set: { archived_at: "now()", status: offline, slug: $slug }
-            ) {
-              id
-              name
-            }
-          }
-        `,
-        variables: {
-          id,
-          slug: `${slug}-archived`,
-        },
-      });
-
-      return data?.flow;
-    } catch (error) {
-      throw Error("Failed to archive flow", { cause: error });
-    }
   },
 
   connect: (src, tgt, { before = undefined } = {}) => {
--- a/apps/editor.planx.uk/src/pages/Team/components/Archive.tsx
+++ b/apps/editor.planx.uk/src/pages/Team/components/Archive.tsx
@@ -23,7 +23,6 @@ type Props = {
   ) => void;
   teamId: number;
   slug: string;
-  fetchFlows: () => void;
 };
 
 const Archive: React.FC<Props> = ({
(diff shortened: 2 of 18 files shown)


apps/hasura.planx.uk/metadata/databases/default/tables/public_flows.yaml
```diff
@@ -460,9 +460,7 @@ update_permissions:
         - summary
         - team_id
         - templated_from
-      filter:
-        archived_at:
-          _is_null: true
```
zz-hh-aa, line 465: Removed filter so that platformAdmins and teamEditors can update a flow where `archived_at` is not null (eg to unarchive it)

apps/editor.planx.uk/src/pages/Team/queries.ts
```diff
@@ -1,6 +1,49 @@
 import { gql } from "@apollo/client";
 import { FlowSummary } from "pages/FlowEditor/lib/store/editor";
 
+export const GET_FLOWS = gql`
```
zz-hh-aa, line 4: Went for a mutation + hook here to keep the flows and archived flow fetches following the same pattern and, more importantly, I could use the Apollo `refetchQueries` prop after the archive or unarchive mutation hooks run to ensure that both lists update. 

But I realise this is kind of repetitive now (eg including the full query itself, especially when we have the `getFlows` query in the store). Is this a use case for gql [fragments](https://github.com/theopensystemslab/planx-new/pull/6421/changes#r3056814254)?

apps/editor.planx.uk/src/routes/_authenticated/app/$team/index.tsx
```diff
@@ -46,7 +46,7 @@ export const Route = createFileRoute("/_authenticated/app/$team/")({
 });
 
 function TeamComponent() {
-  const { flows } = Route.useLoaderData();
+  Route.useLoaderData();
 
-  return <Team flows={flows} />;
```
zz-hh-aa, line 51: If I've understood correctly, the results will be cached already so Team doesn't need to be passed flows explicitly.

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -267,9 +267,6 @@ export interface EditorStore extends Store.Store {
   cutNode: (id: NodeId, parent: NodeId) => void;
   getCutNode: () => CutPayload | null;
   addNode: (node: any, relationships?: Relationships) => void;
-  archiveFlow: (
```
zz-hh-aa, line 270: As [Daf mentioned](https://github.com/theopensystemslab/planx-new/pull/6446#discussion_r3064826200), we want to remove `archiveFlow` from the Zustand store in favour of Apollo hooks

apps/editor.planx.uk/src/pages/Team/components/FlowTable/styles.tsx
```diff
@@ -80,7 +81,7 @@ export const FlowRowLink = styled(CustomLink)(() => ({
 })) as typeof CustomLink;
 
 export const FlowTitleCell = styled(TableCell)(() => ({
-  width: "45%",
+  width: "auto",
```
zz-hh-aa, line 84: Question for @ianjon3s: I fixed the width of `FlowActionsCell` because it was changing between the `Flows` and `Archive` views. But I was still getting a weird width inconsistency before I updated this, and now the `FlowTitleCell` renders smaller on my screen. Is this a problem / is there a better way of handling this?

apps/editor.planx.uk/src/pages/Team/components/Archive.tsx
```diff
@@ -110,8 +109,7 @@ const Archive: React.FC<Props> = ({
                   flow={flow}
                   flows={archivedFlows}
                   key={flow.slug}
-                  refreshFlows={fetchFlows}
-                  showDetails={false}
+                  view={"archive"}
                 />
               ))}
             </DashboardList>
```
zz-hh-aa, line 115: Thinking about renaming this to something more intuitive like `FlowCards` (we have `FlowTable` as the alt view below), any objections?

PR #6468, opened 2026-04-14
chore(lps): Ensure LPS build and deploy is non-blocking

DafyddLlyr:


.github/workflows/pull-request.yml
```diff
@@ -182,43 +182,43 @@ jobs:
...
-  #     - run: pnpm build --site "https://localplanning.${{ env.FULL_DOMAIN }}" --mode pizza
-  #       env:
-  #         ROOT_DOMAIN: ${{ env.FULL_DOMAIN }}
-  #       if: steps.cache-localplanning-build-assets.outputs.cache-hit != 'true'
-  #       working-directory: ${{ env.LOCALPLANNING_SERVICES_DIRECTORY }}
-  
+  build_localplanning_services:
+    name: Build localplanning.services
+    runs-on: ubuntu-24.04
+    continue-on-error: true
```
DafyddLlyr, line 188: This is the only change here - it makes the job non-blocking

.github/workflows/push-main.yml
```diff
@@ -73,58 +73,10 @@ jobs:
           path: ./${{ env.EDITOR_DIRECTORY }}/build
           if-no-files-found: error
 
-  # build_localplanning_services:
```
DafyddLlyr, line 76: The LPS build is now totally stripped from staging and production jobs - it was a leftover from when this was more integrated. I missed this at the time I was working on https://github.com/theopensystemslab/planx-new/pull/5882

Right now we do a build, download the assets, and then do nothing with it - we already have specific `deploy-lps-{{env}}.yaml` actions to handle the build process.

.github/workflows/pull-request.yml
jessicamcinchak: Do we still get an CI error notification if 'continuing' or does it allow it to essentially fail silently?

DafyddLlyr: We don't get a ping to Slack, but it basically should fall silently and show a ⚠️ instead of a ❌

DafyddLlyr:
Merging to `main` so we can test it there are resolve any issues as a follow up - without this other PRs could be blocked.

PR #6383, opened 2026-03-30
feat(editor): Split up loading process for `flow.data` column to improve initial load times

DafyddLlyr:
## What does this PR do?
This PR changes the data loading process for the public-facing `/draft`, `/preview`, and `/published` routes.

### Before
In order to render the basic layout there's some initial date we need such as t
 - team name, logo, settings, theme
 - flow name
 - published flow data (in order to render questions, calculate graph traversal, sections, etc)

This is currently managed via a large data-fetching "waterfall" - all done in a single, blocking, up-front request which means we can't render anything until the request for published flow data is completed.

```
Navigation starts
  └── beforeLoad
        ├── fetchSettingsForPublishedView
        └── getLastPublishedAt
              └── fetchPublishedFlowData                   ← slow, large payload
                    └── store hydrated
                          └── layout renders
                                └── theme visible
                                      └── header visible
                                            └── Questions renders
```

### After
We now remove published flow data from the original `beforeLoad()` request, and lazily load it (non-awaited). Child components (`<PublishedFlow/>`, `<FlattnedFlow/>`) await this data and will not render until we have this data.

This means that we can render the headers, public layout,  SaveAndReturn layouts (mostly, see comments below). This buys us precious time to fetch flow data in the background - by the time a user has navigated the "enter your email" screen it's very likely the fetch has completed already.

```
Navigation starts
  └── beforeLoad
        ├── fetchSettingsForPublishedView + getLastPublishedAt  (merged requests)
        │     └── store hydrated with theme, settings, team, path
        │           └── PublicLayout renders                ← theme, header, footer visible
        │                 └── SaveAndReturnLayout renders   ← "Enter your email" visible
        │                       └── PublishedFlow mounts
        │                             └── useQuery fires (already prefetched, in-flight or completed)
        │                                   └── Questions renders
        └── loader (fire and forget)
              └── prefetch flow.data                        ← starts in parallel with above
```

## Next steps...
This is very much a first pass at this, with an initial set of improvements. We can build on this further in a number of ways - 
 - progressively loading flow data (e.g. first N questions, or first N "layers" of the graph)
 - pre-calculate section data on publish
 - improve reconciliation request to not be dependent on `flow.data` in the frontend (see comments)

## Demo
Not exactly scientific, but here's a video of side-by-side loading, with this Pizza on the right hand side - 

https://github.com/user-attachments/assets/618fc74f-e141-4428-9a4b-aa9825fe52f5

## Testing
[...]

PR #6469, opened 2026-04-14
feat(editor): Split up loading process for `flow.data` column to improve initial load times

DafyddLlyr:
Reverts theopensystemslab/planx-new#6459

## What's the problem?
On custom subdomains, flow data wasn't being fetched once https://github.com/theopensystemslab/planx-new/pull/6383 was merged. The page loaded, alongside other metadata (settings, theme, flow metadata etc), but no `flow.data`).

## What's the solution?
Ensure we call the same loader on the `/_public/_customDomain/$flow` route - to match `/_public/_planXDomain/$team/$flow/published`

Only change here from#6383 is the following commit - [24dd088](https://github.com/theopensystemslab/planx-new/pull/6469/commits/24dd0880030b78b55026871d23cb1ddacac2ce58)

## How was this missed?
I guess there's a few culprits here - 
 - No E2E tests for any custom subdomain implementation
 - No custom subdomains on staging
 - There's no hard link to enforce shared behaviour between these two routes


jessicamcinchak (approved):
Retried a few times to get pizza up here, but think there's an LPS dependency with other CI refactor branch! 

Happy to approve as it's a revert-revert and we talked this through - thank you for fix!!

PR #6478, opened 2026-04-16
chore: react 19 & mui 9

joecarver:
### What's in this PR?

- upgrade React to v19
- upgrade mui and mui-x to v9
- fix all the API and typing changes for the MUI upgrade - this affected a lot of components hence the large diff, the changes are fairly repetitive so here's a summary of the big things to look out for:

#### use `sx` for all style props
- direct shorthand style props e.g. `mt` `fontWeight` moved into `sx` prop   
- e.g. `<Typography mb={2}> -> <Typography sx={{ mb: 2 }}>`

#### `*Props` -> `slotProps`
- e.g. `InputProps` -> `slotProps.input` 

#### Icon import names
e.g. `@mui/icons-material/ErrorOutline -> ErrorOutlined`


apps/editor.planx.uk/package.json
```diff
@@ -62,12 +62,12 @@
     "ot-json0": "^1.1.0",
     "postcode": "^5.1.0",
     "ramda": "^0.29.1",
-    "react": "^18.3.1",
-    "react-beautiful-dnd": "^13.1.1",
+    "react": "^19.2.4",
+    "@hello-pangea/dnd": "^18.0.1",
```
joecarver, line 66: this is a fork of `react-beautiful-dnd`, the original was archived before adding React 19 support. So this package provides a drop-in replacement

apps/api.planx.uk/package.json
```diff
@@ -12,7 +12,7 @@
     "@airbrake/node": "^2.1.9",
     "@aws-sdk/client-s3": "^3.1000.0",
     "@aws-sdk/s3-request-presigner": "^3.1000.0",
-    "@opensystemslab/planx-core": "github:theopensystemslab/planx-core#025b73e",
+    "@opensystemslab/planx-core": "github:theopensystemslab/planx-core#460a548",
```
DafyddLlyr: Once https://github.com/theopensystemslab/planx-core/pull/977 is merged to `main` on `planx-core` please update this ref to the latest commit on `main`.

This will make sure that merging this to `planx-new` doesn't overwrite anything that merged to `planx-core` between now and when this progresses.

apps/editor.planx.uk/src/@planx/components/Content/Public.tsx
```diff
@@ -38,10 +38,6 @@ const Content = styled(Box, {
   },
 }));
 
-Content.defaultProps = {
-  color: "#ffffff",
-};
```
DafyddLlyr, line 43: Good to tidy this up...! `defaultProps` has been deprecated for a while.

apps/editor.planx.uk/src/@planx/components/shared/Schema/InputFields/ChecklistFieldInput.tsx
```diff
@@ -43,7 +43,7 @@ export const ChecklistFieldInput: React.FC<Props<ChecklistField>> = (props) => {
           <FieldInputDescription description={data.description} />
         )}
         {data.options.map((option) => (
-          <Grid item xs={12} key={option.id}>
+          <Grid size={12} key={option.id}>
```
DafyddLlyr, line 46: We're missing the `item` property here.

Looking at the docs, I think this should be `<Item/>` in MUI v9 instead of the old pattern of `<Grid item/>`.

Docs: https://mui.com/material-ui/react-grid/

apps/editor.planx.uk/src/components/Header/Header.test.tsx
DafyddLlyr, line 1: Please see https://github.com/theopensystemslab/planx-new/pull/6494

This looks like a regression that's snuck in during a rebase possibly? These properties should not be reintroduced.

apps/editor.planx.uk/src/lib/gis.ts
```diff
@@ -33,7 +33,14 @@ export const useGeoJSONChange: UseGeoJSONChange = (mapId, callback) => {
       callback(event);
     };
 
-    const map: HTMLElement | null = document.getElementById(mapId);
+    // React 19 sets `id` as a DOM property (not HTML attribute) on custom elements
+    // in some environments (e.g. jsdom), causing getElementById to return null.
+    // Fall back to querying by data-map-id attribute which is always set as an attribute.
```
DafyddLlyr, line 38: Helpful comment!

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Checklist.tsx
```diff
@@ -127,7 +127,9 @@ const Checklist: React.FC<Props> = React.memo((props) => {
             }}
             preload={false}
             onContextMenu={handleContextMenu}
-            ref={drag}
+            ref={(el) => {
+              drag(el);
+            }}
```
DafyddLlyr, line 132: nit: I think this is a no-op change, these are the same before and after. Same on the files below.

apps/editor.planx.uk/src/ui/editor/RichTextInput/RichTextInput.stories.tsx
```diff
@@ -25,7 +25,7 @@ export const Basic = () => {
 
   return (
     <Grid container>
-      <Grid item xs={4}>
+      <Grid size={4}>
```
DafyddLlyr, line 28: Please see previous question on `<Grid/>`!

apps/editor.planx.uk/src/ui/editor/ModalSectionContent.tsx
```diff
@@ -71,8 +71,8 @@ export default function ModalSectionContent({
 }: Props): FCReturn {
   return (
     <SectionContentGrid container>
-      <LeftGutter item>{Icon && <Icon />}</LeftGutter>
-      <SectionContent item>
+      <LeftGutter>{Icon && <Icon />}</LeftGutter>
+      <SectionContent>
```
DafyddLlyr, line 75: Same grid question as before - do we need the `item` prop here?

apps/editor.planx.uk/src/ui/public/Banner.tsx
```diff
@@ -14,7 +14,12 @@ interface BannerProps {
   children?: React.ReactNode;
 }
 
-const Root = styled(Box)(({ theme, color, bgcolor }) => ({
+interface RootProps {
+  $bgcolor?: string;
+  $color?: string;
+}
```
DafyddLlyr, line 20: MUI doesn't handle the `$` syntax for styled-components natively unfortunately (unless added in v8 or v9!). To exclude these from the DOM we still need to use the `shouldForwardProp` syntax.

Docs: https://mui.com/material-ui/customization/how-to-customize/#dynamic-overrides

apps/editor.planx.uk/package.json
DafyddLlyr: Thanks...! Been meaning to look at https://github.com/atlassian/pragmatic-drag-and-drop as an alternative, but this looks like a great (and simple!) switch 👍

apps/editor.planx.uk/src/index.tsx
```diff
@@ -87,6 +87,16 @@ root.render(
...
+    <ToastContainer
+      icon={false}
+      theme="colored"
+      position="top-right"
+      transition={Bounce}
+      autoClose={5000}
+      closeOnClick
+      pauseOnHover
+      pauseOnFocusLoss
+      draggable
```
DafyddLlyr: What's the motivation for this change?

DafyddLlyr (changes requested):
Thanks for this - many changes...! Worked through the files but not yet tested out the Pizza.

Some initial feedback and questions - all pretty small 👍

apps/editor.planx.uk/src/index.tsx
joecarver: Ah yes this probably could have done with a comment! React 19 removed support for defaultProps, which is how toastify previously set these properties. Passing them manually fixed the issue. 

Although I just had a look at the `react-toastify` releases and I think upgrading the package to v10 or 11 would be a better shout https://github.com/fkhadra/react-toastify/releases (ctrl+f `defaultProps`)

apps/editor.planx.uk/src/@planx/components/shared/Schema/InputFields/ChecklistFieldInput.tsx
joecarver: hmm, I removed the `item` prop as it's no longer on the `<Grid />` api - my assumption was that its no longer necessary and that MUI will figure it out automatically.

The docs are a bit misleading,  there is no `<Item />` component - in the example given, Item is just a styled `<Paper />`. Each one is wrapped by a `Grid` (without item or paper so I'm thinking that items are just implicit now.

apps/editor.planx.uk/src/components/Header/Header.test.tsx
joecarver: ah yes - had added these to get the tests passing with the version of planx-core, will remove

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Checklist.tsx
joecarver: yeah, no functionality change but this is to satisfy the type checker. the `drag` function returns a value but `ref` now strictly only accepts functions like `void | (() => VoidOrUndefinedOnly)` - so we need to discard the return value for the type checker to be happy.

apps/editor.planx.uk/src/index.tsx
joecarver: have updated the package and reverted this change now [865ee4c](https://github.com/theopensystemslab/planx-new/pull/6478/commits/865ee4c08eed40c0e988cf3060a90d2bd46092dd)

apps/editor.planx.uk/package.json
joecarver: yeah, i saw some nice altneratives also... was intending to ask about it. in fact, we're using two separate DnD packages, [react-dnd](https://react-dnd.github.io/react-dnd/about) and this one. I assume there is a good reason for this but didn't want to go down the path of potentially aligning them in the scope of this ticket as they both have a fair few separate usages.

DafyddLlyr: Yeah good catch - I think this is just a historic oversight? (cc @jessicamcinchak)

`react-beautiful-dnd` is only used in `<ListManager/>` now, but `react-dnd` is used more widely for graph interactions.

Agreed that this ticket is not the time to address it, but it seems like a sensible future rationalisation 👍

apps/editor.planx.uk/src/@planx/components/shared/Schema/InputFields/ChecklistFieldInput.tsx
DafyddLlyr: Yep, that makes sense thanks - I just scanned the docs and missed that `<Paper/>` setup.

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Checklist.tsx
DafyddLlyr: Thanks for the explanation 👍

apps/editor.planx.uk/src/index.tsx
DafyddLlyr: Brill - thank you!

DafyddLlyr (approved):

DafyddLlyr:
Looking great @joecarver  - thanks for addressing all those points.

apps/editor.planx.uk/package.json
jessicamcinchak: Yes ! `react-beautiful-dnd` decision from before our time https://github.com/theopensystemslab/planx-new/commit/65a00c5d4459b663f01666ddb4e08af8cb472225 (no special requirements here afaik)

Agree future follow-up to align would be very welcome !

PR #6482, opened 2026-04-16
feat: unarchive flow functionality

zz-hh-aa:
Still feature flagged! In sum, creates unarchive functionality in the Archive view, works for pinned flows too.

https://github.com/user-attachments/assets/93ab8a79-ec25-4c7e-b0ca-4139061440de

# What does this PR do? 
More specifically:
- Deletes `archiveFlow` method from the store and creates mutation / hook for it
- Refactors `FlowMenu` into `ArchivedFlowMenu` and `ActiveFlowMenu` (tried making a generic component but the two seemed distinct enough to warrant separate ones)
- Uses Apollo `refetchQueries` param to ensure that list of flows updates after archive / unarchive action (and refactors the pin / unpin flow mutations to do the same)
- ArchivedFlowMenu has an 'Unarchive' menu item (forthcoming PR to add 'Delete')

# Questions
- I guess we want the 'Actions' column to be uniform width across Flows and Archive views? 
- If we unarchive a previously pinned flow, is the expected behaviour that it keeps being pinned? Or should we clear this also?


apps/editor.planx.uk/src/pages/Team/components/FlowCard/queries.ts
```diff
@@ -1,9 +1,10 @@
 import { useMutation } from "@apollo/client";
 import gql from "graphql-tag";
 import { FlowSummary } from "pages/FlowEditor/lib/store/editor";
+import { GET_FLOWS } from "pages/Team/queries";
 
 // this is copied from the getFlows query in FlowEditor/lib/store/editor.ts - might be worth re-using the fragment there as well?
-const FLOW_SUMMARY_FIELDS = gql`
+export const FLOW_SUMMARY_FIELDS = gql`
```
zz-hh-aa: I have a general question about where is best to keep reusable fragments like this, and also about best practice / intended patterns--eg in some places we keep queries and hooks separate, in others in the same file.

apps/editor.planx.uk/src/pages/Team/queries.ts
```diff
@@ -1,53 +1,57 @@
 import { gql } from "@apollo/client";
 import { FlowSummary } from "pages/FlowEditor/lib/store/editor";
 
+import { FLOW_SUMMARY_FIELDS } from "./components/FlowCard/queries"
+
+export const GET_FLOWS = gql`
```
zz-hh-aa: Even though we have this method in the store, I went for a mutation + hook here to keep the fetch pattern the same across Flows and Archive views. And, more importantly, I could use the Apollo refetchQueries prop after the archive or unarchive mutation hooks run to ensure that both lists update.

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -268,9 +268,6 @@ export interface EditorStore extends Store.Store {
   cutNode: (id: NodeId, parent: NodeId) => void;
   getCutNode: () => CutPayload | null;
   addNode: (node: any, relationships?: Relationships) => void;
-  archiveFlow: (
```
zz-hh-aa, line 271: As https://github.com/theopensystemslab/planx-new/pull/6446#discussion_r3064826200, we want to remove archiveFlow from the Zustand store in favour of Apollo hooks

apps/editor.planx.uk/src/pages/Team/components/FlowCard/queries.ts
DafyddLlyr: I don't think there's any hard rule for this - if a hook is going to be reusable, it makes sense to place it in a shared location. The same goes for fragments and queries. We could (should?) enforce something like the ESLint rule `import/no-restricted-paths` for this.

Queries do not need to be in their own files if they're single use, but it can be helpful to do so. We could also write them a `.gql` files and apply formatting / validation but that's a bit trickier on this version of ESLint.

Eventually, we'll migrate them all to a shared library. The only key things really are - 
 - Unique query and mutation names (very helpful for debugging)
 - Wrap in `gql` tags

apps/editor.planx.uk/src/pages/Team/components/FlowCard/queries.ts
```diff
@@ -76,19 +77,28 @@ interface UnpinFlowMutation {
 
 interface PinFlowVars {
   flowId: string;
+  teamId: number;
```
DafyddLlyr, line 43: Do we need to pass this in as a prop, or can we just get from store (as we do for userId)?

apps/editor.planx.uk/src/pages/Team/components/FlowTable/index.tsx
```diff
@@ -30,37 +32,33 @@ interface FlowTableProps {
...
-  refreshFlows,
-  showDetails,
-  updateFlow,
+  view,
 }) => {
-  const [userId] = useStore((state) => [state.user?.id]);
   const { headerText } = useFlowSortDisplay();
 
-  const showPinnedColumn = updateFlow && userId;
+  const showPinnedColumn = view === "flows";
```
DafyddLlyr: I don't quite follow this - it's always going to be `true` as it's nested within the same condition.

apps/editor.planx.uk/src/pages/Team/components/FlowTable/index.tsx
```diff
@@ -108,11 +100,12 @@ const FlowTableRow: React.FC<FlowTableRowProps> = ({
   } = useFlowMetadata(flow);
 
   const { displayTimeAgo, displayActor } = useFlowDates(flow);
+  const isRowLinkActive = view === "flows" ? true : false;
 
-  const showPinnedColumn = updateFlow && userId;
+  const showPinnedColumn = view === "flows";
```
DafyddLlyr: nit: Repeared conditions

apps/editor.planx.uk/src/pages/Team/components/ArchivedFlowMenu.tsx
```diff
@@ -0,0 +1,77 @@
...
+            await unarchiveFlow();
+            toast.success("Unarchived flow");
+        } catch (error) {
+            toast.error(
+            "We are unable to unarchive this flow, refesh and try again or contact an admin",
+            );
+        } finally {
+            setOpenFlowDialog(null);
+        }
+        };
```
DafyddLlyr: nit: Please run linting here (we need to fix this running on pre-commit...! 😅) - there's a few funky spacings and brackets.

apps/editor.planx.uk/src/pages/Team/components/ArchivedFlowMenu.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { FlowMenuProps } from "./StyledSimpleMenu";
+import { StyledSimpleMenu } from "./StyledSimpleMenu";
+
+const ArchivedFlowMenu: React.FC<FlowMenuProps> = ({
+    flow,
+    variant = "card",
+    teamId,
+}) => {
+    type OpenFlowDialog = "unarchive";
+    const [openFlowDialog, setOpenFlowDialog] = useState<OpenFlowDialog | null>(null);
```
DafyddLlyr: nit: This is just a complex way to handle a boolean I think...! 

There's just two possible states - `"unarchive"` or `null`.

```suggestion
    const [isOpen, setIsOpen] = useState<boolean>(false);
```

apps/editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -36,24 +37,23 @@ interface TeamProps {
...
     state.canUserEditTeam,
-    state.getFlows,
     state.teamSettings?.isTrial,
     state.flowCardView,
     state.setFlowCardView,
   ]);
 
-  const [flows, setFlows] = useState<FlowSummary[] | null>(initialFlows);
+  const { data } = useGetFlows(teamId);
+  const flows = data?.flows ?? null;
```
DafyddLlyr: Have a look at the loader for this route - we're fetching a list of flows there upfront so we're guaranteed it will always be present.

If we _just_ rely on the loader data it will go stale as soon as we change anything, and we'll manually need to call `router.invalidate()` (e.g. on each archive, delete, rename, add). This is not what we want - we want nice reactive data that automatically updates as we make changes.

This pattern of adding an Apollo hook is correct - it will mean that as we make calls, `data.flows` will update and so will our UI. The bit that's missing here though is that this component now treats `flows` as potentially falsy - we need to optionally chain all the time (e.g. `flows?.filter()`).

If we make this one change, I think we'll see the behaviour we want and our code will more accurately represent the state of the real data - 


```suggestion
  const { data } = useGetFlows(teamId);
  const flows = data?.flows ?? initialFlows;
```

Not sure how clearly I've laid this out - give it a try and let me know if you want to talk through 👍

apps/editor.planx.uk/src/pages/Team/queries.ts
DafyddLlyr: Please see previous comment about loader data and this hook - this is great!

We can update the loader to also use this, and drop the store method 👍

DafyddLlyr (changes requested):
Looking good..! A few minor changes and comments below.

apps/editor.planx.uk/src/pages/Team/components/ArchivedFlowMenu.tsx
DafyddLlyr: Aha - I've just this this modal again in https://github.com/theopensystemslab/planx-new/pull/6483/changes and this makes a lot more sense now sorry...! 😅 

Please feel free to ignore this comment 👍

apps/editor.planx.uk/src/pages/Team/components/FlowCard/queries.ts
zz-hh-aa: We could, but since `FlowPinButton` siblings also need `teamId` I thought it would make more sense to call it in the parent and pass to multiple children, instead of calling it multiple times, in `FlowPinButton`, `ArchivedFlowMenu`, `ActiveFlowMenu`... does that make sense?

apps/editor.planx.uk/src/pages/Team/index.tsx
zz-hh-aa: Done in [61a0313](https://github.com/theopensystemslab/planx-new/pull/6482/commits/61a0313b32c1293ad66b16a220818596207e8833) but also wondering if the [navigating to empty teams](https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1776676527468709) bug will affect this? 

I quickly tried a simple `network-only` request for `GET_FLOWS` but that didn't fix it.

apps/editor.planx.uk/src/pages/Team/components/ArchivedFlowMenu.tsx
zz-hh-aa: `lint:fix` didn't seem to do it! Fixed manually in [5ad1f60](https://github.com/theopensystemslab/planx-new/pull/6482/commits/5ad1f60c29292e38e25cd417aa99af315d34e2c7)

apps/editor.planx.uk/src/pages/Team/components/FlowTable/index.tsx
zz-hh-aa: Done in [58d57b8](https://github.com/theopensystemslab/planx-new/pull/6482/commits/58d57b8851d048a18e8dca5b5f011ebd1d3d70a2)

apps/editor.planx.uk/src/pages/Team/components/FlowTable/index.tsx
zz-hh-aa: Done in [58d57b8](https://github.com/theopensystemslab/planx-new/pull/6482/commits/58d57b8851d048a18e8dca5b5f011ebd1d3d70a2)

apps/editor.planx.uk/src/pages/Team/components/FlowCard/queries.ts
DafyddLlyr: Yep - makes sense!

DafyddLlyr (approved):

PR #6484, opened 2026-04-16
chore: Setup Canterbury custom subdomain

DafyddLlyr:
- Verified via AWS CM
- Certs added via AWS SM

Once deployed to prod, I'll get the IT team to create CNAME records.

### DIFF
--- a/infrastructure/application/index.ts
+++ b/infrastructure/application/index.ts
@@ -117,6 +117,10 @@ const CUSTOM_DOMAINS: CustomDomain[] =
           domain: "planningservices.horsham.gov.uk",
           name: "horsham",
         },
+        {
+          domain: "planningservices.canterbury.gov.uk",
+          name: "canterbury",
+        },
       ]
     : [];
 


infrastructure/application/index.ts
```diff
@@ -117,6 +117,10 @@ const CUSTOM_DOMAINS: CustomDomain[] =
           domain: "planningservices.horsham.gov.uk",
           name: "horsham",
         },
+        {
+          domain: "planningservices.canterbury.gov.uk",
+          name: "canterbury",
+        },
```
joecarver, line 123: question: is this all that's required for a new client subdomain? what happens behind the scenes when this config is  applied?

zz-hh-aa: @joecarver docs are [here](https://github.com/theopensystemslab/planx-new/blob/main/doc/how-to/how-to-setup-custom-subdomains.md) for the current process but [soon to be updated](https://trello.com/c/8k8xg8AI) by Dan!

zz-hh-aa (approved):

PR #6490, opened 2026-04-17
fix(iac): Invalidate CloudFront cache after deployments

DafyddLlyr:
## What's the problem?
Ticket: https://trello.com/c/2INGMHaM/3519-investigate-blank-screen-when-loading-production-homepage

After a deploy to staging or production, it's possible to see a "white screen of death" - this is happening because CloudFront can serve a stale `index.html` for up to 10 minutes (the current cache TTL). Vite generates random filenames that change every build and a stale `index.html` will point to old files, causing the issue.

## What's the solution?
Invalidate the cache on each deploy. We handle this here for LPS but the step is missing for PlanX.

https://github.com/theopensystemslab/planx-new/blob/main/scripts/deploy-lps.sh#L47-L50

## Next steps...
Let's test this on merge to `main` and `production` - we should see the action running, see an invalidation in the AWS console, and not see the blank white screen on deploys.

### DIFF
--- a/infrastructure/application/index.ts
+++ b/infrastructure/application/index.ts
@@ -490,6 +490,7 @@ export = async () => {
 
   return {
     customDomains,
+    cdnDistributionId: cdn.id,
     metabaseServiceName: metabaseService.service.name,
     hasuraServiceName: hasuraService.service.name,
     apiServiceName: apiService.service.name,
--- a/.github/workflows/push-main.yml
+++ b/.github/workflows/push-main.yml
@@ -107,6 +107,12 @@ jobs:
           edit-pr-comment: true
         env:
           PULUMI_ACCESS_TOKEN: ${{ secrets.PULUMI_ACCESS_TOKEN }}
+      - name: Invalidate CloudFront cache
+        env:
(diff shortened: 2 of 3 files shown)


infrastructure/application/index.ts
```diff
@@ -490,6 +490,7 @@ export = async () => {
 
   return {
     customDomains,
+    cdnDistributionId: cdn.id,
```
DafyddLlyr, line 493: **Note:** This is just the main `editor.planx.(dev|uk)` CDN - this is intentional for a few reasons.

- We're working to retire a mapping of one CDN to one custom subdomain. It seemed redundant to handle this per-CDN here to just remove it in a few days/weeks
- Running this fix in only some envs should allow us to A/B test if it's working as expected

jessicamcinchak (approved):
Great - happy to test on staging here 🤞

DafyddLlyr:
Thanks - will merge this on a separate run from https://github.com/theopensystemslab/planx-new/pull/6491 just to keep the number of variables down on these deployments.

PR #6492, opened 2026-04-17
docs(adr): Migrate secrets to AWS Secrets Manager

DafyddLlyr:
Thoughts very welcome - we can talk through next dev call also.

### DIFF
--- a/doc/architecture/decisions/0013-migrate-secrets-to-aws-secrets-manager.md
+++ b/doc/architecture/decisions/0013-migrate-secrets-to-aws-secrets-manager.md
@@ -1,0 +1,110 @@
+# 13. Migrate secrets to AWS Secrets Manager
+
+Date: 2026-04-17
+
+## Status
+
+Proposed
+
+Extends [4. Secret Generation](0004-secret-generation.md)
+
+## Context
+
+PlanX secrets currently live in two places -
+
+**Platform secrets** — API keys, database credentials, and service tokens used by the application itself — are stored in `Pulumi.staging.yaml` and `Pulumi.production.yaml` as Pulumi-encrypted blobs. They are decrypted at deploy time using a passphrase held in a Pulumi personal access token (PAT), and injected into Fargate containers as plaintext environment variables. This means that -
+
+- Every developer with access to the Pulumi PAT can decrypt every production secret.
+- Secrets are deployment-sensitive - tightly coupled to the CI/CD pipeline. We need a full production deploy to rotate a secret which can prove challenging.
+- There is no auto-rotation support and no audit trail for access.
+
+**Team integration secrets** — per-council API keys for GOV.UK Pay, BOPS, Power Automate, and file upload services — are stored as encrypted values in the `team_integrations` table.
+
+- We are maintaining bespoke encryption logic rather than delegating to a managed service.
+- Updating a team secret (e.g. a council rotating their GOV.UK Pay token) requires developer involvement, as there is no possible self-service path.
+- Staging and production secrets sit in separate columns of the same table, which creates environmental confusion (see [PR #2499](https://github.com/theopensystemslab/planx-new/pull/2499) for history).
+
+AWS Secrets Manager (SM) is already partially in use - custom domain SSL certificates were recently migrated to AWS SM and are now fetched in `infrastructure/application/index.ts` via `aws.secretsmanager.getSecretVersion`.
+
+## Decision
+
+We will migrate secrets from Pulumi config files to AWS Secrets Manager, in three steps.
+
+### Injecting secrets into Fargate
+
+```typescript
+// Before
+environment: [
+  {
+    name: "GOVUK_NOTIFY_API_KEY",
+    value: config.requireSecret("govuk-notify-api-key"),
+  },
+];
+
+// After
+const govukNotifySecret = await aws.secretsmanager.getSecret({
+  name: "govuk-notify-api-key",
+});
+
+secrets: [
+  {
+    name: "GOVUK_NOTIFY_API_KEY",
+    valueFrom: govukNotifySecret.arn,
+  },
+];
+```
+
+### Step 1 - Pilot: GOV.UK Notify API key
+
+We need to rotate this key just now, so it seems like a sensible first step to try out on staging. We can move `govuk-notify-api-key` from `Pulumi.staging.yaml` to AWS SM. Next we would update `infrastructure/application/services/api.ts` to reference it (see above).
+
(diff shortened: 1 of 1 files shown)


doc/architecture/decisions/0013-migrate-secrets-to-aws-secrets-manager.md
```diff
@@ -0,0 +1,110 @@
...
+
+If all goes well, we repeat this for `Pulumi.production.yaml`
+
+### Step 2 - Platform secrets
+
+If all goes well, we can then migrate the remaining secrets from `Pulumi.staging.yaml` and `Pulumi.production.yaml` to AWS SM using the same pattern.
+
+After migration, the Pulumi config files will contain only non-sensitive values (CPU allocations, domain names, URLs, scaling parameters) - the `secure:` entries will be gone.
+
+We do not need to complete this as a "big bang", as we rotate secrets we can move them to the new method.
```
joecarver, line 71: This jumped out at me as potentially a scenario we'd want to avoid... if we gradually migrate secrets then we'll have secrets living in 3 places for some undetermined period of time. This feels slightly more prone to wrong secret bugs and devX pain points.

Would it be too long to do them all in one go? Are there any other issues we might have with an all-in-one change?

doc/architecture/decisions/0013-migrate-secrets-to-aws-secrets-manager.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Step 3 - Team integration secrets
+
+This can happen later, and in-line with https://trello.com/c/A8gB8rSx/3177-configure-council-api-keys-for-file-download-via-teamintegrations-table
+
+We would aim to retire the secrets held in the `team_integrations` table and migrate secrets into AWS SM (using the `teams/{slug}/{key}` naming pattern to group them).
+
+This would open the door to self-service secret management via the PlanX settings pages - a team admin could update their GOV.UK Pay token through the UI, which writes to SM via the API, without developer involvement or a deployment being triggered. This is not currently possible when secrets live in Pulumi config or in `team_integrations`.
```
joecarver, line 79: Just to flag that we'll need proper permission scoping against the secret manager here. I assume its straightforward using the regular AWS IAMs approach to scope an application role to specific sets of secrets.

Also, I suppose as we'd just be using one IAMs role for handling all team integration secrets, we'd also need application-level checks in our code, confirming that the user is updating secrets for a team they have access to.

joecarver (commented):
I think this seems like a good idea and will have nice long term benefits. My only concern would be with the gradual migration approach introducing even more fragmentation, so perhaps we can explore doing it all in one...

doc/architecture/decisions/0013-migrate-secrets-to-aws-secrets-manager.md
```diff
@@ -0,0 +1,110 @@
...
+    valueFrom: govukNotifySecret.arn,
+  },
+];
+```
+
+### Step 1 - Pilot: GOV.UK Notify API key
+
+We need to rotate this key just now, so it seems like a sensible first step to try out on staging. We can move `govuk-notify-api-key` from `Pulumi.staging.yaml` to AWS SM. Next we would update `infrastructure/application/services/api.ts` to reference it (see above).
+
+Once deployed, we can check it's working and then try out a rotation by updating the SM value and restarting the Fargate task.
```
freemvmt, line 61: Does a straight restart of a Fargate task on production imply downtime? I don't know what the proper blue/green method looks like in the console - maybe 'Force new deployment' (of the service)? not sure if we need to say that more explicitly?

doc/architecture/decisions/0013-migrate-secrets-to-aws-secrets-manager.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Step 3 - Team integration secrets
+
+This can happen later, and in-line with https://trello.com/c/A8gB8rSx/3177-configure-council-api-keys-for-file-download-via-teamintegrations-table
+
+We would aim to retire the secrets held in the `team_integrations` table and migrate secrets into AWS SM (using the `teams/{slug}/{key}` naming pattern to group them).
+
+This would open the door to self-service secret management via the PlanX settings pages - a team admin could update their GOV.UK Pay token through the UI, which writes to SM via the API, without developer involvement or a deployment being triggered. This is not currently possible when secrets live in Pulumi config or in `team_integrations`.
+
+Once all team secrets are in SM, the `team_integrations` columns become redundant and can be removed. The `application:encryption-key` Pulumi secret is also retired in step 3, along side the `encrypt` and `decrypt` scripts.
```
freemvmt, line 81: 🔪 😁

doc/architecture/decisions/0013-migrate-secrets-to-aws-secrets-manager.md
```diff
@@ -0,0 +1,110 @@
...
+  --secret-id govuk-notify-api-key \
+  --secret-string "new-value-here"
+```
+
+This isn't as convenient as `pulumi config (set|get) so we could always wrap in a little bash script if we wanted to.
+
+## Consequences
+
+- Rotating a secret is decoupled from a Pulumi deployment
+- Auto-rotation would be possible - right now this is done quite ad-hoc (developer offboarding)
```
freemvmt, line 104: 👏🏼

doc/architecture/decisions/0013-migrate-secrets-to-aws-secrets-manager.md
```diff
@@ -0,0 +1,110 @@
...
+- Access to production secrets can be scoped by IAM role rather than being available to anyone holding the Pulumi PAT.
+- The `Pulumi.staging.yaml` and `Pulumi.production.yaml` files become safe to read without concern about encrypted blobs leaking sensitive information about the shape of the production config (no risk of forgetting `--secret` and committing something)
+- Step 3 removes the requirement to maintain our own encryption logic
+- Step 3 is a prerequisite for self-service secret management in the admin platform (if we want this in future)
+- AWS SM has a small per-secret cost (~$0.40/secret/month) - not a huge cost for the gains I'd argue?
```
freemvmt, line 109: so for approx 100 secrets currently in pulumi across both envs we'd project monthly cost ~ $40

freemvmt (approved):
Sounds like a lot of great improvements and wins tied up in this! Do you think there's any downsides except the additional line on AWS bill?

DafyddLlyr:
Closed in favour of #7066 - right now capacity feels a bit tight to manage this + other migrations (e.g. monorepo). Would still like to do this, but we can revisit in future.

PR #6493, opened 2026-04-17
chore(editor): Remove final instances of `CoreDomainClient`

DafyddLlyr:
## What does this PR do?
- Removes final references of `CoreDomainClient` from the frontend - now all requests are made via the API or Hasura directly (via Apollo) - not via `planx-core`
- Adds hook for team creation
- Removes deprecated `fetchCurrentTeam()` call

## Motivation
I realised whilst working on https://github.com/theopensystemslab/planx-new/pull/6450 that there were just two references to this left and removing it would be another step along the path towards https://github.com/orgs/theopensystemslab/projects/7

### DIFF
--- a/apps/editor.planx.uk/src/lib/feedback.ts
+++ b/apps/editor.planx.uk/src/lib/feedback.ts
@@ -27,10 +27,9 @@ export async function getInternalFeedbackMetadata(): Promise<FeedbackMetadata> {
     breadcrumbs,
     currentCard: node,
     computePassport,
-    fetchCurrentTeam,
+    teamId,
     id: flowId,
   } = useStore.getState();
-  const { id: teamId } = await fetchCurrentTeam();
   const userData = {
     breadcrumbs: breadcrumbs,
     passport: computePassport(),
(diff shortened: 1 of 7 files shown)


apps/editor.planx.uk/src/lib/feedback.ts
```diff
@@ -27,10 +27,9 @@ export async function getInternalFeedbackMetadata(): Promise<FeedbackMetadata> {
     breadcrumbs,
     currentCard: node,
     computePassport,
-    fetchCurrentTeam,
+    teamId,
     id: flowId,
   } = useStore.getState();
-  const { id: teamId } = await fetchCurrentTeam();
```
DafyddLlyr, line 33: I've looked back through git blame to try and work out the origin of this - I believe it was from the old team settings methods, and just incorrectly used here.

`teamId` in the store will always match the `teamSlug` which `fetchCurrentTeam()` is using to look this up.

apps/editor.planx.uk/src/pages/Teams/hooks/useCreateTeam.tsx
DafyddLlyr, line 1: This matches https://github.com/theopensystemslab/planx-core/blob/main/src/requests/team.ts#L80-L134

jessicamcinchak (approved):

PR #6496, opened 2026-04-17
fix: Tidy up zindex and hover states for flow table

ianjon3s:
## What does this PR do?

- Addresses a number of small style regressions with the flow table

### Hover state obscures table row contents

Current:
image

Fixed:
image

(same issue applies for cards, also fixed)

### zIndex overlap for table items

Current:
image

Fixed:
image

(also centre-aligned the pin icon)


DafyddLlyr (approved):
Thanks for picking these up - always really appreciate these PRs..!

PR #6497, opened 2026-04-17
feat: add copy/paste help text operation

augustlindemer:
This PR adds 'copy help text' and 'paste help text' operations to the editor node context menu. It only shows these operations when help text is toggled to display in the graph.

Currently manually excludes node types without help text. See comment.


apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/ContextMenu.tsx
```diff
@@ -142,7 +168,28 @@ export const ContextMenu: React.FC = () => {
       parentIsTemplatedInternalPortal || parentIsChildOfTemplatedInternalPortal;
 
     if (source === "node") {
-      return [
+      const node = self ? getNode(self) : null;
+      const excludedTypes = [
```
augustlindemer, line 172: Thoughts on if there are more robust ways of excluding node types without help text without having to explicitly list them out? I checked if we have a 'hasHelpText' kind of value for node types but I don't think we do. Do we want to check the target node to see if it has help text props (like `howMeasured`) instead?

jessicamcinchak: Good question! This does _not_ exist yet afaik ! 

This is currently tricky for a couple of reasons:
- As I understand it, this is how we currently type `NodeData` in the graph
```ts
export declare const NODE_TAGS: readonly ["toReview", "automation", "analytics", "sensitiveData"];
export type NodeTag = (typeof NODE_TAGS)[number];
export type NodeTags = {
    tags?: NodeTag[];
};
export type NodeFlags = {
    flags?: string[];
};
export type TemplatedNodeData = {
    isTemplatedNode?: boolean;
    templatedNodeInstructions?: string;
    areTemplatedNodeInstructionsRequired?: boolean;
};
export type NodeData = Record<string, Value> & NodeTags & NodeFlags & TemplatedNodeData;
export interface Node {
    id?: NodeId;
    type?: ComponentType;
    edges?: Edges;
    data?: NodeData;
}
type RootNode = {
    edges: Edges;
};
```
- So, from this, we can see that "MoreInformation" props like `howMeasured` etc just fall under the `Record<string,Value>` generic case of `NodeData`. And `NodeData` is _not_ scoped to component `type` in any sense yet.
- In the future, we know we want to have a stronger typed graph & nodes - so I think long term goals here would be to have a `NodeHelpText` type to capture these props consistently, then actually only map them onto relevant component types (eg `NodeData` becomes discriminated union by `type` rather than catchall like now), and ensure editor modal's `MoreInformation` components align correctly.

BUT until we get to that future, I _think_ a simple check like existing `hasHelpText` (see editor/src/pages/FlowEditor/components/Flow/components/Question) would be more future proof/easier to maintain than hard-coded list here?

Does this make sense? Insufficient graph types are something we keep bumping up against, this is another prime example/motivation !

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/ContextMenu.tsx
```diff
@@ -171,6 +218,26 @@ export const ContextMenu: React.FC = () => {
           onClick: handleCut,
         },
       ];
+
+      if (supportsHelpText) {
+        actions.push({
+          id: "copy-help",
+          label: "Copy help text",
```
jessicamcinchak, line 225: nit: I'd love to consistently align "More information" and "Help text" language - eg would there be any consequences to simply renaming editor modal section "Help text" instead of "More information" ?

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -278,6 +278,14 @@ export interface EditorStore extends Store.Store {
...
   copyNode: (id: NodeId) => void;
   getCopiedNode: () => { node: Store.Node; children: Store.Node[] };
+  copyHelpText: (id: NodeId) => void;
+  getCopiedHelpText: () => {
+    info?: string;
+    policyRef?: string;
+    howMeasured?: string;
+    definitionImg?: string | null;
+  } | null;
+  pasteHelpText: (toId: NodeId) => void;
```
jessicamcinchak, line 285: 👍

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -505,6 +513,60 @@ export const editorStore: StateCreator<
 
     return JSON.parse(payload);
   },
+  copyHelpText(id: string) {
+    const { flow } = get();
+    const node = flow[id];
+    if (!node) return;
+
+    const payload = {
+      rootId: id,
```
jessicamcinchak: nit: I'm not entirely following what `rootId` means here and why it's necessary (eg it doesn't seem to be  referenced during local storage retrieval or mapped onto new node during `paste`) ?? 

Can we remove it? If it is necessary, could we call it `sourceNodeId` instead maybe? "Root" has a very specific meaning in the graph!

jessicamcinchak (commented):
This is looking great! A few comments and type suggestions below - will come back to soon for final review once conflicts also resolved :slightly_smiling_face:

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/ContextMenu.tsx
augustlindemer: Yeah for sure. In content language we have always called it 'help text' I think, whereas in code it's always been MoreInformation. Front end used to just be an icon, then replaced with 'More information' text. If we iron these out across everything that's a separate PR though I reckon?

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
augustlindemer: I just copied the actual operation logic from the existing copy operation (copyNode) without double checking applicability. As you say, pasteHelpText doesn't actually use it so it's presumably superfluous!

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/ContextMenu.tsx
augustlindemer: I don't see how a `hasHelpText` check would work here. It would exclude every node that doesn't already have populated help text content, which are the main candidate nodes to paste help text onto. The only nodes this should exclude are nodes whose model does not support help text. This is mostly (but not exclusively) nodes without a user facing render.

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/ContextMenu.tsx
jessicamcinchak: Yes follow-up PR very reasonable here ! Also think possible to make consistent first in editor side without touching "More information" user-facing link yet.

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/ContextMenu.tsx
jessicamcinchak: Ah sorry - you're right ! `hasHelpText` only applies to _if_ there is any populated text available to copy, not the paste operation. 

The paste operation relies on _if_ a given editor modal has `<MoreInformation />`, _not_ whether the underlying props exist or not (in theory they always could _exist_ but not be rendered because `NodeData` is so permissive currently) - so let's go with your list of types for now !

jessicamcinchak (approved):
Changes all look good here, thank you! 

(Let's discuss aligning prettier configs/linters at next dev call - there's more of us now & you're not the first to pickup unrelated changes on rebase!)

PR #6436, opened 2026-04-09
refactor(editor): Implement reducer pattern for complex `FileUploadAndLabel` interactions

DafyddLlyr:
## What does this PR do?
Refactors `FileUploadAndLabel` to use a reducer pattern ([docs](https://react.dev/reference/react/useReducer)). This will allow us to more easily scale and maintain this component. This follows on from [#6416](https://github.com/theopensystemslab/planx-new/pull/6416) and has been a bit of tech debt I've been very keen to address - as the complexity here has grown we've simply continued to manage state in-line which is proving tricky to manage.

This PR does not introduce new functionality, and is simply a refactor of the existing `FileUploadAndLabel` component.

## Why this pattern?
Previously, we relied here on multiple `useState()` hooks paired with reactive `useEffects()` - this is tricky to read and understand, and will be very challenging to integrate the following upcoming features (imho!)
 - Drawing numbers
 - `slots` as a single source of truth (see #6416)
 - Persist data mid-component (e.g. on upload, on tag), not just on "Continue"
 - AI classification and validation (?)

Switching to a pattern using `useReducer()` allows to to understand the component a series of "actions" which the user can trigger. This makes the UI dumber, and the state management is handled centrally by the reducer - this means we don't need to manage cascading updates and effects.

## Testing
In order to test this you'll need the `UPLOAD_LABEL_REBUILD` feature flag enabled. 

## Next steps...
Please see #6416 for more detail - the next step here is to make `slots` a single source of truth.

PR #6501, opened 2026-04-19
feat: Update `FileUploadAndLabel` to use `slots` a single source of truth

DafyddLlyr:
## Context
This follows on from https://github.com/theopensystemslab/planx-new/pull/6416 and https://github.com/theopensystemslab/planx-new/pull/6436 - please see for further context.

Historically, `FileUploadAndLabel` has been driven by two data models - `Slots` and `FileList` - keeping these two in-sync and up to date has caused a number of bugs in the past and leads to pretty complex code within this component. There's a bit more detail on this here - https://github.com/theopensystemslab/planx-new/pull/6216#pullrequestreview-3882350636

## What does this PR do?
This PR removes the dual-state pattern in favor of a strict single source of truth.

**The `slots` array is the single source of truth, and the only mutable state**
When a user selects a tag, it is pushed directly to a new `tags: []` property on the file slot itself.

**`fileList` is now read-only** 
This is only the rules engine defining what the user needs to upload. We not longer update and try to sync this.

**The UI is derived from `slots`**
The requirement checklists, chips, and validation schemas all observe the `slots` array and calculate what to display / validate.

### Key changes
* State simplification - as a result of the above, many helper functions (and tests) have now been safely removed as we no longer need to sync state
* Type cleanup - Similarly, many legacy types (e.g., `UserFile`) are removed or simplified
* Validation - Rather than 3 separate Yup schemas, we have just one which runs the same set of checks against the `slots` array
* Backwards compatible - payload generation (`generatePayload()`) and recovery (`getRecoveredData()`) still account for the previous data structures and will still work if a user is mid-journey (see tests)

## Scope
This PR is unfortunately larger and more involved than originally planned. Last week I tried a few times to get a partially working version up but the truth is that when the change to `<Dropzone<FileUploadAndLabelSlot>>` is made there's a stream of cascading type changes and tests that need to be updated together.

[My original plan](https://github.com/theopensystemslab/planx-new/pull/6416) was to update the types, write to both locations, and then build from there but it soon became clear that this was a little naive. Managing both of these at once would have made this much more complex, just in order to delete one method. Instead of more technical debt and complexity to achieve a slow transition, I thought it best to commit to the full change. Apologies if this makes it more challenging to review - hopefully the comments and commits help somewhat, also very happy to talk this through at dev call.

## Next steps...
One final (small!) PR to finally wire-up the `drawingNumber` field and drop the feature flag..!


apps/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/index.tsx
```diff
@@ -94,80 +89,58 @@ export default function Component(props: Props) {
...
+        abortEarly: false,
         context: { fileList: state.fileList },
-      }),
-      fileLabelSchema.validate(state.fileList, {
-        context: { slots: state.slots },
-      }),
-      fileListSchema.validate(state.fileList, {
-        context: { slots: state.slots },
-      }),
-    ])
```
DafyddLlyr, line 140: Before: 3 different validation schemas which all needed to be checked and managed.

After: Just one, which checks the actual user data.

apps/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/index.tsx
```diff
@@ -196,13 +172,18 @@ export default function Component(props: Props) {
...
-                <Dropzone
+                <Dropzone<FileUploadAndLabelSlot>
                   slots={state.slots}
                   setSlots={handleSetSlots}
                   setFileUploadStatus={handleSetFileUploadStatus}
+                  createSlot={(baseSlot) => ({
+                    ...baseSlot,
+                    tags: [],
+                    drawingNumber: "",
+                  })}
```
DafyddLlyr, line 183: Here's the follow up implementation to https://github.com/theopensystemslab/planx-new/pull/6416

apps/editor.planx.uk/src/@planx/components/FileUploadAndLabel/model.test.ts
```diff
@@ -470,40 +535,20 @@ describe("generatePayload function", () => {
     expect(result.data?.docTypeB[0].filename).toEqual("single-file.pdf");
     expect(result.data?.docTypeB[0].cachedSlot.id).toEqual("slot-001");
   });
-
-  it("maps the most recent validated slot into the payload", () => {
```
DafyddLlyr, line 474: This test was introduced here - https://github.com/theopensystemslab/planx-new/pull/5192

This refactor means that this "synchronisation" category of bug can't happen - there's just a single data source.

apps/editor.planx.uk/src/@planx/components/FileUploadAndLabel/model.test.ts
```diff
@@ -574,160 +632,17 @@ describe("getRecoveredData function", () => {
...
-    expect(fileType1?.slots?.[0].id).toEqual("slot-001");
-
-    // file2 should only have slot2
-    expect(fileType2).toHaveProperty("slots");
-    expect(fileType2?.slots).toHaveLength(1);
-    expect(fileType2?.slots?.[0].id).toEqual("slot-002");
-  });
-});
-
-describe("getTagsForSlot function", () => {
```
DafyddLlyr, line 603: Many tests below are now removed - the associated helper functions no longer exist.

apps/editor.planx.uk/src/@planx/components/FileUploadAndLabel/schema.ts
```diff
@@ -97,64 +95,65 @@ export const slotsSchema = array()
...
     name: "errorStatus",
     message: "Remove files which failed to upload",
-    test: (slots?: Array<FileUploadSlot>) => {
+    test: (slots) => {
       const didAnyUploadFail = slots?.some((slot) => slot.status === "error");
       return !didAnyUploadFail;
     },
-  });
+  })
+  .test({
```
DafyddLlyr, line 103: The validation checks from the previous three schemas are now merged here.

apps/editor.planx.uk/src/@planx/components/shared/PrivateFileUpload/UploadedFileCard.tsx
```diff
@@ -166,89 +149,40 @@ export const UploadedFileCard: React.FC<Props> = ({
               )}
             </Box>
           </Box>
-          {tags !== undefined ? (
-            // Legacy modal UI: icon button delete
```
DafyddLlyr, line 170: This was missed from https://github.com/theopensystemslab/planx-new/pull/6442

jessicamcinchak (approved):

PR #6507, opened 2026-04-20
fix(editor): Add loading state for offline/online chip

DafyddLlyr:
## What does this PR do?
Small UI fix for something which has bugged me for a while...! This chip flashes in incompletely currently before the status is fetched. Video below (scrubbing forwards and backwards over the loading) to demonstrate this - 

https://github.com/user-attachments/assets/43c9e958-1e46-4f65-a32b-9e0682325927

This PR just ensures the chip only renders when we have all the required data.

_TODO: add video when pizza is up_


jessicamcinchak (approved):
Nice !

PR #6508, opened 2026-04-20
style: make FlowTable column widths uniform across views

zz-hh-aa:
Fixes a little styling issue with the Actions column of FlowTable across flows and archive views.

## Before:
main-default
main-default-archive

## After:
feat-default
feat-default-archive

My fix (replacing `TableCell` with `FlowStatusCell` and making the title cell width 100%) does affect some of the other widths. eg, see 'Last edited' column. I checked the changes on a couple other screen sizes too and think using `FlowStatusCell` sometimes works better after all, eg on smaller screens:

**Main**
main-1280

**Feature branch:**
feat-1280

Let me know if there's a better solution here!


DafyddLlyr (approved):
Nice...! 💅

PR #6494, opened 2026-04-17
chore: drop modal from FindProperty component

zz-hh-aa:
Removes outdated modal content linking to external planning site, as [flagged by August](https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1775730393404259). 

- Always allow new address / remove boolean value `allowNewAddresses` prop and conditional rendering
- Remove externalPlanningSite references & db columns
- Bumps `planx-core` `TeamSettings` type (removing externalPlanningSite props)

❓ Am I right in that we expect CI to fail here until my `planx-core` [PR](https://github.com/theopensystemslab/planx-core/pull/974) is merged?

PR #6509, opened 2026-04-20
chore: remove external planning site from seed database

zz-hh-aa:
Missed this in #6494


DafyddLlyr (approved):
Good spot - also didn't consider this..!

jessicamcinchak (approved):
TY!

PR #6514, opened 2026-04-20
feat(editor): Migrate `/global-settings` page to new layout

DafyddLlyr:
## What does this PR do?
 - Migrates the `/global-settings` page to match the layout used for team setting and flow settings
 - Removes `<ListManager/>` setup - this now allows us to save one item at a time instead of saving all at once
 - Updates data fetching methods - remove Zustand store functions, use Apollo queries and mutations
 
 ## Motivation
 Two-fold really..! This is the final settings page, we should have picked this up earlier alongside the others, and updating it to save one footer item at a time will help resolve issues such as the one reported here - https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1776677757992029 (OSL Slack)
 
 ## Demo

https://github.com/user-attachments/assets/40496cd6-4cc7-40cc-a26a-c6cae2c0fad1

## Old layout
image


jessicamcinchak (approved):
Thanks for picking up this migration + bug fix !

PR #6515, opened 2026-04-20
fix: add `flows.deleted_at` to nightly data sync

jessicamcinchak:
Currently throwing `psql:write/flows.sql:26: ERROR:  extra data after last expected colum` otherwise ! 

These are easy to miss! Flows is a table which does a `select * from ...` via `container.sh`, so columns need to be kept manually in sync with Hasura migrations (_but_ only once that migration merges to prod!).


DafyddLlyr (approved):
Thanks for spotting this and fixing..! 🙌 

Merged to `main` so that the synch succeeds this morning.

PR #6516, opened 2026-04-21
feat: display a basic "degraded service" banner on demand

jessicamcinchak:
Follows on from recent custom subdomain downtime bugs (and broken image bugs before that! etc etc) - see https://trello.com/c/myS9aUr6/3549-ability-to-toggle-on-a-custom-banner-across-all-public-links-in-cases-of-degraded-service-or-known-bugs

**Testing notes:** 
- Login to pizza editor. Banner is "off"/invisible by default across all links (editor, public)
- Login to hasura console. Go to "Data" > table "downtime_banner" > edit existing row and toggle "is_visible" to `true` 
- Refresh editor and public links. Banner is visible across all of them  

**UI preview (currently "all" links by default):**
Screenshot from 2026-04-20 21-07-22
Screenshot from 2026-04-20 21-06-58


apps/editor.planx.uk/src/components/DowntimeBanner/DowntimeBanner.tsx
```diff
@@ -0,0 +1,40 @@
...
+        display: "flex",
+        alignItems: "center",
+        justifyContent: "space-between",
+      }}
+    >
+      <Box display="flex" alignItems="center" justifyContent="space-between">
+        <BrokenImageIcon />
+        <Typography variant="body2" ml={1}>
+          <strong>Partial degradation</strong> of some services and features. We
+          are actively working on a fix. Please try again soon.
```
jessicamcinchak, line 39: Copy edit suggestions welcome! Ideally text stays on one-line here

apps/editor.planx.uk/src/hooks/data/useDowntimeBanner.ts
```diff
@@ -0,0 +1,21 @@
+import { gql, useQuery } from "@apollo/client";
+
+const FETCH_DOWNTIME_BANNER_VISIBILITY = gql`
+  query FetchDowntimeBannerVisibility {
+    downtimeBanner: downtime_banner_by_pk(name: "DowntimeBanner") {
+      isVisible: is_visible
+    }
+  }
+`;
```
jessicamcinchak, line 9: We could consider `subscribe` rather than `query` here if it's important for _existing_ sessions to immediately see the banner rather than only after refresh (new visitors will always see as-is)? But, for something that should be invisible 99% time, `query` feels right for now!

apps/hasura.planx.uk/migrations/default/1776805921348_create_table_public_downtime_banner/up.sql
```diff
@@ -0,0 +1,11 @@
...
+  "name" text NOT NULL DEFAULT 'DowntimeBanner', 
+  "is_visible" boolean NOT NULL DEFAULT false, 
+  "updated_at" timestamp NOT NULL DEFAULT now(), 
+  PRIMARY KEY ("name"), 
+  UNIQUE ("name"), 
+  CONSTRAINT "restrict_to_single_row" CHECK (name = 'DowntimeBanner'));
+  
+COMMENT ON TABLE "public"."downtime_banner" IS E'Control whether the <DowntimeBanner /> component is visible, effective immediately without a code deployment';
+
+INSERT INTO downtime_banner (name) VALUES ('DowntimeBanner');
```
jessicamcinchak, line 11: I've chosen to limit this table to a single row for now - I think this helps limit any confusion in a situation where we actually just need/want to turn this on quickly! 

In a more robust future, I could imagine this evolving to support: 
- Different component types for different views/warnings (eg `<EditorDowntimeBanner />`, `<PublicDowntimeBanner />` etc) that can each be turned "on" independently
- Additional column to add custom `message` instead of setting default text directly in the component file
- Other requirements, ideas?

apps/editor.planx.uk/src/hooks/data/useDowntimeBanner.ts
DafyddLlyr: Agreed..!

We can also use the `pollInterval` property to check every few minutes ([docs](https://www.apollographql.com/docs/react/data/queries#polling)).

apps/hasura.planx.uk/migrations/default/1776805921348_create_table_public_downtime_banner/up.sql
DafyddLlyr: This strikes me as the right approach for now - easily expandable in future if we need to.

apps/editor.planx.uk/src/components/DowntimeBanner/DowntimeBanner.tsx
```diff
@@ -0,0 +1,40 @@
...
+  <DowntimeBannerWrapper>
+    <Container
+      maxWidth="contentWrap"
+      sx={{
+        display: "flex",
+        alignItems: "center",
+        justifyContent: "space-between",
+      }}
+    >
+      <Box display="flex" alignItems="center" justifyContent="space-between">
```
DafyddLlyr: a11y: We probably need some markup for this - `role="alert"` and `aria-live="assertive"` (cc @ianjon3s)

apps/editor.planx.uk/src/components/DowntimeBanner/DowntimeBanner.tsx
```diff
@@ -0,0 +1,40 @@
...
+import Box from "@mui/material/Box";
+import Container from "@mui/material/Container";
+import { styled } from "@mui/material/styles";
+import Typography from "@mui/material/Typography";
+import React from "react";
+
+const DowntimeBannerWrapper = styled(Box)(({ theme }) => ({
+  display: "flex",
+  backgroundColor: theme.palette.error.main,
+  color: "#070707",
```
DafyddLlyr: a11y: This doesn't meet contrast requirements

Image

apps/editor.planx.uk/src/components/Header/Header.tsx
```diff
@@ -316,20 +318,27 @@ const Header: React.FC = () => {
   const headerRef = useRef<HTMLDivElement>(null);
   const teamTheme = useStore((state) => state.teamTheme);
 
+  // Query `downtime_banner` visibility direct via db so not deployment sensitive
+  const { data } = useDowntimeBanner();
+  const showDowntimeBanner = data?.downtimeBanner?.isVisible || false;
```
DafyddLlyr, line 323: nit: I'd be inclined to move this across to the `DowntimeBanner` and have it return `null` if the condition is not met as there's already a lot going on in this file.

apps/editor.planx.uk/src/components/DowntimeBanner/DowntimeBanner.stories.tsx
DafyddLlyr, line 1: Great to include this!

DafyddLlyr (changes requested):
A few small a11y questions to look at.

I've toggled the warning to `true` on the Pizza but the banner isn't showing - I can't see the `FetchDowntimeBannerVisibility` query getting called in the network tab. Probably a build issue on the Pizza, just wanted to flag!

apps/editor.planx.uk/src/components/DowntimeBanner/DowntimeBanner.tsx
jessicamcinchak: good shout! text color now updated to use `theme.palette.error.contrastText` :heavy_check_mark:

apps/editor.planx.uk/src/components/DowntimeBanner/DowntimeBanner.tsx
jessicamcinchak: added !

apps/editor.planx.uk/src/components/Header/Header.tsx
jessicamcinchak: Agree `Header` is very long, but conditional return is pretty annoying for Storybook - so leaving as-is for now! Let's aim to breakup header into subcomponents soon though - was reminded here it's not particularly easy to wade through authenticated & public displays here.

jessicamcinchak:
Thanks for review here @DafyddLlyr - indeed experiencing strange Vultr build issues on this one. 

Just destroyed server and rebuilt from scratch and still not able to toggle banner on here - despite continuing to work without issue locally. My minor change to banner text color also confusingly not reflected on Storybook here (still black, whereas now white for me locally. again after full vultr destroy + rebuild) :thinking:

apps/editor.planx.uk/src/components/Header/Header.tsx
DafyddLlyr: Ah yes - very good point...! 👍

DafyddLlyr (approved):
All looking good...!

Let's merge to `main` and test there - I think it's just a build issue here somehow. If we find an issue with another Pizza we can investigate further.

PR #6518, opened 2026-04-21
feat: Extract metabase link generation to custom hooks, remove from store

DafyddLlyr:
## What does this PR do?
 - Removes `flowAnalyticsLink` from the store
 - Creates custom hooks for generating the Metabse links (both team and flow level)
 - Implements these in `EditorNavMenu`

## Context
This came up as a result of this bug report - https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1776675793996109 (OSL Slack)

Currently the `flowAnalyticsLink` can become stale - we don't clear it when navigating away from the flow route, and it depends on an async request to update.

By removing it from the store, and always calculating it when needed, we can guarantee it's always correct without having to manually sync the global store on routing events. For further context on the move away from the Zustand store for server-side data (which `flowAnalyticsLink` is derived from) please see https://github.com/theopensystemslab/planx-new/blob/main/doc/architecture/decisions/0011-standardised-data-fetching-and-state-management.md


zz-hh-aa (approved):
Thanks for catching!! Tested locally (swapping environment to `development`) and works as expected.

PR #6519, opened 2026-04-21
fix: Handle `NavMenuItems` with both a compact and disabled state

DafyddLlyr:
Small bug I noticed when working on https://github.com/theopensystemslab/planx-new/pull/6518

Currently, we don't correctly update the tooltip if a menu item is both compact and disabled (e.g. flow analytics link). Compare the two tooltips from staging below - 

| Team page | Flow page |
|--------|--------|
| image | image |

This PR updates the logic to split out these two conditions

| Team page | Flow page |
|--------|--------|
| image | image |

### DIFF
--- a/apps/editor.planx.uk/src/components/EditorNavMenu/components/NavMenuItem.tsx
+++ b/apps/editor.planx.uk/src/components/EditorNavMenu/components/NavMenuItem.tsx
@@ -19,44 +19,21 @@ const NavMenuItem = ({
   compact,
   onClick,
 }: Props) => {
-  if (compact) {
-    return (
-      <Tooltip title={title} placement="right">
-        <Box component="span">
-          <MenuButton
-            title={title}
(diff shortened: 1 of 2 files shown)


apps/editor.planx.uk/src/components/EditorNavMenu/components/NavMenuItem.tsx
```diff
@@ -19,44 +19,21 @@ const NavMenuItem = ({
...
-            <Icon />
-          </MenuButton>
-        </Box>
-      </Tooltip>
-    );
-  }
-
-  if (disabled) {
-    return (
-      <Tooltip title={`${title} unavailable`} placement="right">
```
DafyddLlyr, line 43: `compact` condition skipped and returned early - this is unreachable.

ianjon3s (approved):
Works great, thanks for spotting this one

PR #6520, opened 2026-04-21
feat: unpin archived flows

zz-hh-aa:
When the `useArchiveFlow` hook runs, ensures that all records for that flow are deleted from the `user_pinned_flows `table. As discussed at S+T, this is the behaviour we would expect when archiving a pinned flow. If an archived flow is unarchived, it is no longer pinned for any users that might have pinned it previously. 

I tested this locally for my own user, as well as for an additional record for the same flow pinned by another user. 

https://github.com/user-attachments/assets/ebb8a4e9-4def-4627-977c-1cc6453df18e

### DIFF
--- a/apps/hasura.planx.uk/migrations/default/1776784884510_create_delete_pinned_flows_on_archive_trigger/down.sql
+++ b/apps/hasura.planx.uk/migrations/default/1776784884510_create_delete_pinned_flows_on_archive_trigger/down.sql
@@ -1,0 +1,2 @@
+DROP FUNCTION IF EXISTS delete_pinned_flows_on_archive();
+DROP TRIGGER IF EXISTS delete_pinned_flows_on_archive_trigger ON flows;
--- a/apps/hasura.planx.uk/migrations/default/1776784884510_create_delete_pinned_flows_on_archive_trigger/up.sql
+++ b/apps/hasura.planx.uk/migrations/default/1776784884510_create_delete_pinned_flows_on_archive_trigger/up.sql
@@ -1,0 +1,16 @@
+CREATE OR REPLACE FUNCTION delete_pinned_flows_on_archive()
+RETURNS TRIGGER AS $$
+BEGIN
+  IF OLD.archived_at IS NULL AND NEW.archived_at IS NOT NULL THEN
+    DELETE FROM user_pinned_flows
+    WHERE flow_id = NEW.id;
+  END IF;
+  
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER delete_pinned_flows_on_archive_trigger
+AFTER UPDATE ON flows
+FOR EACH ROW
+EXECUTE FUNCTION delete_pinned_flows_on_archive();


apps/editor.planx.uk/src/pages/Team/components/FlowCard/queries.ts
```diff
@@ -16,10 +16,24 @@ export const PIN_FLOW = gql`
   }
 `;
 
-// userId not required as a variable, as it is pulled from the jwt by the hasura middleware, and used to restrict operations
```
zz-hh-aa, line 19: `userId` now required as a variable for the unpin flow operations because I changed Hasura settings to enable users to delete _any_ record from the table, not just their own

apps/hasura.planx.uk/metadata/databases/default/tables/public_user_pinned_flows.yaml
```diff
@@ -68,13 +68,9 @@ select_permissions:
 delete_permissions:
   - role: platformAdmin
     permission:
-      filter:
-        user_id:
-          _eq: X-Hasura-User-Id
```
DafyddLlyr, line 73: Allowing this would mean any user could take their JWT, hit the GraphQL API and delete all other user's records here - breaking the permission model.

The right approach would be to be to setup an `ON DELETE CASCADE` for these records - if the parent flow record is deleted, these are also removed as a side effect by the DB.

The catch (there's always something 😅) is that we have a soft delete pattern in place - so it's a `UPDATE` not a `DELETE` we need to look out for. In this instance, we can write a SQL trigger to cover this event. This should allow us to retain our permissions model, and carry out the tidy-up "side effect" for these records.

apps/editor.planx.uk/src/pages/Team/components/ActiveFlowMenu.tsx
```diff
@@ -32,6 +34,7 @@ const ActiveFlowMenu: React.FC<FlowMenuProps> = ({
   const handleArchive = async () => {
     try {
       await archiveFlow();
+      await unpinFlowForAllUsers();
```
DafyddLlyr: Ultimately, I don't think we'll need to apply this directly here (see previous comment), but just a heads up on this pattern.

When we have two async operations like this back-to-back, we should consider a more defensive approach - what would happen if `archiveFlow()` fails somehow but then `unpinFlowForAllUsers()` is hit?

The more robust approach here would be to wrap both mutations in a single GraphQL transaction so that they pass and fail together. 

Docs - https://hasura.io/docs/2.0/mutations/postgres/multiple-mutations/

DafyddLlyr (changes requested):

apps/hasura.planx.uk/metadata/databases/default/tables/public_user_pinned_flows.yaml
zz-hh-aa: Good shout--trigger created in [21e9495](https://github.com/theopensystemslab/planx-new/pull/6520/commits/21e94954a0699ce2e1d80d0ab7a8262ae89500c8). Functionality remains the same as in the screen recording above :)

DafyddLlyr (approved):
Great...!

This would fall neatly into that "invisible functionality" we discussed last week that API E2E tests are good for documenting. If you have capacity this sprint, I'd suggest having a look at setting up somer Gherkin tests which - 

 - pin some flows
 - archive one of them
 - check if the record still exists in `user_pinned_flows`

PR #6529, opened 2026-04-22
feat: handle duplicate flow slugs on unarchive

zz-hh-aa:
# What does this PR do? 
- Updates `RenameDialog` to have two modes: a straightforward `rename` and a `renameAndUnarchive`
    - It now also takes `teamId` as a prop for refetch queries (I noticed that previously the list of flows didn't refresh after one was renamed, which felt buggy)
- Refactors `RenameDialog` original mutation into a hook
- Checks for uniqueness violation when unarchive a flow, and if so, renders `RenameDialog` with name and slug automatically updated to include "unarchived"

# Testing
- Archive a flow
- Create a new flow with the same name
- Unarchive the original flow
- Use `RenameDialog` to rename and successfully unarchive

Regular rename:
Screenshot 2026-04-22 163659

Rename and unarchive:
Screenshot 2026-04-22 163855


apps/editor.planx.uk/src/pages/Team/components/RenameDialog.tsx
```diff
@@ -165,7 +179,9 @@ export const RenameDialog: React.FC<Props> = ({
                   disableRipple
                   disabled={isSubmitting}
                 >
-                  Rename flow
+                  {mode === "rename"
+                    ? "Rename flow"
+                    : "Rename and unarchive flow"}
```
zz-hh-aa, line 185: Is this too clunky for a submit button? (See screenshot in PR description)

apps/editor.planx.uk/src/pages/Team/components/hooks/useRenameAndArchiveFlow.ts
```diff
@@ -0,0 +1,20 @@
...
+import {
+  FlowStatusMutation,
+  GET_ARCHIVED_FLOWS,
+  GET_FLOWS,
+  RENAME_AND_UNARCHIVE_FLOW,
+  RenameFlowMutationVars,
+} from "../../queries";
+
+export const useRenameAndUnarchiveFlow = (flowId: string, teamId: number) =>
+  useMutation<FlowStatusMutation, RenameFlowMutationVars>(
```
zz-hh-aa: Unsure how clear my type names are, suggestions welcome

apps/editor.planx.uk/src/pages/Team/components/RenameDialog.tsx
```diff
@@ -127,6 +136,11 @@ export const RenameDialog: React.FC<Props> = ({
               <DialogContent
                 sx={{ gap: 2, display: "flex", flexDirection: "column" }}
               >
+                {mode === "renameAndUnarchive" && (
+                  <Typography>
+                    Duplicate found, please give the flow a unique name.
```
joecarver: i think maybe a slightly more verbose message might be helpful, as this is a potentially confusing case, perhaps something like:

"A new flow with the same name has been created whilst this one was archived. In order to unarchive it, please first rename it"

apps/editor.planx.uk/src/pages/Team/components/RenameDialog.tsx
joecarver: personally i think its fine and better to be more explicit!

apps/editor.planx.uk/src/pages/Team/queries.ts
```diff
@@ -113,3 +113,35 @@ export const DELETE_FLOW = gql`
...
+
+export type RenameFlowMutationVars = {
+  flowId: string;
+  newSlug: string;
+  newName: string;
+};
+
+export const RENAME_FLOW = gql`
+  mutation RenameFlow($flowId: uuid!, $newSlug: String, $newName: String) {
+    update_flows_by_pk(
```
joecarver: correct me if i'm wrong but i think we should alias the mutation to `flow` like we do in the other mutations:

```suggestion
    flow: update_flows_by_pk(
```

to match the return type `FlowStatusMutation` which has a `flow` property. I guess its not causing an issue right now as we don't do anything with the return value but probably good to keep it aligned regardless

apps/editor.planx.uk/src/pages/Team/components/hooks/useRenameAndArchiveFlow.ts
joecarver: `RenameFlowMutationVars` seems spot on but, `FlowStatusMutation` perhaps is a misnomer, more so when used in the `useRenameFlow` hook as that doesn't actually concern the flow status.

This type is for the response, right? so `FlowMutationResponse` might be most accurate? As it's re-used for a few different responses. Not sure if we have a different naming convention for mutation types though...

joecarver (changes requested):
Works great! The UX with the modals is very smooth I think, best of both worlds. 

Pretty much the only 'confusing' state i could think of is if they re-archived an un-archived flow after using the auto-rename, then they would have an archived flow with 'unarchived' in the name. But this seems like such an edge case, and it would also be potentially more messy to try and address it. 

I left a couple of small comments on the copy / typing but otherwise all good!

apps/editor.planx.uk/src/pages/Team/components/hooks/useRenameAndArchiveFlow.ts
zz-hh-aa: Makes sense to me, don't think there is a different naming convention for mutation types! Updated in [12af47a](https://github.com/theopensystemslab/planx-new/pull/6529/commits/12af47a488e98360022aa98f608e0f12e49a7666)

apps/editor.planx.uk/src/pages/Team/components/RenameDialog.tsx
zz-hh-aa: Went for: "An active flow called "{flow.name}" already exists. To unarchive this one, please give it a unique name first." in [1351ea5](https://github.com/theopensystemslab/planx-new/pull/6529/commits/1351ea531ff318811f0e8c8dc6874529e4363b8c)

apps/editor.planx.uk/src/pages/Team/queries.ts
zz-hh-aa: Good eye 🔍 updated in [82897b6](https://github.com/theopensystemslab/planx-new/pull/6529/commits/82897b6004e51742e4f4fc665adb4d0058eabd38)

apps/editor.planx.uk/src/pages/Team/components/RenameDialog.tsx
joecarver: perfect !

joecarver (approved):

PR #6531, opened 2026-04-23
chore: create analytics_first_used view

zz-hh-aa:
Creates a new view just for `flow_id` and `first_used_at` date. It was slow / expensive to run this as part of an existing view, hence the new table. This can be used to automate Metabot alerts so we can keep on top of new services coming online.

Decision explained in more depth [here](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1776947637441539?thread_ts=1776354127.151379&cid=C01E3AC0C03).


jessicamcinchak (approved):
Smart & simple!

PR #6533, opened 2026-04-23
[infra] enable creation or migration of custom domains to shared CloudFront distribution

freemvmt:
What we're calling the 'SSL migration' is shaping up here (or as it's known in my head, 'one CloudFront to rule them all'). See [relevant ticket](https://trello.com/c/8k8xg8AI) and detailed [scoping doc](https://docs.google.com/document/d/1POAJjpZvhD1rEkV5GVY30YkRlvLC7NjxBLpgX476hak/edit?usp=sharing).

## TODO

- [ ] Document both pathways (migration vs. onboarding from scratch) i.e. replace [this howto](https://github.com/theopensystemslab/planx-new/blob/main/doc/how-to/how-to-setup-custom-subdomains.md) (and [child](https://github.com/theopensystemslab/planx-new/blob/main/doc/how-to/how-to-renew-ssl-certificates.md)) and [associated public notion doc](https://www.notion.so/opensystemslab/9-Set-up-custom-subdomains-3000ef5212cc43a5a88f46563142f82a))

### Enum system

Instead of the 2 bools (`isLegacy` / `isReady`) defining the state of a given domain, we instead define four explicit states that any domain might be in (as pertains to ACM/certificates and CloudFront). The idea is for this to be clearer and more self-documenting than the previous proposal.

See `infrastructure/common/types.ts`.

### Two switches system - rejected!

The infrastructure setup for any custom subdomain should always be in one of the following 5 states:

1. `isLegacy === true`, `isReady === undefined`

Domain should be served by it's own cert and dedicated CloudFront distribution, as is the current condition for all custom subdomains currently running on production (and set up as described in [this how to](https://github.com/theopensystemslab/planx-new/blob/main/doc/how-to/how-to-setup-custom-subdomains.md)).

Pre-existing domains are fine to stay in this state until their imported cert is close to expiry, at which point we should proceed through states (2), (3), (4).

2. `isLegacy === true`, `isReady === false`

Legacy setup should be maintained as is, but meanwhile we are adding the domain to our 'mining' cert to surface a DNS validation record (as a Pulumi output from the `certificates` layer) that needs to be communicated to the council in question and added to their nameserver.

3. `isLegacy === true`, `isReady === true`

Again, legacy setup maintained, but now the custom domain is also added to the single, shared CloudFront distribution (as an additional alias) and associated cert (as an additional SAN). This certificate can only be issued if DNS validation can be completed for *all* domains listed. This is why we have to ensure that councils add said record _before_ we put `isReady: true`.

After this change is on prod, the council can replace the content of their CNAME record for the custom domain (which was pointed at the dedicated/legacy distribution) with the address of the shared distribution, as provided by us.

4. `isLegacy === false || undefined`, `isReady === true`

Once that DNS record is replaced, we can kill the old legacy infrastructure for the given domain. This is the final, most desirable state for all domains, which we would expect all councils to have moved to before April 13th 2027 (the last date any current imported cert will expire) - excluding Camden who are a special case in that they auto-renew their cert, so it's not listed for expiry on our end until 2039.

5. `isLegacy === false || undefined`, `isReady === false`

If we are not migrating a custom domain, but instead on-boarding a new council, we will skip the legacy setup and go straight for the shared CloudFront option. We kick this off by simply adding the `isReady: false` flag, to surface the DNS validation record, as also described in (2).

NB. Keen observers will notice there is a final combination, `isLegacy == false || undefined`, `isReady === undefined` - in this case the implication is that we serve no infrastructure. In this case either Pulumi should error out, i.e. this should not be an allowable config, or we should simply not build anything 🤔

**Migration vs. creation**

If we are migrating an existing legacy domain, it will already be in state (1), and we'll move it through (2), then (3), and end on (4). During this we require the council to update their DNS config twice, once between (2) -> (3) (adding validation record), and once between (3) -> (4) (replacing CNAME).

If we are adding a new custom domain, we will set it up in state (5), and then move straight to (4). In this case we only require the council to act once, adding both DNS records between (5) -> (4).


infrastructure/common/customDomains.ts
```diff
@@ -0,0 +1,118 @@
...
+          domain: "planningservices.canterbury.gov.uk",
+          name: "canterbury"
+        },
+      ]
+    : [
+        // we maintain a 'custom domain' on staging for testing (on a domain we control but which is not PlanX)
+        { 
+          domain: "planningservices.fairhold.org",
+          name: "fairhold",
+          isLegacy: true,
```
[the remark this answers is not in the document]
freemvmt: Intention here is to move this test domain through all possible states (legacy setup -> mining/prep state -> on shared distribution) as a canary for the process (as described at length in the PR description).

infrastructure/common/utils.ts
```diff
@@ -107,3 +107,12 @@ export const createDestinationSgEgressRule = (
     });
   }
 }
+
+// get domains still served by their own dedicated CloudFront distribution + BYO certificate
+export const getLegacyDomains = (customDomains: CustomDomain[]) => customDomains.filter(cd => cd.isLegacy == true);
+
+// get domains with DNS validation pending — added to 'mining' cert to surface records to send to council
+export const getPendingDomains = (customDomains: CustomDomain[]) => customDomains.filter(cd => cd.isReady == false);
```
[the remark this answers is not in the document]
freemvmt: Fair point that we should be explicit about either flag being `undefined`, since this is a possibility.

infrastructure/common/utils.ts
```diff
@@ -107,3 +107,12 @@ export const createDestinationSgEgressRule = (
...
+
+// get domains validated and ready to be served by the single shared CloudFront distribution
+export const getValidatedDomains = (customDomains: CustomDomain[]) => customDomains.filter(cd => cd.isReady == true);
```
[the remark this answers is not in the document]
freemvmt: This is intended behaviour. Afaik there is no problem with having two different CloudFront distributions pointed at the same origin (indeed, we already have this situation), and running both simultaneously enables a smooth transition from one to the other at the point when the council updates their DNS and it fully propagates.

infrastructure/common/customDomains.ts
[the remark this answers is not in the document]
freemvmt: Or to use the numbering given above, to test the transitions (1) -> (2) -> (3) -> (4).

infrastructure/certificates/index.ts
```diff
@@ -35,4 +43,71 @@ const caaRecordWildcard = new cloudflare.DnsRecord("caa-record-wildcard", {
...
+      subjectAlternativeNames: validatedCustomDomains.slice(1).map(d => d.domain),
+      validationMethod: "DNS",
+    },
+    { provider: usEast1 }
+  );
+
+  const customDomainsCertValidation = new aws.acm.CertificateValidation(
+    "sslCertValidation-custom-domains",
+    {
+      certificateArn: customDomainsCert.arn,
```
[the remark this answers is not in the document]
freemvmt: TODO

DafyddLlyr (commented):
As always - appreciate the clear writeup and documentation. At first scan this all looks good and is clearly laid out.

The only (small!) question I have is if there's a simpler way to model the relationship between `isLegacy` and `isReady`. Rather than manage two booleans, could we think of a string (enum) which describes the current state of a given team, e.g.

```ts
// Better names most welcome!
export type DomainState =
  | "legacy"
  | "legacy-validating"
  | "legacy-ready"
  | "active"
  | "new-validating"; 
```

The tradeoff here would be that filtering in the utils may be slightly clunkier (e.g. checking if `["legacy", "legacy-validating", "legacy-ready"].includes(state)` but management and migration may be simpler (domains progress "up" the chain in one clear, documented, direction).

freemvmt:
> The only (small!) question I have is if there's a simpler way to model the relationship between `isLegacy` and `isReady`.

I see the appeal of this but I'm not completely sold. The two flags currently proposed do actually refer to different infrastructures, i.e. in some sense they are mutually independent. 1 or both respective infrastructures could be running at any given time (although ofc we would expect the latter state we to be a temporary measure allowing for smooth re-routing of traffic). And consequently it's not a straightforward linear progression through states (or at least, there are a minimum of 2 expected paths).

Having said that, I do recognise the above is all a little unintuitive. I'll see if I can rationalise a single enum to myself and implement it. Would be a good excuse to draw a state diagram ;)

infrastructure/application/utils/createCDN.ts
```diff
@@ -68,33 +71,39 @@ const responseHeadersPolicy = new aws.cloudfront.ResponseHeadersPolicy(
 );
 
 export const createCdn = ({
-  domain,
+  cdnName,
+  domains,
```
freemvmt, line 75: here we allow `createCdn` to accept a list of domains instead of a single domain (and we apply the `includeWww` flag to each in the expected way)

infrastructure/application/index.ts
```diff
@@ -34,95 +40,9 @@ const data = new pulumi.StackReference(`planx/data/${env}`);
 // You can generate tokens here: https://dash.cloudflare.com/profile/api-tokens
 new pulumi.Config("cloudflare").requireSecret("apiToken");
 
-const CUSTOM_DOMAINS: CustomDomain[] =
```
freemvmt, line 37: this is essentially config which lives in code, and we've moved it to `infrastructure/common` since it is referenced by both the `certificates` and `application` layer

infrastructure/application/index.ts
```diff
@@ -340,85 +260,112 @@ export = async () => {
...
-          {
-            provider: usEast1,
-            replaceOnChanges: ["privateKey"],
-          }
-        );
-
-        acmCertificateArn = certificate.arn;
-      }
+  // ------------------- (legacy) Custom Domains (per-domain CDN + BYO certificate)
+  const createLegacyDistributions = ({
```
freemvmt, line 270: simplified here by removing the IIFE, which wasn't required, but otherwise this code is essentially unchanged

infrastructure/certificates/index.ts
```diff
@@ -1,13 +1,21 @@
...
+import { getPendingDomains, getValidatedDomains } from "../common/utils";
+
 const config = new pulumi.Config();
+const env = pulumi.getStack();
 
 // The @pulumi/cloudflare package doesn't generate errors so this is here just to create a warning in case the Cloudflare API token is missing.
 new pulumi.Config("cloudflare").requireSecret("apiToken");
 
+// ACM certificates for CloudFront must be provisioned in us-east-1
+const usEast1 = new aws.Provider("us-east-1", { region: "us-east-1" });
```
freemvmt, line 17: should possibly move the code in `providers.ts` to `common` and import from there into both layers

infrastructure/common/customDomains.ts
```diff
@@ -0,0 +1,123 @@
...
+          cloudFrontState: "single-plus-validation",
+        },
+        {
+          name: "canterbury",
+          domain: "planningservices.canterbury.gov.uk",
+          cloudFrontState: "single-plus-validation",
+        },
+      ]
+    : [
+        // we spin up 'custom domains' on staging for testing purposes
```
freemvmt, line 112: intention here is to test both pathways, i.e. both migrating the `test-legacy` 'council' from the legacy, independent CDN to the shared CDN, and onboarding `test-new` as a new 'council' from scratch

(I already followed the [instructions](https://github.com/theopensystemslab/planx-new/blob/main/doc/how-to/how-to-setup-custom-subdomains.md) to create a certificate for `planningserviceslegacy.planx.in` and upload it to AWS secrets)

infrastructure/common/teams.ts
```diff
@@ -1,21 +1,4 @@
-// Secrets to be added to Pulumi when a new team is on-boarded
-interface TeamSecrets {
```
freemvmt, line 2: these moved to colocated `types.ts`

infrastructure/common/types.ts
```diff
@@ -1,4 +1,42 @@
...
+  govPayStagingOnly?: boolean;
+};
+
+/**
+ * Describes which CloudFront infrastructure serves a custom domain.
+ *
+ * On-boarding path for new councils: validation-only → shared-only
+ * Migration path for existing councils: single-plus-validation → single-plus-shared → shared-only
+ */
+export type CloudFrontState =
```
freemvmt, line 26: This is the enum which defines the 'state' of a given domain, which is intended to be more 'self-documenting' and explicit than the previous setup with 2 bools/flags. By 'state' here we mean the AWS infrastructure (ACM certificate + CloudFront CDN) which serves the domain. The name is up for grabs, I went with `CloudFrontState` for now because the CDN distribution is arguably the most crucial piece of the puzzle, and the 'single' and 'shared' terms are referring to the distribution in use. 

 tl;dr: A domain must be in one of these states!

freemvmt (commented):
Ok, I think this is looking good!

infrastructure/common/types.ts
freemvmt: kudos to @DafyddLlyr for arguing for this and being on the money as usual ;)

infrastructure/application/utils/createCDN.ts
```diff
@@ -27,6 +27,9 @@ const staticErrorResponses: aws.cloudfront.DistributionArgs["customErrorResponse
   },
 ];
 
+// TODO: should we just be using a pre-baked AWS policy to simplify this?
+// e.g. CORS-with-preflight-and-SecurityHeadersPolicy
+// see: https://github.com/theopensystemslab/planx-new/pull/6491#issue-4281272671
```
DafyddLlyr, line 32: I think we should just go for this in a follow up PR 👍 

Once done, lets update https://www.notion.so/opensystemslab/New-features-30a35d469ad180e4ba81f4ca4bbf5ca4?source=copy_link#34835d469ad180668516ec28d3d45b3e so that we can track this change for next year's pen test.

infrastructure/application/index.ts
DafyddLlyr: Good call - this was always a little tricky as an IIFE!

DafyddLlyr (approved):
Apologies for the delay in getting back to this one...!

All looks good, and we've just had a clean prod deploy + no known issues in the pipeline. I'd be keen to test this out with a few prod deploys this week so we can get it across the line.

freemvmt:
Ok, will be merging this and testing it out on staging tomorrow AM 🚀

PR #6534, opened 2026-04-23
fix(editor): Update flow modals to correctly handle non-unique flow names

DafyddLlyr:
## What's the problem?
Reported here - https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1776866810893219 (OSL Slack)

Uniqueness violations on flow create / move / copy are not being handled correctly - there's either an unhandled error (modal closes, no indication of error) or a generic toast which does not raise a field level error.

## What's the cause?
We're incorrectly checking for the `isAxiosError()` condition. As our `APIClient` (axios wrapper) throws a custom `APIError<T>`, this condition is never met.

## What's the solution?
 - Add better type definitions to `useCreateFlow()` so that the types are inferred correctly when called
 - Remove the `isAxiosError()` check
 
 ## To test...
 Try to copy a flow on staging, retaining the original name (remove ` (copy)` suffix). Nothing happens, no user feedback, just a network and console error.
  
On Pizza, this will show a field level error - 
  
image

### DIFF
--- a/apps/editor.planx.uk/src/pages/Team/components/AddFlow/hooks/useCreateFlow.ts
+++ b/apps/editor.planx.uk/src/pages/Team/components/AddFlow/hooks/useCreateFlow.ts
@@ -1,15 +1,21 @@
 import { useMutation } from "@tanstack/react-query";
+import type { APIError } from "lib/api/client";
 import {
   createFlow,
   createFlowFromCopy,
   createFlowFromTemplate,
 } from "lib/api/flow/requests";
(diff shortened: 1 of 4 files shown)


jessicamcinchak (approved):
An overdue / boomerang one, thank you!

DafyddLlyr:
> An overdue / boomerang one, thank you!

Yeah I think we've fixed / broken / fixed this a few times working through a series of data fetching, tanstack router, and modal improvement PRs.

PR #6535, opened 2026-04-23
chore: devX improvements for e2e tests

joecarver:
### What's in this PR?
Docker volume isolation for regular dev containers vs e2e test containers. This helps reduce friction when switching between dev and test mode. 

#### How does it work?
The idea here is that we still `down` and `up` the containers when moving between local dev and test mode, but its a lot quicker and crucially we keep the volumes separate, using a separate docker project for the test environment. This means that the data we are working with for local dev stays in place. 

#### Extras
- created a new package script 'dev' which calls `./scripts/start-containers-for-dev.sh` which basically has the inverse operations to the `./scripts/start-containers-for-tests.sh` , so the process of switching between the two is a single command in either direction - `pnpm tests` / `pnpm dev`
- Added a `pretest` and `prestart` script that runs `check-containers` before starting the editor/running the e2e tests. If the correct set of containers are not running it exists immediately with a command.
    - Q: should we use this elsewhere as well?

### DIFF
--- a/apps/editor.planx.uk/package.json
+++ b/apps/editor.planx.uk/package.json
@@ -157,6 +157,7 @@
   },
   "packageManager": "pnpm@10.30.2",
   "scripts": {
+    "prestart": "../../scripts/check-containers.sh dev",
     "start": "vite --open",
     "build": "tsc && vite build",
     "serve": "vite preview",
--- a/README.md
+++ b/README.md
@@ -92,7 +92,8 @@ The root of the project has several scripts set up to help you manage your docke
 - `pnpm destroy` will remove volumes (i.e. database data) and can be a useful hard reset when necessary.
 - `pnpm sync-data` will sync production records with modified data in your database
 - `pnpm clean-data` will sync production records and reset any modified data
-- `pnpm tests` will recreate your docker containers and include test services
+- `pnpm dev` will tear down any e2e containers and recreate your dev docker containers from scratch, including seeding the database - useful when switching from a test environment
+- `pnpm tests` will tear down any dev containers and recreate your docker containers and include test services - useful when switching from a dev environment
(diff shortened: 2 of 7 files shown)


apps/editor.planx.uk/package.json
```diff
@@ -157,6 +157,7 @@
   },
   "packageManager": "pnpm@10.30.2",
   "scripts": {
+    "prestart": "../../scripts/check-containers.sh dev",
```
freemvmt, line 160: I didn't know about this general pre/post trick for package scripts, amazing!

freemvmt (approved):
gorgeous! 🤩 

especially the `check-containers` guard. I tested both scripts out and tried to run the wrong thing, and it did indeed stop me :)

does this change how we should be using scripts like `pnpm run up` etc? like should we just default to running `pnpm dev` to build the stack regardless of whether we've been messing around with e2e tests?

joecarver:
> does this change how we should be using scripts like `pnpm run up` etc? like should we just default to running `pnpm dev` to build the stack regardless of whether we've been messing around with e2e tests?

yeah - i meant to raise this as a discussion point actually, i left all the previous commands in there for compatibility on things i may not know about / people's personal preference, but i certainly think they can be consolidated at least... there's a few which do subtly different things but perhaps we can get rid of some at least! Will ask in Slack.

apps/editor.planx.uk/package.json
DafyddLlyr: Yeah it's a standard npm feature which pnpm has carried over - https://docs.npmjs.com/cli/v11/using-npm/scripts#pre--post-scripts

DafyddLlyr (approved):
Sorry it's taken me so long to get to this one. This is great - very much appreciated!

Working locally as expected 👍

PR #6502, opened 2026-04-20
feat: Add drawing numbers to `FileUploadAndLabel`, update passport and payloads

DafyddLlyr:
## What does this PR do?
This PR finialises the work began in https://github.com/theopensystemslab/planx-new/pull/6216 to introduce the `drawingNumbers` property to the file upload and label component.

The UI already exists, behind a feature flag, and this PR introduces the following changes -
 - Drops the feature flag
 - Populates the passport with a `drawingNumber` if provided
 - Basic test coverage for feature
 - Updates `planx-core` ([see change here](https://github.com/theopensystemslab/planx-core/pull/975)) to populate the `file.number` field in the ODP schema
 - Adds Storybook story - https://storybook.6502.planx.pizza/?path=/story/planx-components-fileuploadandlabel--with-drawing-numbers
 
### Testing
  - Upload files on a service, and submit to create a valid payload
  - Check payload - the `numbers` field should be populated as expected
  - Example submission here - TODO

## Next steps...
Add "drawing number" UI and payload creation to single file upload.

Please see https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1776677646846729 (OSL Slack)

PR #6536, opened 2026-04-24
feat(editor): Support the `drawingNumber` field for single file uploads

DafyddLlyr:
## What does this PR do?
This is the final part (hopefully..!) of the work for https://trello.com/c/KMuYlWbz/3269-allow-users-to-add-a-description-a-number-for-each-drawing-and-document-uploaded

This follows on from...
 - https://github.com/theopensystemslab/planx-new/pull/6416
 - https://github.com/theopensystemslab/planx-new/pull/6436
 - https://github.com/theopensystemslab/planx-new/pull/6442
 - https://github.com/theopensystemslab/planx-new/pull/6501
 - https://github.com/theopensystemslab/planx-new/pull/6502

This PR makes the following changes - 
 - Makes `drawingNumber` a prop on all base `FileUploadSlot`s, not just `FileUploadAndLabel`
 - Allows `drawingNumber` to be controlled on `FileUpload` components

The backend is already covered by https://github.com/theopensystemslab/planx-core/pull/975 - any files with an attached drawing number will be populated via the generated payload.

## Exceptions
The `<PrivateFileUpload/>` component touched here is also used in a few other places which I've not updated.
 
### `DrawBoundary`
I don't think it's currently a requirement. An easy fix-forward if we get feedback on this.

### `Schema` (`List` and `Page` components)
This is a trickier one. The `AmendDocuments` schema already exist which have workarounds in place to handle "drawing numbers" by another means. I'll open a Slack thread to check my assumptions here, but my current approach would be to leave the existing pattern in place and revisit this as an option once we return to generating `List` and `Page` schema in-Editor directly

## Demo
image

Storybook - https://storybook.6536.planx.pizza/?path=/story/planx-components-fileupload--with-drawing-numbers

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/FileUpload/Editor.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/FileUpload/Editor.tsx
@@ -101,7 +101,7 @@ function Component(props: Props) {
                   !formik.values.showDrawingNumber,
                 )
               }
-              label="Show a drawing number field for each uploaded file"
+              label="Show an optional drawing number input for each uploaded file"
               disabled={props.disabled}
             />
           </InputRow>
--- a/apps/editor.planx.uk/src/@planx/components/FileUpload/Editor.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/FileUpload/Editor.tsx
@@ -92,6 +92,19 @@ function Component(props: Props) {
               disabled={props.disabled}
             />
           </InputRow>
+          <InputRow>
+            <Switch
+              checked={formik.values.showDrawingNumber}
+              onChange={() =>
+                formik.setFieldValue(
+                  "showDrawingNumber",
+                  !formik.values.showDrawingNumber,
+                )
+              }
+              label="Show a drawing number field for each uploaded file"
+              disabled={props.disabled}
+            />
+          </InputRow>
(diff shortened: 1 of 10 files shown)


apps/editor.planx.uk/src/@planx/components/FileUpload/Editor.tsx
```diff
@@ -92,6 +92,19 @@ function Component(props: Props) {
...
+          <InputRow>
+            <Switch
+              checked={formik.values.showDrawingNumber}
+              onChange={() =>
+                formik.setFieldValue(
+                  "showDrawingNumber",
+                  !formik.values.showDrawingNumber,
+                )
+              }
+              label="Show a drawing number field for each uploaded file"
```
jessicamcinchak: nit: "Show an optional drawing number input for each uploaded file" ? 

Is it worth trying to emphasise to editors that just because you enable, does _not_ guarantee it'll come through payload?

jessicamcinchak (approved):
Great, thank you!

apps/editor.planx.uk/src/@planx/components/FileUpload/Editor.tsx
DafyddLlyr: This is a great shout - updated!

PR #6537, opened 2026-04-24
fix: Implement recent flows component using context, not router state

DafyddLlyr:
## What's the problem?
This follows on from the hotfix in #6530 with a more robust solution. Navigating into a nested flow via an external portal adds breadcrumbs to `recentFlows`. Once breadcrumbs were present, changing the component type in the "new node" modal would silently reset - the modal would snap back to its previous state.

## Root cause
The original implementation stored breadcrumb state in TanStack Router's `location.state`. Because every `navigate()` call drops `location.state` by default, a `useEffect` hook was used to re-hydrate it from `sessionStorage` whenever a drop was detected. This re-hydration fired its own `navigate()` call, which raced with the type-change navigation and overwrote it.

## What's the solution?
In this PR I've moved breadcrumb state out of the router entirely and into a React Context (`RecentFlowsProvider`). This eliminates the re-hydration race because there is no longer any navigation triggered by state management.

The context maintains a `JourneyMap` - an object of `{ historyKey: breadcrumb trail }` which is still persisted to `sessionStorage`. TanStack Router automatically assigns a unique key to every history entry, and restores that key on browser back/forward. The context handles the following events - 

- **New navigation** - record the breadcrumb trail for the new key, catching any pending portal/breadcrumb intent
- **Browser back/forward** - the key is already in the map, so the correct trail is just restored automatically
- **Intra-flow navigation** (opening modals, changing node types) - new key, no pending intent, trail is carried over unchanged. This is the equivalent of the `state={(prev) => prev}` from the previous implementation.

Portal clicks (`addRecentFlow()`) and breadcrumb clicks (`sliceRecentFlows()`) set persistent refs in their `onClick` handlers. These are then read by the provider when the next history key arrives, which allows us to update `recentFlows` correctly.

The PR is larger than planned sorry, but ~600 lines are new test files.

## Outcome
- Changing node types no longer resets the modal
- Browser back/forward correctly restores breadcrumbs
- Multiple routes to the same flow (A→B→X and A→C→X) are tracked independently because they produce unique history keys
- No `navigate()` calls are made by state management code
- No need for a custom Tanstack Link, or relying on `state={(prev) => prev}` calls everywhere


apps/editor.planx.uk/src/pages/FlowEditor/components/RecentFlows/RecentFlows.stories.tsx
```diff
@@ -50,42 +53,29 @@ const meta: Meta<typeof RecentFlows> = {
...
   render: () => <RecentFlows />,
-  parameters: {
-    tanstackRouter: {
-      location: {
-        pathname: "/",
-        state: { recentFlows: [recentFlows[0]] },
-      },
-    },
-  },
+  decorators: [withFlows([recentFlows[0]])],
```
jessicamcinchak, line 78: I don't typically think to reach for `decorators`, but makes really clean mocking solution here 👍

apps/editor.planx.uk/src/pages/FlowEditor/components/RecentFlows/RecentFlowsContext.test.tsx
```diff
@@ -0,0 +1,315 @@
...
+import React from "react";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import {
+  RecentFlowsProvider,
+  startNewRecentFlowsJourney,
+  useRecentFlowsContext,
+} from "./RecentFlowsContext";
+
+const STORAGE_KEY = "planx:recentFlows";
```
jessicamcinchak: nit: can `.test` and `.stories` files import this value rather than re-define?

jessicamcinchak (approved):
Really appreciate thorough write up + explanation here & extra test coverage ! 

All looks good, but hit one question on pizza (is consistent with staging so recognize not introduced here!): 
- I navigated from Barnet apply for LDC &rarr; to OSL LDC more details &rarr; OSL London data hub 
- Why does the "back to" navigation swap order of flows when expanded or contracted? 
  - Shouldn't the top "back to" line _always_ read my _most recent_ flow, and based on the arrow icon the "bottom" flow when expanded should always be _where I started my journey_? Was properly thrown off when I expanded "+1" and these swapped positions! 

Eg 
Image
Image

DafyddLlyr:
> _Shouldn't the top "back to" line always read my most recent flow, and based on the arrow icon the "bottom" flow when expanded should always be where I started my journey? Was properly thrown off when I expanded "+1" and these swapped positions!_

This is intentional behaviour, controlled here - 

https://github.com/theopensystemslab/planx-new/blob/0b05d1537e54b36a091c9613128de3fe10379594/apps/editor.planx.uk/src/pages/FlowEditor/components/RecentFlows/RecentFlows.tsx#L27-L32

We did discuss this at some point during development / design. Can you provide more context @ianjon3s ?

ianjon3s:
Indeed, this is intentional. The expanded view changes to show a full breadcrumb list of where you've been in the current session, emanating from the original flow.

PR #6539, opened 2026-04-26
chore(ci): Split up GHA into reusable composite actions and workflow

DafyddLlyr:
## What does this PR do?
Reduces repetition in our GHA workflows by extracting two composite actions and a reusable workflow. This was one of the first issues I hit last time I started on the monorepo work here - https://github.com/theopensystemslab/planx-new/pull/5458

**`.github/actions/setup-node-pnpm`**
Installs pnpm, sets up Node with pnpm cache, and applies `.gitconfig` - currently repeated 11× across workflows.

**`.github/actions/setup-pizza-env`**
Configures AWS credentials, pulls secrets from S3, and verifies `.env` files exist - currently repeated 8×  across workflows.

**`.github/workflows/build-planx-frontend.yml`**
Reusable workflow for building the PlanX Editor. Replaces a near-identical job duplicated across `push-main.yml` and `push-production.yml`.

### DIFF
--- a/.github/actions/setup-pizza-env/action.yml
+++ b/.github/actions/setup-pizza-env/action.yml
@@ -1,0 +1,39 @@
+name: Setup Pizza environment
+description: Configure AWS credentials, pull secrets from S3, and optionally verify .env files exist
+
+inputs:
+  aws-access-key-id:
+    description: AWS access key ID
+    required: true
+  aws-secret-access-key:
+    description: AWS secret access key
+    required: true
+  api-directory:
+    description: Path to API directory, used for .env.test existence check
+    required: true
+  hasura-directory:
+    description: Path to Hasura directory, used for .env.test existence check
+    required: true
+  check-env-files:
+    description: Whether to verify .env files exist after pulling from S3
+    required: false
+    default: 'true'
+
+runs:
+  using: composite
+  steps:
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v6
+      with:
+        aws-access-key-id: ${{ inputs.aws-access-key-id }}
+        aws-secret-access-key: ${{ inputs.aws-secret-access-key }}
+        aws-region: eu-west-2
+    - name: Copy .env files from Staging S3
+      shell: bash
+      run: ./scripts/pull-secrets.sh
+    - name: Check if .env files exist
+      if: inputs.check-env-files == 'true'
+      uses: andstor/file-existence-action@v2
+      with:
+        files: ".env, .env.staging, ${{ inputs.api-directory }}/.env.test, ${{ inputs.hasura-directory }}/.env.test"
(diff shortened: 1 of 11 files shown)


.github/workflows/build-planx-frontend.yml
```diff
@@ -0,0 +1,65 @@
...
+jobs:
+  build_planx_frontend:
+    name: Build PlanX frontend
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v6
+      - uses: ./.github/actions/setup-node-pnpm
+        with:
+          pnpm-version: ${{ vars.PNPM_VERSION }}
+          node-version: ${{ vars.NODE_VERSION }}
```
jessicamcinchak: nit: now that we have a single `setup-node-pnpm` file, shall we just hardcode/define our pnpm & node version there _once_  and drop these `with` configs ?

They aren't secret, and defining in/reading from `vars` in each referenced step feels less necessary if no longer truly repeating ?

.github/actions/setup-pizza-env/action.yml
```diff
@@ -0,0 +1,39 @@
...
+        aws-secret-access-key: ${{ inputs.aws-secret-access-key }}
+        aws-region: eu-west-2
+    - name: Copy .env files from Staging S3
+      shell: bash
+      run: ./scripts/pull-secrets.sh
+    - name: Check if .env files exist
+      if: inputs.check-env-files == 'true'
+      uses: andstor/file-existence-action@v2
+      with:
+        files: ".env, .env.staging, ${{ inputs.api-directory }}/.env.test, ${{ inputs.hasura-directory }}/.env.test"
```
jessicamcinchak, line 35: nit: similar to other comment, what's the value of `with: ${{ inputs.X }}` here ? Is there a legit use case where one instantiation would use a _unique_ aws key or API directory? 

Can these new shared action files reference the `${{ vars.X }}` or `${{ secrets.X }}` directly and drop need for input params entirely if always expected to be the same?

jessicamcinchak (approved):
Looks good ! A couple minor comments/suggestions about `inputs` versus `vars` !

.github/workflows/build-planx-frontend.yml
DafyddLlyr: Very good catch..! Implemented via default values which simplifies things - no need for `vars` at all.

.github/actions/setup-pizza-env/action.yml
DafyddLlyr: This one is trickier as actions can't read secrets directly - they have to be passed in.

However, I'd missed that I can use default values for the API and Hasura directory paths to skip passing these in. Similarly, the optional `check-env-files` is sort of redundant - we should just alway do it - we don't need this argument really.

PR #6540, opened 2026-04-27
feat(monorepo): Configure pnpm workspaces

DafyddLlyr:
## What does this PR do?
This PR implements most of  Phase 1 of [ADR-0010: Migrate to a monorepo structure](https://github.com/theopensystemslab/planx-new/blob/main/doc/architecture/decisions/0010-migrate-to-a-monorepo-structure.md), which outlines our transition to a fully configured Turborepo monorepo.

### Why pnpm workspaces before Turborepo?
Turborepo needs to sit on top of a workspace-aware package manager. Getting the workspace wiring correct first (single lockfile, single install, consistent dependency resolution) makes adding the Turborepo layer a (hopefully) simple addition rather than a messy debugging exercise with too many moving parts. I learnt this the hard way trying to implement Turborepo directly [on my previous attempt](https://github.com/theopensystemslab/planx-new/pull/5458)...!

This PR is deliberately just setting up pnpm workspaces, nothing else.

### Single workspace
All per-app `pnpm-workspace.yaml` files and lockfiles are replaced with a single root `pnpm-workspace.yaml` and `pnpm-lock.yaml`. All apps, e2e suites, and scripts are now workspace members with explicit `name` fields in their `package.json` files.

### Docker
Both the `api` and `sharedb` Dockerfiles have been updated to work with the workspace configuration.

### CI/CD
All GitHub Actions install steps now run once at the workspace root, instead of per-project installs.

### Basic package alignment
In a pnpm workspace only the root `pnpm.overrides` can be used, so per-app overrides have moved up to here.

## Testing locally
The Pizza here seems to be working, as is the CI/CD pipeline, but I'd be keen for a few devs to try this locally. The main change you should notice is that `pnpm i` just has to be ran once now in the workspace root, rather than per-app.

```sh
# After pulling this branch
pnpm install

# Start the full stack (unchanged)
pnpm start

# Run tests for a specific app
# XXX: Once we setup Turborepo, the syntax for this will change again...!
pnpm --filter api.planx.uk test
pnpm --filter editor.planx.uk test

# Or just cd into the app as before - pnpm will find the root workspace
cd apps/api.planx.uk
pnpm test
```

## Next steps...
This isn't exhaustive, please see the plan in the ADR for that. These are just the next few PRs I have in mind here which should follow along quite soon. 

- [x] Update dependabot config so that it works with a single pnpm workspace
  - Turns out this should "just work"..! [Dependabot already supports pnpm workspaces](https://github.blog/changelog/2025-02-04-dependabot-now-supports-pnpm-workspace-catalogs-ga/), so will pick up dependencies in the root catalog.
- [ ] Use `pnpm catalog` to consolidate shared common packages, were version numbers count and often trip us up - `typescript`, `graphql`, `uuid`, `jsonwebtoken`, `lodash`, `planx-core` (!) 
  - PR here - https://github.com/theopensystemslab/planx-new/pull/6573
- [ ] Setup very basic shared configs in `/packages` - ESLint, Prettier, TypeScript
- [ ] Install Turborepo, update CI, local dev config, and Dockerfiles to use this

Once this is done we can look at migrating `planx-core` 🚀


.github/workflows/pull-request.yml
```diff
@@ -457,15 +452,17 @@ jobs:
...
         working-directory: e2e
       - name: Install Playwright Dependencies
-        run: pnpm playwright install --with-deps
+        run: pnpm playwright install --with-deps chromium
         working-directory: e2e/tests/ui-driven
       - name: Start test containers
         run: ./scripts/start-containers-for-tests.sh
+      - name: Build editor (test mode)
+        run: VITE_APP_ENV=test pnpm build
+        working-directory: apps/editor.planx.uk
```
DafyddLlyr, line 500: Before we had a whole script just for these changes (`e2e/tests/ui-driven/install-dependencies.sh`) - this just meant it was harder to follow the thread here I found. This seems clearer and more explicit.

apps/api.planx.uk/modules/lps/service/generateHTML.ts
DafyddLlyr, line 1: Required type change due this now correctly inferring the types.

apps/api.planx.uk/Dockerfile
```diff
@@ -1,38 +1,33 @@
...
-COPY --from=base /api/dist ./dist
-COPY --from=base /api/node_modules ./node_modules
+COPY --from=base /deploy/api ./
 CMD ["npm", "start"]
 
 ## DEVELOPMENT ##
 FROM base AS development
 ENV NODE_ENV=development
-RUN pnpm install
+RUN npm_config_inject_workspace_packages=true pnpm --filter api.planx.uk deploy --prefer-offline /api
```
DafyddLlyr, line 34: Once we configure Turborepo, the syntax here will be updated again.

README.md
DafyddLlyr, line 1: This file will get a more thorough update explaining the monorepo setup when completed 👍

apps/api.planx.uk/Dockerfile
```diff
@@ -1,38 +1,33 @@
-## BASE ##
```
freemvmt, line 1: keep this if we're keeping the other tags below, e.g. `## PRODUCTION ##`?

apps/api.planx.uk/Dockerfile
```diff
@@ -1,38 +1,33 @@
...
+RUN pnpm --filter api.planx.uk build
+RUN npm_config_inject_workspace_packages=true pnpm --filter api.planx.uk deploy --prod --prefer-offline /deploy/api
 
-RUN pnpm build
-RUN pnpm prune --production
 
 FROM node:24.14.0-alpine AS production
 WORKDIR /api
 
 ## PRODUCTION ##
```
freemvmt: move this above `FROM` to be consistent/clearer?

apps/api.planx.uk/Dockerfile
```diff
@@ -1,38 +1,33 @@
...
+COPY pnpm-workspace.yaml pnpm-lock.yaml package.json ./
 RUN pnpm fetch
 
-# copy and install everything else using cache where possible
-COPY . ./
-RUN pnpm install --recursive --prefer-offline
+COPY . .
+RUN pnpm install --frozen-lockfile --prefer-offline
+RUN pnpm --filter api.planx.uk build
+RUN npm_config_inject_workspace_packages=true pnpm --filter api.planx.uk deploy --prod --prefer-offline /deploy/api
```
freemvmt, line 20: where did you find this long npm flag? nothing of much use comes up when I search ye olde internet? I do understand the the need for it, but closest thing I could find was `inject-workspace-packages`, as discussed in [this github issue](https://github.com/orgs/pnpm/discussions/3938#discussioncomment-12050478)?

freemvmt (approved):
Woah, big one. Looks good! So many deleted files woop. I built the stack on local and everything works as expected. Seemed to take longer than usual but might've been a fluke on my comp.

Possibly another thing for the TODOs is to merge the pnpm workspace in `infrastructure/` into the top level?

zz-hh-aa:
🚀  Tested on WSL, `pnpm install` and `pnpm start` worked like a dream ✨ (and didn't feel slow for me as Dan reported)

Some of my tests timed out but I think they were longer-running React tests? Successfully ran them in any case, both from root and app directories.

.dockerignore
```diff
@@ -0,0 +1,10 @@
+.git
+.claude
+.github
+node_modules
+e2e
+infrastructure
+apps/editor.planx.uk/build
+apps/editor.planx.uk/node_modules
```
joecarver: can we also ignore the editor src files (and in theory everything else except the `package.json` which I guess is needed for the workspace config)? But the `src` would be the biggest thing I guess

joecarver (commented):
still a couple of per-app lockfiles knocking about:
- `apps/sharedb.planx.uk/pnpm-lock.yaml`
- `scripts/encrypt/pnpm-lock.yaml`
- `scripts/image-replace/pnpm-lock.yaml`

which I assume can be removed? correct me if i'm wrong though!

.dockerignore
DafyddLlyr: Good observation thanks - fixed in [7a2d93b](https://github.com/theopensystemslab/planx-new/pull/6540/commits/7a2d93b2f29f998b4f300201c6d4bbb9e7cc814e)

DafyddLlyr:
> still a couple of per-app lockfiles knocking about:

Missed these in the first pass (or they were re-introduced in a rebase?). Now tidied up..! 🗑️ 🧹

apps/api.planx.uk/Dockerfile
DafyddLlyr: Yeah this one took me a while to figure out how to set 😅 

There's actually two things going on here - npm allows variable injection using the `npm_config_` prefix ([docs](https://docs.npmjs.com/cli/v10/using-npm/config#environment-variables)), and pnpm needs `inject_workspace_packages` set to `true` ([docs](https://pnpm.io/cli/deploy)). By combining both of these behaviours we get `npm_config_inject_workspace_packages=true`.

I've added comments to the Dockerfiles explaining this setup, please see  [94135e7](https://github.com/theopensystemslab/planx-new/pull/6540/commits/94135e762a15a50f9323efe903bfa3388db2b662) 👍

jessicamcinchak:
Late to this party, but just confirming that `pnpm install` and `pnpm start` also working locally for me ! :heavy_check_mark: :star2:

PR #6541, opened 2026-04-27
fix: skip `govUkPayment` initialisation when Pay is in "info-only" mode

jessicamcinchak:
See https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1777306960440849

To test :
- On staging, create a basic submission flow with a Pay component in "info-only" mode. When you "continue" through it, check your `lowcal_session.data` for a "govUkPayment" key with status "created"
- On this pizza, for the same flow, confirm your `lowcal_session.data` does _not_ have a `govUkPayment` property at all after "Continue"

(Adding an e2e api-driven test would also be good follow-up here!)

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
@@ -245,8 +245,8 @@ function Component(props: Props) {
   const startNewPayment = async () => {
     dispatch(Action.StartNewPayment);
 
-    // Skip the redirect process if viewing this within the Editor
-    if (environment !== "standalone") {
+    // Skip the redirect process if viewing this within the Editor or using Pay in info-only mode
+    if (environment !== "standalone" || props.hidePay) {
       handleSuccess();
       return;
     }


apps/editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
```diff
@@ -245,8 +245,8 @@ function Component(props: Props) {
   const startNewPayment = async () => {
     dispatch(Action.StartNewPayment);
 
-    // Skip the redirect process if viewing this within the Editor
-    if (environment !== "standalone") {
+    // Skip the redirect process if viewing this within the Editor or using Pay in info-only mode
+    if (environment !== "standalone" || props.hidePay) {
```
jessicamcinchak, line 249: From my understanding here, info-only Pay nodes are in `"init"` status on initial render (per this component's own `ComponentState` machine), and therefore triggering `startNewPayment`. 

We don't intend for this to happen! An info-only Pay node should _never_ actually connect to Gov Pay (eg councils without Gov Pay currently configured should be able to use in this mode!). 

This one-liner seemed simple enough, alternatively we could add a `ComponentState` status for "info-only" mode, but that feels redundant with `hidePay` prop?

DafyddLlyr: This works well..!

DafyddLlyr (approved):

PR #6552, opened 2026-04-28
fix(editor): Use `on` prefix for custom element events (React 19)

DafyddLlyr:
## What's the problem?
The address autocomplete component is not working as expected - nothing happens on address selection, meaning the user cannot proceed.

## What's the cause?
In [React 19](https://github.com/theopensystemslab/planx-new/pull/6478) architectural changes the component body executes strictly before the DOM is updated or painted. As a result, when the component renders, the `<address-autocomplete>` element doesn't exist in the DOM yet, `getElementById()` returns `null`, and the event listener silently fails to attach.

## What's the solution?
Leverage the native custom element support in React 19, but attaching a listener directly via the `on` prefix.

Docs - https://react.dev/reference/react-dom/components#custom-element-events

## Next steps...
Regression tests running here - https://github.com/theopensystemslab/planx-new/actions/runs/25060274739 ✅ 
 - One ITP test is failing - this is related to a separate issue, which we're also looking into.

The `useGeoJSONChange()` hook had already accounted for this React 19 change, but most `<my-map/>` elements do not implement this.

https://github.com/theopensystemslab/planx-new/blob/38e8e722ed084012c74c0bd9720768d49985d975/apps/editor.planx.uk/src/lib/gis.ts#L36-L48

## Testing
The following components have been touched via this change / React 19 bump - 
 - `FindProperty`
 - `DrawBoundary`
 - `List` (with `MapFields`)
 - `Page` (with `MapFields`)
 - `MapAndLabel`

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
@@ -16,8 +16,9 @@ import { squareMetresToHectares } from "@planx/components/shared/utils";
 import buffer from "@turf/buffer";
 import { point } from "@turf/helpers";
 import { Feature } from "geojson";
+import type { GeoJSONChangeEvent } from "lib/gis";
 import { Store, useStore } from "pages/FlowEditor/lib/store";
-import React, { useEffect, useRef, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { FONT_WEIGHT_SEMI_BOLD } from "theme";
 import FullWidthWrapper from "ui/public/FullWidthWrapper";
 import ErrorWrapper from "ui/shared/ErrorWrapper";
@@ -108,28 +109,21 @@ export default function Component(props: Props) {
   useEffect(() => {
     if (isMounted.current) setSlots([]);
     isMounted.current = true;
+  }, [setSlots]);
 
-    const geojsonChangeHandler = ({ detail: geojson }: any) => {
-      if (geojson["EPSG:3857"]?.features) {
-        // only a single polygon can be drawn, so get first feature in geojson "FeatureCollection"
-        setBoundary(geojson["EPSG:3857"].features[0]);
-        setArea(
-          geojson["EPSG:3857"].features[0]?.properties?.["area.squareMetres"],
-        );
-      } else {
-        // if the user clicks 'reset' to erase the drawing, geojson will be empty object, so set boundary to undefined & area to 0
-        setBoundary(undefined);
-        setArea(0);
-      }
-    };
-
-    const map: any = document.getElementById("draw-boundary-map");
-    map?.addEventListener("geojsonChange", geojsonChangeHandler);
-
-    return function cleanup() {
-      map?.removeEventListener("geojsonChange", geojsonChangeHandler);
-    };
-  }, [page, setArea, setBoundary, setSlots]);
+  const geojsonChangeHandler = ({ detail: geojson }: GeoJSONChangeEvent) => {
+    if (geojson["EPSG:3857"]?.features) {
+      // only a single polygon can be drawn, so get first feature in geojson "FeatureCollection"
+      setBoundary(geojson["EPSG:3857"].features[0]);
+      setArea(
+        geojson["EPSG:3857"].features[0]?.properties?.["area.squareMetres"],
+      );
+    } else {
+      // if the user clicks 'reset' to erase the drawing, geojson will be empty object, so set boundary to undefined & area to 0
+      setBoundary(undefined);
+      setArea(0);
+    }
+  };
 
   /**
    * Declare refs to hold a mutable copy the up-to-date validation errors
@@ -307,6 +301,7 @@ export default function Component(props: Props) {
                   collapseAttributions={
                     self.innerWidth < 500 ? true : undefined
                   }
+                  ongeojsonChange={geojsonChangeHandler}
                 />
               </MapContainer>
             </ErrorWrapper>
--- a/apps/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
@@ -227,14 +227,8 @@ export default function Component(props: Props) {
    * Clip map extent to buffered boundary, with a fallback to address point if required
    */
   const clipGeojsonData = (() => {
-    if (boundary)
-      return JSON.stringify(
-        buffer(boundary, bufferInMeters, { units: "meters" }),
-      );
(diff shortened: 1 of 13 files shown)


apps/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
```diff
@@ -108,28 +109,21 @@ export default function Component(props: Props) {
...
-    const map: any = document.getElementById("draw-boundary-map");
-    map?.addEventListener("geojsonChange", geojsonChangeHandler);
-
-    return function cleanup() {
-      map?.removeEventListener("geojsonChange", geojsonChangeHandler);
-    };
-  }, [page, setArea, setBoundary, setSlots]);
+  }, [setSlots]);
+
+  const geojsonChangeHandler = ({ detail: geojson }: GeoJSONChangeEvent) => {
```
DafyddLlyr, line 114: `useEffect()` wrapper dropped, applied directly as an event handler via `ongeojsonChange` prop on `<my-map/>`.

apps/editor.planx.uk/src/@planx/components/FindProperty/Public/Autocomplete.tsx
```diff
@@ -59,14 +59,14 @@ export default function PickOSAddress(props: PickOSAddressProps): FCReturn {
   const { data } = useBLPUCodes();
 
   const { setAddress } = props;
-  useEffect(() => {
-    if (!data?.blpuCodes) return;
-    // Handles mapping the raw Ordnance Survey record to planx's address model
-    const addressSelectionHandler = ({
+
+  const addressSelectionHandler = ({
```
DafyddLlyr, line 63: `useEffect()` wrapper dropped, applied directly as an event handler via `onaddressSelection` prop on `<address-autocomplete/>`.

apps/editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
```diff
@@ -48,6 +49,8 @@ export const DescriptionInput = styled(Box)(({ theme }) => ({
 }));
 
 export default function PlotNewAddress(props: PlotNewAddressProps): FCReturn {
+  const { setAddress, setMapValidationError } = props;
```
DafyddLlyr, line 52: Split these out here to silence `react-hooks/exhaustive-deps` below on a few `useEffect()` calls.

apps/editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
```diff
@@ -67,27 +70,25 @@ export default function PlotNewAddress(props: PlotNewAddressProps): FCReturn {
...
-    };
-
-    const map: any = document.getElementById("plot-new-address-map");
-    map?.addEventListener("geojsonChange", geojsonChangeHandler);
-
-    return function cleanup() {
-      map?.removeEventListener("geojsonChange", geojsonChangeHandler);
-    };
-  }, [setCoordinates]);
+  const geojsonChangeHandler = ({ detail: geojson }: GeoJSONChangeEvent) => {
```
DafyddLlyr, line 73: `useEffect(`) wrapper dropped, applied directly as an event handler via `ongeojsonChange` prop on `<my-map/>`.

apps/editor.planx.uk/src/@planx/components/FindProperty/Public/Map.tsx
```diff
@@ -67,27 +70,25 @@ export default function PlotNewAddress(props: PlotNewAddressProps): FCReturn {
...
+      const geomWebMercator = geojson["EPSG:3857"].features[0]?.geometry;
+      const geomBNG = geojson["EPSG:27700"].features[0]?.geometry;
+
+      // Type-narrowing to exclude GeometryCollection
+      const coordsWebMercator =
+        geomWebMercator?.type === "Point" ? geomWebMercator.coordinates : [];
+      const coordsBNG = geomBNG?.type === "Point" ? geomBNG.coordinates : [];
+
+      const [longitude, latitude] = coordsWebMercator;
+      const [x, y] = coordsBNG;
```
DafyddLlyr, line 85: Small refactor here to satisfy the correct `GeoJSONChangeEvent` type - prior this it was just typed as `any`.

apps/editor.planx.uk/src/lib/gis.ts
DafyddLlyr, line 1: Hook dropped, no longer used - this can be done natively in React 19 🚀

apps/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
```diff
@@ -288,7 +276,7 @@ export default function Component(props: Props) {
                   ariaLabelOlFixedOverlay="An interactive map for providing your location plan boundary"
                   drawMode
                   drawPointer="crosshair"
-                  drawGeojsonData={JSON.stringify(boundary)}
+                  drawGeojsonData={boundary}
```
DafyddLlyr, line 279: **Why are we stripping away `JSON.stringify()`?**
React 19 is passing this _as props_ now, not just pure _attributes_.

> _props that match a property on the Custom Element instance will be assigned as properties_

Source: https://react.dev/blog/2024/12/05/react-19#support-for-custom-elements

jessicamcinchak: Didn't expect this one! 

Do we need to care about uses of `JSON.stringify()` beyond geo-components here - eg `SetFee`, `SummaryList` and `RichTextInput` also show up on quick search through editor? (Also happy to keep an eye out for these on QA & fix forward in separate PR without adding complexity here!)

DafyddLlyr: This is a webcomponent / custom-element specific thing, so other instances of this are fine - these are components that expect `string` input.

What was happening was that before this was...
 - React can only pass values as attributes into webcomponents. HTML attributes have to be strings.
 - Lit was parsing these back to the proper values, thanks to the `@property({ type: Object })` decorators
 
 But now with React 19...
  - React sees `my-map` as a class, and is aware there's a `drawGeojsonData` property on the class
  - It directly assigns to memory - as we were giving it a string, it was a string inside the Lit component when it should have been an object
  - This basically means we can just treat these values as "props" all the way down the chain from now on

jessicamcinchak: Helpful explanation / clarification, thank you!! :heavy_check_mark:

jessicamcinchak (approved):
All working for me locally ! Thank you again!

PR #6555, opened 2026-04-28
feat: limit team sizes to 20 members

zz-hh-aa:
Row-level post-update checks can't seem to run any aggregate functionality, so I can't count the number of rows in `team_members` with a specific `team_id` (correct me if I'm wrong but that's what it looks like [here](https://hasura.io/docs/2.0/auth/authorization/permissions/permissions-operators/)!). So that's why I've gone with an event trigger here. 

In the UI, we also have a new 'Maximum users reached' modal that blocks adding (seemed neater than allowing the interaction only to show an error afterwards).
image

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MaximumUserModal.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MaximumUserModal.tsx
@@ -1,9 +1,7 @@
 import Dialog from "@mui/material/Dialog";
-import DialogActions from "@mui/material/DialogActions";
 import DialogContent from "@mui/material/DialogContent";
 import DialogTitle from "@mui/material/DialogTitle";
 import Typography from "@mui/material/Typography";
-import { Form, Formik } from "formik";
 import React from "react";
 
 import { AddUserModalProps } from "../types";
@@ -12,7 +10,7 @@ export const MaximumUserModal: React.FC<AddUserModalProps> = ({ onClose }) => {
   return (
     <Dialog
       aria-labelledby="dialog-heading"
-      data-testid="dialog-add-user"
+      data-testid="dialog-maximum-users"
       open
       onClose={onClose}
     >
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MaximumUserModal.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MaximumUserModal.tsx
@@ -1,0 +1,30 @@
+import Dialog from "@mui/material/Dialog";
+import DialogActions from "@mui/material/DialogActions";
+import DialogContent from "@mui/material/DialogContent";
+import DialogTitle from "@mui/material/DialogTitle";
+import Typography from "@mui/material/Typography";
+import { Form, Formik } from "formik";
+import React from "react";
+
+import { AddUserModalProps } from "../types";
+
+export const MaximumUserModal: React.FC<AddUserModalProps> = ({ onClose }) => {
+  return (
+    <Dialog
+      aria-labelledby="dialog-heading"
+      data-testid="dialog-add-user"
+      open
+      onClose={onClose}
+    >
+      <DialogTitle variant="h3" component="h1" id="dialog-heading">
+        Maximum users reached
+      </DialogTitle>
+      <DialogContent>
+        <Typography>
+          The maximum number of users already exists for this team. To add
+          another member, please remove an existing one or contact an admin.
+        </Typography>
+      </DialogContent>
+    </Dialog>
+  );
+};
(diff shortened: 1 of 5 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MaximumUserModal.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { Form, Formik } from "formik";
+import React from "react";
+
+import { AddUserModalProps } from "../types";
+
+export const MaximumUserModal: React.FC<AddUserModalProps> = ({ onClose }) => {
+  return (
+    <Dialog
+      aria-labelledby="dialog-heading"
+      data-testid="dialog-add-user"
```
DafyddLlyr: 👀

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
```diff
@@ -166,7 +166,10 @@ export const MembersTable = ({
             {showAddMemberButton && (
               <TableRow>
                 <TableCell colSpan={5}>
-                  <AddButton onClick={addUser}>Add a new member</AddButton>
+                  {/* the hard-coded team member limit value should be kept in sync with the hasura trigger (see enforce_team_member_limit) */}
+                  <AddButton onClick={members.length < 20 ? addUser : maxUsers}>
```
DafyddLlyr: nit: Rather than an inline comment and magic number, I'd recommend something like this...

```
/** The hard-coded team member limit value should be kept in sync with the hasura trigger (see enforce_team_member_limit **/
const MAX_USER_COUNT = 20
```

apps/hasura.planx.uk/migrations/default/1777376005470_create_limit_team_members_trigger/up.sql
DafyddLlyr, line 1: How does this handle "archived users" (soft deleted, email set to `null`, `team_member` records still in place)?

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MaximumUserModal.tsx
zz-hh-aa: Good catch 🔍 I updated it to `dialog-maximum-users` in [4ff1f11](https://github.com/theopensystemslab/planx-new/pull/6555/commits/4ff1f11806321e59bb4ba61b85fc651bf70c6f81) but also noticed that we didn't have tests that used the `dialog-add-user` test in the first place... where would be the best place for those?

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
zz-hh-aa: Done in [4ff1f11](https://github.com/theopensystemslab/planx-new/pull/6555/commits/4ff1f11806321e59bb4ba61b85fc651bf70c6f81)

apps/hasura.planx.uk/migrations/default/1777376005470_create_limit_team_members_trigger/up.sql
zz-hh-aa: Ah thanks for thinking of this! I have updated the trigger function `WHERE` to check that `users.email` `IS NOT NULL`. I tested this by removing the frontend check and then removing users / making sure that the db limit kicked in at 20 active users. See [f3a6687](https://github.com/theopensystemslab/planx-new/pull/6555/commits/f3a6687cd3d024a6d88da617c9f4babeb9890832)

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MembersTable.tsx
DafyddLlyr: Great - much clearer and more maintainable 💪

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/MaximumUserModal.tsx
DafyddLlyr: Yeah we're pretty light on any testing for the Editor interface (settings, user management etc) is the truth, this is something we should aim to deal with - it will be more of a concern as the number of teams and users grow (next ~6 months).

I think we'd get a lot of value from testing these via RTL as frontend tests - we can use `msw` for mock responses to mimic Hasura.

DafyddLlyr (approved):
Looking good - thanks for those changes!

It would be worth following this up with some tests that describe the functionality.

`react-testing-library`
 - The modal pops up if frontend has 20 users already

`e2e/tests/api`
 - Given a team with 20 members, I get an API error if I run a mutation to insert another one

PR #6564, opened 2026-04-29
chore(ci): Configure linting for GitHub actions and workflows

DafyddLlyr:
We've had a few issues in the past where a broken GHA has lead to downstream issues.

This PR sets up a new GHA which will run a linting process when a PR is opened on any changed files using https://github.com/rhysd/actionlint 

This does not currently support actions, just workflows - https://github.com/rhysd/actionlint/issues/46

We could also set up something like this to run on a pre-commit hook - I'll revisit this once we resolve the current Husky / monorepo setup (soon! 🤞)

### DIFF
--- a/.github/workflows/lint-gha.yml
+++ b/.github/workflows/lint-gha.yml
@@ -1,0 +1,19 @@
+name: Lint GHA workflows and actions
+
+on:
+  pull_request:
+    paths:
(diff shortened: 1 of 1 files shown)


DafyddLlyr:
I'll pick this up again shortly as a few recent / queued PRs are touching these workflows.

PR #6568, opened 2026-04-29
chore(editor): ESLint rule to catch MUI dot notation syntax

DafyddLlyr:
To follow on from https://github.com/theopensystemslab/planx-new/pull/6567 - this should fail until all instances of this are resolved.

Fixes implemented by running `npx @mui/codemod@latest v9.0.0/system-props .` in Editor directory, plus sweep up using `pnpm lint:error`

Docs - https://mui.com/material-ui/migration/upgrade-to-v9#system-props

PR #6569, opened 2026-04-29
fix: Update text color properties

ianjon3s:
## Issue

- MUI 9 syntax no longer supports color properties applied to components using theme tokens (for example) `text.primary`

## Solution

- Use semantic color properties, i.e. `textPrimary`


ianjon3s:
Closing as taken care of with https://github.com/theopensystemslab/planx-new/pull/6568

PR #6573, opened 2026-04-30
feat(monorepo): Setup pnpm catalog

DafyddLlyr:
## What does this PR do?
Configures several [pnpm catalogs](https://pnpm.io/catalogs) in order to simplify and standardise our package management.


pnpm-workspace.yaml
```diff
@@ -7,18 +7,57 @@ packages:
   - 'e2e/tests/api-driven'
   - 'infrastructure/*'
   - 'scripts/*'
-  
-minimumReleaseAge: 10080
 
-minimumReleaseAgeExclude:
-  - dompurify@3.4.11
+catalog:
+  "@opensystemslab/planx-core": "github:theopensystemslab/planx-core#2f19e28"
```
DafyddLlyr: Now we only need to update in one place...! 🚀

pnpm-workspace.yaml
```diff
@@ -7,18 +7,57 @@ packages:
...
+    graphql-tag: ^2.12.6
+    graphql: ^16.13.2
+
   infrastructure:
     '@pulumi/aws': ^7.24.0
     '@pulumi/awsx': ^3.4.0
     '@pulumi/cloudflare': ^6.14.0
     '@pulumi/postgresql': ^3.16.2
     '@pulumi/pulumi': ^3.229.0
     '@types/node': ~24.10.15
```
freemvmt: can remove this here too?

freemvmt (approved):
Nice! I can run `pnpm i` from root just fine on the branch.

apps/api.planx.uk/package.json
```diff
@@ -92,35 +93,35 @@
...
     "supertest": "^7.2.2",
     "tsx": "^4.21.0",
-    "typescript": "~5.9.3",
-    "uuid": "^14.0.0",
-    "vite": "^6.4.3",
-    "vitest": "^4.0.18"
+    "typescript": "catalog:typescript",
+    "uuid": "catalog:",
+    "vite": "catalog:",
+    "vitest": "catalog:"
```
jessicamcinchak, line 125: if `vitest` is now `"catalog"` managed, should all other `@vitest/*` dev dependencies above also be (for central reference-ability more than function I guess if not actually shared among apps currently)?

apps/localplanning.services/package.json
```diff
@@ -23,23 +23,23 @@
...
     "@storybook/react-vite": "^10",
-    "@types/node": "~24.10.15",
-    "msw": "^2.14.3",
+    "@storybook/react": "^10",
+    "@types/node": "catalog:",
     "msw-storybook-addon": "^2.0.7",
+    "msw": "^2.14.3",
     "storybook": "^10",
     "typescript-plugin-css-modules": "^5.2.0",
     "vite": "^7.3.5"
```
jessicamcinchak, line 45: Have we missed a `"catalog"` here for vite, or LPS is intentionally on a different version? For my own understanding, is it possible/desirable for us to define/reference a "vite6" _and_ "vite7" in our catalog in cases where we have intentional drift?

pnpm-workspace.yaml
```diff
@@ -7,18 +7,57 @@ packages:
...
+  "@opensystemslab/planx-core": "github:theopensystemslab/planx-core#2f19e28"
+  "@types/jsdom": ^28.0.1
+  "@types/node": ~24.10.15
+  axios: ^1.15.0
+  dompurify: ^3.4.11
+  dotenv: ^17.3.1
+  eslint-config-prettier: ^9.1.2
+  eslint: ^8.57.1
+  husky: ^8.0.3
+  jsdom: ^28.1.0
```
jessicamcinchak, line 21: What's the difference between including `jsdom` and `@types/jsdom` here under `catalog` versus under plural `catalogs` below? Just stylistic/organisational or any functional differences/limitations to be aware of?

jessicamcinchak (approved):
A couple questions, no blockers !

pnpm-workspace.yaml
DafyddLlyr: Good spot - this was unused and now removed!

apps/api.planx.uk/package.json
DafyddLlyr: Yep good catch - I've move the shared `@vitest/eslint-plugin` to the catalog, but left dependencies locally which have just one instance.

pnpm-workspace.yaml
DafyddLlyr: `catalog:` is a single (default) catalog, but `catalogs:*` allows us to name specific grouping (if meaningful). So yep, it's just organisational really 👌

apps/localplanning.services/package.json
DafyddLlyr: This is intentional - we can differentiate and overwrite package numbers where required (both newer and older versions).

Vite here is on a higher version intentionally. If we had a long term version drift we can define `vite6 ` and `vite7` packages within the base catalog (or if we have another app on `vite7`). This would keep the versions consistent and is a good idea if we find ourselves here.

Probably the best solution is just to get everything onto v7 (or v8 now!)

pnpm-workspace.yaml
freemvmt: on this, is the point of the separate catalogs to enable us to manage them separately, e.g. to upgrade all packages related to `vite` at a time, without touching others? or more for readability?

freemvmt: ngl still struggling every time with this spelling of catalogue 😅

PR #6576, opened 2026-04-30
test: e2e tests for templates

joecarver:
### What's in this PR?
Set of e2e tests for template creation, updates, and creating flows from templates.

### Notes
- Separate test blocks for admin actions and editor actions
- Tests are very much still interdependent - the entire file must be run as most tests depend on an action carried out by a previous test
- Added a few new helpers for navigation and creating flows

### DIFF
--- a/e2e/tests/ui-driven/src/helpers/navigateAndPublish.ts
+++ b/e2e/tests/ui-driven/src/helpers/navigateAndPublish.ts
@@ -7,6 +7,13 @@ export const navigateToService = async (page: Page, slug: string) => {
   await expect(page.getByRole("link", { name: slug })).toBeVisible();
 };
 
+export const navigateToTeamPage = async (page: Page) => {
+  await page.goto(`/app/${contextDefaults.team.slug}`);
+
+  await expect(page.getByText(contextDefaults.team.name).first()).toBeVisible();
+  await expect(page.locator("h1", { hasText: "Flows" })).toBeVisible();
+};
+
 export const publishService = async (page: Page) => {
   // Open modal
   await page.getByTestId("check-for-changes-to-publish-button").click();
@@ -26,7 +33,9 @@ export const publishService = async (page: Page) => {
   await page.getByTestId("publish-button").click();
 
   // Wait for confirmation of success
-  await expect(page.getByText("Successfully published changes")).toBeVisible();
+  await expect(
+    page.getByText("Successfully published changes").first(),
+  ).toBeVisible();
 };
 
 export const turnServiceOnline = async (page: Page) => {
--- a/e2e/tests/ui-driven/src/pages/Editor.ts
+++ b/e2e/tests/ui-driven/src/pages/Editor.ts
@@ -117,29 +117,6 @@ export class PlaywrightEditor {
     ).toBeVisible();
   }
 
-  // Create a new node in a templated flow
-  async createNodeWithTemplateConfig(
-    title: string,
-    instructions: string,
-    required: boolean,
-    locator?: Locator,
-  ): Promise<void> {
-    await (locator ?? this.firstNode).click();
-    await this.page.getByRole("dialog").waitFor();
-
-    await this.page.getByPlaceholder("Text").fill(title);
-
-    // force: true is necessary for MuiSwitch inputs
-    await this.page.getByLabel("Allow edits").click({ force: true });
-    await this.page.locator("#templatedNodeInstructions").fill(instructions);
-    if (required) {
-      await this.page.getByLabel("Require edits").click({ force: true });
-    }
-
-    await this.page.locator('button[form="modal"][type="submit"]').click();
-    await this.page.getByRole("dialog").waitFor({ state: "detached" });
-  }
-
   async createQuestion() {
     await createQuestionWithOptions(
       this.page,
--- a/e2e/tests/ui-driven/src/pages/Editor.ts
+++ b/e2e/tests/ui-driven/src/pages/Editor.ts
@@ -78,6 +78,68 @@ export class PlaywrightEditor {
     this.page.getByRole("button", { name: "Add flow " }).click();
   }
 
+  async addNewSourceTemplate(name: string): Promise<void> {
+    const openModalButton = this.page.locator("button", {
+      hasText: "Add a new flow",
+    });
+    await openModalButton.click();
+
+    await this.page.getByLabel("Flow name").fill(name);
+
+    await this.page.getByLabel("Source template").click();
+
+    await this.page.getByRole("button", { name: "Add flow" }).click();
+
+    await expect(this.firstNode).toBeVisible();
+  }
+
+  // Creates a templated flow from a named source template.
+  async addNewFlowFromTemplate(sourceTemplateName: string): Promise<void> {
+    const openModalButton = this.page.locator("button", {
+      hasText: "Add a new flow",
+    });
+    await openModalButton.click();
+
+    // Switch mode to "From a template..."
+    await this.page.locator('[aria-labelledby~="create-flow-mode"]').click();
+    await this.page.getByRole("option", { name: "From a template..." }).click();
+
+    // Select the source template
+    await this.page
+      .locator('[aria-labelledby~="available-templates-select"]')
+      .click();
+    await this.page.getByRole("option", { name: sourceTemplateName }).click();
+
+    // Save flow and redirect
+    await this.page.getByRole("button", { name: "Add flow " }).click();
+    await expect(
+      this.page.getByRole("tab", { name: "Customise" }),
+    ).toBeVisible();
+  }
+
+  // Create a new node in a templated flow
+  async createNodeWithTemplateConfig(
+    title: string,
+    instructions: string,
+    required: boolean,
+    locator?: Locator,
+  ): Promise<void> {
+    await (locator ?? this.firstNode).click();
+    await this.page.getByRole("dialog").waitFor();
+
+    await this.page.getByPlaceholder("Text").fill(title);
+
+    // force: true is necessary for MuiSwitch inputs
+    await this.page.getByLabel("Allow edits").click({ force: true });
+    await this.page.locator("#templatedNodeInstructions").fill(instructions);
+    if (required) {
+      await this.page.getByLabel("Require edits").click({ force: true });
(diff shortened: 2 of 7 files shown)


e2e/tests/ui-driven/src/helpers/navigateAndPublish.ts
```diff
@@ -26,7 +33,9 @@ export const publishService = async (page: Page) => {
   await page.getByTestId("publish-button").click();
 
   // Wait for confirmation of success
-  await expect(page.getByText("Successfully published changes")).toBeVisible();
+  await expect(
+    page.getByText("Successfully published changes").first(),
```
joecarver, line 37: this was failing on my tests because the there were two 'Successfully publishes changes' messages being rendered, one in the toast and one below the publish button. Curious why it hasn't been an issue for other tests using this helper..?

e2e/tests/ui-driven/src/pages/Editor.ts
```diff
@@ -78,6 +78,68 @@ export class PlaywrightEditor {
...
+    await this.page.getByRole("option", { name: sourceTemplateName }).click();
+
+    // Save flow and redirect
+    await this.page.getByRole("button", { name: "Add flow " }).click();
+    await expect(
+      this.page.getByRole("tab", { name: "Customise" }),
+    ).toBeVisible();
+  }
+
+  // Create a new node in a templated flow
```
jessicamcinchak: nit: this should be "Create a new node in a source template"

e2e/tests/ui-driven/src/pages/Editor.ts
```diff
@@ -78,6 +78,68 @@ export class PlaywrightEditor {
...
+  async createNodeWithTemplateConfig(
+    title: string,
+    instructions: string,
+    required: boolean,
+    locator?: Locator,
+  ): Promise<void> {
+    await (locator ?? this.firstNode).click();
+    await this.page.getByRole("dialog").waitFor();
+
+    await this.page.getByPlaceholder("Text").fill(title);
```
jessicamcinchak: Which node/component _type_ is this creating? 

A "sticky note" Question by default if I'm understanding correctly? Can we more explicitly pass a component type here, ideally one that is actually seen/interacted with my users more often like an input or Question _with options_ ?

e2e/tests/ui-driven/src/templates.spec.ts
```diff
@@ -0,0 +1,419 @@
...
+      // Close modal without saving
+      await page.locator('button[aria-label="close"]').click();
+      await page.getByRole("button", { name: "Discard changes" }).click();
+      await page.getByRole("dialog").waitFor({ state: "detached" });
+
+      // check node is not visible
+      await expect(editor.nodeList).not.toContainText([REQUIRED_NODE_TITLE]);
+    });
+
+    test("can add required and optional nodes to a source template flow", async ({
```
jessicamcinchak, line 124: Can we expand this source template a bit to include: 
- [x] Required templated node 
- [x] Optional templated node
- [x] Non-templated node (then assert has no styled banner in source template and renders as read-only in templated flow)
- [x] Templated folder (to test fully editable in templated flow, including drag + drop)

e2e/tests/ui-driven/src/templates.spec.ts
```diff
@@ -0,0 +1,419 @@
...
+      const page = await getTeamPage({
+        browser,
+        userId: context.user!.id!,
+        teamName: context.team.name,
+      });
+      await navigateToService(page, SOURCE_TEMPLATE_SLUG);
+
+      await publishService(page);
+
+      await turnServiceOnline(page);
```
jessicamcinchak, line 276: nit: rather grouping publish + make it available in same test, I wonder if it'd be more meaningful to split like: 
- Publish source template only
- Go to "New flow" modal and confirm _not_ available in "From template" options yet
- Set source to "online"
- Repeat "New flow" modal and now selectable/visible

e2e/tests/ui-driven/src/templates.spec.ts
```diff
@@ -0,0 +1,419 @@
...
+      await page.locator('[aria-labelledby~="Templates-label"]').click();
+      await page.getByRole("option", { name: "Source template" }).click();
+
+      await expect(page.getByText(SOURCE_TEMPLATE_NAME)).toBeVisible();
+      await expect(page.getByText(REGULAR_FLOW_NAME)).not.toBeVisible();
+    });
+  });
+
+  // team editor creating a flow from source template, customising it and publishing it
+  // relies on the source template being online and published from the first block
```
jessicamcinchak, line 327: 👍 helpful reminder these are dependent tests

e2e/tests/ui-driven/src/templates.spec.ts
```diff
@@ -0,0 +1,419 @@
...
+      await navigateToService(page, TEMPLATED_FLOW_SLUG);
+
+      // The "Customise" tab is the default active tab for templated flows
+      const customiseTab = page.getByRole("tab", { name: "Customise" });
+      await expect(customiseTab).toBeVisible();
+      await expect(customiseTab).toHaveAttribute("aria-selected", "true");
+
+      const editor = new PlaywrightEditor(page);
+      await editor.checkNodeExists(REQUIRED_NODE_TITLE);
+      await editor.checkNodeExists(OPTIONAL_NODE_TITLE);
```
jessicamcinchak, line 363: nit: shall we also check that the "customise" cards are clickable - eg I can click on one and it opens a node modal with instructions visible at the top?

e2e/tests/ui-driven/src/templates.spec.ts
```diff
@@ -0,0 +1,419 @@
...
+        templatedCard.getByTestId("templated-flow-star"),
+      ).toBeVisible();
+
+      // The source template should not appear in the filtered results
+      await expect(
+        page.getByText(SOURCE_TEMPLATE_NAME, { exact: true }),
+      ).not.toBeVisible();
+    });
+
+    // TODO - test("Cannot add new nodes in a templated flow");
```
jessicamcinchak: This is an important one to flesh out please! 

Also a few other ideas here too of visual differences in a templated flow graph which we could check have rendered correctly: 
- Lack of add-new-node "hangers" (small dot button on branches)
- Disabled context menu (copy/clone/cut) on right click of a node (templated or non)

e2e/tests/ui-driven/src/templates.spec.ts
```diff
@@ -0,0 +1,419 @@
...
+      await expect(
+        requiredCustomisationTask.getByTestId("customisation-complete"),
+      ).toBeVisible();
+
+      await publishService(page);
+
+      await navigateToTeamPage(page);
+      await expect(page.getByText(TEMPLATED_FLOW_NAME)).toBeVisible();
+    });
+  });
```
jessicamcinchak, line 572: Publish tests are great, thanks !

e2e/tests/ui-driven/src/templates.spec.ts
```diff
@@ -0,0 +1,419 @@
...
+      await page.getByTestId("next-step-publish-button").click();
+
+      // The Publish step should show template-specific messaging and list affected flows
+      await expect(
+        page.getByRole("heading", { name: "Publish" }),
+      ).toBeVisible();
+      await expect(page.getByText("This flow is a template")).toBeVisible();
+      await expect(
+        page.getByText(new RegExp(TEMPLATED_FLOW_SLUG)),
+      ).toBeVisible();
```
jessicamcinchak, line 611: ➕

jessicamcinchak (commented):
Overall looking very good, thanks for both getting into the E2E framework + improving helpers and also learning Templates as a core product feature ! 

A few minor comments + questions throughout. Main ask is that you complete one outstanding "TODO" around editing templated flows before final approval please! 

Also noticed a few auto-comments around `{ force: true }` syntax that might be worth double-checking?

e2e/tests/ui-driven/src/pages/Editor.ts
joecarver: **done** https://github.com/theopensystemslab/planx-new/pull/6576/commits/5d83073b11184aef49a3ac55bed9d4d12eb1cbbb - i actually just moved the templated bits into the generic `createBaseComponent` function as that already has all the switching logic for different component types

e2e/tests/ui-driven/src/templates.spec.ts
joecarver: done https://github.com/theopensystemslab/planx-new/pull/6576/commits/101fe91ad12fadfd190c1d69bcd711419167d5ab

(sorry for combined commit with the prev comment around extending the source template)

e2e/tests/ui-driven/src/templates.spec.ts
joecarver: done https://github.com/theopensystemslab/planx-new/pull/6576/commits/101fe91ad12fadfd190c1d69bcd711419167d5ab

note i had to change the `getByPlaceholder` call for `InternalPortal` as it was looking for the wrong string. I don't think this had been covered by any e2e tests before from what I could tell.

(sorry for combined commit with the next comment around splitting tests up)

e2e/tests/ui-driven/src/templates.spec.ts
```diff
@@ -0,0 +1,658 @@
...
+        page.getByRole("menuitem", { name: "Copy" }),
+      ).toHaveAttribute("aria-disabled", "true");
+      await expect(
+        page.getByRole("menuitem", { name: "Clone" }),
+      ).toHaveAttribute("aria-disabled", "true");
+      await expect(page.getByRole("menuitem", { name: "Cut" })).toHaveAttribute(
+        "aria-disabled",
+        "true",
+      );
+    });
```
jessicamcinchak, line 501: These are meaningful and helpful, thank you !

jessicamcinchak (approved):
Changes all look good here, thank you !

Only final question: this is a significant number of new ui-driven tests, should any be tagged as `regression` only? 

It looks like e2e CI on your branch here is still completing in < 10 minutes so happy to keep all "active" on every run for now (which is also easier when dependent tests I think), but maybe one to just keep an eye on and adjust down the line if we notice longer CI times!

PR #6577, opened 2026-04-30
ci: e2e tests optimisations

joecarver:
I'm expecting this to be the same speed on the first run but then hopefully on subsequent runs, most (all?) layers will be cached in the GHA cache... lets see

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -52,6 +52,8 @@ jobs:
               - 'apps/api.planx.uk/**'
             e2e:
               - 'e2e/**'
+              - 'docker-compose.e2e.yml'
+              - 'scripts/start-containers-for-tests.sh'
             hasura:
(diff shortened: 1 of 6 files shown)


DafyddLlyr (commented):
This is a great shout - wasn't even aware this was something we could do to be perfectly honest.

Reading the docs the setup and syntax looks right, but re-running the action and following logs I can't see that it's working as expected here. There's no indication (I can see) of a cache hit, and the total time taken is pretty similar to other PRs.

Locally I've tried debugging the `docker buildx bake` command by adding the `--print` flag and nothing seems wrong - `tags`, `cache-from` and `cache-to` are populated as expected. 🤔

joecarver:
> I can't see that it's working as expected here

Yeah I noticed the same when I opened it last week. Am yet to revisit but will have a poke around today, might just be some GH config that needs changing to let us access the cache or something.

PR #6578, opened 2026-04-30
feat: combine pinned flows and unpinned into same list

joecarver:
### What's in this PR?
- addressing this UI suggestion by Silvia & Ian https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1777541377498229
- combines pinned and unpinned flows into the same list so that filters have the same effect on both sets, making it easier to manage pinned flows when using a filtered view

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -226,7 +226,7 @@ jobs:
         uses: actions/cache@v5
         with:
           path: ./${{ env.EDITOR_DIRECTORY }}/build
-          key: ${{ runner.os }}-${{ hashFiles(format('{0}/**', '!{0}/build/**', env.EDITOR_DIRECTORY)) }}
+          key: ${{ runner.os }}-${{ hashFiles(format('{0}/**', env.EDITOR_DIRECTORY)) }}
       - name: Download LPS build assets
         id: cache-localplanning-build-assets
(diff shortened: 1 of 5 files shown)


ianjon3s:
This is working brilliantly for card view, however list view still shows two tables.

I think combining into one table should be possible by setting up a spacer row element (with styles), then populate the table:
- map over pinned flows
- add spacer row (if pinned flows are present)
- map over unpinned flows

The spacer row seems a bit old school, but as we're using old school tables (which need a defined structure) it may be the simplest way to get it done.

joecarver:
> combining into one table should be possible

@ianjon3s good shout! changed [fc4941f](https://github.com/theopensystemslab/planx-new/pull/6578/commits/fc4941f2c7b7bbeb7c7705b0731b710b7b414090)

DafyddLlyr (approved):
Great..!

ianjon3s:
@joecarver this is working great, 2 small issues:

- The spacer cell is visible when there are no flows pinned
- Spacer cell has borders left and right (this might be a bit fiddly to fix, and may require some `:first-of-type` & `:last-of-type` targeting on rows, after removing the border from the wrapping `<table>` element)

image

DafyddLlyr (approved):

ianjon3s (approved):

PR #6575, opened 2026-04-30
test(editor): Optimise RTL setup for React 19

DafyddLlyr:
## What does this PR do?
After bumping to React 19 / MUI 9, React tests are timing out more frequently in CI. They usually pass on a re-run, and pass locally also. Stats show that these are now running at ~11m, up from ~8m30s historically - https://github.com/theopensystemslab/planx-new/actions/metrics/performance?dateRangeType=DATE_RANGE_TYPE_CUSTOM&tab=jobs&range=1777507200000-1777507200000

This PR attempts to improve to the performance of these tests by identifying issues (mainly turning off the `--silent` flag, as well as identifying where the timeouts are happening). Multiple re-runs on this PR have the React tests running in ~8m30m.

I think the underlying change here is concurrent rendering and stricter batching behaviour in React 19.

Most the changes should be grouped in commits, but here's the broad strokes - 

 - Move `/testUtils.ts` to be colcated with other test setup files in `/test/utils.ts` - noisy change sorry..!
 - Setup the `user` prior to render in our test setup util
 - Setup jsdom mocks to suppress warnings (makes it easier to spot actual issues)
 - Replace `user.type()` for long strings / frequently called helpers with `user.paste()` - one event rather than one per-key stroke.

## Next steps...
There's plenty more room for improvement here, but hopefully this gets us over the initial timeouts. I think from here we can work through tests in batches trying to resolve warnings, and focus on methods for improving long-running ones. I'll aim to pick up one suite a day...! 🛠️

PR #6579, opened 2026-05-01
test(editor): Fix warnings in MapAndLabel tests

DafyddLlyr:
Follow on from https://github.com/theopensystemslab/planx-new/pull/6575

The `MapAndLabel` tests will now run without any warning (minus jsdom ones resolved above).

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
@@ -18,7 +18,7 @@ import { point } from "@turf/helpers";
 import { Feature } from "geojson";
 import type { GeoJSONChangeEvent } from "lib/gis";
 import { Store, useStore } from "pages/FlowEditor/lib/store";
-import { useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 import { FONT_WEIGHT_SEMI_BOLD } from "theme";
 import FullWidthWrapper from "ui/public/FullWidthWrapper";
 import ErrorWrapper from "ui/shared/ErrorWrapper";
@@ -228,8 +228,18 @@ export default function Component(props: Props) {
    */
   const clipGeojsonData = (() => {
     if (boundary) return buffer(boundary, bufferInMeters, { units: "meters" });
-    if (addressPoint) return buffer(addressPoint, bufferInMeters, { units: "meters" });
+    if (addressPoint)
+      return buffer(addressPoint, bufferInMeters, { units: "meters" });
   })();
+
+  const mapRef = useCallback((node: HTMLElement | null) => {
+    if (node) {
+      (
+        node as HTMLElement & { ariaLabelOlFixedOverlay: string }
+      ).ariaLabelOlFixedOverlay =
+        "An interactive map for providing your location plan boundary";
+    }
+  }, []);
 
   function getBody(mapValidationError?: string, fileValidationError?: string) {
     if (page === "draw") {
@@ -272,8 +282,8 @@ export default function Component(props: Props) {
                 )}
                 {/* @ts-ignore */}
                 <my-map
+                  ref={mapRef}
                   id="draw-boundary-map"
-                  ariaLabelOlFixedOverlay="An interactive map for providing your location plan boundary"
                   drawMode
                   drawPointer="crosshair"
                   drawGeojsonData={boundary}
--- a/apps/editor.planx.uk/src/@planx/components/MapAndLabel/Public/CopyFeature.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/MapAndLabel/Public/CopyFeature.tsx
@@ -31,10 +31,18 @@ export const CopyFeature: React.FC<Props> = ({
 
   const formik = useFormik({
     initialValues: {
-      sourceLabel: sourceFeatures[0]?.properties?.label,
+      sourceLabel: sourceFeatures[0]?.properties?.label ?? "",
     },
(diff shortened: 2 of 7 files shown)


apps/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
```diff
@@ -272,8 +282,8 @@ export default function Component(props: Props) {
                 )}
                 {/* @ts-ignore */}
                 <my-map
+                  ref={mapRef}
                   id="draw-boundary-map"
-                  ariaLabelOlFixedOverlay="An interactive map for providing your location plan boundary"
```
DafyddLlyr, line 276: This triggers this warning - https://react.dev/warnings/invalid-aria-prop

There's a few more instances of this below.

```
stderr | src/@planx/components/FindProperty/Public/Public.test.tsx > render states > renders correctly when allowing non-UPRN addresses
Invalid ARIA attribute `ariaLabelOlFixedOverlay`. ARIA attributes follow the pattern aria-* and must be lowercase.
```

apps/editor.planx.uk/src/@planx/components/MapAndLabel/Public/CopyFeature.tsx
```diff
@@ -52,7 +60,7 @@ export const CopyFeature: React.FC<Props> = ({
             bordered
             required
             title={"Copy from"}
-            value={formik.values.sourceLabel}
+            value={sourceLabel}
```
DafyddLlyr, line 63: The fixes here address this issue - 

```sh
stderr | src/@planx/components/MapAndLabel/test/public.backNavigation.test.tsx > navigating back after adding two features > should maintain labelling when removing a feature
MUI: You have provided an out-of-range value `2` for the select (name="select-0") component.
Consider providing a value that matches one of the available options or ''.
The available values are "".
```

apps/editor.planx.uk/src/@planx/components/MapAndLabel/Public/index.tsx
```diff
@@ -271,11 +271,26 @@ const Root = () => {
...
   // If coming "back" or "changing", load initial features & tabs onto the map
   // Pre-populating form fields within tabs is handled via formik.initialValues in Context.tsx
-  if (previouslySubmittedData?.data?.[fn]?.features?.length > 0) {
-    addInitialFeaturesToMap(previouslySubmittedData?.data?.[fn]?.features);
-  }
+  useEffect(() => {
+    if (previouslySubmittedData?.data?.[fn]?.features?.length > 0) {
+      addInitialFeaturesToMap(previouslySubmittedData?.data?.[fn]?.features);
+    }
+  }, [addInitialFeaturesToMap, previouslySubmittedData, fn]);
```
DafyddLlyr, line 293:
```
stderr | src/@planx/components/MapAndLabel/test/public.test.tsx > back navigation > sets the latest feature as the active tab when coming back
Cannot update a component (`MapAndLabelProvider`) while rendering a different component (`Root`). To locate the bad setState() call inside `Root`, follow the stack trace as described in https://react.dev/link/setstate-in-render
```

apps/editor.planx.uk/src/@planx/components/MapAndLabel/test/utils.ts
```diff
@@ -18,7 +18,9 @@ export const addFeaturesToMap = async (
       "EPSG:3857": { features },
     },
   });
-  await act(() => map.dispatchEvent(mockEvent));
+  await act(async () => {
+    map.dispatchEvent(mockEvent);
+  });
```
DafyddLlyr, line 23:
```
stderr | src/@planx/components/MapAndLabel/test/public.test.tsx > validation and error handling > should show all fields are required, for all feature tabs
An update to ForwardRef(Tabs) inside a test was not wrapped in act(...).

When testing, code that causes React state updates should be wrapped into act(...):

act(() => {
  /* fire events that update state */
});
/* assert on the output */

This ensures that you're testing the behavior the user would see in the browser. Learn more at https://react.dev/link/wrap-tests-with-act
```

apps/editor.planx.uk/src/@planx/components/MapAndLabel/test/mocks/mockPayload.tsx
DafyddLlyr, line 1: I think this is just a bug / typo causing this error. As we're not correctly merging these objects, we end up with an `undefined` value.

```
stderr | src/@planx/components/MapAndLabel/test/public.backNavigation.test.tsx > navigating back after adding two features > should maintain labelling when removing a feature
A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://react.dev/link/controlled-components
```

apps/editor.planx.uk/src/@planx/components/DrawBoundary/Public/index.tsx
```diff
@@ -228,9 +228,19 @@ export default function Component(props: Props) {
...
   })();
 
+  const mapRef = useCallback((node: HTMLElement | null) => {
+    if (node) {
+      (
+        node as HTMLElement & { ariaLabelOlFixedOverlay: string }
+      ).ariaLabelOlFixedOverlay =
+        "An interactive map for providing your location plan boundary";
+    }
+  }, []);
```
jessicamcinchak, line 242: 👍

jessicamcinchak (approved):
Appreciate you working through these! Code looks good, happy for this to merge once CI clearled up ✔️

PR #6596, opened 2026-05-04
chore: add Gov Notify template IDs for new "general" templates

jessicamcinchak:
** **Not** hooked up to anything yet! I've simply created new templates within Gov Notify web console and recorded IDs and template config "types" here for hand-off of this ticket: https://trello.com/c/IOY0ZHEC/3121-add-a-second-govnotify-template-to-make-language-for-discretionary-services-less-application-planning-focused

In Gov Notify web, this now looks like screenshot below. Moving the _existing_ templates into a new "application" subfolder has _not_ changed their ID or URL structure!
Screenshot from 2026-05-03 20-36-09


apps/api.planx.uk/lib/notify/templates/saveAndReturn/expiry.ts
```diff
@@ -1,14 +1,23 @@
...
+ * GovNotify application template: https://www.notifications.service.gov.uk/services/012e65af-0eb2-45d5-87bd-4248354c4c22/templates/9619f89d-5d33-4cb0-a365-42c431ea9db3
  */
 export const expiryTemplate: NotifyTemplate<Config> = {
   id: "9619f89d-5d33-4cb0-a365-42c431ea9db3",
   access: "private",
   config: {} as Config,
 };
 
+/**
+ * GovNotify generic template: https://www.notifications.service.gov.uk/services/012e65af-0eb2-45d5-87bd-4248354c4c22/templates/b38532ea-9808-4367-b53d-3317d2389f8a
```
jessicamcinchak: FYI: I'm classifying templates as "application" (current default) or "general" (new option) - I don't _think_ the usual language of statutory/discretionary is very fitting here as things like "pre-apps" are discretionary "applications", so they'd still use the default "application" template.  

The "general" template is suitable for "reports [of planning breaches]", "calls [for sites]", "notifications [of commencement]", etc where councils do _not_ want to say "We have received your application."

apps/api.planx.uk/lib/notify/templates/saveAndReturn/expiry.ts
```diff
@@ -18,3 +27,7 @@ type Config = NotifyConfig<
     serviceLink: string;
   }
 >;
+
+type GenericConfig = Omit<Config, "personalisation"> & { 
+  personalisation: Omit<Config["personalisation"], "projectType">
+};
```
jessicamcinchak: I'm still assuming all "general" submissions have an "address" at minimum; "projectType" is the only property unique to "applications" at this time (final templates still to be confirmed with Alice / original drafts not currently accessible via Google)

apps/api.planx.uk/lib/notify/templates/saveAndReturn/expiry.ts
DafyddLlyr: This works for me..!

DafyddLlyr (approved):
Great!

Only small `nit:` would be that it might be worth continuing to mirror the GovNotify folder structure here (adding `/application` and `/general` subfolders).

jessicamcinchak:
**Updated per dev call decisions:**
- Now includes "general" templates for _all_ templates (save and return, invite to pay, and send to email)
- "General" template now shares exact same `Config` type (template-able variables) as "application" templates, rather than original assumption that "general" would be subset of variables (eg no project type). Both will have same and rely on same fallback "Not provided" values
- (Future PR) We will _not_ support separate help text configs (eg reply-to email) per template selection. Instead, in the future, we may change existing three help text template-able variables to a single string to allow councils to customise better

joecarver:
I may be missing something here but just taking a look at how this works in code and do we not need to register the templates in templates/index.ts `templateRegistry`?

edit - i can just do this on the main feature branch as i guess it comes more under the scope of acually wiring things up :~)

apps/api.planx.uk/lib/notify/templates/saveAndReturn/expiry.ts
jessicamcinchak: Update from dev call (5 May): both templates should actually keep the same types and we'll continue using fallback values like "Not provided" if not applicable! Above comment no longer applies.

jessicamcinchak:
@joecarver yes exactly ! 

This PR is really just aiming to do the first tedious bit of creating all 14 new templates in the Notify web dashboard and copying over UUIDs to the codebase; please carry forward "hooking up" however makes sense for you! This should be totally safe to merge as-is, new templates are added/exported, but not yet imported/registered anywhere :slightly_smiling_face:

PR #6605, opened 2026-05-05
chore: Bump planx-core

DafyddLlyr:
This will resolve a few security dependabot issues now that `planx-core` issues have been resolved.


zz-hh-aa (approved):
Thanks for working through dependabot updates!

PR #6606, opened 2026-05-05
test: maximum members per-team

zz-hh-aa:
Adds frontend and e2e tests for maximum team member limit, as per Daf's [comment on the feature](https://github.com/theopensystemslab/planx-new/pull/6555#pullrequestreview-4210750203)

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/mockMembers.ts
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/mockMembers.ts
@@ -1,0 +1,57 @@
+import { TeamMember } from "../types";
+
+const firstNames = [
+  "Alice",
+  "Bob",
+  "Charlie",
+  "Diana",
+  "Edward",
+  "Fiona",
+  "George",
+  "Hannah",
+  "Joe",
+  "Liz",
+];
+const lastNames = [
+  "Smith",
+  "Jones",
+  "Brown",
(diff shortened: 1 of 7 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/mockMembers.ts
```diff
@@ -0,0 +1,57 @@
...
+    email: "",
+    role: "teamEditor",
+    isPlatformAdmin: false,
+    isAnalyst: false,
+    defaultTeamId: 1,
+  }),
+);
+
+export const mockTeamMembersAt19 = mockTeamMembers.slice(0, 19);
+export const mockTeamMembersAt20 = mockTeamMembers.slice(0, 20);
```
DafyddLlyr, line 57: Neat 👍

DafyddLlyr (approved):
Great to see - thanks...! 

I think I'd be tempted to add an `@regression` tag here to the E2E test - this feels like a feature which we aim to have pretty simple, stable, and untouched going forwards.

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/mockMembers.ts
zz-hh-aa: Lazy like a fox... 🦊

PR #6608, opened 2026-05-05
feat: set up teamAdmin inherited from teamEditor role

zz-hh-aa:
Sets up a teamAdmin role that inherits from teamEditor. 

This is the Hasura step in the ["Create teamAdmin" role ticket](https://trello.com/c/j63eNFql). It specifies that this should configure basic row-level access, but leave specific user management permissions on another ticket, so this is very bare bones.

When we want to override / give teamAdmin more permissions, we can do this:
>...by creating a separate specific permission. After creation, it will override the inherited permission, if any.

(see [Hasura docs for more info](https://hasura.io/docs/2.0/auth/authorization/inherited-roles/))

Ended up reading about this last Friday so went ahead and did this first step, just a heads up to @joecarver !

### DIFF
--- a/apps/hasura.planx.uk/metadata/inherited_roles.yaml
+++ b/apps/hasura.planx.uk/metadata/inherited_roles.yaml
@@ -1,1 +1,3 @@
-[]
+- role_name: teamAdmin
+  role_set:
+    - teamEditor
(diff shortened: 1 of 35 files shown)


DafyddLlyr (approved):
We'll need to remember introspection tests here to describe the new role - these should be quick / easy. Worth picking up just now I'd say..!

PR #6625, opened 2026-05-06
chore: renew gloucester ssl

zz-hh-aa:
Hopefully the last one using the current / soon-to-be old method...

- Updated certs tested on AWS CM staging (useful--caught a mistaken BOPS cert!)
- Secrets added to AWS SM production
- Secrets removed from Pulumi config

~~⚠️ As per usual, I manually entered the secret into AWS prod, taking care to preserve line breaks. (Never managed to get the script to run locally, tried it again today just in case). I noticed this error when switching back to key/value view after checking line breaks, though the plaintext looks right to me. Is this an issue?~~

Formatting sorted on AWS; script still unsuccessful despite numerous config retries.


DafyddLlyr (changes requested):
@zz-hh-aa I think this is an issue - we rely on the secrets being key:value in the codebase. I'm not quite sure how a plaintext object would get handled here, but I'd guess it would not work (and it's not a risk we probably want to take here if it could result in downtime for the subdomain).

Did you enter the secret as KV?

We need to fix the script so this is as pain-free as possible - what issues did you hit?

zz-hh-aa:
>Did you enter the secret as KV?

Yes but after I edited in plaintext to ensure that line breaks were preserved, I could no longer view as key / value. I've just tried re-entering them again but I can't insert line breaks in the key / value view (and I know the docs specify to preserve them + I can see that other secrets display with line breaks). 

> We need to fix the script so this is as pain-free as possible - what issues did you hit?

I keep getting this, even after I have tried reconfiguring AWS with `aws configure`, manually inputting the values copied from "Get credentials for ssl-dev-tasks", as well as  trying `aws configure sso`. 

```
An error occurred (ExpiredTokenException) when calling the PutSecretValue operation: The security token included in the request is expired
Secret not found, creating it...

An error occurred (ExpiredTokenException) when calling the CreateSecret operation: The security token included in the request is expired
```

Which is the same as I've encountered before.

zz-hh-aa:
Reformatted the cert and key again and have succeeded in viewing as KV in AWS! I think this is now ready to go (please feel free to check in AWS). 

Still didn't have any luck troubleshooting my local `aws configure` both with `sso` and manual.

DafyddLlyr (approved):
Looks good - thanks for sorting..!

The new SSL setup Dan is working on will eliminate this manual step (which is good news!) but I'd still like to get to the bottom of this one, lets pair on this if there's another cert renewal.

Once merged, please open a PR to `production` 🚀

PR #6626, opened 2026-05-06
fix(editor): Improve test coverage for `FeedbackLog`

DafyddLlyr:
This PR increases test coverage for the `FeedbackLog` component. We're now testing the toolbar so this suite would have failed an caught the regression introduced on the MUI data grid update.


jessicamcinchak (approved):
As a quite recent-memory feature, I hadn't even realised we were lacking test coverage around this - appreciate you picking up + filling in !

DafyddLlyr:
Yeah I think the truth is there's a number of these Editor facing features which ran over during development and we then neglected to spent time on tests. 

_Relatively_ simple and stable features, for a small user-base, and time constraints I think were the killers here - something we probably need to be more aware of going forward I think.

PR #6627, opened 2026-05-06
fix(editor): Restore toolbar to all `<DataTable/>` instances

DafyddLlyr:
## What's the problem?
The toolbar is missing from all `<DataTable/>` instances - feedback, admin panel, submissions.

## What caused this?
When upgrading `@mui/x-data-grid` from v7 to v9 we missed this change from the the migration guide - 

> _The `showToolbar` prop is now required to display the toolbar._

Source: https://mui.com/x/migration/migration-data-grid-v7/#changes-to-the-public-api

## What's the solution?
Add the prop, update and fix style and behavioural issues.

## Also done this PR...
- [x] Run the codemod and all migration guide steps for v8 (this would have auto-fixed this)
- [x] Run the codemod and all migration guide steps for v9
- [x] Update `PlatformAdminPanel` and associated components (see code comments)

This highlights that we really need to be more conscious of test coverage for these Editor facing features - this is a change which has broken several Editor facing features, and remained undetected for a few days. I've started this here - https://github.com/theopensystemslab/planx-new/pull/6626

### DIFF
--- a/apps/editor.planx.uk/src/ui/shared/DataTable/DataTable.tsx
+++ b/apps/editor.planx.uk/src/ui/shared/DataTable/DataTable.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable no-restricted-imports */
 import Box from "@mui/material/Box";
+import type { GridToolbarProps, ToolbarPropsOverrides } from "@mui/x-data-grid";
 import {
   DataGrid,
   GridColDef,
@@ -25,6 +26,32 @@ import {
   getValueOptions,
 } from "./utils";
 
+declare module "@mui/x-data-grid" {
+  interface ToolbarPropsOverrides {
+    customTools?: React.FC[];
+    csvExportFileName?: string;
+  }
+}
+
+const CustomToolbar = (props: GridToolbarProps & ToolbarPropsOverrides) => {
+  const { customTools, csvExportFileName } = props;
+
+  return (
+    <GridToolbarContainer>
+      {customTools &&
+        customTools.map((CustomTool, index) => <CustomTool key={index} />)}
+      <GridToolbarColumnsButton />
+      <GridToolbarFilterButton />
+      <GridToolbarDensitySelector />
+      <GridToolbarExport
+        csvOptions={{
+          fileName: csvExportFileName,
+        }}
+      />
+    </GridToolbarContainer>
+  );
+};
+
 export const DataTable = <T,>({
   rows,
   columns,
@@ -44,23 +71,6 @@ export const DataTable = <T,>({
     }
     const ComponentRenderer = columnCellComponentRegistry[column.type];
     return ComponentRenderer(params.value, filterValues);
-  };
-
-  const CustomToolbar = () => {
-    return (
-      <GridToolbarContainer>
-        {customTools &&
-          customTools.map((CustomTool, index) => <CustomTool key={index} />)}
-        <GridToolbarColumnsButton />
-        <GridToolbarFilterButton />
-        <GridToolbarDensitySelector />
-        <GridToolbarExport
-          csvOptions={{
-            fileName: csvExportFileName,
-          }}
-        />
-      </GridToolbarContainer>
-    );
   };
 
   const [filterValues, setFilterValues] = useState<string[]>([]);
@@ -129,6 +139,12 @@ export const DataTable = <T,>({
           }
           slots={{
             toolbar: CustomToolbar,
+          }}
+          slotProps={{
+            toolbar: {
+              customTools,
+              csvExportFileName,
+            },
           }}
           showToolbar
           getRowId={(row) => row.id}
--- a/apps/editor.planx.uk/src/ui/shared/DataTable/DataTable.tsx
+++ b/apps/editor.planx.uk/src/ui/shared/DataTable/DataTable.tsx
@@ -130,6 +130,7 @@ export const DataTable = <T,>({
           slots={{
             toolbar: CustomToolbar,
           }}
+          showToolbar
           getRowId={(row) => row.id}
           processRowUpdate={onProcessRowUpdate}
           checkboxSelection={checkboxSelection}
--- a/apps/editor.planx.uk/src/ui/shared/DataTable/components/MultipleOptionSelectFilter.tsx
+++ b/apps/editor.planx.uk/src/ui/shared/DataTable/components/MultipleOptionSelectFilter.tsx
@@ -1,7 +1,7 @@
 import FormControl from "@mui/material/FormControl";
+import { outlinedInputClasses } from "@mui/material/OutlinedInput";
 import TextField from "@mui/material/TextField";
-import { GridFilterFormProps } from "@mui/x-data-grid";
-import React from "react";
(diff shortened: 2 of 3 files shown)


DafyddLlyr:
Converted to draft - it's not quite as simple as setting this prop, there's a few styles issues to deal with as a result of the upgrade also...!

image

I'll re-open for review once I've got a more thorough solution.

DafyddLlyr:
The above style issues are now resolved, PR back open for review.

apps/editor.planx.uk/src/ui/shared/DataTable/components/MultipleOptionSelectFilter.tsx
```diff
@@ -10,54 +10,56 @@ import {
...
   OptionalAutocompleteProps<T> &
-  GridFilterFormProps & { applyValue: (args: Record<string, any>) => void };
+  GridFilterInputValueProps;
 
 export function MultipleOptionSelectFilter<T>(props: Props<T>) {
-  const { item, applyValue } = props;
+  const { item, applyValue, options, focusElementRef, ...otherProps } = props;
 
-  const [chipData, setChipData] = React.useState(item.value);
+  const currentValue = Array.isArray(item.value) ? item.value : [];
```
DafyddLlyr, line 18: Rather than track state manually via `chipData`, we can just use the `value` directly - adding and deleting are automatically handled here.

apps/editor.planx.uk/src/ui/shared/DataTable/components/MultipleOptionSelectFilter.tsx
```diff
@@ -10,54 +10,56 @@ import {
...
         disableCloseOnSelect
         multiple
+        disableClearable
+        value={currentValue}
+        sx={{
+          backgroundColor: "transparent",
+          [`&. ${outlinedInputClasses.root}`]: {
+            backgroundColor: "transparent",
+          },
+        }}
```
DafyddLlyr, line 36: This was (I think) an existing style issue - the input here has a white background on staging / prod unlike all other inputs.

apps/editor.planx.uk/src/ui/shared/DataTable/components/MultipleOptionSelectFilter.tsx
```diff
@@ -10,54 +10,56 @@ import {
...
             slotProps={{
               ...params.slotProps,
               inputLabel: { ...params.slotProps.inputLabel, shrink: true },
             }}
-            label="Option"
-            variant="standard"
+            label="Value"
+            variant="outlined"
+            inputRef={focusElementRef}
+            size="small"
```
DafyddLlyr, line 47: Updates to match other input styles.

apps/editor.planx.uk/src/ui/shared/DataTable/components/MultipleOptionSelectFilter.tsx
```diff
@@ -10,54 +10,56 @@ import {
...
-              );
-              setChipData((prev: string[]) => {
-                prev.splice(chipIndex, 1);
-              });
-              applyValue({ ...item, chipData });
-              setChipData(item.value);
+          },
+          popper: {
+            sx: {
+              width: "max-content !important",
```
DafyddLlyr, line 57: @ianjon3s I couldn't achieve this without `!important` - very open to other solutions here...!

This style ensures that the menu shows the full options and doesn't overflow per-line.

**Without `max-content`**

Image

**With `max-content`**

Image

apps/editor.planx.uk/src/ui/shared/DataTable/DataTable.tsx
```diff
@@ -25,6 +26,32 @@ import {
...
 } from "./utils";
 
+declare module "@mui/x-data-grid" {
+  interface ToolbarPropsOverrides {
+    customTools?: React.FC[];
+    csvExportFileName?: string;
+  }
+}
+
+const CustomToolbar = (props: GridToolbarProps & ToolbarPropsOverrides) => {
```
DafyddLlyr, line 36: Moving this outside the `DataTable` means we don't re-render on each change - without this there's a re-render flash and the toolbar popover jumps to position `0,0` (top left) on the screen each time there's a change of filter.

The tradeoff is slightly more complex props in order to pass in our `csvExportFileName` and `customTools`

apps/editor.planx.uk/src/ui/shared/DataTable/DataTable.tsx
```diff
@@ -25,6 +26,32 @@ import {
...
+        customTools.map((CustomTool, index) => <CustomTool key={index} />)}
+      <GridToolbarColumnsButton />
+      <GridToolbarFilterButton />
+      <GridToolbarDensitySelector />
+      <GridToolbarExport
+        csvOptions={{
+          fileName: csvExportFileName,
+        }}
+      />
+    </GridToolbarContainer>
```
DafyddLlyr, line 51: These are deprecated APIs which will be removed in a future MUI version.

Image

Some can already be migrated, but others are currently incomplete (see warning on docs - https://mui.com/x/react-data-grid/components/filter-panel/).

I'd vote we - 
 - leave this be for now, tackle in one go on next upgrade
 - at the same time, re-assess MUI DataGrid - we know multiple filters are a limitation today, might be other solid options here in future

apps/editor.planx.uk/src/ui/shared/DataTable/DataTable.tsx
```diff
@@ -130,6 +140,13 @@ export const DataTable = <T,>({
           slots={{
             toolbar: CustomToolbar,
           }}
+          slotProps={{
+            toolbar: {
+              customTools,
+              csvExportFileName,
+            },
+          }}
+          showToolbar
```
DafyddLlyr, line 149: This single new prop is the fundamental change - this is now no longer `true` by default.

apps/editor.planx.uk/src/ui/shared/DataTable/components/MultipleOptionSelectFilter.tsx
ianjon3s: Agree, trying to match the list width to the "parent" input just doesn't work here, sometimes a blunt force override is necessary for MUI components 👍

ianjon3s (approved):
Working great visually and functionally. I do however recall the previous toolbar controls being placed on the left. If it's a simple update to shift it back to the left then worth doing, if it's more complex then I'm sure we can live with on the right.

DafyddLlyr:
> _I do however recall the previous toolbar controls being placed on the left. _

@ianjon3s Yeah that's correct - it was previously on the left, but the new default it right aligned. I actually think this works pretty well - especially now we have our own menus on the left. It feels a little less cramped and dense?

No strong feelings here though - very happy to follow your judgement..!

I've merged to `main` so that we can get a fix across to `production` asap, but we can tag along an alignment solution as a quick follow up - just let me know 👍

jessicamcinchak (commented):
Late to this one, but thanks for working through these changes - helpful comments as always throughout 👍

ianjon3s:
Also no strong opinions, let's go with right!

PR #6630, opened 2026-05-07
 feat: GovNotify general template - backend

joecarver:
### What's in this PR
DB Migrations, Hasura Metadata, backend changes, and tests for the 'general' worded templates

### Why?
Currently the emails we send always have 'application' centered language, whereas not all flows concern an application. Some are more general submissions, so we want different lanugage in the email.
https://trello.com/c/IOY0ZHEC/3121-add-a-second-govnotify-template-to-make-language-for-discretionary-services-less-application-planning-focused

### Notes
- New templates previously written by Jess and added to GovNotify, and added to `lib/notify/templates`
- DB Migrations previously written by Jess in https://github.com/theopensystemslab/planx-new/pull/6597 - have cherry picked that commit into this branch
-  Requires that new 'general' templates are added to `GENERAL_TEMPLATE_MAP` in notify/index.ts
    - I did consider just prepending `general` to the template name but the explicit map is a bit more robust with the types, and in either case we have to handle the default case of 'no corresponding general template' and its a bit cleaner this way


apps/api.planx.uk/lib/notify/index.ts
```diff
@@ -28,7 +53,7 @@ const sendEmail: SendEmail = async (template, emailAddress, config) => {
       expiryDate?: string;
     } = { message: "Success" };
 
-    if (template === "save") {
+    if (template === "save" || template === "general-save") {
```
jessicamcinchak, line 56: nit: currently, `save` and `general-save` templates have the same config type but in a future where they might differ - could this be more flexible and dynamically set like
```
const saveConfig = config as TemplateRegistry[template]["config"];
```  
?? I suppose small risk here of different configs differing on returned expiry date field here, so that suggestion alone above may not be type safe _enough_ !

apps/api.planx.uk/modules/pay/service/inviteToPay/sendConfirmationEmail.ts
```diff
@@ -40,6 +52,7 @@ type PayeeAndAgentEmailData = {
   payeeEmail: string;
   projectTypes: string[];
   emailReplyToId: string;
+  emailTemplate: "application" | "general";
```
jessicamcinchak: nit: can we reuse/import the `GovNotifyEmailTemplate` type defined above here?

apps/api.planx.uk/modules/saveAndReturn/service/resumeApplication.ts
```diff
@@ -25,7 +25,14 @@ const resumeApplication = async (teamSlug: string, email: string) => {
     reference: null,
     emailReplyToId: team.settings.emailReplyToId,
   };
-  const response = await sendEmail("resume", email, config);
+  const hasGeneralSession = sessions.some(
+    (s) => s.flow.email_template === "general",
+  );
```
jessicamcinchak, line 30: Expected logic if many sessions / in-progress services 👍

apps/api.planx.uk/modules/send/email/newLink/service.ts
```diff
@@ -25,7 +28,12 @@ export const getSession = async (sessionId: string) => {
   const { session } = await $api.client.request<{
     session: {
       submittedAt: string;
-      flow: { id: string; name: string; team: { id: number } };
+      flow: {
+        id: string;
+        name: string;
+        email_template: "application" | "general";
```
jessicamcinchak: nit: same as above / can we import existing type here?  An at least two more instances below too!

jessicamcinchak (changes requested):
Looks great, thanks again for taking this one over ! 

- ✅ I tested an existing submission service via pizza editor and am receiving "application" emails by default as expected
- 👀 I'd also like to test via the API docs and manually trigger "general" templates to myself for an existing session ID; but the Swagger picker is now out of sync - can you update this please too? https://api.6630.planx.pizza/docs/#/send%20email
- 👍 A few other nit picky comments below, no show stoppers!

Once Swagger updated, happy to test again & approve !

apps/api.planx.uk/lib/notify/index.ts
joecarver: unfortunately we can't pass template in as the type index, as its a runtime variable and you can only pass types in as type arguments. `typeof template` also won't do what we want here as it returns `T extends Template` i.e. the full set of templates, not narrowed down by the if statement.

we can however, just use the union of `"save" | "general-save"` as the index instead, which gives us the compile time checks in the case of the two configs diverging. have changed that now https://github.com/theopensystemslab/planx-new/pull/6630/commits/2c7b325ceb122655c47d0af82d42942facff933f

jessicamcinchak (approved):
All working as expected for me here, thanks for changes!

PR #6633, opened 2026-05-08
fix: drop `demoUser` option from `user_roles` enum

jessicamcinchak:
Noticed this while manually adding a team membership to the database per latest #planx-bugs question - I surprisingly still had the option to select `demoUser` even though we've removed this via the UI. 

It was due to a lingering `user_roles` enum value missed as part of #6419, now corrected!

Screenshot from 2026-05-07 21-00-36


DafyddLlyr (approved):
Good spot - thanks!

PR #6635, opened 2026-05-08
feat: GovNotify general template - frontend

joecarver:
### What's in this PR?
Frontend changes for the 'general' email template settings

Associated backend PR: https://github.com/theopensystemslab/planx-new/pull/6630

### Notes
- Added a separate page for this setting
- Mutations/Schema etc in separate files
- Does not include the team settings contact changes - coming in a separate PR


jessicamcinchak:
Couple quick requests ahead of properly reviewing this one: 
- Looks like data sync failed on pizza which means not possible to login / can confirm this via empty `users` table: https://hasura.6635.planx.pizza/console/data/default/schema/public/tables/users/browse 
  - To fix, you can login to Vultr, destroy service instance + 2x DNS records then should re-create successfully on next CI run. Shout if any issues doing this if your first time! 
- Could you please rebase to `main` to omit API changes from "changed files" count? 

Thanks !

joecarver:
> Couple quick requests

@jessicamcinchak done! Pizza should be ready soon 🧑‍🍳

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Flow/Notifications/components/EmailTemplateForm.tsx
```diff
@@ -0,0 +1,82 @@
...
+    <SettingsFormContainer<
+      GetFlowEmailTemplate,
+      UpdateFlowEmailTemplate,
+      FormValues
+    >
+      query={GET_FLOW_EMAIL_TEMPLATE}
+      mutation={UPDATE_FLOW_EMAIL_TEMPLATE}
+      validationSchema={validationSchema}
+      legend="Email template"
+      description="Choose the email template category for notifications sent to users of this service. Use 'Application' for statutory planning services and 'General' for discretionary services."
```
jessicamcinchak: nit: can we update phrasing to something like: 
```
Choose an email template for communicating with users of this service. Use 'Application' for statutory or apply-for services and 'General' for discretionary services.
```

This addresses two minor things: 
- Avoids "notification" language around this feature
- Pre-applications are a common but distinct edge case in that they are technically "discretionary" but also apply-for/application language!

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Flow/Layout.tsx
```diff
@@ -40,6 +41,7 @@ const FlowSettingsLayout: React.FC<Props> = ({ children }) => {
       icon: StarIcon,
       condition: Boolean(data?.flow.templatedFrom),
     },
+    { label: "Notifications", path: "/notifications", icon: EmailIcon },
```
jessicamcinchak: nit: I think we need to avoid the language "Notifications" here as this is going to be introduced soon to mean in-editor tasks + alerts. Perhaps "Emails" or "Communication" are better options?

jessicamcinchak (commented):
Thanks for rebase here! 

Please note still unable to login to pizza - I assume data sync failed on same issue identified here this morning? https://github.com/theopensystemslab/planx-new/pull/6652

One more Vultr destroy + rebase should do the trick? (PR data syncs only run on _create_ step, not upsert, in order to preserve session data etc on feature branches).

A couple small language nits in the meantime!

joecarver:
> still unable to login to pizza

@jessicamcinchak sorry forgot to check it had worked before! just signed in now after a fresh rebase & rebuild so the sync is all fixed it seems

ianjon3s:
One small styling suggestion: there is a `compact` variant of radio buttons that we use for non public facing forms

jessicamcinchak (approved):
Functionally all working as expected for me! Thanks Ian for small style spot here too!

joecarver:
> there is a `compact` variant of radio buttons

thanks @ianjon3s - there wasn't actually a `compact` variant for `DescriptionRadio` yet so I've added it to mirror the `BasicRadio` - let me know what you think of the padding values I've gone with. 

Also my first time updating anything in the component library, is there anything else I need to do surrounding the actual component change?

ianjon3s:
Thanks @joecarver , I had assumed there was a compact variant for description radios as used in the editor modal of the Section component, but looking at the code this was manually styled with descriptions (one for us to retrofit with your new component variant!)

The main concern when working with the component library is a given components usage across the platform. We've spilt the directories into `/editor` `/shared` and `/public` to indicate their usage, with `/shared` being the one where we reuse mostly public-facing components as UI in the editor. 

As you've added an optional variant that isn't referenced in any other usage of the UI component (the main one here being the public facing Question component), it's less of a concern, but when doing more fundamental changes to shared components it's always worth checking in a public flow.

An easy way of doing this is to find a flow that you know has that component in (which should already exist on both staging and your pizza for comparison), but you can also quickly create the component in a testing flow:
https://6635.planx.pizza/testing/radio-testing/preview
and compare against the same on staging:
https://editor.planx.dev/testing/radio-testing/preview

PR #6636, opened 2026-05-08
style: Flow view toggle position

ianjon3s:
## What does this PR do?

Quick one:

- Moves position of flow list toggle to prevent earlier wrapping of filters and inconsistent position:

**Current issue:**
image

**Updated:**
image


DafyddLlyr (approved):
Solid change 👍

PR #6637, opened 2026-05-08
fix: Hook up search for archived flows

ianjon3s:
## What does this PR do?

- Hooks up search to work when viewing the flow archive (currently only works with "flows" view)
- Maintains search term when switching between flows and archive

### DIFF
--- a/apps/editor.planx.uk/src/pages/Team/index.tsx
+++ b/apps/editor.planx.uk/src/pages/Team/index.tsx
@@ -222,7 +222,7 @@ const Team: React.FC<TeamProps> = (initialFlows) => {
             teamId={teamId}
             slug={slug}
             archivedFlows={displayedArchivedFlows}
-            loading={archivedFlowsLoading}
+            loading={archivedFlowsLoading && !displayedArchivedFlows}
             error={archivedFlowsError}
             isFiltered={archiveIsFiltered}
             onClearSearch={() => {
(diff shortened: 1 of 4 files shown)


apps/editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -202,6 +221,14 @@ const Team: React.FC<TeamProps> = (initialFlows) => {
             handleViewChange={handleViewChange}
             teamId={teamId}
             slug={slug}
+            archivedFlows={displayedArchivedFlows}
+            loading={archivedFlowsLoading}
```
DafyddLlyr: This is an issue not directly introduced here, but this should be as follows I think.

```suggestion
            loading={loading={archivedFlowsLoading && !displayedArchivedFlows}}
```

As we're using a `cache-and-network` strategy, `loading` will be true when we're re-running the request. Currently, we're showing a loading spinner unnecessarily on each load of the tab despite having data in our local cache.

After making this change we'll need to look at how the "0 archived flows" is showing as well, as I suspect this will be affected by the change.

DafyddLlyr (approved):
Great..! One small comment below at loading states to take a look at.

Sometime (soon!) I think we'll want to revisit this whole loading / filtering strategy as it's getting pretty complex here for a _relatively_ simple feature.

apps/editor.planx.uk/src/pages/Team/index.tsx
ianjon3s: Thanks for this @DafyddLlyr , I've made the change. "0 archived flows" appears to be outputting correctly.

DafyddLlyr:
Merged as we're queuing up a prod deploy 👍 Thanks for making those changes!

PR #6638, opened 2026-05-08
fix: publish requests should set body rather than query params

jessicamcinchak:
See #planx-bug report https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1778240234996769

Our old approach of passing `templatedFlowIds` as query params triggered qs' default `arrayLimit` length of 20 (FOI source template now officially has ~24 children) and was automatically converted from an array into an object, therefore triggering zod validation errors. 

**To test:** 
- Try to publish FOI source template on this pizza - it should succeed & response will include all scheduled event IDs as expected/usual https://6638.planx.pizza/app/opensystemslab/find-out-if-you-need-planning-permission
- Try to publish FOI source template on staging - it will fail !


apps/api.planx.uk/server.ts
```diff
@@ -33,14 +35,16 @@ import userRoutes from "./modules/user/routes.js";
...
-import { defaultCors } from "./cors.js";
 
 const app = express();
 
 useSwaggerDocs(app);
 
 app.set("trust proxy", 1);
 
+// Set arrayLimit to 100 (default is 20, which is too low for `templatedFlowIds`)
+app.use(express.query({ arrayLimit: 100 }));
```
jessicamcinchak: note that we're still on express v4 which uses `qs` as its' defualt query parser. Express v5 moves to `node:querystring` as the default which seems to be much more liked 😅 

Ref `https://github.com/expressjs/express/discussions/5783`

apps/api.planx.uk/server.ts
```diff
@@ -33,7 +35,6 @@ import userRoutes from "./modules/user/routes.js";
 import webhookRoutes from "./modules/webhooks/routes.js";
 import { apiLimiter } from "./rateLimit.js";
 import { registerSessionStubs } from "./session.js";
-import { defaultCors } from "./cors.js";
```
jessicamcinchak, line 36: These are simply re-ordered / alphabetised imports! Sorry for extra noise

apps/api.planx.uk/server.ts
jessicamcinchak: These comments now **outdated** ! My original solution here was to attempt to simply set qs' `arrayLimit` override on our express app.

But the line above, nor other attempted syntax variations like snippet below actually worked as expected here ! Hence prompting move to swapping out query params all together (more future-proof anyways!)
```
// Set arrayLimit to 100 (default is 20, which is too low for `templatedFlowIds`)
app.set("query parser", (str: string) => {
  return require("qs").parse(str, { arrayLimit: 100 });
});
```

apps/api.planx.uk/pnpm-lock.yaml
joecarver, line 1: how come the lockfile has changed although none of the dependencies have changed? just looks like some minor /patch bumps from what i can see but not sure if its intentional?

joecarver (approved):
Interesting bug, remember experiencing this before with Express. Definitely the right fix and makes more sense to use the request body for POSTs anyway!

apps/api.planx.uk/server.ts
joecarver: there's a browser limit on total query string lengths, so it's definitely correct to switch out and use POST here instead. Not sure if we have any other endpoints which might be affected now or in the future? Having a quick look now

DafyddLlyr (approved):
Thanks for picking this one up 👍

apps/api.planx.uk/server.ts
jessicamcinchak: Yes - browser limit is based on characters (~250 for most based on what I read?) and this sample request was still about half of that, so it was still the `qs` limit being hit at this stage - but definitely not future-proof fix if that route!!

I also checked other endpoints and don't think there are any other concerns :heavy_check_mark:

apps/api.planx.uk/pnpm-lock.yaml
jessicamcinchak: Lockfile has been regenerated as in earlier commits I was sorting which version of `qs` was a pre-existing peer dependent of our Express version. All still works as expected here!

PR #6640, opened 2026-05-08
feat: conditionally render settings for flows vs services

zz-hh-aa:
# What does this PR do?
- Adds a boolean `flows.is_service` column, allows select and update permissions to `platformAdmin` and `teamEditor`
- Conditionally renders most of the settings depending on whether a flow `isService` or not
- New `IsService` component to make the change (which is one way with a warning modal, see discussion with [August about this](https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1778245685599309?thread_ts=1778067730.871219&cid=C04DZ1NBUMR). We don't anticipate this will change back and forth, can handle edge cases, and this also seemed useful to distinguish this functionality from online / offline status)

https://github.com/user-attachments/assets/d0c71917-6b48-4dd6-9db6-4460d95f1f6f

# Testing
- Open a flow that is currently offline
- It should be a flow
- Change to a service
- Other settings should now be visible and change cannot be reverted

# Next steps
- [X] Tests
- [ ] Make sure `is_service` propagates from source template, if templated
- [ ] When this merges, we'll have to update sync scripts! 
- [x] Tidy up where we use the term 'service' vs 'flow' elsewhere
Screenshot 2026-05-08 173905

# Questions
- Should this be behind a feature flag or okay to merge as is?

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Flow/Visibility/IsService/index.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Flow/Visibility/IsService/index.tsx
@@ -17,7 +17,7 @@ import { validationSchema } from "./schema";
 import type {
   GetIsServiceResponse,
   IsServiceFormValues,
-  UpdateIsServiceVars,
+  IsServiceVars,
 } from "./types";
 
 const IsService: React.FC = () => {
@@ -30,7 +30,7 @@ const IsService: React.FC = () => {
   return (
     <SettingsFormContainer<
       GetIsServiceResponse,
-      UpdateIsServiceVars,
+      IsServiceVars,
       IsServiceFormValues
     >
       query={GET_IS_SERVICE}
@@ -38,7 +38,7 @@ const IsService: React.FC = () => {
       validationSchema={validationSchema}
       legend={"Is this flow a service?"}
       description={
-        "A service is user-facing and includes more customisable settings. Only services can accept responses."
+        "A service is user-facing and can accept responses when turned online. Only services can accept responses."
       }
       getInitialValues={({ flow: { isService } }) => ({ isService })}
       queryVariables={{ flowId }}
@@ -66,17 +66,22 @@ const IsService: React.FC = () => {
                 variant="body1"
                 sx={{ fontWeight: FONT_WEIGHT_BOLD, mr: 1, mb: 2 }}
               >
-                This is currently a {formik.values.isService === true ? "service" : "flow"}
+                {formik.values.isService === true ? "This is a service" : "This is currently a flow"}
               </Typography>
-              <Typography
-                variant="body1"
-              >
-                {formik.values.isService === true ? 
-                "A service is user-facing and can accept submissions" : 
-                "A flow is content that will be nested in other services, and cannot directly accept submissions"}
-                .
-              </Typography>
+              {formik.values.isService === true ? 
+                <Typography
+                  variant="body2"
+                >
+                  If this flow has been made a service in error, please contact an admin to revert it to a flow.
+                </Typography> :
+                <Typography
+                  variant="body1"
+                  >
+                  A flow is content that will be nested in other services, and cannot directly accept responses.
+                </Typography>
+              }
             </Box>
+              {formik.values.isService !== true &&
               <Box sx={{ display: "flex" }}>
                 <Button
                   id="set-is-service-button"
@@ -86,11 +91,9 @@ const IsService: React.FC = () => {
                   variant="contained"
                   onClick={() => setDialogOpen(true)}
                 >
-                  {formik.values.isService === false
-                    ? "Change to user-facing service"
-                    : "Change to flow"}
+                  {"Make this a user-facing service"}
                 </Button>
-              </Box>
+              </Box>}
               </>
             )}
             {isProduction && (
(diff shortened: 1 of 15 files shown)


apps/hasura.planx.uk/migrations/default/1778154943749_alter_table_public_flows_add_column_is_service/up.sql
```diff
@@ -0,0 +1,5 @@
+alter table "public"."flows" add column "is_service" boolean
+ not null default 'false';
+
+UPDATE "public"."flows" SET is_service = true
+    WHERE "public"."flows"."status" = 'online';
```
zz-hh-aa, line 5: Using online status in this migration as a proxy for `is_service`--does that seem reasonable? Is it fine to put that in this migration rather than just running on prod? Seemed like the best way of initially populating the column to me.

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Flow/Visibility/IsService/queries.ts
```diff
@@ -0,0 +1,48 @@
...
+  flow: flows_by_pk(id: $flowId) {
+    isService: is_service
+    team {
+      settings: team_settings {
+        isTrial: is_trial
+      }
+    }
+  }
+}
+
```
jessicamcinchak: nit: empty line / formatting of `gql` block looks a little off here !

e2e/tests/ui-driven/src/helpers/navigateAndPublish.ts
```diff
@@ -38,8 +38,18 @@ export const publishService = async (page: Page) => {
...
+};
+
+export const makeFlowAService = async (page: Page) => {
+  await page.getByText("Make this a user-facing service").click();
+  await page.getByText("Set to service").click();
+
+  await expect(page.getByText("Settings updated successfully")).toBeVisible();
+};
+
+export const turnServiceOnline = async (page: Page) => {
```
jessicamcinchak, line 52: 👍

apps/hasura.planx.uk/migrations/default/1778154943749_alter_table_public_flows_add_column_is_service/up.sql
jessicamcinchak: This makes sense to me! 

I do think it might highlight a number of "flows" which have been incorrectly/unnecessarily "online" for quite some time that we may want to manually review/update in near future!

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Flow/Visibility/IsService/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+              cancelText="Cancel"
+            >
+              Are you sure you want to set this flow as a user-facing service?
+            </ConfirmationDialog>
+          </>
+        );
+      }}
+    </SettingsFormContainer>
+  );
+};
```
jessicamcinchak, line 113: Copy all reads really clear here ➕ "services" versus "flows" historically not easy to weed through and explain !

apps/hasura.planx.uk/migrations/default/1778154943749_alter_table_public_flows_add_column_is_service/up.sql
jessicamcinchak: Checking out pizza now and clicking through settings pages and seeing a few inconsistent results here that suggest migration hasn't applied as expected? Any ideas what's going on?
 
- ❓ "Offline" + unexpectedly "service" https://6640.planx.pizza/app/barnet/apply-for-works-to-trees-template/settings/visibility
- ❓ "Online" + unexpectedly "flow" https://6640.planx.pizza/app/barnet/apply-for-planning-permission/settings/visibility
- ✅ "Offline" + "flow" https://6640.planx.pizza/app/barnet/confirmation-pages/settings/visibility

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Flow/Visibility/index.tsx
```diff
@@ -1,15 +1,23 @@
...
+  const isService = isServiceData?.flow.isService;
+
   return (
     <>
-      <FlowStatus />
+      <IsService />
+      {isService && <FlowStatus />}
       <FlowCopy />
-      <LPSListing />
+      {isService && <LPSListing />}
```
jessicamcinchak, line 20: nit: When I toggle on `isService`, I think it'd be visually clearer if any conditional settings were grouped/visible immediately below what I just toggled? (In future maybe even with left indent border - one for @ianjon3s when back?)

Versus right now, the always-visible `FlowCopy` sits _between_ the two conditional ones and it's a bit confusing to visually parse what's changed! Compounded by "Flow status" still reading as "Flow", when now only possible if "Service" !

jessicamcinchak (changes requested):
This is looking great! Everything working as expected when I change a flow to a service, but one issue noted below about unexpected initial "service" assignment results on pizza

apps/hasura.planx.uk/migrations/default/1778154943749_alter_table_public_flows_add_column_is_service/up.sql
zz-hh-aa: Thanks Jess--I checked my local environment and `is_service` matched `status = online`; I rebuilt and re-synced data / re-ran migrations a couple times locally too and my db was always as expected. 

I also rebuilt the Pizza and then ran 
```
UPDATE "public"."flows" SET is_service = true
    WHERE "public"."flows"."status" = 'online';
```

And saw that data is as expected (a `COUNT(*)` of flows where `status = online` and `is_service IS TRUE` was 312, and online + false counted as 0). (Pizza db now what we would expect it to be.)

I don't know why these would have been inconsistent before but I can't seem to replicate the inconsistencies. Any other ideas? Or do you think it's okay to merge to `main` and immediate check that the migration applies as expected?

jessicamcinchak: Sorry for delay coming back to this & thanks for looking into ! Let's count as a weird pizza quirk for now, and happy to immediately double check this after merging to `main` to ensure initial migration runs as expected there!

jessicamcinchak (approved):

PR #6641, opened 2026-05-11
fix: store redirectTo in sessionStorage so it works after oauth

joecarver:
I noticed that the redirectTo query param was getting discarded throughout the google oauth process (probably the same with the Microsoft flow as well). Storing the url in SessionStorage and retrieving it before accessing the app should be a pretty robust fix for this.

### DIFF
--- a/apps/editor.planx.uk/src/routes/(auth)/login.tsx
+++ b/apps/editor.planx.uk/src/routes/(auth)/login.tsx
@@ -36,6 +36,6 @@ export const Route = createFileRoute("/(auth)/login")({
   pendingComponent: DelayedLoadingIndicator,
 });
 
-function isValidRedirect(path: string | null): path is string {
+function isValidRedirect(path: string | null): boolean {
   return Boolean(path && path.startsWith("/") && !path.startsWith("//"));
 }
--- a/apps/editor.planx.uk/src/routes/(auth)/login.tsx
+++ b/apps/editor.planx.uk/src/routes/(auth)/login.tsx
@@ -1,7 +1,10 @@
-import { createFileRoute } from "@tanstack/react-router";
+import { createFileRoute, redirect } from "@tanstack/react-router";
 import DelayedLoadingIndicator from "components/DelayedLoadingIndicator/DelayedLoadingIndicator";
+import { useStore } from "pages/FlowEditor/lib/store";
 import Login from "pages/Login/Login";
 import { z } from "zod";
+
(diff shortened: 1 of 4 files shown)


apps/editor.planx.uk/src/routes/_authenticated/route.tsx
```diff
@@ -20,6 +20,19 @@ export const Route = createFileRoute("/_authenticated")({
         replace: true,
       });
     }
+
+    const pendingRedirect = sessionStorage.getItem(
+      "planx_redirect_after_login",
+    );
```
DafyddLlyr: nit: Can we import this key?

apps/editor.planx.uk/src/routes/(auth)/login.tsx
```diff
@@ -1,14 +1,41 @@
...
+      });
+    }
+  },
   component: Login,
   pendingComponent: DelayedLoadingIndicator,
 });
+
+function isValidRedirect(path: string | null): boolean {
+  return Boolean(path && path.startsWith("/") && !path.startsWith("//"));
+}
```
DafyddLlyr: This has a rather innocuous name but is actually an important security check (e.g. it should block `https://editor.planx.uk?redirectTo=https://evil.com`)

I think we should - 
 - rename it appropriately to highlight that it's a security check
 - add a simple suite of unit tests
 - maybe use `new URL()` to offload parsing, e.g. 


```ts
// We provide a dummy base URL
// If 'path' is "https://evil.com", the URL object will adopt that origin
// If 'path' is "/dashboard", it will use the dummy origin
const url = new URL(path, window.location.origin);
    
// 1. Ensure the origin matches exactly
// 2. Still check it doesn't start with // (which URL() might resolve)
return url.origin === window.location.origin && !path.startsWith("//");
```

DafyddLlyr (changes requested):

apps/editor.planx.uk/src/routes/(auth)/login.tsx
joecarver: good shout! Done here https://github.com/theopensystemslab/planx-new/pull/6641/commits/908effc2b99e919ae9fe94494471e4149782b115

DafyddLlyr (approved):
Fab..! Appreciate those changes. Once this has moved to production please update the following card on Trello - https://trello.com/c/047oaBJh/2962-fix-redirects-after-auth?filter=redirect

PR #6644, opened 2026-05-11
fix: ResponsiveChecklist test timeout

joecarver:
Increaes the test timeout for 
   `✓ Responsive Checklist editor component > can construct a valid payload  5173ms`
   
 to 15000ms
 
 I don't think there's any actual race conditions in this test, it just takes a long time, around 5s hence why it was intermittently failing. So increasing the timeout seems like a decent short term bandage for CI.


DafyddLlyr (approved):
Also had a look at this in parallel and came to the same conclusion...!

PR #6647, opened 2026-05-11
feat(ci): Generate sourcemaps on build, and upload to Airbrake

DafyddLlyr:
## What does this PR do?
 - Generates sourcemaps for the Editor build, and uploads to Airbrake via API client
 -  Generates sourcemaps for the API, which Node can read from the disk due to the `--enable-source-maps` flag. This is then written to the stack trace which is sent to Airbrake
 
This will mean that when an error is thrown in Airbrake, narrowing down the cause will be a much simpler and more reliable process.

## Motivation
This has been on the to-do list for quite a while, and I've tried a few times and been blocked by out-of-memory issues by Vite (see https://github.com/vitejs/vite/issues/2433). For now I think using an override is fine, and it should be resolved for Vite v8 anyway.

We've seen a number of bugs creep in recently, and with a growing team I think it's important to make the debugging and fixing process as streamlined as possible. Going forward I'll also put some work into these caught issues and work to resolve more of them..!

## Testing...
 - Run builds locally, see `.map.js` files generated for both projects
 - Throw an error on Editor, see sourcemaps being used on Airbrake. Example here, pointing to exact line of non-minified code - https://planx.airbrake.io/projects/329753/groups/4330208537499245641?tab=overview
 - Throw an error on API, see sourcemaps being used on Airbrake. Example here - https://planx.airbrake.io/projects/329753/groups/4330248737529109929?tab=overview

### DIFF
--- a/.github/workflows/build-planx-frontend.yml
+++ b/.github/workflows/build-planx-frontend.yml
@@ -23,6 +23,10 @@ on:
         description: Value for VITE_APP_SHAREDB_URL
         required: true
         type: string
+      vite-app-url:
+        description: Public base URL of the deployed frontend (e.g. https://editor.planx.uk)
+        required: true
+        type: string
       editor-directory:
         description: Path to the editor app directory
         required: false
@@ -52,9 +56,25 @@ jobs:
           VITE_APP_HASURA_WEBSOCKET: ${{ inputs.vite-hasura-websocket }}
           VITE_APP_MAPBOX_ACCESS_TOKEN: ${{ secrets.MAPBOX_ACCESS_TOKEN }}
           VITE_APP_SHAREDB_URL: ${{ inputs.vite-sharedb-url }}
+      - name: Upload source maps to Airbrake
+        working-directory: ${{ inputs.editor-directory }}
+        env:
+          AIRBRAKE_PROJECT_ID: ${{ secrets.AIRBRAKE_PROJECT_ID }}
+          AIRBRAKE_PROJECT_KEY: ${{ secrets.AIRBRAKE_PROJECT_KEY }}
+        run: |
+          for mapfile in build/assets/*.map; do
+            jsname=$(basename "${mapfile%.map}")
+            curl -sS --fail -X POST \
+              "https://api.airbrake.io/api/v4/projects/${AIRBRAKE_PROJECT_ID}/sourcemaps" \
+              -H "Authorization: Bearer ${AIRBRAKE_PROJECT_KEY}" \
+              -F "file=@${mapfile}" \
+              -F "name=${{ inputs.vite-app-url }}/assets/${jsname}"
+          done
       - name: Upload PlanX frontend build artifact
         uses: actions/upload-artifact@v7
         with:
           name: planx_frontend_build
-          path: ./${{ inputs.editor-directory }}/build
+          path: |
+            ./${{ inputs.editor-directory }}/build
+            !./${{ inputs.editor-directory }}/build/**/*.map
           if-no-files-found: error
--- a/apps/api.planx.uk/modules/test/controller.ts
+++ b/apps/api.planx.uk/modules/test/controller.ts
@@ -5,3 +5,7 @@ export const testSessionMethods: RequestHandler = (req, res) => {
   const hasSave = typeof req.session?.save === "function";
   res.status(200).json({ hasRegenerate, hasSave });
 };
+
+export const testAirbrake: RequestHandler = (_req, _res, next) => {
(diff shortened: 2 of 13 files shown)


.github/workflows/build-planx-frontend.yml
```diff
@@ -52,9 +56,25 @@ jobs:
...
+          AIRBRAKE_PROJECT_KEY: ${{ secrets.AIRBRAKE_PROJECT_KEY }}
+        run: |
+          for mapfile in build/assets/*.map; do
+            jsname=$(basename "${mapfile%.map}")
+            curl -sS --fail -X POST \
+              "https://api.airbrake.io/api/v4/projects/${AIRBRAKE_PROJECT_ID}/sourcemaps" \
+              -H "Authorization: Bearer ${AIRBRAKE_PROJECT_KEY}" \
+              -F "file=@${mapfile}" \
+              -F "name=${{ inputs.vite-app-url }}/assets/${jsname}"
+          done
```
DafyddLlyr, line 72: Docs: https://docs.airbrake.io/docs/devops-tools/api/#source-maps-v4

.github/workflows/build-planx-frontend.yml
```diff
@@ -52,9 +56,25 @@ jobs:
...
       - name: Upload PlanX frontend build artifact
         uses: actions/upload-artifact@v7
         with:
           name: planx_frontend_build
-          path: ./${{ inputs.editor-directory }}/build
+          path: |
+            ./${{ inputs.editor-directory }}/build
+            !./${{ inputs.editor-directory }}/build/**/*.map
```
DafyddLlyr, line 79: Upload all build files, excluding `.js.map` files.

apps/api.planx.uk/modules/test/controller.ts
DafyddLlyr, line 1: This is just a temp endpoint for testing on this PR, I'll remove after approval and before merging 🗑️ 🧹

apps/editor.planx.uk/src/airbrake.test.ts
DafyddLlyr, line 1: Currently we don't actually log Airbrake errors thrown on Pizzas - I think the original thinking here was that it could be too noisy? There was also a bug I spotted below that seems to suggest this may not have been intentional though 🤔 

I'd be keen to try this out for a period and then toggle off if we decided to. I'm keen to catch these issues early where possible.

apps/editor.planx.uk/src/airbrake.ts
```diff
@@ -35,7 +35,7 @@ function getEnvForAllowedHosts(host: string) {
       return "staging";
 
     default:
-      "pullrequest";
+      return "pizza";
```
DafyddLlyr, line 38: Bug! There was never a `return` here so this always returned `undefined`, meaning Airbrake was never running here.

I've also updated this from `pullrequest` -> `pizza` as this matched what we use in Airbrake on the API and matches how we describe our environments both in the codebase and elsewhere. I think this is the only real `"pullrequest"` reference in the casebase vs many other `"pizza"` references.

apps/editor.planx.uk/package.json
```diff
@@ -159,7 +159,7 @@
   "scripts": {
     "prestart": "../../scripts/check-containers.sh dev",
     "start": "vite --open",
-    "build": "tsc && vite build",
+    "build": "tsc && NODE_OPTIONS=--max-old-space-size=4096 vite build",
```
DafyddLlyr, line 162: Please see PR description and the following issues - 

 - https://github.com/vitejs/vite/issues/10067
 - https://github.com/vitejs/vite/issues/13666

docker-compose.pizza.yml
```diff
@@ -11,6 +11,8 @@ services:
       - "/api/dist"
     environment:
       NODE_ENV: "pizza"
+      AIRBRAKE_PROJECT_ID: ${AIRBRAKE_PROJECT_ID}
+      AIRBRAKE_PROJECT_KEY: ${AIRBRAKE_PROJECT_KEY}
```
DafyddLlyr, line 15: These env vars have also been added to `root.env` on AWS S3.

apps/editor.planx.uk/src/airbrake.test.ts
jessicamcinchak: If you search #planx-errors for `pullrequest` there's lots of results/history for errors logged by pizzas ? 

I agree we want to (keep) logging these !

jessicamcinchak (approved):
Thanks for this one!

PR #6660, opened 2026-05-12
[infra] rewrite HOWTO for setting up custom domains

freemvmt:
After #6533, we document here the process for either onboarding a new council, or migrating an existing custom domain from legacy infra to the new shared CDN (with auto-renewing SSL certs managed on our end).

Still to do: prepare an externally facing explanation to replace [this doc](https://www.notion.so/opensystemslab/9-Set-up-custom-subdomains-3000ef5212cc43a5a88f46563142f82a)


DafyddLlyr (approved):
Very clear - thank you!

jessicamcinchak (approved):

freemvmt:
This howto will change again but merging it in order to revise further alongside infra changes.

PR #6667, opened 2026-05-13
fix: Reduce notification lag

ianjon3s:
## What does this PR do?

Feature flag:
```
```

Currently notifications are running on a Hasura scheduled event, causing a significant lag to notification updates.

Switching to `useSubscription` allows near instant updates.

### DIFF
--- a/apps/api.planx.uk/modules/flows/publish/service.ts
+++ b/apps/api.planx.uk/modules/flows/publish/service.ts
@@ -10,6 +10,7 @@ import { dataMerged, getMostRecentPublishedFlow } from "../../../helpers.js";
 import { createScheduledEvent } from "../../../lib/hasura/metadata/index.js";
 import type { CreateScheduledEventResponse } from "../../../lib/hasura/metadata/types.js";
 import { userContext } from "../../auth/middleware.js";
+import { resolveNotification } from "../../notifications/service.js";
 import { buildNodeTypeSet, createFlowTypeMap } from "../validate/helpers.js";
 
 interface PublishFlow {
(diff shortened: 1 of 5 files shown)


apps/api.planx.uk/modules/flows/publish/service.ts
```diff
@@ -120,47 +119,31 @@ export const publishFlow = async (
...
-          webhook: `{{HASURA_PLANX_API_URL}}/flows/${flowId}/update-templated-flow/${templatedFlowId}`,
-          schedule_at: new Date(
-            new Date().getTime() + (i > 0 ? i * 10 : 0) * 1000,
-          ), // Stagger events by 10 seconds starting at "now"
-          payload: {
-            sourceFlowId: flowId,
-            templatedFlowId: templatedFlowId,
-          },
-          comment: `update_templated_flow_${templatedFlowId}`,
-        }),
```
jessicamcinchak, line 140: Curious why we have moved away from scheduled events to direct requests here? Are you trying to reduce 10-second stagger? The stagger is currently intentional in order to not overwhelm Hasura with 20+ simultaneous flow update requests (which can be large/timely!), but 10 seconds is probably conservative and we could try reduing this to ~5-7sec? 

For me, key benefits of scheduled events over direct requests are: 
- ability to easily retry or manually trigger single updates (on failure or while debugging)
- more visible logging/auditability via Hasura metadata tables (versus filtering AWS logs)
  -  we also know we may want to show source template owners dashboards in the future of each templated flow child status/last update and this would rely on querying metadata views (eg easier to do if scheduled events stay in place) !!

Can we "reduce lag" with editor-side changes only? The switch to `subscription` alone should solve the page-refresh feedback!

ianjon3s: Appreciate the sense check here, yes the intention is to both reduce the 10 second lag and switch to auto-refresh within the component. I was looking for an approach that would avoid repeated requests to Hasura, as I was informed (by Claude code) that this could become an issue when reducing the stagger time.

I've reverted the changes so that they affect the editor-side only, with the backend restored to the scheduled event approach. As you've suggested I've reduced the stagger to 5s.

jessicamcinchak (approved):
This works for me, thanks for API reverts here ! 

On the pizza, my notification badge was resolved on publish without refreshing the page as expected ➕

ianjon3s:
Thanks @jessicamcinchak !

PR #6668, opened 2026-05-13
feat: Add resolved by info to notification cards

ianjon3s:
## What does this PR do?

Feature flag:
```
```

- Adds record of "resolved by" to notifications, so that cards can output as below:

image

### DIFF
--- a/apps/api.planx.uk/modules/notifications/service.ts
+++ b/apps/api.planx.uk/modules/notifications/service.ts
@@ -1,5 +1,7 @@
+import { CoreDomainClient } from "@opensystemslab/planx-core";
 import { gql } from "graphql-request";
 import { $api } from "../../client/index.js";
+import { buildJWTForAPIRole } from "../auth/service/jwt.js";
 
 interface UpdateNotifications {
   notifications: {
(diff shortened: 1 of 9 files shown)


DafyddLlyr (dismissed):
This is a perfect use-case for column presents.

If we just attach a `resolved_by` column, a user could just add any value they want here via the API. A column present would make it simpler for the frontend (no changes on write) as well as more secure / reliable. We can just populate this column by reading from the `x-hasura-user-id` session variable embedded in the JWT.

Docs: https://hasura.io/docs/2.0/auth/authorization/permissions/column-presets/

ianjon3s:
Thanks for this @DafyddLlyr , I've created the column preset `resolved_by: x-hasura-user-id`, and fixed formatting issues (which were causing API test fails)

apps/api.planx.uk/modules/notifications/service.ts
```diff
@@ -12,11 +14,24 @@ interface UpdateNotifications {
...
+        auth: { jwt: buildJWTForAPIRole(publisherId) },
+        targetURL: process.env.HASURA_GRAPHQL_URL!,
+      })
+    : $api;
+
   // If many unresolved notifications of the same "type" exist for this flow, resolve them all
   //   If no notifications match these conditions, it'll simply return [] and still "succeed"
   const resolveNotificationResponse =
-    await $api.client.request<UpdateNotifications>(
+    await client.client.request<UpdateNotifications>(
```
jessicamcinchak: Sorry if I'm misunderstanding a prior decision here but: 
- Why do we need to swap out the client here? (and change the `buildJWTForAPIRole` entirely?)
- Why can't we simply add `resolved_by: $publisherId` to the `_set` in the mutation? 
  - Then using a Hasura "relationship" on resolvedBy <> user we could get the same first name / last name we need in the subscription/query? (I think the foreign key on the new column might be too strict as-is, eg we want to be able to keep notification history even if an editor is deleted? Relationship would be more flexible?)

Don't want to add more unnecessary back + forth here so I'll open a branch shortly which updates the data model only without touching API client here. Then Ian you could rebase/cherry pick only editor changes on top of that? (Apologies for weird division of work here, I should have spotted data limitations here earlier and picked up sooner so you could crack on with UI-only!)

jessicamcinchak: Okay here's a simplified alternative proposal: https://github.com/theopensystemslab/planx-new/pull/6690

If it's okay by you Ian - I'd like to avoid merging the change to `buildJWTForAPIRole` on this branch which I don't fully understand the motivations/implications of !

Apologies again here, hope the UI bits are straightforward enough to re-apply on top of 6690! Let's stay in touch about this next week :slightly_smiling_face: (And please never hesitate to tag me in for data/backend changes when required)

ianjon3s: Thanks for this @jessicamcinchak , I'm rebasing off your branch now, will simplify and push up this branch once I've worked through the rebase

jessicamcinchak (approved):
Super - thanks for rebasing here, much nicer diff !

ianjon3s:
Beat me to it, thanks @jessicamcinchak !

PR #6673, opened 2026-05-14
feat(lps): Storybook

DafyddLlyr:
## What does this PR do?
 - Configures Storybook for the LocalPlanning.services project
 - Adds basic stories for React components, broadly following [Atomic Design principles](https://atomicdesign.bradfrost.com/)
 
Right now I've not considered hosting this somewhere - we could easily get it up on Netlify or something though if we wanted to. For this PR, it's local-dev only via `pnpm storybook`.
 
### Why just React components?
It is possible to configure Storybook and Astro, using something like https://storybook-astro.org/

However, I think the best way to implement this would be to upgrade to Astro v6 first, then implement this - I had a crack at it and it wasn't _that_ simple, the PR was looking pretty big.

Most of the high complexity stuff is in React components anyway, so this is a nice 80/20 solution.

Once we've upgraded to v6, it would be nice to take a look at this again and add new Atoms / Molecules / Organisms / Templates / Pages.

## Motivation
I've wanted to do this for a while but put it off due to the Astro / Storybook compatibility mentioned above. Working on https://github.com/theopensystemslab/planx-new/pull/6666 this week reminded me both how awkard it is to set up the right mock data for local dev as well as the low test coverage here. I figured implementing Storybook would be a nice solution here - easier to test and develop, and set up mock data.

We'll also be running an a11y audit on LPS this year, and hopefully this can be expanded by then to be a helpful tool in that process.

## Demo

https://github.com/user-attachments/assets/0050c68e-9f7e-4ae8-a704-12f45826bb8b

### DIFF
--- a/apps/localplanning.services/.storybook/main.ts
+++ b/apps/localplanning.services/.storybook/main.ts
@@ -1,0 +1,43 @@
+import type { StorybookConfig } from "@storybook/react-vite";
+import path from "path";
+import { fileURLToPath } from "url";
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+const config: StorybookConfig = {
+  stories: ["../src/**/*.stories.@(ts|tsx)"],
+  addons: [
+    "@storybook/addon-a11y",
+    "msw-storybook-addon",
+  ],
+  framework: {
+    name: "@storybook/react-vite",
+    options: {},
+  },
+  async viteFinal(config) {
+    const { default: tailwindcss } = await import("@tailwindcss/vite");
+
+    config.plugins = [...(config.plugins ?? []), tailwindcss()];
+
+    config.resolve = {
+      ...config.resolve,
+      alias: {
+        ...(config.resolve?.alias ?? {}),
+        // Mock Astro's virtual modules
+        "astro:env/client": path.resolve(__dirname, "./mocks/astro-env.ts"),
+        "astro:transitions/client": path.resolve(__dirname, "./mocks/astro-transitions.ts"),
+        // Resolve path aliases to match tsconfig.json
+        "@components": path.resolve(__dirname, "../src/components"),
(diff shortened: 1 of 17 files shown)


apps/localplanning.services/public/mockServiceWorker.js
DafyddLlyr, line 1: Autogenerated file when installing `msw`

apps/localplanning.services/.storybook/main.ts
```diff
@@ -0,0 +1,43 @@
...
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+const config: StorybookConfig = {
+  stories: ["../src/**/*.stories.@(ts|tsx)"],
+  addons: [
+    "@storybook/addon-a11y",
+    "msw-storybook-addon",
+  ],
+  framework: {
+    name: "@storybook/react-vite",
```
jessicamcinchak, line 14: Curious if we were to introduce further Astro components here in the future post-upgrade as mentioned in description, could the Astro + React frameworks actually live side-by-side here? Or would LPS have _two_ Storybook instances - one covering React and the other Astro?

jessicamcinchak (approved):
Works perfectly locally, thanks for spinning up!

apps/localplanning.services/.storybook/main.ts
DafyddLlyr: I've not gone as far as getting this up and running with https://storybook-astro.org/ but I'd certainly want all components to live in a single Storybook instance - I think anything else would would feel quite unsatisfactory as an outcome.

jessicamcinchak: Yep totally fair not something we need to answer now, was a vaguely familiar problem from looking into Lit Storybook a long time ago and it seeming very hard to run multiple frameworks within single instance back then. But trust there are probably better solutions out there now!

PR #6680, opened 2026-05-14
[infra] define sharper cutover states for custom domains

freemvmt:
Relates to [this ticket](https://trello.com/c/8k8xg8AI). Seeks to address issues not foreseen in #6533, as discussed on [Slack](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1778612103221119?thread_ts=1778584338.647659&cid=C01E3AC0C03). Namely:

- that we cannot delete the old shared cert until the CloudFront distribution that relies on it has been attached to the new cert (and Pulumi will hang forever if we try, because AWS does not allow it)
- that AWS also does not allow 2 CloudFront distributions to have the same alias, throwing a `CNAMEAlreadyExists` - see [docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/alternate-domain-names-move.html)) - and hence we need to manually transfer the alias (custom domain) at the right stage

This PR achieves up to step 3 of the `B. Migrating a council from legacy setup` flow of the howto. After merge, we can test an appropriate `associate-alias` command works as expected, and proceed through the `cutover-ongoing` and `shared-final` states for the `planningtestlegacy` custom domain that we're using as a test candidate.


doc/how-to/how-to-setup-custom-domains.md
```diff
@@ -4,27 +4,32 @@
...
-| State | Meaning |
+| `cloudFrontState` | Actual state of infrastructure |
 | --- | --- |
-| `validation-only` | No CDN. Domain is on the "mining" cert to surface DNS validation records. |
-| `single-plus-validation` | Legacy per-domain CDN is live. Domain is also on mining cert. |
-| `single-plus-shared` | Legacy CDN + shared CDN both live and running in parallel (during cutover). |
-| `shared-only` | Shared CDN is live. This will be the eventual final state for all domains. |
+| `validation-only` | Domain is not associated with any CDN, but is on the "mining" certificate (cert) to surface DNS validation records (which we will send to council). |
+| `legacy-with-validation` | Domain is associated with (i.e. is an alias of) legacy CDN, which is backed by a council-provided cert. Domain is also on mining cert. |
+| `cutover-init` | Domain is still associated with legacy CDN, but shared CDN also certainly exists, and is backed by a shared cert provisioned by us, which includes the domain as an SAN.
```
jessicamcinchak, line 19: nit: what's SAN mean?

doc/how-to/how-to-setup-custom-domains.md
```diff
@@ -4,27 +4,32 @@
...
+| `shared-final` | Domain is associated with shared CDN. The legacy CDN, if this was a migration, has been torn down (otherwise, it never existed). This will be the eventual final state for all domains. |
+
+NB. We largely use 'CDN' and '[CloudFront] distribution' interchangeably although this is a simplification.
 
 ### Deploy ordering
 
-The `certificates` stack must always be deployed (manually) **before** the `application` stack. The application stack reads the shared certificate ARN from the certificates stack via a Pulumi stack reference.
+Correct sequencing is important! Usually, the `certificates` stack must be deployed (manually) _before_ the `application` stack, for example because it exports the shared certificate ARN via a Pulumi stack reference.
+
+However, there is nuance here. When it comes to tearing down legacy CloudFront distributions, we have to delete those first (i.e. deploy `application`) _before_ we delete the attached cert (i.e. deploy `certificates`.)
```
jessicamcinchak, line 29: Helpful !

doc/how-to/how-to-setup-custom-domains.md
```diff
@@ -143,37 +153,46 @@ Path: `validation-only` → `shared-only`
...
-The usual prompt for this process will be an impending expiry of a council's SSL cert, as flagged in the `#planx-notifications-ssl` Slack channel.
+The usual prompt for this process will be an impending expiry of a council's SSL cert, as flagged in the `#planx-notifications-ssl` Slack channel. You can check these dates at source in the [AWS console](https://us-east-1.console.aws.amazon.com/acm/certificates/list?region=us-east-1).
 
 NB. No new councils will be onboarded in the legacy mode, so when we migrate the last council, we can revise our documentation (i.e. delete this section).
 
-1. Advance the domain in question to the `single-plus-shared` state. Update the entry in `customDomains.ts`:
+1. Do steps 3-5 from flow A.
+
+    > NB. **If this is not the first domain on the shared CDN** and you want to expedite:
+    > - Skip the production deploy in step 3 (**TODO**: does this make sense?)
```
jessicamcinchak, line 169: I think makes sense! If I understand: if it's _not_ the first domain, then we skip certifcate layer deployment and only deploy application layer state change?

doc/how-to/how-to-setup-custom-domains.md
```diff
@@ -143,37 +153,46 @@ Path: `validation-only` → `shared-only`
...
     ```
 
-    Then **merge your PR** to deploy the `application` layer and update the source of truth on `main`.
+    Then **merge your PR to `main`** and **deploy your changes to `production`** to deploy the `application` layer via CI.
+
+    - The `certificates` deploy makes a new shared cert with the additional domain on it. At this point, we retain the old cert, because until the `application` deploy has run, the shared CDN still relies on it.
+    - The `application` deploy attaches the new cert to the shared CDN (and will create it first, if it doesn't already exist).
 
-    This adds the domain to the shared cert and shared CDN **while keeping the legacy per-domain CDN running**. Both CDNs serve traffic — which one the user hits depends on where the council's DNS points.
+    Keen observers will notice that we now have an orpaned cert which is not attached to any CDN - but we will clean that up later.
```
jessicamcinchak, line 193: 🙂 🧹

infrastructure/certificates/index.ts
```diff
@@ -78,7 +77,12 @@ if (validatedCustomDomains.length > 0) {
       subjectAlternativeNames: validatedCustomDomains.slice(1).map(d => d.domain),
       validationMethod: "DNS",
     },
-    { provider: usEast1 }
+    { provider: usEast1,
+      // ensure cert is replaced but *not* deleted - that is, deletion will be delayed to the next deployment of this layer
+      // (immediate deletion would fail since cert is in use by a CloudFront distribution, pending deployment of application layer)
```
jessicamcinchak, line 82: 👍

infrastructure/common/types.ts
```diff
@@ -18,16 +18,19 @@ export interface Team {
 };
 
 /**
- * Describes which CloudFront infrastructure serves a custom domain.
+ * Describes which AWS ACM + CloudFront infrastructure serves a custom domain.
  *
- * On-boarding path for new councils: validation-only → shared-only
- * Migration path for existing councils: single-plus-validation → single-plus-shared → shared-only
+ * On-boarding path for new councils: validation-only → shared-final
+ * Migration path for existing councils: legacy-with-validation → cutover-init -> cutover-ongoing → shared-final
```
jessicamcinchak, line 24: ⏯️ 👍

jessicamcinchak (approved):
Thanks for working through this!

doc/how-to/how-to-setup-custom-domains.md
freemvmt: Subject Alternative Name! will add this in :)

doc/how-to/how-to-setup-custom-domains.md
freemvmt: I think we definitely have to deploy the cert layer, but I thought we could get away without application layer - however I'm now changing my mind on this, because when we add the additional domain to the shared cert, we have to make a *new* cert, and we therefore need a run of the application layer to associate the shared CDN with that new cert, before we can proceed :)

doc/how-to/how-to-setup-custom-domains.md
freemvmt: It's possible Pulumi is smart enough to delete the retained cert on next run, but haven't confirmed this yet - otherwise it's a manual job we can bundle into the last step!

infrastructure/common/types.ts
freemvmt: It's a bit cumbersome tbh but I don't see a way to reduce steps!

PR #6683, opened 2026-05-15
chore: re-organise `dataMerged` and improve test coverage

jessicamcinchak:
https://trello.com/c/cD2Fpav1/3155-make-preview-panel-match-position-in-the-graph

Building out continuous preview panel navigation is going to rely on flattened `draftDataOnly` flow data. Currently it's slow and expensive to fetch. Motivated to try to optimise existing methods a bit first! 

**Changes:**
- [x] Moves `dataMerged` out of `helpers.ts` to its' own file under `shared/`
- [x] Updates `dataMerged` test mocks and scenarios to cover more complex cases (eg multiple nested flows, same nested flow referenced in multiple locations while only fetched + flattened once, etc)

**Next PR:**
- [ ] Refactor recursive `dataMerged` function to use a linear stack instead (would be too hard to review in this diff)
  - Record some baseline `/flatten-data` loading times here, should be faster :crossed_fingers:

### DIFF
--- a/apps/api.planx.uk/shared/dataMerged.test.ts
+++ b/apps/api.planx.uk/shared/dataMerged.test.ts
@@ -1,0 +1,644 @@
+import {
+  ComponentType,
+  type FlowGraph,
+  type Value,
+} from "@opensystemslab/planx-core/types";
+
+import { queryMock } from "../tests/graphqlQueryMock.js";
+import { dataMerged } from "./dataMerged.js";
+
+describe("dataMerged() function", () => {
+  beforeEach(() => {
+    queryMock.mockQuery({
+      name: "GetFlowData",
+      matchOnVariables: true,
+      variables: {
+        id: "parent-flow",
+      },
+      data: {
+        flow: {
+          data: draftParentFlow,
+          slug: "parent-flow",
+          team_id: 1,
+          team: {
+            slug: "testing",
+          },
+          publishedFlows: [{ data: flattenedParentFlow }],
+        },
+      },
+    });
+  });
+
+  it("flattens published nested flows by default", async () => {
+    queryMock.mockQuery({
+      name: "GetFlowData",
+      matchOnVariables: true,
+      variables: {
+        id: "nested-a",
+      },
+      data: {
+        flow: {
+          data: draftNestedFlowA,
+          slug: "nested-a",
+          team_id: 1,
+          team: {
+            slug: "testing",
+          },
+          publishedFlows: [
+            {
+              data: publishedNestedFlowA,
+            },
+          ],
+        },
+      },
+    });
+
+    queryMock.mockQuery({
+      name: "GetFlowData",
+      matchOnVariables: true,
+      variables: {
+        id: "nested-b",
+      },
+      data: {
+        flow: {
+          data: draftNestedFlowB,
+          slug: "nested-b",
+          team_id: 1,
+          team: {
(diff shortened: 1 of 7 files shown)


apps/api.planx.uk/shared/dataMerged.test.ts
```diff
@@ -0,0 +1,644 @@
...
+    });
+
+    await expect(dataMerged("parent-flow")).rejects.toThrow();
+  });
+});
+
+/**
+ * Draft parent flow data which includes 3 nested flow nodes (2 unique flows) and 1 folder
+ */
+const draftParentFlow: FlowGraph = {
```
jessicamcinchak, line 313: Visually, the `parent` flow I'm testing against looks like this: 
Image

Our prior mock here was a very simple flow with a single external portal/nested flow in exactly one location, no regular folders, etc. Updating the complexity of our mock here will give us more confidence in updating this function next !

apps/api.planx.uk/shared/dataMerged.test.ts
```diff
@@ -0,0 +1,644 @@
...
+    type: ComponentType.Content,
+  },
+};
+
+const draftNestedFlowB: FlowGraph = {
+  _root: {
+    edges: ["g4fB7FE0cz"],
+  },
+  g4fB7FE0cz: {
+    type: 250,
```
joecarver: nit: should this use the `ComponentType.Content` constant instead?

apps/api.planx.uk/shared/dataMerged.test.ts
```diff
@@ -0,0 +1,644 @@
...
+    },
+  },
+};
+
+const publishedNestedFlowB: FlowGraph = {
+  _root: {
+    edges: ["g4fB7FE0cz"],
+  },
+  g4fB7FE0cz: {
+    type: 250,
```
joecarver: same comment here w.r.t. `ComponentType.Content`

joecarver (commented):
Looks good and tests are passing! Just had a tiny comment around using a constant for ComponentType.

The other thing I was wondering is if we should add a test for this case, I guess its checking if the merge is stale and if so doing a re-merge.

```ts
const mostRecentPublishedFlowId =
        await getMostRecentPublishedFlowVersion(nodeId);
if (mostRecentPublishedFlowId !== node.data?.publishedFlowId) {
  await dataMerged(nodeId, ob, true, draftDataOnly);
}
```

I ran the tests with coverage on the dataMerged.ts file and these are the only lines not covered, so might be worth checking them off if its not too tricky to test!

apps/api.planx.uk/shared/dataMerged.test.ts
jessicamcinchak: Yes good catch - I update these manually to make mocks a little more human-readable, so easy to miss but doesn't actually break anything! Now fixed :white_check_mark:

jessicamcinchak:
Thanks Joe ! re: coverage, I've updated the first test (`flattens published nested flows by default`) to now also cover `getMostRecentPublishedFlowVersion` - thanks for noticing this gap !

joecarver (approved):

DafyddLlyr (approved):
Looking good - nothing to add, exciting first step...! Agree the challenge will be to fetch the relevant flows (once) upstream and then iterate and transform in the next PR 🚀

PR #6686, opened 2026-05-15
feat(ci): Trigger regression tests via PR comment

DafyddLlyr:
Follow up to post-mortem doc here - https://www.notion.so/opensystemslab/Postmortem-35135d469ad180da8c60caf36ef81396

This PR creates a new GHA which allows regression tests to be triggered by commenting `/regression-tests` on a PR. The aim here is to lower the friction for kicking these off on a given branch.

### Workflow
1. Comment `/regression-tests` on a PR,  👀 reaction appears
2. A comment posts - "⏳ Regression tests running on `{{branch-name}}`..."
3. Tests run against branch
4. That same comment is edited to either -  
    - ✅ Regression tests passed + link to run
    - ❌ Regression tests failed + link to run
    
This cannot be tested on this branch, we'll need to merge to `main` and then test/debug unfortunately..!

### DIFF
--- a/.github/workflows/regression-tests-on-comment.yml
+++ b/.github/workflows/regression-tests-on-comment.yml
@@ -8,7 +8,8 @@ jobs:
   dispatch:
     if: |
       github.event.issue.pull_request != null &&
-      startsWith(github.event.comment.body, '/regression-tests')
+      startsWith(github.event.comment.body, '/regression-tests') &&
+      contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association)
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
@@ -60,7 +61,6 @@ jobs:
               workflow_id: 'regression-tests.yml',
               ref: '${{ steps.pr.outputs.ref }}',
               inputs: {
-                pr_number: String(context.issue.number),
                 comment_id: String(${{ steps.comment.outputs.id }}),
               },
             });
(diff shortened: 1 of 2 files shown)


.github/workflows/regression-tests-on-comment.yml
```diff
@@ -0,0 +1,66 @@
...
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  dispatch:
+    if: |
+      github.event.issue.pull_request != null &&
+      startsWith(github.event.comment.body, '/regression-tests')
```
joecarver: Could we also check whether the user who posted the comment is a repo member? Otherwise any member of the public could trigger the CI with a comment

.github/workflows/regression-tests.yml
```diff
@@ -7,6 +7,15 @@ on:
     # Runs every morning at 06:06 (https://crontab.guru/#06_6_*_*_*)
     - cron: "06 6 * * *"
   workflow_dispatch:
+    inputs:
+      pr_number:
```
joecarver: doesn't look like we actually need/use the pr_number as the comment_id is enough to find the comment to update?

DafyddLlyr: Fixed in [9977e4b](https://github.com/theopensystemslab/planx-new/pull/6686/commits/9977e4b0f706e87e209b8ce3c70b4bd40fa52809)

.github/workflows/regression-tests-on-comment.yml
DafyddLlyr: Very very good spot 🤦 

Fixed in [9977e4b](https://github.com/theopensystemslab/planx-new/pull/6686/commits/9977e4b0f706e87e209b8ce3c70b4bd40fa52809)

joecarver (approved):
Nice!

DafyddLlyr:
Tested in #6688 and working as expected ✅

PR #6690, opened 2026-05-15
feat: add and populate `notifications.resolved_by`

jessicamcinchak:
Please see https://github.com/theopensystemslab/planx-new/pull/6668#discussion_r3250096139

This is just backend piece, no UI! 

To test, "resolve" a notification via publishing then `notifications` table to confirm your user ID / relationship is populated. 
Screenshot from 2026-05-15 14-38-54

### DIFF
--- a/apps/api.planx.uk/modules/flows/publish/service.ts
+++ b/apps/api.planx.uk/modules/flows/publish/service.ts
@@ -154,6 +154,7 @@ export const publishFlow = async (
       payload: {
         flowId: flowId,
         type: "updated_templated_flow",
+        resolvedBy: parseInt(userId),
       },
       comment: `resolve_notification_on_templated_flow_publish_${flowId}`,
     });
--- a/apps/hasura.planx.uk/metadata/databases/default/tables/public_notifications.yaml
+++ b/apps/hasura.planx.uk/metadata/databases/default/tables/public_notifications.yaml
@@ -23,34 +23,46 @@ object_relationships:
         remote_table:
           name: teams
           schema: public
+  - name: user
+    using:
+      manual_configuration:
+        column_mapping:
+          resolved_by: id
+        insertion_order: null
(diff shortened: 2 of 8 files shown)


apps/api.planx.uk/modules/flows/publish/service.ts
```diff
@@ -154,6 +154,7 @@ export const publishFlow = async (
       payload: {
         flowId: flowId,
         type: "updated_templated_flow",
+        resolvedBy: parseInt(userId),
```
jessicamcinchak, line 157: Because we resolve templated flow notification types as a side-effect of publishing, we _already have_ the user ID available to us and don't need to infer from the `x-hasura-user-id`.

As a scheduled event, this request is scoped to the `$api` client and there shouldn't really be any risk of unpredictable values being set here!

apps/hasura.planx.uk/metadata/databases/default/tables/public_notifications.yaml
```diff
@@ -23,34 +23,46 @@ object_relationships:
...
           schema: public
+  - name: user
+    using:
+      manual_configuration:
+        column_mapping:
+          resolved_by: id
+        insertion_order: null
+        remote_table:
+          name: users
+          schema: public
```
jessicamcinchak, line 34: I've chosen to link `resolved_by` to `users` via a Hasura "relationship" rather than "foreign_key" to avoid any unexpected conflicts on nightly staging data syncs (which drops & re-creates `users`, but we want staging `notifications` to persist) or `users` updates/deletions. 

The Hasura "relationship" accomplishes exactly the same in terms of being able to do a relational query. It's not as strict as an fkey on `insert`, but again `insert`/`update` permissions are otherwise very controlled here!

joecarver (approved):
All looks good to me!

PR #6691, opened 2026-05-18
chore(perf): refactor `dataMerged` to use a linear stack instead of recursion

jessicamcinchak:
Follows on from #6683. 

Precursor to https://trello.com/c/cD2Fpav1/3155-make-preview-panel-match-position-in-the-graph

`dataMerged` is the main function relied on by the `flows/:flowId/flatten-data` API endpoint which is called when we're loading draft, preview and published routes in the editor. "Merging"/"flattening" logic stays exactly the same here (tests all pass as-is), we've simply swapped how we're iterating over each flow (eg `Set` allows for faster lookups, `merged` tracks that we're not unnecessarily fetching flow data if the same external portal is nested in multiple locations).

Here's some quick baseline before/after metrics for flow id `49ae2f9b-f4a1-4f66-8cfc-b27bdbc835f1` (Barnet FOI, which nests `permitteddevelopment` - usually the common culprit for big slow flows):
- **Staging**
  - https://api.editor.planx.dev/flows/49ae2f9b-f4a1-4f66-8cfc-b27bdbc835f1/flatten-data &rarr; 9.1sec
  - https://api.editor.planx.dev/flows/49ae2f9b-f4a1-4f66-8cfc-b27bdbc835f1/flatten-data?draft=true &rarr; 4.6sec
- **Pizza**
  - https://api.6691.planx.pizza/flows/49ae2f9b-f4a1-4f66-8cfc-b27bdbc835f1/flatten-data &rarr; 4.8sec (nearly 50% faster!)
  - https://api.6691.planx.pizza/flows/49ae2f9b-f4a1-4f66-8cfc-b27bdbc835f1/flatten-data?draft=true &rarr; 3.1sec

For future gains, I think we'd want to consider Postgres-side experiments (partitions? missing indexes?) on the `published_flows` table? Pizzas and staging only sync the lastest published version per flow, whereas prod has full history and is a slower table to query - so these aren't perfect samples !


apps/api.planx.uk/shared/dataMerged.ts
```diff
@@ -2,95 +2,133 @@ import type { FlowGraph } from "@opensystemslab/planx-core/types";
 import { ComponentType } from "@opensystemslab/planx-core/types";
 
 import { getFlowData, getMostRecentPublishedFlowVersion } from "../helpers.js";
-import type { Flow, Node } from "../types.js";
+import type { Node } from "../types.js";
```
freemvmt, line 5: apparently this is deprecated - should we be using a type from `planx-core`?

apps/api.planx.uk/shared/dataMerged.ts
```diff
@@ -2,95 +2,133 @@ import type { FlowGraph } from "@opensystemslab/planx-core/types";
...
+        throw new Error(
+          `Publish flow ${team.slug}/${slug} before proceeding. All nested flows must be published.`,
+        );
+      }
+    }
+
+    // Fetch portals/nested flows
+    for (const [nodeId, node] of Object.entries(data)) {
+      const isExternalPortalRoot =
+        nodeId === "_root" && Object.keys(ob).length > 0;
```
freemvmt: just for my own understanding, when would there be a `_root` node other than the first entry in `data`?

e.g. for your example in PR desc (`49ae2f9b-f4a1-4f66-8cfc-b27bdbc835f1`), I think there is just 1 `_root` node, but several with type `310` i.e. `ExternalPortal` (and indeed 1 with type `300` i.e. `InternalPortal`)

so my understanding of the code here is to 'internalise' any external portal, which might manifest in two different ways? but why would there already be an internal portal? sorry, showing my ignorance here 😅 

... perhaps the existing internal portal is an example of a possibly 'stale' portal that has already been flattened?

apps/api.planx.uk/shared/dataMerged.ts
```diff
@@ -2,95 +2,133 @@ import type { FlowGraph } from "@opensystemslab/planx-core/types";
...
+        }
       }
     }
   }
 
   return ob as FlowGraph;
 };
+
+const fetchAndMergeStack = async (
+  ob: { [key: string]: Node } = {},
```
freemvmt: could we give `ob` a more explicative name?

apps/api.planx.uk/shared/dataMerged.ts
```diff
@@ -2,95 +2,133 @@ import type { FlowGraph } from "@opensystemslab/planx-core/types";
...
-
-      // recursively merge flow
-      if (!isMerged) {
-        await dataMerged(node.data?.flowId, ob, true, draftDataOnly);
-      }
-    }
-
-    // merge all other nodes
-    else ob[nodeId] = node;
+  if (stack.length > 0) {
```
freemvmt: what's the case where `stack` is empty already? looking at the code just above I reckon this could never happen, by our own construction?

apps/api.planx.uk/shared/dataMerged.ts
```diff
@@ -2,95 +2,133 @@ import type { FlowGraph } from "@opensystemslab/planx-core/types";
...
+  flowId: string,
   ob: { [key: string]: Node } = {},
   isPortal = false,
   draftDataOnly = false,
 ): Promise<FlowGraph> => {
-  // get the primary draft flow data, including its' latest published version
-  const response = await getFlowData(id);
-  const { slug, team, publishedFlows } = response;
-  let { data } = response;
+  const stack: Array<{ flowId: string; isPortal: boolean }> = [
```
freemvmt: we have this same `stack` type declared in multiple places - could pull out as a type at top of file?

freemvmt (approved):
Ok think I've largely got my head round this! Awesome work.

A few comments but nothing structural - much of it really just for my own understanding of the code here, or minor nits re read-/maintain-ability :)

freemvmt:
PS. I ran `curl http://localhost:7002/flows/49ae2f9b-f4a1-4f66-8cfc-b27bdbc835f1/flatten-data > out.json` on both `main` and this branch, and the results are not identical - should they be? The differences are concentrated at the very start and very beginning of the files.

Haven't tried to get my head round this yet but thought I would flag!

(I guess it could be that some portal was republished in the 5 mins or so between running each command, but seems unlikely!)

e.g. both results start like

```
{
  "_root": {
    "edges": [
      "zfOe2z5yfb",
      "CkMsTNgNcG",
      "DkNeIqsNJt",
      "rWYud0mRc6",
      "DDFEkgbcxz",
      "385V0sTHKP",
      "licB4ZyQyu",
      "OqMJOSUlOH",
      "GuKjdxjqCQ",
      "eJmoqYwOhV",
      "3TFp0N016m",
      "ngZeyiTMuF",
      "5pOt5ja49r",
      "b8TDjb1TiM",
      "Ymdh2XrVXK",
      "fqelgx027o",
      "VgtuwpWB7Z",
      "HATy2VvDnM"
    ]
  },
  "027ItK9qLb": { "data": { "text": "No" }, "type": 200 },
  "05rhf5ULaG": {
    "data": { "val": "flag.pp.priorApproval", "text": "Prior approval" },
    "type": 200,
    "edges": ["u3DbSdRIat"]
  },
  "0gEYCvLV3V": {
    "data": {
      "val": "internal",
      "text": "Internal alterations",
      "flags": ["flag.lbc.consentNeeded"]
    },
    "type": 200,
    "edges": ["oBGkJbsWg7"]
  },
  "2rYrCr1dWv": {
    "data": {
      "val": "true",
      "text": "Yes",
      "flags": ["flag.lbc.consentNeeded"]
    },
    "type": 200,
    "edges": ["oBGkJbsWg7"]
  },
  "385V0sTHKP": {
    "type": 300,
    "edges": ["6689164d-2a5e-422f-8d6a-d3ea11a2fbb9"],
    "data": { "flattenedFromExternalPortal": true }
  },
  ...
```

but then the result on `main` continues like...

```
  "6689164d-2a5e-422f-8d6a-d3ea11a2fbb9": {
    "edges": [
      "6FL5k5zovo",
      "JkyukIyZ5r",
      "ztEGQPy8eA",
      "pLO3rdvUCO",
      "PdibnhWYLn",
      "NMv6bveXnn",
      "u9NGAEwffA",
      "zUg8XBn5IM",
      "OaX0ysc0Aq",
      "3jtleKMI7B"
    ],
    "type": 300,
    "data": {
      "text": "opensystemslab/property-types",
      "flattenedFromExternalPortal": true,
      "templatedFrom": null,
      "publishedFlowId": 6748,
      "publishedAt": "2025-12-03T16:28:39.170881+00:00",
      "publishedBy": 84,
      "summary": "Remove unclassified fallback value."
    }
  },
  "0S7BqkuumE": {
    "data": {
      "val": "commercial.retail.licensedPremises",
      "text": "Off-licence"
    },
    "type": 200
  },
  ...
```

while the result on `jess/refactor-dataMerged` continues like...

```
  "3TFp0N016m": {
    "type": 300,
    "edges": ["86d80cbe-04ad-4cec-aaba-4d1833f19033"],
    "data": { "flattenedFromExternalPortal": true }
  },
  "3UPxHjdZcH": {
    "data": {
      "info": "<p>Your enquiry will not be submitted to a planning officer. This question will be used only to point you towards any further useful information or links.</p><p>This service is intended to provide you with information, but not advice or a decision. If you decide to proceed with the changes, and it turns out that planning permission was needed, you may be liable to enforcement action by planning officers in the future.</p>",
      "tags": [],
      "steps": [
        {
          "url": "https://www.barnet.gov.uk/planning-and-building-control/planning-applications-and-permissions/do-i-need-planning-permission/when-planning-permission-isnt-needed",
          "title": "Our website has helpful advice about Permitted Development Rights and how you can apply"
        },
        { "title": "Tell us what you think" }
      ],
      "title": "What would you like to do next?",
      "description": "<p>Choose how you would like to proceed.</p>",
      "isTemplatedNode": true,
      "templatedNodeInstructions": "You need to provide a URL for all applicable options, with the exception of 'Give us your feedback on this service' which should be left with a blank URL",
      "areTemplatedNodeInstructionsRequired": true
    },
    "type": 730
  },
  ...
```

apps/api.planx.uk/shared/dataMerged.ts
jessicamcinchak: Good question - extracted this to top of file as `MergedGraph` type and added a note ! 

We still rely on a few of these deprecated `Node` types (in more places than here!) because they are looser/less strict than planx-core's `FlowGraph` and `Node` which assume there's always a `_root` node, which is _not_ the case while we're in-the-process of flattening many things here ! 

It's a known piece of tech debt !

apps/api.planx.uk/shared/dataMerged.ts
jessicamcinchak: Yes good call now `mergedGraph` ! 

As `dataMerged` is no longer recursive, updating this also made me realise we don't need to pass it as a function argument anymore but rather than just track in internal function state :heavy_plus_sign:

apps/api.planx.uk/shared/dataMerged.ts
jessicamcinchak: Yep good spot - this is unnecessary + now removed !

apps/api.planx.uk/shared/dataMerged.ts
jessicamcinchak: Done!

apps/api.planx.uk/shared/dataMerged.ts
jessicamcinchak: Good question ! 

Let me attempt to explain:
- Yes, a "flow" data type is a dictionary of node that always starts with a single `_root`
- Flows can nest other flows - this is done via the 'external portal' component type
  - In "draft" data, an external portal component just has a reference to the nested flow's ID, not it's actual content. The "actual content" of an external portal is just a flow again, with it's own `_root`
- Flows can also use folders (aka 'internal portal' component type) to organise other nodes, including external portal node types (so these are actually distinct types that can exist side-by-side, one doesn't _always_ "become" the other!)
- When "flattened" (aka when `dataMerged` is run), internal portal types stay the same and external portal types _become_ internal portal types
  - This means that rather than an external portal being a _reference_, we've actually fetching its' content and we want to merge it into the main flow - but the main flow can only ever have a single `_root` ! 
  - So the new transformed internal portal node acts like an invisible-to-the-user folder to hold the `edges` of the original external portal node so that we still know which order to navigate its' content. 

Does that sort of make sense? Screenshot from this related PR with test mocks might help visualise a bit more https://github.com/theopensystemslab/planx-new/pull/6683#discussion_r3245701386

jessicamcinchak:
> PS. I ran `curl http://localhost:7002/flows/49ae2f9b-f4a1-4f66-8cfc-b27bdbc835f1/flatten-data > out.json` on both `main` and this branch, and the results are not identical - should they be? The differences are concentrated at the very start and very beginning of the files.
> 
> Haven't tried to get my head round this yet but thought I would flag!

Thanks for trying this & flagging !! 

After some double-checking, quite confident this is simply because flows are stored as jsonb and key order of returned object isn't guaranteed on query :melting_face: (hence vital role of node `edges`)

I ran a fresh data sync in my local env and on staging to ensure equal content and confirmed: 
- unflattened live flow data line-length was equal
- flattened draft flow data line-length was equal
- flattened preview flow data line-length was equal
  - flattened nodes were indeed not in same order here for me either, but I spot-checked many nodes and all existed in each env as expected with correct/equal props, just in different locations ! 

We have pretty good mock data for these API unit tests, but unfortunately don't mock these full council FOI flows because it'd tediously be like 30+ graphql query mocks (1 per external portal/nested flow reference) each with draft flow data + latest published version - would love to think of other/better ways in future! Eg is there ever a version of an end-to-end suite that would actually read flows from staging env (rather than static mocks) and assert certain results?

apps/api.planx.uk/shared/dataMerged.ts
freemvmt: Thanks Jess this helps a lot! Really appreciate you taking the time :)

freemvmt:
> After some double-checking, quite confident this is simply because flows are stored as jsonb and key order of returned object isn't guaranteed on query 🫠 (hence vital role of node `edges`)

Thanks for double checking this!

PR #6693, opened 2026-05-19
feat: analytics_enhancements view

zz-hh-aa:
Following the existing pattern of analytics materialized views, this PR creates a new view that unfurls the `analytics_logs.allow_list_answers._enhancements` jsonb into a new Metabase-digestible view for enhanced project description analytics.

# Questions
❓ Is it overkill to follow the existing pattern and make this a materialized view refreshed with a nightly cron job, when we don't have much data? I opted for this for the sake of consistency and the fact that it seemed more future-proof, but it could be adding more cron job cruft. 
❓ I also wondered if it might be more appropriate to create an additional column for the type of enhancements, again with a view to when we add additional enhancements in the future--but decided not to for now because of possible cruft. Please say if anyone thinks we should do otherwise here...

### DIFF
--- a/apps/hasura.planx.uk/metadata/databases/default/tables/public_analytics_enhancements.yaml
+++ b/apps/hasura.planx.uk/metadata/databases/default/tables/public_analytics_enhancements.yaml
@@ -1,0 +1,3 @@
+table:
+  name: analytics_enhancements
+  schema: public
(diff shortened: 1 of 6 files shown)


jessicamcinchak (approved):
Great! 

Good questions here, but personally not too worried about creating CRON cruft here and agree it's nice to have a consistent pattern that we know scales ➕

PR #6695, opened 2026-05-19
fix: ensure archived flows are not copiable or nestable

zz-hh-aa:
As reported [here](https://trello.com/c/T74ZBaOR).

The main change is adding an `{ _is_null: true }` check to `archived_at` in the query, but while I was here I refactored to use our [existing hook pattern](https://github.com/theopensystemslab/planx-new/blob/main/doc/guides/data-fetching-and-state-management.md). 

# Testing

1. Start creating a new flow from copy and check the list of copiable flows
2. Archive one of those flows
3. Start creating a new flow from copy and check the list again--the newly archived flow should no longer show up as an option


jessicamcinchak (approved):
Works as expected, thank you! 

Here, or in a follow-up PR, could you also please double-check that archived flows are similarly omited from list of nestable flows? (eg `getExternalPortals` helper/query)

PR #6699, opened 2026-05-19
[fix] pin docker build platform

freemvmt:
Various fixes/improvements for the CI pizza build.

Prompted by all pizza builds (that is, runs of `pull-request.yml` GHA workflow) suddenly failing. Careful read of logs revealed the following line:

```
Error response from daemon: no matching manifest for linux/arm64 in the manifest list entries: no match for platform in manifest: not found
```

We conject that this is a result of Vultr bumping their Alpine build to latest (i.e. 3.23, or perhaps applying a more recent security patch thereof), which includes a [major version upgrade](https://gitlab.alpinelinux.org/alpine/apk-tools/-/releases/v3.0.0?__goaway_challenge=cookie&__goaway_id=d71abc4027187129922685ad0389d226&__goaway_referer=https%3A%2F%2Fwiki.alpinelinux.org%2F) (to v3) of the `apk` package manager, which is somehow more fussy about having the correct manifest/being on right platform? Unsure of cause but could be e.g. something to do with Alpine no longer being able to emulate it's way out of wrong platform build (see [here](https://github.com/moby/buildkit/issues/6400#issuecomment-3670695463))...

Alternatively it could be that the default github runners have been swapped out for `arm64` when they were previously `amd64`, and are therefore building for the former by default without being told otherwise? Although that would be weird because we do explicitly request [ubuntu-24.04](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md) 🤔

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -310,7 +310,7 @@ jobs:
 
       - if: steps.create.outputs.ip_address != null
         name: Create commands
-        uses: appleboy/ssh-action@master
+        uses: appleboy/ssh-action@v1
         with:
           host: ${{ env.FULL_DOMAIN }}
           username: root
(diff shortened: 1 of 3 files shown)


jessicamcinchak (approved):
Thanks for sorting ! 🍕

PR #6702, opened 2026-05-19
fix: Notification createdBy and time formatting

ianjon3s:
## What does this PR do?

Tidy up tasks and fixes with notifications:
- Ensure timestamp is formatted the same as FlowCard/table
- Ensure ResolvedBy is populated correctly
- Update copy for "no notifications"


apps/hasura.planx.uk/migrations/default/1779223924522_alter_table_public_notifications_use_timestamptz/up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "public"."notifications"
+  ALTER COLUMN "created_at" TYPE timestamptz,
+  ALTER COLUMN "resolved_at" TYPE timestamptz;
```
jessicamcinchak, line 3: Good catch !

jessicamcinchak (approved):

PR #6706, opened 2026-05-20
feat(api): assign teamAdmin role to JWT

zz-hh-aa:
- Creates a `teamAdmin` enum in I `user_role` (missed this in #6608)
- Assigns role in API

Permissions have been tested locally too and, as expected, `teamEditor` permissions are granted / inherited when I assign myself a `teamAmin` role in Hasura.

### DIFF
--- a/apps/api.planx.uk/modules/auth/middleware.ts
+++ b/apps/api.planx.uk/modules/auth/middleware.ts
@@ -319,9 +319,14 @@ export const useRoleAuth: UseRoleAuth =
 export const useTeamViewerAuth = useRoleAuth([
   "teamViewer",
   "teamEditor",
+  "teamAdmin",
   "platformAdmin",
 ]);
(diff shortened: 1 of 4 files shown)


jessicamcinchak (approved):
Great! 

Should we also update `TeamRole` in planx-core (then bump here or in follow-up PR) to stay in sync? See https://github.com/theopensystemslab/planx-core/blob/main/src/types/roles.ts

zz-hh-aa:
> Should we also update `TeamRole` in planx-core (then bump here or in follow-up PR) to stay in sync? See https://github.com/theopensystemslab/planx-core/blob/main/src/types/roles.ts

I think you beat me to it actually :) 

https://github.com/theopensystemslab/planx-core/blob/5fe4e59622a3d87d6ce1b076fe22a88470e660e9/src/types/roles.ts#L11

jessicamcinchak:
Ha yes sorry - was remembering this, but not reading close enough :upside_down_face:

PR #6707, opened 2026-05-20
fix[tests]: bump API test timeout for `modules/gis/serivce/findLpa` and add mock for Resend

jessicamcinchak:
This has been regularly timing out on nightly regression test runs and local branches lately. Eg https://github.com/theopensystemslab/planx-new/actions/runs/26148104122

Timeout added here is consistent with a pre-existing similar one in main GIS module  (`modules/gis/service/index.test.ts`).

### DIFF
--- a/apps/api.planx.uk/modules/gis/service/findLpa/index.test.ts
+++ b/apps/api.planx.uk/modules/gis/service/findLpa/index.test.ts
@@ -56,6 +56,6 @@ describe("checking several points against local planning authority lookup API", 
             ),
           ).toEqual(location.lpas);
         });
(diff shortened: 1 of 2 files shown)


freemvmt (approved):
Looks good! Just as an aside, do we have any idea why this code was timing out?

jessicamcinchak:
@freemvmt good question - this is definitely the lazy "fix" approach :sweat_smile: 

Think the better check here would be to review how we're mocking Ordnance Survey API here? GIS mocks, as large geojson, may also just be bigger/slower than usual here !

PR #6709, opened 2026-05-20
chore: add Liverpool metadata

augustlindemer:
This is a **DRAFT** list. To be confirmed with Liverpool.

Captures A4 directions and large sites with historic planning condition.

We need to explore separately how anything not mapped to metadata here can be treated as a separate 'planning conditions' constraint (with no expectation that these are mapped to rules via content).


apps/api.planx.uk/modules/gis/service/digitalLand.ts
```diff
@@ -70,6 +71,7 @@ export const localAuthorityMetadata: Record<string, LocalAuthorityMetadata> = {
   "greater-cambridge-shared-planning": greaterCambridge,
   horsham,
   lambeth,
+  "liverpool-city": liverpoolCity,
```
jessicamcinchak, line 74: Unless Liverpool actually requested that this be their editor team `name`, I wonder if we should clean up editor team names while still easy (eg no live services yet) to remove unnecessary "City" suffixes, which quite a few new teams have? Maybe we need to ask Alki?

Will make these A4 mappings a bit easier to do in future / this is an extra tedious step to remember!

PR #6710, opened 2026-05-21
chore: re-build editor when running pnpm tests

joecarver:
This behavour is kind of assumed to be done as part of the main set up step, but wasn't being done. Uses the existing `build-editor.sh` script to only rebuild if there are changes in the working tree. Also added a note to README to remind that it needs to be re-built in between changes.


zz-hh-aa (approved):
Super useful, thanks! Tested and worked as expected ✨

jessicamcinchak (approved):

PR #6712, opened 2026-05-21
feat: dashboard / explore: flags, routes and stubbed pages 

joecarver:
Adds feature flags, new routes, and empty pages for the new pages.

I guess eventually we may want to set dashboard as the landing page and move flows to `/flows` but I decided against doing this just yet as the switching behaviour with the feature flag becomes a bit complicated, due to the data fetching and query param integrations.


apps/editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
```diff
@@ -134,12 +136,32 @@ function EditorNavMenu() {
...
             Icon: EditorIcon,
             route: `/app/${teamSlug}`,
             accessibleBy: "*",
           },
+          ...(hasFeatureFlag("EXPLORE")
+            ? [
+                {
+                  title: "Explore",
+                  Icon: ExploreIcon,
+                  route: `/app/${teamSlug}/explore`,
```
jessicamcinchak: This is expected nav menu location based on figmas, but I _think_ it's a bit unexpected for `/explore` route to be a child of `${teamSlug}` if for planx-level stats? Should it be `/app/explore` ? (or will its' queries actually rely on a `teamSlug` param to fetch data?)

apps/editor.planx.uk/src/routes/_authenticated/app/$team/explore.tsx
```diff
@@ -0,0 +1,13 @@
+import { createFileRoute, redirect } from "@tanstack/react-router";
+import { hasFeatureFlag } from "lib/featureFlags";
+import Explore from "pages/Explore";
+import React from "react";
+
+export const Route = createFileRoute("/_authenticated/app/$team/explore")({
+  beforeLoad: ({ params }) => {
+    if (!hasFeatureFlag("EXPLORE")) {
+      throw redirect({ to: "/app/$team", params: { team: params.team } });
```
jessicamcinchak: Related to above comment, I think I'd expect this redirect to simply go to `"/app"` ?

jessicamcinchak (changes requested):
Thanks for incremental PRs here ! One question re: nesting `explore` under team!

apps/editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
joecarver: ah yes very good point! will change :~)

apps/editor.planx.uk/src/routes/_authenticated/app/$team/explore.tsx
joecarver: done! [2f3f5ea](https://github.com/theopensystemslab/planx-new/pull/6712/commits/2f3f5ea02197fc9eda6a6205418400790de80e41)

apps/editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
joecarver: done! [2f3f5ea](https://github.com/theopensystemslab/planx-new/pull/6712/commits/2f3f5ea02197fc9eda6a6205418400790de80e41)

jessicamcinchak (approved):
Perfect!

PR #6715, opened 2026-05-21
[infra] simplify cutover to single state

freemvmt:
Relates to [this ticket](https://trello.com/c/8k8xg8AI). Builds on #6533, #6660 and #6680.

Should deliver the final deploy of the `application` layer on staging/`main` that I need to be confident of the migration process. Assuming this goes to plan, I just need to polish the howto (see TODOs) and we're done.

Did a light review below with some comments for further explanation.

### DIFF
--- a/doc/how-to/aws/how-to-setup-aws-sso-credentials.md
+++ b/doc/how-to/aws/how-to-setup-aws-sso-credentials.md
@@ -1,19 +1,26 @@
 # How to set up AWS SSO credentials
+
 This document describes the process of setting up credentials for your local AWS CLI. This will enable you to interact directly with our infrastructure using the CLI, as well as provision infrastructure using Pulumi.
 
 ## Process
+
 As part of onboarding, you should be set up to access AWS via [the SSO portal](https://opensystemslab.awsapps.com/start#/). Please ensure that you have set up and enabled MFA.
 
 1. Download the AWS CLI - see https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html
-1. Run `aws configure sso` - some of the required details can be found via [the SSO portal](https://opensystemslab.awsapps.com/start#/) in the 'Access keys' section:
+
+2. Run `aws configure sso` - some of the required details can be found via [the SSO portal](https://opensystemslab.awsapps.com/start#/) in the 'Access keys' section:
     - e.g. `SSO start URL` and `SSO Region` are listed under `AWS IAM Identity Center credentials (Recommended)`.
     - `SSO registration scopes` can be left as the default (`sso:account:access`).
-1. Allow access from the authorisation page that should open in your browser.
-1. You should be prompted for more details in the terminal:
+
+3. Allow access from the authorisation page that should open in your browser.
+
+4. You should be prompted for more details in the terminal:
     - `CLI default client region` should be `eu-west-2`
     - `CLI default output format` should be `json`
     - For the `CLI profile name` field, please use the format `planx-<STACK>`, e.g. `planx-staging`
-1. Repeat for all environments (staging, production)
-1. Test! You should be able to call commands using the CLI, for example `aws sts get-caller-identity --profile planx-staging` and get the expected result.
+
+5. Repeat for all environments (staging, production)
+
+6. Test! You should be able to call commands using the CLI, for example `aws sts get-caller-identity --profile planx-staging` and get the expected result.
 
 Occasionally, you'll be prompted to refresh your credentials, this can be done via `aws sso login --profile <PROFILE>`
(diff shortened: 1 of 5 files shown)


doc/how-to/aws/how-to-setup-aws-sso-credentials.md
```diff
@@ -1,19 +1,26 @@
 # How to set up AWS SSO credentials
+
 This document describes the process of setting up credentials for your local AWS CLI. This will enable you to interact directly with our infrastructure using the CLI, as well as provision infrastructure using Pulumi.
 
 ## Process
+
```
freemvmt, line 6: Just improved the formatting of this doc as I came through.

doc/how-to/how-to-setup-custom-domains.md
```diff
@@ -16,25 +18,24 @@ All custom domains are defined in [`infrastructure/common/customDomains.ts`](../
 | --- | --- |
 | `validation-only` | Domain is not associated with any CDN, but is on the "mining" certificate (cert) to surface DNS validation records (which we will send to council). |
 | `legacy-with-validation` | Domain is associated with (i.e. is an alias of) legacy CDN, which is backed by a council-provided cert. Domain is also on mining cert. |
-| `cutover-init` | Domain is still associated with legacy CDN, but shared CDN also certainly exists, and is backed by a shared cert provisioned by us, which includes the domain as an SAN.
```
freemvmt, line 19: Working up this PR has been a lesson in how CloudFront actually works - a ramification of that is that we only need one state to handle the 'cutover' (i.e. migration from the legacy CDN to the shared CDN), during which we execute multiple deploys of both `certificates` and `application`, and transfer the alias manually from the former to the latter using the AWS CLI.

doc/how-to/how-to-setup-custom-domains.md
```diff
@@ -185,30 +208,25 @@ NB. No new councils will be onboarded in the legacy mode, so when we migrate the
...
 
     ```sh
     cd infrastructure/application
     pulumi stack output customDomainsCdnDomainName --stack production
+    pulumi stack output customDomainsDistributionId --stack production
     ```
 
-    This returns a value like `d1234abcd.cloudfront.net`.
-
-4. **TODO**: Run the `associate-alias` command from AWS CLI to move the domain from the legacy CloudFront distribution to the shared CloudFront distribution. Explain why this works.
```
freemvmt, line 204: It turns out [associate-alias](https://docs.aws.amazon.com/cli/latest/reference/cloudfront/associate-alias.html) does not work for our purposes, despite seeming simpler and the [docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/alternate-domain-names-move-options.html) certainly suggesting that we should be able to use it to transfer aliases between standard distributions on the same account.

That is, I tried to run the following command:

```
aws cloudfront associate-alias \
      --alias planningserviceslegacy.planx.in \
      --target-distribution-id E123...
```

But it fails, demanding that we add DNS TXT records to validate ownership of the domain, even though the [docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/alternate-domain-names-move-create-target.html) say quite explicitly that this is only necessary when moving aliases between accounts.

Therefore we use the more fully featured [update-domain-association](https://docs.aws.amazon.com/cli/latest/reference/cloudfront/update-domain-association.html) command :)

infrastructure/common/types.ts
```diff
@@ -21,16 +21,15 @@ export interface Team {
...
  * See doc/how-to/how-to-setup-custom-domains.md for more detail.
  */
 export type CloudFrontState =
   | "validation-only"           // no CDN yet; mining cert surfaces DNS validation records
   | "legacy-with-validation"    // legacy per-domain CDN running; mining cert surfaces validation records
-  | "cutover-init"              // initialise migration: shared CDN spun up if needed; alias still on legacy CDN
-  | "cutover-ongoing"           // mid-migration: alias transferred to shared CDN; legacy CDN still running
-  | "shared-final";             // shared CDN only; legacy CDN torn down (if it exists)
+  | "cutover-ongoing"           // migrating: shared CDN spun up if needed; alias to be transferred from legacy to shared CDN
+  | "shared-final";             // shared CDN only; legacy CDN torn down (if it existed)
```
jessicamcinchak, line 32: Simpler ⭐

jessicamcinchak (approved):
All looks good to me, thanks again for working through all this and leaving clear notes !

infrastructure/common/types.ts
freemvmt: Was hoping it would end up like this! :)

PR #6717, opened 2026-05-21
feat: enable teamAdmin to be assigned in editor

zz-hh-aa:
# What does this PR do?
- In frontend, adds `teamAdmin` as intermediate option to permissions for both `platformAdmin` and `teamEditor`
- Adds a conditionally rendered `Switch` to the `AddUserModal` and `EditUserModal` components (only renders for team members and not `platformAdmin`). 
- Updates some copy ('Admin' chip -> 'Platform admin' and 'Team admin', etc.)

image


apps/editor.planx.uk/src/pages/FlowEditor/lib/store/user.ts
```diff
@@ -7,6 +7,10 @@ import { TeamStore } from "./team";
 export const getDisplayRole = (user: User): string => {
   if (user.isPlatformAdmin) return "Platform Admin";
   if (user.isAnalyst) return "Analyst";
+
+  const isTeamAdmin = user.teams.find((team) => team.role === "teamAdmin");
+  if (isTeamAdmin) return "teamAdmin";
```
joecarver: should we return `"Team Admin"` here to match the other return values?

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/user.ts
```diff
@@ -7,6 +7,10 @@ import { TeamStore } from "./team";
 export const getDisplayRole = (user: User): string => {
   if (user.isPlatformAdmin) return "Platform Admin";
   if (user.isAnalyst) return "Analyst";
+
+  const isTeamAdmin = user.teams.find((team) => team.role === "teamAdmin");
```
joecarver: nit: using .some() rather than .find() here might be more semantic

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/user.ts
joecarver: in fact, might make sense for the whole function to use `ROLE_LABELS` for its return type?

joecarver (approved):
Functionality seems spot on to me and nice and clean implementation! Couple of tiny suggestions with `getDisplayRole` otherwise all good

PR #6720, opened 2026-05-22
Revert "chore(deps): bump ws from 8.19.0 to 8.20.1 in /apps/sharedb.planx.uk in the npm_and_yarn group across 1 directory"

jessicamcinchak:
Reverts theopensystemslab/planx-new#6701

See bug thread https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1779446296118569 - this seems like most likely culprit in latest deploy for sharedb-related bugs! 

Nothing in particular stuck out in `ws` changelog and I'd checked that we'd merged last couple minor versions successfully (eg not dependabot ignored them), but if my memory is right this has definitely been a problematic package in the past to update ??

### DIFF
--- a/apps/sharedb.planx.uk/package.json
+++ b/apps/sharedb.planx.uk/package.json
@@ -8,7 +8,7 @@
     "jsdom": "^28.1.0",
     "pg": "^8.19.0",
     "sharedb": "^5.2.2",
-    "ws": "^8.20.1"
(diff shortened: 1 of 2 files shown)


jessicamcinchak:
Not root issue! Rather issue was onlyl recreatable when multiple editors were working in same flow at same time, which instead connects us way back to #839

PR #6723, opened 2026-05-26
feat: dashboard / explore: layout and display components

joecarver:
### What's in this PR?
- `StatsBanner` component - ready for linking up to real stats
- `DashboardWidget` component - re-usable card-like component for the dashboard / explore grid items (perhaps this could do with a better name)
- layout presenting these

These should provide a good basis for developing the actual content that will go inside them.

### DIFF
--- a/apps/editor.planx.uk/src/pages/Dashboard/index.tsx
+++ b/apps/editor.planx.uk/src/pages/Dashboard/index.tsx
@@ -1,10 +1,14 @@
 import Box from "@mui/material/Box";
 import Container from "@mui/material/Container";
 import Typography from "@mui/material/Typography";
-import { DashboardWidget } from "ui/editor/DashboardWidget";
 import React from "react";
+import { DashboardWidget } from "ui/editor/DashboardWidget";
+
+import { useStore } from "../../pages/FlowEditor/lib/store";
 
 export default function Dashboard() {
+  const team = useStore((state) => state.getTeam());
+
   return (
     <Container maxWidth="lg" sx={{ py: 4 }}>
       <Typography variant="h2" component="h1" gutterBottom>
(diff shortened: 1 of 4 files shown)


apps/editor.planx.uk/src/pages/Dashboard/index.tsx
```diff
@@ -1,13 +1,57 @@
...
+
+import { useStore } from "../../pages/FlowEditor/lib/store";
+import StatsBanner from "./components/StatsBanner";
 
 export default function Dashboard() {
+  const team = useStore((state) => state.getTeam());
+
   return (
-    <Container maxWidth="lg">
+    <Container maxWidth="lg" sx={{ py: 4 }}>
```
ianjon3s: Let's use the width of the current team/flows page for continuity (`contentWide`)

ianjon3s (approved):
Really simple and smart responsive layout as a starter. I've added one comment about reusing the `contentWide` container max-width for continuity.

Also about background colours, we tend to do the opposite to this setup when showing cards, using `background.paper` as the page background, and `background.default` (white) for cards.

apps/editor.planx.uk/src/ui/editor/DashboardWidget/index.tsx
```diff
@@ -0,0 +1,56 @@
+import Box from "@mui/material/Box";
+import { styled, SxProps, Theme } from "@mui/material/styles";
+import Typography from "@mui/material/Typography";
+import React from "react";
+import { CustomLink } from "ui/shared/CustomLink/CustomLink";
+
+interface DashboardWidgetProps {
+  title: string;
+  linkTarget?: string;
```
DafyddLlyr: We're losing type-safety here which we should be getting inferred from `CustomLink`. I guess we maybe have something wrong with the setup here if it's accepting `string`.

It's possible on this PR for a widget to link to `/app/someRandomPath/etc` - a non-existent path. Similarly we could update a route and this component would silently break.

As we're using Tanstack Router we can benefit from leaning on this for route type-safety. Ideally we'd accept a `link?: LinkOptions<RegisteredRouter>` here, passed in with `linkOptions()` for type-safety which would then enforce that only valid routes can get used and the required params must be provided.

It's possible that there's a fix for `CustomLink` we can apply which should basically force us to follow this pattern if the TS inference runs correctly.

```ts
import { linkOptions } from "@tanstack/react-router";

// Props pattern to aim for I think...!
  <DashboardWidget
    title="Flows"
    link={linkOptions({
      to: "/app/$team",
      params: { team: team.slug },
      text: "view all flows"
    })}
  >
```

Docs: https://tanstack.com/router/latest/docs/guide/link-options

DafyddLlyr (changes requested):
Please see comment on Tanstack Router implementation.

apps/editor.planx.uk/src/ui/editor/DashboardWidget/index.tsx
joecarver: ah yes nice spot, thanks for the guide, done here https://github.com/theopensystemslab/planx-new/pull/6723/commits/ffe17874dd2ddc4d8e7957344c1eb0bf0e2b3119

apps/editor.planx.uk/src/pages/Dashboard/index.tsx
joecarver: done https://github.com/theopensystemslab/planx-new/pull/6723/commits/6a7ccf0feb8aba91d2e8bbc32c535265e9489034

apps/editor.planx.uk/src/ui/editor/DashboardWidget/index.tsx
DafyddLlyr: Perfect...! 👌 

image

DafyddLlyr (approved):

PR #6724, opened 2026-05-26
fix: Resolve `InputField` bug (out of bounds index)

DafyddLlyr:
## What's the problem?
I noticed a new Airbrake error this morning, originating from Report A Planning Breach, specifically the `MapAndLabel` component which relied on `InputField` components under the hood.

Error - https://planx.airbrake.io/projects/329753/groups/4340288067798671310?tab=overview

image

## What's the cause?
This looks like a rendering issue / race condition between `setFeatures` and `formik.setFieldValue` meaning that `activeIndex` ends up out of bounds.

When all features are removed and then re-added, `activeIndex` (React state) is updated synchronously by `addFeatureToMap()`, while `schemaData` (Formik) is updated separately by `addFeatureToForm()`. A React render between these two updates sees an `activeIndex` that is ahead of `schemaData`, causing `schemaData[activeIndex]` to be `undefined`. Reading any field property then throws.

The fix adds an guard to `SchemaFields` so it renders nothing when `activeIndex` is out of bounds, and also correctly resets the index during `removeFeature()`.

## Testing...
Go to https://storybook.planx.uk/?path=/docs/planx-components-mapandlabel--docs

1. Add two items
2. Delete two items
3. Add four items
4. Delete fourth ❌ Error hit..!

There's a few ways to make this state update get hit but this seems the most recreatable.

https://github.com/user-attachments/assets/e2fc4def-889d-4003-a948-09ad0d233464

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/shared/Schema/SchemaFields.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/shared/Schema/SchemaFields.tsx
@@ -28,12 +28,16 @@ export const SchemaFields: React.FC<SchemaFieldsProps> = ({
   formik,
   sx,
   activeIndex = 0,
-}) => (
-  <Box sx={sx}>
-    {schema.fields.map((field, i) => (
-      <InputRow key={i}>
-        <InputFields {...field} activeIndex={activeIndex} formik={formik} />
-      </InputRow>
-    ))}
-  </Box>
-);
+}) => {
+  if (!formik.values.schemaData[activeIndex]) return null;
+
+  return (
+    <Box sx={sx}>
+      {schema.fields.map((field, i) => (
+        <InputRow key={i}>
+          <InputFields {...field} activeIndex={activeIndex} formik={formik} />
+        </InputRow>
+      ))}
+    </Box>
+  );
+};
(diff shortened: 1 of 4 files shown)


apps/editor.planx.uk/src/@planx/components/shared/Schema/SchemaFields.tsx
```diff
@@ -28,12 +28,16 @@ export const SchemaFields: React.FC<SchemaFieldsProps> = ({
...
-  <Box sx={sx}>
-    {schema.fields.map((field, i) => (
-      <InputRow key={i}>
-        <InputFields {...field} activeIndex={activeIndex} formik={formik} />
-      </InputRow>
-    ))}
-  </Box>
-);
+}) => {
+  if (!formik.values.schemaData[activeIndex]) return null;
```
DafyddLlyr, line 32: Additional guard clause - this shouldn't really be hit..!

DafyddLlyr:
For some reason CI is't running on this branch... 🤔 

Update: Looks like it's a GitHub issue again - https://www.githubstatus.com/

apps/localplanning.services/pnpm-lock.yaml
DafyddLlyr, line 1: Seems to have been an issue with https://github.com/theopensystemslab/planx-new/pull/6719 - CI is hitting issues without a `pnpm i` here.

joecarver (approved):
all looks good and thanks for the thorough write up + new tests

PR #6733, opened 2026-05-26
feat: dashboard stats panel

joecarver:
### What's in this PR?
- Implements the dashboard stats panel. Uses a DB view to provide easy querying through hasura. Will check the query time for this on some prod-scale data, if its at all slow then we can switch to a materialized view instead.


apps/editor.planx.uk/src/pages/Dashboard/components/useTeamDashboardStats.ts
```diff
@@ -0,0 +1,34 @@
...
+      sessions_previous
+      submissions_current
+      submissions_previous
+      guidance_sessions_current
+      guidance_sessions_previous
+      online_flows
+      online_flows_previous
+    }
+  }
+`;
```
DafyddLlyr, line 29: nit: Please destructure these to camelCase so we can keep out TS interfaces consistently formatted in camelCase.

```gql
const GET_TEAM_DASHBOARD_STATS = gql`
  query GetTeamDashboardStats($teamSlug: String!) {
    teamDashboardStats: team_dashboard_stats(where: { team_slug: { _eq: $teamSlug } }) {
      sessionsCurrent: sessions_current
      sessionsPrevious: sessions_previous
      ...
```

apps/hasura.planx.uk/metadata/databases/default/tables/public_team_dashboard_stats.yaml
DafyddLlyr, line 1: When creating a new table, please also add introspection tests in `apps/hasura.planx.uk/tests`

apps/editor.planx.uk/src/pages/Dashboard/components/StatsBanner.tsx
```diff
@@ -40,27 +36,100 @@ const StatsGrid = styled(Box)(({ theme }) => ({
...
+          sx={{
+            display: "flex",
+            alignItems: "center",
+            justifyContent: "center",
+            minHeight: 72,
+          }}
+        >
+          <DelayedLoadingIndicator inline msDelayBeforeVisible={300} />
+        </Box>
+      ) : (
```
DafyddLlyr: nit(style): I'd suggest we put a bit more work into this loading layout as it's going to have a pretty prominent position on the page - maybe we keep the labels and just us a loader for the value (maybe just a pulse animation on the `—` placeholder?). (cc @ianjon3s)

apps/editor.planx.uk/src/pages/Dashboard/components/StatsBanner.stories.tsx
DafyddLlyr, line 1: Super helpful...!

apps/editor.planx.uk/src/pages/Dashboard/components/StatsBanner.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+    stats: {
+      online_flows: 12,
+      online_flows_previous: 10,
+      sessions_current: 1840,
+      sessions_previous: 1523,
+      submissions_current: 304,
+      submissions_previous: 318,
+      guidance_sessions_current: 97,
+      guidance_sessions_previous: 84,
+    },
```
DafyddLlyr, line 20: Please see above comment on camelCase vs snakeCase

apps/editor.planx.uk/src/pages/Dashboard/components/StatsBanner.tsx
```diff
@@ -40,27 +36,100 @@ const StatsGrid = styled(Box)(({ theme }) => ({
...
+      delta: stats
+        ? stats.submissions_current - stats.submissions_previous
+        : null,
+    },
+    {
+      label: "Total guidance sessions",
+      value: stats?.guidance_sessions_current ?? null,
+      delta: stats
+        ? stats.guidance_sessions_current - stats.guidance_sessions_previous
+        : null,
```
DafyddLlyr, line 84: This conditional handling of `stats` multiple times throughout this list and the wider component is a bit of a sign that our data-fetching is spilling out into our presentation layer.

I'd recommend - 
 - Handling data fetching purely in the parent `ConnectedStatsBanner`
 - This would mean `stats` would always be defined in `StatsBanner` - no conditionality

```ts
export default function ConnectedStatsBanner() {
  const teamSlug = useStore((state) => state.getTeam().slug);
  const { data, loading } = useTeamDashboardStats(teamSlug);
  
  const stats = data?.team_dashboard_stats?.[0];

  // Handle loading state at the container level
  if (loading) {
    return (
      <Root>
        <Box sx={{ display: "flex", alignItems: "center", justifyContent: "center", minHeight: 140 }}>
          <DelayedLoadingIndicator inline msDelayBeforeVisible={300} />
        </Box>
      </Root>
    );
  }

  // Handle empty or error states safely 
  if (!stats) {
    return null;
  }

  // Guaranteed to have data here
  return <StatsBanner teamSlug={teamSlug} stats={stats} />;
}
```

apps/hasura.planx.uk/migrations/default/1779792000000_create_team_dashboard_stats_view/up.sql
```diff
@@ -0,0 +1,58 @@
+CREATE OR REPLACE VIEW "public"."team_dashboard_stats" AS
+WITH team_sessions AS (
+  SELECT
+    f.team_id,
+    COUNT(DISTINCT CASE WHEN a.created_at >= now() - INTERVAL '30 days' THEN a.id END) AS sessions_current,
```
DafyddLlyr: Not sure without measuring how much of a performance difference this would make, but `id` fields are guaranteed to be unique so `DISTINCT` is not really doing much here.

```sql
COUNT(CASE WHEN a.created_at >= now() - INTERVAL '30 days' THEN 1 END) AS sessions_current,
```

apps/hasura.planx.uk/migrations/default/1779792000000_create_team_dashboard_stats_view/up.sql
```diff
@@ -0,0 +1,58 @@
...
+    COUNT(DISTINCT CASE WHEN a.created_at >= now() - INTERVAL '60 days' AND a.created_at < now() - INTERVAL '30 days' THEN a.id END) AS sessions_previous,
+    COUNT(DISTINCT CASE WHEN a.created_at >= now() - INTERVAL '30 days' AND f.category = 'guidance' THEN a.id END) AS guidance_sessions_current,
+    COUNT(DISTINCT CASE WHEN a.created_at >= now() - INTERVAL '60 days' AND a.created_at < now() - INTERVAL '30 days' AND f.category = 'guidance' THEN a.id END) AS guidance_sessions_previous
+  FROM flows f
+  JOIN analytics a ON a.flow_id = f.id
+  WHERE f.deleted_at IS NULL
+  GROUP BY f.team_id
```
DafyddLlyr: `analytics` is huge (too big..!). We should also add a backstop here to just narrow down to the last 60 days (same applies for `team_submissions`) below.

```sql
WHERE a.created_at >= now() - INTERVAL '60 days'
```

apps/hasura.planx.uk/migrations/default/1779792000000_create_team_dashboard_stats_view/up.sql
```diff
@@ -0,0 +1,58 @@
...
+    COUNT(DISTINCT CASE WHEN ls.submitted_at >= now() - INTERVAL '60 days' AND ls.submitted_at < now() - INTERVAL '30 days' THEN ls.id END) AS submissions_previous
+  FROM flows f
+  JOIN lowcal_sessions ls ON ls.flow_id = f.id
+  WHERE f.deleted_at IS NULL AND ls.submitted_at IS NOT NULL
+  GROUP BY f.team_id
+),
+online_flows_current AS (
+  SELECT team_id, COUNT(*)
+  FROM flows
+  WHERE status = 'online' AND deleted_at IS NULL
```
DafyddLlyr: Another possible optimisation - we could just handle these clauses once up front (`WITH filtered_flows AS ...`)

apps/hasura.planx.uk/migrations/default/1779792000000_create_team_dashboard_stats_view/up.sql
DafyddLlyr, line 1: I suspect this will be pretty slow on production due to the `analytics` table. You should be able to test this though via the Hasura console (using a version of this query with just a `SELECT` and/or `EXPLAIN ANALYZE`). 

For similar views for Metabase we've opted for a materialized view + pg_cron to refresh nightly.

apps/hasura.planx.uk/metadata/databases/default/tables/public_team_dashboard_stats.yaml
```diff
@@ -0,0 +1,51 @@
...
+        - online_flows
+        - online_flows_previous
+      filter:
+        team:
+          members:
+            _and:
+              - user_id:
+                  _eq: x-hasura-user-id
+              - role:
+                  _eq: teamEditor
```
DafyddLlyr, line 50: Have we ever discussed permissions for this dashboard page? 

I guess it does make sense to restrict these stats to `teamEditors` only, but we'd need an approach for this entire page - e.g. a route guard to block access for non-teamEditors (for the given team).

DafyddLlyr (changes requested):
A few small comments below.

Looking solid though - appreciate the thought and work that's gone into this. Easy to read and follow 👍

apps/editor.planx.uk/src/pages/Dashboard/components/useTeamDashboardStats.ts
joecarver: done https://github.com/theopensystemslab/planx-new/pull/6733/commits/b0ba309a3cb98db930ddcc5121cb72a10adbf8e2

apps/hasura.planx.uk/metadata/databases/default/tables/public_team_dashboard_stats.yaml
joecarver: done https://github.com/theopensystemslab/planx-new/pull/6733/commits/6892c10912fd3b9f631fe23130f7f89664dc889c

apps/hasura.planx.uk/migrations/default/1779792000000_create_team_dashboard_stats_view/up.sql
joecarver: done https://github.com/theopensystemslab/planx-new/pull/6733/commits/eb329ccff881519c1df6573c1e1b4d46e9c25014

apps/hasura.planx.uk/migrations/default/1779792000000_create_team_dashboard_stats_view/up.sql
joecarver: done https://github.com/theopensystemslab/planx-new/pull/6733/commits/54102b5a498730909875bd8e84106405b59e8c2e

apps/hasura.planx.uk/migrations/default/1779792000000_create_team_dashboard_stats_view/up.sql
joecarver: done https://github.com/theopensystemslab/planx-new/pull/6733/commits/7704133196f90738ae4e334b661199e14ee9d394

apps/hasura.planx.uk/migrations/default/1779792000000_create_team_dashboard_stats_view/up.sql
joecarver: Have run this against prod on Hasura and it's completing fairly quickly (176.080 ms) - so maybe no need for the materialized view just yet?

I've attached the EXPLAIN ANALYZE output here and it looks like the main cost is scanning and sorting analytics as you suggested. We can cut this down with a composite index on (created_at, flow_id) if we want to cut this down & safeguard us as the table grows. 

Or / (and?) if we want to materialize the data, its a fast query so we can probably run the refresh very often  - every minute or so. What do you think?

[team_dashboard_stats_prod_explain_analyze.rtf](https://github.com/user-attachments/files/28340927/team_dashboard_stats_prod_explain_analyze.rtf)

apps/hasura.planx.uk/metadata/databases/default/tables/public_team_dashboard_stats.yaml
joecarver: ah yeah good point - is there any reason why teamViewers shouldn't be able to see them though? They can see analytics so surely this is similar? I can imagine that other panels should be hidden for viewers but stats seems fine to me. Happy to hide it if we prefer though!

apps/editor.planx.uk/src/pages/Dashboard/components/StatsBanner.tsx
joecarver: done https://github.com/theopensystemslab/planx-new/pull/6733/commits/e08714997f15e5a57f442e8afecc8ede6396e3d2

apps/hasura.planx.uk/migrations/default/1779792000000_create_team_dashboard_stats_view/up.sql
DafyddLlyr: Thanks for checking this out..! I'd vote to not prematurely optimise here - 176ms is significantly better than I was expecting. If we find there's an issue here in future we can revisit 👍

apps/hasura.planx.uk/metadata/databases/default/tables/public_team_dashboard_stats.yaml
DafyddLlyr: There's no real reason to hide these stats (there's no "security" reason to hide them). The simplest solution might be to keep these permissions as-is and route-guard the `/explore` page (it's not really intended for outside consumers?).

Let's keep this as-is for now and cross check on Slack with wider team, and/or next design call 👍

apps/editor.planx.uk/src/pages/Dashboard/components/StatsBanner.tsx
DafyddLlyr: Nice..!

DafyddLlyr (approved):

PR #6735, opened 2026-05-27
feat: dashboard notifications panel

joecarver:
### What's in this PR?
Adds the 'notifications' panel for the dashboard page. Fairly straightforward as we already have this data displayed in a very similar format.

This one doesn't match the designs exactly, as I re-used the whole `NotificationCard` component that's already being used and is very close. The only difference I could spot is the button styling, where there is a separate button rather than just the whole card being clickable. For now I think it's fine to keep as it, when we introduce more notification types that have multiple actions associated with them, we will want to re-design this card component both for the sidebar and the panel.


ianjon3s (approved):
Working great! The difference from designs is due to an update to the notifications card after the dashboard design was put together, this brings it in nicely.

DafyddLlyr (approved):

PR #6738, opened 2026-05-27
feat: dashboard feedback panel

joecarver:
### What's in this PR?
Adds the 'feedback' panel for the dashboard page.

I looked into making the list items link to specific feedback entries on the feedback page but I don't think its possible as the route just renders a table, and the table filters do not persist in the URL. 

Decided to write a new query rather than using the existing one on the feedback page because there's a lot less information required for this table and we'll want it to load as fast as possible, being a landing page summary.


ianjon3s (approved):
Approving from a visual perspective, looks great in Storybook!

DafyddLlyr (approved):
Great...!

PR #6740, opened 2026-05-27
feat: dashboard activity panel

joecarver:
### What's in this PR?
Adds the 'activity' panel for the dashboard page. Needed a new gql query & hook, and needed to allow aggregations on lowcal sessions in order to fetch it, so you'll need to make sure metadata is re-applied to test this locally.

### DIFF
--- a/apps/editor.planx.uk/src/hooks/data/useActivityData.ts
+++ b/apps/editor.planx.uk/src/hooks/data/useActivityData.ts
@@ -1,0 +1,82 @@
+import { gql, useQuery } from "@apollo/client";
+import { useStore } from "pages/FlowEditor/lib/store";
+
+const GET_ACTIVITY_BY_SERVICE = gql`
+  query GetActivityByService($teamId: Int!) {
+    flows(
+      where: { team_id: { _eq: $teamId }, deleted_at: { _is_null: true } }
+    ) {
+      name
+      sessions: lowcal_sessions_aggregate(
+        where: { deleted_at: { _is_null: true } }
+      ) {
+        aggregate {
+          count
+        }
+      }
+      submissions: lowcal_sessions_aggregate(
+        where: {
+          deleted_at: { _is_null: true }
+          submitted_at: { _is_null: false }
+        }
+      ) {
+        aggregate {
+          count
+        }
+      }
+    }
+  }
+`;
+
+type FlowActivity = {
+  name: string;
+  sessions: { aggregate: { count: number } | null };
+  submissions: { aggregate: { count: number } | null };
+};
+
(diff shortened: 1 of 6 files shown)


ianjon3s (approved):
Very smart, seeing it in action I think we can lose the subtle border between each row (graph line does a good enough job of dividing), otherwise this looks good to go.

apps/editor.planx.uk/src/hooks/data/useActivityData.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const flows = data?.flows ?? [];
+
+  const sessions = flows
+    .map(({ name, sessions }) => ({
+      name,
+      count: sessions.aggregate?.count ?? 0,
+    }))
+    .sort((a, b) => b.count - a.count)
+    .slice(0, TOP_LIMIT);
```
DafyddLlyr, line 71: It's _probably_ possible to offload these to the GQL query - sorting by count and adding a limit clause.

```gql
# Not tested...!
order_by: { lowcal_sessions_aggregate: { count: desc } }
limit: $limit 
```

DafyddLlyr (approved):
This is really solid - well structured, clear, and nicely sized..!

One small comment on shifting sorting and ordering to the query, maybe worth trying out - doubt there's a huge benefit to the switch.

apps/editor.planx.uk/src/hooks/data/useActivityData.ts
joecarver: OK so i've tested this against a bunch of test data locally, and you don't actually get the same result when doing the `order_by` at the GQL level - because the `order_by` is applied to all the rows in the table, it does not take into account the `where` clause in the query. As far as I can tell there is no way to pass the `where` clause into the `order_by` statement.

So for submissions, if you had loads of total sessions for a service but only a few of them were submitted, the service would rank higher than it should in the search, because its ordering by total sessions not by submitted sessions. 

And of course the limit has to be applied at the same level as the sort for either to be meaningful.

So I think the in-code approach is the only way to get what we want... unless I'm missing something in my limited GQL / Hasura knowledge!

joecarver: We could also do it with a Postgres view of course if we wanted to make sure this is all delegated to the DB layer

DafyddLlyr: Thanks for trying this out and sorry for sending you down a dead end...!

Very happy for this to merge as-is, and we can look at implementing a view at a later date if we find it's necessary 👍

DafyddLlyr (approved):

PR #6750, opened 2026-05-28
fix(editor): Fallback to relative links within Editor sidebar and Storybook

DafyddLlyr:
## What's the problem?
Components that rely on the `usePublicContext()` hook throw an error anywhere outside the "public" context (`/_planXDomain` or `/_customSubdomain` routes). This means certain components cannot render within the Editor sidebar or Storybook

Exaple here - https://storybook.planx.uk/?path=/docs/planx-components-feedback--docs

The `Feedback` component just throws a `notFound()` error, and does not render.

Seen here in Airbrake - https://planx.airbrake.io/projects/329753/groups/4341880470066273589?tab=notice-detail

## What's the solution?
Degrate links in these contexts to being just relative ones, by passing an `undefined` `from` argument. This will mean links will be broken within the Editor sidebar and Storybook, but it's not a public-facing user error and is an improvement on the current situation.

## Context
This came up when working on https://github.com/theopensystemslab/planx-new/pull/6739 (now reverted). Once this PR has been merged I'll take another look at the above PR and move forward with it.

### DIFF
--- a/apps/editor.planx.uk/src/pages/Preview/ContentPage.tsx
+++ b/apps/editor.planx.uk/src/pages/Preview/ContentPage.tsx
@@ -4,15 +4,8 @@ import Container from "@mui/material/Container";
 import IconButton from "@mui/material/IconButton";
 import { styled } from "@mui/material/styles";
 import Typography from "@mui/material/Typography";
-import {
-  notFound,
-  useLocation,
-  useNavigate,
-  useRouter,
-} from "@tanstack/react-router";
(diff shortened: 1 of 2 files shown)


apps/editor.planx.uk/src/pages/Preview/ContentPage.tsx
DafyddLlyr, line 1: Just tidying up unused imports here.

jessicamcinchak (approved):

PR #6759, opened 2026-05-28
feat: show placeholder card for nested flows in editor Preview tab

jessicamcinchak:
See https://opensystemslab.slack.com/archives/C01E3AC0C03/p1779994695866529

Known issues: 
- Not currently setting a breadcrumb `onSubmit` (do we need it to? how do folders handle this?)
- Missing `wasVisited` green border (relies on breadcrumb? or use another method of setting?)
  - This is actually consistent with folders
- Flow ID obviously not friendly, should we already fetch and display as `team-slug/flow-slug` ?


jessicamcinchak:
Not progressing this approach !

PR #6763, opened 2026-05-29
fix: replace existing help text with blank on paste 

augustlindemer:


jessicamcinchak (approved):
Perfect! Thanks for picking this one up!

PR #6764, opened 2026-05-29
fix: persist templated node props when flattening nested flows

jessicamcinchak:
See bug report here: https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1780055732310229

The FOI source template was recently updated and there are two new required customisations - specifically nested flow / external portal type nodes. 

The graph's "Customise" sidebar tab immediately recognised these as outstanding because the nodes in the live flow `data` have templated node props. 

But the "Publish" modal was failing identify these and throw a validation error. When nested flow node types are _flattened_ on preview (which is the data shape publish checks read from), we were failing to persist these templated node props. 

This has always been overlooked/missed since templates introduced from what I can tell, not a recent regression from `dataMerged` refactor.

**Testing:** 
- Open publish dialog on the following link and confirm that there is now a failing publish validation check: https://6764.planx.pizza/app/barnet/find-out-if-you-need-planning-permission (for same flow on prod/staging all checks are passing)
Screenshot from 2026-05-29 09-04-41

### DIFF
--- a/apps/api.planx.uk/shared/dataMerged.test.ts
+++ b/apps/api.planx.uk/shared/dataMerged.test.ts
@@ -418,6 +418,8 @@ const flattenedParentFlow: FlowGraph = {
     edges: ["nested-a"],
     data: {
       flattenedFromExternalPortal: true,
+      isTemplatedNode: false,
+      areTemplatedNodeInstructionsRequired: false,
     },
   },
   "nested-a": {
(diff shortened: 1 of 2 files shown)


DafyddLlyr (approved):
Thanks for quick fix and detailed write up...! 🏅 

Getting expected results on Pizza vs Staging ✅

PR #6766, opened 2026-05-29
fix: polyfill `at` for ITP users on old browsers

jessicamcinchak:
See https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1780063916845819

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
@@ -1,5 +1,3 @@
-import "core-js/stable/array/at"; // at polyfill
-
 import ErrorOutline from "@mui/icons-material/ErrorOutlined";
 import Button from "@mui/material/Button";
 import { styled } from "@mui/material/styles";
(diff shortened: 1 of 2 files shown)


DafyddLlyr (approved):

apps/editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
```diff
@@ -1,3 +1,5 @@
+import "core-js/stable/array/at"; // at polyfill
```
freemvmt: nice! one thing - I think let's group it with the other `core-js` import in `src/index.tsx`, since there are other examples of `.at()` in the codebase that could otherwise not be caught by this?

freemvmt (approved):

apps/editor.planx.uk/src/@planx/components/Pay/Public/InviteToPayForm.tsx
jessicamcinchak: Yep good shout - thanks for closer attention to detail here than I had :sweat_smile:

jessicamcinchak: Updated in latest commit! Based on other instances of `.at`, we also need to polyfill for _both_ string & array !

freemvmt: @jessicamcinchak fab nice catch, LGTM!

PR #6776, opened 2026-06-02
feat: try rendering flattened flow in preview tab only

jessicamcinchak:
does _not_ work, just trying to get a few different options up for dev call chat/demo ~wednesday ! 

open questions: 
- do we need to instantiate + run separate/parallel stores if side tab is reading from different data than the graph? `<Questions />` is a gnarly component heavily coupled to many navigation state methods


jessicamcinchak:
Not progressing this approach !

PR #6777, opened 2026-06-02
chore: update copy in conditional settings

zz-hh-aa:
To specify 'service' when relevant.


e2e/tests/ui-driven/src/create-flow.spec.ts
```diff
@@ -143,6 +143,8 @@ test.describe("Flow creation, publish and preview", () => {
     await page.goto(`/${context.team.slug}/${serviceProps.slug}`);
     await navigateToFlowSettings(page);
     await makeFlowAService(page);
+    await page.reload();
```
zz-hh-aa, line 146: Tests were working fine before / didn't need to refresh the page because it was automatically refetching data--looking into why this is the case now but hoping to resolve in another PR if poss.

DafyddLlyr (approved):

PR #6779, opened 2026-06-02
chore(editor): Update vitest to v4

DafyddLlyr:
## What does this PR do?
Update Vitest from v3 to v4 within the Editor project, matching versions used elsewhere (Hasura, API)

## Motivation
 - Solves https://github.com/theopensystemslab/planx-new/security/dependabot/903 & https://github.com/theopensystemslab/planx-new/security/dependabot/904
 - Ensures we can use [a single catalog version](https://github.com/theopensystemslab/planx-new/pull/6573) of vitest in future
 - Background task during strategy days calls...!

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/MapAndLabel/test/public.backNavigation.test.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/MapAndLabel/test/public.backNavigation.test.tsx
@@ -16,11 +16,13 @@ beforeAll(() => {
     window.customElements.define("my-map", MyMap);
   }
 
-  const ResizeObserverMock = vi.fn(() => ({
-    observe: vi.fn(),
-    unobserve: vi.fn(),
-    disconnect: vi.fn(),
-  }));
+  const ResizeObserverMock = vi.fn(function () {
+    return {
+      observe: vi.fn(),
+      unobserve: vi.fn(),
+      disconnect: vi.fn(),
+    };
+  });
 
   vi.stubGlobal("ResizeObserver", ResizeObserverMock);
 });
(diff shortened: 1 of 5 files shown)


apps/editor.planx.uk/src/@planx/components/MapAndLabel/test/public.backNavigation.test.tsx
```diff
@@ -16,11 +16,13 @@ beforeAll(() => {
...
-    unobserve: vi.fn(),
-    disconnect: vi.fn(),
-  }));
+  const ResizeObserverMock = vi.fn(function () {
+    return {
+      observe: vi.fn(),
+      unobserve: vi.fn(),
+      disconnect: vi.fn(),
+    };
+  });
```
DafyddLlyr, line 25: > _Note that now if you provide an arrow function, you will get [<anonymous> is not a constructor error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Not_a_constructor) when the mock is called._

Docs: https://vitest.dev/guide/migration.html#spyon-and-fn-support-constructors

apps/editor.planx.uk/vitest.config.ts
```diff
@@ -1,10 +1,11 @@
-import { defineConfig } from "vite";
 import tsconfigPaths from "vite-tsconfig-paths";
+import { defineConfig } from "vitest/config";
 
 export default defineConfig({
   test: {
     environment: "jsdom",
     globals: true,
+    clearMocks: true,
```
DafyddLlyr, line 8: Docs: https://vitest.dev/guide/migration.html#changes-to-mocking

freemvmt: also useful https://vitest.dev/config/clearmocks

freemvmt (approved):
nice 👌🏼

PR #6799, opened 2026-06-02
feat: auto-suggest editor data values for `property.localAuthorityDistrict`

jessicamcinchak:
Was mentioned during strat days updates that case-sensitivity of passport data fields still a big gotcha / editor pain point. 

`property.localAuthorityDistrict` is one that _every_ council needs to customise and is particularly odd casing ! 

We can fetch a list of valid values to suggest from https://www.planning.data.gov.uk/entity.json?limit=400&dataset=local-authority-district&exclude_field=geometry,point&entries=all

Screenshot from 2026-06-02 14-26-10

Because all our auto-suggests allow overwriting, templated node instructions & defaults are un-altered:
image

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/shared/BaseChecklist/Editor/components/Options/ExclusiveOrOptionManager.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/shared/BaseChecklist/Editor/components/Options/ExclusiveOrOptionManager.tsx
@@ -7,6 +7,7 @@ import {
   AnyChecklist,
   OptionGroup,
 } from "@planx/components/shared/BaseChecklist/model";
+import { useLocalAuthorityDistricts } from "@planx/components/shared/hooks";
 import { getOptionsSchemaByFn } from "@planx/components/shared/utils";
 import { getIn } from "formik";
 import React from "react";
@@ -39,6 +40,7 @@ export const ExclusiveOrOptionManager = <T extends AnyChecklist>({
   isTemplatedNode,
 }: Props<T>) => {
   const { schema, currentOptionVals } = useCurrentOptions(formik);
+  const { data: localAuthorityDistrictSchema } = useLocalAuthorityDistricts();
 
   return (
     <Box>
@@ -103,6 +105,7 @@ export const ExclusiveOrOptionManager = <T extends AnyChecklist>({
                 formik.values.fn,
                 schema,
                 currentOptionVals,
+                localAuthorityDistrictSchema,
               ),
             }}
             isTemplatedNode={isTemplatedNode}
--- a/apps/editor.planx.uk/src/@planx/components/shared/hooks.ts
+++ b/apps/editor.planx.uk/src/@planx/components/shared/hooks.ts
@@ -1,3 +1,5 @@
+import { useQuery } from "@tanstack/react-query";
+import { getLocalAuthorityDistricts } from "lib/planningData/requests";
 import { useEffect, useState } from "react";
 
 export type UseFileUrlProps =
@@ -30,4 +32,15 @@ export const useFileUrl = (props: UseFileUrlProps) => {
   return {
     fileUrl,
   };
+};
+
+export const useLocalAuthorityDistricts = () => {
+  const query = useQuery({
+    queryKey: ["localAuthorityDistrictNames"],
+    queryFn: () => getLocalAuthorityDistricts(),
+    refetchOnWindowFocus: false,
+    staleTime: Infinity,
+  });
+
+  return query;
 };
--- a/apps/editor.planx.uk/src/@planx/components/shared/utils.ts
+++ b/apps/editor.planx.uk/src/@planx/components/shared/utils.ts
@@ -69,6 +69,7 @@ export const getOptionsSchemaByFn = (
   fn?: string,
   defaultOptionsSchema?: string[],
   currentOptions?: (string | undefined)[],
+  localAuthorityDistrictSchema?: string[] | undefined,
 ) => {
   let schema = defaultOptionsSchema;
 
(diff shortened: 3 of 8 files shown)


apps/editor.planx.uk/src/@planx/components/shared/BaseChecklist/Editor/components/Options/ExclusiveOrOptionManager.tsx
```diff
@@ -39,6 +40,7 @@ export const ExclusiveOrOptionManager = <T extends AnyChecklist>({
   isTemplatedNode,
 }: Props<T>) => {
   const { schema, currentOptionVals } = useCurrentOptions(formik);
+  const { data: localAuthorityDistrictSchema } = useLocalAuthorityDistricts();
```
jessicamcinchak, line 43: Because I'm fetching local authority districts via a query hook, I can't call it directly inside of `getOptionsSchemaByFn` and have instead hoisted 4x to (responsive)question/checklist components. 

The query will cache after first run, so I'm not particulary worried about performance here - but curious if there's a neater way of handling that I'm missing?

apps/editor.planx.uk/src/@planx/components/shared/utils.ts
```diff
@@ -79,6 +80,10 @@ export const getOptionsSchemaByFn = (
     schema = getValidSchemaValues("ProjectType");
   if (fn === "property.type") schema = getValidSchemaValues("PropertyType");
 
+  // For other data fields, suggest based on external data source like Planning Data
+  if (fn === "property.localAuthorityDistrict" && localAuthorityDistrictSchema)
+    schema = localAuthorityDistrictSchema;
```
jessicamcinchak, line 85: Other potential future candidates for this same pattern: 
| fn  | Data source  | 
|---|---|
|`property.localPlanningAuthority` | https://www.planning.data.gov.uk/entity.json?limit=400&dataset=local-planning-authority&exclude_field=geometry,point  |
|`property.region`  | https://www.planning.data.gov.uk/entity.json?limit=400&dataset=region&exclude_field=geometry,point  | 
|`property.ward`  | https://www.planning.data.gov.uk/entity.json?limit=400&dataset=ward&exclude_field=geometry,point  |

Maybe a few others I'm forgetting ? 

Have started with `property.localAuthorityDistrict` as this is by far most prevalent as a customisable node across all templated services, etc.

apps/editor.planx.uk/src/@planx/components/shared/hooks.ts
```diff
@@ -31,3 +33,14 @@ export const useFileUrl = (props: UseFileUrlProps) => {
     fileUrl,
   };
 };
+
+export const useLocalAuthorityDistricts = () => {
+  const query = useQuery({
+    queryKey: ["localAuthorityDistrictNames"],
+    queryFn: () => getLocalAuthorityDistricts(),
+    refetchOnWindowFocus: false,
+    staleTime: Infinity,
```
DafyddLlyr, line 42: Perfect..!

apps/editor.planx.uk/src/@planx/components/shared/BaseChecklist/Editor/components/Options/ExclusiveOrOptionManager.tsx
DafyddLlyr: Yeah this is a tricky one. As a one-off I probably would have gone for this approach also, but reading your comment below about other candidates it's probably worth planning for this (good!) outcome here and solving once.

How does something like this sound...
 - Rather than a specific `/lib/planningData` function for just LADs, we have something like `getPlanningDataEntityNames(dataset)` which can accomodate all future candidates
 - A shared hook for taking an `fn`, and querying Planning Data entities based on this (we'll need to map Passport vars -> dataset names, then call the above lib function)
  - If we key this using `enabled` on a matching dataset only, a query will never run if `fn` doesn't match (e.g. `food.italian` will never trigger a request, but `property.localAuthorityDistrict` will get `"local-authority-district"` results)
- We can then just pass in a single static argument into `getOptionsSchemaByFn()`, keeping it pure and synchronous - we can do this already via `defaultOptionsSchema`


We then end up with something like this - 

```ts
const { data: externalSchema } = usePlanningDataOptionsSchema(formik.values.fn);

...

schema: getOptionsSchemaByFn(
  formik.values.fn,
  externalSchema ?? schema,
  currentOptionVals,
)
```

Once done, adding new datasets is just a one-line case of adding the mapping from Passport variable to dataset name I think (e.g. `"property.region": "region"`).

Not sure how clear this is, happy to talk through to sense check it..!

DafyddLlyr (approved):
Happy for this to merge-as is - solves the problem..!

I've added an idea for how we can solve this in a bit more of an expansive way, capturing the other entities without too much effort. Maybe worth a look, or it can be a future follow-up.

apps/editor.planx.uk/src/@planx/components/shared/BaseChecklist/Editor/components/Options/ExclusiveOrOptionManager.tsx
jessicamcinchak: This does make sense I think, thanks for thoughts here ! 

Going to merge this one as-is, but per [this #planx-content thread](https://opensystemslab.slack.com/archives/C07F7215W7P/p1780426978798729), I'm going to pickup these other candidates sooner than later so will have a go at this shared hook in that PR :+1:

jessicamcinchak: Implemented here ! https://github.com/theopensystemslab/planx-new/pull/6814

PR #6803, opened 2026-06-03
chore: tidy flows is_service

zz-hh-aa:
August (and Mel for WikiHouse team) have confirmed which flows shouldn't be services. This migration sets `status = offline` and `is_service = false` for the list of flow IDs. 

In addition to the listed flows, anything in OSL or Testing teams (that is not a source template) should also have these values set. 

I've left the down migration empty but let me know if that should be filled in (I don't think we could work backwards for the OSL / Testing flows, however).

### DIFF
--- a/apps/hasura.planx.uk/migrations/default/1780479622910_tidy_flows_is_service/up.sql
+++ b/apps/hasura.planx.uk/migrations/default/1780479622910_tidy_flows_is_service/up.sql
@@ -1,0 +1,43 @@
+UPDATE flows
+SET status = 'offline',
+    is_service = FALSE
+WHERE flows.id IN (
+    'fcf3770c-6fdf-416f-be90-60522beb7d88',
+    '1a679d42-ab7a-43dc-9f57-607e394c1ae7',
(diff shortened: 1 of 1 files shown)


DafyddLlyr (commented):
I think it would just be simpler to run this as a one-off manually on prod as opposed to making it a migration.

Migrations are (generally!) structural, and we try to only cover data changes if required as part of that structural change (e.g. move data from one column to another or something).

As this is purely a data change, we can just run as a one-off on prod 👍 

We aren't super strict on this pattern, but we've been caught out a few times by trying to handle data updates this way (what happens if an ID is added between now and the prod deploy?).

zz-hh-aa:
As discussed in dev call, pros and cons to either option (updating data as migration or directly editing on prod). I think the conclusion is to go with migration for better visibility / history (this also gives me less anxiety :)) but always pair if editing on prod. 

In this case, if an ID is added in between now and prod deploy, it's not a big issue anyway (just having an extra flow miscategorised as a service--something we will manually handle fixes for in the future).

DafyddLlyr (approved):
Sorry for the slow return with an approval here to unblock this one...! ✅

PR #6805, opened 2026-06-03
feat: inherit or write isService value on flow creation

zz-hh-aa:
# What does this PR do? 

When a new flow is a copy or template, inherit the `flows.is_service` value.
When a new flow is created from scratch, write a `flows.is_service` value using an editor-facing switch.

Practically speaking, this means
- Update Hasura permissions (for selecting, inserting `flows.is_service` column)
- New `Switch` in `BaseFormSection` only shown when `values.mode === "new"`
- API writes `isService` value

# Testing

Test on a pizza and check `flows.is_service` value in Hasura. I have tested:
- Create a new flow from scratch that is just a flow
- Create a new flow from scratch that is a service
- Create a flow copied from a flow
- Create a flow copied from a service
- Create a templated flow that is a service
    - Because the get query here filters for `status = online`, and only services can be online, then source templates that are services can be used to create new ones

### DIFF
--- a/apps/api.planx.uk/helpers.ts
+++ b/apps/api.planx.uk/helpers.ts
@@ -13,6 +13,7 @@ export interface GetFlowDataResponse {
   data: Flow["data"];
   summary: string | null;
   description: string | null;
+  isService: boolean;
   limitations: string | null;
   team_id: number;
   team: { slug: string };
@@ -41,6 +42,7 @@ const getFlowData = async (id: string): Promise<GetFlowDataResponse> => {
           name
           summary
           description
+          isService: is_service
           limitations
           team_id
           team {
@@ -79,6 +81,7 @@ const createFlow = async ({
   summary,
   description,
   limitations,
+  isService,
 }: {
   teamId: number;
   slug: string;
@@ -90,6 +93,7 @@ const createFlow = async ({
   summary?: string;
   description?: string;
   limitations?: string;
+  isService?: boolean;
 }) => {
   const { client: $client } = getClient();
   const userId = userContext.getStore()?.user?.sub;
@@ -112,6 +116,7 @@ const createFlow = async ({
           $summary: String
           $description: String
           $limitations: String
+          $is_service: Boolean
         ) {
           insertFlow: insert_flows_one(
             object: {
@@ -126,6 +131,7 @@ const createFlow = async ({
               summary: $summary
               description: $description
               limitations: $limitations
+              is_service: $is_service
             }
           ) {
             id
@@ -143,6 +149,7 @@ const createFlow = async ({
         summary: summary,
         description: description,
         limitations: limitations,
+        is_service: isService,
       },
     );
 
--- a/apps/api.planx.uk/helpers.ts
+++ b/apps/api.planx.uk/helpers.ts
@@ -15,16 +15,16 @@ export interface GetFlowDataResponse {
   description: string | null;
   isService: boolean;
   limitations: string | null;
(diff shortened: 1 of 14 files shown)


apps/api.planx.uk/helpers.ts
```diff
@@ -13,6 +13,7 @@ export interface GetFlowDataResponse {
   data: Flow["data"];
   summary: string | null;
   description: string | null;
+  isService: boolean;
   limitations: string | null;
   team_id: number;
```
zz-hh-aa: There was a mix of snake and camel case in here--should we update all to camel?

apps/editor.planx.uk/src/pages/Team/components/AddFlow/BaseFormSection.tsx
```diff
@@ -67,16 +67,26 @@ export const BaseFormSection: React.FC = () => {
...
-            name="isTemplate"
-            checked={values.flow.isTemplate}
+            name="isService"
+            checked={values.flow.isService}
             onChange={() =>
-              setFieldValue("flow.isTemplate", !values.flow.isTemplate)
+              setFieldValue("flow.isService", !values.flow.isService)
             }
-            label={"Source template"}
+            label={"Is this a user-facing service?"}
```
zz-hh-aa: Platform admins will see two toggles here, which is maybe a little strange but I think is okay for now since it's only us. Other opinions / suggestions welcome though!

eg platform admin view:
Image

team editor / admin view:
Image

apps/api.planx.uk/helpers.ts
DafyddLlyr: Yep this is worth tidying up imho..!

apps/editor.planx.uk/src/pages/Team/components/AddFlow/index.tsx
```diff
@@ -45,6 +46,7 @@ export const AddFlow: React.FC = () => {
     values,
     { setFieldError, setStatus },
   ) => {
+    console.log({ values });
```
DafyddLlyr: 👀

apps/editor.planx.uk/src/pages/Team/components/AddFlow/BaseFormSection.tsx
DafyddLlyr: The only real thought here I have is that this is a pretty fundamental change for users creating new flows - it should maybe be more prominent / clear what this toggle is doing? 

I feel the "ON/OFF" toggle (whilst consistent) doesn't really serve us well here. We might need to change the layout somewhat and add a line of explainer here around this? cc @ianjon3s

DafyddLlyr (changes requested):
One small `console.log()` left in, and a UI question - not sure this is quite clear enough given the weight of this option.

apps/editor.planx.uk/src/pages/Team/components/AddFlow/BaseFormSection.tsx
ianjon3s: Agreed, something like the following would give more context
image

zz-hh-aa: Thanks both!! Done in [de3cce9](https://github.com/theopensystemslab/planx-new/pull/6805/commits/de3cce9e1e96955b444069ec9e346d31856dff6c)

image

DafyddLlyr (approved):

PR #6771, opened 2026-06-01
chore(editor): Upgrade to Storybook v10

DafyddLlyr:
## What does this PR do?
Upgrades the PlanX Storybook from v8 to v9 by following these migration guides - 
 - [Migration guide from Storybook 8.x to 9.1](https://storybook.js.org/docs/releases/migration-guide-from-older-version)
 - [Migration guide for Storybook 10](https://storybook.js.org/docs/releases/migration-guide)

### Notable changes
- Drop custom Tanstack router decorator, now working from `@storybook/tanstack-react` framework which has this built in
- Additional `devDependencies` to silence Storybook warnings

## Motivation
Worth doing before a11y audit, easy background task for strategy days calls, and will resolve https://github.com/theopensystemslab/planx-new/security/dependabot/893

PR #6806, opened 2026-06-03
feat(editor): Autofocus `TeamSelect` on open

DafyddLlyr:
Small issue which has bugged me for a while...! 😅 

When the `TeamSelect` is now opened, the search box is immediately focussed which allows the user to type right away.

Test it out here - https://storybook.6806.planx.pizza/?path=/docs/design-system-molecules-teamselect--docs

### DIFF
--- a/apps/editor.planx.uk/src/components/EditorNavMenu/components/TeamSelect.stories.tsx
+++ b/apps/editor.planx.uk/src/components/EditorNavMenu/components/TeamSelect.stories.tsx
@@ -1,7 +1,13 @@
 import type { Meta, StoryObj } from "@storybook/tanstack-react";
+import {
+  createMemoryHistory,
+  createRootRoute,
+  createRoute,
+  createRouter,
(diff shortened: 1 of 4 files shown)


apps/editor.planx.uk/src/components/EditorNavMenu/components/TeamSelect.stories.tsx
DafyddLlyr, line 1: Small Storybook fix, problem introduced here I believe - https://github.com/theopensystemslab/planx-new/pull/6771

apps/editor.planx.uk/src/hooks/useSearch.ts
DafyddLlyr, line 1: Another small fix, locally I hit an issue here occasionally which I think is a race condition possibly? Can't recreate it on staging or production.

Image

jessicamcinchak (approved):

PR #6807, opened 2026-06-03
feat: assign permissions to teamAdmin

zz-hh-aa:
# What does this PR do? 

Assigns `teamAdmin` permissions. `teamAdmin` is an _inherited role_ (from `teamEditor`, which means that when permissions are simple for a whole set (eg `select` on `flows`), the roles should have access parity. 

# Why? 

When `teamAdmin` has a defined permission, it should override `teamEditor` permissions. Since many tables had custom checks (eg where user is `teamEditor` for specific team), `teamAdmin` should have a matching set, specifying `teamAdmin`. That's what most of these permission changes are (adding a `teamAdmin` version to all `teamEditor` custom checks, which would fail when run for a `teamAdmin` because they wouldn't be a `teamEditor` for the team they're currently editing!). 

_NB: when we see a `teamAdmin` row in the Hasura Permissions tabs, this can be misleading—it will show all X’s when no custom permissions exist, but will have the functionality of a `teamEditor`._

# Testing
**As a `teamAdmin` I have tested:**

New functionality
- creating new members: both `teamEditor` and `teamAdmin`
    - updating member roles (`teamEditor` ↔ `teamAdmin`)

Existing functionality
- viewing flows
- publishing flows
- archive / unarchive flows
- pin / unpin flows
- copy flows

**As a `teamEditor` I have tested:**
- cannot provision a `teamAdmin`

Would recommend testing the same on Pizza! Hoping that tests would've caught anything here, but is there any other important functionality missing?

🚫 I have revoked all OSL platform admin privileges on the Pizza for role testing ease. 

# Next steps
- [ ] Migrate select `teamEditors` to `teamAdmin`


apps/hasura.planx.uk/metadata/databases/default/tables/public_submission_services_summary.yaml
```diff
@@ -53,6 +53,45 @@ select_permissions:
...
+        - user_clicked_save
+        - user_invited_to_pay
+        - user_role
+      filter:
+        team:
+          members:
+            _and:
+              - user_id:
+                  _eq: X-Hasura-User-Id
+              - id:
```
DafyddLlyr, line 92: Should this be `role:` and not `id:`?

DafyddLlyr (commented):
One small issue spotted..! (typo?)

This `filter:` issue is pretty awkward and in many ways removes the benefit of inherited permissions here unfortunately. Appreciate the write up and the extra work this leads to, shame I didn't land on it initially sorry.

Another option here that would work is just to drop the `team_members_user_id_team_id_key` constraints and allow multi-granting of both roles to the user. If they have both `teamAdmin` and `teamEditor` as available roles we can...
 - not inherit roles
 - `teamAdmin` is just a small footprint of additional responsibilities
 - keep `teamEditor` exactly as is
 - grant both as `x-hasura-allowed-roles` when creating a JWT

The cost of this approach is that..
- we will need some slight rework in a few places where a single role is checked
- we'd need to specify in the client which role to use for a given operation (setting the `x-hasura-role` header)
 - role granting and management is a little more complex (we want to abstract this from the user - they don't need to know that we're setting two roles under the hood)

The upside is much simpler role management going forward in Hasura, and simpler and smaller roles.

Happy to talk this one through...!

PR #6808, opened 2026-06-03
fix: Question options & Autocomplete 

joecarver:
### What's in this PR?
Two separate fixes but closely related, in response to [this Slack thread](https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1780482734978799)
1. Clear 'Data Field' values on question & checklist options when the data field value for the question is cleared
2. Always persist the entered option in `DataFieldAutocomplete` when the input field loses focus

Tests added for both as well.

I'm expecting that this will fix George's issue in which it was possible to Update a Question node, as the Question looked like it had a Data Field selected, but really it didn't, because it hadn't persisted when the input lost focus.


jessicamcinchak (approved):
Changes all look good here, and really appreciate extra test scenarios !

PR #6809, opened 2026-06-03
test(iac): Use Node test runner for Lambda@Edge tests

DafyddLlyr:
## What does this PR do?
 - Replaces custom test harness implementation for standard Node testing modules (e.g. `assert()`, `test()`)
 - Adds `package.json` script to kick off these tests
 - Does not auto-run as part of CI in any way - I don't think this is necessary given the footprint, I see these tests more as development aids

## Motivation
I'm currently investigating LPS downtime, which I've identified that it's an issue related to the deployment strategy. I'll give a full write-up in a follow up PR but essentially because of how stripping `.html` extensions from files is working I think we're clearing the bucket before writing new files - leading to downtime. My approach is to simplify the LPS build and just handle routing via a Lamba@Edge function in Cloudfront (similar to how `flow_link_preview.js` is working).

Before picking this up I wanted to improve the test harness currently in place - rather than a custom implementation we can just use in-build Node methods ([docs](https://nodejs.org/api/test.html)).


infrastructure/application/aws/lambda/README.md
```diff
@@ -51,13 +53,11 @@ Run the test script against a live Hasura instance. `HASURA_URL` defaults to pro
...
 # Run both default URL types (planx.uk domain + custom council domain)
-node test_flow_link_preview.js
+node --test flow_link_preview.test.js
 
-# Test a specific planx.{dev,uk} URL
-HASURA_URL=https://hasura.editor.planx.dev/v1/graphql node test_flow_link_preview.js https://editor.planx.dev/barnet/find-out-if-you-need-planning-permission/published
+# Alternatively, run the following from project root -
 
-# Test a specific custom domain URL
-node test_flow_link_preview.js https://planningservices.camden.gov.uk/find-out-if-you-need-planning-permission
```
DafyddLlyr, line 60: I've stripped this out as I think it's easy enough to add a custom test scenario for a given URL, rather than pass data in as an argument.

infrastructure/application/aws/lambda/flow_link_preview.test.js
```diff
@@ -0,0 +1,123 @@
+const { test, mock } = require("node:test");
+const assert = require("node:assert/strict");
+const { handler } = require("./flow_link_preview");
+
+const DEFAULT_HASURA_URL = "https://hasura.editor.planx.uk/v1/graphql";
```
jessicamcinchak: Default here needs to be production in order to test custom subdomain cases below, am I understanding that right? 

Feels counter-intertuitive to test against prod, is it maybe possible to default to `.dev` and reference the `planx.in` subdomains we've been testing ssl against recently?

jessicamcinchak (approved):
Thanks for working towards bottom of this LPS downtime issue - one question, not a blocker!

infrastructure/application/aws/lambda/flow_link_preview.test.js
DafyddLlyr: As we're mocking the global `fetch` anyway, I've actually just simplified this by removing the env var manipulation - it's never actually used.

PR #6810, opened 2026-06-04
feat: Feature flag toggles in global settings

ianjon3s:
## What does this PR do?

- Enables feature flags to be toggled in global settings
- Feature flags are added/removed from local storage
- Auto-refresh happens after saving so that the feature flag takes immediate effect

image

**Testing:**
https://6810.planx.pizza/app/global-settings/feature-flags


DafyddLlyr (approved):
Great..! 

Sometimes we have 0 available flags in production - we should also properly account for this UI state. We should also maybe make "experimental" a little clearer maybe - a curious user could toggle one of these, break something, and have no UI feedback that's it's linked to this.

ianjon3s:
Thanks for this @DafyddLlyr, I've updated the following:
- No feature flag state now uses `EmptyState` component
- Feature flags are present as a menu item to indicate how many flags are currently applied

As feature flags are currently only visible to PlatformAdmins I think this is a worthwhile addition without being too noisy

DafyddLlyr (approved):
Ah yep - missed the `platformAdmin` guard on this page sorry 👍 

Great...!

PR #6811, opened 2026-06-04
feat: filter flow search by isService too

zz-hh-aa:
Hides chips when a flow is not a service
Screenshot 2026-06-05 093031

Does this in the editor breadcrumbs also
eg, when a flow is 'just' a flow:
Screenshot 2026-06-04 151142

vs when a flow is a service:
Screenshot 2026-06-04 154029

Filter functionality all updated too!


apps/editor.planx.uk/src/pages/Team/components/FlowTable/index.tsx
```diff
@@ -179,16 +180,18 @@ const FlowTableRow: React.FC<FlowTableRowProps> = ({
...
+                  statusVariant={statusVariant}
+                >
+                  {statusVariant}
+                </FlowTag>
+              )}
             </Box>
           </FlowStatusCell>
           <FlowStatusCell>
-            {isSubmissionService && (
+            {isSubmissionService && isService && (
```
DafyddLlyr: I know these two values aren't truly linked currently but this reads as quite strange - it seems like the second condition is redundant. Maybe we should take `isService` into account when writing `isSubmissionService`, or consider a different naming pattern?

DafyddLlyr (approved):
Looking good! Works as expected, appreciate the test coverage.

One architectural question here around fetching flow data, please see below.

apps/editor.planx.uk/src/pages/Team/components/FlowTable/index.tsx
zz-hh-aa: Good point! Since it's reading the value from `has_send_component` anyway I think just renaming to `hasSendComponent` is simpler for now (done in [210b719](https://github.com/theopensystemslab/planx-new/pull/6811/commits/210b719ab2f69dd724e062e92a04a67ee61f7639))

DafyddLlyr: This is a good solution!

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -243,6 +243,7 @@ export interface FlowSummary {
   templatedFrom: string | null;
   isTemplate: boolean;
   isListedOnLPS: boolean;
+  isService: boolean;
```
DafyddLlyr, line 246: Flagging this one against [ADR#11](https://github.com/theopensystemslab/planx-new/blob/main/doc/architecture/decisions/0011-standardised-data-fetching-and-state-management.md) - `isService` is server-derived flow metadata, and the ADR asks us to keep new server state out of Zustand and fetch it via Apollo instead.

I certainly see why you've gone for this pattern and put this here though, `flowStatus`, `isFlowPublished` etc already live in the store, so adding it seems to keep the component internally consistent.

The ADR is approved and agreed, so ideally `Breadcrumbs` / `OpenServiceMenu` would pull `isService` straight from the Apollo cache via a `useQuery`, with no Zusantand field at all. The catch here is you'd then be getting `isService` one way and other fields another way in the same component, which is probably messier and more complex here.

My suggestion here would be to have a pass at getting these fields altogether in a single hook, and out of the store. No strong feelings if this is done after this PR (I'll still approve this one), or before / in-parallel to this PR.

DafyddLlyr (commented):
Hey sorry - just realised I never hit submit on this comment I was referencing in my previous review @zz-hh-aa 🤦

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
zz-hh-aa: Thanks for review & catching this! Makes sense--refactor trying this out is in [e3a9f77](https://github.com/theopensystemslab/planx-new/pull/6811/commits/e3a9f7717a63d48af55e96f268353cba30aed76e)

apps/editor.planx.uk/src/components/Breadcrumbs.tsx
```diff
@@ -42,15 +44,18 @@ const BreadcrumbsLink = styled(CustomLink)(({ theme }) => ({
...
   const isStandalone = useStore(
     (state) => state.previewEnvironment === "standalone",
   );
-  const flowSlug = useStore((state) => state.flowSlug);
-  const flowStatus = useStore((state) => state.flowStatus);
+
   const canUserEditTeam = useStore((state) => state.canUserEditTeam);
 
-  if (!params.flow) return null;
+  if (!params.team || !params.flow || !data?.flows) return null;
```
jessicamcinchak: nit: we've destructured `params` already above, can we re-use those here like `(!teamSlug || !flowParam || !data?.flows)` ? 

A couple similar cases below when passing props. Just a bit easier to read/follow imo !

apps/editor.planx.uk/src/hooks/analyticsLinks/useFlowAnalyticsLink.tsx
```diff
@@ -120,10 +120,10 @@ export const useFlowAnalyticsLink = (): string | undefined => {
   const hasAnalytics = Boolean(flow.onlineHistory.length);
   if (!hasAnalytics) return;
 
-  const isSubmissionService = Boolean(flow.publishedFlows[0]?.hasSendComponent);
+  const hasSendComponent = Boolean(flow.publishedFlows[0]?.hasSendComponent);
   const dashboardId = getAnalyticsDashboardId({
     flowSlug,
-    isSubmissionService,
+    hasSendComponent,
```
jessicamcinchak, line 126: 👍 Appreciate more explicit naming

jessicamcinchak (approved):
Changes look good ! 

Just noticed one minor thing on pizza filters: "Service type" -> "Offline" unexpectly returned 0 results for me (Is this because they are now offline _flows_, not services? Is there maybe some copy we can use in this 0 case to clarify?).

"Online", as well as all other categories, working as expected.

Image

apps/editor.planx.uk/src/components/Breadcrumbs.tsx
zz-hh-aa: Ah much nicer 😅 thanks for review! Done in [fae5ece](https://github.com/theopensystemslab/planx-new/pull/6811/commits/fae5ece66bbf20c33d7076693afa1c7e873356ca)

PR #6814, opened 2026-06-04
feat: auto-suggest editor data values for all Planning Data fns

jessicamcinchak:
Follow-up to #6799 ! 

**Now covering:** 
- `property.localAuthorityDistrict`
- `property.localPlanningAuthority`
- `property.region`
- `property.developmentCorporation`
 
**Outstanding / next PR:** 
- `property.constraints.planning` including granular A4s

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/shared/hooks.ts
+++ b/apps/editor.planx.uk/src/@planx/components/shared/hooks.ts
@@ -1,5 +1,5 @@
 import { useQuery } from "@tanstack/react-query";
-import { getLocalAuthorityDistricts } from "lib/planningData/requests";
+import { getEntityNames } from "lib/planningData/requests";
 import { useEffect, useState } from "react";
 
 export type UseFileUrlProps =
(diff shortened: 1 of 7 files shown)


apps/editor.planx.uk/src/@planx/components/shared/hooks.ts
```diff
@@ -34,10 +34,19 @@ export const useFileUrl = (props: UseFileUrlProps) => {
...
 };
 
-export const useLocalAuthorityDistricts = () => {
+export const usePlanningDataEntityNames = (fn: string) => {
+  const fnToDataset: Record<string, string> = {
+    "property.localAuthorityDistrict": "local-authority-district",
+    "property.localPlanningAuthority": "local-planning-authority",
+    "property.region": "region",
+    "property.developmentCorporation": "development-corporation",
+    // "property.ward": "ward" // only 'turn on' once we setup query pagination and if UI is okay (current PD API limit = 500, but there's ~7k wards)
```
jessicamcinchak, line 43: I've left this one here as a reminder, but we do _not_ want to auto-suggest `property.ward` values at this time for a couple of reasons! 
- There are ~7k records
- That's an unideal autocomplete user experience 😅 
- Planning Data API returns max 500 records (which is plenty to accomodate these other datasets), but we'd need to handle paginated requests for wards
  - Or alternatively get more clever and query wards by team boundary, then only suggest intersecting entitiy names

AFAIK, only Doncaster actually automate against these values for now, so I think safe to omit !

DafyddLlyr: Fun problem 😅 

I guess we could also only auto-suggest here after n letters have been typed / aggressive debounce for short-named wards in order to make the list smaller (within the _second_ field). But yep very very much support not tackling this one now...!

apps/editor.planx.uk/src/@planx/components/shared/utils.ts
```diff
@@ -69,21 +69,17 @@ export const getOptionsSchemaByFn = (
   fn?: string,
   defaultOptionsSchema?: string[],
   currentOptions?: (string | undefined)[],
-  localAuthorityDistrictSchema?: string[] | undefined,
 ) => {
   let schema = defaultOptionsSchema;
 
   // For certain data fields, suggest based on full ODP Schema enums rather than current flow schema
+  //   Note that many `property.{}` fields which auto-suggest based on Planning Data entities set their schema one level up
```
DafyddLlyr, line 76: Worthwhile comment..!

DafyddLlyr (approved):
Exactly what I had in mind...!

apps/editor.planx.uk/src/@planx/components/shared/utils.ts
jessicamcinchak: Yes, only downside of this dynamic hook is that it's hard to see all that auto-suggestable `fn`'s in a single place now ! Also working on a Notion doc for 'data fields and automation' to collect/explain these at a high-level.

PR #6815, opened 2026-06-05
feat(lps): Use Lambda@Edge to handle URL extensions

DafyddLlyr:
## What does this PR do?
- Created a Lambda@Edge function to map URLs (see below)
- Configures this function as part of the Pulumi IAC setup
- Basic test coverage

## Context
LPS is a static Astro site served from a private S3 bucket via a CloudFront REST origin (OAI). A REST origin performs no directory-index / pretty-URL resolution, so `/about` maps to an S3 key literally named `about`. Astro (`build.format: "file"`) outputs `about.html`.

Previously the deploy script bridged this gap by stripping `.html` from keys on upload (`about.html` → `about`). That created a key-space mismatch between `dist/` and the bucket, so the static `aws s3 sync --delete` deleted every live page on each deploy and re-created it moments later. During this window CloudFront served 403/404 responses, recorded by UptimeRobot as downtime on every deploy.

This function moves the mapping to the edge instead, so the site is deployed to S3 as-is and running `s3 sync --delete`  on deploy only ever overwrites live objects.

## Testing
I'd love to merge this to `main` to test the Pulumi staging setup, usually hit a few issues at this stage as there's not an easy way to test this without merging. I've aimed to follow the pattern used by our other Lambda function, but I'd also like to see how builds are running here once in place so we can fix or revert this one.

### DIFF
--- a/apps/localplanning.services/astro.config.mjs
+++ b/apps/localplanning.services/astro.config.mjs
@@ -51,8 +51,9 @@ export default defineConfig({
     },
   ],
   build: {
-    // AWS Cloudfront requires /about.html not /about/index.html
-    format: isCloudfrontBuild 
+    // Emit /about.html not /about/index.html for AWS Cloudfront
+    // The LPS URL rewrite Lambda@Edge function maps pretty URLs (/about) to *.html keys
+    format: isCloudfrontBuild
       ? "file"
       : "directory"
   }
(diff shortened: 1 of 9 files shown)


jessicamcinchak (approved):
Happy for this to merge/test/revert-if-needed today !

PR #6817, opened 2026-06-05
fix(editor): Broken header layout when DowntimeBanner displayed

DafyddLlyr:
## What's the problem?
The `DowntimeBanner` element can break the layout (dependent on other children). This shows up as an expanded header in places.

image

## What's the cause?
This is a CSS Grid issue - we declare in `RootContainer` that there will be three rows, but `DowntimeBanner` makes it four children and breaks the layout - 

https://github.com/theopensystemslab/planx-new/blob/7f084798b072a5c111861e715dd308becbabddcb/apps/editor.planx.uk/src/pages/layout/PublicLayout.tsx#L21-L26

## What's the solution?
Just wrap in a `<Box/>` so the number of children stays consistent - there's a single `<Header/>` child instead of two breaking out from the fragment.

## Testing...
- Toggle downtime banner on staging, navigate through a few nodes, see expanded header ❌ 
- Toggle downtime banner on pizza, navigate through a few nodes, see correctly sized header ✅

### DIFF
--- a/apps/editor.planx.uk/src/components/Header/Header.tsx
+++ b/apps/editor.planx.uk/src/components/Header/Header.tsx
@@ -323,7 +323,7 @@ const Header: React.FC = () => {
   const showDowntimeBanner = data?.downtimeBanner?.isVisible || false;
 
   return (
-    <>
+    <Box>
       {showDowntimeBanner && <DowntimeBanner />}
       <Root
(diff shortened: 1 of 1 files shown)


DafyddLlyr:
Closed in favour of the more thorough #6818

PR #6820, opened 2026-06-05
feat: Toggle feature flags within panel

ianjon3s:
## What does this PR do

After testing and thinking more about the feature flag global settings page introduced in https://github.com/theopensystemslab/planx-new/pull/6810, it seems disruptive to jump back to a global level to toggle feature flags.

This PR reuses the panel setup introduced by notifications, to toggle feature flags without changing location. Also improved is the marker to show how many flags are available, and how many of which have been applied.

- **No feature flags:**
image

- **Flags available, none applied:**
image

- **Flags available, one or more applied:**
image


jessicamcinchak (approved):
Handy!

PR #6821, opened 2026-06-05
feat(editor): Capture OS error state mocks

DafyddLlyr:
Trying to make the best of a bad situation...! 😅 

This PR adds two new stories, which cover error states when OS APIs go down. 

The aim here would be to improve over time how we handle this visually - for example a more meaningful error in `FindProperty` and maybe an OpenStreetMap tile fallback for the `<my-map/>` component (right now I think this kicks in if keys/proxy endpoints are missing, not in the case of 500s).


jessicamcinchak (approved):
Thanks for grabbing these!

PR #6824, opened 2026-06-09
feat[api]: auto-suggest editor data values for `property.constraints.planning`

jessicamcinchak:
Follows on from #6814

Relies on https://github.com/theopensystemslab/planx-core/pull/1009

API-side changes required to expose a simple public planning constraints schema that we can then use to auto-suggest data values in the editor. 

**Examples:** 
- Generic public schema (team agnostic): https://api.6824.planx.pizza/planning-constraints-schema
- Schema plus granular article four variables:
  - For a team which has A4s configured: https://api.6824.planx.pizza/planning-constraints-schema?localAuthority=camden
  - For a team which does not have A4s configured (gently fallsback to default generic schema): https://api.6824.planx.pizza/planning-constraints-schema?localAuthority=westminster
  
In future we may configure a special "wildcard" team like `?localAuthority=opensystemslab` to return _all_ granular A4s across all councils rather than fallback to default, but I've decided to leave that out of this initial PR - it's most important we have council-specific auto-suggest working !


apps/api.planx.uk/modules/gis/service/digitalLand.ts
```diff
@@ -9,83 +9,7 @@ import { gql } from "graphql-request";
...
-  "south-gloucestershire": southGloucestershire,
-  "south-staffordshire": southStaffordshire,
-  southwark,
-  "st-albans": stAlbans,
-  "stockport-metropolitan": stockport,
-  "stoke-on-trent": stoke,
-  tewkesbury,
-  torbay,
-  "west-berkshire": westBerkshire,
-};
```
jessicamcinchak, line 88: Since we now want to share A4 configs between our planning constraints geo-intersection endpoint and this new basic schema endpoint, I've moved these out to `helpers.ts`

cc @augustlindemer impacts A4 setup PRs, but just file location change here, mapping rules stay exactly the same!

apps/api.planx.uk/modules/gis/service/digitalLand.ts
```diff
@@ -379,6 +304,38 @@ async function go(
...
+    }
+  });
+
+  if (subVariablesFound < subVariables.length) {
+    return responseObject;
+  } else {
+    return resObjWithDesignated;
+  }
+};
+
```
jessicamcinchak, line 338: This function used to be in `helpers` but is only referenced here - just a little re-organisation, no functional change!

apps/api.planx.uk/modules/gis/service/helpers.ts
```diff
@@ -1,33 +1,110 @@
...
+}
 
-  // If any of the subvariables are true, then set "designated" to true
-  subVariables.forEach((s) => {
-    if (resObjWithDesignated[`designated.${s}`]?.value) {
-      resObjWithDesignated["designated"] = { value: true };
-    }
-  });
+/**
+ * When a team publishes their granular Article 4 data, add them to this list. Key must match team slug
```
freemvmt, line 41: 🫡

apps/api.planx.uk/modules/gis/service/helpers.ts
```diff
@@ -1,33 +1,110 @@
...
+ * The database column team_setting.has_article4_schema also needs to be updated via the Hasura console
+ */
+export const localAuthorityMetadata: Record<string, LocalAuthorityMetadata> = {
```
freemvmt, line 44: minor nit, imo it's a little bit confusing that the object `localAuthorityMetadata` is not actually an example of `LocalAuthorityMetadata` - maybe it could be called something like `localAuthorityMetadataLookup` or similar? this might also just be a me thing 😅

apps/api.planx.uk/modules/gis/service/helpers.ts
```diff
@@ -1,33 +1,110 @@
...
-  }
+  return localAuthorityMetadata[localAuthority];
 };
 
-export { addDesignatedVariable };
+// Return the granular article four data values, if applicable, for a local authority
+export const getLocalAuthorityArticleFourSchema = (
+  localAuthority?: string,
+): string[] | undefined => {
+  if (
```
freemvmt, line 94: can we leverage `getLocalAuthorityMetadata` here to avoid repeating the if statement?

freemvmt (approved):
Beaut! 📜 

Could absolutely merge as is, or I have a couple minor/cosmetic points.

PR #6828, opened 2026-06-09
fix: stagger `update-templated-flow` events further

jessicamcinchak:
See https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1781020983863579

This was recently _reduced from 10s to 5s_ when we introduced notifications here https://github.com/theopensystemslab/planx-new/pull/6667 (I believe mostly for ease of UI testing, council editors are surely not logging in this frequently and noticing a "tardy" notification!)

The first occurances of `Failed to update templated flow...` begin on Airbrake in late May, right around merge of that reduction :dart: 

Based on errors here, we're overwhelming Hasura at point of event creation which is resulting in errors like `One-off Scheduled event timed out while processing`, _before_ it's even able to attempt to hit an API endpoint. 

I've also chosen to bump the retry interval from 10s to 30s here across the board (everywhere we use scheduled events including send, not just updating templated flows). Again, no functional change just hopefully giving Hasura and the API a bit more recovery room if needed.

See Slack thread for most robust ideas / investigation points! This increased stagger intends to immediately ease issues, but recognise not necessarily sustainable !

### DIFF
--- a/apps/api.planx.uk/lib/hasura/metadata/index.ts
+++ b/apps/api.planx.uk/lib/hasura/metadata/index.ts
@@ -57,6 +57,7 @@ export const createScheduledEvent = async (
         ],
         retry_conf: {
           num_retries: 1,
+          retry_interval_seconds: 30,
         },
       },
     });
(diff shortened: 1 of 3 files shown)


zz-hh-aa (approved):
Thanks for the explainer Jess and clear notes in Slack thread--green ticking cause yes it seems to make sense to stagger the events further / given that we halved them recently, even if maybe not the solution here...

PR #6829, opened 2026-06-09
[infra] fix logic for alias allocation to legacy distributions 

freemvmt:
~Changes to Doncaster cert represented here already on prod, hence Doncaster's [custom domain](https://planningservices.doncaster.gov.uk/) being back up without this having rolled through.~

This PR fixes an error in the logic in the `application` layer, which is a holdover from a previous stage of development when we planned to have two intermediate states between `legacy-with-validation` and `shared-final` (called `cutover-init` and `cutover-ongoing`), which we later collapsed into `cutover-ongoing` for simplicitly. The logic as-was resulted in the alias being unexpectedly removed from the legacy distribution, which led to downtime for the Doncaster custom domain, which was [complicated to fix](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1781032501830019).

Additionally/relatedly, I didn't take sufficient account of the effect of unrelated prod deploys in the period in which we have already moved the domain from legacy -> shared CloudFront distribution. Without an additional state to represent this configuration, we have to be especially careful with sequencing. I will bring a follow up PR with changes to the [migration steps in the howto](https://github.com/theopensystemslab/planx-new/blob/main/doc/how-to/how-to-setup-custom-domains.md#b-migrating-a-council-from-legacy-setup) to rectify this (#[6837](https://github.com/theopensystemslab/planx-new/pull/6837))


jessicamcinchak (approved):

infrastructure/application/index.ts
```diff
@@ -323,7 +323,7 @@ export = async () => {
     });
 
     // only include domains for which we have not yet moved alias to shared CDN via AWS CLI (deploy would fail o/w)
-    const domains = cloudFrontState !== "cutover-ongoing" ? [domain] : [];
```
freemvmt, line 326: Note that before the fix, this logic resulted in the alias being removed during cutover. We actually want the alias to be applied in all states for which the domain is returned by `getLegacyDomains`.

infrastructure/common/customDomains.ts
```diff
@@ -24,8 +24,7 @@ export const getCustomDomains = (env: string): CustomDomain[] =>
         {
           name: "doncaster",
           domain: "planningservices.doncaster.gov.uk",
-          cloudFrontState: "cutover-ongoing",
-          certificateLocation: "pulumiConfig",
+          cloudFrontState: "shared-final",
```
freemvmt, line 27: Doncaster has been migrated (not especially smoothly) - rolling this out to prod will destroy the legacy distribution and associated cert.

jessicamcinchak (approved):
Changes look good, thanks for working through Doncaster bit recently !

PR #6830, opened 2026-06-10
fix: remove row-level update permission checks for `api` role on `flows` table

jessicamcinchak:
Trying to figure out if this changes anything related to this bug: https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1781020983863579


jessicamcinchak:
Not a solution, these permissions are needed to match UI state to database.

PR #6834, opened 2026-06-10
infra: bump ShareDB and Hasura `memory` on staging

jessicamcinchak:
More attempts at resolving https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1781094283435389?thread_ts=1781020983.863579&cid=C088K9ZL8EA

See notes from this specific comment:
> In trying to pinpoint what's different about GDPO source template, I'm comparing unflattened flow size (flows.data which ShareDB & Hasura operate on) by simply counting number of nodes:
> - HPP = 68
> - LDC = 146
> - FOI source template = 169
> - A4 = 793 (more customisable nodes than GDPO)
> - GDPO = 4,408 :melting_face: 
>   - Non-source-template osl/permitteddevelopment =  14,171 for comparison though which ShareDB typically connects to fine (eg no "Page not found" issue)
>
> Is this literally just a much larger source template data structure than we've had to operate on before? Can we "just" allocate more memory/larger containers/($$) to the situation until we better understand how to optimise data operations?

There won't be any noticable difference on pizza here, we'll have to test this via Pulumi/AWS staging infra. Our API `memory` is already at `4096` so I haven't bumped that any further here.

### DIFF
--- a/apps/editor.planx.uk/src/lib/graphql/links.ts
+++ b/apps/editor.planx.uk/src/lib/graphql/links.ts
@@ -70,6 +70,8 @@ const getWsLink = () => {
       uri: import.meta.env.VITE_APP_HASURA_WEBSOCKET!,
       options: {
         reconnect: true,
+        lazy: true,
+        inactivityTimeout: 30000, // 30s
         connectionParams: async () => {
           const jwt = await getJWT();
           return {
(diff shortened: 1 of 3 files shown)


apps/editor.planx.uk/src/lib/graphql/links.ts
```diff
@@ -70,6 +70,8 @@ const getWsLink = () => {
       uri: import.meta.env.VITE_APP_HASURA_WEBSOCKET!,
       options: {
         reconnect: true,
+        lazy: true,
+        inactivityTimeout: 30000, // 30s
```
jessicamcinchak, line 74: These two lines come from another thread where I was looking into Hasura's "socket hang up" error thrown on create flow from template attempt - https://github.com/hasura/graphql-engine/issues/4509#issuecomment-882503283

I don't think it hurts to try them! Looking into these has also highlighted that `@apollo/client/link/ws` is no longer maintained and we should look into migrating this: https://www.apollographql.com/docs/react/api/link/apollo-link-ws

zz-hh-aa (approved):

apps/editor.planx.uk/src/lib/graphql/links.ts
freemvmt: shall we capture this point about apollo in a ticket?

jessicamcinchak: yep will add to Trello tech debt :heavy_plus_sign:

jessicamcinchak: https://trello.com/c/lDJVF69G/3651-upgrade-apollo-client-link-ws-package-to-recommended-alternatives-for-editor-websocket-subscriptions

PR #6835, opened 2026-06-10
chore: batch `updateTemplatedFlow` mutations into single request

jessicamcinchak:
A small optimisation effort related to https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1781020983863579

Unfortunately the `data` payload is still a large jsonb (from profiling, this is the spot where this function _was_ hanging and _is_ still most likely to hang!). We don't have a better answer for this yet, common suggestions to only update relevant jsonb keys don't really work in our case and we need to overwrite the full flow graph.

Nonetheless, combining these three independent mutations into a single network request may help a smidge!

### DIFF
--- a/apps/api.planx.uk/modules/flows/updateTemplatedFlow/service.ts
+++ b/apps/api.planx.uk/modules/flows/updateTemplatedFlow/service.ts
@@ -71,7 +71,7 @@ export const updateTemplatedFlow = async (
   const updateTemplatedFlowTransactionResponse =
     await $api.client.request<UpdateTemplatedFlowTransactionResponse>(
       gql`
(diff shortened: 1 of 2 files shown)


zz-hh-aa (approved):
Thanks Jess! Makes sense and behaves as expected on pizza. 

Did the previous version run the risk that one operation would succeed but not the others? eg if merged data was updated, but then both "History" comment and "Notification" failed--does combining them into one transaction sidestep this possibility?

jessicamcinchak:
@zz-hh-aa Yes exactly ! Transaction also has benefit of "all of nothing" update here. From profiling, the hangup point was definitely the first update of the three, so it's very unlikely we ever hit that issue in this case though.

PR #6839, opened 2026-06-11
feat: Implement new component select

ianjon3s:
## What does this PR do?

- Hooks up new component select with feature flag `COMPONENT_SELECT`
- Updates tests to use new component adding method

**To test:**
- https://6839.planx.pizza/
- Toggle feature flag `COMPONENT_SELECT`
- Add a new component to a flow


DafyddLlyr (approved):
Small question here about how URLs are handles, possibly missed due the Storybook first approach. https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1781608534121709

DafyddLlyr (changes requested):
Sorry, hit the wrong option above...!

ianjon3s:
@DafyddLlyr as discussed, I've dropped the unique URL for adding components. There's a bit of cross-targeting in the CSS and tests now where both `<a>` and `<button>` are present for each respective version of the component. I'll drop references to the link when removing the feature flag.

apps/editor.planx.uk/src/pages/FlowEditor/lib/componentSelectorState.ts
DafyddLlyr, line 1: Here we're manually implementing another store from scratch, when we already have an established pattern for this (using the `EditorUIStore`). I'd recommend removing this custom implementation and reaching for our current state management solution.

We could also use query params (which we then strip from history I guess?) to manage this if we wanted to.

On a side note - we probably need to look at an `AGENTS.md` and/or `CLAUDE.md` to help enforce these conventions 👌

DafyddLlyr (changes requested):
Working as expected...! One small suggestion on implementation.

apps/editor.planx.uk/src/pages/FlowEditor/lib/componentSelectorState.ts
ianjon3s: Thanks @DafyddLlyr , I've removed the custom store in favour of using the `EditorUIStore`. I'll follow up with a boilerplate for an `AGENTS` file to keep this consistency

DafyddLlyr (approved):
Nice!

PR #6840, opened 2026-06-11
feat: assign permissions to teamAdmin

zz-hh-aa:
v2--new version of #6807 but, per [Daf's suggestions](https://github.com/theopensystemslab/planx-new/pull/6807#pullrequestreview-4429107372), reworked to not use inherited roles anymore

Much of this (~400 lines?) is updated tests and helper functions--if it's helpful for reviewing, I can separate into Hasura + editor PRs now that the model is clearer! 

# What does this PR do?
- Removes `teamAdmin` as inherited role 
    - Changes `team_members` so that the constraint is on `team_id`, `user_id` _and_ `role` (not just `team_id` and `user_id`)
    - `teamAdmin` is an _additional_ role granted to `teamEditors`
- Adds relevant mutations for all of these scenarios
    - eg updating a user's name or email only will just mutate `users` table
    - promoting a `teamEditor` to `teamAdmin` will add a row to `team_members`, and demoting will remove a row
- Introspection tests + editor tests

# Why?
- The benefits of inherited roles weren't actually relevant when we got into role-based custom checks on row-level permissions, this is a simpler way of ensuring that team admins have all the permissions of team editors but with an additional layer of permissions where relevant (subscription page, team membership management)

# Next steps
- [ ] Make the Team Admin toggle into a dropdown
- [ ] Migrate select users to new Team Admin role

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/List/Public/List.stories.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/List/Public/List.stories.tsx
@@ -1,8 +1,8 @@
 import { Meta, StoryObj } from "@storybook/tanstack-react";
 
 import Wrapper from "../../fixtures/Wrapper";
-import { SCHEMAS } from "../Editor";
 import Editor from "../Editor";
+import { SCHEMAS } from "../Editor";
 import ListComponent from ".";
 import Public from ".";
 
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
@@ -92,7 +92,6 @@ export const EditUserModal: React.FC<EditUserModalProps> = ({
       formatted.email !== (member.email ?? "").toLowerCase();
 
     const roleChanged = values.role !== member.role;
-    console.log({ userInfoChanged, roleChanged });
 
     if (userInfoChanged) {
       await updateUserOnly({
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
@@ -74,7 +74,7 @@ export const EditUserModal: React.FC<EditUserModalProps> = ({
         },
       ],
       onCompleted: (data) => {
-        if (data.delete_team_members.affected_rows > 0) {
+        if (data.DeleteTeamMembers.affected_rows > 0) {
           handleCompleted("Successfully updated a user");
         } else {
           toast.warning("No admin role found to remove");
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
@@ -12,11 +12,7 @@ import React from "react";
 import { Switch } from "ui/shared/Switch";
 
 import { upsertMemberSchema } from "../formSchema";
-import {
-  GET_USERS_FOR_TEAM_QUERY,
-  UPDATE_TEAM_MEMBER,
-  UPDATE_USER_DETAILS,
-} from "../queries";
+import { GET_USERS_FOR_TEAM_QUERY, UPDATE_TEAM_MEMBER } from "../queries";
 import { type EditUserModalProps, type UserFormValues } from "../types";
 import { EmailField } from "./Fields/EmailField";
 import { NameFields } from "./Fields/NameFields";
@@ -35,17 +31,6 @@ export const EditUserModal: React.FC<EditUserModalProps> = ({
     toast.success(successMessage);
   };
 
-  const [updateUserOnly] = useMutation(UPDATE_USER_DETAILS, {
-    refetchQueries: [
-      {
-        query: GET_USERS_FOR_TEAM_QUERY,
-        variables: { teamSlug: useStore.getState().teamSlug },
-      },
-    ],
-    onCompleted: () => handleCompleted("Successfully updated user details"),
-    onError: () => toast.error("Failed to update the user, please try again"),
-  });
-
   const [updateTeamMember, { loading }] = useMutation(UPDATE_TEAM_MEMBER, {
     refetchQueries: [
       {
@@ -60,31 +45,18 @@ export const EditUserModal: React.FC<EditUserModalProps> = ({
   const handleSubmit = (values: UserFormValues) => {
     const formatted = { ...values, email: values.email.toLowerCase() };
 
-    if (showTeamAdminSwitch) {
-      updateTeamMember({
-        variables: {
-          userId: member.id,
-          teamId: teamId,
-          role: values.role,
-          userValues: {
-            first_name: formatted.firstName,
-            last_name: formatted.lastName,
-            email: formatted.email,
-          },
+    updateTeamMember({
+      variables: {
+        userId: member.id,
+        teamId: teamId,
+        role: values.role,
+        userValues: {
+          first_name: formatted.firstName,
+          last_name: formatted.lastName,
+          email: formatted.email,
         },
-      });
-    } else {
-      updateUserOnly({
-        variables: {
-          userId: member.id,
-          userValues: {
-            first_name: formatted.firstName,
-            last_name: formatted.lastName,
-            email: formatted.email,
-          },
-        },
-      });
-    }
+      },
+    });
   };
 
   return (
@@ -117,21 +89,19 @@ export const EditUserModal: React.FC<EditUserModalProps> = ({
               <Box sx={{ mt: 2, mb: 2 }}>
                 <NameFields />
               </Box>
-              {showTeamAdminSwitch && (
-                <Switch
-                  name="role"
-                  checked={formik.values.role === "teamAdmin"}
-                  onChange={() =>
-                    formik.setFieldValue(
-                      "role",
-                      formik.values.role === "teamEditor"
-                        ? "teamAdmin"
-                        : "teamEditor",
-                    )
-                  }
-                  label={"Team Admin"}
-                />
-              )}
+              <Switch
+                name="role"
+                checked={formik.values.role === "teamAdmin"}
+                onChange={() =>
+                  formik.setFieldValue(
+                    "role",
+                    formik.values.role === "teamEditor"
+                      ? "teamAdmin"
+                      : "teamEditor",
+                  )
+                }
+                label={"Team Admin"}
+              />
             </DialogContent>
             <DialogActions>
               <ModalActions
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
@@ -29,6 +29,7 @@ export const EditUserModal: React.FC<EditUserModalProps> = ({
 }) => {
   const toast = useToast();
   const teamId = useStore((state) => state.teamId);
+  const isPlatformAdmin = member.role === "platformAdmin";
 
   const handleCompleted = (successMessage: string) => {
     onClose();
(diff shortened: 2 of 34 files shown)


apps/editor.planx.uk/src/@planx/components/List/Public/List.stories.tsx
```diff
@@ -1,8 +1,8 @@
 import { Meta, StoryObj } from "@storybook/tanstack-react";
 
 import Wrapper from "../../fixtures/Wrapper";
-import { SCHEMAS } from "../Editor";
 import Editor from "../Editor";
+import { SCHEMAS } from "../Editor";
```
zz-hh-aa, line 5: Linter caught a couple things...

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
```diff
@@ -25,56 +26,76 @@ import { ModalActions } from "./ModalActions";
...
-    refetchQueries: [
-      {
-        query: GET_USERS_FOR_TEAM_QUERY,
-        variables: { teamSlug: useStore.getState().teamSlug },
-      },
-    ],
-    onCompleted: () => handleCompleted("Successfully updated user details"),
-    onError: () => toast.error("Failed to update the user, please try again"),
-  });
+  const [updateUserOnly, { loading: loadingUser }] = useMutation(
```
zz-hh-aa, line 40: While there is a fair bit of repetition here, I got partway through a refactor with a generic hook when I realised there was also enough variation that the generic hook didn't feel useful / like it was adding unnecessary complexity.

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
```diff
@@ -25,56 +26,76 @@ import { ModalActions } from "./ModalActions";
...
-        },
-      });
-    } else {
-      updateUserOnly({
+    const userInfoChanged =
+      formatted.firstName !== member.firstName ||
+      formatted.lastName !== member.lastName ||
+      formatted.email !== (member.email ?? "").toLowerCase();
+
+    const roleChanged = values.role !== member.role;
```
DafyddLlyr, line 111: Not sure if it's actually a simpler or better alternative to this (this code is clear and solid...!), but just a reminder that we also have a `.dirty` property that can be used in these instances. I'm not sure if there's a way of identifying easily _which_ fields are dirty though in a way which is meaningfully different or simpler than this.

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
```diff
@@ -25,56 +26,76 @@ import { ModalActions } from "./ModalActions";
...
+    REMOVE_TEAM_ADMIN_ONLY,
+    {
+      refetchQueries: [
+        {
+          query: GET_USERS_FOR_TEAM_QUERY,
+          variables: { teamSlug: useStore.getState().teamSlug },
+        },
+      ],
+      onCompleted: (data) => {
+        if (data.delete_team_members.affected_rows > 0) {
```
DafyddLlyr: nit: snake_case creeping into TS code - is it possible to rename these in the query to camelCase?

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/queries.ts
```diff
@@ -37,49 +41,53 @@ export const UPDATE_USER_DETAILS = gql`
...
+      first_name
+      last_name
+      email
     }
   }
 `;
 
-export const CREATE_AND_ADD_USER_TO_TEAM = gql`
-  mutation CreateAndAddUserToTeam(
+/** Our teamAdmin permission is a layer _on top of_ teamEditor
```
DafyddLlyr:
```suggestion
/**
 * Our teamAdmin permission is a layer _on top of_ teamEditor
```

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/queries.ts
```diff
@@ -99,20 +107,53 @@ export const REMOVE_TEAM_MEMBER = gql`
   }
 `;
 
-export const ADD_EXISTING_USER_TO_TEAM = gql`
-  mutation AddExistingUserToTeam(
-    $role: user_roles_enum
-    $teamId: Int!
-    $userId: Int!
-  ) {
+export const ADD_EXISTING_USER_TO_TEAM_AS_EDITOR = gql`
```
DafyddLlyr, line 112: The `_AS_EDITOR` and `_AS_ADMIN` prefixes here are a little unclear I think and slightly blur the boundary between permissions and operations.

The operation (which record(s) to insert) are a pure GQL mutation, the permissions are derived from the available roles in the user's JWT.

If we were to rename these so that the operations just describe what's happening, I think this would be clearer. The permission layer is then handled via the `x-hasura-role` header attached to the request.

Docs: https://hasura.io/docs/2.0/auth/authentication/jwt/#hasura-jwt-claim-description

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/utils.tsx
```diff
@@ -8,8 +8,11 @@ const getRole = (user: User): Role => {
   if (user.isPlatformAdmin) return "platformAdmin";
   if (user.isAnalyst) return "analyst";
 
+  const isUserTeamAdmin = user.teams.some((user) => user.role === "teamAdmin"); // teamAdmins have two team_member records per-team: teamEditor _and_ teamAdmin. when we check for 'some' here, this is _within_ the team (ie checking if member x is a teamAdmin of this _specific_ team, not teamAdmin of _any_ team)
```
DafyddLlyr: nit: If this was a JSDoc, IDEs would pick it up better.

Image

DafyddLlyr (changes requested):
A few small comments and nits below, and then a wider blocking permission issue.

On the tests, it's great to see the additional coverage here...! It's probably worth a comment here somewhere saying that these are testing the UI state only and not the underlying permission model.

Question - how is this working without the change to `getDefaultRoleForUser()` in the API? Most requests made in-editor by teamAdmin users will now have `x-hasura-default-role: "teamAdmin"` set which should block functionality (e.g., requesting submissions or feedback). The default role should be the one of least-privilege, which we can then override with `x-hasura-role` for select operations.

Testing this out on the Pizza I can see that this is indeed the case - if I log in as a `teamAdmin` user I cannot access a bunch of pages - all blocked by a permission error as I'm making requests which the wrong role.

I'd expect that we need to...
 - grant two roles on user creation (done here!)
 - if a user has two roles, assign the appropriate default
 - when making specific privileges roles (e.g. adding a `teamAdmin`), we need to dictate which role to use. If this is not within a user's `x-hasura-allowed-roles`, the operation will fail.

We've had a few scattered discussions over a few PRs (and a few weeks!) - very happy to walk through the approach here at dev call if you'd like, or the above is unclear?

DafyddLlyr:
@zz-hh-aa Just spotted that dev call is tomorrow not today - free to catch up anytime tbh if helpful...!

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/queries.ts
zz-hh-aa: Ok I think this is further evidence that the naming was unclear! The roles here are specifying what role the team member _being added_ has, not the user conducting the operation. 

I started renaming them to `ADD_EXISTING_EDITOR_TO_TEAM` and `ADD_EXISTING_ADMIN_TO_TEAM` but that also didn't quite seem right (eg because the existing user by definition does not have these roles yet). 

So I've opted for a JSDoc comment to make it clearer that the name is about the new team member's role and not the current user's role.

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/utils.tsx
zz-hh-aa: Done in [876d13d](https://github.com/theopensystemslab/planx-new/pull/6840/commits/876d13d042cfb2833d3c2cd94adc1a0c97ac11df)

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/queries.ts
zz-hh-aa: Done in [876d13d](https://github.com/theopensystemslab/planx-new/pull/6840/commits/876d13d042cfb2833d3c2cd94adc1a0c97ac11df)

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
zz-hh-aa: Done in [876d13d](https://github.com/theopensystemslab/planx-new/pull/6840/commits/876d13d042cfb2833d3c2cd94adc1a0c97ac11df)

zz-hh-aa:
Thanks Daf, I think I've worked out the kinks this time but also happy to talk through at dev call!

re permissions denied issue when viewing a team as 'Team viewer': this was a small frontend error, fixed in [06bde01](https://github.com/theopensystemslab/planx-new/pull/6840/commits/06bde019460ec8c7ad5fde7b0c180a97858a83af)

I have removed `teamAdmin` as a potential default role assigned in `getDefaultRoleForUser` and it should now still be `teamEditor` [d0df752](https://github.com/theopensystemslab/planx-new/pull/6840/commits/d0df752be3ce2d186994a916836f8cf31d32904e). (NB: seems inaccurate to say that this assigns the role of least privilege though, if `platformAdmin` is the first to be checked / returned?) 

`teamAdmin` exists as an allowed role in `"x-hasura-allowed-roles" but this part of the JWT code didn't need updating because `getAllowedRolesForUser(user)` was already pulling `teamAdmin` from `user.teams`. 

So instead of changing and checking a user's _default_ role when making certain requests (eg to `team_members`), I've been checking if there is an entry in the db for this user with `x-hasura-user-id` and the `teamAdmin` role for a given team. Does that make sense, and if so, is there a reason as to why we shouldn't do it this way? I previously hadn't thought about using `x-hasura-allowed-roles` for the check and used the pattern I'd seen in the db previously (if `x-hasura-user-id` has x role). 

eg:
```
  - role: teamAdmin
    permission:
      check:
        team:
          members:
            _and:
              - user_id:
                  _eq: X-Hasura-User-Id
              - role:
                  _eq: teamAdmin
```

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/queries.ts
```diff
@@ -142,7 +145,7 @@ export const CREATE_TEAM_ADMIN_ONLY = gql`
 
 export const REMOVE_TEAM_ADMIN_ONLY = gql`
   mutation RemoveTeamAdminOnly($teamId: Int!, $userId: Int!) {
-    delete_team_members(
+    DeleteTeamMembers: delete_team_members(
```
DafyddLlyr: nit: PascalCase slipped in here instead of camelCase. It's an easy one to make as PascalCase is used or the mutation / query name (e.g. `mutation RemoveTeamAdminOnly`) but this line is just a mapping of the `delete_team_members` table.

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
DafyddLlyr: nit: One still outstanding issue here.

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/queries.ts
DafyddLlyr: Aha yep - I have just read this in a funny way - it's grammatically correct just ambiguous. JSDocs are a great solution here, thanks for clarifying.

DafyddLlyr (changes requested):
Just taken another look at this, and I think I can clarify and demonstrate this. The DB checks you outlined in your comment above are correct - but we're never hitting them.

- I log into the Pizza as a `teamAdmin` (user and role must be configured to match on staging)
- My JWT has `x-hasura-default-role` as `teamEditor` ([jwt.io](https://www.jwt.io/) is nice for this)
- I can get a list of flows (because `teamEditor`s can access them)
- I cannot access the "Subscription" page, despite Hasura permissions being correctly configured ❌ 
  - This is happening because we're falling back to the default `x-hasura-default-role: "teamEditor"`
  - We need to add an override to set this header to  `x-hasura-role: "teamAdmin"` - this defines it as `teamAdmin` query

The same pattern applies to other `teamAdmin` mutations / queries.

Image

DafyddLlyr (approved):
Great...! Let's merge to staging and test there - all looking good to me 🚀

PR #6843, opened 2026-06-11
fix: when archiving or deleting `flows`, hard delete any associated notifications

jessicamcinchak:
See https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1781183979127809

On staging, we had an odd error of a couple, not all, team routes failing to load. 

**This was due to a few factors:**
- We intentionally do not sync `notifications` from prod to staging, instead you'd re-trigger notifications unique to staging when publishing a source template on staging. BUT `notifications` were persisting on staging between nightly data syncs. This is because they have a soft Hasura "relationship" to flows, not a foreign key with `on delete cascade` behavior
- This meant that a templated flow could be archived or deleted from production or staging, while its' associated notification was still active/unresolved
- The `NotificationCard` component relies on being able to link to an existing flow

**How we fix this:**
- We do not need to persist notifications on staging between data syncs. We add a foreign key relationship to flows: when flows are truncated and re-synced each night, that will now cascade to notifcations too
- When an editor "soft-deletes" a flow by archiving or "deleting" it in the UI, we also now cascade this to any associated notifications via a batch mutation (database transaction) 
  - I've chosen to hard-delete associated notifications here rather than "mark them as resolved" because they are just an internal, ephemeral alert system - we don't actually need to care about data retention here imo! I also don't think we want to get into the complexity of archiving/unarchiving notifications. We'll still always have an audit of _creating_ the original notification in the `templated_flow_update_requests`, and new notifications will queue up for active flows!
 
 **How to test:**
 - Publish a source template on this pizza (eg osl/article-4)
 - Go to a team which has a templated flow of that source. They should have a notification that it's been updated
 - Archive the flow and the notification should automatically be removed (whether notification is active or you resolve it first by publishing templated flow before archiving it!)
 
 **When this merges to staging:**
 - In order to clear up Adur & Worthing and West Berks on staging, I think we can simply manually hard-delete any notifications which no longer map to an active flow in their team? Alternatively once this change is on prod, the first nightly data sync will clear all existing staging notifcations !


apps/hasura.planx.uk/metadata/databases/default/tables/public_notifications.yaml
```diff
@@ -70,6 +70,25 @@ select_permissions:
...
+        - flow_id
+      filter:
+        team:
+          members:
+            _and:
+              - user_id:
+                  _eq: x-hasura-user-id
+              - role:
+                  _eq: teamEditor
+    comment: ""
```
jessicamcinchak: `teamAdmins` were missing permission to "view" notifications equivalent to `teamEditors` - I've brought that into sync here. Each now also have permission to "delete" notifications within their own teams.

zz-hh-aa: Thanks for working through this! 

I don't think we'll actually need `teamAdmin` permissions here. Since the role is a subset of permissions granted _on top of_ `teamEditor` (ie with the [v2 approach](https://github.com/theopensystemslab/planx-new/pull/6840), which is already proving much simpler than inherited roles, all `teamAdmin` users will have a `teamEditor` record too, and this `teamEditor` will be the role used in most cases, excluding subscription page and managing team members).

jessicamcinchak: Aha, of course ! Thanks for this reminder, now updated to remove `teamAdmin` permissions https://github.com/theopensystemslab/planx-new/pull/6843/commits/0777ca2b7afc1151b301ffa9900933af87699bb7

apps/editor.planx.uk/src/pages/Team/queries.ts
```diff
@@ -112,6 +133,11 @@ export const DELETE_FLOW = gql`
     ) {
       id
     }
+    notifications: delete_notifications(where: { flow_id: { _eq: $id } }) {
```
zz-hh-aa, line 136: In theory this should never happen because a flow needs to be archived before being deleted (and archive would run `delete_notifications` as above) but maybe good to keep as a precaution anyway?

apps/hasura.planx.uk/migrations/default/1781199980872_set_fk_public_notifications_flow_id/up.sql
```diff
@@ -0,0 +1,6 @@
+-- When flow records are deleted, cascade to notifications (this means notifications will not persist on staging, but be reset nightly)
+alter table "public"."notifications"
```
zz-hh-aa, line 2: Just for my own understanding--is this basically doing the same as above but at the db level--so even if the flow delete operation isn't run as the above mutation, it will still necessarily delete any associated notifications?

zz-hh-aa (approved):
Thanks for fix! Works as expected on Pizza :) checked both in UI and in Hasura notifications table too

apps/editor.planx.uk/src/pages/Team/queries.ts
jessicamcinchak: Yes exactly ! It's a second-line precaution if there's any _existing_ archived templated flows that have notifications hanging around still. But as you say, the first "archive" step is most important here / will be the one actually handling the cascade for all _new_ archives going forward !

apps/hasura.planx.uk/migrations/default/1781199980872_set_fk_public_notifications_flow_id/up.sql
jessicamcinchak: Yes! The main difference here is: 
- On production, we _never_ hard-delete `flows`, we _only_ have a soft delete pattern. This allows to always recover flow content etc. So on production, I never expect notifications to actually be deleted via this foreign key, only via the batch mutation on (soft) archive/delete initiated by an editor ! 
- But on staging, because [our sync script fully truncates `flows` each night](https://github.com/theopensystemslab/planx-new/blob/main/scripts/seed-database/container.sh#L46) (eg hard-deletes rows without associated user action), we can use this database-level foreign key to handle removing associated notifications

Make sense? The staging data sync is always a bit of a tricky mental model!

zz-hh-aa: Ah yes totally, wasn't thinking about the sync here. Makes sense, thanks for clarifying :)

PR #6844, opened 2026-06-12
fix: refactor `previousCard`  and `sortBreadcrumbs` to align backwards and forwards navigation of clones

jessicamcinchak:
See this thread for context: https://opensystemslab.slack.com/archives/C01E3AC0C03/p1781008677115969

**Testing:**
Production: https://editor.planx.uk/app/testing/backwards-navigation
This pizza: https://6844.planx.pizza/app/testing/backwards-navigation

Proceed forwards in the order outlined in `backNavigationWithClones.test.ts` ("Site a caravan" &rarr; "Building" &rarr; "[cloned] Content" &rarr; "Notice").

From "Notice", go "back":
- On production, it'll skip the cloned "Content" and take you immediately back to the question "Is it a building or a caravan"
- On this pizza, it should follow the same/expected forwards order in reverse (eg "Notice" &rarr; "Content" &rarr; "Is it a building or a caravan" question)


apps/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/mocks/breadcrumbsDependentOnPassport.json
```diff
@@ -164,6 +169,7 @@
           "species": "Elm"
         }
       }
-    }
+    },
+    "seq": 6
```
jessicamcinchak, line 173: All breadcrumbs generated today (and for 'active' sessions in the last 28 days) will have a `seq`; in order for tests relying on this mock to now pass, we need to bring it up to date

apps/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/preview/removeNodesDependentOnPassport.test.ts
```diff
@@ -269,7 +269,6 @@ describe("nodesDependentOnPassport with previousCard", () => {
     const breadcrumbs = {
       findProperty: breadcrumbsDependentOnPassport.findProperty,
       text: breadcrumbsDependentOnPassport.text,
-      text2: breadcrumbsDependentOnPassport.text2,
```
jessicamcinchak, line 272: This was an impossible scenario based on the corresponding mock flow, there are many non auto-answerable questions between 'text' and 'text2' and we would never generate breadcrumb "holes" like mocked here IRL. These changes made this test fail as it should have been!

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -188,33 +188,32 @@ export const previewStore: StateCreator<
   },
 
   previousCard: (node: Store.Node | null) => {
-    const { breadcrumbs, flow, _nodesPendingEdit, changedNode } = get();
+    const { breadcrumbs, _nodesPendingEdit, changedNode } = get();
+
+    // Do not calculate "previous" if "changing" an answer rather than going "back"
+    if (changedNode && node?.id === changedNode) return;
```
jessicamcinchak, line 190: In this scenario, a "change" link already encodes the exact node ID it needs to navigate to, we don't need to calculate it against our current position like here !

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -188,33 +188,32 @@ export const previewStore: StateCreator<
...
+
+    // Only "go back" to nodes which were visible to the user (not automated) while navigating forwards
     const goBackable = Object.entries(breadcrumbs)
       .filter(([, v]) => !v.auto)
       .map(([k]) => k);
-    if (changedNode && node?.id === changedNode) return;
 
+    // Nodes pending edit capture dependencies like FindProp
```
jessicamcinchak, line 197: I honestly still don't fully understand this pre-existing `return` case, but tests in `removeNodesDependentOnPassport` fail without it ! One to look into further "later" ?

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -188,33 +188,32 @@ export const previewStore: StateCreator<
...
-    }
+    // Breadcrumbs are sequenced, find the most recent one which is eligible to go back to
+    const maxGoBackSeq = Math.max(
+      ...Object.entries(breadcrumbs)
+        .filter(([k]) => goBackable.includes(k))
+        .map(([, v]) => v.seq!),
+    );
+    const goBackId = Object.entries(breadcrumbs).filter(
+      ([, v]) => v.seq === maxGoBackSeq,
+    )?.[0]?.[0];
```
jessicamcinchak, line 210: This is the most important bit of the refactor:
- Now that `breadcrumbs` have a "sequence" directly written to them, we can simply find the max sequence among eligible breadcrumbs (eg subset seen by user, not automated)
- We no longer need to join back to `flow` via `sortIdsDepthFirst` which is notoriously very expensive and leads to the clone ordering issue in the first place

As sorting/filtering on `seq` is far less expensive, I've dropped the `memoization` here as well - any concerns about this?

apps/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/preview/backNavigationWithClones.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    // Manually call setCurrentCard() as we're not using record() as part of our setup
+    setCurrentCard();
+
+    // Begin going backwards from FinalNotice and confirm "back" button is visible
+    expect(getCurrentCard()?.id).toBe("FinalNotice");
+    expect(canGoBack(getCurrentCard())).toBe(true);
+
+    expect(previousCard(flow["FinalNotice"])).toBe("ClonedContent"); // currently returns "RightPathFirstQuestion"
+    // expect(previousCard(flow["ClonedContent"])).toBe("RightPathFirstQuestion");
+    // expect(previousCard(flow["RightPathFirstQuestion"])).toBe("RootQuestion"); // Skips ClonedSetValue because was automated
```
jessicamcinchak: ‼️ Fun twist ! We're now calculating the _correct_ `previousCard` ID, but the actual "back" click is still skipping it 🫠 🙈 

Looking into this next ! (Suspect other store methods which rely on `previousCard`?)

**👇 Production returns the ID of the question _and_ navigates to it**
Image
**👇 Locally, I'm returning the correct ID of the cloned content node now, but still _actually_ navigating to the question prior**
Image

jessicamcinchak: Okay sorted this! 
- We call `previousCard` to _calculate_ the correct node ID to go back to
- But then `record()` is called on the previous card ID to actually _navigate_ to it
  - `record()` calls `sortBreadcrumbs` which was still relying on a join to the `flow` and `sortIdsDepthFirst`, which re-introduces the clone ordering problem! 
  - By also updating `sortBreadcrumbs` to sort via `seq`, then our navigation _matches_ our calculation :white_check_mark:

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
```diff
@@ -984,15 +978,13 @@ export const removeOrphansFromBreadcrumbs = ({
...
 ) => {
   return editingNodes?.length
     ? nextBreadcrumbs
-    : sortIdsDepthFirst(flow)(new Set(Object.keys(nextBreadcrumbs))).reduce(
-        (acc, id) => ({ ...acc, [id]: nextBreadcrumbs[id] }),
-        {} as Store.Breadcrumbs,
-      );
+    : (Object.fromEntries(
+        Object.entries(nextBreadcrumbs).sort(([, a], [, b]) => a.seq! - b.seq!),
+      ) as Store.Breadcrumbs);
```
jessicamcinchak, line 987: In order to introduce `seq` on _existing_ breadcrumbs without a migration of historic session data, the type is `number | undefined`. But this change has been live for long enough that all active session `breadcrumbs` should always have this so I'm not expecting the `seq!` to cause any issues here !

jessicamcinchak: As a follow-up to this PR, I'll also update `sortBreadcrumbs` in planx-core next: https://github.com/theopensystemslab/planx-core/blob/main/src/models/session/logic.ts#L69

This planx-new change fixes 'back' navigation and order of previous answers on Result & Resume/Reconciliation pages. Until planx-core change, schema `responses` and overview.html anwers may also display the wrong order of cloned content.

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
freemvmt: oh nice so this is just hoisted rather than further down!

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
freemvmt: I don't really understand your comment, what is a 'dependency' here?

to my understanding, this is invoked if there are some non-zero number of 'nodes pending edit', and we are either on the 1st node (i.e. no breadcrumbs generated yet because no answers given) or 2nd node (so 1 non-automated breadcrumb) which the user has seen?

and if that's the case then we get sent back by 1 node, which probably means back to the start?

but also if `goBackable.length === 0`, which seems to be allowed here, then `goBackable.pop() === undefined`? I don't understand this 😅

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
freemvmt: brilliant! agree memoization seems more work than it's worth, and not used elsewhere right

what happens if `goBackable === []`? can that happen?

apps/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/preview/removeNodesDependentOnPassport.test.ts
freemvmt: reassuring!

apps/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/preview/previousCard.test.ts
```diff
@@ -40,8 +40,8 @@ describe("store.previousCard is", () => {
   test("the most recent human-answered card id", () => {
     setup({
       breadcrumbs: {
-        a: { auto: false }, // a human answered this card
-        b: { auto: true },
+        a: { auto: false, seq: 1 }, // a human answered this card
+        b: { auto: true, seq: 2 },
```
freemvmt, line 44: ah so auto-answered breadcrumbs also get a `seq`, makes sense

apps/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/preview/backNavigationWithClones.test.ts
```diff
@@ -0,0 +1,232 @@
...
+    resetPreview();
+    setState({ flow });
+  });
+
+  test("It navigates forwards in the expected order", () => {
+    expect(upcomingCardIds()).toEqual(["RootQuestion", "FinalNotice"]);
+
+    // Answer the first question
+    clickContinue("RootQuestion", { auto: false, answers: ["RightPath"] });
+    // The clone has been incorrectly hoisted from the left branch into first position
```
freemvmt, line 28: ok so this test currently passes not because it works as we would like, but because we are encoding an expected/well-understood problem? gotcha

freemvmt (approved):
Awesome work here Jess!

I think it all makes sense, as far as my understanding/comprehension goes ✨

apps/editor.planx.uk/src/@planx/components/Review/Public/Presentational.tsx
```diff
@@ -25,7 +25,6 @@ function Component(props: Props) {
   // ensure questions & answers display in expected order
   const sortedBreadcrumbs: Store.Breadcrumbs = sortBreadcrumbs(
     props.breadcrumbs,
-    props.flow,
```
DafyddLlyr, line 28: Very nice that these can now be sorted independently of the flow! 🙌

apps/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/preview/backNavigationWithClones.test.ts
```diff
@@ -0,0 +1,232 @@
...
+      "ClonedSetValue",
+      "RightPathFirstQuestion", // Answer left path
+      "ClonedContent",
+      "FinalNotice",
+    ];
+    expect(visitedNodes().concat(upcomingCardIds())).toEqual(forwardsSequence);
+
+    // In a future where clones are fixed, forwards order _should_ be:
+    // ["RootQuestion", "RightPathFirstQuestion", "ClonedContent", "ClonedSetValue", "FinalNotice"]
+  });
```
DafyddLlyr, line 54: Thanks for adding this..!

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
DafyddLlyr, line 987: Lodash is nice here for sorting objects - just syntactic sugar that makes this a bit more readable.


```suggestion
    : sortBy(nextBreadcrumbs, ["seq"]) as Store.Breadcrumbs);
```

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
DafyddLlyr: Agree that it's nice to drop the memoisation here - it's just adding complexity that I think we no longer need.

DafyddLlyr: > what happens if goBackable === []? can that happen?

This is accounted for in the function signature which returns `NodeId | undefined` already. The return type is checked by consumers, and access to the function is similarly guarded by the UI in places (e.g. back button is hidden on essentially the same condition).

DafyddLlyr (approved):
Great..! Thanks for detailed write up + testing and coordinating with content team 🚀

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
jessicamcinchak: Dependencies are a good question - it's essentially the situation where: 
- If I change my address via FindProperty component, then that change needs to immediately invalidate my answers to DrawBoundary, PlanningConstraints or anything else that was _based on my address_ 
- Because "back" navigation is step-by-step, you wouldn't get to FindProperty without first stepping back through PlanningConstraints &rarr; DrawBoundary _before_ FindProperty anyways, the the direct "change" navigation route from Review pages would skip that nav and take you direct to FindProp
- I think this is where the `_nodesPendingEdit` come in - the ones that have been invalidated and now need to be re-seen fresh (not with prior answer auto-populated like usual when navigating forwards again after going back)

Per testing, behavior has _not_ broken or changed here so we should be good ! But it's a quite old part of the codebase that will benefit from a re-visit soon I suspect (eg both to revisit general logic/optimisation and because new component types like MapAndLabel, List/Page with map field may be considered new "dependencies" that should also be handled)

jessicamcinchak: Daf has a good thorough answer for `goBackable` length below - short answer is we handle those cases in the UI of `<Questions />` component, hide "back" button etc :slightly_smiling_face:

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
jessicamcinchak: This looks tidy, but types very much don't match up or work as suggested here! Fiddled with for a bit, but not straightforward with nested `Object.entries` - going to merge as is!

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts
freemvmt: @jessicamcinchak this was super helpful, thanks for taking the time to explain!

PR #6848, opened 2026-06-12
infra: bump Hasura memory and cpu on production

jessicamcinchak:
Follows on from https://github.com/theopensystemslab/planx-new/pull/6834 ahead of prod deploy

ShareDB not set via yaml, so would _already_ apply to all envs here.


ianjon3s (approved):

infrastructure/application/Pulumi.production.yaml
```diff
@@ -47,11 +47,11 @@ config:
   application:hasura-admin-secret:
     secure: AAABAFXUOfERqmY2ht+UN/6j7fNncuB3uwqeEGGUhXzHLXLt5Mu7iGbuLfjMpRC0Niw301Zejtc=
   application:hasura-cpu: "1024"
```
freemvmt: I think we doubled CPU as well?

freemvmt (changes requested):

infrastructure/application/Pulumi.production.yaml
jessicamcinchak: yes good catch :sweat_smile: now included!

freemvmt (approved):
niiiice 🚀

PR #6849, opened 2026-06-12
onboard Stockport for custom domain

freemvmt:
As per title ([ticket](https://trello.com/c/573YBL7y)). Also fixes slug for Coventry and adds reminder in howto about that.

### DIFF
--- a/doc/how-to/how-to-setup-custom-domains.md
+++ b/doc/how-to/how-to-setup-custom-domains.md
@@ -58,7 +58,7 @@ Path: `validation-only` → `shared-final`
     },
     ```
 
-    Then commit this change. Note that `certificateLocation` is not necessary here.
+    Here `name` should match the `teams.slug` value in the [db](https://hasura.editor.planx.uk/console/data/default/schema/public/tables/teams/browse). Note also that `certificateLocation` is not necessary here. Then commit this change.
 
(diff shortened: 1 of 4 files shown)


doc/how-to/how-to-setup-custom-domains.md
```diff
@@ -58,7 +58,7 @@ Path: `validation-only` → `shared-final`
     },
     ```
 
-    Then commit this change. Note that `certificateLocation` is not necessary here.
+    Here `name` should match the `teams.slug` value in the [db](https://hasura.editor.planx.uk/console/data/default/schema/public/tables/teams/browse). Note also that `certificateLocation` is not necessary here. Then commit this change.
```
jessicamcinchak, line 61: This is such a tedious pain currently 😅  have a ticket to migrate / make more consistent (which is also going to be very tedious to do, but worth it I think!) https://trello.com/c/kkOc4lvJ/3631-improve-how-team-slug-maps-to-granular-article-4-data-field-to-ensure-future-submissions-validate

jessicamcinchak (approved):

doc/how-to/how-to-setup-custom-domains.md
freemvmt: ah nice, didn't know about that! wasn't actually sure whether the names here were important or not but I thought they be consistent for sanity's sake at minimum ;)

PR #6851, opened 2026-06-15
fix: don't restrict `Subsription` page access from `teamEditors` yet

jessicamcinchak:
Per report from South Glos here: https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1781519277542449

We don't want to restrict permissions here yet because we have _not_ actually assigned `team_members` "teamAdmin" roles per #6840! Everyone on the team is still a "teamEditor" and needs to maintain access in interim.

Just a minor sequencing glitch, we can revert this later !


jessicamcinchak:
Admin-merging this one as we're low capacity today and this is a user-facing bug!

PR #6853, opened 2026-06-15
[infra] add pki.goog + letsencrypt to CAA records

freemvmt:
Handles a TODO, motivated by edge certs for services on staging expiring in the coming week ([Slack](https://opensystemslab.slack.com/archives/C08H0AEN4DR/p1781395318805389)).

Basically the CAA record determines which Certificate Authority's (CAs) can provision certs for our domains (in this case `planx.dev` and `planx.uk`).

Our Cloudflare setup uses Google Trust Services for this on staging, and Let's Encypt on prod (I don't know why the diff). We had previously fixed this on prod by manually adding certs for the latter. I've already rolled out these changes to both environments, and deleted said manual certs in the process, so that these are all managed by Pulumi from the `certificates` layer.


jessicamcinchak (approved):
Thanks for explanation and sorting !

PR #6856, opened 2026-06-16
fix: archive & delete handling

zz-hh-aa:
- Fixes a typo (`refesh` -> `refresh`)
- Jonny also hit the edge case of duplicate archived slugs--this made me realise that I hadn't fully implemented Daf's suggestion from before (append `-archived` or `-deleted` [_until_ unique](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1776854434544869?thread_ts=1776852554.614039&cid=C01E3AC0C03)). But given the possibility of slugs that look like `test-deleted-deleted-deleted-deleted-deleted`, maybe it's just neater to append a date? Thoughts welcome!

### DIFF
--- a/apps/editor.planx.uk/src/pages/Team/components/ActiveFlowMenu.tsx
+++ b/apps/editor.planx.uk/src/pages/Team/components/ActiveFlowMenu.tsx
@@ -35,7 +35,7 @@ const ActiveFlowMenu: React.FC<FlowMenuProps> = ({
       toast.success("Archived flow");
     } catch (error) {
       toast.error(
-        "We are unable to archive this flow, refesh and try again or contact an admin",
+        "We are unable to archive this flow, refresh and try again or contact an admin",
       );
     } finally {
       setOpenDialog(null);
--- a/apps/editor.planx.uk/src/pages/Team/components/ActiveFlowMenu.tsx
+++ b/apps/editor.planx.uk/src/pages/Team/components/ActiveFlowMenu.tsx
@@ -21,7 +21,7 @@ const ActiveFlowMenu: React.FC<FlowMenuProps> = ({
   type OpenDialog = "archive" | "copy" | "rename" | "move";
   const [openDialog, setOpenDialog] = useState<OpenDialog | null>(null);
   const dateTime = Date.now();
-  const archivedSlug = flow.slug.concat(`-archived${dateTime}`);
+  const archivedSlug = flow.slug.concat(`-archived-${dateTime}`); // this is just to make the slug unique, not as an actual timestamp
   const [archiveFlow] = useArchiveFlow(flow.id, archivedSlug, teamId);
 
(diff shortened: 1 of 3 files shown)


apps/editor.planx.uk/src/pages/Team/components/ActiveFlowMenu.tsx
```diff
@@ -20,7 +20,8 @@ const ActiveFlowMenu: React.FC<FlowMenuProps> = ({
 }) => {
   type OpenDialog = "archive" | "copy" | "rename" | "move";
   const [openDialog, setOpenDialog] = useState<OpenDialog | null>(null);
-  const archivedSlug = flow.slug.concat("-archived");
+  const dateTime = Date.now();
+  const archivedSlug = flow.slug.concat(`-archived${dateTime}`);
```
DafyddLlyr: Oh yeah this is much neater and clevered 😅 👏 

Maybe just worth a comment here to say that this is used as a unique identifier and not a real "timestamp".

apps/editor.planx.uk/src/pages/Team/components/ArchivedFlowMenu.tsx
```diff
@@ -23,10 +23,11 @@ const ArchivedFlowMenu: React.FC<FlowMenuProps> = ({
     null,
   );
 
-  const unarchivedSlug = flow.slug.replace("-archived", "");
+  const unarchivedSlug = flow.slug.replace(/-archived\d+$/, "");
   const [unarchiveFlow] = useUnarchiveFlow(flow.id, unarchivedSlug, teamId);
 
-  const deletedSlug = flow.slug.replace("-archived", "-deleted");
+  const dateTime = Date.now();
+  const deletedSlug = flow.slug.replace(/-archived\d+$/, `-deleted${dateTime}`);
```
DafyddLlyr: nit: Maybe an additional `-` here will also be a bit more readable for both archive and delete cases?


```suggestion
  const deletedSlug = flow.slug.replace(/-archived\d+$/, `-deleted-${dateTime}`);
```

DafyddLlyr (approved):
I've not checked out CI failures, but if these slugs are used in tests anywhere you might need to call on `vi.useFakeTimers()` ([docs](https://vitest.dev/guide/mocking/timers)).

apps/editor.planx.uk/src/pages/Team/components/ActiveFlowMenu.tsx
zz-hh-aa: Credits to Jonny for the clever idea!

Good shout on both counts, both little changes made in [6f778ed](https://github.com/theopensystemslab/planx-new/pull/6856/commits/6f778edb73f774dcf6db01b6c61ae6ddea7b3948)

PR #6859, opened 2026-06-16
feat[dashboard]: link feedback items directly to the service it concerns

joecarver:
As per Silvia's suggestion, links each feedback item to the service feedback page. Required fetching the service slug with the feedback items.

### DIFF
--- a/apps/editor.planx.uk/src/pages/Dashboard/components/useUnreadFeedback.tsx
+++ b/apps/editor.planx.uk/src/pages/Dashboard/components/useUnreadFeedback.tsx
@@ -10,12 +10,12 @@ export interface FlowSummary {
 
 export const useUnreadFeedback = (teamSlug: string) => {
   const { data, loading, error } = useQuery<{
-    feedback_summary: { flowName: string; flowSlug: string }[];
+    feedbackSummary: { flowName: string; flowSlug: string }[];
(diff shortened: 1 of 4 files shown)


jessicamcinchak (approved):
Looks good!

apps/editor.planx.uk/src/pages/Dashboard/components/useUnreadFeedback.tsx
```diff
@@ -4,26 +4,29 @@ import { GET_UNREAD_FEEDBACK_SUMMARY } from "./queries";
...
 export interface FlowSummary {
   flowName: string;
+  flowSlug: string;
   count: number;
 }
 
 export const useUnreadFeedback = (teamSlug: string) => {
   const { data, loading, error } = useQuery<{
-    feedback_summary: { flowName: string }[];
+    feedback_summary: { flowName: string; flowSlug: string }[];
```
zz-hh-aa: nit: think we want to avoid snake case here (and below in `aggregated`) / rename to camel in the query! (only noticing cause Daf [just spotted](https://github.com/theopensystemslab/planx-new/pull/6840#discussion_r3419862049) me doing this)

zz-hh-aa (approved):
Works as expected and all round looks great ✨

PR #6863, opened 2026-06-17
chore: migrate certain teamEditors to teamAdmin

zz-hh-aa:
This PR
- Creates migrations to promote select editors to team admin
- Re-gates the subscription page (following early gating and revert)

Selected teams and users are from the spreadsheet shared by Silvia of either editors that should be made admins, or teams that should be promoted wholesale (please feel free to check my work!) https://opensystemslab.slack.com/archives/C01E3AC0C03/p1779975749440209

To-do 2/2 from #6840


apps/hasura.planx.uk/migrations/default/1781694636453_migrate_teams_to_team_admin/down.sql
```diff
@@ -0,0 +1,3 @@
+DELETE FROM team_members    
+WHERE team_id IN (22, 27, 65, 36, 3, 50, 31, 37, 6, 28, 25)
+  AND role = 'teamAdmin';
```
zz-hh-aa, line 3: No `teamAdmin` exist in the db yet so figured this works as a down migration

DafyddLlyr (approved):

PR #6864, opened 2026-06-17
chore: add flow ID to "Failed to publish flow" error messages

jessicamcinchak:
Would be helpful to have more context on these in light of latest GDPO bugs - see example https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1781693402979589

### DIFF
--- a/apps/api.planx.uk/modules/flows/publish/controller.ts
+++ b/apps/api.planx.uk/modules/flows/publish/controller.ts
@@ -32,9 +32,10 @@ export const publishFlowController: PublishFlowController = async (
   res,
   next,
 ) => {
+  const { flowId } = res.locals.parsedReq.params;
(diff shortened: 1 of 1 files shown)


apps/api.planx.uk/modules/flows/publish/controller.ts
```diff
@@ -32,9 +32,10 @@ export const publishFlowController: PublishFlowController = async (
   res,
   next,
 ) => {
+  const { flowId } = res.locals.parsedReq.params;
+  const { summary, templatedFlowIds } = res.locals.parsedReq.body;
```
jessicamcinchak, line 36: We have a zod schema for these, therefore they should be safe to keep outside of the `try` block!

freemvmt (approved):
great :)

PR #6866, opened 2026-06-17
chore: update template copiable permissions

zz-hh-aa:
As [discussed in Slack](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1781009706080469), we want templatable flows to be controlled via copy permissions, and not the previous `status = online` model (since that now requires `is_service = true` and not all templates will be services). 

This PR:
- Sets all online, `is_service` templates to `can_create_from_copy = true`
- Updates the templatable flows to search on this column instead
- Updates the copiable flows query to filter out templates

Questions for @augustlindemer:
-  I think after this merges to prod, what we want is for the service team to go through and manually remove `is_service = true` for relevant templates (eg Article 4s), does that sound right to you? 
- Is using the flow copy component (as is) to control template functionality confusing / do you want any updates to the UI here? 
image


jessicamcinchak (approved):
Code and DB changes look good here, thanks for sorting this logic! 

Will defer to August on all open questions you've listed and final testing approval 🙂

PR #6867, opened 2026-06-17
a11y: public-facing fixes

joecarver:
Fixes the key public facing violations that were flagged:
- A few missing `aria-` props
- fix `ImgInput` nested interactive warning by moving the Tooltip inside it
- rework `PublicFileUploadButton` to fix further `no-nested-interactive` warnings
- rework `MoreInformation` so it still works with new `ImgInput` style

### DIFF
--- a/apps/editor.planx.uk/src/ui/editor/MoreInformation/MoreInformation.tsx
+++ b/apps/editor.planx.uk/src/ui/editor/MoreInformation/MoreInformation.tsx
@@ -51,7 +51,14 @@ export const MoreInformation = <T extends BaseNodeData>({
             />
           </InputLabel>
           {/* ImgInput must be outside InputLabel: role="button" inside <label> = nested-interactive */}
-          <Box sx={{ display: "flex", gap: "5px", alignItems: "flex-end" }}>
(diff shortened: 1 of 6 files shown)


apps/editor.planx.uk/src/ui/editor/MoreInformation/MoreInformation.tsx
```diff
@@ -44,29 +44,32 @@ export const MoreInformation = <T extends BaseNodeData>({
...
-                  formik.handleChange({
-                    target: { name: "definitionImg", value: newUrl },
-                  });
-                }}
-                disabled={disabled}
-              />
-            </InputRow>
-          </InputLabel>
+          {/* ImgInput must be outside InputLabel: role="button" inside <label> = nested-interactive */}
+          <Box sx={{ display: "flex", gap: "5px", alignItems: "flex-end" }}>
```
ianjon3s: I think this just needs a 100% width property to maintain alignment

Image

ianjon3s (approved):
One small comment about input grouping width, but otherwise really solid changes!

PR #6878, opened 2026-06-18
fix(api): Single DOMPurify instance per-request

DafyddLlyr:
This is a follow up to #6860 which is attempting to improve performance here.

This PR converts the DOMPurify instance from a module level singleton to a factory pattern - this instantiates and closes one instance per-request to the API.

When profiling this on the GPDO flow, I noticed that the current implementation just bloats memory usage and slows down per-iteration. This could be the underlying cause of the issues here - even if not, it's worth resolving and ruling out.

```
Shared instance (current pattern):
  round 0:   2049 ms | heapUsed 176 MB
  round 1:   9464 ms | heapUsed 268 MB
  round 2:   9269 ms | heapUsed 396 MB
  round 3:  15261 ms | heapUsed 474 MB
  round 4:  21077 ms | heapUsed 557 MB
  round 5:  25421 ms | heapUsed 685 MB     ← ~12× slower than round 0

Fresh instance per round (this PR):
  round 0:   1246 ms
  round 1:   1089 ms
  round 2:   1688 ms
  round 3:   2260 ms
  round 4:   1371 ms
  round 5:   1021 ms                       ← flat, no upward trend
  ```

### DIFF
--- a/apps/api.planx.uk/modules/webhooks/service/validateInput/schema.ts
+++ b/apps/api.planx.uk/modules/webhooks/service/validateInput/schema.ts
@@ -1,6 +1,6 @@
 import { z } from "zod";
 import type { ValidatedRequestHandler } from "../../../../shared/middleware/validate.js";
-import { isCleanHTML, isObjectValid } from "./utils.js";
+import { HTMLSanitiser, isObjectValid, makeIsCleanHTML } from "./utils.js";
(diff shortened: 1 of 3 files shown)


jessicamcinchak (approved):
Makes sense, thanks for this one!

PR #6879, opened 2026-06-18
fix(api): Do not re-validate source template data in insert or update

DafyddLlyr:
## What does this PR do?

This is a follow-up to #6860.

Creating or updating a flow from a source template re-ran the `clean-html` `validate_input` webhook against the entire `flows.data` JSONB. For template operations this validation is entirely redundant - the data comes straight from `getFlowData()` (a DB read of already-sanitised data), so it always passes. Running it repeatedly  on large flows causes memory spikes.

This PR stops that redundant validation, and re-establishes the authorisation pattern currently in place.

## Approach

The Hasura input validation in Hasura is configured per role + table + operation, so it can't be skipped per-query within a given role securely (e.g. `skipValidation` query param). Instead we route the template insert/update through the `api` role, and remove the input validation from that role's `flows` insert and update permissions.

## Trade-offs

The `api` role has no Hasura row-level check, so switching to it dropped two guarantees that the user role enforced at the DB layer (can the user insert into this team?). Both are re-added - 

1. `creator_id` - previously set via the `set: creator_id: x-hasura-user-id` preset. The `api` role's JWT carries no `x-hasura-user-id`, so `creator_id` is now set explicitly in the insert from the authenticated user context.

2. Per-team authorisation - previously enforced by the `teamEditor` row check (`team.members.user_id == x-hasura-user-id`). The route guard `useTeamEditorAuth` only checks the global role claim, not membership of the target team, so this would otherwise allow a `teamEditor` of team A to create a flow into team B. A new `requireTeamMembership` middleware refills this gap in the API layer. It runs after `validate`, reads the typed request, and authorises platformAdmins plus `teamEditor`/`teamAdmin` members of the target team.

Regression tests passing here - https://github.com/theopensystemslab/planx-new/actions/runs/27940200368 ✅

### DIFF
--- a/apps/api.planx.uk/modules/auth/requireTeamMembership.ts
+++ b/apps/api.planx.uk/modules/auth/requireTeamMembership.ts
@@ -1,0 +1,52 @@
+import type { NextFunction, Request, Response } from "express";
+
+import { $api } from "../../client/index.js";
+import { userContext } from "./middleware.js";
+
+/**
+ * Authorise that the current user can edit the team targeted by a request
+ *
+ * `useRoleAuth` only checks the user's global role claim, leaving per-team
+ * authorisation to Hasura's row-level check (are they a teamEditor within *this* team?).
+ * Routes that mutate via the `api` role (which has no row-level check) lose that
+ * enforcement, so this middleware refills the permission check in the API layer.
+ *
+ * This middleware should be used sparingly - only when it's not possible to rely on
+ * Hasura row-level checks (e.g. to skip redundant and memory intensive input validation
+ * for template INSERT actions)
+ */
+export const requireTeamMembership =
+  <T>(getTeamId: (parsedReq: T) => number) =>
+  async (_req: Request, res: Response, next: NextFunction) => {
+    const userId = userContext.getStore()?.user?.sub;
+    if (!userId)
+      return next({
+        status: 403,
+        message: "Access denied - userId missing from request",
+      });
+
+    const teamId = getTeamId(res.locals.parsedReq as T);
+
+    const user = await $api.user.getById(Number(userId));
+    if (!user)
+      return next({
+        status: 403,
+        message: `Access denied - unable to find user matching ID ${userId}`,
+      });
+
+    const isUserInTeam = user.teams.some(
+      ({ team, role }) =>
+        team.id === teamId && (role === "teamEditor" || role === "teamAdmin"),
+    );
+
+    const isAuthorised = user.isPlatformAdmin || isUserInTeam;
+
+    if (!isAuthorised) {
+      return next({
+        status: 403,
+        message: `Access denied. User ${userId} is not a member of team ${teamId} with permission to edit it`,
+      });
+    }
+
+    return next();
+  };
--- a/apps/editor.planx.uk/src/components/Breadcrumbs.tsx
+++ b/apps/editor.planx.uk/src/components/Breadcrumbs.tsx
@@ -54,7 +54,7 @@ const Breadcrumbs: React.FC = () => {
 
   const canUserEditTeam = useStore((state) => state.canUserEditTeam);
 
-  if (!teamSlug || !flowSlug || !data?.flows) return null;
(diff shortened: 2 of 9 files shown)


DafyddLlyr:
Converted to draft to take a look at e2e failures.

e2e/tests/ui-driven/src/templates.spec.ts
```diff
@@ -584,88 +584,93 @@ test.describe("Templates", () => {
...
   });
 
-  // relies on the templated flow being created in the "As a team editor" block
-  test.describe("As a platform admin", () => {
-    test("can publish changes to the source template and sees affected templated flows listed", async ({
-      browser,
-    }) => {
-      const page = await getTeamPage({
+  // These tests rely on the resources created in the above blocks and must sequentially run later
+  test.describe("Hybrid actions", () => {
```
DafyddLlyr: Playwright is actually collapsing the repeated `describe()` blocks here meaning that despite being at the bottom of the file, these both fail on first pass and then re-run - 

**Current run**
 - Platform admin tests run and pass ✅
 - Final platform admin test fails, there's no derived template which is required ❌ 
 - Team admin tests run and pass ✅
 - Final team admin test fails, there's no re-publish event which is required ❌ 
 - Final platform admin test passes on a re-run - resource in place ✅ 
 - Final team admin test passes on a re-run - re-publish has happened ✅ 

**This PR**
 - Platform admin tests run and pass ✅ 
 - Team admin tests run and pass ✅ 
 - "Hybrid" tests run last, and pass first time ✅ 

As far as I can tell this is only within the VSCode UI - these tests run in order and all pass first time with the CI tool. Making this change though makes these easier to understand and debug locally and makes the layout and intention a little clearer within this file.

Image

_Single `describe()` block on `main`, despite two in file_

apps/hasura.planx.uk/metadata/databases/default/tables/public_flows.yaml
```diff
@@ -153,20 +151,12 @@ insert_permissions:
         - name
         - settings
         - slug
+        - status
```
DafyddLlyr, line 154: Missing permission for this column, required for this role to write.

apps/hasura.planx.uk/metadata/databases/default/tables/public_flows.yaml
```diff
@@ -137,8 +137,6 @@ insert_permissions:
   - role: api
     permission:
       check: {}
-      set:
-        creator_id: x-hasura-user-id
```
DafyddLlyr, line 141: Incorrectly configured - the `api` role will never have an attached `x-hasura-user-id`.

apps/editor.planx.uk/src/components/Breadcrumbs.tsx
```diff
@@ -54,7 +54,7 @@ const Breadcrumbs: React.FC = () => {
 
   const canUserEditTeam = useStore((state) => state.canUserEditTeam);
 
-  if (!teamSlug || !flowSlug || !data?.flows) return null;
+  if (!teamSlug || !flowSlug || !data?.flows.length) return null;
```
DafyddLlyr, line 57: Small visual bug noticed when working on e2e tests - if `data.flows` is a `[]`, `flowStatus` below fails to be destructured.

e2e/tests/ui-driven/src/templates.spec.ts
jessicamcinchak: Tedious but make much more sense now / better mental model match!

apps/api.planx.uk/modules/auth/requireTeamMembership.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    if (!isAuthorised) {
+      return next({
+        status: 403,
+        message: `Access denied. User ${userId} is not a member of team ${teamId} with permission to edit it`,
+      });
+    }
+
+    return next();
+  };
```
jessicamcinchak, line 52: Very clear code and helpful comment ➕

jessicamcinchak (approved):

PR #6880, opened 2026-06-18
chore(editor): Additional logging for connectToFlow route

DafyddLlyr:


jessicamcinchak (approved):
Thanks! Shall we explicitly try to pass a flow id/slug here too, or safe to assume that'll always be in `error` ?

DafyddLlyr:
> Thanks! Shall we explicitly try to pass a flow id/slug here too, or safe to assume that'll always be in error ?

As this is just a console log (for now) this information will always be caught in the URL. If we upgrade to an Airbrake error (once we confirm it's recreatable in a meaningful way) then the flow slug and team slug will also get handled automatically as context via the URL 👍

PR #6883, opened 2026-06-18
feat[db]: create internal audit view `daily_failed_submissions`

jessicamcinchak:
See https://trello.com/c/JdLmMcxP/3645-implement-a-nightly-cron-job-to-check-for-failed-submission-events-without-a-successful-retry-resubmit

Airbrake errors remain the first line of defense for spotting submission errors. But during a recent blip of API downtime, we failed to correctly log failed events for a submission to Lambeth. 

A few days passed before a Lambeth editor noticed and asked for it to be resubmitted (it _was_ successfully sent to email the original day, but failed to send to 3x other destinations which Lambeth uses to dual-process). We want to introduce a second-line of defense for more proactively spotting these errors.

On production, this view returns this (as expected; redundant with Airbrake when all systems in working order!):
Screenshot from 2026-06-18 18-54-40

**Next steps:**
- Let's decide how we want to access this view
  - Hasura cron event/webhook which queries the view and if >= 1 row is returned, logs it to #planx-notifications-internal via `postToSlack`?
  - Metabase bot maybe? The view itself isn't PPI, can we configure non-ODP channel for that bot?
  - Just manually check if we experience API downtime to not create redundant noise on most days?

### DIFF
--- a/apps/hasura.planx.uk/migrations/default/1781823231755_create_view_daily_failed_submissions/up.sql
+++ b/apps/hasura.planx.uk/migrations/default/1781823231755_create_view_daily_failed_submissions/up.sql
@@ -1,0 +1,25 @@
+CREATE OR REPLACE VIEW daily_failed_submissions AS (
+	with send_event_statuses_per_session as (
+			select
+					session_id,
+			event_type,
+			flow_name,
+			team_id,
+					array_agg(status) as statuses,
+			max(created_at) as last_event
+			from submission_services_log
+			where event_type != 'Pay'
+			and created_at >= NOW() - INTERVAL '24 hours'
+			group by session_id, event_type, flow_name, team_id
+	)
+	select 
+		t.name,
+		s.flow_name,
+		s.session_id, 
+		s.event_type,
+		s.statuses,
+		s.last_event
+	from send_event_statuses_per_session s
+		join teams t on s.team_id = t.id
+	where 'Success' != ALL(statuses)
+);
--- a/apps/hasura.planx.uk/metadata/databases/default/tables/public_daily_failed_submissions.yaml
+++ b/apps/hasura.planx.uk/metadata/databases/default/tables/public_daily_failed_submissions.yaml
@@ -1,0 +1,3 @@
+table:
+  name: daily_failed_submissions
+  schema: public
--- a/apps/hasura.planx.uk/metadata/databases/default/tables/tables.yaml
+++ b/apps/hasura.planx.uk/metadata/databases/default/tables/tables.yaml
@@ -19,6 +19,7 @@
 - "!include public_application_access_tokens.yaml"
 - "!include public_blpu_codes.yaml"
 - "!include public_bops_applications.yaml"
+- "!include public_daily_failed_submissions.yaml"
 - "!include public_downtime_banner.yaml"
 - "!include public_email_applications.yaml"
 - "!include public_email_events_log.yaml"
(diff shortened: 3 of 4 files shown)


apps/hasura.planx.uk/migrations/default/1781823231755_create_view_daily_failed_submissions/up.sql
```diff
@@ -0,0 +1,25 @@
...
+	select 
+		t.name,
+		s.flow_name,
+		s.session_id, 
+		s.event_type,
+		s.statuses,
+		s.last_event
+	from send_event_statuses_per_session s
+		join teams t on s.team_id = t.id
+	where 'Success' != ALL(statuses)
```
jessicamcinchak, line 24: This is postgres' syntax for "array does _not_ contain". We do _not_ want to pick up events which failed but succeeded on auto retry, we only want to pickup unique event types which failed all together and remain unsubmitted !

apps/hasura.planx.uk/migrations/default/1781823231755_create_view_daily_failed_submissions/up.sql
```diff
@@ -0,0 +1,25 @@
...
+			select
+					session_id,
+			event_type,
+			flow_name,
+			team_id,
+					array_agg(status) as statuses,
+			max(created_at) as last_event
+			from submission_services_log
+			where event_type != 'Pay'
+			and created_at >= NOW() - INTERVAL '24 hours'
```
DafyddLlyr, line 12: How would this handle weekends, holidays etc - wouldn't a table containing any outstanding failures be more robust / less ephemeral?

DafyddLlyr (approved):
This is great to see - missed the trello ticket - this will be super useful! One small question to flag but perhaps this was previously discussed elsewhere.

Very much on board with this suggestion - 

> _Hasura cron event/webhook which queries the view and if >= 1 row is returned, logs it to `#planx-notifications-internal` via `postToSlack`?_

I'm not too sure how views work in Hasura when it comes to event triggers - I would guess that a new update that adds a record would not trigger an `INSERT` event. This means that we'll need to query the table on a schedule and then hit the API as you suggest 👍

zz-hh-aa:
Thanks for picking up Jess! Just chiming in to rule out the Metabot option--recent investigations showed that we can only connect it to one Slack workspace on our current plan, which is obviously already taken with weekly ODP updates.

apps/hasura.planx.uk/migrations/default/1781823231755_create_view_daily_failed_submissions/up.sql
jessicamcinchak: Yes fair question ! 

I'm currently thinking:
- Nightly CRON query, _not_ database event trigger, because setup as a view as you say!
- If >= 1 row in daily view, log details of this failure to #planx-notifications-internal (session ID, team + flow)
- Weekends/holidays handled exactly as any other bug triaging channels / SSL notifications etc? 

As this view is just a subset of `submission_services_log` db view which _is_ ephemeral, it should be easy enough to get to the robust details as long as session ID is in Slack message? 

A table of all outstanding failures would currently be noisy (>= 1 row rule would apply/log every day?) as we _never_ expect to resolve the Southwark Power Automate failures? (Honestly maybe this is the core issue and we just need to finally admin-publish to turn this off?)

DafyddLlyr: Yep this makes sense to me and will work well - we can always revisit and tweak if we find we want to explore an alternative. I think the logging to `#planx-notifications-internal` is exactly right for now - much less noisy, much more meaningful.

PR #6885, opened 2026-06-22
chore: automate discretionary dashboards on `is_service` [skip pizza]

zz-hh-aa:
Removing the unwieldy allow list of discretionary slugs in favour of using the db `flows.is_service` value. Skipped a Pizza here since the link generation checks the environment (to test, check out this branch and update `import.meta.env.VITE_APP_ENV === "production"` to `development`. Any service that has been online at any point should have analytics this way.


DafyddLlyr (approved):
Nice...!

PR #6886, opened 2026-06-22
chore(iac): Inital setup for East Riding certs

DafyddLlyr:
Step 1 here - https://github.com/theopensystemslab/planx-new/blob/main/doc/how-to/how-to-setup-custom-domains.md#a-onboarding-a-new-council

`certificates` layer not manually deployed, we should pair on this as an upskilling / onboarding session for @DafyddLlyr and @zz-hh-aa with @freemvmt 👌 🏫

### DIFF
--- a/infrastructure/common/customDomains.ts
+++ b/infrastructure/common/customDomains.ts
@@ -120,6 +120,11 @@ export const getCustomDomains = (env: string): CustomDomain[] =>
           domain: "planningservices.stockport.gov.uk",
           cloudFrontState: "validation-only",
         },
+        {
+          name: "east-riding-of-yorkshire",
+          domain: "planningservices.eastriding.gov.uk",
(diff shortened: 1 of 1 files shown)


freemvmt (approved):
looks good!

freemvmt (commented):
Just to note we can already bundle the application level changes mentioned in point 13 of the [onboarding guide](https://github.com/theopensystemslab/planx-new/blob/main/doc/how-to/how-to-setup-custom-domains.md#appendix-b-cloudfront-routing) in this PR if we like (although also fine to wait).

This can be merged to main or kept around as an open PR, but I think for the sake of consistency on `main` we should avoid having these custom domain PRs outstanding for too long.

This is an essentially harmless contribution because if this made it to prod in it's current state, nothing would happen!

PR #6887, opened 2026-06-22
fix: reload settings after ejecting from template

zz-hh-aa:
Noticed this little bug while doing some other templates settings work. Should now correctly redirect route / reload settings after ejecting from templates. 

Before:

https://github.com/user-attachments/assets/0ffd98a7-1c85-41a6-9851-aff607409a2a

After:

https://github.com/user-attachments/assets/10a30b04-256e-4df8-88e9-c2d524622460


DafyddLlyr (approved):
Nice, thanks for spotting and fixing 🛠️

PR #6894, opened 2026-06-23
feat: New component select for change component function

ianjon3s:
## What does this PR do?

- Refactors the editor modal so that the header bar is used for the component title and conditional "change" link
  - Component icon and title is moved to the header, making better use of available space and enabling a sticky component type header
  - On component creation (still "draft"), a change link appears

The above is in anticipation of our forthcoming work to allow saved components to be changed to a different component type.

**Testing:**
- Enable feature flag `COMPONENT_SELECT`
- Add a new component to a flow, the "change" link will be visible when the component is still draft, and the component title will persist in all states

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
@@ -159,7 +159,7 @@ const TextInputToggle: React.FC<{
   };
 
   return (
-    <Box sx={{ position: "absolute", right: 82, top: 30, zIndex: 1 }}>
+    <Box sx={{ px: 6, height: 0, py: 1.5 }}>
       <Switch
         label={
           <>
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
@@ -23,7 +23,7 @@ import {
   nodeIsChildOfTemplatedInternalPortal,
   nodeIsTemplatedInternalPortal,
 } from "pages/FlowEditor/utils";
-import React, { createContext, useMemo, useRef, useState } from "react";
+import React, { useMemo, useRef, useState } from "react";
 import { ErrorBoundary } from "react-error-boundary";
 import type { NodeSearchParams } from "routes/_authenticated/app/$team/$flow/_flowEditor/nodes/route";
 import { Switch } from "ui/shared/Switch";
@@ -33,7 +33,6 @@ import { fromSlug, SLUGS } from "../../data/types";
 import { useStore } from "../../lib/store";
 import ChangeComponentHeader from "./ChangeComponentHeader";
 
-export const ModalFormContext = createContext({ hideComponentTypeHeader: false });
 
 const StyledDialog = styled(Dialog)(({ theme }) => ({
   // Target all modal sections (the direct child is the backdrop, hence the double child selector)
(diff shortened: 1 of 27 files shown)


apps/editor.planx.uk/src/ui/editor/ModalSectionContent.tsx
```diff
@@ -69,11 +70,14 @@ export default function ModalSectionContent({
   author,
   Icon,
 }: Props): FCReturn {
+  const { hideComponentTypeHeader } = useContext(ModalFormContext);
+  const suppressHeader = hideComponentTypeHeader && Boolean(Icon) && Boolean(title);
```
ianjon3s: This will be simplified when the feature flag is dropped

DafyddLlyr (commented):
https://github.com/user-attachments/assets/8dcf3fe3-cb96-491e-a093-5ae15e006d86

Some quick visual regression feedback on the enhanced text input component.

Will follow up with a more thorough review in the morning 👍

ianjon3s:
Good catch @DafyddLlyr ! I'll update to unify the title and address the toggle positioning

ianjon3s:
@DafyddLlyr the toggle is now inline with the form, which feels like a more coupled interaction

https://github.com/user-attachments/assets/8c014409-72f0-4143-b24f-5caee3650dac

apps/editor.planx.uk/src/ui/editor/ModalSectionContent.tsx
DafyddLlyr:
```suggestion
  const suppressHeader = hasFeatureFlag("COMPONENT_SELECT") && Boolean(Icon) && Boolean(title);
```

Using context here just seems like a more complex way of checking the feature flag - can't we strip this out and check the following?

DafyddLlyr (changes requested):
Small question re: simplification of the use of `useContext()`

There's also still a visual issue with text input / enhanced text input on the Pizza. It was demoed in my above video but I didn't call it out specifically sorry - the toggle persists when the component is changed, meaning it remains "true" even when it shouldn't.

https://github.com/user-attachments/assets/6b276d9b-0cd1-4798-b721-4d2199e78db4

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
```diff
@@ -130,15 +130,14 @@ const TextInputToggle: React.FC<{
...
-  const [checked, setChecked] = useState(type === "enhanced-text-input");
   const { flow, team } = useParams({ from: "/_authenticated/app/$team/$flow" });
 
   if (!["text-input", "enhanced-text-input"].includes(type)) return null;
 
-  const toggleTextInput = () => {
-    // Toggle visual state immediately without waiting for route change
-    setChecked(!checked);
+  // Ensure toggle stays in sync with component state
+  const checked = type === "enhanced-text-input";
```
DafyddLlyr, line 137: Perfect!

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
```diff
@@ -17,19 +17,23 @@ import { type BaseNodeData, parseFormValues } from "@planx/components/shared";
...
 import { ErrorBoundary } from "react-error-boundary";
 import type { NodeSearchParams } from "routes/_authenticated/app/$team/$flow/_flowEditor/nodes/route";
 import { Switch } from "ui/shared/Switch";
 import { getNodeRoute } from "utils/routeUtils/utils";
 
 import { fromSlug, SLUGS } from "../../data/types";
 import { useStore } from "../../lib/store";
+import ChangeComponentHeader from "./ChangeComponentHeader";
+
+export const ModalFormContext = createContext({ hideComponentTypeHeader: false });
```
DafyddLlyr: nit: This is now unused as we're just checking the feature flag directly.

DafyddLlyr (approved):
Great...! Thanks for the quick changes. One small tidy up comment below and then good to go 🚀

ianjon3s:
Thanks for this feedback @DafyddLlyr , beat me to it. I've done a small refactor here, as it became apparent that the previous method was stripping the header from all instances of `ModalSectionContent`. Introducing `ModalComponentHeader` keeps a clear repeatable semantic structure. (had considered an `isModalHeader` prop on the first container, but easy to miss, or indeed misplace)

PR #6898, opened 2026-06-24
fix: Hide actions column for non team editors

ianjon3s:
## What does this PR do?

Quick one:

- Hides the action column in the flow table for non team editors or platform admins, rather than just the cells

**Before:**
image

**After:**
image


DafyddLlyr (approved):
Good catch 👍 

We probably want to expand the `<Permission/>` component to cover this common scenario at some point.

PR #6899, opened 2026-06-24
feat: use dropdown for team Role instead of switch

zz-hh-aa:
1/2 to-dos from #6840

v2 of #6855 after rebase problems

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/AddUserModal.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/AddUserModal.tsx
@@ -84,14 +84,9 @@ export const AddUserModal: React.FC<AddUserModalProps> = ({
                     bordered
                     required
                     data-testid="user-role-select"
-                    onChange={() =>
-                      formik.setFieldValue(
-                        "role",
-                        formik.values.role === roleOptions[0].id
-                          ? roleOptions[1].id
-                          : roleOptions[0].id,
-                      )
-                    }
+                    onChange={(e) => {
+                      formik.setFieldValue("role", e.target.value);
+                    }}
                   >
                     {roleOptions.map(({ id, name }) => (
                       <MenuItem
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/AddUserModal.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/AddUserModal.tsx
@@ -70,7 +70,12 @@ export const AddUserModal: React.FC<AddUserModalProps> = ({
                   <Box sx={{ mt: 2, mb: 2 }}>
                     <NameFields />
                   </Box>
-                  <Typography sx={{ pb: 0.5 }} variant="body2">
+                  <Typography
+                    component="label"
+                    id="User-role-label"
+                    sx={{ pb: 0.5 }}
+                    variant="body2"
+                  >
                     Role
                   </Typography>
                   <SelectInput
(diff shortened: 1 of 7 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/AddUserModal.tsx
```diff
@@ -61,19 +70,35 @@ export const AddUserModal: React.FC<AddUserModalProps> = ({
                   <Box sx={{ mt: 2, mb: 2 }}>
                     <NameFields />
                   </Box>
-                  <Switch
-                    name="role"
-                    checked={formik.values.role === "teamAdmin"}
+                  <Typography sx={{ pb: 0.5 }} variant="body2">
```
ianjon3s: This needs to be associated with the input as a proper `<label>`

adding `component="label"` and `id="User-role-label"` should do this correctly

(the input already has the property `aria-labelledby="User-role-label` which will link these up)

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/AddUserModal.tsx
```diff
@@ -61,19 +70,35 @@ export const AddUserModal: React.FC<AddUserModalProps> = ({
...
+                    Role
+                  </Typography>
+                  <SelectInput
+                    value={formik.values.role}
+                    name="User role"
+                    bordered
+                    required
+                    title={"User role"}
```
ianjon3s: the title can be removed once the label association is created

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
```diff
@@ -167,19 +170,38 @@ export const EditUserModal: React.FC<EditUserModalProps> = ({
...
-                      "role",
-                      formik.values.role === "teamEditor"
-                        ? "teamAdmin"
-                        : "teamEditor",
-                    )
-                  }
-                  label={"Team Admin"}
-                />
+                <>
+                  <Typography sx={{ pb: 0.5 }} variant="body2">
```
ianjon3s: same comments as above 👍

ianjon3s (approved):
Works great, just a couple of comments about input + label association

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
```diff
@@ -167,19 +170,38 @@ export const EditUserModal: React.FC<EditUserModalProps> = ({
...
+                    data-testid="user-role-select"
+                    labelId="user-role-select"
+                    onChange={() =>
+                      formik.setFieldValue(
+                        "role",
+                        formik.values.role === roleOptions[0].id
+                          ? roleOptions[1].id
+                          : roleOptions[0].id,
+                      )
+                    }
```
DafyddLlyr: We've updated the UI to be a `<select/>` but we're still maintaining a "toggle" functionality here. It would be more idiomatic (and extendable) to just read and set from `e.target.value`.

DafyddLlyr (changes requested):

apps/editor.planx.uk/src/pages/FlowEditor/components/Team/components/EditUserModal.tsx
zz-hh-aa: Ah yes thanks for catching, done in [80a0a22](https://github.com/theopensystemslab/planx-new/pull/6899/commits/80a0a226d13e232d6bf107e1cad4a7780b25513d)

DafyddLlyr (approved):
Great!

PR #6900, opened 2026-06-24
feat: Update Permission component for can/cannot edit and new TeamAdmin role

ianjon3s:
## What does this PR do?

Updates the `<Permission>` component:
- Adds `CanEdit` and `CannotEdit` to be used in view logic, replacing calls to the store for `canUserEditTeam`
- Applies the above to:
  - Flow table actions menu
  - Flow card actions menu
  - Edit history comment function
  - "Show individual node changes" accordion in public modal
 - Also adds `TeamAdmin` for future use (includes `PlatformAdmin` as per hierarchy)
 
 No expected change in functionality.


DafyddLlyr (approved):
Looks good - thanks! 👌

PR #6902, opened 2026-06-25
chore: Drop COMPONENT_SELECT feature flag

ianjon3s:
## What does this PR do?

- Removes `COMPONENT_SELECT` feature flag
- `<Hanger>` simplified to fire `addComponent` modal
- `ModalComponentHeader` no longer needs to pass the title or icon `ModalSectionContent`
- Titles and icons are pulled from `componentData` rather than being defined again in every editor modal


apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Hanger.tsx
```diff
@@ -96,38 +94,12 @@ const Hanger: React.FC<HangerProps> = ({ before, parent, hidden = false }) => {
...
-              bottom: rect.bottom,
-              left: rect.left,
-              right: rect.right,
-            });
-          }}
-        >
-          {canDrop && item && item.text}
-        </Link>
-      )}
+      <button
```
ianjon3s, line 97: Simplified to output a button that triggers the new `addComponent` functionality

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
```diff
@@ -47,76 +45,6 @@ const CloseButton = styled(IconButton)(({ theme }) => ({
   color: theme.palette.grey[600],
 }));
 
-const TypeSelect = styled("select")(() => ({
```
ianjon3s, line 50: Removing in favour of firing the `addComponent` modal

apps/editor.planx.uk/src/ui/editor/ModalComponentHeader.tsx
DafyddLlyr, line 1: This component doesn't really do anything - it's just a pure pass-through wrapper (renamed) for `ModalSectionContent`.

This has always been the case, and it's a bit odd (imho!) that the pattern is...

```tsx
<Section>
  <Header/>
  <Content/>
</Section>
<Section>
  <Content />
</Section>
<Footer/>
```

... and not something more expected like

```tsx
<Header/>
<Section>
  <Content/>
</Section>
<Section>
  <Content />
</Section>
<Footer/>
```

This might be a decent time to rectify this? It might be as simple as just dropping the redundant / misleading `ModalComponentHeader`.

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
DafyddLlyr: Docs will also need to update to account for this change, explainig that components are not categorised here but within `ALL_CATEGORIES` (`apps/editor.planx.uk/docs/adding-a-new-component.md`)

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
```diff
@@ -361,17 +289,11 @@ const FormModal: React.FC<FormModalProps> = ({
...
-          ) : (
-            !handleDelete && (
-              <NodeTypeSelect value={type} onChange={changeNodeType} />
-            )
-          )}
+          <ChangeComponentHeader
+            type={type}
+            onChange={changeNodeType}
+            canChange={!handleDelete}
+          />
```
DafyddLlyr, line 296: One downside of moving this here (which I'd missed previously sorry) is that it removes the title from our Storybook examples of all these components.

Image

`FormModal` does feel like the right structural location for this so I'm not too sure what the right solution is here - it would be nice to maintain fidelity in Storybook if we can. Possibly we eject a version of this header (minus the "change" link) into the Storybook `<Wrapper/>` (like we do the "Update" button?).

apps/editor.planx.uk/src/@planx/components/AddressInput/Editor.test.tsx
```diff
@@ -13,7 +13,7 @@ describe("AddressInputComponent - Editor Modal", () => {
         <AddressInputComponent id="test" />
       </DndProvider>,
     );
-    expect(screen.getByText("Address input")).toBeInTheDocument();
+    expect(screen.getByRole("form")).toBeInTheDocument();
```
DafyddLlyr: nit: This test is not as valuable in its new state I'd argue - we could have the wrong form or an empty form here.

It's tricky as we've now lost the identifying header we were using. I guess we should probably check for an input or other text value which truly identifies this component? This applies to a number of `Editor.test.tsx` files below also.

apps/editor.planx.uk/src/@planx/components/TaskList/Editor.tsx
```diff
@@ -94,7 +94,7 @@ const TaskListComponent: React.FC<Props> = (props) => (
           }
         />
         <ModalSection>
-          <ModalComponentHeader title="Task list" Icon={ICONS[TYPES.TaskList]}>
```
DafyddLlyr, line 97: nit: A few examples of `ICONS` remaining as an unused import across these files.

Image

DafyddLlyr (changes requested):
A few nits and suggestions below to take a look at - nothing too big.

It often turns out that removing a feature flag is not quite as simple as initially expected 😅 One solution to this is to mock the feature flags within the test environment, so we end up with two PRs - 

1. Update tests to match new pattern
2. Very simple and small "toggle" PR which can easily and quickly be reverted

Not sure this is entirely necessary to implement here - I'll let you decide - but a good pattern to have on your radar. We can toggle feature flags in tests by just setting up localStorage to include the relevant flag in our original setup calls for the test enviroments.

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
ianjon3s: I've tried to introduce an approach that keeps a single source of truth for component titles, introducing a `ComponentTitles` file that can be used across both the editor and Storybook

apps/editor.planx.uk/src/@planx/components/TaskList/Editor.tsx
ianjon3s: Thanks, all cleaned up

apps/editor.planx.uk/src/@planx/components/AddressInput/Editor.test.tsx
ianjon3s: Agreed, it did feel a bit flaky to introduce. In the absence of a title I've updated to target specific elements: https://github.com/theopensystemslab/planx-new/pull/6902/changes/6fbdea1c471df8b3bbc3e7e4070f904b1c35c49c

apps/editor.planx.uk/src/ui/editor/ModalComponentHeader.tsx
ianjon3s: I've removed `ModalComponentHeader` and introduced `ComponentTypeHeader` which is used both in the editor and Storybook to display the title and icon, pulling from single respective files

ianjon3s:
Thanks for the review @DafyddLlyr, agree I've introduced some wider complexity by removing the repeated title definition from editor modals, but hopefully this approach offers a clearer and easier to maintain structure

apps/editor.planx.uk/docs/adding-a-new-component.md
```diff
@@ -131,12 +131,21 @@ case TYPES.SetValue:
...
 
-```jsx
-<option value={TYPES.SetValue}>SetValue</option>
+Add an entry for the new component in the appropriate category within `ALL_CATEGORIES`. Each entry requires a `type`, `slug`, `title`, and `description`:
+
+```typescript
+{
+  type: TYPES.SetValue,
+  slug: "set-value",
+  title: t(TYPES.SetValue),
```
DafyddLlyr: nit: Typo?

apps/editor.planx.uk/docs/adding-a-new-component.md
```diff
@@ -131,12 +131,21 @@ case TYPES.SetValue:
...
+  description: "Set a data value in the passport",
+},
 ```
 
+> Note: components are no longer categorised in `FormModal.tsx`. The component picker UI is driven entirely by `ALL_CATEGORIES` in `componentData.ts`.
+
```
DafyddLlyr: nit: Not sure this is really necessary?

DafyddLlyr (approved):
Thanks for those changes...!

apps/editor.planx.uk/docs/adding-a-new-component.md
ianjon3s: I've updated the `t` shortcut to `title` so it's more obvious as a prefix

apps/editor.planx.uk/docs/adding-a-new-component.md
ianjon3s: Agreed, I've removed

PR #6904, opened 2026-06-26
feat(api): Generic Slack endpoint for Hasura events

DafyddLlyr:
## What does this PR do?
Adds a generic, reusable way to post Slack messages from Hasura events, and tidies up the existing Slack code along the way.

- New generic endpoint -  `POST /webhooks/hasura/send-slack-message` accepting `{ channel, message, username? }`. The message text and target channel are built in the Hasura event trigger's JSON template, so that adding a new notification is can be done purely via metadata (a new event-trigger YAML block, no new `controller/schema/route`.)
- Shared `sendSlackMessage` util. All existing callers (submission notifications, pay flow, sanitise + analytics crons) now route through it.
- Existing notifications (e.g. submission pings that need session/exemption lookups from DB) untouched and keep their custom logic.

## Motivation

* Tidying up - one Slack helper instead of the same client/`.send()` re-implemented in four places.
* Expandability - pinging a channel from a new event no longer means writing (and deploying) a bespoke endpoint each time
* Possible implementation details of https://github.com/theopensystemslab/planx-new/pull/6883

### DIFF
--- a/apps/api.planx.uk/modules/pay/service/utils.ts
+++ b/apps/api.planx.uk/modules/pay/service/utils.ts
@@ -2,8 +2,8 @@ import type { GovUKPayment } from "@opensystemslab/planx-core/types";
 import { $api } from "../../../client/index.js";
 import { gql } from "graphql-request";
 
-import SlackNotify from "slack-notify";
 import type { Request } from "express";
+import { sendSlackMessage } from "../../slack/utils.js";
 
 export const addGovPayPaymentIdToPaymentRequest = async (
   paymentRequestId: string,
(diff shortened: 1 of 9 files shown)


jessicamcinchak (approved):
Much appreciated! Scheduling 'daily failed submissions' on my to-do list today, started last Friday and quickly hit this exact issue of lots of slightly different/contradictory ways to manage Slack notifications ➕ 

Do you suspect we need to change/clean up any 'configurations'/'integrations' in Slack dashboard as result of this? Shout if extra eyes needed there

DafyddLlyr:
> _Do you suspect we need to change/clean up any 'configurations'/'integrations' in Slack dashboard as result of this? Shout if extra eyes needed there_

I'll let you know if I hit this, but I hope not - all being well the new endpoint should work as-is 👍

PR #6906, opened 2026-06-29
chore(editor): Simplify display of flows (sort / filter / search)

DafyddLlyr:
## What does this PR do?
Refactors search/filter/sort on the Team flows dashboard (active and archived flows). This falls into three parts - 

**1. Split the search UI into focused components**
The old `SearchBox` included both presentation with state management. During this refactor I found it quite challenging to work with - I kept expanding it to cover all options. Instead it's now split into three parts - 
   -  `<SearchInput/>` - simple presentational shell (the field, search icon, clear button / spinner). No internal state.
   -  `<SearchBox/>` - owns a local Fuse search. It's parent passes a list, gets filtered results back via `setRecords`. This is the basic behaviour currently in place.
   -  `<DebouncedSearchInput/>` - a controlled, debounced input for URL-synced lists. Internal state so the input updates immediately, and writes back to the parent on a debounce.

**2. Persist the flows search in the URL (`?search=`)**
Search is now bookmarkable/shareable and survives refresh and back/forward operations. This is consistent with how filters/sort already worked and was the original intention here.

**3. Unify search / filter / sort into one hook - `useDisplayedFlows()`**
Previously these were three separate mechanisms, which made managing state pretty awkward here (three pieces of local search state - `searchedFlows`, `searchedArchivedFlows`, `shouldClearSearch`. Now there's a single URL-param driven pipeline which runs `search → filter → sort`.

This PR ended up a little larger than expected sorry, but ~450 lines (over half) is newly added test coverage.

## Motivation
Search, sort, filter, and the archive view on the flows dashboard have grown quite organically, each bolted on its own way. This has left three separate sources of truth for "which flows do we show" wired together. We've noted this in a few PRs touching these pages recently, and we want to make it simple to manage this code (e.g. add / change filters).

This PR addresses this Trello ticket - [Refactor and simplify flows search](https://trello.com/c/kCVEWPWh/3656-refactor-and-simplify-flows-search-for-both-active-and-archived-flows) and is a prerequesite for what [we believe is a related bug](https://trello.com/c/Mv80HECs).


apps/editor.planx.uk/src/pages/Team/helpers/sortAndFilterOptions.ts
DafyddLlyr, line 1: **`isService` behaviour change** cc @zz-hh-aa 

I _think_ there's been a drift here in how these `validationFn`s were working, the `isService` property was added to the manual filter chain (see comment), but missed from the `validationFn` definiton. 

This hook makes the `validationFn`s the single source of truth for how a filter should work, so they now include the `isService` check - the Service status filter and the Submission flow-type filter only match actual services (not flows). I believe this is correct and the intended behaviour but just flagging as it's a pretty small change in a noisy PR sorry..!

apps/editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -57,79 +61,11 @@ const Team: React.FC<TeamProps> = (initialFlows) => {
...
-          flow.status === searchParams["service-status"] &&
-          flow.isService === true,
-      );
-    }
-
-    if (searchParams["flow-type"]) {
-      if (searchParams["flow-type"] === "submission") {
-        result = result.filter(
-          (flow) => flow.publishedFlows[0]?.hasSendComponent && flow.isService,
-        );
```
DafyddLlyr, line 90: This is where `isService` was previously managed (please see https://github.com/theopensystemslab/planx-new/pull/6906/changes#r3490459050)

apps/editor.planx.uk/src/pages/Team/helpers/sortAndFilterOptions.ts
```diff
@@ -23,17 +23,23 @@ export const sortOptions: SortableFields<FlowSummary>[] = [
...
-  if (value === "service charge enabled")
-    return flow.publishedFlows[0]?.hasEnabledServiceCharge;
+  const hasSendComponent =
+    Boolean(flow.publishedFlows[0]?.hasSendComponent) && flow.isService;
+  const isFeeCarrying = Boolean(flow.publishedFlows[0]?.hasVisiblePayComponent);
+  const hasServiceChargesEnabled = Boolean(
+    flow.publishedFlows[0]?.hasEnabledServiceCharge,
+  );
+
+  if (value === "submission") return hasSendComponent;
```
jessicamcinchak, line 39: Should "submission" logic change here? 

Your new L32 now checks for send component _plus_ `isService` but I think we still want send component _only_ ? Eg it's a flow I'm actively working on and haven't set online/`isService` yet but still has a send component to test submissions?

apps/editor.planx.uk/src/ui/shared/SearchBox/DebouncedSearchInput.tsx
```diff
@@ -0,0 +1,82 @@
...
+      value={localValue}
+      onChange={handleChange}
+      onClear={handleClear}
+      showClear={Boolean(localValue)}
+      isSearching={isSearching}
+      hideLabel={hideLabel}
+      compact={compact}
+    />
+  );
+};
```
jessicamcinchak, line 82: ➕

jessicamcinchak (approved):
One comment below ! Otherwise much appreciated and looks good!

apps/editor.planx.uk/src/pages/Team/helpers/sortAndFilterOptions.ts
DafyddLlyr: Yeah this was the "drift" I was describing - the `isService` check is in the old `/Team/index.tsx` here - https://github.com/theopensystemslab/planx-new/pull/6906/changes#r3490464451

And this file was unused - the left side of the diff here was previously never being hit. We missed this as part of our Tanstack Router migration I think (see https://github.com/orgs/theopensystemslab/projects/8/views/1?pane=issue&itemId=154371885)

This means there's no logical change here despite how this diff looks - the condition set in this PR is the same on we currently have on `main`.

The underlying question is a good one though - our previous assumption from the original PR https://github.com/theopensystemslab/planx-new/pull/6811 was that this filter should catch only _services_ with send components. I'll maybe defer to @zz-hh-aa here on this decision - was there a discussion about this previously?

For now I'll merge this and we can preserve the behaviour, but it's certainly worth checking this 👍

apps/editor.planx.uk/src/pages/Team/helpers/sortAndFilterOptions.ts
zz-hh-aa: Yep I totally missed this before, thanks for catching!

Confirming that the Service status filter and Submission flow-type filter should only match _services_

apps/editor.planx.uk/src/pages/Team/helpers/sortAndFilterOptions.ts
zz-hh-aa: I've done some digging for the original Slack thread but to no avail, but confirming that we decided that we should be consistent about this being a '_Service_ type' column (only services can ever accept submissions). Editors will need to mark something as a service to be able to see it categorised as "Submission" here.

PR #6909, opened 2026-06-29
feat(editor): Support validation of reference codes for LPGs

DafyddLlyr:
## What's the problem?
The `team.reference_code` value `CAB-SCA` is valid, but cannot be entered via the UI. To update this I've had to set the values directly via the DB.

image

Please see the folloiwng thread for further context - https://opensystemslab.slack.com/archives/C0ARCE260LF/p1782396116018359 (OSL Slack)

## What's the solution?
 - Update validation schema
 - Add basic test coverage


jessicamcinchak (approved):
Thanks!

PR #6911, opened 2026-06-30
feat: show message when no edit history for the last 6 months

zz-hh-aa:
Adds a new conditionally rendered message because it looks like a bug when there's no history to render
Screenshot 2026-06-30 103238

To test, create a new flow, open Hasura with the flow ID and update `published_flows` and `operations` table entries to last year or similar (outside of the 6 month query window), then refresh History panel.


DafyddLlyr (approved):
Perfect...!

PR #6912, opened 2026-06-30
fix(ci): Resolve failing cache-keys for build

DafyddLlyr:
## What's the problem?
Following the merge of https://github.com/theopensystemslab/planx-new/pull/6540 the Pulumi build is failing with the following error - 

```
docker-build:index:Image (07a03b10-container):
      error: failed to solve: failed to compute cache key: failed to calculate checksum of ref wn2yotpyeufgkdaklj8k8k865::yiyateh0eoji9imozhkftqo8i: "/pnpm-workspace.yaml": not found

    docker-build:index:Image (508f1352-container):
      error: failed to solve: failed to compute cache key: failed to calculate checksum of ref wn2yotpyeufgkdaklj8k8k865::2qlct8ekp0f029rb9l413y0uh: "/pnpm-workspace.yaml": not found
```

## What's the solution?
- Correctly set local build context so the single new `pnpm-lock.yaml` file can be located correctly

PR #6916, opened 2026-06-30
fix(ci): Fargate build of new API dockerfile

DafyddLlyr:
Follow up to #6912 - Pulumi is now timing out.

Locally this builds expected, but it might just be a case of correctly configuring these paths.

### DIFF
--- a/apps/api.planx.uk/Dockerfile
+++ b/apps/api.planx.uk/Dockerfile
@@ -12,7 +12,7 @@ RUN npm install -g pnpm@10.30.2
 COPY pnpm-workspace.yaml pnpm-lock.yaml package.json ./
 RUN pnpm fetch
 
-COPY apps/api.planx.uk ./apps/api.planx.uk
+COPY . .
 RUN pnpm install --frozen-lockfile --prefer-offline
 RUN pnpm --filter api.planx.uk build
 # `pnpm deploy` refuses to run unless `inject-workspace-packages=true`, failing with ERR_PNPM_DEPLOY_NONINJECTED_WORKSPACE
--- a/infrastructure/application/services/api.ts
+++ b/infrastructure/application/services/api.ts
@@ -92,9 +92,7 @@ export const createApiService = async ({
     repositoryUrl: repo.url,
     context: "../..",
     dockerfile: "../../apps/api.planx.uk/Dockerfile",
-    args: {
-      target: "production",
-    },
+    target: "production",
   });
   const apiLogGroup = new aws.cloudwatch.LogGroup("api", {
     name: "/ecs/api",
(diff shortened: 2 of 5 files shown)


infrastructure/application/services/api.ts
```diff
@@ -92,9 +92,7 @@ export const createApiService = async ({
     repositoryUrl: repo.url,
     context: "../..",
     dockerfile: "../../apps/api.planx.uk/Dockerfile",
-    args: {
-      target: "production",
-    },
+    target: "production",
```
DafyddLlyr, line 95: 🐛 Hit this along the way (x2 - see below)

We're currently running this service on AWS staging and prod with `--target development` as this is the last layer in the Dockerfile. We can verify this via Cloudwatch logs - 

```
2026-06-05T12:56:01.568+01:00 | > api.planx.uk@ dev /api
2026-06-05T12:56:01.568+01:00 | > tsx watch index.ts
```

The above is the result of `pnpm dev`, not `npm start`.

infrastructure/application/services/sharedb.ts
```diff
@@ -54,6 +54,7 @@ export const createSharedbService = async ({
     repositoryUrl: repo.repository.repositoryUrl,
     context: "../..",
     dockerfile: "../../apps/sharedb.planx.uk/Dockerfile",
+    target: "production",
```
DafyddLlyr, line 57: 🐛 Same as above, ShareDB has been running in `--target development`

```
2026-06-23T11:19:31.845+01:00 | > sharedb.planx.uk@ dev /sharedb
2026-06-23T11:19:31.845+01:00 | > node --watch server.js
```

This is `pnpm dev`, not `node server.js`

infrastructure/application/services/api.ts
freemvmt: wow nice catch!

apps/api.planx.uk/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN npm install -g pnpm@10.30.2
 COPY pnpm-workspace.yaml pnpm-lock.yaml package.json ./
 RUN pnpm fetch
 
-COPY apps/api.planx.uk ./apps/api.planx.uk
+COPY . .
```
freemvmt, line 15: how come these reversions? this wasn't the fix actually?

freemvmt (approved):

apps/api.planx.uk/Dockerfile
DafyddLlyr: Yeah - the previous fix did not work as expected - I think this Dockerfile combo + `{context: "../..", dockerfile: "../../apps/api.planx.uk/Dockerfile"}` is correct.

Frustratingly it's not something that can be tested locally (afaik?)

freemvmt: Hm, did you try `pulumi preview` or something? agree it is hard to get a read on veracity of a pulumi fix from local

PR #6918, opened 2026-07-01
feat: configure daily CRON event for logging failed submissions

jessicamcinchak:
Follows on from #6883 

See https://trello.com/c/JdLmMcxP/3645-implement-a-nightly-cron-job-to-check-for-failed-submission-events-without-a-successful-retry-resubmit

Configures a simple Hasura scheduled CRON event which:
- Queries the `daily_failed_submissions` database view
- If environment = production and rows returned, formats the failing session IDs/event types in a Slack notification to #planx-notifications-internal to investigate further

**Testing:**
- Cron events historically a bit fiddly to test! 
- If this pizza is still open at ~5p UK time today, then we should be able to log into pizza Hasura console -> Events -> Cron events -> "Invoked" tab and see successful event with message that indicates "skipping" because non-prod
- Else we're merge and monitor the same on staging console! 
- (Or of course clone branch locally and adjust cron timing in your local console to every minute or similar)


apps/api.planx.uk/modules/webhooks/service/dailyFailedSubmissions/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  if (data.failedSubmissions?.length === 0)
+    return "No recent submission failures";
+
+  const failures: string[] = [];
+  for (const failure of data.failedSubmissions) {
+    // Skip notifications related to Southwark Power Automate, which we expect to fail as of June 2026
+    if (
+      failure.teamName !== "Southwark" &&
+      failure.eventType !== "Upload to AWS S3"
+    ) {
```
jessicamcinchak, line 38: Are we happy to skip notifications for this combo to avoid noise, or useful to still log/see ? (If anything, will be easiest way to confirm this notification works every couple days 😅)

apps/api.planx.uk/modules/webhooks/controller.ts
```diff
@@ -296,3 +299,29 @@ export const updateTemplatedFlowEditsController: UpdateTemplatedFlowEditsControl
...
+  _req,
+  res,
+  next,
+) => {
+  try {
+    if (process.env.APP_ENVIRONMENT !== "production") {
+      return res.status(200).send({
+        message: `Not production, skipping failed submissions notification`,
+      });
+    }
```
jessicamcinchak, line 313: We'll never send Slack notifications for non-production submissions! This matches #planx-notifications etc.

apps/hasura.planx.uk/metadata/cron_triggers.yaml
```diff
@@ -7,6 +7,15 @@
     - name: authorization
       value_from_env: HASURA_PLANX_API_KEY
   comment: Parses allow-list answers for submitted sessions to retain for analytics post-sanitation
+- name: log_failed_submissions
+  webhook: '{{HASURA_PLANX_API_URL}}/webhooks/hasura/failed-submissions'
+  schedule: 0 17 * * *
```
jessicamcinchak, line 12: My thought is a ~5pm UK time log is nice for a couple reasons:
- We're all pretty much around, rather than only UK-folks waking up to notifications
- If any failures from earlier in the day, it gives us a chance to resolve them same date

Open to other preferences though!

DafyddLlyr: This works for me...! Lets run it like this for a good while and see if we want to adjust it 👍

apps/api.planx.uk/modules/webhooks/service/dailyFailedSubmissions/index.ts
DafyddLlyr: Skipping this seems right to me 👍

DafyddLlyr (approved):
Brill - thanks for this follow up - I'm sure it will be very helpful 👌

PR #6935, opened 2026-07-01
feat: auto-suggest data values for minor editor enums like EPC form

jessicamcinchak:
A quick solution off the back of some recent service testing bugs! See this thread https://opensystemslab.slack.com/archives/C5Q59R3HB/p1782910757324799

Relies on https://github.com/theopensystemslab/planx-core/pull/1019

We've been quite good at auto-suggesting "big" schema enums like `proposal.projectType`, `property.type` etc. But there are also many minor enums (eg less than 5 possible values) which we have _not_ handled, mostly because we publish these in a different format in the JSON Schema (_not_ `anyOf` value/description pairs, but rather actual `enum` properties). 

The planx-core helper method sets up basic handling for some (most?) of these and allows us to auto-suggest values for a range of new values here - notably `property.EPCKnown.form` which has been a recent culprit!

Further context about which data fields (`fn`s) included here in #planx-content thread, it can be a bit separate chat than code review! https://opensystemslab.slack.com/archives/C07F7215W7P/p1782942853790559

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/shared/utils.ts
+++ b/apps/editor.planx.uk/src/@planx/components/shared/utils.ts
@@ -96,6 +96,10 @@ export const getOptionsSchemaByFn = (
     case "application.leadDeveloper":
       schema = getValidSchemaValuesByEnumPath("LeadDeveloper", "type");
       break;
+    case "application.preAppAdvice.form":
+      // this one is not defined/accessible _within_ ODP Schema JSON, but mappings do rely on "Yes" so we still want to auto-suggest !
+      schema = ["Yes"];
+      break;
     case "application.type":
       schema = getValidSchemaValues("ApplicationType");
       break;
--- a/apps/editor.planx.uk/src/@planx/components/shared/utils.ts
+++ b/apps/editor.planx.uk/src/@planx/components/shared/utils.ts
@@ -1,4 +1,7 @@
-import { getValidSchemaValues } from "@opensystemslab/planx-core";
+import {
+  getValidSchemaValues,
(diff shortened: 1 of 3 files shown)


apps/editor.planx.uk/src/@planx/components/shared/utils.ts
```diff
@@ -79,11 +82,60 @@ export const getOptionsSchemaByFn = (
...
-    schema = getValidSchemaValues("ApplicationType");
-  if (fn === "proposal.projectType")
-    schema = getValidSchemaValues("ProjectType");
-  if (fn === "property.type") schema = getValidSchemaValues("PropertyType");
+  switch (fn) {
+    case "applicant.type":
+      schema = getValidSchemaValuesByEnumPath("BaseApplicant", "type");
+      break;
+    case "applicant.siteContact.role":
+      // .concat() on user role is lazy manual handling of discriminated union enum type for BaseApplicant/SiteContact/Role !
```
DafyddLlyr, line 90: 👍

DafyddLlyr (approved):
Great!

apps/editor.planx.uk/src/@planx/components/shared/utils.ts
```diff
@@ -96,6 +96,10 @@ export const getOptionsSchemaByFn = (
     case "application.leadDeveloper":
       schema = getValidSchemaValuesByEnumPath("LeadDeveloper", "type");
       break;
+    case "application.preAppAdvice.form":
```
DafyddLlyr, line 99: Helpful comment! 👍

DafyddLlyr (approved):

PR #6938, opened 2026-07-02
feat: disable update button when no change in component

zz-hh-aa:
In `FormModal`, this PR adds a new `isSubmitDisabled` var (instead of just `disabled`--we use this elsewhere but the logic for disabling submit is different from that of disabling the rest of the component!)

`isSubmitDisabled` now checks for:
- if the user has editing privileges
- if it's an existing node & formik is not dirty

Changes are predominantly in `FormModal` following Daf's [previous suggestions on a relevant PR](https://github.com/theopensystemslab/planx-new/pull/5851#pullrequestreview-3551984380). 

Most components were using `useFormikWithRef` but a handful were using `innerRef` (eg Enhanced Text Input, Pay), so they have been refactored here to use the hook too.

e2e tests also updated as needed. 

Before:
https://github.com/user-attachments/assets/02db3697-16e0-4f96-b0ad-3aa7a8c6da3a

After:
https://github.com/user-attachments/assets/9c56cd6d-a77e-4a61-bba6-93dd2a79134d

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/ExternalPortal/Editor.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/ExternalPortal/Editor.tsx
@@ -78,26 +78,23 @@ const ExternalPortalForm: React.FC<{
   areTemplatedNodeInstructionsRequired = false,
   formikRef,
 }) => {
-  const formik = useFormikWithRef<ExternalPortalFormData>(
-    {
-      initialValues: {
-        flow: flows.find((flow) => flow.id === flowId) || null,
-        flowId: flowId || null,
-        tags,
-        notes,
-        isTemplatedNode,
-        templatedNodeInstructions,
-        areTemplatedNodeInstructionsRequired,
-      },
-      onSubmit: (data) => {
-        if (handleSubmit) {
-          handleSubmit({ type: TYPES.ExternalPortal, data });
-        }
-      },
-      validationSchema,
+  const formik = useFormikWithRef<ExternalPortalFormData>({
+    initialValues: {
+      flow: flows.find((flow) => flow.id === flowId) || null,
+      flowId: flowId || null,
+      tags,
+      notes,
+      isTemplatedNode,
+      templatedNodeInstructions,
+      areTemplatedNodeInstructionsRequired,
     },
-    formikRef,
-  );
+    onSubmit: (data) => {
+      if (handleSubmit) {
+        handleSubmit({ type: TYPES.ExternalPortal, data });
+      }
+    },
+    validationSchema,
+  });
 
   return (
     <form id="modal" onSubmit={formik.handleSubmit} data-testid="form">
--- a/apps/editor.planx.uk/src/@planx/components/ExternalPortal/Editor.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/ExternalPortal/Editor.tsx
@@ -65,7 +65,10 @@ const ExternalPortalForm: React.FC<{
   isTemplatedNode?: boolean;
   templatedNodeInstructions?: string;
   areTemplatedNodeInstructionsRequired?: boolean;
-  formikRef?: MutableRefObject<FormikProps<ExternalPortalFormData> | null>;
+  formikRef?: {
+    current: FormikProps<ExternalPortalFormData> | null;
+    onDirtyChange: (dirty: boolean) => void;
+  };
 }> = ({
   handleSubmit,
   flowId,
@@ -78,23 +81,26 @@ const ExternalPortalForm: React.FC<{
   areTemplatedNodeInstructionsRequired = false,
   formikRef,
 }) => {
-  const formik = useFormikWithRef<ExternalPortalFormData>({
-    initialValues: {
-      flow: flows.find((flow) => flow.id === flowId) || null,
-      flowId: flowId || null,
-      tags,
-      notes,
-      isTemplatedNode,
-      templatedNodeInstructions,
-      areTemplatedNodeInstructionsRequired,
+  const formik = useFormikWithRef<ExternalPortalFormData>(
+    {
+      initialValues: {
+        flow: flows.find((flow) => flow.id === flowId) || null,
+        flowId: flowId || null,
+        tags,
+        notes,
+        isTemplatedNode,
+        templatedNodeInstructions,
+        areTemplatedNodeInstructionsRequired,
+      },
+      onSubmit: (data) => {
+        if (handleSubmit) {
+          handleSubmit({ type: TYPES.ExternalPortal, data });
+        }
+      },
+      validationSchema,
     },
-    onSubmit: (data) => {
-      if (handleSubmit) {
-        handleSubmit({ type: TYPES.ExternalPortal, data });
-      }
-    },
-    validationSchema,
-  });
+    formikRef,
+  );
 
   return (
     <form id="modal" onSubmit={formik.handleSubmit} data-testid="form">
--- a/apps/editor.planx.uk/src/@planx/components/Filter/Editor.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/Filter/Editor.tsx
@@ -5,26 +5,22 @@ import {
   DEFAULT_FLAG_CATEGORY,
   flatFlags,
 } from "@opensystemslab/planx-core/types";
+import type { Child, EditorProps } from "@planx/components/shared/types";
 import { useFormikWithRef } from "@planx/components/shared/useFormikWithRef";
-import type { FormikProps } from "formik";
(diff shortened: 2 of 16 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
```diff
@@ -146,40 +148,33 @@ const FormModal: React.FC<FormModalProps> = ({
...
-      return "";
-    }
 
-    if (Array.isArray(obj)) {
-      return obj.map(normalizeFormValues);
-    }
+  const node = id ? flow[id] : undefined;
+  const isEditingExistingNode = Boolean(handleDelete);
+
+  // checking formik periodically because useEffect would create infinite re-render
```
DafyddLlyr: Yeah this is getting pretty complex...! 😅 

The original sin here is that we have the forms living in child state, but the buttons etc in the `FormModal` parent - upwards data flow is not that viable a pattern in React and we're hitting the edges of it now.

Longer term, it's probably worth taking another look at the setup here but that would be a pretty fundamental change.

Right now I think there's two options to simplify this that are worth trying out (you might still end up circling back to your current solution but I think at least option 1 would be viable).

1. Move the `dirty` handling to the hook

If we had a signature something like this I think we'd be able to lose the interval - the hook re-runs on changes anyway

```ts
const [isFormDirty, setIsFormDirty] = useState(false);
const formikRef = useMemo<FormikRef<BaseNodeData & unknown>>(
  () => ({ current: null, onDirtyChange: setIsFormDirty }),
  [],
);
```

2. Work out why an `op` is actually firing (won't solve the UI state)
When a modal is opened, no changes are made, and "Update" is hit we should not get a resulting update of `flows.data` (which in turn is marking a `templated_flow_edits` record as complete). If you can find examples where hitting "Update" without changes marks a customisation as "Done" you should be able to take a look at the underlying operation (logged in console, also checkout WS connection in Dev Tools > Network). There might be an issue with form initialisation values which is causing these phantom edits. Solving this will still leave the button as "enabled" though

DafyddLlyr (commented):
This does work as expected but we might be missing the underlying issue here which is driving this problem.

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
zz-hh-aa: So following dev call yesterday I've made two commits:
- one that proceeds with disabling `Create` while maintaining an allow-list of components that do not need config [0e310ab](https://github.com/theopensystemslab/planx-new/pull/6938/commits/0e310ab7d81f4fcf2227eb4378ba46a5774dbefd)
- one that only disables `Update` and not `Create`

I guess option 2 is probably the simpler, expected option? 

In either case, I've also made a card to make validation handling consistent across components--I tested them all, some fail with a tooltip-esque message, some with error wrapper, some fail silently, some create without config, and a few create without config as we'd expect (eg Pay, Enhanced Text Description) https://trello.com/c/HDboptPY
Screenshot 2026-07-08 093955
Screenshot 2026-07-08 093944

apps/editor.planx.uk/src/@planx/components/ExternalPortal/Editor.tsx
```diff
@@ -78,26 +78,23 @@ const ExternalPortalForm: React.FC<{
...
+    initialValues: {
+      flow: flows.find((flow) => flow.id === flowId) || null,
+      flowId: flowId || null,
+      tags,
+      notes,
+      isTemplatedNode,
+      templatedNodeInstructions,
+      areTemplatedNodeInstructionsRequired,
     },
-    formikRef,
```
DafyddLlyr, line 99: Why is `formikRef` now dropped? Presumably this is because of the type - we don't want to disable `Update` here right? 

If this is the case, we should make this dependency much clearer - if `formikRef` is optional for some component types, for a reason (and with behavioural implications) we should document this (ideally in code!).

apps/editor.planx.uk/src/@planx/components/Filter/Editor.tsx
```diff
@@ -5,26 +5,22 @@ import {
...
 
-export interface Props {
-  id?: string;
-  handleSubmit?: (data: any, children?: any) => void;
-  node?: any;
-  autoAnswer?: Node["id"];
-  disabled?: boolean;
-  formikRef?: MutableRefObject<FormikProps<FilterData> | null>;
-}
+export type Props = EditorProps<
```
DafyddLlyr, line 19: Appreciate these tidy ups / standardisations 👌

apps/editor.planx.uk/src/pages/Preview/Node.tsx
```diff
@@ -220,10 +220,15 @@ const Node: React.FC<Props> = (props) => {
       );
 
     case TYPES.Filter: {
-      const filterProps = getComponentProps<Filter>();
-      const autoAnswer = nodeId ? autoAnswerableFlag(nodeId) : undefined;
+      const filterProps = getComponentProps<FilterData>();
+      const autoAnswers = nodeId ? autoAnswerableFlag(nodeId) : undefined;
```
DafyddLlyr, line 224: I'm pretty sure this is the only evocation of `autoAnswerableFlag()` - why not just change the signature to `(filterId?: NodeId) => NodeId[] | undefined` and simplify things here?

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
```diff
@@ -216,6 +191,8 @@ const FormModal: React.FC<FormModalProps> = ({
   //  2. The user has edit access to this team, but it is:
   //    - a templated flow
   //    - and the node itself is not marked "isTemplatedNode" or a child of an internal portal marked "isTemplatedNode"
+  //  3. The form is not dirty
+  //    - except for when a user is creating an allow-listed component (this is because they can be created and used as-is)
```
DafyddLlyr: See previous comment - this allow-list is now an invisible side-effect, and only discernible from reading PR comments here.

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
DafyddLlyr, line 1: Changes in this file really validate the new approach 🙌 

Much simpler and clearer, this will be easier to maintain and understand going forwards.

DafyddLlyr (changes requested):
A few minor comments questions, but looking really solid 👍 

I'd recommend updating the PR description here now to explain the _current_ changes - there's been a big change of direction which is not clearly captured. Updating this now will make life for us easier in the future when we `git blame` these changes.

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/FormModal.tsx
zz-hh-aa: Good catch, updated in [013f5e6](https://github.com/theopensystemslab/planx-new/pull/6938/commits/013f5e62a54495680e8795b03fb5fa93dd468e2e)

apps/editor.planx.uk/src/pages/Preview/Node.tsx
zz-hh-aa: I think I understood you correctly, let me know if I haven't! [bbb773f](https://github.com/theopensystemslab/planx-new/pull/6938/commits/bbb773f7f5a78fe5e8820c89087d10a6a9a265b9)

apps/editor.planx.uk/src/@planx/components/ExternalPortal/Editor.tsx
zz-hh-aa: Eagle-eyed catch, this was a mistake! I confused myself with a some stash- un-stash activity, fixed in [26ae92f](https://github.com/theopensystemslab/planx-new/pull/6938/commits/26ae92f430a64aa3c8d85efb81e58e598ad42259)

zz-hh-aa:
> I'd recommend updating the PR description here now to explain the _current_ changes - there's been a big change of direction which is not clearly captured. Updating this now will make life for us easier in the future when we `git blame` these changes.

Thanks Daf, good shout--worked through the changes and have updated the description accordingly too!

apps/editor.planx.uk/src/pages/Preview/Node.tsx
DafyddLlyr: Yep pretty much...! Only minor difference in mine was `filterId?: NodeId` (optional input) to save you from the ternary here 👌

DafyddLlyr (approved):
Great...!

PR #6941, opened 2026-07-02
Production deploy

jessicamcinchak:
Small ones but nice to get across before end of week?


DafyddLlyr (approved):
Currently working on follow ups to the `application_access_tokens` fixes discussed earlier here which would also be worth getting across to prod - https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1782990379303119 (OSL Slack)

I'd say lets merge this small deployment and, based on progress, maybe get another one across the line tomorrow?

PR #6942, opened 2026-07-02
fix(api): Ensure `InsertApplicationAccessToken` can handle failures and retries without issue

DafyddLlyr:
## What's the problem?
When a "send to email" event fails (e.g. due to Resend downtime) we have an automatic retry in place. This fails on second run due to the unique constraint on the `application_access_tokens` table - we cannot insert a second record.

## What's the solution?
Account for this by adding na `on_conflict` clause to the mutation. On a retry, we'll just bump the `expires_at` value allowing us to proceed as expected. We care about updating this value as a retry may not be an automatic Hasura retry, but a manual dev retry at a slightly later time.

## Testing
There's no meaningful way of testing this via mocks in the API tests - we can't mock the actual DB conflict. I'd suggest that if we hit issues here in the future we maybe aim to get some meaningful E2E tests in place.

### DIFF
--- a/apps/api.planx.uk/modules/send/email/service.ts
+++ b/apps/api.planx.uk/modules/send/email/service.ts
@@ -306,6 +306,10 @@ const insertApplicationAccessTokenRecord = async (
       ) {
         applicationAccessTokens: insert_application_access_tokens_one(
           object: { expires_at: $expiresAt, session_id: $sessionId }
+          on_conflict: {
+            constraint: application_access_tokens_session_id_key
+            update_columns: [expires_at]
+          }
         ) {
           token
         }


apps/api.planx.uk/modules/send/email/service.ts
```diff
@@ -306,6 +306,10 @@ const insertApplicationAccessTokenRecord = async (
       ) {
         applicationAccessTokens: insert_application_access_tokens_one(
           object: { expires_at: $expiresAt, session_id: $sessionId }
+          on_conflict: {
+            constraint: application_access_tokens_session_id_key
```
freemvmt, line 310: just ran this to make sure constraint was right, as per [docs](https://hasura.io/docs/2.0/mutations/postgres/upsert/#convert-insert-mutation-to-upsert) - looks good!

`SELECT * FROM "information_schema"."table_constraints" WHERE table_name='application_access_tokens'`

Image

freemvmt (approved):
Fab ✨

apps/api.planx.uk/modules/send/email/service.ts
DafyddLlyr: You can also get this from the GraphiQL playground in Hasura which has "intellisense" - it will show errors and suggest values - 

image

image

The "Explorer" panel also exposes these - 

image

freemvmt: ah these are nice tricks, thanks!

PR #6945, opened 2026-07-02
fix: Dont enforce fixed height on map containers

ianjon3s:
## What does this PR do?

- Fixes the issue that causes maps to have a margin-bottom on taller screens
- Simplifies code by making maps a standard height regardless of environment

**Before:**
image

**After:**
image


joecarver (approved):
Nice fix, thanks!

DafyddLlyr (approved):
Nice solution! Good to see it was fixable in PlanX also - my comment about it being a map repo issue was incorrect.

PR #6946, opened 2026-07-03
feat(api): Improve Resend error handling

DafyddLlyr:
## What does this PR do?
- Improved error handling and logging for Resend, allowing us to better understand failures
- Protects against retries by setting [an idempotency key](https://resend.com/docs/dashboard/emails/idempotency-keys)
- Follow up to #6942

PR #6949, opened 2026-07-03
chore: Remove deprecated "send to email" migration route

DafyddLlyr:
Whilst working on https://github.com/theopensystemslab/planx-new/pull/6946 I realised we hadn't revisited this cleanup from https://github.com/theopensystemslab/planx-new/pull/6319

This code was used to trigger a new-style Resend email, when an _existing_ GovNotify email link was clicked. As none of these exist any longer, and the 28 day waiting period has passed, we can retire this.

Currently, when a "send to email" event is triggered an email is generated via Resend which includes a link to `${process.env.EDITOR_URL_EXT}/download-submission?token=${token}`


jessicamcinchak (approved):
Great ✂️ !

PR #6950, opened 2026-07-03
feat: make dashboard the default team view

zz-hh-aa:
And remove feature flag. (It seemed to make more sense to do this in one go, then to do conditional routing based on a feature flag?)

### DIFF
--- a/apps/editor.planx.uk/src/hooks/data/useActivityData.ts
+++ b/apps/editor.planx.uk/src/hooks/data/useActivityData.ts
@@ -6,7 +6,7 @@ import { useMemo } from "react";
 const GET_ACTIVITY_BY_SERVICE = gql`
   query GetActivityByService($teamId: Int!, $thirtyDaysAgo: timestamptz!) {
     flows(
-      where: { team_id: { _eq: $teamId }, deleted_at: { _is_null: true } }
+      where: { team_id: { _eq: $teamId }, archived_at: { _is_null: true } }
     ) {
       name
       sessions: analytics_aggregate(
--- a/apps/editor.planx.uk/src/pages/Team/index.tsx
+++ b/apps/editor.planx.uk/src/pages/Team/index.tsx
@@ -58,7 +58,7 @@ const Team: React.FC<TeamProps> = (initialFlows) => {
   ]);
 
   const navigate = useNavigate();
-  const searchParams = useSearch({ from: "/_authenticated/app/$team/" });
+  const searchParams = useSearch({ from: "/_authenticated/app/$team/flows" });
 
   const { data } = useGetFlows(teamId);
   const flows = data?.flows ?? initialFlows.flows;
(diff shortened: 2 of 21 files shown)


apps/editor.planx.uk/src/hooks/data/useActivityData.ts
```diff
@@ -6,7 +6,7 @@ import { useMemo } from "react";
 const GET_ACTIVITY_BY_SERVICE = gql`
   query GetActivityByService($teamId: Int!, $thirtyDaysAgo: timestamptz!) {
     flows(
-      where: { team_id: { _eq: $teamId }, deleted_at: { _is_null: true } }
+      where: { team_id: { _eq: $teamId }, archived_at: { _is_null: true } }
```
zz-hh-aa, line 9: Revoked my platformAdmin privileges and hit an error here cause teamEditors can't select `deleted_at`, figured we don't want to show activity from archived views here anyway

joecarver:
Have just double checked this and the issue is indeed with the `useSearch()` - There's 4 separate files that need to be updated from:

`const searchParams = useSearch({ from: "/_authenticated/app/$team" });`

to 

`const searchParams = useSearch({ from: "/_authenticated/app/$team/flows" });`

the files are:
`useDisplayedFlows.tsx`
`useFlowSortDisplay.tsx`
`Filter.tsx`
`SortControl.tsx`

apps/editor.planx.uk/src/pages/Team/index.tsx
```diff
@@ -58,7 +58,7 @@ const Team: React.FC<TeamProps> = (initialFlows) => {
   ]);
 
   const navigate = useNavigate();
-  const searchParams = useSearch({ from: "/_authenticated/app/$team/" });
+  const searchParams = useSearch({ from: "/_authenticated/app/$team/flows" });
```
joecarver, line 61: ah yeah just seen you've done this change already here - basically just a few more files to do the same change on :)

joecarver (changes requested):
suggestion: Would it be worth updating the directory structure / file names in `editor/src/pages` ? Currently `Team/index.tsx` still renders the flows which is no longer the case semantically speaking! We'd probably want a separate `pages/Flows` directory and then can maybe keep the `pages/Team` directory for some of the higher level layout specific stuff...

zz-hh-aa:
Ah thanks v much @joecarver that did it!! Explains why my attempt to disable search to isolate the issue wasn't fruitful either...

I was thinking the same about renaming `Team`, I'll do that in a separate PR since it'll touch so many files.

joecarver (approved):
All looking good to me now, thanks for picking this up I had intended to do it after finishing dashboard but slipped my mind! 

I sent a message in Slack asking if we need to do any client comms before release so might need to hold off on merging but happy to approve

PR #6951, opened 2026-07-03
feat: notes [backend / db]

joecarver:
Contains the Hasura changes only for the new notes data structure, so:
- db migrations
- permissions
- metadata for gql 
- introspection tests

Another PR coming with the frontend changes which should make some of the choices here a bit more clear, but just to mention a couple of design decisions:
- Notes are always associated directly to one node via the node's ID
- The `placement` column determines how the note renders in the chart:
    - `attached_to_node` - stuck onto the bottom of a regular node
    - `attached_to_option` - stuck onto the bottom of a question/checklist option
    - `before_node` - not stuck to a specific node or option, located on the actual graph as a standalone note
- When a node is deleted:
    - any notes with placement `attached_to_` it are deleted alongside
    - any notes with placement `before_node` are moved to the next node in the flow
    - this is handled in the frontend


apps/hasura.planx.uk/migrations/default/1783036800000_create_table_public_flow_node_notes/up.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE "public"."flow_node_notes" (
+	"id" uuid NOT NULL DEFAULT gen_random_uuid(),
+	"flow_id" uuid NOT NULL,
+	"node_id" text NOT NULL,
+	"placement" text NOT NULL CHECK (placement IN ('attached_to_node', 'attached_to_option', 'after_node', 'before_node')),
```
jessicamcinchak: A couple loose thoughts on this: 
- I've been thinking of notes as having one of two possible positions:
  - Attached directly to an existing node ID (in which case column `node_id` is not null, `placement` is null)
  - Between two existing node IDs (in which case column `node_id` is null, `placement` is not null and must have `parent`, `before` keys - borrowing existing language from @planx/graph/index.ts operations)
- I'm not exactly sure how postgres types column types will play with TS discriminated union types, but does the above pseudo-code make sense? 
  - Is there a way to enforce those null/not-null constraints in PG? (( Maybe we want single `placement` JSONB column then enforce `{nodeId: string} | {parent: string, before: string}` in frontend methods which will update? ))
  - Rather than introduce new `attached_to_node`, `attached_to_option` (options _are_ node types/have their own unique IDs so not sure I understand this one?), `after_node` and `before_node` keys - can we re-use existing `parent` and `before` properties for describing location of notes? 
    - Even though we're saying notes themselves are _not_ nodes, I think their location shold be able to be described the same? 
    - And similarly "add", "delete" graph operations?! With the exception of allowing notes to be added with `parent` = question/checklist/filter node type ID and `before` = option node type ID?

apps/hasura.planx.uk/migrations/default/1783036800000_create_table_public_flow_node_notes/up.sql
```diff
@@ -0,0 +1,34 @@
...
+	"text" text NOT NULL DEFAULT '',
+	"color" text NOT NULL DEFAULT '#fffdb0',
+	"created_by" integer NOT NULL,
+	"updated_by" integer,
+	"created_at" timestamptz NOT NULL DEFAULT now(),
+	"updated_at" timestamptz NOT NULL DEFAULT now(),
```
jessicamcinchak: Am I understanding correctly that this will only ever capture/attribute the note _creation_ and _most recent edit_ only? 

Alternatively,
- Should `flow_notes` table only capture created by ID/timestamp?
- Should we create a separate linked table `flow_note_edits` to store updates as individual rows so we'll always have the option of displaying a complete edit history for a single note? (again, mental model closer to operations" this way?)

apps/hasura.planx.uk/migrations/default/1783036800000_create_table_public_flow_node_notes/up.sql
```diff
@@ -0,0 +1,34 @@
...
+	PRIMARY KEY ("id"),
+	FOREIGN KEY ("flow_id") REFERENCES "public"."flows" ("id") ON UPDATE RESTRICT ON DELETE CASCADE,
+	FOREIGN KEY ("created_by") REFERENCES "public"."users" ("id") ON UPDATE RESTRICT ON DELETE RESTRICT,
+	FOREIGN KEY ("updated_by") REFERENCES "public"."users" ("id") ON UPDATE RESTRICT ON DELETE RESTRICT
```
jessicamcinchak: nit: if we establish postgres `foreign keys` against `flows` and `users`, this means that notes will NOT persist on data syncs between environments (when flows and users are truncated, it will cascade here). 

This may well be the behaviour we want, but just flagging early so we're clear about consequences/intentions! Alternatively, Hasura has a gentler concept of "relationships" (which you can define in the data table GUI) which will enable relational GraphQL queries without the strict update/delete behaviour of true foreign keys. Easier to add a PG fkey later than drop one imo!

apps/hasura.planx.uk/migrations/default/1783036800000_create_table_public_flow_node_notes/up.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE "public"."flow_node_notes" (
```
jessicamcinchak: nit: is `flow_notes` a simpler, less prescriptive name (especially if notes are not _always_ attached to an existing flow node?) ?

Also maybe we'll support future note placements/positions that don't reference nodes at all - like a coordinate on the canvas, in a flow settings tab etc !

jessicamcinchak (commented):
Apologies it's taken me a minute to come around to this one. Have left a couple thoughts below - happy to chat through more tmrw at dev call maybe?

apps/hasura.planx.uk/migrations/default/1783036800000_create_table_public_flow_node_notes/up.sql
joecarver: yeah I was wondering whether we might want full history rather than latest version only, totally happy to do that if we think its important to track!

apps/hasura.planx.uk/migrations/default/1783036800000_create_table_public_flow_node_notes/up.sql
joecarver: Ah yes good shout! Definitely good to leave out the FKs for now.

apps/hasura.planx.uk/migrations/default/1783036800000_create_table_public_flow_node_notes/up.sql
joecarver: thanks for these thoughts, the modelling i chose for note position was somewhat arbitrary as an approach to solve the problem - so i'm happy to switch it up, especially if we have existing convention for `before` / `parent` - I did spot this but wasn't sure if it'd be appropriate.

> `Is there a way to enforce those null/not-null constraints in PG?`

assuming you mean the constraint that exactly one of `placement` / `node_id` must be set? You can do this in a `CHECK` constraint I believe, which accepts any custom SQL statement that evaluates to a boolean

 Can discuss more in dev call :)

apps/hasura.planx.uk/migrations/default/1783036800000_create_table_public_flow_notes/up.sql
```diff
@@ -0,0 +1,35 @@
+CREATE TABLE "public"."flow_notes" (
+	"id" uuid NOT NULL DEFAULT gen_random_uuid(),
+	"flow_id" uuid NOT NULL,
+	"node_id" text,
+	"placement" jsonb,
+	CHECK (
+		(node_id IS NOT NULL AND placement IS NULL) OR
+		(node_id IS NULL AND placement IS NOT NULL)
+	),
```
jessicamcinchak, line 9: `CHECK` is new-to-me and very cool / perfect for this case ! :star2:

apps/hasura.planx.uk/migrations/default/1783036800000_create_table_public_flow_notes/up.sql
```diff
@@ -0,0 +1,35 @@
...
+	"text" text NOT NULL DEFAULT '',
+	"color" text NOT NULL DEFAULT '#fffdb0',
+	"created_by" integer NOT NULL,
+	"updated_by" integer,
```
jessicamcinchak: nit: should `updated_by` be NOT NULL too?

jessicamcinchak (approved):
Thanks for working through feedback here!

PR #6952, opened 2026-07-03
chore: Bump `planx-core`

DafyddLlyr:
Please see https://github.com/theopensystemslab/planx-core/pull/1024

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -9,7 +9,7 @@ packages:
   - 'scripts/*'
 
 catalog:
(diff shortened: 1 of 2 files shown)


pnpm-workspace.yaml
```diff
@@ -9,7 +9,7 @@ packages:
   - 'scripts/*'
 
 catalog:
-  "@opensystemslab/planx-core": "github:theopensystemslab/planx-core#8471b48"
+  "@opensystemslab/planx-core": "github:theopensystemslab/planx-core#34004ae"
```
freemvmt, line 12: So great this is just in one place now 🚀

freemvmt (approved):
Yep!

PR #6954, opened 2026-07-03
[api] update API packages to get latest `fast-xml-parser`

freemvmt:
⚠️ All actual changes are in 1st commit, 2nd is just result of running `pnpm lint:fix` in the API package in order to fix the [otherwise-failing tests](https://github.com/theopensystemslab/planx-new/actions/runs/28816147262/job/85455949334?pr=6954).

The `deleteApplicationFiles` operation of the nightly data sanitation cronjob [has been failing](https://opensystemslab.slack.com/archives/C04EEBR1E0P/p1783044059748249) for 3 days with the following error:

`Failed to delete S3 files: Error: [EntityReplacer] Invalid character '#' in entity name: "#xD"`

My theory is this is to do with the XML parser in the AWS SDK. Something like: if there's a carriage return (`\r`) internal to one of the files up for deletion from `user-data`, then when the AWS SDK formats the request to the AWS API in XML (example [here](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html#API_DeleteObjects_RequestSyntax)), that carriage return gets encoded for XML as `&#xD`.

Then due to some [regression](https://github.com/NaturalIntelligence/fast-xml-parser/issues/823) introduced in `fast-xml-parser` 5.7.1, which we happened to downgrade to (from 5.7.3) in #6540, this entity cannot be handled. This tracks also because the issue only started after this was merged (we pushed it up to prod [same day](https://github.com/theopensystemslab/planx-new/pull/6917)).

So my proposed solution is just to update API packages to ensure we're on latest AWS SDK/XML parser.

I checked out that the encode/decode logic works along those lines in node:

```
> encodeURIComponent("some\rfile.png")
'some%0Dfile.png'
> decodeURIComponent("some%0Dfile.png")
'some\rfile.png'
```


package.json
```diff
@@ -30,7 +30,6 @@
       ]
     },
     "overrides": {
-      "fast-xml-parser": "^5.4.1",
```
freemvmt, line 44: redundant at this point

apps/api.planx.uk/package.json
```diff
@@ -12,51 +12,51 @@
   ],
   "dependencies": {
     "@airbrake/node": "^2.1.9",
-    "@aws-sdk/client-s3": "^3.1000.0",
-    "@aws-sdk/s3-request-presigner": "^3.1000.0",
+    "@aws-sdk/client-s3": "^3.1075.0",
```
freemvmt: this is the important bump

pnpm-workspace.yaml
```diff
@@ -1,64 +1,59 @@
 packages:
-  - 'apps/*'
-  - 'apps/hasura.planx.uk/tests'
-  - 'packages/*'
-  - 'e2e'
-  - 'e2e/tests/ui-driven'
-  - 'e2e/tests/api-driven'
-  - 'infrastructure/*'
-  - 'scripts/*'
+  - apps/*
```
freemvmt, line 2: most of this is just the result of me running `pnpm update` from root and presumably it has opinions about how to format the file

DafyddLlyr (approved):
Thanks for the fix and writeup here 👍 

The version regression here must have been introduced via one of the many rebases https://github.com/theopensystemslab/planx-new/pull/6540 went through. Aiming to work through outstanding dependabot issues today in case any others were missed.

PR #6961, opened 2026-07-06
chore: omit `flows.status` and `is_service` columns in sync

zz-hh-aa:
To aid testing--we don't want these values to overwritten on staging every night. If a service is frequently being tested on staging, we don't want editors to have to make `is_service -> true` and then `status -> online` each day.

Since `status` depends on `is_service` (eg no nestable flow should ever be online) I thought it made more sense to exclude both columns from being overwritten on conflict (eg to avoid a scenario where `status = online` persists but `is_service = true` gets overwritten with prod value `is_service = false`).

I tested this locally by updating certain `is_service` and `status` values, syncing data, and seeing that they persisted as expected. 

`sync-data` was failing locally so there's also a fix in here for that.

### DIFF
--- a/apps/hasura.planx.uk/migrations/default/1783336034412_fix_team_member_limit_count_distinct/up.sql
+++ b/apps/hasura.planx.uk/migrations/default/1783336034412_fix_team_member_limit_count_distinct/up.sql
@@ -1,0 +1,21 @@
+CREATE OR REPLACE FUNCTION check_team_member_limit()
+RETURNS trigger AS $$
+DECLARE
+  member_count INTEGER;
+BEGIN
+  SELECT COUNT(DISTINCT tm.user_id)
+    INTO member_count
+    FROM team_members tm
+    JOIN users u ON u.id = tm.user_id
+   WHERE tm.team_id = NEW.team_id
+     AND u.email IS NOT NULL;
+
+  IF member_count >= 20 THEN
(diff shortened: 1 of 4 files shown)


apps/hasura.planx.uk/migrations/default/1783336034412_fix_team_member_limit_count_distinct/up.sql
```diff
@@ -0,0 +1,21 @@
+CREATE OR REPLACE FUNCTION check_team_member_limit()
+RETURNS trigger AS $$
+DECLARE
+  member_count INTEGER;
+BEGIN
+  SELECT COUNT(DISTINCT tm.user_id)
```
zz-hh-aa, line 6: This is the only line that is different to the `down` migration. `sync-data` was failing locally for me with error:
```
ERROR:  Team cannot have more than 20 members
CONTEXT:  PL/pgSQL function check_team_member_limit() line 13 at RAISE
```

And I realised I never updated the max 20 members per-team trigger after we changed the permissions model to allow multiple roles per user per team.

jessicamcinchak: Good catch!

jessicamcinchak (approved):
Thanks for this one!

PR #6964, opened 2026-07-07
fix: add router context decorator to FlowsPanel story

ianjon3s:
## What does this PR do?

- Follows the pattern used in https://github.com/theopensystemslab/planx-new/pull/6955 to fix rendering of `FlowsPanel` in storybook
- Also updates mock dates to use `date-fns` as used across the codebase

**Testing:**
https://storybook.6964.planx.pizza/?path=/docs/editor-components-dashboard-flowspanel--docs

### DIFF
--- a/apps/editor.planx.uk/src/pages/Dashboard/components/FlowsPanel.stories.tsx
+++ b/apps/editor.planx.uk/src/pages/Dashboard/components/FlowsPanel.stories.tsx
@@ -1,4 +1,11 @@
 import type { Meta, StoryObj } from "@storybook/tanstack-react";
+import {
+  createMemoryHistory,
+  createRootRoute,
+  createRouter,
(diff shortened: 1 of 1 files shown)


DafyddLlyr:
Thanks for catching and fixing 🏅 

If we keep hitting this lets work out a more re-usable pattern - this is great for now though.

I've also added a Trello ticket which will encourage us to run Storybook as a CI/CD task - this should catch broken stories like this right away in future https://trello.com/c/KoeXjfg0/3672-investigate-storybook-addon-vitest-for-automated-accessibility-testing

DafyddLlyr (approved):

ianjon3s:
Thanks @DafyddLlyr , agree Storybook run on CI would make a lot of sense given how many components and elements are at play

PR #6965, opened 2026-07-07
feat: Add AGENTS.md file

ianjon3s:
## What does this PR do?

- Adds a guidance file to be used by AI agents in the codebase
- Also adds a `CLAUDE.md` file that reference the main agents file


DafyddLlyr (approved):
All looks good to me - lets run with this and see how we get on..! 🚀  🤖

ianjon3s:
Thanks @DafyddLlyr, I mostly left it to do it's own thing with some specific pointers about use of the store, and tooling around `date-fns` (as those were fresh in my head). Let's see how useful this is when conversing with an agent.

PR #6919, opened 2026-07-01
feat: Configure eslint, prettier, and husky once across monorepo

DafyddLlyr:
## What does this PR do?
This is part of [the monorepo migration (ADR 0010](https://github.com/theopensystemslab/planx-new/blob/main/doc/architecture/decisions/0010-migrate-to-a-monorepo-structure.md)), building on [the pnpm catalog work](https://github.com/theopensystemslab/planx-new/pull/6573).

### Shared config packages (`packages/`)
- `@planx/eslint-config` - ESLint 9 flat config, typescript-eslint v8. Linting rules are split by toolchain and composed per app
- `@planx/prettier-config` - shared Prettier options applied across all apps

### Husky fix
We had three separate `.husky/` dirs, each running its own `husky install`. This is now replaced with a single root `.husky/` (v9) and a singel `lint-staged` config. The pre-commit hook now fires for any workspace, from any directory, in one commit.

I've aimed to deliberately keep this small and reviewable - there's a few intentional `TODO`s in the code which I'll pick up as small and discrete follow up tasks. I did originally try bringing them in here but it was getting way too big.

This is still scoped to the three workspaces that already had linting in place (`editor`, `api`, `e2e`) - expanding to others (e.g. LPS) can be a follow up task.

## Testing

- `pnpm lint` passes in api, e2e, editor - 0 errors locally and in CI
- `tsc` passes in all three also
- Making a commit on/from this branch will auto-fix committed files in any directory in a single commit
  - Make a breaking prettier and/or eslint change in Editor and API files (e.g. spacing, newlines at end of file, incorrect import order)
  - `cd` to any directory
  - Make a commit
  - See all auto-fixes applied, not just those nested within the current project

## Next steps...
A few quick follow ups for this PR that will have a small config change, but will touch many files - please see `TODO`s in the diff here and associated comments. I'll handle these as small follow ups.

PR #6969, opened 2026-07-07
chore: Enable `@typescript/no-unused-expression` rule across codebase

DafyddLlyr:
Follow up to https://github.com/theopensystemslab/planx-new/pull/6919

### DIFF
--- a/apps/api.planx.uk/package.json
+++ b/apps/api.planx.uk/package.json
@@ -65,6 +65,7 @@
     "start": "node --enable-source-maps ./dist/index.js",
     "lint": "eslint '**/*.{js,ts}' && prettier -c .",
     "lint:fix": "eslint --fix '**/*.{js,ts}' && prettier -w .",
+    "lint:error": "eslint --quiet '**/*.{js,ts}' && prettier -c . --ignore-path ../../.prettierignore",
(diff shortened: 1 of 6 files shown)


apps/api.planx.uk/package.json
```diff
@@ -65,6 +65,7 @@
     "start": "node --enable-source-maps ./dist/index.js",
     "lint": "eslint '**/*.{js,ts}' && prettier -c .",
     "lint:fix": "eslint --fix '**/*.{js,ts}' && prettier -w .",
+    "lint:error": "eslint --quiet '**/*.{js,ts}' && prettier -c . --ignore-path ../../.prettierignore",
```
freemvmt, line 68: just quickly, what does this do?

freemvmt (approved):
yup :)

apps/api.planx.uk/package.json
DafyddLlyr: Ah sorry I should have commented this - we also have `pnpm lint:error` in the Editor.

If you run this it will just expose the errors and not the warnings - otherwise they're intermingled with warnings (of which there are many more) and can be hard to spot/fix.

freemvmt: ah v useful thanks!

PR #6977, opened 2026-07-08
fix: Expose dashboard stats to `TeamViewer` and `Analyst` roles

ianjon3s:
## What does this PR do

- Ensure all permission levels can see the `StatsBanner` dashboard stats summary for a given team
- Hide the `View analytics` link for permission group `CannotEdit`

Before (view of non `TeamEditor`):

image

After:
image

### DIFF
--- a/apps/editor.planx.uk/src/pages/Dashboard/components/StatsBanner.tsx
+++ b/apps/editor.planx.uk/src/pages/Dashboard/components/StatsBanner.tsx
@@ -5,6 +5,7 @@ import Tooltip from "@mui/material/Tooltip";
 import Typography from "@mui/material/Typography";
 import { useTeamAnalyticsLink } from "hooks/analyticsLinks/useTeamAnalyticsLink";
 import { cardBoxShadow, FONT_WEIGHT_SEMI_BOLD } from "theme";
+import Permission from "ui/editor/Permission";
 
 import { useStore } from "../../FlowEditor/lib/store";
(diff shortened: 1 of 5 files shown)


apps/editor.planx.uk/src/pages/Dashboard/components/StatsBanner.tsx
```diff
@@ -110,22 +111,24 @@ export function StatsBanner({
...
-            target="_blank"
-            rel="noopener noreferrer"
-            underline="always"
-          >
-            view analytics
-          </AnalyticsLink>
-        ) : (
-          <Tooltip title="Analytics unavailable" placement="bottom">
-            <AnalyticsLinkDisabled variant="body2">
+        <Permission.CanEdit>
```
jessicamcinchak: nit: if Metabase dashboards are all _public_, do we actually need to restrict this "view analytics" link? 

Makes sense to me that everyone can see everyone else's stats !

jessicamcinchak (approved):

apps/editor.planx.uk/src/pages/Dashboard/components/StatsBanner.tsx
ianjon3s: Agreed, I was following the logic in `EditorNavMenu` without questioning it. I've removed the `CanEdit` wrapper and also exposed analytics in the editor menu to all.

PR #6978, opened 2026-07-08
fix: restrict analyst access and redirect to admin panel

joecarver:
Analyst users should only have access to /app/admin-panel, so this PR adds an auto-redirect to take them there. It also adds guards round other routes to show the 404 page if they try access them. Also adds some tests for the util function and sidebar rendering.


jessicamcinchak:
Can this one be closed now in favor of #6977 & confirmations in this thread? https://opensystemslab.slack.com/archives/C5Q59R3HB/p1783585025488799

PR #6981, opened 2026-07-08
feat: Setup Turborepo

DafyddLlyr:
## What does this PR do?
Adopts [Turborepo](https://turborepo.com) as a task runner over the existing pnpm workspace. Turbo doesn't replace pnpm - it reads the existing workspace structure from `pnpm-workspace.yaml`, runs their existing `package.json` scripts, and caches each task's inputs and outputs.

This PR is just for local dev - CI/CD, Docker builds, and remote caching will be follow up steps.
- ⚠️ Apart from LPS which does need a change which touches CI/CD - see comment below

image

## Motivation
Wec now have one single command from the root to `typecheck` / `lint` / `build` the whole repo, with caching. This means re-runs are basically instant - not a huge win right now but will have a big impact once we have more `/packages` (e.g. `planx-core`).

## Testing
This is worth trying out locally - currently this is not running anywhere else. Commands below all run from root.

### Type-checking every workspace
```sh
# First run - will take maybe a minute
pnpm typecheck

# Second run - basically instant, not repeated (>>> FULL TURBO 🌈 )
pnpm typecheck

# Make a change to some code (e.g. modify a file in api.planx.uk)
# Only API type-check will be repeated, much quicker than first run (9/10 cached)
pnpm typecheck
```

This pattern of first run, second run, change files, additional run can be repeated for the commands below to test them. You can also delete the `/.turbo` directory to clear the cache or add the `--force` flag to skip a cache ([docs](https://turborepo.dev/docs/reference/run#--force)).

### Other tasks and commands
```sh
# Linting
pnpm lint

# Combined typecheck + lint
pnpm check

# Build Editor / API / LPS
pnpm build

# Scope to a single package
pnpm turbo run build --filter=api.planx.uk
```

## Next steps...
- CI integration
- Remote caching - caches can be shared across runs via Vercel (or self-hosted). This should have a significant impact on CI times (as well as local dev)
- Use Turbo for Docker builds

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,17 @@ The root of the project has several scripts set up to help you manage your docke
 - `pnpm analytics` will recreate your docker containers and include [Metabase](https://www.metabase.com/)
 - `pnpm logs` will print docker log entries (this can be filtered by appending `-- [service name]`, for example `pnpm logs -- api`)
 
+### Task running (Turborepo)
+
+[Turborepo](https://turborepo.com) runs and caches tasks across the pnpm workspace. From the project root:
+
+- `pnpm typecheck` - type-check every workspace (`tsc --noEmit` / `astro check`)
+- `pnpm lint` - ESLint + Prettier checks (Editor, API, E2E)
+- `pnpm check` - `typecheck` and `lint` together
+- `pnpm build` - build the Editor, API and LPS
+
+Turbo caches by content hash, so re-running an unchanged task is near-instant (`>>> FULL TURBO`). This can be scoped to one package with `--filter` (e.g. `pnpm turbo run build --filter=api.planx.uk`).
+
 ### Documentation
 
 This project uses Architecture Decision Records (ADRs) to record significant changes and decisions. Further details of this can be [found here](https://github.com/theopensystemslab/planx-new/blob/main/doc/architecture/decisions/0001-record-architecture-decisions.md).
--- a/apps/localplanning.services/astro.config.mjs
+++ b/apps/localplanning.services/astro.config.mjs
@@ -1,16 +1,19 @@
+/* eslint-disable no-restricted-syntax */
+
 import react from "@astrojs/react";
 import sitemap from "@astrojs/sitemap";
 import tailwindcss from "@tailwindcss/vite";
 import { defineConfig, envField, fontProviders } from "astro/config";
 import icon from "astro-icon";
 
-// Check args to access Astro mode
-// Env var is not available in Astro config files
-const mode = process.argv.at(-1).replaceAll("-", "");
+// BUILD_MODE must be an env var for Turbo caching
+// The dev / pizza / staging / prod builds must have their own cache-key
+const mode = process.env.BUILD_MODE ?? "dev";
 const isCloudfrontBuild = ["staging", "production"].includes(mode);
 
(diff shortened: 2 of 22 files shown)


ianjon3s (approved):
Worked through all commands, working as described 🚀

ianjon3s:
One thing I noticed is that I'm getting a bunch (75 in this case) of log files as diffs after running the commands, do we need to update the `gitignore` file?

image

DafyddLlyr:
> _One thing I noticed is that I'm getting a bunch (75 in this case) of log files as diffs after running the commands, do we need to update the `gitignore` file?_

Good question - this PR does already add `.turbo` folders to the `.gitignore` so these _should_ be skipped (they're not in this PR diff, and are show as gitignored locally for me). Possibly a local issue - are these files _just_ appearing as a diff when you exit this branch (e.g. `git checkout main` > files *now* show)? If this is the case it's fine - once this is merged and they're included on the `main` branch `.gitignore` it will be fine.

image

ianjon3s:
Indeed, the diffs appeared when I switched branches, all good then 🙏

apps/localplanning.services/astro.config.mjs
DafyddLlyr, line 1: Changes are required here as we need turbo to per _per-env_ - the dev build is not the same as the prod build for example. Turbo can do this via env vars ([docs](https://turborepo.dev/docs/crafting-your-repository/using-environment-variables)).

This is a bit awkward in Astro as it can't read the correctly evaluated env files within this config file (e.g. `.env.production`) - please see https://github.com/theopensystemslab/planx-new/pull/5262 for more context here.

The solution here is to still populate a readable `process.env` (via GHA) so that builds are cachable, per-env.

Locally we can test this via the following methods (and repeat to hit caches) - 

```
# Dev mode - read from `.env.dev` (default)
pnpm turbo run build --filter=localplanning.services

# Pizza mode - read from `.env.pizza`
BUILD_MODE=pizza SITE_URL=https://1234.editor.planx.pizza pnpm turbo run build --filter=localplanning.services

# Staging mode - read from `.env.staging`
BUILD_MODE=staging SITE_URL=https://localplanning.editor.planx.dev pnpm turbo run build --filter=localplanning.services

# Production mode - read from `.env.production`
BUILD_MODE=production SITE_URL=https://localplanning.services pnpm turbo run build --filter=localplanning.services
```

README.md
```diff
@@ -99,6 +99,17 @@ The root of the project has several scripts set up to help you manage your docke
...
+
+[Turborepo](https://turborepo.com) runs and caches tasks across the pnpm workspace. From the project root:
+
+- `pnpm typecheck` - type-check every workspace (`tsc --noEmit` / `astro check`)
+- `pnpm lint` - ESLint + Prettier checks (Editor, API, E2E)
+- `pnpm check` - `typecheck` and `lint` together
+- `pnpm build` - build the Editor, API and LPS
+
+Turbo caches by content hash, so re-running an unchanged task is near-instant (`>>> FULL TURBO`). This can be scoped to one package with `--filter` (e.g. `pnpm turbo run build --filter=api.planx.uk`).
+
```
jessicamcinchak, line 112: 🌟

jessicamcinchak (approved):
Cool !!

PR #6985, opened 2026-07-09
refactor(lps): Open accordions using URL query param

ianjon3s:
## What does this PR do?

Issue identified by DAC, on providing LPS links for testing.

Currently on an LPA services page on LPS, we default open accordions depending on the route that the user took to reach the page, for example:
- Entering through https://localplanning.services/search — no accordion open by default
- Entering through https://localplanning.services/search/notify — "Start a planning application" accordion open
- Entering through https://localplanning.services/search/guidance — "Get planning guidance" accordion open

However the above only works if you navigate through the homepage, visiting a search URL directly (for example https://localplanning.services/search/guidance) doesn't result in the expansion of any accordions.

This PR refactors to introduce a URL query param, rather than the in-memory nanostore set by a homepage card click.

**Testing (visit URL directly):**
https://localplanning.6985.planx.pizza/search/guidance/


apps/localplanning.services/src/content/action.ts
DafyddLlyr, line 1: nit: This isn't a `/stores` file any longer - it's just some shared types / constants.

Maybe move it to `/content/actions`?

DafyddLlyr (approved):
Great - one small tidy up `nit:`

apps/localplanning.services/src/content/action.ts
ianjon3s: Nice catch, have updated

PR #6986, opened 2026-07-09
feat(ci/cd): Integrate Turbo into CI processes

DafyddLlyr:
## What does this PR do?
This PR wires up  Turborepo (with a Vercel remote cache) through the CI pipeline so that unchanged work is restored instead of recomputed at each step - shared across jobs, runs, branches, and local dev. 

## Caching
```mermaid
flowchart LR
  T[Turbo task] --> L{Is there a local<br/>.turbo cache?}
  L -- hit --> R[Skip task]
  L -- miss --> V{Is there a <br/>remote cache?}
  V -- hit --> R
  V -- miss --> X[Run task] --> S[Save to local<br/>+ remote]
```

We have two caches set up here - 
- **Remote** - Held by Vercel, this is the cross-job / cross-run cache
- **Local** - Held per-job, synched to remote. Held in `.turbo` folder

## Changes
- Added a `verify` job which runs once across all workspaces - previous the Editor and Pulumi did not have this as part of CI
- Turbo now handles test runs
  - Editor and API can be cached - there's no external inputs, these are always deterministic
  - Hasura and E2E tests are marked as `cache: false` - because they rely on containerised services we can't reliable hash results
 - Turbo now handles build processes

## How to test...
- CI is passing on this PR, Pizza is building
- GHA action logs show that the remote cache is hit (`Remote caching enabled` → `>>> FULL TURBO`)
- Local dev still works as before

### DIFF
--- a/.github/workflows/build-planx-frontend.yml
+++ b/.github/workflows/build-planx-frontend.yml
@@ -36,6 +36,10 @@ on:
 permissions:
   contents: read
 
+env:
+  TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+  TURBO_TEAM: ${{ secrets.TURBO_TEAM }}
+
 jobs:
   build_planx_frontend:
     name: Build PlanX frontend
@@ -44,8 +48,13 @@ jobs:
       - uses: actions/checkout@v7
       - uses: ./.github/actions/setup-node-pnpm
       - run: pnpm install --frozen-lockfile
-      - run: pnpm build
-        working-directory: ${{ inputs.editor-directory }}
+      - name: Restore Turbo cache
+        uses: actions/cache@v6
+        with:
+          path: .turbo
+          key: ${{ runner.os }}-turbo-${{ github.sha }}
+          restore-keys: ${{ runner.os }}-turbo-
+      - run: pnpm turbo run build --filter=editor.planx.uk
         env:
           VITE_APP_AIRBRAKE_PROJECT_ID: ${{ secrets.AIRBRAKE_PROJECT_ID }}
           VITE_APP_AIRBRAKE_PROJECT_KEY: ${{ secrets.AIRBRAKE_PROJECT_KEY }}
--- a/.github/workflows/build-planx-frontend.yml
+++ b/.github/workflows/build-planx-frontend.yml
@@ -52,8 +52,8 @@ jobs:
         uses: actions/cache@v6
         with:
           path: .turbo
-          key: ${{ runner.os }}-turbo-${{ github.sha }}
-          restore-keys: ${{ runner.os }}-turbo-
+          key: ${{ runner.os }}-turbo-${{ github.job }}-${{ github.sha }}
+          restore-keys: ${{ runner.os }}-turbo-${{ github.job }}-
       - run: pnpm turbo run build --filter=editor.planx.uk
         env:
           VITE_APP_AIRBRAKE_PROJECT_ID: ${{ secrets.AIRBRAKE_PROJECT_ID }}
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -141,8 +141,13 @@ jobs:
       - name: Setup Node and PNPM
         uses: ./.github/actions/setup-node-pnpm
       - run: pnpm install --frozen-lockfile
-      - run: pnpm build
-        working-directory: ${{ env.EDITOR_DIRECTORY }}
+      - name: Restore Turbo cache
+        uses: actions/cache@v6
+        with:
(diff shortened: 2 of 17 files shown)


.github/workflows/pull-request.yml
```diff
@@ -67,6 +69,26 @@ jobs:
...
+      - name: Setup Node and PNPM
+        uses: ./.github/actions/setup-node-pnpm
+      - run: pnpm install --frozen-lockfile
+      - name: Restore Turbo cache
+        uses: actions/cache@v6
+        with:
+          path: .turbo
+          key: ${{ runner.os }}-turbo-${{ github.job }}-${{ github.sha }}
+          restore-keys: ${{ runner.os }}-turbo-${{ github.job }}-
+      - run: pnpm turbo run typecheck lint --affected
```
DafyddLlyr, line 88: The `--affected` flag is graph-aware which means it only runs for changed packages and dependents.

.github/workflows/build-planx-frontend.yml
```diff
@@ -36,6 +36,10 @@ on:
 permissions:
   contents: read
 
+env:
+  TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+  TURBO_TEAM: ${{ secrets.TURBO_TEAM }}
```
DafyddLlyr, line 41: Added to shared `.env` on AWS S3 + repo secrets

I'm not sure if these will work locally without a `npx turbo login` - let's pair and test this at some point.

.github/workflows/pull-request.yml
```diff
@@ -93,8 +115,6 @@ jobs:
     name: Run API Tests
     runs-on: ubuntu-24.04
     timeout-minutes: 10
-    needs: [changes]
-    if: ${{ needs.changes.outputs.api == 'true' || needs.changes.outputs.pnpmworkspace == 'true' }}
```
DafyddLlyr, line 97: Removed - API tests could also fail if an internal monorepo package has been updated. The `--affected` flag now captures this.

The same pattern is repeated below.

apps/editor.planx.uk/package.json
```diff
@@ -156,7 +156,8 @@
     "start": "vite --open",
     "storybook": "storybook dev -p 6006",
     "test:silent": "vitest --silent",
-    "test": "vitest",
+    "test": "vitest run",
+    "test:watch": "vitest",
```
DafyddLlyr, line 160: Updating the names here to be accurate means a shared `turbo test` has the same behaviour across packages.

apps/localplanning.services/turbo.json
```diff
@@ -5,7 +5,8 @@
     "build": {
       "dependsOn": ["^build"],
       "outputs": ["dist/**"],
-      "env": ["BUILD_MODE", "SITE_URL", "ROOT_DOMAIN"]
+      "env": ["BUILD_MODE", "SITE_URL", "ROOT_DOMAIN"],
+      "cache": false
```
DafyddLlyr, line 9: Turns out we can't (currently) cache this just on the above `ENV` values - if there's a new LPA, or new service, we need a rebuild.

For now, I've just disabled this caching - it's a quick build.

Longer term, the answer is to run the LPS / services query upfront and use this to decide if we need to rebuild or not (we've previously discussed this approach).

jessicamcinchak (approved):
Cool !

.github/workflows/build-planx-frontend.yml
freemvmt: I wasn't convinced my turbo commands were hitting the remote cache, so I think it _is_ necessary to login from local for the true 'multiplayer' experience? i.e. I don't think just having these in `.env` is enough.

`npx turbo login` alone prompts you to login via browser, but afaik we don't all have Vercel logins and it's not in 1pass? So I ended up using `npx turbo login --manual`. I had to supply the Remote Cache URL (it doesn't assume the default of `https://api.vercel.com`), then the `TURBO_TEAM` and `TURBO_TOKEN` values as given here.

Now it seems to work!

DafyddLlyr: I missed this comment earlier, could you add this step to the `README.md` so we can all get access to remote caching? I have Vercel access so `npx turbo login` does work directly.

DafyddLlyr: I missed this comment earlier, could you add this step to the `README.md` so we can all get access to remote caching? I have Vercel access so `npx turbo login` does work directly.

freemvmt: I try to address this here: https://github.com/theopensystemslab/planx-new/pull/7184

PR #6988, opened 2026-07-09
fix: ensure Submissions log always displays an address

jessicamcinchak:
Submissions logs currently have a number of blank "address" column values. If you check payloads, these are always for cases where the site address is _proposed_, rather than an existing OS site address. 

According to our address types, _all_ addresses have a `"title"`, but only existing OS addresses have a `"single_line_address"`. This updates the query to use `coalesce()` and always fallback to title ! 

See https://trello.com/c/LxU8rBR9/3664-handle-propose-a-new-address-journeys-in-submissions-table-and-on-overviewhtml

### DIFF
--- a/apps/hasura.planx.uk/migrations/default/1783614492150_update_submission_services_log_address/up.sql
+++ b/apps/hasura.planx.uk/migrations/default/1783614492150_update_submission_services_log_address/up.sql
@@ -1,0 +1,83 @@
+CREATE OR REPLACE VIEW "public"."submission_services_log" AS 
+ WITH payments AS (
+         SELECT ps.session_id,
+            ps.payment_id AS event_id,
+            'Pay'::text AS event_type,
+            initcap(ps.status) AS status,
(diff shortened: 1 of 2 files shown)


apps/hasura.planx.uk/migrations/default/1783614492150_update_submission_services_log_address/up.sql
```diff
@@ -0,0 +1,83 @@
...
+    ae.status,
+    ae.response,
+    ae.created_at,
+    ae.retry,
+    f.team_id,
+    f.name AS flow_name,
+    coalesce(
+      ((((ls.data -> 'passport'::text) -> 'data'::text) -> '_address'::text) ->> 'single_line_address'::text), 
+      ((((ls.data -> 'passport'::text) -> 'data'::text) -> '_address'::text) ->> 'title'::text)
+   )  AS address
```
jessicamcinchak, line 78: 75-78 are the only updated lines here ! 

Because we're updating an existing column, we're able to "replace" the existing view in place and don't need to worry about also dropping/recreating/replacing any of its' dependents like `daily_failed_submissions`.

DafyddLlyr (approved):
Perfect!

PR #6989, opened 2026-07-09
fix: account for undefined `data` in daily failed submission query

jessicamcinchak:
See https://planx.airbrake.io/projects/329753/groups/4373085423522651345

### DIFF
--- a/apps/api.planx.uk/modules/webhooks/service/dailyFailedSubmissions/index.ts
+++ b/apps/api.planx.uk/modules/webhooks/service/dailyFailedSubmissions/index.ts
@@ -4,40 +4,42 @@ import { $api } from "../../../../client/index.js";
 import { sendSlackMessage } from "../../../slack/utils.js";
 
 interface FailedSubmissions {
-  data: {
-    failedSubmissions:
-      | {
-          sessionId: string;
-          eventType: string;
-          teamName: string;
-        }[]
-      | [];
-  };
+  failedSubmissions:
+    | {
+        sessionId: string;
+        eventType: string;
+        teamName: string;
+      }[]
+    | [];
 }
 
 export const getDailyFailedSubmissions = async (): Promise<string> => {
-  const { data } = await $api.client.request<FailedSubmissions>(gql`
-    query GetDailyFailedSubmissions {
-      failedSubmissions: daily_failed_submissions {
-        sessionId: session_id
-        eventType: event_type
-        teamName: name
+  const { failedSubmissions } = await $api.client.request<FailedSubmissions>(
+    gql`
+      query GetDailyFailedSubmissions {
+        failedSubmissions: daily_failed_submissions {
+          sessionId: session_id
+          eventType: event_type
+          teamName: name
+        }
       }
-    }
-  `);
+    `,
+  );
 
-  if (!data || data.failedSubmissions?.length === 0)
-    return "No recent submission failures";
+  if (failedSubmissions?.length === 0) return "No recent submission failures";
 
   const failures: string[] = [];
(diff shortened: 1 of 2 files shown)


apps/api.planx.uk/modules/webhooks/service/dailyFailedSubmissions/index.ts
```diff
@@ -26,7 +26,7 @@ export const getDailyFailedSubmissions = async (): Promise<string> => {
     }
   `);
 
-  if (data.failedSubmissions?.length === 0)
+  if (!data || data.failedSubmissions?.length === 0)
```
jessicamcinchak: I had assumed (based on Hasura API console) that when the view is empty (hopefully its usual state), my response would be: 
```
data: {
  failed_submissions: []
}
```

But based on latest local testing here, `data` is actually `undefined` ! I'm not getting/seeing any "...not in query root" errors so I don't think undefined is a permissions issue/bug?

apps/hasura.planx.uk/metadata/cron_triggers.yaml
```diff
@@ -9,7 +9,7 @@
   comment: Parses allow-list answers for submitted sessions to retain for analytics post-sanitation
 - name: log_failed_submissions
   webhook: '{{HASURA_PLANX_API_URL}}/webhooks/hasura/failed-submissions'
-  schedule: 0 17 * * *
+  schedule: 0 9,17 * * *
```
jessicamcinchak, line 12: Logging twice a day until confident this is sorted !

apps/api.planx.uk/modules/webhooks/service/dailyFailedSubmissions/index.ts
DafyddLlyr: Yeah this is a weird one...!

Running this query in the console gives me the response above, not `undefined`. Permissions all look correct also so agree this isn't the issue.

I actually think the issue is the interface `FailedSubmissions` which has the additional `data` wrapper the API client is stripping. As it's included in the interface it doesn't raise a type error here, but if you have a look at other examples the `data` wrapper is not included - we just returns the records directly.

DafyddLlyr (changes requested):
Think I might have spotted the issue - could you try without `data`?

apps/api.planx.uk/modules/webhooks/service/dailyFailedSubmissions/index.ts
DafyddLlyr, line 37: Just spotted this also - I think this will reject any S3 uploads who are not Southwark. The below would satisfy this condition and not log to Slack -

```json
{ 
  teamName: "Camden",
  eventType: "Upload to AWS S3"
}
```

I think this would be the intention here - 
```ts
if (
  failure.teamName === "Southwark" &&
  failure.eventType === "Upload to AWS S3"
) {
  continue;
}
failures.push(`- *${failure.sessionId}* ${failure.eventType}`);
```

apps/api.planx.uk/modules/webhooks/service/dailyFailedSubmissions/index.ts
jessicamcinchak: Ah what a tedious one - yes you're totally right! I'd forgotten that API client strips away `data` here !

apps/api.planx.uk/modules/webhooks/service/dailyFailedSubmissions/index.ts
jessicamcinchak: Good spot, now updated !

DafyddLlyr (approved):
Fab!

PR #6992, opened 2026-07-10
feat: Exclude system teams from dashboard

ianjon3s:
## What does this PR do?

As [discussed in Slack](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1783509114465439?thread_ts=1783507676.099849&cid=C01E3AC0C03), for some admin/system teams we don't need a dashboard, and it's better to directly access the flows page.

- Set up allow-list `systemTeams` which exports the helper `isSystemTeam`
- List includes
  - Open Systems Lab
  - Testing
  - PlanX Academy
  - Templates (obsolete team, but included as it still exists)
- Council Onboarding has not been included as it contains live flows


jessicamcinchak (approved):
Great!

PR #6996, opened 2026-07-10
feat: add the postcode, outcode, incode, title, and street from address data to user passport

augustlindemer:
This PR adds more of our address data to the user passport to allow for automation in services.


jessicamcinchak (approved):
Perfect !

PR #6998, opened 2026-07-10
chore: Remove old Lambeth cert details

DafyddLlyr:
Not deploy sensitive now the CDN has been removed, this is just a tidy up task from https://github.com/theopensystemslab/planx-new/blob/main/doc/how-to/how-to-setup-custom-domains.md (step B12)


jessicamcinchak (approved):
If the "new" method ended up coming through, I think we can remove _all_ Lambeth details from this file now (as previously done for Doncaster, etc)? 

Eg `ssl-lambeth-cert` and `ssl-lambeth-key` in addition to `ssl-lambeth-chain` removed here?

DafyddLlyr:
🤦 I've removed them all locally but not committed - thanks for checking my homework here..!

Updated in [2b73684](https://github.com/theopensystemslab/planx-new/pull/6998/commits/2b736848a17690d53c8d620ebe1b968c0b19339a)

jessicamcinchak (approved):

PR #7002, opened 2026-07-10
fix: ensure test suites run when bumping shared packages in pnpm-workspace.yml

jessicamcinchak:
Noticed while opening #7001 - see https://opensystemslab.slack.com/archives/C01E3AC0C03/p1783703935338949


DafyddLlyr (approved):
Nice solution..!

Yep, this was certainly missed - good spot. Once Turborepo is in one place and we have [task graphs](https://turborepo.dev/docs/core-concepts/package-and-task-graph#task-graph) set up I _think_ that this will allow us to drop the `needs.changes.outputs` approach, as it will be baked into the lint/test/build processes.

PR #7003, opened 2026-07-13
feat: add 'Customise' tab to source templates

augustlindemer:
This PR adds a 'Customise' tab to the side bar of source templates.

Removes the 'Preview' tab from Source Templates, principally to space the new tab and given current inability to actually test source templates directly. Potentially to be re-introduced following future 'source template testing' feature suite.

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Customisations/index.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Customisations/index.tsx
@@ -5,13 +5,25 @@ import Typography from "@mui/material/Typography";
 import { sortIdsDepthFirst } from "@planx/graph";
 import DelayedLoadingIndicator from "components/DelayedLoadingIndicator/DelayedLoadingIndicator";
 import { useStore } from "pages/FlowEditor/lib/store";
-import React from "react";
+import React, { useEffect } from "react";
 
 import { CustomisationCard } from "./CustomisationCard";
 import type { FlowEdits } from "./types";
 
 const Customisations = () => {
-  const [flowId, flow] = useStore((state) => [state.id, state.flow]);
+  const [flowId, flow, isTemplatedFrom, isTemplate, setOrderedFlow] = useStore(
+    (state) => [
+      state.id,
+      state.flow,
+      state.isTemplatedFrom,
+      state.isTemplate,
+      state.setOrderedFlow,
+    ],
+  );
+
+  useEffect(() => {
+    if (flow) setOrderedFlow();
+  }, [flow, setOrderedFlow]);
 
   // Get the nodes within this flow that are customisable and sort them top-down to match graph
   const customisableNodeIds = new Set(
@@ -37,6 +49,7 @@ const Customisations = () => {
       variables: {
         flow_id: flowId,
       },
+      skip: !isTemplatedFrom,
     },
   );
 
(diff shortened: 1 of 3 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Customisations/index.tsx
```diff
@@ -5,13 +5,25 @@ import Typography from "@mui/material/Typography";
...
+    (state) => [
+      state.id,
+      state.flow,
+      state.isTemplatedFrom,
+      state.isTemplate,
+      state.setOrderedFlow,
+    ],
+  );
+
+  useEffect(() => {
```
augustlindemer, line 24: similar to the review tab it will otherwise render the customisation cards prior to having their data and not re-render unless manually refreshed.

apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Customisations/index.tsx
```diff
@@ -37,6 +49,7 @@ const Customisations = () => {
       variables: {
         flow_id: flowId,
       },
+      skip: !isTemplatedFrom,
```
augustlindemer, line 52: I'm not sure this is needed (doesn't seem to error without), but maybe good to explicitly skip? Happy for alternative suggestions!

augustlindemer:
E2E just times out without helpful errors. Could there be a test that requires the removed 'Preview' tab in source templates? Can't pinpoint which one that would be. Maybe one of the UI driven node creation ones that expect to find its title from the Preview tab?

jessicamcinchak:
Think I've pinpointed what's going on with your E2E failures here! 

If you check GitHub logs, this is the snippet that's actually helpful (recognise there's a LOT of noise to sift through here): 
Screenshot from 2026-07-14 15-37-55

It tells us which specific test has hung, and that the templated folder name _is_ visible, but actually TWICE (once or "exact match" is implicitly required by "strict mode"). 
  - If we think about the graph page, this makes sense - previously the templated folder name would have been exclusively visible in the graph, now it's in the graph _and_ default-visible "Customise" tab node card

To fix, I think we just want to use a stricter qualifier here than `getByText`. In e2e ui-driven `templates.spec.ts`, on line 235, I'd try: 
``` diff
- await expect(page.getByText(TEMPLATED_FOLDER_NAME)).toBeVisible();
+ await expect(page.getByRole("link", { name: TEMPLATED_FOLDER_NAME })).toBeVisible();
```

apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Customisations/index.tsx
jessicamcinchak: Yep agree makes sense to be explicit!

apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Customisations/index.tsx
```diff
@@ -60,12 +73,24 @@ const Customisations = () => {
...
+          </Typography>
+        </>
+      )}
+      {isTemplate && (
+        <>
+          <Typography variant="h4" sx={{ mb: 1 }}>
+            Nodes set to 'Allow edits'
+          </Typography>
+        </>
+      )}
```
jessicamcinchak, line 93: 👍 Nice distinction

jessicamcinchak (commented):
Changes all look great here, thanks for picking this one up ! 

Left a comment/suggestion around e2e failure, if still giving you grief - please shout and happy to look at again / can jump on branch to work out 🙂

augustlindemer:
> To fix, I think we just want to use a stricter qualifier here than `getByText`. In e2e ui-driven `templates.spec.ts`, on line 235, I'd try:
> 
> ```diff
> - await expect(page.getByText(TEMPLATED_FOLDER_NAME)).toBeVisible();
> + await expect(page.getByRole("link", { name: TEMPLATED_FOLDER_NAME })).toBeVisible();
> ```

wow nice spot, worked like a charm!

jessicamcinchak (approved):

PR #7004, opened 2026-07-14
feat: notes frontend

joecarver:
### What's in this PR?
The core functionality for the new notes frontend:
- Add notes attached to nodes (incl options)
- Add notes via hangers to the graph itself
- Notes can be added via the context menu on right click or via the new component popup menu
- When deleting a node: 
    - notes attached to it are also deleted
    - notes after it are re-positioned after the next preceding node
- Some unit and UI tests for key functionality

### What's not in this PR?
- script to convert notes from old format to new
- handle notes when copy+pasting nodes
These are both in progress and coming in separate PRs

### Key files in this PR
It's pretty big in terms of lines and files touched but there's a bunch of boilerplate/tests/less interesting stuff. Most of the logic can be found in:
- `apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/lib/notePlacement.ts` - graph traversal for figuring out `{before, parent}` values for notes
- `apps/editor.planx.uk/src/pages/FlowEditor/lib/store/notes.ts` - queries and business logic for calling the above functions

### Some assumptions I've made which are up for discussion
- Notes are editable and deleteable by any member of the team, not just the creator
- When you move a node that has a note following it (not attached), the note following it moves with the parent
- Notes cannot be attached directly after a question and before an option as we do not render hangers here - this one I feel like maybe we do want, but I was having issues with adding new hangers so wanted to check before spending more time on i
t.
- We do not insert new hangers before notes either, so if you want multiple notes in the same graph position they just stack on each other. I think it'd add a new level of complexity to do this because notes are no longer real nodes so they cannot be positioned relative to one another

### DIFF
--- a/apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
+++ b/apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
@@ -1,0 +1,87 @@
+import { gql, useSubscription } from "@apollo/client";
+import { useStore } from "pages/FlowEditor/lib/store";
+
+export const DEFAULT_NOTE_COLOR = "#fffdb0";
+
+export interface NotePlacement {
+  parent: string;
+  before?: string;
+}
+
+export interface FlowNote {
+  id: string;
+  flowId: string;
+  nodeId: string | null;
+  placement: NotePlacement | null;
+  text: string;
+  color: string;
+  createdBy: number;
+  updatedBy: number;
+  createdAt: string;
+  updatedAt: string;
+}
+
+const GET_FLOW_NOTES = gql`
+  subscription GetFlowNotes($flowId: uuid!) {
+    flow_notes(where: { flow_id: { _eq: $flowId } }) {
+      id
+      flow_id
+      node_id
+      placement
+      text
+      color
+      created_by
+      updated_by
+      created_at
+      updated_at
+    }
+  }
+`;
+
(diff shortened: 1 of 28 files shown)


DafyddLlyr (changes requested):
I want to push back here on the size of the PR - this is too big I think for us to effectively review, test, and iterate on in one go. We don't have hard rules, or a `CONTRIBURING.md` (maybe we should?) but generally we aim for max 500 lines (where possible). The aim is to ensure that we hit the sweet spot of comprehensibility and speed of iteration - from experience when we get large PRs they tend to drift towards one of two outcomes - either a skim and "LGTM! ✅ " approval, or a quagmire of 30+ comments which is just horrible to work through and track.

Looking at the very clear commit history and file changes, it looks like there's a few ways of splitting this one meaningfully. Maybe something like...

 - Note component + styling (Storybook?)
 - note placement logic + tests
 - create/edit/delete entry points
 - graph placement / drag and drop

Most of these would hopefully end up pretty small/simple and then we'd get maximum attention on the meaningful bits - presumably node placement as you highlighted in the PR description.

We also want to think about the migration plan here (in terms of both data and feature roll out). In order to migrate the data we (presumably) want to turn off the current "create note" (empty question) method, migrate data (via script in https://github.com/theopensystemslab/planx-data-migrations), then enable this new functionality.

PR #7005, opened 2026-07-14
fix: Design settings team name

ianjon3s:
## What does this PR do?

Fixes display of team name on settings page:
- Ensures text is visible with `contrastText`
- Drops `PlanX /` prefix, as we removed this on the public header a few months back

**Before:**
image

**After:**
image

**Testing:**
https://7005.planx.pizza/app/a-new-team/settings/design


jessicamcinchak (approved):
Great, thank you for catching this !

PR #7008, opened 2026-07-15
feat: Improve usability of node drag and drop

ianjon3s:
## What does this PR do?

When dragging a component in the PlanX editor it is not immediacy communicated where the component can be moved to. We know the following is true:
- Immediate sibling hangers cannot be dropped into, as they would present no relative movement
- Other hangers can be dropped into, but they do not indicate as such

Update to give the droppable hangers a distinct style with + character, making it obvious where the potential drop targets are.

https://github.com/user-attachments/assets/bc90740f-a537-4a81-83f8-50ddb06e8d3f

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Hanger.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Hanger.tsx
@@ -8,7 +8,7 @@ import {
   nodeIsTemplatedInternalPortal,
 } from "pages/FlowEditor/utils";
 import React, { useCallback } from "react";
-import { useDrop } from "react-dnd";
+import { useDragLayer, useDrop } from "react-dnd";
 
 import { useStore } from "../../../lib/store";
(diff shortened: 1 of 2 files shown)


DafyddLlyr (approved):
Great - works as expected and having a visual cue is very helpful.

Only small nit/question would be the use of the ➕ symbol - this feels like it means "create" or "add" not "drop off" / "move" / "relocate here". Might just be me though, happy to await user feedback and it's an easy change in future if required..!

ianjon3s:
That's a good point, we do use the plus symbol heavily for an "add new" operation. I'll put together an alternative

ianjon3s:
I've updated with something that closer resembles the file dropzone, and also increased the drop target in the process. Seems more consistent (and immediately obvious) as a target?


https://github.com/user-attachments/assets/f44a528e-7d18-4f1c-be4b-b3b66179bbbe

DafyddLlyr (approved):
This is much clearer and more consistent I think...! 👍

PR #7009, opened 2026-07-15
feat: allow `platformAdmins` to author patterns

jessicamcinchak:
See https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1784169341576099


apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Publish/CheckForChangesButton.tsx
```diff
@@ -96,6 +96,7 @@ export const CheckForChangesToPublishButton: React.FC<{
 
   const isDisabled =
     !useStore.getState().canUserEditTeam(teamSlug) ||
+    useStore.getState().isPattern ||
```
jessicamcinchak: Here's an interesting question I hit: do patterns need to be publish-able? Would it be sufficient for `platformAdmins` to control when they are visible/available or not to select/insert _only_ ?

If we do make them publishable, then:
- Assumption becomes that when I'm inserting an available pattern I should be inserting its' latest published version
- If I fetch `published_flows.data`, this is going to be a _flattened_ data structure
- Before inserting/pasting in the pattern into my own graph, I'll need to _unflatten_ it (we haven't crossed this bridge yet! Possible to write an inverse of `dataMerged` but will add complexity!)

Key example / user story here I think is "Find property" pattern which will use a nested flow (and I don't think we'll ever care about the published content of that nested flow in the context of patterns, only inserting a live reference to it).

apps/editor.planx.uk/src/pages/Flows/components/AddFlow/BaseFormSection.tsx
```diff
@@ -72,6 +72,16 @@ export const BaseFormSection: React.FC = () => {
...
+            <Switch
+              name="isPattern"
+              checked={values.flow.isPattern}
+              onChange={() =>
+                setFieldValue("flow.isPattern", !values.flow.isPattern)
+              }
+              label={
+                "Pattern (exclusive; cannot be source template or service)"
+              }
+            />
```
jessicamcinchak, line 82: This is admittedly pretty odd here ! BUT as discussed at design call: 
- We want patterns to be `flows` in their data structure so that ShareDB "just works", immediately giving us familiar graph operations and interactions, edit history, etc
- We could build an "Add pattern" modal in global settings and present a catalog there, but I realised we'll quickly hit up against questions around which parts of the larger `FlowsLayout` to rebuild or strip away (archive, filter and sorting, search, etc all seem quite useful to keep available!)
- _Only_ `platformAdmins` will ever see options to create patterns _and_ actual patterns as flow entries within a team. "Normal" `teamEditors` will exclusively discover patterns via the "Add component" modal.

I _think_ this solution feels right for MVP, and ultimately shouldn't limit any future design shifts here - eg filter all the `flows.isPattern` entries and present them elsewhere in the editor more intentionally! There might be a good case for doing the same with source templates honestly!

apps/editor.planx.uk/src/pages/Flows/components/AddFlow/types.ts
```diff
@@ -7,6 +7,7 @@ export const validationSchema = object().shape({
     slug: string().required("Slug is required"),
     name: string().required("Name is required"),
     teamId: number().integer().required("Team ID is required"),
+    isPattern: boolean(), // TODO error if isTemplate or isService are true?
```
jessicamcinchak, line 10: I've added a Postgres-level `CHECK` constraint for this, but not yet enforced in UI. Poor error feedback currently only visible to `platformAdmins` at least

apps/editor.planx.uk/src/pages/Flows/components/FlowCard/styles.tsx
```diff
@@ -29,6 +29,23 @@ export const CardBanner = styled(Box)(({ theme }) => ({
   gap: theme.spacing(0.4),
 }));
 
+export const PatternCardBanner = styled(Box)(({ theme }) => ({
+  width: "100%",
+  // TODO Ian to update? Fun but a bit easter egg
```
jessicamcinchak, line 34: cc @ianjon3s 😅 

Image

apps/editor.planx.uk/src/theme.ts
```diff
@@ -104,6 +104,12 @@ const DEFAULT_PALETTE: Partial<PaletteOptions> = {
     dark: "#C099FF",
     icon: "#380F77",
   },
+  pattern: {
+    main: "#f5ffdb",
+    light: "#f5ffdb",
+    dark: "#e3f1be",
+    icon: "#568203",
+  },
```
jessicamcinchak, line 112: cc @ianjon3s again - very much not set on any of these 🥑 hues, suggestions most welcome!

apps/hasura.planx.uk/metadata/databases/default/tables/public_flows.yaml
```diff
@@ -230,27 +232,32 @@ select_permissions:
...
-        - category
-        - created_at
+        - is_pattern
+        - is_service
+        - is_template
+        - summary
         - creator_id
-        - data
+        - team_id
+        - settings
```
jessicamcinchak: Any ideas why it's noisily re-ordered fields here? Was hoping this would be less common with new pnpm workspace/turbo repo setup! Am I possibly missing something locally?

apps/hasura.planx.uk/migrations/default/1784132841864_alter_table_public_flows_add_column_is_pattern/up.sql
```diff
@@ -0,0 +1,8 @@
+alter table "public"."flows" add column "is_pattern" boolean
+ not null default 'false';
+alter table "public"."flows" add constraint "pattern_check" check (is_pattern = TRUE
+ AND templated_from IS NULL
+ AND is_template = FALSE
+ AND is_service = FALSE
+ AND is_listed_on_lps = FALSE
+OR is_pattern = FALSE);
```
jessicamcinchak, line 8: We _never_ want a pattern to be combined with _at least_ these flow properties ! I am assuming we may want to "copy" patterns

apps/editor.planx.uk/src/pages/Flows/components/AddFlow/BaseFormSection.tsx
jessicamcinchak: Another alternative to global settings could also be dedicated "Patterns" tab here? Perk would be conditional "Add a new pattern" button and simplified modal rather than "Add a new flow" with extra toggle, but then possible con is "Archive" is mix of flows _and_ patterns or we need 2x archives?? Again maybe a bit odd :shrug: 
image

apps/api.planx.uk/modules/flows/createFlow/controller.ts
```diff
@@ -19,6 +19,7 @@ export const createFlowSchema = z.object({
     name: z.string().trim(),
     isTemplate: z.boolean().optional().default(false),
     isService: z.boolean().optional().default(false),
+    isPattern: z.boolean().optional().default(false),
```
DafyddLlyr, line 22: This is mitigated against below with the constraints check on `INSERT` but I think this pattern generally shows us that an instance of a flow can (should?) have a single "type" - template, service, pattern, flow (default).

apps/editor.planx.uk/src/pages/Dashboard/components/FlowsPanel.tsx
```diff
@@ -177,7 +199,9 @@ export const FlowsPanel: React.FC<FlowsPanelProps> = ({
 export default function ConnectedFlowsPanel() {
   const team = useStore((state) => state.getTeam());
   const { data, loading } = useGetFlows(team.id);
-  const flows = data?.flows ?? [];
+
+  // Patterns never need to be visible on dashboard
+  const flows = data?.flows?.filter((flow) => !flow.isPattern) ?? [];
```
DafyddLlyr, line 204: Alternatively: Check against the `is_pattern` column in `GET_FLOWS` to exclude these`, and just get patterns where we need them (e.g. on another tab?).

DafyddLlyr: Ah I can see a comment about this below - the "archive" question is a good one, I think it's a case of ensuring that...
 - the flows tab fetches only flows via `GET_FLOWS`
 - the patterns tab fetches only patterns via a new query `GET_PATTERNS`
 - the archive tab fetches **all** (update clauses on `GET_ARCHIVED_FLOWS`)

This would give a single archive tab, with intermingled patterns and flows which I think is ok / intelligible?

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Flow/Visibility/index.tsx
```diff
@@ -22,9 +22,9 @@ const VisibilitySettings: React.FC = () => {
 
   return (
     <>
-      <IsService />
-      {isService && <FlowStatus />}
-      <FlowCopy isService={isService} />
+      {!isPattern && <IsService />}
+      {(isService || isPattern) && <FlowStatus />}
+      {!isPattern && <FlowCopy isService={isService} />}
       {isService && <LPSListing />}
```
DafyddLlyr, line 28: Another example where a `flowType` might prove helpful?

See routing comments in main review body - another solution is handling this via routing, we maybe just need a `patterns/settings` page.

apps/api.planx.uk/modules/flows/createFlow/controller.ts
DafyddLlyr: Appreciate this isn't necessarily a small change - there's DB migrations, cascading type changes, updated to conditionals etc. There might be an lighter weight approach where the application data model maps theses db columns to/from a `flowType` value (or we use a computed field). None of these are ideal though...!

apps/editor.planx.uk/src/pages/Flows/components/AddFlow/types.ts
DafyddLlyr: A few options I guess if we want to tackle this (Yup is not as nice as Zod here for this, I don't think there's a built in way to make union types).

1. A `flowType` enum (see previous comments)
2. A `lazy()` union + `switch/case` (see `EnhancedTextInput`)
3. Adding a `test()` block to the validation schema to ensure it's valid (`isPattern` is not in an impossible state)

apps/editor.planx.uk/src/pages/Flows/components/FlowTable/styles.tsx
```diff
@@ -48,24 +49,34 @@ export const StyledTableHead = styled(TableHead)(({ theme }) => ({
 }));
 
 export const StyledTableRow = styled(TableRow, {
-  shouldForwardProp: (prop) => prop !== "isTemplated" && prop !== "clickable",
-})<{ isTemplated?: boolean; clickable?: boolean }>(({
+  shouldForwardProp: (prop) =>
+    prop !== "isTemplated" && prop !== "isPattern" && prop !== "clickable",
```
DafyddLlyr, line 53: Well remembered..!

apps/hasura.planx.uk/metadata/databases/default/tables/public_flows.yaml
DafyddLlyr, line 1: In terms of permissions, I think ideally we want...
 - `teamAdmins` and `teamEditors` to have read access to patterns
 - `platformAdmins` to have full CRUD
 - Everybody else (e.g. `public`) - I'm not sure...! No harm in read access but it's not really needed?

I can't think of a _particularly_ simple way of managing this. 

Hasura row level permissions will work - certain roles get read access, but we update `INSERT`, `UPDATE` and `DELETE` blocks to check the status of this column. Something like this would work as long as we also toggle off column write access for `is_pattern`. 

```jsonc
{
  // Add this _and wrapper...
  "_and": [
    {
      "team": {
        "members": {
          "_and": [
            {
              "user_id": {
                "_eq": "x-hasura-user-id"
              }
            },
            {
              "role": {
                "_eq": "teamEditor"
              }
            }
          ]
        }
      }
    },
    // ..then add this check
    {
      "is_pattern": {
        "_eq": false
      }
    }
  ]
}
```

The downside is it's a little fiddly to manage and isn't very expandable. But it should work well for now..!

Other options would be a `patterns` table - but this isn't that viable an option (see ShareDB comments).

This might be a non-issue though - if we decide that patterns may be a non-platform admin concern in future, the current `team_members` check prevents writes to patterns made by us - we just allow users to create/update/destroy their own patterns via API without a corresponding UI which will get flagged as OWASP "Broken Access Control".

DafyddLlyr (commented):
Appreciate this is just a draft, but I've had a quick pass through with a few high-level comments. As always, happy for you to apply judgement as to what will work, and what's outside the scope of an MVP. Very much in favour of a speedy delivery here..!

### ShareDB
Following on from our conversation yesterday - our assumptions were correct re:new `patterns` table and ShareDB. We've hardcoded a few places that that ShareDB connects to the `flows` table, notably in `apps/sharedb.planx.uk/sharedb-postgresql.js` - it's not as simple as writing a new `connectToPattern()` which calls `connect("patterns", id)` in the frontend. The ShareDB docs show that this hardcoded "flows" string is something we've introduced on fork - there's a `_collections` param we should be using to make this properly generic (allowing us to connect easily to `flows` / `patterns` / `schemas` "docs"). I'm not that keen to explore this right now - it would sit better alongside a wider look at ShareDB and possible alternatives I think unless we start hitting against the edges of this.

There's a few permissions / querying implications of this I think but hopefully no blockers - more on this below.

Based on all this I'm fully on board with "patterns" being instances of the `flows` table initially 👍 

### Routing
Managing patterns at a given route might make some of the UI setup less tedious/repetitive. For example, if a pattern lives at `/:team/patterns/:slug` we could make reference this once up top to manage layouts (e.g. publish UI, setting pages).

I think the same would apply if we places these at `/patterns/:slug`, outside the scope of a team. We'd need to update a few flow editor components to have fallbacks for missing team details though which might lead to a few cascading changes. The upside is there's a stronger divide between flows and patterns, and we can more easily keep these concerns separate. Downside would be a bit more work upfront to work through making these components agnostic of team (and managing flows with `null` `team_id` value I guess.

As this was already discussed with the content team at a design call, and the current implementation has been demoed without issue I'm very happy to move ahead with the current approach - it's pragmatic and will work well. Also very happy to have a quick pass at trying out these routing questions + re-working of `FlowEditor` components to handle re-use outside a `/:team` context.

apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Publish/CheckForChangesButton.tsx
DafyddLlyr: Agree that "publishing" (in the sense of the current procedure) doesn't really make sense. It might be the case we just drop in a slightly different UI component here which does a "pattern publish" which is a different thing but still affords Editors the ability to batch changes and set them "live".

apps/hasura.planx.uk/metadata/databases/default/tables/public_flows.yaml
DafyddLlyr: I think this is just Hasura authoring these files - they also get re-ordered sometimes on change for me. 

If you go to a random Hasura `yaml` file, make a change in-IDE (e.g. change a column name from `settings` to `settings_jess_test` and save the file, I would guess it's not your IDE re-ordering these and sorting?

I reckon Hasura sorts on update but no insert or some permutation like this.

apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Publish/CheckForChangesButton.tsx
jessicamcinchak: Per #planx-design thread, seems we're in consensus that they shouldn't change often and we can proceed without publishing for now :heavy_check_mark:

apps/api.planx.uk/modules/flows/createFlow/controller.ts
jessicamcinchak: There's also the complication here that source templates can be _flows_ or _services_ - they are not 4 mutually exclusive categories/types!

Going to largely ignore this in this PR as a change will have a lot of cascading noise as noted, but agree one to keep an eye on !

apps/editor.planx.uk/src/pages/Dashboard/components/FlowsPanel.tsx
jessicamcinchak: Yes, the unique route option would be a bit cleaner for permissions. 

Per #planx-design thread, I think direction for now is to proceed with _current tabs_ and Ian's latest "create" modal which account for patterns _within_ "from scratch" - no votes or clear preference for the tabbed design proposal expressed there :heavy_check_mark: 

Again though, since only us admins will see this - would be a very low friction change to filter patterns out of "flows" tab in the future into their own!

I think if we go the separate tabs route in the future, there is a bit of a new visual language question - eg if "Patterns" are in their own tab we don't have to "pick a colour banner", they'd just be naturally filtered. But then if "Archive" is a combo, we're back to them needing a visual cue - which is maybe confusing condsidering "Flows" tab has many categories of things within ("flows", "services", "source templates")?

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Flow/Visibility/index.tsx
jessicamcinchak: Have updated this to match Silvia's feedback / be consistent with source templates - we actually want to use/expose `FlowCopy` to control pattern discover-ability, _not_ `FlowStatus` :white_check_mark: 

Per other comments, have _not_ introduced a `flowType` here to avoid noise/sprawl, but perhaps one to revisit!

apps/hasura.planx.uk/metadata/databases/default/tables/public_flows.yaml
jessicamcinchak: Yes good shout on insert/update row-level permissions here - have added :white_check_mark:

DafyddLlyr (approved):

PR #7010, opened 2026-07-16
fix(editor): Manage overflow in `AddFlow` modal

DafyddLlyr:
Small visual issue I spotted when testing https://github.com/theopensystemslab/planx-new/pull/7009

The `AddFlow` modal has an invisible overflow, instead of the standard fixed `DialogAction` element (like `FormModal`). Removing the redundant wrapping `Box` solves this. 

_`EditorModal` for reference showing how this is usually handled_
image

| Before | After |
|--------|--------|
| image | image |


ianjon3s (approved):
Nice catch and solution 👍

PR #7011, opened 2026-07-16
docs: Update custom subdomain guide

DafyddLlyr:
## What does this PR do?
Reworks `doc/how-to/how-to-setup-custom-domains.md` after running through the process last week and getting caught out in a few places.

The changes made in this PR can be read as markdown here - https://github.com/theopensystemslab/planx-new/blob/a93096360ce90a90c7418cbde1a6c7515cc22763/doc/how-to/how-to-setup-custom-domains.md

#### Dropped the two-part CNAME process
Councils just get both records (validation + routing) at once, with a single table explaining each. Removed the confusing "send this one now, that one later" notes scattered through both flows which are basically redundant. This is they key change and motivation for the rework.

#### Linear processes
There's a bit more text and repetition, but less jumping between steps and both processes are self-contained

#### Added state diagrams & explanation
A Mermaid lifecycle diagram plus a state → infrastructure table (mining cert / shared cert / legacy CDN / alias location / who serves traffic) so the `cloudFrontState` values are actually legible. I found myself coming to this process a little cold after each update from a council and it's helpful having a clear understanding of the states and high-level processes here.

#### Updated troubleshooting section
Covers a few additional things I hit...!

### DIFF
--- a/doc/how-to/how-to-setup-custom-domains.md
+++ b/doc/how-to/how-to-setup-custom-domains.md
@@ -4,329 +4,353 @@
 
 Teams can access PlanX via a custom subdomain on their own domain (e.g. `https://planningservices.medway.gov.uk/`).
 
-Custom domains are ideally served by a **shared CloudFront distribution** backed by a single DNS-validated ACM certificate (which we abbreviate as 'cert' hereafter). This replaces the legacy model where each council's custom domain had its own dedicated CloudFront distribution and had to provide us with an SSL cert.
+Custom domains are served by a single **shared CloudFront distribution** ("shared CDN") backed by a single **DNS-validated ACM certificate** ("shared cert") that we provision. This replaces the legacy model where each council had its own dedicated CloudFront distribution backed by a council-provided SSL certificate.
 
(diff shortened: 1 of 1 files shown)


doc/how-to/how-to-setup-custom-domains.md
```diff
@@ -4,329 +4,353 @@
...
 
-    > NB. **If this is not the first domain on the shared CDN** and you want to expedite:
-    > - Do steps 5-6 **now** and send the council both DNS records together (you still have to wait for them to add the records before proceeding to step 2).
+```sh
+cd infrastructure/certificates
+pulumi up --refresh --stack production
+```
 
-3. Advance the domain in question to the `cutover-ongoing` state by updating `customDomains.ts`:
+> 🚩 **Sanity-check the Pulumi preview before confirming.** A correct run _adds_ an SAN and **replaces** `sslCert-custom-domains` (Pulumi status - `retain[replace]`) - the SAN count should go **up**. If you see the shared cert or mining cert being **deleted outright**, or the SAN count shrinking, you are almost certainly on the wrong Pulumi stack or using the wrong AWS credentials. Do not proceed.
```
zz-hh-aa, line 230: 🙏

zz-hh-aa (approved):
V clear, thanks for all the detail :)

jessicamcinchak (approved):

freemvmt:
This is great, thanks Daf! ✨

PR #7018, opened 2026-07-16
feat: add team settings email as Confirmation component default

augustlindemer:
This PR adds responsive default content to the Confirmation component.

Instead of defaulting to the placeholder 'ADD YOUR COUNCIL CONTACT', which has unintentionally ended up in live services, the Confirmation component now reads from the team settings helpEmail and set it as the default contact. Principally a QoL change that makes it easier adding Confirmation components to flows.

helpEmail should always be available for all teams, but I still retained the placeholder as a fallback just in case.

Limitation: If helpEmail itself is a placeholder (e.g. example@council.gov.uk) or not available, this still doesn't prevent inadequate content from potentially going live.


DafyddLlyr (approved):
Nice solution 👌

PR #6725, opened 2026-05-26
chore(sharedb): Remove node-dev dependency (CVE-2026-41907)

DafyddLlyr:
Fixes https://github.com/theopensystemslab/planx-new/security/dependabot/889, closes https://github.com/theopensystemslab/planx-new/pull/6718

`node-dev` can be replaced by `node --watch` since Node v19 👍

PR #7020, opened 2026-07-16
chore(sharedb): CVE-2026-41907

DafyddLlyr:
Fixes https://github.com/theopensystemslab/planx-new/security/dependabot/1052

### DIFF
--- a/apps/sharedb.planx.uk/package.json
+++ b/apps/sharedb.planx.uk/package.json
@@ -13,8 +13,5 @@
   "scripts": {
     "dev": "node --watch server.js",
     "start": "node server.js"
-  },
-  "devDependencies": {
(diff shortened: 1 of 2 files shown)


apps/sharedb.planx.uk/package.json
```diff
@@ -13,8 +13,5 @@
   "scripts": {
     "dev": "node --watch server.js",
     "start": "node server.js"
-  },
-  "devDependencies": {
-    "node-dev": "^8.0.0"
```
DafyddLlyr, line 18: This was actually introduced during a rebase of the monorepo work I believe - this was originally removed here and is not relied on at all - https://github.com/theopensystemslab/planx-new/pull/6725

jessicamcinchak (approved):

PR #7024, opened 2026-07-16
feat: Highlight adding of new nodes into graph

ianjon3s:
## What does this PR do?

- Introduces a "pulse" style highlight when a new node is added or pasted into the graph
- Adds `useFlashOnNodeAdded` hook to flash a node's border when it's first added to the graph

https://github.com/user-attachments/assets/8a5481eb-f351-4615-9bb7-2b9b41a9d259

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Question.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Question.tsx
@@ -67,13 +67,12 @@ const Question: React.FC<Props> = React.memo((props) => {
   const hasHelpText =
     props.data?.policyRef || props.data?.info || props.data?.howMeasured;
 
-  const ref = useFlashOnNodeAdded<HTMLLIElement>(props.id);
+  const ref = useFlashOnNodeAdded<HTMLDivElement>(props.id);
 
   return (
     <>
       <Hanger hidden={isDragging} before={props.id} parent={parent} />
       <li
-        ref={ref}
         className={classNames(
           "card",
           "decision",
@@ -87,6 +86,7 @@ const Question: React.FC<Props> = React.memo((props) => {
         )}
       >
         <TemplatedNodeContainer
+          ref={ref}
           isTemplatedNode={props.data?.isTemplatedNode}
           areTemplatedNodeInstructionsRequired={
             props.data?.areTemplatedNodeInstructionsRequired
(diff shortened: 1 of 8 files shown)


apps/editor.planx.uk/src/ui/editor/TemplatedNodeContainer.tsx
```diff
@@ -82,14 +82,20 @@ const getTemplatedNodeStatus = (
...
 
-export const TemplatedNodeContainer: React.FC<TemplatedNodeContainerProps> = ({
-  children,
-  isTemplatedNode = false,
-  areTemplatedNodeInstructionsRequired = false,
-  isComplete = false,
-  showStatus = false,
-  className,
-}) => {
+export const TemplatedNodeContainer = forwardRef<
```
DafyddLlyr: As of React 19 I don't think you need `forwardRef` here - `ref` is just available as a prop now. You'll just need to declare it in the interface with an additional `ref?: Ref<HTMLDivElement>;`

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Question.tsx
```diff
@@ -103,7 +107,7 @@ const Question: React.FC<Props> = React.memo((props) => {
             }}
             preload={false}
             onContextMenu={handleContextMenu}
-            ref={(el) => {
+            ref={(el: HTMLAnchorElement | null) => {
```
DafyddLlyr: Is this needed, or can it just be picked up via inference?

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
```diff
@@ -301,7 +301,9 @@ export interface Template {
 export interface EditorStore extends Store.Store {
   cutNode: (id: NodeId, parent: NodeId) => void;
   getCutNode: () => CutPayload | null;
-  addNode: (node: any, relationships?: Relationships) => void;
+  addNode: (node: any, relationships?: Relationships) => NodeId;
+  lastAddedNodeId?: NodeId;
+  clearLastAddedNodeId: () => void;
```
DafyddLlyr: nit: I think `EditorUIStore` is a better home for these given they have a purely UI consideration.

DafyddLlyr (changes requested):
A few very small comments - nice addition..!

apps/editor.planx.uk/src/pages/FlowEditor/lib/store/editor.ts
ianjon3s: Agreed, added to UI store

ianjon3s:
Thanks for these @DafyddLlyr , I've actioned all

DafyddLlyr (approved):
Great...! 👌

PR #7026, opened 2026-07-17
fix(editor): Correct sync of `Restore to this point` for history events

ianjon3s:
## What does this PR do

Current the history "Restore to this point" function is one operation out of sync, and to restore to a point you need to select the subsequent operation.

A good demonstration of this is creating a single operation on a graph, "Restore to this point" is active on the operation rather than the initial state, and "restoring to" the operation will reset to the initial state ("Created flow"):

image

By adjusting the sequencing and making sure there is a "restore to" function on the initial Created flow state, we can ensure the "restore to" always returns to the correct operation state:

image

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
@@ -9,6 +9,7 @@ import Box from "@mui/material/Box";
 import IconButton from "@mui/material/IconButton";
 import Tooltip from "@mui/material/Tooltip";
 import Typography from "@mui/material/Typography";
+import { ROOT_NODE_KEY } from "@planx/graph";
 import type { OT } from "@planx/graph/types";
 import type {
   CommentHistoryItem,
@@ -87,10 +88,16 @@ export const EditHistoryTimeline = ({
     const hasNewerOperation = events
       .slice(0, i)
       .some((e) => e.type === "operation");
+
+    // Once the flow is already empty, do not allow restoring to the initial "Created flow" operation
+    const isAlreadyAtInitialState =
+      !event.actorId && !flow[ROOT_NODE_KEY]?.edges?.length;
+
     return (
       event.type === "operation" &&
       canUserEditTeam(teamSlug) &&
-      hasNewerOperation
+      hasNewerOperation &&
+      !isAlreadyAtInitialState
     );
   };
 
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
@@ -83,22 +83,22 @@ export const EditHistoryTimeline = ({
     return ["operation", "comment"].includes(type);
   };
 
+  // Show restore only for editable, undoable operations that have something newer to undo
   const showUndoButton = (event: HistoryItem, i: number): boolean => {
-    // Only show the restore button for operations within teams I can edit, and only when an operation to actually restore from exists
+    if (event.type !== "operation") return false;
+    if (!canUserEditTeam(teamSlug)) return false;
+
     const hasNewerOperation = events
       .slice(0, i)
       .some((e) => e.type === "operation");
+    if (!hasNewerOperation) return false;
 
     // Once the flow is already empty, do not allow restoring to the initial "Created flow" operation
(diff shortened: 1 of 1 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
```diff
@@ -48,8 +49,8 @@ export const EditHistoryTimeline = ({
     ]);
 
   const handleUndo = (i: number) => {
-    // Get all events since & including the selected one
-    const eventsToUndo = events.slice(0, i + 1);
+    // Get all events since the selected one, excluding it
+    const eventsToUndo = events.slice(0, i);
```
DafyddLlyr, line 51: Good spot...! I would guess this is a regression introduced here which makes the `+ 1` redundant now https://github.com/theopensystemslab/planx-new/pull/6762

apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
```diff
@@ -82,12 +83,21 @@ export const EditHistoryTimeline = ({
...
+
     return (
       event.type === "operation" &&
-      Boolean(event.actorId) &&
-      canUserEditTeam(teamSlug)
+      canUserEditTeam(teamSlug) &&
+      hasNewerOperation &&
+      !isAlreadyAtInitialState
     );
   };
```
DafyddLlyr, line 99: I'm struggling with comprehension here - there were already a lot of stacked conditional and booleans and we've added to them in this PR.

Have a shot at simplifying this - maybe guard clauses?

DafyddLlyr (changes requested):
One small nit..!

apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
ianjon3s: I believe this has been an issue since the feature was built, as we originally scoped the work around "undo" rather than "restore to"

apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
ianjon3s: Rewritten in a sequence of guard clauses, hopefully easier to follow now

apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
```diff
@@ -83,22 +83,22 @@ export const EditHistoryTimeline = ({
     return ["operation", "comment"].includes(type);
   };
 
+  // Show restore only for editable, undoable operations that have something newer to undo
   const showUndoButton = (event: HistoryItem, i: number): boolean => {
-    // Only show the restore button for operations within teams I can edit, and only when an operation to actually restore from exists
+    if (event.type !== "operation") return false;
+    if (!canUserEditTeam(teamSlug)) return false;
```
DafyddLlyr: Would a wrapper of `<Permission.CanEdit/>` also work here?

Not a required change, but just realised you recently added this.

DafyddLlyr (approved):
Great..!

apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
jessicamcinchak: Since this change slightly changes _definition_ of this feature as you say - can we also get in front of content team please to test before merging just to ensure doesn't break any of their workflows / assumptions? 

Has this change been prompted by bug report? Or just something you noticed Ian?

I don't have a perfect memory of the in-depth reasons for "undo" versus "restore" at the time, but remember it was a bit of a tangly one! Will see if I dig any of this up.

ianjon3s: It's an issue that we've been aware of for some time, in that "Restore to this point" is disjointed from the point in question.

I raised it as a bug last year here : https://opensystemslab.slack.com/archives/C01E3AC0C03/p1744197462671939
but didn't get round to acting on it.

Agree on putting in front of the content team, I'll drop a message in Slack now.

jessicamcinchak (approved):

PR #7027, opened 2026-07-17
fix: Handle browser back navigation from GovPay

DafyddLlyr:
## What does this PR do?
- Handles instances of a user navigating "back" from GovPay (browser back, or copy/pasting URL)
- This (as far as I can tell) is not a regression and has been an issue since day 1, we've just never hit it before

### DIFF
--- a/e2e/tests/ui-driven/src/pay.spec.ts
+++ b/e2e/tests/ui-driven/src/pay.spec.ts
@@ -348,6 +348,47 @@ test.describe("Gov Pay integration @regression", () => {
     await page.locator("a").getByText("View your payment summary").click();
     await expect(page.getByText("Form sent")).toBeVisible();
   });
+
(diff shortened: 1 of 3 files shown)


e2e/tests/ui-driven/src/pay.spec.ts
```diff
@@ -348,6 +348,47 @@ test.describe("Gov Pay integration @regression", () => {
...
+    // Abandon the payment by navigating directly back to PlanX
+    await page.goto(`${previewURL}&sessionId=${sessionId}`);
+
+    await expectSessionResumedAtRetryPayment({
+      page,
+      context,
+      sessionId,
+      adminGQLClient,
+    });
+  });
```
jessicamcinchak, line 391: ✅ thank you!

jessicamcinchak (approved):

PR #7035, opened 2026-07-21
feat: new notes placement logic

joecarver:
**What's in this PR **
As per @DafyddLlyr's suggestion, this is the first of a few PRs comprising the new notes feature. This one has the core logic around how to place notes in a flow, i.e. going from a visual hanger -> `NotePlacement` object, and back. 

**How to review?**
There's no UI here so really just looking for eyes on the logic, the test coverage etc.

Later there will be extra logic for moving/cloning etc that will introduce further behaviour in these files.

### DIFF
--- a/apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
+++ b/apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
@@ -29,7 +29,10 @@ type FlowNoteRow = SnakeCasedProperties<FlowNote>;
 
 const GET_FLOW_NOTES = gql`
   subscription GetFlowNotes($flowId: uuid!) {
-    flow_notes(where: { flow_id: { _eq: $flowId } }) {
+    flow_notes(
+      where: { flow_id: { _eq: $flowId } }
+      order_by: { created_at: asc, id: asc }
+    ) {
       id
       flow_id
       node_id
--- a/apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
+++ b/apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
@@ -67,17 +67,21 @@ const toFlowNote = (row: FlowNoteRow): FlowNote => ({
 interface UseFlowNotesResult {
   notes: FlowNote[];
   loading: boolean;
+  error: unknown;
 }
 
 export const useFlowNotes = (): UseFlowNotesResult => {
   const flowId = useStore((state) => state.id);
 
-  const { data, loading } = useSubscription<QueryResult>(GET_FLOW_NOTES, {
-    variables: { flowId },
-    skip: !flowId,
-  });
+  const { data, loading, error } = useSubscription<QueryResult>(
+    GET_FLOW_NOTES,
+    {
+      variables: { flowId },
+      skip: !flowId,
+    },
+  );
 
   const notes = (data?.flow_notes ?? []).map(toFlowNote);
 
-  return { notes, loading };
+  return { notes, loading, error };
 };
--- a/apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
+++ b/apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
@@ -2,8 +2,6 @@ import { gql, useSubscription } from "@apollo/client";
 import { useStore } from "pages/FlowEditor/lib/store";
 import type { SnakeCasedProperties } from "type-fest";
 
-export const DEFAULT_NOTE_COLOR = "#fffdb0";
-
 /** Anchored immediately after a sibling node */
 export interface SiblingAnchoredPlacement {
   /** id of the preceding sibling node */
--- a/apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
+++ b/apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
@@ -1,0 +1,91 @@
+import { gql, useSubscription } from "@apollo/client";
+import { useStore } from "pages/FlowEditor/lib/store";
+
+export const DEFAULT_NOTE_COLOR = "#fffdb0";
+
+export interface NotePlacement {
+  parent: string;
+  before?: string;
+  /** True when `parent` is the container to insert into (leading/first-child slot) */
+  parentIsContainer?: boolean;
+  /** The container `parent` was resolved against at creation time - disambiguates `parent` when it's a cloned node referenced from multiple containers */
+  container?: string;
+}
+
+export interface FlowNote {
+  id: string;
+  flowId: string;
+  nodeId: string | null;
+  placement: NotePlacement | null;
+  text: string;
+  color: string;
+  createdBy: number;
+  updatedBy: number;
+  createdAt: string;
+  updatedAt: string;
+}
+
+const GET_FLOW_NOTES = gql`
+  subscription GetFlowNotes($flowId: uuid!) {
+    flow_notes(where: { flow_id: { _eq: $flowId } }) {
+      id
+      flow_id
+      node_id
+      placement
+      text
+      color
+      created_by
+      updated_by
+      created_at
+      updated_at
+    }
+  }
+`;
+
+interface FlowNoteRow {
+  id: string;
+  flow_id: string;
+  node_id: string | null;
+  placement: NotePlacement | null;
+  text: string;
+  color: string;
+  created_by: number;
+  updated_by: number;
+  created_at: string;
+  updated_at: string;
+}
+
+interface QueryResult {
+  flow_notes: FlowNoteRow[];
+}
+
+const toFlowNote = (row: FlowNoteRow): FlowNote => ({
+  id: row.id,
+  flowId: row.flow_id,
+  nodeId: row.node_id,
+  placement: row.placement,
+  text: row.text,
+  color: row.color,
+  createdBy: row.created_by,
+  updatedBy: row.updated_by,
+  createdAt: row.created_at,
+  updatedAt: row.updated_at,
+});
+
+interface UseFlowNotesResult {
+  notes: FlowNote[];
+  loading: boolean;
+}
+
+export const useFlowNotes = (): UseFlowNotesResult => {
+  const flowId = useStore((state) => state.id);
+
+  const { data, loading } = useSubscription<QueryResult>(GET_FLOW_NOTES, {
+    variables: { flowId },
+    skip: !flowId,
+  });
+
+  const notes = (data?.flow_notes ?? []).map(toFlowNote);
+
+  return { notes, loading };
+};
--- a/apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
+++ b/apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
@@ -4,20 +4,31 @@ import type { SnakeCasedProperties } from "type-fest";
 
 export const DEFAULT_NOTE_COLOR = "#fffdb0";
 
-export interface NotePlacement {
+/** Anchored immediately after a sibling node */
+export interface SiblingAnchoredPlacement {
+  /** id of the preceding sibling node */
   parent: string;
-  before?: string;
-  /** True when `parent` is the container to insert into (leading/first-child slot) */
-  parentIsContainer?: boolean;
   /** The container `parent` was resolved against at creation time - disambiguates `parent` when it's a cloned node referenced from multiple containers */
   container?: string;
+  parentIsContainer?: false;
+  before?: undefined;
 }
 
-export interface FlowNote {
+/** Anchored to a container's leading/first-child slot (no preceding sibling to anchor to) */
+export interface ContainerAnchoredPlacement {
+  /** id of the container to insert into */
+  parent: string;
+  parentIsContainer: true;
+  before?: string;
+  container?: undefined;
+}
+
+export type NotePlacement =
+  SiblingAnchoredPlacement | ContainerAnchoredPlacement;
+
+interface FlowNoteBase {
   id: string;
   flowId: string;
-  nodeId: string | null;
-  placement: NotePlacement | null;
   text: string;
   color: string;
   createdBy: number;
@@ -25,6 +36,18 @@ export interface FlowNote {
   createdAt: string;
   updatedAt: string;
 }
+
+export interface AttachedNote extends FlowNoteBase {
+  nodeId: string;
+  placement: null;
+}
+
+export interface PositionedNote extends FlowNoteBase {
+  nodeId: null;
+  placement: NotePlacement;
+}
+
+export type FlowNote = AttachedNote | PositionedNote;
 type FlowNoteRow = SnakeCasedProperties<FlowNote>;
 
 const GET_FLOW_NOTES = gql`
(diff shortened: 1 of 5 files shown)


apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export const DEFAULT_NOTE_COLOR = "#fffdb0";
+
+export interface NotePlacement {
+  parent: string;
+  before?: string;
+  /** True when `parent` is the container to insert into (leading/first-child slot) */
+  parentIsContainer?: boolean;
+  /** The container `parent` was resolved against at creation time - disambiguates `parent` when it's a cloned node referenced from multiple containers */
+  container?: string;
```
DafyddLlyr: Helpful comments, handy to have as JSDocs 👌

apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
```diff
@@ -0,0 +1,91 @@
...
+      color
+      created_by
+      updated_by
+      created_at
+      updated_at
+    }
+  }
+`;
+
+interface FlowNoteRow {
```
DafyddLlyr: Also possible via `SnakeCasedProperties<FlowNote>` I think - we use [`type-fest`](https://github.com/sindresorhus/type-fest) in the API as a dev dependency for this and it's super nice.

This is nice as it saves on both typing out these interfaces and synchronising them 👌 

The plan is to eventually (soon-ish?) generate all these Hasura / DB sources types and be able to infer/import them 🚀 https://github.com/theopensystemslab/planx-new/blob/main/doc/architecture/decisions/0011-standardised-data-fetching-and-state-management.md#graphql-api-interactions

apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export const useFlowNotes = (): UseFlowNotesResult => {
+  const flowId = useStore((state) => state.id);
+
+  const { data, loading } = useSubscription<QueryResult>(GET_FLOW_NOTES, {
+    variables: { flowId },
+    skip: !flowId,
+  });
+
+  const notes = (data?.flow_notes ?? []).map(toFlowNote);
```
DafyddLlyr, line 109: nit: I _think_ you can skip this mapping actually by just destructing the query, e.g. `createdBy: created_by` ?

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/lib/partitionNotes.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    const noteB = makeNote({
+      placement: { parent: "container-b", before: "X" },
+    });
+    const { positioned } = partitionNotes([noteA, noteB], {});
+
+    expect(positioned.get(placementKey("container-a", "X"))).toEqual([noteA]);
+    expect(positioned.get(placementKey("container-b", "X"))).toEqual([noteB]);
+  });
+
+  it("preserves insertion order for multiple notes at the same coordinate", () => {
```
DafyddLlyr, line 100: There's no `order_by` on the `GET_FLOW_NOTES` query, how does this work in production?

I get how it's working via the `makeNote()` helper here as we're just handing it an array, but I don't think the query is giving us the same output here without an explicit request to order by `created_at`.

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/lib/notePlacement.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  const container =
+    placement.container ?? findContainerOf(flow, placement.parent);
+  if (!container) {
+    return { container: placement.parent, before: undefined };
+  }
+
+  const siblings = flow[container]?.edges ?? [];
+  const index = siblings.indexOf(placement.parent);
+  return { container, before: siblings[index + 1] };
```
DafyddLlyr, line 70: When we come back to handling delete operations, I guess we'll need to handle `indexOf()` here returning a `-1` (so it will always jump to index `0`) if the parent is missing. Not exactly sure how/if this could happen - will watch out for this in the next PR.

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/lib/notePlacement.ts
```diff
@@ -0,0 +1,64 @@
...
+ */
+export const resolveNotePlacement = (
+  flow: Graph,
+  container: string,
+  before?: string,
+): NotePlacement => {
+  const containerEdges = flow[container]?.edges ?? [];
+  const precedingId = before
+    ? containerEdges[containerEdges.indexOf(before) - 1]
+    : containerEdges[containerEdges.length - 1];
```
DafyddLlyr, line 38: It shouldn't happen, but it might be worth checking that `before` is valid - e.g. `containerEdges.includes(before)`. We're assuming here that is is always valid - if not, a placed note would jump to the top of the graph (index `0`).

apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
```diff
@@ -0,0 +1,91 @@
...
+  flowId: string;
+  nodeId: string | null;
+  placement: NotePlacement | null;
+  text: string;
+  color: string;
+  createdBy: number;
+  updatedBy: number;
+  createdAt: string;
+  updatedAt: string;
+}
```
DafyddLlyr: This doesn't really match the `CHECK` constraint we have on the DB - the better we model this I think the easier it will be to maintain. 

There's a few ways of doing this, but a type like this would correctly model that a `FlowNote` has either a `placement` or a `NodeId` (never both, never none - currently possible within this type).

```ts
interface FlowNoteBase {
  id: string;
  flowId: string;
  text: string;
  color: string;
  createdBy: number;
  updatedBy: number;
  createdAt: string;
  updatedAt: string;
}

export interface AttachedNote extends FlowNoteBase {
  nodeId: string;
  placement: null;
}

export interface PositionedNote extends FlowNoteBase {
  nodeId: null;
  placement: NotePlacement;
}

export type FlowNote = AttachedNote | PositionedNote;
```

The upside here I hope would be that we both gain clarity and strictness - a few functions may need additional guards to handle the differing behaviour of attached or positioned notes.

This type probably also negates my previous comment of GQL destructing - a `toFlowNote()` function is probably required to make TS happy!

We probably need to think of this within the context of the clone discussion we had yesterday as well.

apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
```diff
@@ -0,0 +1,91 @@
...
+  const flowId = useStore((state) => state.id);
+
+  const { data, loading } = useSubscription<QueryResult>(GET_FLOW_NOTES, {
+    variables: { flowId },
+    skip: !flowId,
+  });
+
+  const notes = (data?.flow_notes ?? []).map(toFlowNote);
+
+  return { notes, loading };
```
DafyddLlyr: Consumers probably want `error` here as well.

apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
```diff
@@ -0,0 +1,91 @@
+import { gql, useSubscription } from "@apollo/client";
+import { useStore } from "pages/FlowEditor/lib/store";
+
+export const DEFAULT_NOTE_COLOR = "#fffdb0";
```
DafyddLlyr: nit: Maybe better in `theme.ts`, within the palete?

apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export interface NotePlacement {
```
DafyddLlyr: Similar to my comment on `FlowNote` - could we consider this as a union of `SiblingAnchoredPlacement | ContainerAnchoredPlacement` (my understanding of the model here - better names most welcome).

Handling multiple states within a single type can prove quite tricky when they shouldn't ever be possible.

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/lib/notePlacement.test.ts
```diff
@@ -0,0 +1,204 @@
+import type { Graph } from "@planx/graph";
+import { describe, expect, it } from "vitest";
+
+import {
+  findContainerOf,
+  resolveNotePlacement,
+  resolveNoteRenderCoordinate,
+} from "./notePlacement";
+
+// _root -> nodeA -> nodeB -> nodeC (a plain linear root-level sequence,
```
DafyddLlyr, line 10: Very helpful 👍

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/lib/notePlacement.ts
```diff
@@ -0,0 +1,64 @@
...
+  flow: Graph,
+  nodeId: string,
+): string | undefined => {
+  for (const [id, node] of Object.entries(flow)) {
+    if (node.edges?.includes(nodeId)) {
+      return id;
+    }
+  }
+  return undefined;
+};
```
DafyddLlyr, line 18: chore: We have a number of graph traversal utils spread throughout the codebase - many in `planx-core`, some in the API, `src/@planx/graph.ts` in this repo, and elsewhere.

Very much looking forward to moving these to a shared package sometime soon 🚀 

Not a suggestion to move this - just an observation that this is a known shortcoming of the current codebase and very much on out to do list to improve.

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/lib/partitionNotes.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export interface PartitionedNotes {
+  attached: Map<string, FlowNote[]>;
+  positioned: Map<string, FlowNote[]>;
+}
+
+export const placementKey = (parent: string, before?: string): string =>
+  `${parent}::${before ?? ""}`;
+
+export const partitionNotes = (
```
DafyddLlyr, line 22: docs: Something to explain the "why" of this would be helpful here - why do we need to partition them. 

The "how" is handled very clearly in code below. Test cases cover do describe the motivation and uses very clearly, but it's probably worth an explanation here at source.

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/lib/notePlacement.test.ts
```diff
@@ -0,0 +1,204 @@
...
+      nodeA: linearFlow.nodeA,
+      folder: { type: 300, edges: [] },
+    };
+    expect(resolveNoteRenderCoordinate(flow, { parent: "folder" })).toEqual({
+      container: "_root",
+      before: undefined,
+    });
+  });
+
+  it("round-trips through encode then decode back to the original gap", () => {
```
DafyddLlyr, line 141: These "round-trip" tests are great 👍

DafyddLlyr (changes requested):
Really solid piece of work! Very solid test coverage, nice design in the hook and lib functions.

Appreciate you splitting it up - it's made it much easier for me to engage with the specifics here thanks.

There's a few comments below to review - I think the biggest win here would be improving the type design to match the actual data models.

apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
joecarver: ahh nice! i've rolled my own type helper for this kind of thing before but really nice that we've got a library type for it here. Have changed.

apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
joecarver: leaving this on account of the comment re: separate types for attached and positioned

apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
joecarver: those names make sense to me and yep definitely better to discriminate the two states with types here!

apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
joecarver: good shout! have done this and will keep it in mind for the cloning work later as well

apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
joecarver: this is not even used actually as we're not currently allowing custom colours, its just hardcoded into the note CSS. Will remove it for now.

apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
DafyddLlyr: Agreed - should have come back to delete this comment really. Thanks!

apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
DafyddLlyr: Yep spotted this actually in the UI PR - thanks!

apps/editor.planx.uk/src/hooks/data/useFlowNotes.ts
```diff
@@ -29,7 +29,10 @@ type FlowNoteRow = SnakeCasedProperties<FlowNote>;
 
 const GET_FLOW_NOTES = gql`
   subscription GetFlowNotes($flowId: uuid!) {
-    flow_notes(where: { flow_id: { _eq: $flowId } }) {
+    flow_notes(
+      where: { flow_id: { _eq: $flowId } }
+      order_by: { created_at: asc, id: asc }
```
DafyddLlyr, line 55: 👍

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/lib/notePlacement.test.ts
```diff
@@ -69,6 +69,16 @@ describe("resolveNotePlacement (encode)", () => {
...
+
+  it("throws an error when before node is not found in container", () => {
+    const flow: Graph = {
+      _root: { edges: ["folder"] },
+      folder: { type: 300, edges: ["nodeA"] },
+    };
+    expect(() => resolveNotePlacement(flow, "folder", "nodeB")).toThrow(
+      "before node nodeB not found in container folder",
+    );
+  });
```
DafyddLlyr, line 81: 👍

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/lib/partitionNotes.ts
```diff
@@ -11,6 +11,14 @@ export interface PartitionedNotes {
...
   `${parent}::${before ?? ""}`;
 
+/**
+ * Notes render at two different kinds of location, so the flow renderer needs two different lookups
+ * attached notes are keyed by the nodeId they're pinned to
+ * positioned notes sit in a gap between nodes and must be keyed by the resolved render coordinate instead (container, before)
+ *
+ * a note's stored `placement` doesn't map 1:1 to that coordinate (e.g. a cloned anchor can resolve differently depending its parent)
+ * so this resolution has to happen once, up front, rather than per-render.
+ */
```
DafyddLlyr, line 21: Great 👌

DafyddLlyr (approved):
Fab - thanks for the quick follow ups and fixes here 🚀

PR #7037, opened 2026-07-21
[chore] fix errors from pnpm check at root

freemvmt:
Was just testing out the turborepo implementation from #6981 and was getting a couple of issues on `pnpm check` - fixed here.


DafyddLlyr (approved):
Nice - thanks!

These will also be picked up here (when rebased - just one there currently), and we'll be triggering this on CI to avoid any more sneaking in 👌 https://github.com/theopensystemslab/planx-new/pull/6986/changes#diff-9370e2c0af3b5a28cdc75bde4c04b517760cced38878cc707d56e1da6c197c06

freemvmt:
@DafyddLlyr excellent, I saw that in draft and thought you might be doing something like that! 🛠️

PR #7038, opened 2026-07-21
feat: Set up UI and Storybook for PlanX in numbers widget

ianjon3s:
## What does this PR do?

Sets up UI and Storybook file for "PlanX in numbers" widget (I've called it `NumbersWidget` for now, better naming suggestions welcome)
- Adjusts styling of Explore page to match layout and style of Dashboard
- Adds placeholder search button
- Sets up `NumbersWidget` with same loading pattern as `StatsBanner`
- Sets up Storybook file with mock data

**Testing:**
- Toggle feature flag for `explore` - https://7038.planx.pizza/app/explore
- Storybook with mock data - https://storybook.7038.planx.pizza/?path=/docs/editor-components-explore-numberswidget--docs

### DIFF
--- a/apps/editor.planx.uk/src/pages/Explore/components/NumbersWidget.tsx
+++ b/apps/editor.planx.uk/src/pages/Explore/components/NumbersWidget.tsx
@@ -1,0 +1,88 @@
+import Box from "@mui/material/Box";
+import Divider from "@mui/material/Divider";
+import Typography from "@mui/material/Typography";
+import React from "react";
+
+export interface NumbersWidgetStats {
+  lpasOnPlanX: number;
+  lpasOnPlanXPrevious: number;
+  onlineServices: number;
+  onlineServicesPrevious: number;
+  totalSessions: number;
+  totalSessionsPrevious: number;
+  totalSubmissions: number;
+  totalSubmissionsPrevious: number;
+}
+
+interface NumbersWidgetProps {
+  stats?: NumbersWidgetStats;
+}
+
+function formatDelta(delta: number): string {
+  const formatted = Math.abs(delta).toLocaleString("en-GB");
+  return delta >= 0 ? `+${formatted}` : `-${formatted}`;
+}
+
+export function NumbersWidget({ stats }: NumbersWidgetProps) {
+  const rows = [
+    {
+      label: "LPAs on PlanX",
+      value: stats?.lpasOnPlanX ?? null,
+      delta: stats ? stats.lpasOnPlanX - stats.lpasOnPlanXPrevious : null,
+    },
+    {
+      label: "online services",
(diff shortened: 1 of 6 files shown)


apps/editor.planx.uk/src/pages/Explore/components/NumbersWidget.tsx
```diff
@@ -0,0 +1,88 @@
...
+  totalSubmissions: number;
+  totalSubmissionsPrevious: number;
+}
+
+interface NumbersWidgetProps {
+  stats?: NumbersWidgetStats;
+}
+
+function formatDelta(delta: number): string {
+  const formatted = Math.abs(delta).toLocaleString("en-GB");
```
DafyddLlyr: Looks like a repeat of previous function - we should share util functions like this across components for both simplicity and maintainability.

apps/editor.planx.uk/src/pages/Explore/components/NumbersWidget.tsx
```diff
@@ -0,0 +1,88 @@
...
+                <Typography variant="body2" sx={{ color: "text.secondary" }}>
+                  {label}
+                </Typography>
+                {delta !== null && (
+                  <Typography
+                    variant="body2"
+                    sx={{
+                      color: delta >= 0 ? "success.main" : "error.main",
+                      fontWeight: "bold",
+                    }}
```
DafyddLlyr, line 119: Not sure on the answer here - how should we handle `+0` situations?

This will be quite common on additional LPA counts for example.

Image

DafyddLlyr (changes requested):
Some things missing here I think - such as error or loading states.

Fetching data in the parent and passing here is a nice pattern, but we still need to consider at least a `loading` prop I think - the `Empty` Storybook example is the "error" state I guess?

apps/editor.planx.uk/src/pages/Explore/components/NumbersWidget.tsx
ianjon3s: Nice catch, I've moved to the utils file 👍

apps/editor.planx.uk/src/pages/Explore/components/NumbersWidget.tsx
ianjon3s: I think the cleanest approach is to not show a figure at all for change=0, have updated

ianjon3s:
Thanks for this @DafyddLlyr ,

I've taken the approach of creating a dedicated loading bar UI component that can be used across different editor components as needed.

This is used in both `NumbersWidget` and `StatsBanner` for now, falling back to a solid `disabled` fill when there is no data.

DafyddLlyr (approved):
Thanks for the changes 👍

PR #7039, opened 2026-07-21
[feat] expand file types accepted by API/Dropzone

freemvmt:
Here we finally enable users to upload file types beyond PNGs, PDFs and JPEGs. Apparently this is a popular feature request! After merge we will accept all files current [supported by Planning Portal](https://www.planningportal.co.uk/services/help/faq/applications/what-file-formats-are-accepted-for-online-applications) (bar the holdouts for concerns re macros, see below) - and more! See [ticket](https://trello.com/c/kkt7dKkB).

## Changes

Includes changes to the API endpoints `/file/private/upload` and `/file/public/upload` (so will affect both editors and end-users), and fits out the `Dropzone` component to accept these new files.

We make a few related cosmetic changes, like removing the file types from the subtext of that component, now that there's so many options, and improving the error messages that appear in the dialog. Like so:

image

In terms of code organisation we pull out the `ALLOWED_EXTENSIONS_BY_MIME_TYPE` source-of-truth into a common package called `file-upload`. This is our first proper package that is not config! Thanks to @DafyddLlyr for inspiring this and making the monorepo vision a reality (see discussion [below](https://github.com/theopensystemslab/planx-new/pull/7039#discussion_r3628807657)).

API-side we added some additional logic which uses the excellent [file-type](https://www.npmjs.com/package/file-type?activeTab=readme) package to inspect the content of the upload (where it's capable of doing so) and verify that it agrees with the provided extension/type. This has the additional boon of ensuring that an `.xlsx` file, for example, is not actually an `.xlsm` file (which could therefore contain potentially-malicious macros).

## Malicious macros?

On that note, macros are something we need to consider. The NCSC has a [whole article](https://www.ncsc.gov.uk/guidance/macro-security-for-microsoft-office) on the security implications of allowing macros at all (downstream). But as per the defense-in-depth heuristic, we should also take what precautions we can on our end.

This could look like, for example, using a package like [SheetJS](https://docs.sheetjs.com/docs/csf/features/vba#copying-macros) to read any `.ods` or `.xls` file and save them as `.xlsx` (which by definition does not include macros), so that any VBA code therein would be removed before we push the file to S3 (and therefore, more importantly, before it ever hits a council's server).

Alternatively we could run a Python microservice/sidecar and do something with [oletools/olevba](https://github.com/decalage2/oletools/wiki/olevba) to straight up reject any file with macros (and ask people to go away and come back with a clean file).

We do of course have Scanii, which runs every file against a database of YARA patterns, but I've nonetheless held out on enabling `.doc` and `.xls` files until we've either discussed this more or implemented a solution. Happy for people to disagree on this though, could be overzealous.

## Final commit

Had to do some fiddly stuff at the end with `pnpm build` steps to get CI and Turbo to play nicely with the new `file-upload` package (like, so that `dist/` exists when we try to run API tests etc), but I imagine that will be happily superceded by #6986 soon enough! 

## Test

Play around with file uploads on the pizza e.g. [here](https://7039.planx.pizza/testing/upload-and-label/preview). Try uploading big files, unsupported files, files with the wrong extension, etc :)

## TODO

- [ ] Add new file types to service spec on merge
- [x] Decide how to proceed for DOC/XLS files (do any councils still deal with these!?)


apps/editor.planx.uk/src/routes/$.tsx
```diff
@@ -6,6 +6,7 @@ export const Route = createFileRoute("/$")({
   component: CatchAllComponent,
 });
 
+// TODO: export from elsewhere (and import here) to avoid code-split warning
```
DafyddLlyr, line 9: Thanks! This has been on my radar as well. I have a half-baked branch which handles a bunch of the warnings on build as well as this one - will try to progress it this week.

apps/api.planx.uk/modules/file/middleware/useFileUpload.ts
```diff
@@ -9,45 +9,86 @@ import path from "path";
 const FILE_SIZE_LIMIT = 30 * 1024 * 1024;
 
 /**
- * Should match MIME type restrictions in frontend
+ * Must be kept synchronised with the ALLOWED_EXTENSIONS_BY_MIME_TYPE map in the frontend.
```
DafyddLlyr: Appreciate this is a WIP PR but just wanted to jump on here to say that this is now the perfect example of a warning comment we no longer need in our new monorepo world. We can define all these variables once in a shared package, and import into consuming apps.

Briefly, here's what this would look like if you want a crack at it (also very happy for this to be a second pass, or I'm happy to give it a shot once this is merged also) - 

 - Make a package, maybe `/packages/file-upload` (`@planx/file-upload`)
 - Setup tsconfig and build
 - Export the single sources of truth `ALLOWED_EXTENSIONS_BY_MIME_TYPE`, `ALLOWED_EXTENSIONS`, `ALLOWED_EXTENSIONS_LABELS` and `MAX_UPLOAD_SIZE_BYTES` (I guess?)
 - Import into API and Editor, use as needed
 - Satisfying delete of these "⚠️  keep these in sync ⚠️ " warnings 😄

apps/api.planx.uk/modules/file/middleware/useFileUpload.ts
```diff
@@ -9,45 +9,86 @@ import path from "path";
 const FILE_SIZE_LIMIT = 30 * 1024 * 1024;
```
DafyddLlyr: So this does not match the frontend `MAX_UPLOAD_SIZE_MB` which is `30 * 1000 * 1000` it turns out 🤦  

If it was a single instance within `@planx/file-upload` we could avoid this 😉

apps/api.planx.uk/modules/file/middleware/useFileUpload.ts
freemvmt: brave new world! I will try and bring this in tomorrow, how exciting ✨

apps/api.planx.uk/modules/file/middleware/useFileUpload.ts
freemvmt: nice catch! additional motivation hehe

apps/api.planx.uk/modules/file/middleware/useFileUpload.ts
freemvmt: would it make sense to just have a `constants` package? like, if all we're doing is exporting constants, that could probably be useful across many others cases (e.g. where things in Editor and API need to agree) and seems a bit laborious to make a new package in each case. unless there is a reason to keep separation?

DafyddLlyr: Yeah I also wondered this so did some reading up - I also thought of a `constants` or `utils` or something.

The recommended approach tends to be several smaller, more targeted, packages as a shared "miscellaneous bucket" can end up slightly abused and over-shared. I think `planx-core` has slightly become this - we have requests, types, data processing, document generation... All of this is bundled into e.g. the `scripts/encrypt` package which just needs a single util. Across the footprint of the repo this can really build up.

This does sort of feel like it's close to an edge case as it's so small and simple, but I feel we're best off building good fences early here.

Decent writeup here - https://dev.to/abdelaaziz_ouakala/the-shared-library-is-a-lie-fixing-your-nx-monorepo-architecture-3mie

DafyddLlyr: I'm also hoping that, over time, we can improve the process of setting up and making a simple package - maybe a script or template. 

The TS build process is also a small config step we need, but tooling is improving to read `.ts` files directly.

freemvmt: heard!

ianjon3s:
FYI, I have made changes to `Dropzone.tsx` in order to proceed with fixes for the accessibility audit report.

https://github.com/theopensystemslab/planx-new/pull/7140

This isn't surfacing as a merge conflict, but thought I'd point it out as your work touches this file.

.github/workflows/pull-request.yml
DafyddLlyr, line 1: I _think_ that these issues are now resolved now that https://github.com/theopensystemslab/planx-new/pull/6986 has merged - can you rebase and check?

apps/api.planx.uk/modules/file/middleware/useFileContentValidation.ts
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * Notably, this distinguishes genuine .docx/.xlsx (OOXML) from macro-enabled .docm/.xlsm by inspecting
+ * the zip package internals - so a macro-enabled file renamed to .docx/.xlsx is still caught.
+ *
+ * If file-type can't determine a type at all, we allow the upload - we only want to reject
+ * confirmed mismatches, not penalise formats we have no reliable way to verify.
+ *
+ * NB. Function could be expanded (e.g. to verify SVGs) by integrating further detectors.
+ * See: https://github.com/sindresorhus/file-type#available-third-party-file-type-detectors
+ */
```
DafyddLlyr, line 27: Great write up and solid addition to the process 👌

apps/editor.planx.uk/src/lib/api/upload.ts
DafyddLlyr, line 1: This looks like a leftover, also confirmed via `pnpm dlx knip`

apps/editor.planx.uk/package.json
```diff
@@ -19,6 +19,7 @@
     "@mui/x-data-grid": "^9.0.0",
     "@opensystemslab/map": "1.0.0-alpha.14",
     "@opensystemslab/planx-core": "catalog:",
+    "@planx/file-upload": "workspace:*",
```
DafyddLlyr, line 22: 🚀

package.json
```diff
@@ -17,6 +17,7 @@
     "prepare": "husky",
     "typecheck": "turbo run typecheck",
     "lint": "turbo run lint",
+    "lint:fix": "turbo run lint:fix",
```
DafyddLlyr, line 20: Thanks!

package.json
```diff
@@ -27,7 +28,8 @@
     "apps/tests/**/*.{js,ts}": "eslint --fix --flag v10_config_lookup_from_file",
     "e2e/tests/**/*.{js,ts}": "eslint --fix --flag v10_config_lookup_from_file",
     "apps/localplanning.services/**/*.{js,jsx,ts,tsx,mjs}": "eslint --fix --flag v10_config_lookup_from_file",
-    "{apps/editor.planx.uk/src,apps/api.planx.uk,apps/localplanning.services,e2e/tests,apps/hasura.planx.uk/tests}/**/*.{js,jsx,ts,tsx,json,css,scss,md,html}": "prettier --write"
+    "packages/file-upload/**/*.ts": "eslint --fix --flag v10_config_lookup_from_file",
+    "{apps/editor.planx.uk/src,apps/api.planx.uk,apps/localplanning.services,e2e/tests,apps/hasura.planx.uk/tests,packages/file-upload}/**/*.{js,jsx,ts,tsx,json,css,scss,md,html}": "prettier --write"
```
DafyddLlyr, line 32: Adding and managing this is going to get tedious fast - very open to ideas here..!

DafyddLlyr (commented):
Looking good..! Codebase is in a better place than you found it thanks to content-checking and other added validation - great to see. Level of tests also looks good.

On the macro / executables point it's a bit more complex I think than just blocking certain file type and stripping out macros (potentially breaking the files). I tried to do a bit of reading up here on how this is generally handled (e.g. in gmail where sending a `.xlsx` or `.doc` is fine).

These are the file types that can have executable content, we already allow PDF and SVG notably.

| File types | Examples | Status | Can carry active content? |
|---|---|---|---|
| **Photos** | JPG, PNG, plus new: BMP, GIF, TIFF, WebP | PNG/JPG live; rest new | No |
| **Vector images** | SVG | **Already live** | **Yes** - can hold scripts |
| **PDFs** | PDF | **Already live** | Some - can hold scripts |
| **Modern office** | DOCX, XLSX | New | Low - macros blocked; but Excel formulas can auto-run |
| **Text & data** | TXT, CSV | New | CSV formulas can auto-run in Excel |
| **Rich text** | RTF | New | **Yes** - can bundle other files inside |
| **CAD / design** | DWG, DXF, IFC, RVT, SKP, BIM | New | **Yes** - AutoCAD & Revit can hold macros |
| **Geographic data** | GML | New | No |
| **Video** | MP4, MOV, AVI, MKV, WebM, WMV, MPG | New | Mostly no |

I think the following approach should work - 
 - Allow all file types, regardless of active content
 - Ensure we never execute / open these files anywhere - no preview, and downloads are forced to save, not open in the browser (I think we need to cross-check this)
 - Scan via Scanii
 - Council downloads, having also ran their own scans

I think this means we need to do the following - 

- Serve files as `attachment` and set the content-type ourselves instead of trusting the uploader. The mimetype is also read directly from the uploaded file which I think is fine as we've rejected by this point. This is the line here - https://github.com/theopensystemslab/planx-new/blob/main/apps/api.planx.uk/modules/file/service/uploadFile.ts#L88 
- Check Scanii status before allowing download I guess - not just rely on it executing (see https://trello.com/c/g4q2hI9k/2498-verify-that-user-uploaded-images-have-been-checked-by-scanii-before-handing-over-to-integrations)
- Confirm nothing renders uploads inline anywhere in the Editor (thumbnails on image upload?)

I also think that from a product / legal perspective we want to double check we're clear on expecting consumers to also scan files.

apps/editor.planx.uk/src/lib/api/upload.ts
freemvmt: deleted this file 🔪

freemvmt:
PR has been rebased and last commit comments out all new file types in order to address the points made [here](https://github.com/theopensystemslab/planx-new/pull/7039#pullrequestreview-4884317780) in new PRs. See the [ticket](https://trello.com/c/kkt7dKkB) for a more detailed plan of remaining work.

DafyddLlyr (approved):
Tested pizza and working as expected - can upload jpg, but not file types commented out.

apps/editor.planx.uk/src/routes/$.tsx
DafyddLlyr: Better late than never..!

https://github.com/theopensystemslab/planx-new/pull/7302

PR #7041, opened 2026-07-21
feat: ui components for positioned and attached notes

joecarver:
**Whats in this PR?**
- Display components for positioned and attached notes
- Storybook entries and tests

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Hanger.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Hanger.tsx
@@ -12,6 +12,9 @@ import { useDragLayer, useDrop } from "react-dnd";
 
 import { useStore } from "../../../lib/store";
 import { getParentId } from "../lib/utils";
+import { useFlowNotesContext } from "../notes/FlowNotesContext";
+import { placementKey } from "../notes/lib/partitionNotes";
+import { PositionedNoteCard } from "../notes/PositionedNoteCard";
 
 interface HangerProps {
   hidden?: boolean;
@@ -33,12 +36,18 @@ const Hanger: React.FC<HangerProps> = ({ before, parent, hidden = false }) => {
     from: "/_authenticated/app/$team/$flow",
   });
 
-  const [moveNode, isTemplatedFrom, flow, orderedFlow] = useStore((state) => [
-    state.moveNode,
-    state.isTemplatedFrom,
-    state.flow,
-    state.orderedFlow,
-  ]);
+  const [moveNode, isTemplatedFrom, flow, orderedFlow, showNotes] = useStore(
+    (state) => [
+      state.moveNode,
+      state.isTemplatedFrom,
+      state.flow,
+      state.orderedFlow,
+      state.showNotes,
+    ],
+  );
+
+  const { positioned } = useFlowNotesContext();
+  const notes = positioned.get(placementKey(parent, before)) ?? [];
 
   // When working in a templated flow, if any internal portal is marked as "isTemplatedNode", then the Hanger should be visible to add children
   const indexedParent = orderedFlow?.find(({ id }) => id === parent);
@@ -97,26 +106,35 @@ const Hanger: React.FC<HangerProps> = ({ before, parent, hidden = false }) => {
   );
 
   const isHidden = hidden || hideHangerFromUser;
+  const showNoteCards = showNotes && notes.length > 0;
 
   return (
-    <li
-      className={classnames("hanger", {
-        hidden: isHidden,
-        "hanger--drop-target": isDropTargetVisible && !isHidden,
-        "hanger--drop-target-active":
-          isDropTargetVisible && !isHidden && isOver && canDrop,
-      })}
-      ref={(el) => {
-        drop(el);
-      }}
-    >
-      <button
-        onContextMenu={handleContextMenu}
-        onClick={handleHangerButtonClick}
+    <>
+      {showNoteCards && (
+        /* decoartive only - connector between parent node and note*/
+        <li className="hanger note-connector" aria-hidden="true" />
+      )}
+      {showNoteCards &&
+        notes.map((note) => <PositionedNoteCard key={note.id} note={note} />)}
+      <li
+        className={classnames("hanger", {
+          hidden: isHidden,
+          "hanger--drop-target": isDropTargetVisible && !isHidden,
+          "hanger--drop-target-active":
+            isDropTargetVisible && !isHidden && isOver && canDrop,
+        })}
+        ref={(el) => {
+          drop(el);
+        }}
       >
-        {canDrop && item && item.text}
-      </button>
-    </li>
+        <button
+          onContextMenu={handleContextMenu}
+          onClick={handleHangerButtonClick}
+        >
+          {canDrop && item && item.text}
+        </button>
+      </li>
+    </>
   );
 };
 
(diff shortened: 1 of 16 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/floweditor.scss
```diff
@@ -672,27 +759,13 @@ $fontMonospace: "Source Code Pro", monospace;
...
       &::before {
-        content: "";
-        position: absolute;
-        right: -$nodeBorderWidth;
-        bottom: -$nodeBorderWidth;
-        width: 0;
-        height: 0;
-        border-bottom: 12px solid #fff;
-        border-left: 12px solid transparent;
+        @include paperFoldFront;
```
DafyddLlyr, line 815: Nice!

apps/editor.planx.uk/src/pages/FlowEditor/floweditor.scss
```diff
@@ -59,6 +60,29 @@ $fontMonospace: "Source Code Pro", monospace;
   }
 }
 
+// Triangle shapes to simulate a paper fold on a note's corner
+@mixin paperFoldFront($size: 12px, $background: #fff) {
```
DafyddLlyr, line 65: Is the change of style here for the folder corner intentional?

**Staging**
No "second page" post-it yellow behind corner, just white/transparent
Image

**This PR**
"Second" post-it visible in yellow behind fold
Image

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/PositionedNoteCard.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+        flexDirection: "column",
+        alignItems: "center",
+        listStyle: "none",
+        padding: 0,
+        margin: 0,
+      }}
+    >
+      <li className="card decision type-Question">
+        <div className="card-wrapper">
+          {/* eslint-disable-next-line jsx-a11y/anchor-is-valid -- decorative mock of the real node markup */}
```
DafyddLlyr, line 47: 👍

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/FlowNotesContext.tsx
DafyddLlyr, line 1: Great!

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/FlowNotesContext.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const flow = useStore((state) => state.flow);
+  const { attached, positioned } = useMemo(
+    () => partitionNotes(notes, flow),
+    [notes, flow],
+  );
+
+  const value = useMemo(
+    () => ({ attached, positioned, loading }),
+    [attached, positioned, loading],
+  );
```
DafyddLlyr: (Please check this!) I'm pretty sure we can just remove these `useMemo()` calls as we're now using the React compiler.

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/PositionedNoteCard.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+          <a>
+            <span>Node A</span>
+          </a>
+        </div>
+      </li>
+      <li className="hanger">
+        <button type="button" />
+      </li>
+      <li className="hanger note-connector" aria-hidden="true" />
+      <PositionedNoteCard note={note} />
```
DafyddLlyr, line 56: I'm not sure if this is a CSS issue or a layout issue here in the Story which won't be in the graph, but there's an unexpected extra space here above the note.

Image

Image

DafyddLlyr: Looking at `Hanger.tsx` and how `note-connector` is assigned this actually just looks like a Storybook issue - we have two `<li className="hanger"/>` elements here when the graph will only render one.

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/PositionedNoteCard.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+  text: "this note is positioned in between nodes",
+  color: "#fffdb0",
+  createdBy: 1,
+  updatedBy: 1,
+  createdAt: "2026-01-01T00:00:00Z",
+  updatedAt: "2026-01-01T00:00:00Z",
+};
+
+const meta = {
+  title: "Editor Components/Notes/PositionedNoteCard",
```
DafyddLlyr, line 23: So these are the first time the Graph has made it into Storybook..! 😅 

I think `Editor Components/Graph/Notes/PositionedNoteCard` would be the right hierarchy here.

At some point we probably want a holistic think through our Storybook setup (outside components which I think is well structured) - we have these Editor components which are disperate, and then a bunch of [Atomic Design](https://bradfrost.com/blog/post/atomic-web-design/) oriented components below. I'd _probably_ vote to keep the component structure as is, and then lean into the atomic design system. The question here would be how we keep a meaningful split between public and editor setups. (cc @ianjon3s)

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Hanger.tsx
```diff
@@ -87,20 +96,32 @@ const Hanger: React.FC<HangerProps> = ({ before, parent, hidden = false }) => {
     [parent, before],
   );
 
+  const showNoteCards = showNotes && notes.length > 0;
```
DafyddLlyr, line 109: question: What does feature flagging look like for notes?

For migration of `flow.data` -> `notes` we'll (presumably?) want to toggle off the current "note" creation method (enforcing that `Question` components must have at least one `Option` child at time of creation). 

I think it's most coherent to only have one note creation/management method in place at any one given time, so we probably want to turn this "off" (hide it) until we're ready to switch.

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/AttachedNotes.test.tsx
DafyddLlyr, line 1: Super clear and meaningful 👌

DafyddLlyr (approved):
Great stuff - a few comments/questions, but no show stoppers.

apps/editor.planx.uk/src/pages/FlowEditor/floweditor.scss
joecarver: nope, fixed!

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Hanger.tsx
joecarver: good question, have been thinking about this, I _think_ the feature flag would just need to do the following:

- do not display any notes from the new table (maybe just don't even execute the query to fetch them)
- do not render any of the new entry points

and if the feature flag is set:
- do not display any notes from the old structure 
- do not allow creating a question/checklist with no answers

on the migration step, it kind of raises the question as to whether we want it to be a 'one-shot' migration i.e. old notes are deleted at the same time as we insert new notes, or whether we want to separate the creation and deletion steps. I'm guessing the latter? In which case we wouldn't actually have to handle blocking the old notes via feature flag.

joecarver: @DafyddLlyr have added feature flag in this branch and used it to conditionally render the `FlowNotesProvider` which i think should do the trick for this part of things at least! https://github.com/theopensystemslab/planx-new/pull/7041/changes/b39376b26370fc32bd92c784f61f7479d72021b0

PR #7043, opened 2026-07-21
fix: URL preview when creating new flows

zz-hh-aa:
As flagged by @ianjon3s in dev call, I missed this in #6950

Tested for:
- all methods of new flow creation
- flow rename
- new team creation

(basically everywhere `URLPrefix` is used)

I figured it was fine to use a `.replace` in `URLPrefix` but let me know if that's too brittle somehow?

### DIFF
--- a/apps/editor.planx.uk/src/ui/editor/URLPrefix.tsx
+++ b/apps/editor.planx.uk/src/ui/editor/URLPrefix.tsx
@@ -1,9 +1,19 @@
 import InputAdornment from "@mui/material/InputAdornment";
 import { typographyClasses } from "@mui/material/Typography";
+import { useParams } from "@tanstack/react-router";
 import React, { useState } from "react";
 
 type Props = {
   mode: "team" | "flow";
+};
+
+const getPath = (mode: Props["mode"], team?: string) => {
+  switch (mode) {
+    case "team":
+      return "/app";
+    case "flow":
+      return `/app/${team}`;
+  }
 };
 
 /**
@@ -11,10 +21,11 @@ type Props = {
  * Used to display URLs when creating teams and flows
  */
 export const URLPrefix: React.FC<Props> = ({ mode }) => {
+  const { team } = useParams({ strict: false });
+
   // Store in component state so this does not update when user submits form
   const [urlPrefix] = useState(() => {
-    const { origin, pathname } = window.location;
-    const path = mode === "team" ? pathname : pathname.replace("/flows", ""); // if creating a new flow, the route of the flow itself should not be a child of `/flows`
+    const path = getPath(mode, team);
     return `${origin}${path}/`;
   });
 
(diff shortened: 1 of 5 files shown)


apps/editor.planx.uk/src/ui/editor/URLPrefix.tsx
```diff
@@ -10,7 +10,7 @@ export const URLPrefix: React.FC = () => {
   // Store in component state so this does not update when user submits form
   const [urlPrefix] = useState(() => {
     const { origin, pathname } = window.location;
-    const path = pathname !== "/" ? pathname : "";
+    const path = pathname !== "/" ? pathname.replace("/flows", "") : "";
```
DafyddLlyr: This is actually a bit hard to follow (as is the original). The intention is not very clear

Why not just make this more explicit, by maybe taking a `mode: "team" | "flow"` prop and then outputting the required URL. Thinking ahead of other possible uses (adding patterns - if they have non-team URLs in future, adding schemas etc).

DafyddLlyr (approved):
Happy to approve - fixes a bug without any issues...! One small suggestion on a possible improvement / future use case.

apps/editor.planx.uk/src/ui/editor/URLPrefix.tsx
zz-hh-aa: That makes sense, thanks for the tip! Think [e054c30](https://github.com/theopensystemslab/planx-new/pull/7043/commits/e054c30e6ee4506e9be158523583534356f7a62b) suffices?

apps/editor.planx.uk/src/ui/editor/URLPrefix.tsx
```diff
@@ -2,15 +2,19 @@ import InputAdornment from "@mui/material/InputAdornment";
...
  * InputAdornment for text inputs displaying the current URL
  * Used to display URLs when creating teams and flows
  */
-export const URLPrefix: React.FC = () => {
+export const URLPrefix: React.FC<Props> = ({ mode }) => {
   // Store in component state so this does not update when user submits form
   const [urlPrefix] = useState(() => {
     const { origin, pathname } = window.location;
-    const path = pathname !== "/" ? pathname.replace("/flows", "") : "";
+    const path = mode === "team" ? pathname : pathname.replace("/flows", ""); // if creating a new flow, the route of the flow itself should not be a child of `/flows`
```
DafyddLlyr: Re-reading my comment, it wasn't that clear 😅 

I had something more like this in mind - a strong relationship between `mode` and `path`

```ts
const getPath = (mode: Props["mode"], team?: string) => {
  switch (mode) {
    case "team":
      return "/app";
    case "flow":
      return `/app/${team}`;
  }
};

/**
 * InputAdornment for text inputs displaying the current URL
 * Used to display URLs when creating teams and flows
 */
export const URLPrefix: React.FC<Props> = ({ mode }) => {
  const { team } = useParams({ strict: false });
  const { origin } = window.location;

  // Store in component state so this does not update when user submits form
  const [urlPrefix] = useState(() => {
    const path = getPath(mode, team);
    return `${origin}${path}/`;
  });
```

zz-hh-aa: That's so much clearer than what I did 😅 thanks for explaining! Updated again in [e710323](https://github.com/theopensystemslab/planx-new/pull/7043/commits/e710323ceea50af383af354badb8bf96c8257ed9)

PR #7045, opened 2026-07-22
feat: grouped submissions log

zz-hh-aa:
This PR starts the grouped submissions log feature:

**db changes**
- Creates a `submissions_grouped` view that is similar to the `submission_services_log` view _but only has one row per session_ and does not include columns such as `response`
    - This replicates a lot of the logic from `submission_services_log` but I figured it was better to avoid complicated dependencies _and_ would be better for performance to not be querying a non-materialized view
    - The aggregation logic happens here--if the most recent event for _any_ event fails, then we will show a failure; if there are no failures, show the status of the most recent event
    - We might still use `submission_services_log` for the detailed view of events (and there is a `daily_failed_submissions` dependency to keep in mind, as Jess flagged!)

**Frontend changes**
- Creates `GROUPED_SUBMISSIONS` feature flag and duplicates certain components (eg `EventsLog` and `Submissions` -> `EventsLogGrouped` and `SubmissionsGrouped`)
    -  At dev call we decided that route-level feature flagging with replicated code was probably the simplest way of feature flagging this work
- Removes 'View submission', 'Download', 'Response' and 'Resubmit' columns from submissions log (info to go into detail modal)
- Makes rows clickable
- Creates placeholder submission detail modal

Before:
Screenshot 2026-07-22 114331

After:
Screenshot 2026-07-22 114312

with modal placeholder:
Screenshot 2026-07-22 114317

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/SubmissionsGrouped.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/SubmissionsGrouped.tsx
@@ -49,6 +49,11 @@ const Submissions: React.FC<SubmissionsProps> = ({ flowSlug }) => {
   const submissions = useMemo(() => data?.submissions || [], [data]);
 
   const sessionSummaries = useMemo(() => {
+    type PartialSummary = Omit<
+      SubmissionSummary,
+      "eventCount" | "mostRecentEventType" | "mostRecentStatus" | "status"
+    >;
+
     const grouped = submissions.reduce(
       (acc, submission) => {
         if (!acc[submission.sessionId]) {
@@ -71,17 +76,26 @@ const Submissions: React.FC<SubmissionsProps> = ({ flowSlug }) => {
 
         return acc;
       },
-      {} as Record<string, SubmissionSummary>,
+      {} as Record<string, PartialSummary>,
     );
 
-    return Object.values(grouped).map((session) => ({
-      ...session,
-      eventCount: session.events.length,
-      // if ANY most recent event failed, session fails
-      sessionStatus: session.events.some((e) => e.status !== "Success")
-        ? "Failed"
-        : "Success",
-    }));
+    return Object.values(grouped).map((session) => {
+      // Find the most recent event
+      const mostRecentEvent = session.events.reduce((latest, event) =>
+        event.createdAt > latest.createdAt ? event : latest,
+      );
+
+      return {
+        ...session,
+        eventCount: session.events.length,
+        mostRecentEventType: mostRecentEvent.eventType,
+        mostRecentStatus: mostRecentEvent.status,
+        // if ANY most recent event failed, session fails
+        status: session.events.some((e) => e.status !== "Success")
+          ? ("Failed" as const)
+          : ("Success" as const),
+      };
+    });
   }, [submissions]);
 
   // filter by flow if flowId prop is passed from route params
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/SubmissionsGrouped.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/SubmissionsGrouped.tsx
@@ -8,33 +8,26 @@ import { slugify } from "utils";
 
 import { useStore } from "../../lib/store";
 import EventsLogGrouped from "./components/EventsLogGrouped";
-import type { Submission, SubmissionsProps, SubmissionSummary } from "./types";
+import type { SubmissionsProps, SubmissionSummary } from "./types";
 
 const Submissions: React.FC<SubmissionsProps> = ({ flowSlug }) => {
   const [teamId] = useStore((state) => [state.teamId]);
 
-  // submission_services_log view is already filtered for events >= Jan 1 2024
-  const { data, loading, error } = useQuery<{ submissions: Submission[] }>(
+  const { data, loading, error } = useQuery<{
+    submissions: SubmissionSummary[];
+  }>(
     gql`
       query GetSubmissions($team_id: Int!) {
-        submissions: submission_services_log(
+        submissions: submissions_grouped(
           where: { team_id: { _eq: $team_id } }
-          order_by: [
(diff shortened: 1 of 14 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/SubmissionsGrouped.tsx
```diff
@@ -0,0 +1,132 @@
...
+          ]
+          distinct_on: [session_id, event_type]
+        ) {
+          flowId: flow_id
+          sessionId: session_id
+          eventId: event_id
+          eventType: event_type
+          status: status
+          retry: retry
+          response: response
```
DafyddLlyr: Please take a look at https://trello.com/c/A77JvEt2

If we never show the response info here, we shouldn't fetch it upfront - it's a huge amount of data in some cases.

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/SubmissionsGrouped.tsx
```diff
@@ -0,0 +1,132 @@
...
+    {
+      variables: { team_id: teamId },
+      skip: !teamId,
+      pollInterval: 10_000,
+    },
+  );
+
+  const submissions = useMemo(() => data?.submissions || [], [data]);
+
+  const sessionSummaries = useMemo(() => {
```
DafyddLlyr: Rather than fetching all data, and reformatting in the frontend, it may be better to handle this on the backend with a new view.

As the number of submissions grow (see Bucks prod for an example) we will incrementally hit issues here I suspect unless we aim to improve how we fetch and handle this data.

The data we're querying here is already a view of submission data - I think it makes sense to construct what we need on the backend directly otherwise when we retire the current non-grouped submissions we will be - 
 - taking raw data from db
 - making into an view (now unused directly anywhere)
 - reformatting and grouping in the frontend

jessicamcinchak: Agree that we likely want _new_ views here (and should actually make it a bit easier to iterate on data structures while still work-in-progress) ! 

Plus a couple other quick points to keep in mind: 
- The "minimum" view which powers the table should include all session IDs currently in `submission_services_log` _but also_ every session ID which has a non-expired `payment_requests` now (eg "awaiting payment" session statuses)
  - _Then_, on row click only, we'd query full details (either `submissions_services_log` filtered on session ID or another new view?) 
- Also reminder that `submission_services_log` view has a _dependent view_ `daily_failed_submissions` - so we'll either plan to leave in place/not retire or also refactor `daily_failed_submissions` to point at new stuff !

jessicamcinchak (commented):
Great start, thanks again for picking this one up! 

Just a quick comment/reply to Daf's first review here.

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/SubmissionsGrouped.tsx
zz-hh-aa: Thanks both! Useful tips and context. 

I've made a new view and updated the plumbing to frontend accordingly. More detailed description updated in the PR, think I've addressed Jess's first point by recycling logic from the existing view? And will keep the dependent view in mind...

apps/hasura.planx.uk/migrations/default/1784812787512_created_submissions_grouped_view/up.sql
```diff
@@ -0,0 +1,149 @@
...
+      ) THEN 'Send to email' :: text
+      WHEN (
+        (se.webhook_conf) :: text ~~ '%?notify=false%' :: text
+      ) THEN 'Upload to AWS S3 (no notification)' :: text
+      WHEN (
+        (se.webhook_conf) :: text ~~ '%upload-submission%' :: text
+      ) THEN 'Upload to AWS S3' :: text
+      WHEN ((se.webhook_conf) :: text ~~ '%idox%' :: text) THEN 'Submit to Idox Nexus' :: text
+      ELSE (se.webhook_conf) :: text
+    END AS event_type,
```
jessicamcinchak, line 55: nit: per Ian's mockup, I think we can align these `event_type` names directly with what we want to display in frontend - eg "Submit to BOPS" &rarr; "Sent to BOPS" ?

apps/hasura.planx.uk/migrations/default/1784812787512_created_submissions_grouped_view/up.sql
```diff
@@ -0,0 +1,149 @@
+CREATE OR REPLACE VIEW "public"."submissions_grouped" AS WITH payments AS (
```
jessicamcinchak, line 1: This overall looks great! I think I'd expect one more CTE here like `WITH invitations_to_pay AS ( ... )`, then unioned together with payments and submissions below! 

ITP can query `payment_requests` for records created within last 28 days (eg not expired) and always default to "Success" status I think?

jessicamcinchak (approved):
New view definition looks good - a couple minor questions/nits, happy for them to be addressed here or in follow-up PRs though!

apps/hasura.planx.uk/migrations/default/1784812787512_created_submissions_grouped_view/up.sql
zz-hh-aa: Thanks for the keen eye 👁️ had a think I'm going to leave this one for now as I'm not 100% clear on how we want to display events in the table itself (and the detail view won't use `submissions_grouped`)

apps/hasura.planx.uk/migrations/default/1784812787512_created_submissions_grouped_view/up.sql
zz-hh-aa: Great shout, have done in [e236c2f](https://github.com/theopensystemslab/planx-new/pull/7045/commits/e236c2fc689e2b4714c95136401c23d655adba77)

PR #7064, opened 2026-07-24
fix: make useFormikWithRef hook consistent across components

zz-hh-aa:
Bug as flagged by [August](https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1784904959081099). I didn't check every single component type in #6938, and when going through the "Accessibility testing 2026" flow trying to I replicated the issue in FindProperty and PropertyInformation, in addition to Send. 

The issues were that these components used `useFormik` and not the `useFormikWithRef` hook.

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/Send/Editor.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/Send/Editor.tsx
@@ -4,7 +4,8 @@ import Link from "@mui/material/Link";
 import Typography from "@mui/material/Typography";
 import type { SendIntegration } from "@opensystemslab/planx-core/types";
 import { ComponentType as TYPES } from "@opensystemslab/planx-core/types";
-import { Formik, getIn, useFormikContext } from "formik";
+import { useFormikWithRef } from "@planx/components/shared/useFormikWithRef";
+import { FormikProvider, getIn } from "formik";
 import type { SubmissionEmailInput } from "pages/FlowEditor/components/Settings/Team/Integrations/SubmissionEmails/types";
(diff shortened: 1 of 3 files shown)


apps/editor.planx.uk/src/@planx/components/Send/Editor.tsx
```diff
@@ -29,187 +30,14 @@ import { parseSend, validateSchema } from "./model";
...
-  const [teamSlug, teamId, flowSlug] = useStore((state) => [
+const SendComponent: React.FC<Props> = (props) => {
+  const [teamSlug, teamId, flowSlug, id] = useStore((state) => [
     state.teamSlug,
     state.teamId,
     state.flowSlug,
+    state.id,
   ]);
 
-  const formik = useFormikContext<Send>();
```
zz-hh-aa, line 43: The most substantial / most complex changes are in Send. In multiple submission email work, [Daf suggested to split the Send](https://github.com/theopensystemslab/planx-new/pull/6318#discussion_r3044164938) component into an inner and outer component for better separation of concerns. But I don't think the nested formik instances will work with the `useFormikWithRef` hook, so they're combined into one component again here... not sure if that is satisfactory?

DafyddLlyr: Yep no issues with this - especially in the context of fixing an issue...!

Thanks for flagging and linking back to old discussion 👌

DafyddLlyr (approved):
Thanks! Merged to `main` as you're away this week, will update `#planx-bugs` 👌

PR #7074, opened 2026-07-28
feat: Add tooltip with exact time to history items

DafyddLlyr:
When debugging an issue with Lambeth's pre-app I noticed that there's no way in the Editor to get an exact date - just a relative date.

This PR adds a tooltip to display this - idea and formatting taken from GitHub. Comments most welcome, [cross-posted to `#planx-design` also](https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1785251673196849)

image

image

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/EditHistory/Timeline.tsx
@@ -11,7 +11,6 @@ import Tooltip from "@mui/material/Tooltip";
 import Typography from "@mui/material/Typography";
 import { ROOT_NODE_KEY } from "@planx/graph";
 import type { OT } from "@planx/graph/types";
-import { format, parseISO } from "date-fns";
 import type {
   CommentHistoryItem,
   HistoryItem,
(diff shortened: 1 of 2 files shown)


joecarver (approved):
good idea to add this as a tooltip! 

nit: the default sizing and positioning could potentially be improved upon a bit? to my eyes it looks a couple of font px too large and a couple of vertical px too low. feel free to disagree though!

DafyddLlyr:
Thanks @joecarver ! Updated to solve both the positioning and font size issues

image

I've also extracted this out into a reusable `<RelativeTime/>` component which can handle date formatting in a consistent manner across the app if we want to reuse it (cc @ianjon3s)

ianjon3s (approved):
Nice, good to a have a reusable component for this!

One small nit is that this seems to introduce a strange formatting for very recent operations, wheres currently the output is consistent (same operation time reads `Less than 5 seconds ago` on production.

Image

DafyddLlyr:
Good spot 🤔 

This PR doesn't change for the formatting method at all - it uses existing once. Looking at the docs the "in" comes from a time in the future - https://date-fns.org/v4.4.0/docs/formatDistanceToNow

What's happened here might be _slight_ drift between your browser clock and the Vultr server? I can't actually recreate this - tried triggering it by spamming events and it's always "less than 5 seconds ago".

If we see this on staging or prod we can take a closer look, and come up with a solution.

PR #7075, opened 2026-07-28
chore: Setup Stoke subdomain

DafyddLlyr:
Please see progress here - https://trello.com/c/u56xHbGF/3709-create-stoke-custom-subdomain

### DIFF
--- a/infrastructure/common/customDomains.ts
+++ b/infrastructure/common/customDomains.ts
@@ -133,6 +133,11 @@ export const getCustomDomains = (env: string): CustomDomain[] =>
           domain: "planningservices.kingston.gov.uk",
           cloudFrontState: "shared-final",
         },
+        {
+          name: "stoke",
+          domain: "planningservices.stoke.gov.uk",
+          cloudFrontState: "validation-only",
+        },
       ]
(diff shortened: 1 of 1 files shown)


jessicamcinchak (approved):

infrastructure/common/customDomains.ts
```diff
@@ -133,6 +133,11 @@ export const getCustomDomains = (env: string): CustomDomain[] =>
           domain: "planningservices.kingston.gov.uk",
           cloudFrontState: "shared-final",
         },
+        {
+          name: "stoke",
```
jessicamcinchak, line 137: ‼️ After thought on this one, sorry for not catching on first review !! 

As I understand, `name` here should match team slug, which in this case is still `stoke-on-trent` 🫠 

It's on my migration list to simplify to `stoke` - but we'll want to fix/align timing on this before actual cert layer gets deployed !! 

There's a few in this list (`westberks`, `stalbans`) which also do _not_ align to current team slugs so maybe not crucial, but confusing nonetheless?? Maybe those are only working because they are not actually migrated away from old method yet?

PR #7076, opened 2026-07-29
wip: migrate team slugs

jessicamcinchak:
Tracking scope of all changes here https://docs.google.com/spreadsheets/d/1Jlb7HVNDphfNOu-A44dxAZWQf8Z-Ni6pyD-puZazcJ0/edit?gid=0#gid=0

Coordinated PRs / follow-ups coming soon:
- `planx-data-migration` script to find + replace all updated `articleFour.{slug}.something` values across each flow + most recent published version (and "active" `lowcal_sessions` breadcrumbs + passports depending on decision to touch existing "live" teams?) https://github.com/theopensystemslab/planx-data-migrations/pull/21
- deploy certifcates layer because custom subdomain changes (I think?)
- update UI and mutation for "add a new team" modal to allow unique name _and_ slug inputs, with slug still defaulting to `slugify(name)`


apps/hasura.planx.uk/migrations/default/1785287957182_align_team_slugs/up.sql
```diff
@@ -0,0 +1,77 @@
...
+SET name = 'Stockport',
+slug = 'stockport'
+WHERE name = 'Stockport Metropolitan';
+
+UPDATE teams
+SET name = 'Stoke-on-Trent',
+slug = 'stoke'
+WHERE name = 'Stoke on Trent';
+
+-- Teams with updated slug only, where is "old" slug
```
jessicamcinchak, line 34: This category of changes is arguably a lot more tedious and immediate benefits feel less clear? (When updating name _and_ slug, it feels like very clear benefit of correct user-facing council name.) 

Slug-only is also going to be the easiest to continue to get wrong when onboarding new teams as "name" can intentionally and regularly drift from "slug" here (eg `South Gloucester`, `southglos`). 

Both will need to be uniquely input/edited in "add a team modal" which is not currently possible:
Image

Are we absolutely sure that we want to align PlanX team workspace slugs to council web domains? Alternatively, do we just want to fix "new" worst offenders and not touch existing teams?

apps/api.planx.uk/modules/gis/service/local_authorities/metadata/bristol.ts
```diff
@@ -0,0 +1,32 @@
...
+      "articleFour.bristol.cliftonWest": "A4D12",
+      "articleFour.bristol.lawrenceAshleyCabotCotham": "A4D13",
+      "articleFour.bristol.groveWood": "A4D14",
+      "articleFour.bristol.cothamRedland": "A4D15",
+      "articleFour.bristol.henleaze": "A4D16",
+    },
+  },
+};
+
+export { planningConstraints };
```
jessicamcinchak, line 32: Diffs are a little strange here sorry! like Liverpool, this is actually a _renamed_ file and the only line-level changes are renaming keys from `articleFour.bristolCity...` &rarr; `articleFour.bristol...`

jessicamcinchak:
Closing in favor of #7114 - have narrowed scope there to ensure we do NOT touch existing teams with LIVE services, as those would require additional migration step for active sessions plus 'downloadable' submissions from last 28 days which is unnecessarily risky.

PR #7077, opened 2026-07-29
fix: Tidy up editor secondary buttons

ianjon3s:
## What does this PR do?

Quick one:
- Use correct `contained` / `secondary` button style (subtle border) for back buttons in publish modal
- Remove a few obsolete background colour properties


joecarver (approved):
all good!

PR #7079, opened 2026-07-29
feat: make notes clonable [backend]

joecarver:
DB schema changes for making notes cloneable, using the discussed approach:

- separate out `flow_notes` table into separate tables `flow_note_content` and `flow_note_positions` - where a cloned note has a single row in `flow_note_positions` and multiple rows in `flow_note_positions`, one per cloned instance
- up and down migrations include backfilling / converting data between the two data structures. obviously the down migration is lossy as it breaks the clone linking but hopefully this shouldn't come up
- graphql metadata and instrospection tests


apps/hasura.planx.uk/migrations/default/1784708519240_split_flow_notes_content_and_positions/up.sql
```diff
@@ -0,0 +1,81 @@
...
+RETURNS TRIGGER AS $$
+BEGIN
+  DELETE FROM "public"."flow_note_content"
+  WHERE "id" = OLD."note_id"
+  AND NOT EXISTS (
+    SELECT 1 FROM "public"."flow_note_positions" WHERE "note_id" = OLD."note_id"
+  );
+  RETURN OLD;
+END;
+$$ LANGUAGE plpgsql;
```
DafyddLlyr, line 66: 👍

DafyddLlyr (approved):

PR #7082, opened 2026-07-29
feat: copy notes when a flow is copied

joecarver:
Adds missing feature behaviour to copy notes for a flow when the flow is copied. Handles the case for cloned notes by creating a single new flow_note_content row and multiple flow_note_position rows.

Copying flows is an api-implemented feature so we had to redeclare types here that had already been declared in the editor. I'm guessing we will be creating more shared types soon now that monorepo is almost ready, so I left this for now.

### DIFF
--- a/apps/api.planx.uk/helpers.ts
+++ b/apps/api.planx.uk/helpers.ts
@@ -421,19 +421,34 @@ const getChildren = (
 };
 
 /**
+ * Rename a single node id the same way `makeUniqueFlow` renames the flow it belongs to
+ * (replace last n characters), leaving `_root` and nullish ids untouched.
+ */
+function renameNodeId(nodeId: string, replaceValue: string): string;
+function renameNodeId(
+  nodeId: string | null | undefined,
+  replaceValue: string,
+): string | null | undefined;
+function renameNodeId(
+  nodeId: string | null | undefined,
+  replaceValue: string,
+): string | null | undefined {
+  if (!nodeId || nodeId === "_root") return nodeId;
+  return nodeId.slice(0, -replaceValue.length) + replaceValue;
+}
+
+/**
  * For a given flow, make it unique by renaming its' node ids (replace last n characters) while preserving its' content
  */
 const makeUniqueFlow = (
   flowData: Flow["data"],
   replaceValue: string,
 ): Flow["data"] => {
-  const charactersToReplace = replaceValue.length;
-
   Object.keys(flowData).forEach((node) => {
     // if this node has edges, rename them (includes _root.edges)
     if (flowData[node]["edges"]) {
-      const newEdges = flowData[node]["edges"]?.map(
-        (edge) => edge.slice(0, -charactersToReplace) + replaceValue,
+      const newEdges = flowData[node]["edges"]?.map((edge) =>
+        renameNodeId(edge, replaceValue),
       );
       delete flowData[node]["edges"];
       flowData[node]["edges"] = newEdges;
@@ -441,7 +456,7 @@ const makeUniqueFlow = (
 
     // rename this top-level node if it's not _root
     if (node !== "_root") {
-      const newNodeId = node.slice(0, -charactersToReplace) + replaceValue;
+      const newNodeId = renameNodeId(node, replaceValue);
       flowData[newNodeId] = flowData[node];
       delete flowData[node];
     }
@@ -478,4 +493,5 @@ export {
   getMostRecentPublishedFlowVersion,
   isLiveEnv,
   makeUniqueFlow,
+  renameNodeId,
 };
--- a/apps/api.planx.uk/modules/flows/copyFlow/copyFlowNotes.ts
+++ b/apps/api.planx.uk/modules/flows/copyFlow/copyFlowNotes.ts
@@ -1,0 +1,159 @@
+import { gql } from "graphql-request";
+
+import { getClient } from "../../../client/index.js";
+import { renameNodeId } from "../../../helpers.js";
+import type { Flow } from "../../../types.js";
+import { userContext } from "../../auth/middleware.js";
+
+interface NotePlacement {
+  parent: string;
+  container?: string;
+  before?: string;
+  parentIsContainer?: boolean;
+}
+
+interface FlowNotePositionRow {
+  note_id: string;
+  node_id: string | null;
+  placement: NotePlacement | null;
+  note: {
+    text: string;
+    color: string;
+  };
+}
+
+const GET_FLOW_NOTES = gql`
+  query GetFlowNotesForCopy($flowId: uuid!) {
+    flow_note_positions(where: { flow_id: { _eq: $flowId } }) {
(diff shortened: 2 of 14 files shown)


apps/api.planx.uk/modules/flows/copyFlow/copyFlowNotes.ts
```diff
@@ -0,0 +1,159 @@
...
+  placement: placement ? remapPlacement(placement, replaceValue) : null,
+});
+
+/**
+ * Copy a flow's notes to a newly-copied flow, remapping nodeIds/placements to match the destination flow's renamed nodes.
+ * editing a note in the new flow instance will not affect the original flow's note
+ *
+ * notes that were clones of one another in the source flow are copied as clones of one another in the destination flow too
+ * - they are just no longer clones of the original notes
+ */
```
DafyddLlyr, line 87: 👍

apps/api.planx.uk/modules/flows/copyFlow/copyFlowNotes.ts
```diff
@@ -0,0 +1,159 @@
...
+  const { client: $client } = getClient();
+
+  const { flow_note_positions } = await $client.request<{
+    flow_note_positions: FlowNotePositionRow[];
+  }>(GET_FLOW_NOTES, { flowId: sourceFlowId });
+
+  if (!flow_note_positions.length) return;
+
+  const groupsByOriginalNoteId = new Map<string, FlowNotePositionRow[]>();
+  for (const row of flow_note_positions) {
```
DafyddLlyr: Where ever possible we aim to keep snake_case out of TS code - we should be able to map via GQL to camelCase.

apps/api.planx.uk/modules/flows/copyFlow/service.ts
```diff
@@ -28,6 +29,13 @@ const copyFlow = async ({
       isService: flow.isService,
       isPattern: flow.isPattern,
     });
+
+    // Copy this flow's notes too, as new (not cloned) notes pointing at the new nodeIds
+    await copyFlowNotes({
+      sourceFlowId: flowId,
+      newFlowId: insertedFlow.id,
+      replaceValue,
+    });
```
DafyddLlyr, line 38: How should be handle rollbacks here? If we insert a flow, but there's a failure on notes, we'll have what looks to the user like a partially copied flow (missing notes).

apps/api.planx.uk/helpers.ts
```diff
@@ -420,28 +420,43 @@ const getChildren = (
...
+ */
+function renameNodeId(nodeId: string, replaceValue: string): string;
+function renameNodeId(
+  nodeId: string | null | undefined,
+  replaceValue: string,
+): string | null | undefined;
+function renameNodeId(
+  nodeId: string | null | undefined,
+  replaceValue: string,
+): string | null | undefined {
```
DafyddLlyr, line 435: Nice use of overloads - I don't use these often but it's a nice solution here 👌

apps/api.planx.uk/modules/flows/copyFlow/service.ts
DafyddLlyr: One way could be to write this as psql ([exposed as a custom function in Hasura](https://hasura.io/docs/2.0/schema/postgres/postgres-guides/functions/)) which we call within the same mutation as create flow. See `diff_latest_published_flow` as an example of this approach.

This means we get a single transaction, one Hasura query instead of 1 flow insert + N note inserts.

The downside is you have to write (and test) psql - we could use the e2e/api harness for this though.

apps/api.planx.uk/modules/flows/copyFlow/copyFlowNotes.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  // One new flow_note_content row per unique original note, created via the first position that used it
+  const newNoteIdByOriginalNoteId = new Map<string, string>();
+  await Promise.all(
+    Array.from(groupsByOriginalNoteId.entries()).map(
+      async ([originalNoteId, [first]]) => {
+        const { insert_flow_note_positions_one } = await $client.request<{
+          insert_flow_note_positions_one: { note_id: string };
+        }>(INSERT_FLOW_NOTE_POSITION_WITH_CONTENT, {
```
DafyddLlyr, line 122: Same as above re:transactions - we sort of have two potential issues here - 

1. Create flow succeeds, failure / error before create notes calls - flow without any notes
2. Create flow succeeds, create notes fails part way through - flow with some notes

DafyddLlyr (changes requested):
Making progress on the reviews...! 😅 

Once again - very happy with tests and code quality - very clean, clear and well structured.

A few minor comments below, the real area of interest is the question around a single transaction here - something to take a look at.

apps/api.planx.uk/modules/flows/copyFlow/copyFlowNotes.ts
joecarver: https://github.com/theopensystemslab/planx-new/pull/7082/commits/cefe8a40551ef485ddf8bac837842d9e82ad8615

apps/api.planx.uk/modules/flows/copyFlow/service.ts
joecarver: This is a good spot - just got a couple questions to clarify here:

- did a search for `diff_latest_published_flow` and this just looks like SELECT functionality? A Postgres FUNCTION that we call from the gql query `GetFlowDiff`? So not sure which part of this we should be imitating, do you just mean to wrap our inserts in a Postgres function? That seems like a good idea if so.
- Correct me if I'm wrong but this is an existing issue with copying flows? And actually with anywhere we call `createFlow`? This function appears to:
    - insert the flow
    - insert the operations
    - publish the flow
in three separate queries/requests. There's a catch statement round the whole block but it doesn't roll anything back.

Could be misreading but I think that looks like a similar issue with an even wider scope. So could be worth a separate fix.

DafyddLlyr: 1. Yep - `diff_latest_published_flow ` was intended just as as example of how we can expose custom functions via gql. This is maybe a better docs example as it covers mutations - https://hasura.io/docs/2.0/schema/postgres/custom-functions/

Wrapping the inserts in a function _I think_ will be needed as we'll be iterating over variables - I don't think this is possible in a plain mutation.

2. Absolutely a known issue - should have been clear here 😅 A pragmatic way forwards here is maybe a noisy failure and we can fix forwards if/when this appears as a problem? The difference here I think is that a missing published flow is a relatively simple manual fix (and easy to spot) if anything goes wrong. Updating notes based on a partial failure or complete failure would be a more difficult issue to solve.

DafyddLlyr (approved):
The transaction/rollback issue is pre-existing but we're building on this here. Something I'm happy for us to fix forwards 👌

PR #7083, opened 2026-07-29
feat: add drag-n-drop support for notes

joecarver:
Fairly straightforward change, hooking into the existing drag and drop implementation but enabling this behaviour for positioned notes, and handling this case in the Hanger. Attached notes do not need this support, as they are moved by default along with the node they are attached to.

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Hanger.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Hanger.tsx
@@ -13,6 +13,7 @@ import { useDragLayer, useDrop } from "react-dnd";
 import { useStore } from "../../../lib/store";
 import { getParentId } from "../lib/utils";
 import { useFlowNotesContext } from "../notes/FlowNotesContext";
+import { useMoveFlowNotePosition } from "../notes/hooks/useMoveFlowNotePosition";
 import { placementKey } from "../notes/lib/partitionNotes";
 import { PositionedNoteCard } from "../notes/PositionedNoteCard";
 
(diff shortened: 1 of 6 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Hanger.tsx
```diff
@@ -71,8 +78,13 @@ const Hanger: React.FC<HangerProps> = ({ before, parent, hidden = false }) => {
 
   const [{ isOver, canDrop, item }, drop] = useDrop({
     accept: DROPPABLE_TYPES,
-    drop: (item: Item) => {
-      moveNode(item.id, item.parent, before, parent);
+    drop: (item: Item | NoteItem, monitor) => {
+      if (monitor.getItemType() === "NOTE") {
+        moveFlowNotePosition((item as NoteItem).positionId, parent, before);
+      } else {
+        const nodeItem = item as Item;
```
DafyddLlyr, line 85: suggestion: A type predicate can help us out here!

We can wrap the `monitor.getItemType()` call in a type predicate (`is` keyword) to avoid casting and protect ourselves in future if `DROPPABLE_TYPES` grows - 

```ts
  const isNoteItem = (monitor: DropTargetMonitor<Item | NoteItem, unknown>, _item: Item | NoteItem): _item is NoteItem => monitor.getItemType() === "NOTE"
```

When called TS will then correctly infer the type 👍

DafyddLlyr (approved):
Nice and simple..! Approving from a code POV, I can't get the Pizza to load to test out manually. I'll circle back to this one and manually test once the stack has shrunk and the Pizza is up.

Any issues here can be ironed out later and fixed forwards I'm sure 👍

PR #7085, opened 2026-07-29
feat(editor): Additional warning step on rename flow action

DafyddLlyr:
## What does this PR do?
Adds an additional warning step when updating a flow slug. Suggestions on copy most welcome..!

Trello ticket - https://trello.com/c/iGugBHpq/3690-add-additional-warning-verification-step-on-flow-name-change

https://github.com/user-attachments/assets/29c6b705-9245-4b6e-beaf-0c0743425191

### DIFF
--- a/apps/editor.planx.uk/src/pages/Flows/components/RenameDialog.tsx
+++ b/apps/editor.planx.uk/src/pages/Flows/components/RenameDialog.tsx
@@ -46,6 +46,39 @@ const validationSchema = object().shape({
   flowSlug: string().trim().required("Slug is required"),
 });
 
+const RenameWarning: React.FC = () => (
+  <ErrorSummary format="warning" heading="Renaming will break existing links">
+    <Typography variant="body2" sx={{ mt: 2 }}>
+      If applicable:
+      <List>
+        <ListItem>
+          <CheckCircleIcon color="success" fontSize="small" sx={{ mr: 1 }} />
+          Any nested flow references will update automatically
+        </ListItem>
+        <ListItem>
+          <CheckCircleIcon color="success" fontSize="small" sx={{ mr: 1 }} />
+          Local Planning Services (LPS) listing will update automatically
+        </ListItem>
+        <ListItem>
+          <WarningAmber color="warning" fontSize="small" sx={{ mr: 1 }} />
+          Any links to this service on your council website will need to be
+          updated by your IT team
+        </ListItem>
+        <ListItem>
+          <WarningAmber color="warning" fontSize="small" sx={{ mr: 1 }} />
+          Any Next Steps components pointing to this service will need to be
+          updated and published
+        </ListItem>
+        <ListItem>
+          <WarningAmber color="warning" fontSize="small" sx={{ mr: 1 }} />
+          Any users with active magic links or bookmarks will get an error if
+          this is a public-facing service and need to be manually redirected.
+        </ListItem>
+      </List>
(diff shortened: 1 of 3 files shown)


apps/editor.planx.uk/src/pages/Flows/components/RenameDialog.tsx
```diff
@@ -134,58 +150,92 @@ export const RenameDialog: React.FC<Props> = ({
...
-              </DialogActions>
+              {isConfirming ? (
+                <>
+                  <DialogContent
+                    sx={{ gap: 2, display: "flex", flexDirection: "column" }}
+                  >
+                    <ErrorSummary
+                      format="warning"
+                      heading="Renaming will break existing links"
+                      message={`This will change the service's URL. Anyone using a link or bookmark to the old URL will get an error. Coordinate with your web team to update links before continuing.`}
```
jessicamcinchak: nit: maybe this covers more cases impacted by name/slug change? (also aware we don't use flow/service so interchangeably anymore and trying to be more specific/consistent!)
> Renaming will change the URL. If applicable:
>  - ✔️ Any nested flow references will update automatically 
>  - :heavy_check_mark:  Local Planning Services (LPS) listing will update automatically
> - :warning: Any Next Steps components pointing to this service will need to be updated and published
> - ⚠️ Any users with active magic links or bookmarks will get an error
>
> If this is a public-facing service, please coordinate with your IT team to update links on your website before continuing.

Coordinating with web team solves new users _finding_ the service still (great if guidance only), but not broken S&R or Pay magic links for submission services - is there another suggestion we can offer there? 
Should we query/display number of sessions which could be resumed or have outstanding invitations to pay for further context?

DafyddLlyr: Thanks - this copy is much better and more comprehensive.

The more I think about it, the gnarlier changing flow slug becomes - implementing some sort of lookup from old slug -> new slug solves many problems (mentioned within the context of redirects here https://trello.com/c/ydxI7WEG/3691-configure-redirects-if-a-service-name-url-changes). Solving as a lookup seems neater than a big find+replace across flows.

I'd say for now we...
 - Add a simple warning, with updated copy
 - Don't add open session counts
 - Focus on solving the problem by tracking changes to `flows.slug`

DafyddLlyr: Please see [b758121](https://github.com/theopensystemslab/planx-new/pull/7085/commits/b7581218ea809e2a4459d5fa0bf37736eee981d0)

Icons taken from `<ValidationChecks/>` (cc @ianjon3s)

image

jessicamcinchak: Agree with way forward here! 

Nit picky copy suggestion: in the last point, I think it currently suggests that _if_ you coordinate with your IT team to update links _then_ that will fix/negate broken magic links ? This is not the case! Shall we just chop off everything from ", please ..." ? Or perhaps "...will get an error and need to be manually re-directed" ? (to indicate please expect folks to reach out with issues?)

Maybe better phrasing for IT team bit is a 5th bullet / warning: "Any links to this service on your council website will need to be updated by your IT team" (and placed _after_ LPS point since related/similar?)

jessicamcinchak (approved):
One final copy nit, but technically looks great !

PR #7086, opened 2026-07-30
feat(editor): Download service charge CSV

DafyddLlyr:
## What does this PR do?
 - Wires up `.csv` download from service charges page
 - Adds Storybook coverage

Storybook demo here - https://storybook.7086.planx.pizza/?path=/docs/editor-components-subscription-servicecharges--docs

Trello ticket - https://trello.com/c/FJjck1KD/3680-fix-download-service-charge-payment-records-csv-link-in-editor

### DIFF
--- a/apps/api.planx.uk/package.json
+++ b/apps/api.planx.uk/package.json
@@ -25,7 +25,7 @@
     "cookie-session": "^2.1.1",
     "copyfiles": "^2.4.1",
     "cors": "^2.8.6",
-    "csv-stringify": "^6.8.0",
+    "csv-stringify": "catalog:",
     "date-fns": "^4.4.0",
     "dompurify": "catalog:",
     "express": "^4.22.2",
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Subscription/components/ServiceCharges.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Subscription/components/ServiceCharges.tsx
@@ -22,6 +22,7 @@ import Caret from "ui/icons/Caret";
 
 import type { SubscriptionProps } from "../types";
 import {
+  downloadServiceChargesCsv,
   formatUKFiscalYear,
   getUKFiscalYear,
   getUKFiscalYearQuarter,
@@ -96,7 +97,15 @@ const ActiveServiceCharges = ({ serviceCharges }: SubscriptionProps) => {
         />
       </Box>
       <Box sx={{ mt: 2, textAlign: "right" }}>
-        <Link component="button" onClick={() => console.log("todo")}>
+        <Link
+          component="button"
+          onClick={() =>
+            downloadServiceChargesCsv(
+              serviceChargesInActiveYear,
+              `planx-service-charges-${activeYear}.csv`,
+            )
+          }
+        >
           <Typography variant="body2">
             {"Download service charge payment records (.csv)"}
           </Typography>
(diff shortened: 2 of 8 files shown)


apps/api.planx.uk/package.json
```diff
@@ -25,7 +25,7 @@
     "cookie-session": "^2.1.1",
     "copyfiles": "^2.4.1",
     "cors": "^2.8.6",
-    "csv-stringify": "^6.8.0",
```
joecarver, line 28: nice! not seen this before but seems handy

apps/editor.planx.uk/src/pages/FlowEditor/components/Subscription/utils.ts
```diff
@@ -35,3 +36,31 @@ export const formatUKFiscalYear = (fiscalYear: number): string => {
...
+  serviceCharges: ServiceCharge[],
+): string =>
+  stringify(serviceCharges, {
+    header: true,
+    columns: [
+      { key: "flowName", header: "Flow name" },
+      { key: "sessionId", header: "Session ID" },
+      { key: "paymentId", header: "Payment ID" },
+      { key: "amount", header: "Amount (excl VAT)" },
+      { key: "paidAt", header: "Paid at" },
```
joecarver, line 50: nit: only potential issue i can think of is that some of these values could intefer with csv parsing if they contained commas or even more special characters that are interpreted as formulas/rules in excel? i guess it would only be flowName as that's the only customisable value, but seems likely it could contain a comma in some cases at least? so might be worth sanitising it here.

apps/editor.planx.uk/src/pages/FlowEditor/components/Subscription/components/ServiceCharges.tsx
```diff
@@ -96,7 +97,15 @@ const ActiveServiceCharges = ({ serviceCharges }: SubscriptionProps) => {
...
-        <Link component="button" onClick={() => console.log("todo")}>
+        <Link
+          component="button"
+          onClick={() =>
+            downloadServiceChargesCsv(
+              serviceChargesInActiveYear,
+              `planx-service-charges-${activeYear}.csv`,
+            )
+          }
+        >
```
joecarver, line 109: nit: do we want to style this like an actual button? or are we happy just with the plain text style link?

joecarver (approved):
Looks good and working well! couple of v minor suggestions but nothing blocking.

Also just occurred to me that maybe we want to add service_charges to the sync script so we have data to test this on pizza and local?

apps/editor.planx.uk/src/pages/FlowEditor/components/Subscription/utils.ts
DafyddLlyr: I've been caught out by this before too many times 😅 This is why I went for `csv-stringify` which handles escape characters etc instead of a hand-rolled implementation.

I'll add test coverage here though for this - if we swap or remove the library but don't handle this edge case we want something to break.

DafyddLlyr: Please see [c06870f](https://github.com/theopensystemslab/planx-new/pull/7086/commits/c06870f69ad93ddadb76d3ac9e4aafe120cb4dfa)

apps/editor.planx.uk/src/pages/FlowEditor/components/Subscription/components/ServiceCharges.tsx
DafyddLlyr: Good catch, please see [5055455](https://github.com/theopensystemslab/planx-new/pull/7086/commits/5055455ded21153cd9b703431dae4b9a896ec0a3)

image

jessicamcinchak (commented):

PR #7091, opened 2026-07-30
chore(editor): Refactor AddComponent in prep for new Patterns tab

DafyddLlyr:
## What does this PR do?
This PR tidies up the `AddComponentModal` before the upcoming patterns tab work, so the new feature complexity has a clear structure as opposed to just adding to what's here.

There's no behavioural changes here, just structural refactors - the popover is all working as before.

## Structure
- `AddComponentModal/index.tsx` - main popover wrapper - handles sizing, positioning, and tab management (soon!)
- `AddComponentModal/ComponentsTab/` - the searchable component list, child row component, and data (`componentData.ts`)
- `FormModal/` - also grouped into a folder, with the `ChangeComponentHeader` grouped here


apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/ComponentsTab/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const filteredCategories = useMemo<Category[]>(() => {
+    if (!searchedItems) return ALL_CATEGORIES;
+    const visibleSlugs = new Set(searchedItems.map((item) => item.slug));
+    return ALL_CATEGORIES.map((cat) => ({
+      ...cat,
+      items: cat.items.filter((item) => visibleSlugs.has(item.slug)),
+    })).filter((cat) => cat.items.length > 0);
+  }, [searchedItems]);
+
+  // TODO: useRef instead of getElementById
```
DafyddLlyr, line 46: Intentional `TODO` left here - will take another look at this when I handle search in the Patterns component.

jessicamcinchak (approved):
Great ! 

One minor question / maybe regression: I notice that as a `platformAdmin`, I can _see_ hangers on templated flows  but _not_ click on them to open the component modal (specifically in the "root" of the templated flow, templated folders are okay. This is same on staging, prod right now which suggests this PR isn't culprit, but another recent component modal or even notes one might be?). 

If my memory is right here, `platformAdmins` _should_ be able to change structure of any flow, templated or not (else why not hide hanger all together when templated?) ?? 

Raising here because hopefully quick to fix and not-not related to modal !

DafyddLlyr:
> _One minor question / maybe regression: I notice that as a platformAdmin, I can see hangers on templated flows but not click on them to open the component modal (specifically in the "root" of the templated flow, templated folders are okay. This is same on staging, prod right now which suggests this PR isn't culprit, but another recent component modal or even notes one might be?)._
>
> _If my memory is right here, platformAdmins should be able to change structure of any flow, templated or not (else why not hide hanger all together when templated?) ??_
> 
> _Raising here because hopefully quick to fix and not-not related to modal !_

Let me take a look at this one - thanks for catching..! I'll fix in a clean PR so as to not hold this up 👍

PR #7000, opened 2026-07-10
prototype: Show patterns select in AddComponentModal

ianjon3s:
## What does this PR do?

Working prototype for building and inserting patterns:
- Patterns are created in `Patterns` team
- Patterns tab added to `AddComponentModal`
  - Lists Patterns created
  - Attempts to render the graph as a simplified graphic
  - Allows insertion into a flow

PR #7092, opened 2026-07-30
feat(editor): Initial add pattern UI

DafyddLlyr:
## What does this PR do?
This PR adds the Patterns tab to the `AddComponentModal`, behind the `PATTERN_SELECT` feature flag. The flag is intended to be very short lived, just covering this feature, but it means we can easily merge discrete PRs onto `main` / `production`.

This PR is strictly UI only. Patterns are static mock data, and clicking "Insert" does nothing yet.

UI is largely pulled from https://github.com/theopensystemslab/planx-new/pull/7000 (thanks @ianjon3s !), but with a few notable exclusions - 
 - No "preview" in sidepanel (see comment on `PatternDetailPanel.tsx`)
 - No component count (see comment on `pattern.data`)

https://github.com/user-attachments/assets/62be0e7f-228c-4f8a-a336-054996cfddc1

## Testing
 - Please try this out on Storybook, link here - https://storybook.7092.planx.pizza/?path=/docs/editor-components-modal-addcomponentmodal--docs

## Next steps...
 - [ ] Create queries to populate real flows https://github.com/theopensystemslab/planx-new/pull/7095
 - [ ] Tests
 - [ ] Insert patterns into flows
 - [ ] Fix search focus across tabs

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/ComponentsTab/ComponentRow.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/ComponentsTab/ComponentRow.tsx
@@ -3,7 +3,7 @@ import Tooltip from "@mui/material/Tooltip";
 import Typography from "@mui/material/Typography";
 import { ICONS } from "@planx/components/shared/icons";
 import React, { useEffect, useState } from "react";
-import { FONT_WEIGHT_SEMI_BOLD } from "theme";
+import { focusStyle, FONT_WEIGHT_SEMI_BOLD } from "theme";
 import { AiChip } from "ui/editor/AiChip";
 
 import type { ComponentItem } from "./componentData";
@@ -41,6 +41,8 @@ export const ComponentRow: React.FC<Props> = ({
       slotProps={{ tooltip: { sx: { maxWidth: 240 } } }}
     >
       <Box
+        component="button"
+        type="button"
         onClick={onClick}
         data-component-type={item.type}
         sx={{
@@ -49,11 +51,18 @@ export const ComponentRow: React.FC<Props> = ({
           gap: 1,
           px: 1.5,
           py: 0.875,
+          width: "100%",
+          border: 0,
+          backgroundColor: "transparent",
+          font: "inherit",
+          color: "inherit",
+          textAlign: "left",
           cursor: "pointer",
           "&:hover": {
             backgroundColor: "action.hover",
             "& .component-title": { fontWeight: FONT_WEIGHT_SEMI_BOLD },
           },
+          "&:focus-visible": focusStyle,
         }}
       >
         {Icon && (
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/PatternDetailPanel.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/PatternDetailPanel.tsx
@@ -1,0 +1,71 @@
+import CloseIcon from "@mui/icons-material/Close";
+import Box from "@mui/material/Box";
+import Button from "@mui/material/Button";
+import IconButton from "@mui/material/IconButton";
+import Typography from "@mui/material/Typography";
+import React from "react";
+import { FONT_WEIGHT_SEMI_BOLD } from "theme";
+
+import type { Pattern } from ".";
+
+export const DETAIL_PANEL_WIDTH = 300;
+
+interface Props {
+  pattern: Pattern | null;
+  onInsert: (patternId: string) => void;
+  onClear: () => void;
+}
+
+export const PatternDetailPanel: React.FC<Props> = ({
+  pattern,
+  onInsert,
+  onClear,
+}) => (
+  <Box
+    sx={{
+      width: DETAIL_PANEL_WIDTH,
+      flexShrink: 0,
+      overflowY: "auto",
+      p: 2,
+      display: "flex",
+      flexDirection: "column",
+      gap: 1.5,
+    }}
+  >
+    {pattern ? (
+      <>
+        <Box sx={{ display: "flex", alignItems: "flex-start", gap: 1 }}>
+          <Typography
+            variant="body1"
+            sx={{ fontWeight: FONT_WEIGHT_SEMI_BOLD, flexGrow: 1 }}
+          >
+            {pattern.name}
+          </Typography>
+          <IconButton
+            onClick={onClear}
+            aria-label={`Close ${pattern.name} details`}
+            size="small"
+            sx={{ mt: -0.5, mr: -0.5 }}
+          >
+            <CloseIcon fontSize="small" />
+          </IconButton>
+        </Box>
+        {pattern.summary && (
+          <Typography variant="body2">{pattern.summary}</Typography>
+        )}
+        <Button
+          variant="contained"
+          size="small"
+          // disabled={!pattern.data}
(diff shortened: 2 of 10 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  name: string;
+  summary: string | null;
+  // data: Graph | null;
+}
+
+const summary =
+  "Lorem, ipsum dolor sit amet consectetur adipisicing elit. Praesentium voluptatum excepturi at rem enim distinctio officia";
+
+// TODO: Replace mock data with a query
+const MOCK_PATTERNS: Pattern[] = [
```
DafyddLlyr, line 22: Next PR will be adding queries here, just using mock data here.

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/index.tsx
```diff
@@ -19,25 +27,33 @@ const AddComponentModal: React.FC<Props> = ({
...
+        flow,
+        ...(parent && { parent }),
+        ...(before && { before }),
+      },
+      search: { type: slug as NodeSearchParams["type"] },
+    });
+  };
+
+  const handleInsertPattern = (patternId: string) =>
+    console.log(`Inserting pattern ${patternId}!`);
```
DafyddLlyr, line 56: Intentional `TODO` - inserting (via existing `paste()` method) will be a follow up PR)

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/index.tsx
```diff
@@ -60,14 +76,27 @@ const AddComponentModal: React.FC<Props> = ({
...
+            mt: showBelow ? "4px" : "-4px",
+          },
         },
         backdrop: {
           sx: { backgroundColor: "rgba(0, 0, 0, 0.3)" },
         },
       }}
     >
-      <ComponentsTab onSelect={handleSelect} />
+      {showPatterns ? (
```
DafyddLlyr, line 90: Very temp feature flag, allowing these PRs just to merge as and when in small chunks.

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/ComponentsTab/ComponentRow.tsx
DafyddLlyr, line 1: This is mostly a11y changes - allowing keyboard navigation to work.

Current list on `main` - 

https://github.com/user-attachments/assets/3a0f2f71-2f39-4ac7-ae3c-fbdd8a0f9bc3

This PR - 

https://github.com/user-attachments/assets/e51a017c-4326-4038-958b-db5073fcd7c6

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { TabHeader } from "../TabHeader";
+import { PatternDetailPanel } from "./PatternDetailPanel";
+import { PatternRow } from "./PatternRow";
+
+export interface Pattern {
+  id: string;
+  slug: string;
+  name: string;
+  summary: string | null;
+  // data: Graph | null;
```
DafyddLlyr, line 15: I've not yet worked out how to best handle this...!

Fetching _all_ flow data, for all patterns, once upfront if probably a bad idea? We won't notice an issue immediately but it will just deteriorate over time as we add more and more patterns.

A few ideas...
 - Only fetch a flow when clicked, and preview panel opens
 - Fetch flows on scroll into view
 - Background fetch during idle time, e.g. when `AddComponentModal` is opened and keep a long cache
 - Not worry about it and don't over-optimise 😅 

The downside of not fetching all data right up front is potential latency before insert is possible (very minimal, can easily be handled) and we don't get the nice "component count" that was in the prototype.

Image

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/PatternDetailPanel.tsx
```diff
@@ -0,0 +1,71 @@
...
+            <CloseIcon fontSize="small" />
+          </IconButton>
+        </Box>
+        {pattern.summary && (
+          <Typography variant="body2">{pattern.summary}</Typography>
+        )}
+        <Button
+          variant="contained"
+          size="small"
+          // disabled={!pattern.data}
```
DafyddLlyr, line 59: Intentional `TODO`, see previous comment on loading `pattern.data` - we'll need it before / at insert.

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/PatternDetailPanel.tsx
DafyddLlyr, line 1: **Why no preview?**

Basically this is a pragmatic choice for now to speed up delivery. There's three real approaches here I think...

**1. Natively render simplified graph**
#7000 gets pretty complex, pretty fast with this approach. This seems like the easiest to rule out now given the composition of the code base - we need to repeat graph display logic (branching + styles) in a simplified format for this to really work.

**2. Generate static `.png`s at pattern publish**
On publish, trigger an endpoint to take a screenshot of `/:team/:flows/pattern-preview` which is a non-accessible URL which renders the graph in a simplified / styled manner. Stick screenshot in S3, link to pattern via DB column. High fidelity, pretty high implementation complexity, but no reinvention of graph display.

**3. Mock graphs / presets selected by Editors**
Create a number of simplified fake graph previews, apply either randomly or allow platformAdmins to select which one applies to which pattern. Very quick and easy, kind of totally pointless..? 😅 

To an extent, these all have _some_ merit as approaches but all involve some dedicated time and will have compromises baked into them. My favoured approach would be 2 - I can see these previews/thumnails being useful in a number of places (e.g. flow cards, template previews, OG images?)

Image

jessicamcinchak: Agree "skip" is right approach for MVP ! Thanks for clearly explaining future options here to revisit later.

jessicamcinchak (approved):
Great !

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/ComponentsTab/ComponentRow.tsx
ianjon3s: Wonderful, thanks for picking this up

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/index.tsx
ianjon3s: I think in the absence of a realistic screenshot (which I know we are hopefully working towards), the count is very important, in that we're indicating the size of the thing (set of components) that will be pasted into a flow.

My only suggestion to add to the above (which may be well off the mark, especially as we're not using publishing for patterns) is to add a component number value to each flow that is stored alongside the flow details.

PR #7094, opened 2026-07-31
fix: Option opening read-only in customisible template folder

ianjon3s:
## Issue

Trello card: https://trello.com/c/GsoCxT2d/3665-when-copying-pasting-content-in-customisable-folders-ensure-editable-styles-persist-on-paste

Reported here in Slack (originally flagged as caused by copying, but this seems to be resolved, however option clicks still cause bug): https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1775728748087909

- When editing a customisable folder within a template, clicking any option opens the component as read-only

Click any option to see:
https://editor.planx.dev/app/testing/article-4-template-bug%2CHUF6h0hpiR

## Solution

- Pass a folder ID into the option, so when clicking an option `FormModal` doesn't incorrectly compute the whole form as read-only

Click any option to see:
https://7094.planx.pizza/app/testing/article-4-template-bug%2CHUF6h0hpiR

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Checklist.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Checklist.tsx
@@ -186,6 +186,7 @@ const Checklist: React.FC<Props> = React.memo((props) => {
                   {children.map((child) => (
                     <Node
                       parent={props.id}
+                      containerFolderId={parent}
                       key={child.id}
                       {...child}
                       showTemplatedNodeStatus={props.showTemplatedNodeStatus}
@@ -200,6 +201,7 @@ const Checklist: React.FC<Props> = React.memo((props) => {
             {childNodes.map((child) => (
               <Node
                 parent={props.id}
+                containerFolderId={parent}
                 key={child.id}
                 {...child}
                 showTemplatedNodeStatus={props.showTemplatedNodeStatus}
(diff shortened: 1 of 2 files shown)


DafyddLlyr (changes requested):
Unfortunately this doesn't seem to hold up - it only fixes it for checklists at the root of a folder - https://editor.planx.dev/app/testing/article-4-template-bug%2CHUF6h0hpiR

See example flow where I've added a parent question - 
 - Question options still open as read-only
 - Checklist options now also read as read-only

Happy to take a closer look if you want a second pair of eyes - the issue should be related to the `parentNodeIsTemplatedInternalPortal()` function. 

Are we just missing a `parentId` somewhere, or is the parent check too naive and only looking one level up?

ianjon3s:
@DafyddLlyr if I do the above on the pizza link it appears to work correctly:
https://7094.planx.pizza/app/testing/article-4-template-bug%2CHUF6h0hpiR

DafyddLlyr:
Yeah so I tested this on the staging link not the pizza link... 🤦 😂

DafyddLlyr (approved):
Ok - apologies for the previous rejection here...! 😅 

After testing _the pizza_, I can confirm this all works as expected in any permutation I can find, and the fix makes sense. We need to know the container `parentId`, but options only go up one layer currently - stopping at the `Question`/`Checklist` - we care about the parent of _this_ node when working out permissions.

ianjon3s:
Thanks @DafyddLlyr !

PR #7095, opened 2026-07-31
feat(editor): Query pattern list and data

DafyddLlyr:
## What does this PR do?
- Replaces static mock data, with live queries to get actual pattern data here
- Handles error and loading states
- Addresses [this question](https://github.com/theopensystemslab/planx-new/pull/7092#discussion_r3689129514) by deferring `flow.data` loading until we open the panel

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/AddComponentModal.stories.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/AddComponentModal.stories.tsx
@@ -1,10 +1,12 @@
 import Paper from "@mui/material/Paper";
 import type { Meta, StoryObj } from "@storybook/tanstack-react";
+import { graphql, HttpResponse } from "msw";
 import React, { useState } from "react";
 
 import { COMPONENT_LIST_WIDTH, componentListFrameSx } from "./ComponentsTab";
 import type { ModalTab } from "./ModalTabs";
 import { ModalTabs } from "./ModalTabs";
+import { mockPatternData, mockPatterns } from "./PatternsTab/mocks";
 import { DETAIL_PANEL_WIDTH } from "./PatternsTab/PatternDetailPanel";
 
 /**
@@ -14,6 +16,22 @@ import { DETAIL_PANEL_WIDTH } from "./PatternsTab/PatternDetailPanel";
 const meta: Meta<typeof ModalTabs> = {
   title: "Editor Components/Modal/AddComponentModal",
   component: ModalTabs,
+  parameters: {
+    msw: {
+      handlers: [
+        graphql.query("GetPatterns", () =>
+          HttpResponse.json({ data: { patterns: mockPatterns } }),
+        ),
+        graphql.query("GetPatternData", ({ variables }) =>
+          HttpResponse.json({
+            data: {
+              pattern: { id: variables.id, data: mockPatternData },
+            },
+          }),
+        ),
+      ],
+    },
+  },
 };
 
 export default meta;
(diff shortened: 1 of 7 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/PatternDetailPanel.tsx
```diff
@@ -20,52 +22,88 @@ export const PatternDetailPanel: React.FC<Props> = ({
...
-        )}
-        <Button
-          variant="contained"
+}) => {
+  const { data, loading, error } = usePatternData(pattern?.id ?? null);
+  const graph = data?.pattern?.data;
+
+  // Both null while there's no graph data (loading, error, or not yet selected)
+  const componentCount = graph ? Object.keys(graph).length - 1 : null;
+  const componentCountLabel =
```
jessicamcinchak, line 31: nit: patterns will use nested flows (eg osl/property-types) and we're counting un-flattened data here (which will omit all nested flow nodes) - do we care about communicating this somehow? 

Is "count" meaningful (especially as this will count each option/answer uniquely too); is it actually some concept of "dependencies" that matters more? (Eg "This pattern uses osl/property-types and osl/some-other-nested-flow") .

Appreciate just following original mock-ups for now and no expectations to change here, but one to keep thinking about maybe !

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/PatternDetailPanel.tsx
```diff
@@ -20,52 +22,88 @@ export const PatternDetailPanel: React.FC<Props> = ({
...
+    componentCount !== null && componentCount >= 1
+      ? `${componentCount} component${componentCount === 1 ? "" : "s"}`
+      : null;
+
+  // A pattern needs at least one component to be insertable
+  // We should aim to catch this when a pattern is made copyable
```
jessicamcinchak, line 37: 👍

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/queries.ts
```diff
@@ -0,0 +1,51 @@
...
+      }
+      order_by: { name: asc }
+    ) {
+      id
+      name
+      slug
+      summary
+    }
+  }
+`;
```
jessicamcinchak, line 20: 👍 simple !

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/AddComponentModal.stories.tsx
```diff
@@ -14,6 +16,22 @@ import { DETAIL_PANEL_WIDTH } from "./PatternsTab/PatternDetailPanel";
...
+        graphql.query("GetPatterns", () =>
+          HttpResponse.json({ data: { patterns: mockPatterns } }),
+        ),
+        graphql.query("GetPatternData", ({ variables }) =>
+          HttpResponse.json({
+            data: {
+              pattern: { id: variables.id, data: mockPatternData },
+            },
+          }),
+        ),
```
jessicamcinchak, line 31: Nice conditional, deferral pattern for `data` for now ➕

jessicamcinchak (approved):

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/PatternDetailPanel.tsx
DafyddLlyr: Yeah I also wondered about this when it comes to nested flows, but hadn't considered `Option` components.

For now I think this simple approach works and gives the user some sense of cause and effect, but I agree this is one to keep an eye on and come back to. I really like the dependencies idea - this feels more meaningful to me.

PR #7097, opened 2026-07-31
fix: Clone appearance when cloning or releasing

ianjon3s:
## Issue

Reported in Slack here:
https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1783676184671859

- When cloning a node, the original/source of the clone did not immediately receive clone bordered styling (browser refresh corrected this)
- Similarly when releasing a clone (make unique or delete on one of a pair of clones, the other/remaining node didn't not immediately lose clone styling

**Testing:**
https://editor.planx.dev/app/testing/clones

## Solution

Several things happening here:
- `isClone(props.id)` is being cached in `Question.tsx`, preventing the border from updating from first render
- Moving `isClone(props.id)` inside the Zustand selector solves the above
- Also refresh node data uniformly in `Node.tsx` and un-memoizing the `Make unique` button in `FormModal.tsx`, both of which had smaller instances of the same staleness problem

**Testing:**
https://7097.planx.pizza/app/testing/clones

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Node.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Node.tsx
@@ -26,7 +26,6 @@ const Node: React.FC<any> = (props) => {
 
   const allProps = {
     ...props,
-    ...node,
     wasVisited,
     tags: node.data?.tags,
     isTemplatedNode: node.data?.isTemplatedNode,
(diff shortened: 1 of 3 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Node.tsx
DafyddLlyr, line 1: I don't quite understand this change - we pass in the whole destructured `...node` into all components, but it's never referenced directly in any consumer? Is there a prop like `id` which is already used, but the re-render is needed?

DafyddLlyr (approved):
Solution in `Question` works and makes sense..!

One question on the change in `Node.tsx` which I don't think is necessary.

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Node.tsx
ianjon3s: Good catch, and leftover from a first attempt to get this working. I've reverted 👍

PR #7108, opened 2026-08-02
feat: Shared `insertGraph()` function

DafyddLlyr:
## What does this PR do?
- Creates a new `insertGraph()` function. The aim of this is to share business logic across both `pasteNode()` and `insertPattern()` - these should both function consistently under the hood.
- Test coverage for this shared logic
- Implements this directly into `pasteNode()` - I'll prompt the content team to test this on the Pizza

### DIFF
--- a/apps/editor.planx.uk/src/@planx/graph/__tests__/insert_graph.test.ts
+++ b/apps/editor.planx.uk/src/@planx/graph/__tests__/insert_graph.test.ts
@@ -1,0 +1,207 @@
+import type { Graph } from "..";
+import { insertGraph } from "..";
+
+let count = 0;
+const deterministicId = () => `ID_${count++}`;
+
+beforeEach(() => {
+  count = 0;
+});
+
+/** All graph nodes, minus _root */
+const nodeIds = (graph: Graph) =>
+  Object.keys(graph).filter((id) => id !== "_root");
+
+const emptyFlow = (): Graph => ({ _root: { edges: [] } });
+
+const pattern: Graph = {
+  _root: { edges: ["question", "notice"] },
+  question: { type: 100, data: { text: "Question?" }, edges: ["yes", "no"] },
+  yes: { type: 200, data: { text: "Yes" } },
+  no: { type: 200, data: { text: "No" } },
+  notice: { type: 8, data: { title: "Notice" } },
+};
+
+describe("inserting a graph", () => {
+  describe("basic behaviour", () => {
+    test("adds every node from the source, with fresh ids", () => {
+      const [graph] = insertGraph(pattern, { idFn: deterministicId })(
+        emptyFlow(),
+      );
+
+      expect(graph).toEqual({
+        _root: { edges: ["ID_0", "ID_3"] },
+        ID_0: {
+          type: 100,
+          data: { text: "Question?" },
+          edges: ["ID_1", "ID_2"],
+        },
+        ID_1: { type: 200, data: { text: "Yes" } },
+        ID_2: { type: 200, data: { text: "No" } },
+        ID_3: { type: 8, data: { title: "Notice" } },
+      });
+    });
+
+    test("does not carry over the source's root node", () => {
+      const [graph] = insertGraph(pattern, { idFn: deterministicId })(
+        emptyFlow(),
+      );
+
+      // The only root is the destination's own, holding the new top level nodes
+      expect(Object.keys(graph)).not.toContain("question");
+      expect(graph["_root"]).toEqual({ edges: ["ID_0", "ID_3"] });
+    });
+
+    test("keeps the order of top level nodes", () => {
+      const [graph] = insertGraph(pattern, { idFn: deterministicId })(
+        emptyFlow(),
+      );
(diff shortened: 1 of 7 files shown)


apps/editor.planx.uk/src/@planx/graph/__tests__/insert_graph.test.ts
```diff
@@ -0,0 +1,207 @@
...
+
+      // Each insert maintains its own clone...
+      const firstCloneId = cloneBeneath(twice, firstQuestionId);
+      const secondCloneId = cloneBeneath(twice, secondQuestionId);
+
+      // ...but the clones never meet
+      // Clones are not shared across repetitions of the same pattern
+      expect(firstCloneId).not.toBe(secondCloneId);
+      expect(nodeIds(twice)).toHaveLength(8);
+    });
```
DafyddLlyr, line 182: Would love a sense check here...!

This behaviour makes sense to me, but I'm not sure how well it hold up in terms of planned pattern content, or for users. I do think this is the expected behaviour but I'd rather be explicit about it upfront as we introduce this.

apps/editor.planx.uk/src/@planx/graph/index.ts
```diff
@@ -813,3 +813,60 @@ export const buildGraphFromNodes = (
...
   };
 };
+
+/**
+ * Insert a copy of an entire graph (e.g. a pattern or pasted content) into current graph
+ *
+ * All mutations are collected into a single set of ops (due to wrap()), so an insert
+ * is a single ShareDB transaction (and so a single "undo")
+ */
+export const insertGraph =
```
DafyddLlyr, line 823: Logic below is almost entirely the current `pasteNode()` logic. 

Only real changes are the signature (this needs a `Graph` with `_root` node), and extraction of `stipTemplateNodeProps()` which I think is a paste only concern.

apps/editor.planx.uk/src/@planx/graph/__tests__/insert_graph.test.ts
```diff
@@ -0,0 +1,207 @@
...
+      // Both copies are present in full, sharing no nodes between them
+      expect(firstCopy).toHaveLength(4);
+      expect(secondCopy).toHaveLength(4);
+      expect(nodeIds(twice)).toHaveLength(8);
+
+      // Inserting again leaves the first copy untouched - edits can't bleed across
+      // No unexpected "clone" behaviour introduced
+      firstCopy.forEach((id) => expect(twice[id]).toEqual(once[id]));
+    });
+  });
```
jessicamcinchak, line 115: 👍

apps/editor.planx.uk/src/@planx/graph/__tests__/insert_graph.test.ts
```diff
@@ -0,0 +1,207 @@
...
+    );
+
+    const createdNodeIds = ops
+      .filter((op) => "oi" in op && op.p.length === 1)
+      .map((op) => op.p[0]);
+
+    expect(createdNodeIds.sort()).toEqual(
+      ["ID_0", "ID_1", "ID_2", "ID_3"].sort(),
+    );
+  });
```
jessicamcinchak, line 130: 👍

apps/editor.planx.uk/src/@planx/graph/__tests__/insert_graph.test.ts
jessicamcinchak: Tricky one! I _think_ this is right, but would be curious to hear if content team has differing opinion !

apps/editor.planx.uk/src/@planx/graph/index.ts
jessicamcinchak: Really clear comments, thanks !

jessicamcinchak (approved):
Think this all looks good from code review! Seems like an ideal pizza to share with content team though just in case any missed assumptions here !

apps/editor.planx.uk/src/@planx/graph/__tests__/insert_graph.test.ts
DafyddLlyr: Will post links + question Monday before merging 👌

jessicamcinchak: Actually more I think about about this, I don't think there's any reasonable way to _maintain_ cloned relationships when inserting a pattern more than once ! Eg because the original insert may have already been edited further, etc ! And plain "matching" could actually run the risk of linking a non-patterned node as now a clone?

I do think we should make sure this is an explicit known rule of using patterns - but am changing my mind that there's actually an "open question" about _how_ to implement here so I'd say don't let it block merging (at least to staging) !

DafyddLlyr: Agreed - I also hit this - we'd need to indefinitely track source/lifecycle of a node and it would get very complex, very fast, for little to no gain. It's just something I think we need users to understand - clones _within_ a single pattern will work, but won't work across inserted patterns.

PR #7112, opened 2026-08-02
chore(editor): Drop unused dev dependencies

DafyddLlyr:
As far as I can tell these are all unused..! Spotted this when dependabot opened https://github.com/theopensystemslab/planx-new/pull/7105 - this wasn't a dependency I recognised.

Verified these as unused via the following methods - 
 - `pnpm dlx knip`
 - `pnpm dlx depcheck`
 - `Ctrl + F` in the code base
 - Build Editor locally
 - All CI checks pass.
 
The above tool also identified a number of other unused files and imports. I'll also take a look at these and run the above for other projects. It might be worth us setting up a GHA which runs these checks?


jessicamcinchak (approved):
Great ! 🧹

PR #7121, opened 2026-08-04
feat: Mop-up cronjob for GovPay

DafyddLlyr:
## What does this PR do?
This PR adds the "mop-up job" GOV.UK Pay docs recommend here - https://docs.payments.service.gov.uk/integrate_with_govuk_pay/#use-an-automatic-mop-up-job-recommended

This is a Hasura cron that re-checks GOV.UK Pay for any session that began a payment but was never submitted, corrects `payment_status` with the true result, and alerts `#planx-notifications-internal` for anything paid with no submission.

Trello ticket - https://trello.com/c/gzmBq67h/3609-implement-mop-up-job-for-govuk-payments

## Design
I've aimed to keep this lightweight as we're aiming to migrate away relatively soon. If this is lacking in any area we can always build on it in future.
- No new tables or migrations - we don't write failures to the db, just log to Slack
- Re-uses reuse the existing `GET /pay/:team/:paymentId` route rather than a new GOV.UK Pay request. This already resolves the team's token, calls GOV.UK Pay, and writes the result to `payment_status` as a side effect.
- No auto-resubmission on failure - just notify us via Slack.

```mermaid
sequenceDiagram
    participant Cron as Hasura Cron
    participant API as API (reconcile-payments)
    participant DB as Hasura (payment_status / lowcal_sessions)
    participant GovPay as GOV.UK Pay
    participant Slack

    Cron->>API: POST /reconcile-payments (every 30 min)
    API->>DB: query unsubmitted sessions with recent payment activity
    DB-->>API: candidate sessions

    loop each candidate outside the 3h grace period
        alt already failed / cancelled / error
            Note over API: skip - no money taken
        else already known success
            Note over API: flag as unsubmitted
        else needs re-check
            API->>GovPay: GET /pay/:team/:paymentId (self-call)
            GovPay-->>API: current payment state
            API->>DB: write true status to payment_status
            alt status = success
                Note over API: flag as unsubmitted
            else
                Note over API: discard
            end
        end
    end

    opt unsubmitted sessions found
        API->>Slack: post alert to planx-notifications-internal
        Slack-->>API: (failure is logged, retried next run)
    end

    API-->>Cron: 200 { checked, unsubmitted, errors }
```

PR #7122, opened 2026-08-04
fix(editor): Allow retries on GovUK Pay failures

DafyddLlyr:
## What's the problem?
This is spun out from https://github.com/theopensystemslab/planx-new/pull/7121

We've had incidents where payments have been made, without the related submission getting through.

When GOV.UK Pay redirects an applicant back, we fire a single `GET /pay/:team/:paymentId`. This request is the only thing which writes a `success` row to `payment_status` and lets the flow advance to the `Send` node.

`refetchPayment` sets `IncompletePaymentConfirmed` inside the `try/catch` block and the `catch` only logs an error. This means that any issue like an API 5xx, slow GovPay response, or a connection issue can leave the user stuck on `fetching_payment` with a perpetual "Loading payment information" spinner.

If the applicant gives up at this point, the payment is made but the submission is not completed.

We have no _direct_ proof that this is what's happened to users in the past unfortunately, but it's a failure method which we're not accounting for and could lead to this result. See video below from Playwright running the test added here against `main` - the user ends up stuck in a loading state.

[video.webm](https://github.com/user-attachments/assets/24307596-128b-4a23-b10b-23191f1fe570)

## What's the solution?
A new `status_unknown` component state, which both failure paths (the `catch`, and a missing `payment_id`) now dispatch. It shows the user a "Check payment status" button which just re-runs the status request, plus a warning explaining that we couldn't read their payment status.

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/Pay/Public/Pay.test.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/Pay/Public/Pay.test.tsx
@@ -1,12 +1,14 @@
+import type { GovUKPayment } from "@opensystemslab/planx-core/types";
 import { PaymentStatus } from "@opensystemslab/planx-core/types";
 import { ComponentType as TYPES } from "@opensystemslab/planx-core/types";
-import { act, screen } from "@testing-library/react";
+import { act, screen, waitFor } from "@testing-library/react";
 import { logger } from "airbrake";
 import ErrorFallback from "components/Error/ErrorFallback";
+import { http, HttpResponse } from "msw";
 import type { FullStore, Store } from "pages/FlowEditor/lib/store";
 import { useStore } from "pages/FlowEditor/lib/store";
-import React from "react";
 import { ErrorBoundary } from "react-error-boundary";
+import server from "test/mockServer";
 import { setup } from "test/utils";
 import type { Breadcrumbs } from "types";
 import { ApplicationPath } from "types";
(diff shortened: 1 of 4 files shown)


apps/editor.planx.uk/src/@planx/components/Pay/Public/Pay.test.tsx
```diff
@@ -228,6 +230,221 @@ describe("Pay component when fee is undefined or £0", () => {
...
+          handleSubmit={handleSubmit}
+          govPayMetadata={[]}
+        />
+      </ErrorBoundary>,
+    );
+
+    expect(await screen.findByText("Check payment status")).toBeInTheDocument();
+    expect(
+      screen.getByText(/We could not check the status of your payment/),
+    ).toBeInTheDocument();
```
jessicamcinchak, line 343: Great !

apps/editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
```diff
@@ -206,22 +218,22 @@ function Component(props: Props) {
...
       }
+
       dispatch(Action.IncompletePaymentConfirmed);
     } catch (err) {
       // XXX: There's probably been an issue fetching the payment status,
-      //      but there's a chance that the user might've made a successful
-      //      payment. We silently log the error and the service continues.
+      //      allow user to re-run status check instead of silently failing
       logger.notify(err);
+      dispatch(Action.PaymentStatusUnknown);
```
jessicamcinchak, line 236: ➕

jessicamcinchak (approved):
Clever solution, nice & straightfoward UI ! 

This is a really good edge-casey journey for us to also be keeping in mind going into Stripe migration. Thanks again for investigating and spotting !

apps/editor.planx.uk/src/@planx/components/Pay/Public/Pay.tsx
DafyddLlyr: Yeah the note we left ourselves here was the giveaway that kicked off this realisation..!

PR #7126, opened 2026-08-04
fix: Support for FME submission retries

DafyddLlyr:
## What's the problem?
`fme` has been missed a submission type, meaning we're unable to retry submissions directly within the Editor.

Updating types here resolves this. These types are driven by the view `submission_services_log` and aren't used elsewhere. Once we have typegen for Hasura setup these will be generated and imported, no need to manually track and sync.

Frontend types now match the following `event_type` values - 

https://github.com/theopensystemslab/planx-new/blob/5af3d2192fc81525255bbfbeb603347b67e937b8/apps/hasura.planx.uk/migrations/default/1783614492150_update_submission_services_log_address/up.sql#L25-L33

Further context - https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1785795651320889


jessicamcinchak (approved):
Thanks for catching + aligning !

PR #7127, opened 2026-08-04
chore: disable Question and Checklist pseudo-notes ahead of migration

jessicamcinchak:
We're preparing to migrate the existing pseudo-notes _nodes_ to new `flow_notes` structure.

In order to not have a moving target while migrating, I'm proposing we disable adding _new_ pseudo-note nodes but still ensure existing ones are visible and navigate correctly in the interim. We'll move forward with full deprecation once migrated.

Please note this does not prevent nodes or flows with pseudo-notes from being cloned, copied, templated, nested in the meantime. It's imperfect !

Applies to each Question, Checklist, Responsive Question, Responsive Checklist.

(This PR is actually independent to internal notes property one, just "stacking" to get a feel for it!)

Screenshot from 2026-08-04 12-04-52


apps/editor.planx.uk/src/@planx/components/Checklist/Editor.tsx
```diff
@@ -42,6 +43,13 @@ export const ChecklistEditor: React.FC<Props> = (props) => {
           );
         }
       },
+      validate: ({ options, ...values }) => {
+        const errors: FormikErrors<FormikValues> = {};
+        if (values.text && (!options || !options.length)) {
+          errors.text =
+            "Checklists can no longer be used as notes and must have at least one option";
+        }
+      },
```
jessicamcinchak, line 52: Validation error is easy solution which is hit _both_:
- When trying to create new node
- When trying to update existing pseudo-note node text

Will be able to fully revert once types are updated in the future though to prevent `!options.length` !

e2e/tests/ui-driven/src/templates.spec.ts
```diff
@@ -151,6 +152,7 @@ test.describe("Templates", () => {
         ComponentType.Question,
         OPTIONAL_NODE_TITLE,
         { instructions: OPTIONAL_NODE_INSTRUCTIONS, required: false },
+        ["Yes", "No"],
```
jessicamcinchak, line 155: This change has highlighted that our mock e2e source template was a bit buggy! We were creating/testing pseudo-notes, when we should create true questions _with options_.

apps/editor.planx.uk/src/@planx/components/shared/BaseQuestion/Editor/index.tsx
```diff
@@ -63,9 +63,6 @@ const BaseQuestionComponent: React.FC<Props> = (props) => {
 
   const isTemplate = useStore.getState().isTemplate;
 
-  const title =
-    type === ComponentType.Question ? "Question" : "Responsive question";
```
DafyddLlyr, line 67: Nice tidy up!

DafyddLlyr (approved):
Great!

PR #7131, opened 2026-08-05
chore: sync `flow_notes_*` from prod to other envs

jessicamcinchak:
Flow data is synced (fully truncated + replaced on other envs) so corresponding notes need to be too.

We want the data sync in place _before_ we migrate! For migrating `internal_notes` specifically, I'd like to start from an empty `flow_notes_*` so that we don't need to account for "crashes" between migrating content and "test notes" attached to existing nodes.

I've run the following direct on prod to enable sync: 
```sql
GRANT SELECT ON "public"."flow_note_content" TO github_actions;
GRANT SELECT ON "public"."flow_note_positions" TO github_actions;
```


scripts/seed-database/write/flow_note_content.sql
```diff
@@ -0,0 +1,25 @@
+-- insert flow_note_content overwriting conflicts
+CREATE TEMPORARY TABLE sync_flow_note_content (
+  id uuid,
+  text text,
+  color text,
```
jessicamcinchak, line 5: Is `color` actually used in final designs (eg no in-modal picker), can this column be dropped in a future Hasura migration and simply default to sticky note yellow in UI instead?

scripts/seed-database/write/flow_note_positions.sql
```diff
@@ -0,0 +1,27 @@
+-- insert flow_note_positions overwriting conflicts
+CREATE TEMPORARY TABLE sync_flow_note_positions (
+  id uuid,
+  note_id uuid,
+  flow_id uuid,
+  node_id text,
+  placement jsonb,
+  created_by integer,
+  created_at timestamptz,
+  updated_at timestamptz
```
jessicamcinchak, line 10: Also had another question about this schema while hooking up sync: 
- For `_content` we track who/when created and who/when updated
- But for `_positions` we only track who/when created and _when_ updated - but _any_ user with editing positions to this flow can move an existing note, should we also be tracking `updated_by` ? Is this important?
  - When we think about displaying who "last edited" a note, should that mean updating content, updating position, or both/either? (For nodes, it's both/either)

scripts/seed-database/write/flow_note_content.sql
joecarver: Ah yeah I think we can drop this - it's not used in any of the UI or expected in components or anything, so it can go.

scripts/seed-database/write/flow_note_positions.sql
joecarver: Ah good spot! Yes I think we should be tracking `updated_by` for positions as well, and agreed on "last edited" being both/either as consistent with nodes.

joecarver (approved):
Looks good thanks for adding this! Just need to remember to update these when we make schema changes for dropping colour / adding `updated_by`

scripts/seed-database/write/flow_note_content.sql
jessicamcinchak: Picked this up here https://github.com/theopensystemslab/planx-new/pull/7172

Don't want to keep introducing new changes (`copy_flow` psql function, migration scripts) while unused column is still lingering.

PR #7134, opened 2026-08-06
fix(a11y): Add href to SkipLink, fix references

ianjon3s:
## Issue

Relates to:
p.24 - DAC_Name_Role_Value_01 - https://trello.com/c/q26P6L7s/3717-a-name-role-value-01

> JavaScript event handlers have been attached to <a> elements without a href to emulate
links. Assistive technology may not be able to recognise these components as links in order
to present this purpose to users as a href is usually required for the link role to be
recognised.

## Solution

I believe the JavaScript handler was introduced due to a previously flagged (by DAC) bug that meant using a href link caused the page to refresh, and thus return focus to the top of the page. This appears to no longer be an issue, so we can return to using a regular href link.

- Remove JavaScript in favour of vanilla href link
- Contain skip to content text in `contentWrap`, simplify styling
- Add missing `#main-content` ids to `StatusPage` and `TestWarningPage`

### Testing
https://7134.planx.pizza/buckinghamshire/get-ecology-advice/preview

### DIFF
--- a/apps/editor.planx.uk/src/pages/Preview/StatusPage.tsx
+++ b/apps/editor.planx.uk/src/pages/Preview/StatusPage.tsx
@@ -33,7 +33,12 @@ const StatusPage: React.FC<PropsWithChildren<Props>> = ({
   const resetPreview = useStore().resetPreview;
 
   return (
-    <>
+    <Box
+      component="main"
+      id="main-content"
+      tabIndex={-1}
+      sx={{ width: "100%" }}
+    >
       <Box sx={{ width: "100%" }}>
         <Banner
           heading={bannerHeading}
@@ -84,7 +89,7 @@ const StatusPage: React.FC<PropsWithChildren<Props>> = ({
           </>
         )}
       </Card>
-    </>
+    </Box>
   );
 };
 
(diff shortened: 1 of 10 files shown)


apps/editor.planx.uk/src/pages/Preview/StatusPage.tsx
```diff
@@ -33,7 +33,12 @@ const StatusPage: React.FC<PropsWithChildren<Props>> = ({
   const resetPreview = useStore().resetPreview;
 
   return (
-    <>
+    <Box
+      component="main"
+      id="main-content"
+      tabIndex={-1}
+      sx={{ width: "100%" }}
+    >
```
DafyddLlyr: We already have a `<Main/>` element for this (which should be wrapping already.

It's possible we missed this during the Tanstack Router migration if it's no longer here.

Can we please - 
 - Implement the shared `<Main`/>
 - Cross-check it's not missing elsewhere

DafyddLlyr (changes requested):
One small question about using `<Main/>`..!

apps/editor.planx.uk/src/pages/Preview/StatusPage.tsx
ianjon3s: Thanks for catching this @DafyddLlyr , took me a while to untangle the various routes, but things were a lot more complex than needed. The named `<MainContainer>` in `PublicRoute` was using a regular box/div, meaning `<main>` references were being added downstream. Correcting this in `PublicRoute` (using the existing `Main.tsx` component) means we can remove all repetition.

DafyddLlyr (approved):
Great - thanks!

PR #7138, opened 2026-08-07
feat(editor): Dummy UI for Stripe onboarding workflow

DafyddLlyr:
## What does this PR do?
Adds rough dummy UI for Stripe onboarding workflow - user clicks button, async action takes place (realistically a redirect for SSO login w/ Stripe Connect?), and then account status is shown. I've also dropped the `<Status/>` placeholder - I think we can incorporate this into a single section.

https://github.com/user-attachments/assets/bee53cc3-2fc3-4332-a379-60487f854446


ianjon3s (approved):
Really smart, would be great to show a teaser demo of this in a future show & tell (if we're confident the interaction will be this smooth)

PR #7151, opened 2026-08-10
refactor: Perform 'copy flow' operation in a transaction

joecarver:
Following on from [this discussion](https://github.com/theopensystemslab/planx-new/pull/7082#discussion_r3702299529), this PR changes the copy flow behaviour to run in an single SQL transaction, instead of a series of independent mutations.

Added/updated a number of tests for this:
- `api/copyFlow.test.ts` - this now just needs to test routing and auth as the logic for copying has moved
- `hasura/copy_flow.test.js` - this tests what goes in and out of hasura itself, inserting seed data using hasura, and invoking `copy_flow` mutation which executes the SQL
- `hasura/copy_flow.test.sql` - this tests that the sql transaction rolls back appropriately when given invalid notes - we can't do this via hasura tests because hasura validates the inputs itself
- e2e tests - full pipeline tests


apps/api.planx.uk/modules/flows/copyFlow/service.ts
```diff
@@ -18,24 +23,38 @@ const copyFlow = async ({
...
+          is_service: flow.isService,
+          is_pattern: flow.isPattern,
+          replace_value: replaceValue,
+          creator_id: Number(userId),
+        },
+      },
+    );
+
+    // Publish immediately, same as createFlow() does for every newly-created flow
+    await publishFlow(insertedFlow.id, "Created flow");
```
jessicamcinchak: nit: now that this is a unique function and not sharing `createFlow`, can we update this message to be "Created flow from copy" for better clarity in editor "History"?

jessicamcinchak (changes requested):
Took a quick first pass through this pizza and unable to copy any flows (with _and_ without notes)‼️ 

`copy-flow` endpoint is returning a 500 with error: 
```
{
    "error": "Failed to copy flow. Error: Error: field 'copy_flow' not found in type: 'mutation_root' ... "
}
```

You've added/updated many tests here which are all _passing_, so does this "real" UI failure point to a gap in coverage or insufficient mock? Can we add Hasura introspection tests _for pg functions_ the same way we do for tables?

I've added notes here in case useful for testing / I'll re-review once a fix pushed ! https://7151.planx.pizza/app/barnet/pre-application-planning-advice

joecarver:
> Took a quick first pass through this pizza and unable to copy any flows (with _and_ without notes)‼️

This was due to the `public` schema not being specified with the new function name so hasura couldn't find it, was fixed in https://github.com/theopensystemslab/planx-new/pull/7151/commits/b25e2b8328eb43afa3887661169774558a250b3e but the Pizza upsert does not re-apply database migrations I guess. So I'm re-building the pizza from scratch, will let you know when ready!

apps/api.planx.uk/modules/flows/copyFlow/service.ts
joecarver: https://github.com/theopensystemslab/planx-new/pull/7151/commits/6cf467a3a8179c6c722e4942a75b7ec70bfc4d61

joecarver:
@jessicamcinchak Pizza is freshly baked and tested - seems to be working for me now. Also updated the summary text as requested.

jessicamcinchak (approved):
All working for me here now, thanks for updating pizza and explaining failure point / lots of test coverage is re-assuring here !

PR #7153, opened 2026-08-10
chore: run `reconcile payments` CRON every 3 hours, not every 30 minutes

jessicamcinchak:
This felt quite noisy yesterday ! Every 3 hours should let us ID proactively - see https://opensystemslab.slack.com/archives/C04EEBR1E0P/p1786318900791539

Skipping pizza as this only runs on prod anyways!

### DIFF
--- a/apps/hasura.planx.uk/metadata/cron_triggers.yaml
+++ b/apps/hasura.planx.uk/metadata/cron_triggers.yaml
@@ -18,7 +18,7 @@
   comment: 'Checks for any failed submissions and logs them to #planx-notifications-internal to investigate further'
 - name: reconcile_payments
   webhook: '{{HASURA_PLANX_API_URL}}/webhooks/hasura/reconcile-payments'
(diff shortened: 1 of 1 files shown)


apps/hasura.planx.uk/metadata/cron_triggers.yaml
```diff
@@ -18,7 +18,7 @@
   comment: 'Checks for any failed submissions and logs them to #planx-notifications-internal to investigate further'
 - name: reconcile_payments
   webhook: '{{HASURA_PLANX_API_URL}}/webhooks/hasura/reconcile-payments'
-  schedule: '*/30 * * * *'
+  schedule: 0 */3 * * *
```
jessicamcinchak, line 21: I was unsure about the dropped `'` here, but based on L30 below looks like should be fine un-quoted !

joecarver (approved):

PR #7157, opened 2026-08-10
feat: make notes searchable

joecarver:
Adds notes as a separate result list for search tab, using just the text for the search facet. Also tests.

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/facets.ts
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/facets.ts
@@ -1,6 +1,7 @@
 import type { IndexedNode } from "@opensystemslab/planx-core/types";
 import { flatFlags } from "@opensystemslab/planx-core/types";
 import type { FuseOptionKey, FuseOptionKeyObject } from "fuse.js";
+import type { FlowNote } from "hooks/data/useFlowNotes";
 import { get } from "lodash";
 
 export type SearchFacets = Array<FuseOptionKey<IndexedNode>>;
@@ -172,3 +173,6 @@ export const ALL_FACETS: SearchFacets = [
   ...feedback,
   ...DATA_FACETS,
 ];
+
+/** Notes aren't flow nodes - they're a separate `FlowNote` shape with a single searchable field */
+export const NOTE_FACETS: Array<FuseOptionKey<FlowNote>> = ["text"];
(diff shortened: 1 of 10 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/facets.ts
```diff
@@ -172,3 +173,6 @@ export const ALL_FACETS: SearchFacets = [
   ...feedback,
   ...DATA_FACETS,
 ];
+
+/** Notes aren't flow nodes - they're a separate `FlowNote` shape with a single searchable field */
+export const NOTE_FACETS: Array<FuseOptionKey<FlowNote>> = ["text"];
```
jessicamcinchak, line 178: 👍

apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/index.tsx
```diff
@@ -78,20 +75,45 @@ const Search: React.FC = () => {
...
+  const { results: noteResults, search: searchNotes } = useSearch({
+    list: noteSearchList,
+    keys: NOTE_FACETS,
+  });
+
+  const results = useMemo<SearchableResult[]>(
+    () => [...nodeResults, ...noteResults],
+    [nodeResults, noteResults],
+  );
+
```
jessicamcinchak, line 106: I think there's one key bit of unexpected behavior about this current `merge` logic: "Notes" are always last in the list of results, even if the "best" match ! 

Eg search for the word "attached" here: https://7157.planx.pizza/app/lambeth/request-copies-of-enforcement-notices-and-register (I get 9 results, many "fuzzy" matched nodes and a direct match on a note; but the note is the bottom of the list and I have to scroll to see it). 

Is it possible to merge in a way that preserves the _quality_ of each match, rather than Note matches always "last" like now?

jessicamcinchak (commented):
Thanks for this, appreciate test coverage 🌟 

One key question/comment about result ordering below, otherwise all working as expected !

apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/index.tsx
joecarver: https://github.com/theopensystemslab/planx-new/pull/7157/commits/72231ea702151a0fbb87a77d845c1c27f136e48c - seems to do a good job of it by just using the score that Fuse returns!

jessicamcinchak (approved):
Including `score` is perfect solution here, order of results feels much more accurate and natural now - thanks !

PR #7158, opened 2026-08-10
fix: don't delete notes when deleting nodes - so restore still works

joecarver:
The implication of this is that we will end up with dangling notes that have no node, in the case where the node is deleted but not restored. Perhaps we should have a cleanup job to deal with those? Or maybe its fine to leave them.

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/lib/repositionNotes.test.ts
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/notes/lib/repositionNotes.test.ts
@@ -26,8 +26,8 @@ beforeEach(() => {
 });
 
 describe("repositionNotesForDeletedNodes", () => {
-  it("deletes all attached notes whose node was deleted, regardless of author", async () => {
-    let deleteIds: string[] | undefined;
+  it("leaves attached notes whose node was deleted untouched (orphaned), regardless of author", async () => {
+    let mutationCalled = false;
 
     server.use(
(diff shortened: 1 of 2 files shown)


jessicamcinchak (commented):
A couple quick thoughts on this one! 
- This "re-anchors" _attached_ notes as expected, but I definitely agree we don't want a bunch of orphaned "notes" in the database and will need some cleanup mechanism for these
  - One idea: Since we really don't allow "undos" after six months (minimal "History" is displayed), can we add a `deleted_at` timestamp column on `flow_notes_` that is null by default, populated on `DeleteFlowNotePositions`, and re-nullified if `ReanchorFlowNotePosition` is called? Then add these tables to our existing data sanitation crons to drop any rows where `deleted_at` >= 6 months ago? Is that as simple as adding a second variable to these existing mutations, or more tedious?
- Even though "attached" notes are working here, I think it's really strange to have a mix of behaviour where standalone notes can be retained even if I restore to a time _before_ I added them (as long as they aren't a parent/before of removed nodes)
  - And generally the visual language still feels a bit off here because notes are _not_ in history, but they can be [partially] restored?

So I think big open questions here are:
- Can we land on a solution that works consistently for notes attached to nodes and standalone notes?
- Can we land on a solution that doesn't lead to orphaned records in notes tables?
- Have you sense-checked any of these "what do we want/expect history, deletion, and restore to entail" questions / proposed solutions with content team? They are ultimate end users here!! (Eg are they actually fine with "hard" deletion and inability to restore and we're over-complicating this ??)

PR #7162, opened 2026-08-11
feat: parallel grouped submissions log v2

zz-hh-aa:
Reopens #7045, which was reverted in #7057 following deployment and rollback issues, description duplicated again below. 

This PR starts the grouped submissions log feature:

**db changes**

- Creates a submissions_grouped view that is similar to the submission_services_log view but only has one row per session and does not include columns such as response
- This replicates a lot of the logic from submission_services_log but I figured it was better to avoid complicated dependencies and would be better for performance to not be querying a non-materialized view
- The aggregation logic happens here--if the most recent event for any event fails, then we will show a failure; if there are no failures, show the status of the most recent event
- We might still use submission_services_log for the detailed view of events (and there is a daily_failed_submissions dependency to keep in mind, as Jess flagged!)

**Frontend changes**

- Creates GROUPED_SUBMISSIONS feature flag and duplicates certain components (eg EventsLog and Submissions -> EventsLogGrouped and SubmissionsGrouped)
    - At dev call we decided that route-level feature flagging with replicated code was probably the simplest way of feature flagging this work
- Removes 'View submission', 'Download', 'Response' and 'Resubmit' columns from submissions log (info to go into detail modal)
- Makes rows clickable
- Creates placeholder submission detail modal

Before:
image

After:
image

with modal placeholder:
image

### DIFF
--- a/apps/editor.planx.uk/src/lib/featureFlags.ts
+++ b/apps/editor.planx.uk/src/lib/featureFlags.ts
@@ -1,6 +1,7 @@
 // add/edit/remove feature flags in array below
 export const AVAILABLE_FEATURE_FLAGS = [
   "EXPLORE",
+  "GROUPED_SUBMISSIONS",
   "NOTES",
   "PATTERN_SELECT",
   "STRIPE_MIGRATION",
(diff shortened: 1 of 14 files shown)


jessicamcinchak (approved):
Per notes here https://opensystemslab.slack.com/archives/C01E3AC0C03/p1786441766967639

Let's try it!🤞🏁

PR #7163, opened 2026-08-11
ui: display last updated information in edit note modal

joecarver:


jessicamcinchak (approved):
Great !

Per prior conversation, I think there's still an open question here of if "Last edited" should refer to _just editing content_ (completed here) and _moving position of note_ (not here, will require db schema change first - then adjust queries here to return both, and just display "max" date?).

Happy for second condition to be picked up in follow-up PR in future though so first piece is still demo-able at S&T tomorrow!

PR #7167, opened 2026-08-11
feat: copy notes along with node copies

joecarver:
Creates new attached notes for the new node when a node is pasted. Also changed the 'Add note' -> 'Attach Note' wording as @ianjon3s suggested.


ianjon3s (approved):
Works great

PR #7168, opened 2026-08-11
feat: submission detail modal and route

zz-hh-aa:
PR shows many lines changed but most are in `routeTree.gen.ts`!

- Creates a new route for the submission detail modal, using established pattern to load child route & navigating back to submissions log when modal closes
- Gets events log per session ID in `SubmissionDetailModal`
- Rough rendering of detail view

Screenshot 2026-08-11 155306


jessicamcinchak (approved):
Great !

PR #7171, opened 2026-08-12
feat: styled grouped submissions log

zz-hh-aa:
This PR:
- Fixes submission log date desc ordering
- Consolidates the status chip, combining event type and status into 5 different chips
    - '[Send event] failed'
    - 'Sending' (eg if Pay was Submitted)
    - 'Invited to pay' or 'Payment in progress' (wanted to preserve both for more transparency / as ITP is a common help-issue)
    - 'Success'
- Adds two custom colour styles for StatusChip

While there is interest in "Paid" status existing _alongside_ rather than being replaced by the "Success" status, I'm thinking of using this single consolidated status for now and we can have multiple statuses as a nice-to-have or additional small ticket later on? Just to avoid too much creep / sprawl!

See [thread](https://opensystemslab.slack.com/archives/C04DZ1NBUMR/p1786539752941109?thread_ts=1783699904.490609&cid=C04DZ1NBUMR) for further context.

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/SubmissionsGrouped.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/SubmissionsGrouped.tsx
@@ -23,6 +23,7 @@ const Submissions: React.FC<SubmissionsProps> = ({ flowSlug }) => {
         ) {
           flowId: flow_id
           id: session_id
+          eventType: event_type
           status: status
           address: address
           eventCreatedAt: created_at
@@ -38,6 +39,7 @@ const Submissions: React.FC<SubmissionsProps> = ({ flowSlug }) => {
   );
 
   const submissions = useMemo(() => data?.submissions || [], [data]);
+
   // filter by flow if flowId prop is passed from route params
   const filteredSubmissions = submissions.filter(
     (submission) => !flowSlug || slugify(submission.flowName) === flowSlug,
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/SubmissionsGrouped.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/SubmissionsGrouped.tsx
@@ -20,8 +20,6 @@ const Submissions: React.FC<SubmissionsProps> = ({ flowSlug }) => {
       query GetSubmissions($team_id: Int!) {
         submissions: submissions_grouped(
           where: { team_id: { _eq: $team_id } }
-          order_by: [{ session_id: asc }, { created_at: desc }]
-          distinct_on: session_id
         ) {
           flowId: flow_id
           id: session_id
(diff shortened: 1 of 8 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/SubmissionsGrouped.tsx
```diff
@@ -20,11 +20,10 @@ const Submissions: React.FC<SubmissionsProps> = ({ flowSlug }) => {
       query GetSubmissions($team_id: Int!) {
         submissions: submissions_grouped(
           where: { team_id: { _eq: $team_id } }
-          order_by: [{ session_id: asc }, { created_at: desc }]
-          distinct_on: session_id
```
zz-hh-aa, line 24: This was a relic from an earlier version (before creating a new view in the db) and was messing up the order by date desc--which the view already does!

apps/editor.planx.uk/src/themeOverrides.d.ts
```diff
@@ -10,6 +10,22 @@ declare module "@mui/material/Chip" {
   }
 }
 
+declare module "@mui/material/Chip" {
+  interface ChipPropsColorOverrides {
```
zz-hh-aa, line 14: First time doing MUI style customisation--tried to follow existing patterns but not totally confident in my approach, feedback welcome :)

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/SubmissionsGrouped.tsx
jessicamcinchak: 💡 good catch !

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+    // "Submitted" status only possible on Pay events, hence "Sending" status
+    return "Sending";
+  } else if (status === "Success") {
+    return "Success";
+  } else if (typeof status === "string" && status.includes("Failed")) {
+    return eventType.replace("(no notification)", "").concat(" failed"); // shortening possible status "Upload to AWS S3 (no notification)"
+  } else {
+    return "Failed";
+  }
+};
```
jessicamcinchak, line 19: 👍

apps/editor.planx.uk/src/themeOverrides.d.ts
jessicamcinchak: Also not an expert with these by any means, but looks good and works as expected for me !

jessicamcinchak (approved):
Code changes and pizza look good, thanks for mock submissions in Barnet for testing ! 

Totally happy to go with single consolidated statuses here for initial launch/refactor, agree something we can always reconsider in future ➕

PR #7172, opened 2026-08-12
chore: drop unused column `flow_note_content.color`

jessicamcinchak:
Per comment here https://github.com/theopensystemslab/planx-new/pull/7131#discussion_r3722737775

"Color" was never used/implemented so we should remove it from the schema and all permissions, queries, and mutations for simplicity prior to any migrations.

Notes are always sticky post-it yellow by default (set as `$noteBackground` in .scss file), the color has _never_ read from the database column - which you can test on this pizza behind feature flag! 
image

### DIFF
--- a/apps/hasura.planx.uk/metadata/databases/default/functions/public_copy_flow.yaml
+++ b/apps/hasura.planx.uk/metadata/databases/default/functions/public_copy_flow.yaml
@@ -2,6 +2,7 @@ function:
   name: copy_flow
   schema: public
 configuration:
+  custom_root_fields: {}
   exposed_as: mutation
 permissions:
   - role: api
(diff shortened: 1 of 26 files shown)


apps/hasura.planx.uk/metadata/databases/default/functions/public_copy_flow.yaml
```diff
@@ -2,6 +2,7 @@ function:
   name: copy_flow
   schema: public
 configuration:
+  custom_root_fields: {}
```
zz-hh-aa, line 5: Just wanted to check my understanding--this change was auto-generated by Hasura and doesn't actually affect anything, is that right?

zz-hh-aa (approved):
Great catch 🔍 note functionality + copying flows etc. works as expected on Pizza!

apps/hasura.planx.uk/metadata/databases/default/functions/public_copy_flow.yaml
jessicamcinchak: Yes exactly ! Still a bit of hard-to-explain inconsistency here between personal machines I think for Hasura auto-generated changes (eg also re-alphabetising etc), but empty `{}` has been auto-generated and should just be harmless default here :heavy_plus_sign:

PR #7173, opened 2026-08-13
fix: File upload info only headings, print button disclaimer

ianjon3s:
## Issue

Addresses two issues present when viewing upload and label in information only mode:
- Headings appear when no information types are present in category
- Print disclaimer reads `This is a record of your responses so far and does not confirm submission.`, which is misleading at this point

image

## Solution
- Make headings conditional on information type being present for that category
- Add an optional `showSubmissionDisclaimer` to use wording around responses so far

image

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/index.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/index.tsx
@@ -141,7 +141,9 @@ export default function Component(props: Props) {
   };
 
   const isCategoryVisible = (category: keyof typeof state.fileList) => {
-    if (props.hideDropZone) return true;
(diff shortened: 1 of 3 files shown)


apps/editor.planx.uk/src/@planx/components/FileUploadAndLabel/Public/index.tsx
```diff
@@ -141,7 +141,9 @@ export default function Component(props: Props) {
   };
 
   const isCategoryVisible = (category: keyof typeof state.fileList) => {
-    if (props.hideDropZone) return true;
+    if (props.hideDropZone) {
+      return state.fileList[category].length > 0;
+    }
```
jessicamcinchak, line 146: Nice tidy up ➕

jessicamcinchak (approved):

PR #7175, opened 2026-08-13
refactor: Integrate patterns to Add new flow form

ianjon3s:
## What does this PR do?

Integrates `Pattern` as a category of flow type, alongside `Flow` and `Service`. This prevents a separate pattern toggle, prevents conflict around `pattern=true` + `source template=true`, and allows conditional wording of the form and button elements.

- Refactor from `isService` as a boolean option to defining a string union type `FlowTypeOption`, which includes `flow`, `service` and `pattern`
- Template toggle only shown for `flow` or `service`, move up to sit with other config options
- Flow name label and add button respond to selected flow type
- Update `BasicRadio` to allow an inline description
- Reorder form so that all options are selected before choosing a name

https://github.com/user-attachments/assets/74a925c5-0688-468e-86f7-2edbdac613fb


DafyddLlyr (approved):
Great!

PR #7176, opened 2026-08-13
fix: status chip styling

zz-hh-aa:
Most changes are generated ones in routetree file again...

This PR moves my StatusChip changes from #7171 into a parallel component, in keeping with general approach to feature flag here--noticed I accidentally affected styling on non-feature-flagged route, eg see screenshot from staging:
Screenshot 2026-08-13 170909


ianjon3s (approved):
Seems like a solid approach, keeps existing chip functional

PR #7177, opened 2026-08-14
prototype: Simplify pattern select to single column

ianjon3s:
## What does this PR do?

As discussed in the design call earlier in the week:
- Prototype of pattern select as a single column, without a "preview" panel on hover


ianjon3s:
Closing as agreed this isn't a direction to pursue

PR #7181, opened 2026-08-14
chore: double `sanitise_application_data` hasura cron timeout

zz-hh-aa:
Previously was the default value of 60, proposing bump because of recent string of failures, see thread: https://opensystemslab.slack.com/archives/C04EEBR1E0P/p1786586483482069

### DIFF
--- a/apps/hasura.planx.uk/metadata/cron_triggers.yaml
+++ b/apps/hasura.planx.uk/metadata/cron_triggers.yaml
@@ -30,6 +30,11 @@
   schedule: 0 2 * * *
   include_in_metadata: true
   payload: {}
+  retry_conf:
+    num_retries: 0
+    retry_interval_seconds: 10
+    timeout_seconds: 120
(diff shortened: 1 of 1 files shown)


apps/hasura.planx.uk/metadata/cron_triggers.yaml
```diff
@@ -30,6 +30,11 @@
   schedule: 0 2 * * *
   include_in_metadata: true
   payload: {}
+  retry_conf:
+    num_retries: 0
+    retry_interval_seconds: 10
+    timeout_seconds: 120
+    tolerance_seconds: 21600
```
zz-hh-aa, line 37: All still default values _except_ for `timeout_seconds`; think they weren't written into the yaml before because all were set to default before.

freemvmt (approved):

DafyddLlyr (approved):

PR #7184, opened 2026-08-14
[chore] get turbo tests working with remote cache

freemvmt:
So I've been trying to use turbo to do all the things in the repo. Everything works except running all test suites, which I want to be able to do via turbo by just running `pnpm test`. 

I first tried to solve logging in so that I could hit the remote cache (and/or contribute to it when I do my own local runs i.e. proper 'multiplayer' mode) - more detail on that [in this comment](https://github.com/theopensystemslab/planx-new/pull/6986#discussion_r3768034740). Have added the relevant commands to the README.

Then I was trying to get all tests to actually run to success, but hit some snags. Most immediately each `pnpm test` command scattered throughout the codebase requires a certain set of conditions to run (`pull-request.yml` is the most obvious reference for these different conditions) e.g. to run `e2e` tests, you have to build the test containers.

But also, not every `pnpm test` was actually fit for purpose, and turbo attempting to run `e2e` as well as the 2 packages it bundles at same time led to race conditions.

On top of that the editor tests were overwhelming my machine (which has reasonable cpu/memory etc, if not excessive), and I wanted to resolve that too, which I come at from various angles.

While working through all this I also took the opportunity to smooth out some of the inconsistencies with managing the `e2e` vs. `dev` docker stacks, e.g. I should be able to bring either down with `pnpm down`.

🥳  With all that said and done, `pnpm tests` + `pnpm test` works a dream!

### An outstanding question 🤔 

Do the [turbo runs in CI context](https://github.com/theopensystemslab/planx-new/blob/main/.github/workflows/pull-request.yml#L82) definitely contribute to the same shared remote cache as we're attempting to interact with here? They read more immediately to me as just making use of the GitHub Action cache rather than connecting to the Vercel API?


apps/editor.planx.uk/src/@planx/components/ContactInput/Editor.test.tsx
```diff
@@ -40,7 +40,8 @@ describe("ContactInputComponent - Editor Modal", () => {
     );
 
     const title = screen.getByPlaceholderText("Title");
-    await user.type(title, "What is your contact info?");
```
freemvmt, line 43: A fair few people seem to have [good mileage](https://github.com/testing-library/user-event/discussions/977#discussioncomment-10974166) with this trick, but perhaps there's a reason we want to stick with `.type`?

apps/editor.planx.uk/src/@planx/components/List/Public/tests/index.payloadGeneration.test.tsx
```diff
@@ -23,108 +23,98 @@ beforeEach(() => {
 });
 
 describe("Payload generation", () => {
-  it(
-    "generates a valid payload on submission (Zoo)",
-    { timeout: 35_000 },
```
freemvmt, line 28: just removing these selective `35_000` timeouts since that's now the global value

apps/editor.planx.uk/vitest.config.ts
```diff
@@ -7,10 +7,15 @@ export default defineConfig({
     globals: true,
     clearMocks: true,
     retry: 1,
+    // Interaction-heavy jsdom tests are wall-clock sensitive and can vary wildly by machine
+    // due to e.g. cores, processes, etc. So we are deliberately generous with timeouts.
+    testTimeout: 35_000,
```
freemvmt, line 12: `35_000` was the maximum timeout we had programmed in for problematic tests, so seems a good place to start here?

e2e/tests/api-driven/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@planx/e2e-api",
   "private": true,
   "scripts": {
-    "test": "cucumber-js --tags 'not @regression'",
+    "test:run": "cucumber-js --tags 'not @regression'",
```
freemvmt, line 5: this ensures that the two separate `e2e` test suites (UI vs. API) are not invoked independently by turbo, but rather belong to the overall `e2e` test suite

e2e/tests/ui-driven/package.json
```diff
@@ -13,15 +13,15 @@
...
     "graphql": "catalog:graphql",
+    "graphql-request": "^6.1.0",
     "isomorphic-fetch": "^3.0.0",
     "jsonwebtoken": "catalog:jwt",
     "serve": "^14.2.6",
     "uuid": "catalog:"
   },
   "devDependencies": {
-    "@playwright/test": "^1.58.2",
+    "@playwright/test": "^1.62.1",
```
freemvmt, line 24: this bump brings support for Ubuntu 26.04 LTS (i.e. me)

scripts/start-containers-for-dev.sh
```diff
@@ -35,7 +35,7 @@ function setupContainers(){
   echo "Starting docker…"
 
   # Bring dev containers back up, preserving existing volumes so local data changes are not lost.
-  # For a clean first-time setup, use `pnpm up` instead.
+  # For a clean first-time setup (including seeding the db), use `pnpm run up` instead.
```
freemvmt, line 38: emphasise that this script *does not* run `sync-data`, as we seemed to suggest before (here and in README)

.gitignore
```diff
@@ -48,7 +48,9 @@ __pycache__
 # Tooling
 .aider*
 .claude
-.turbo/
 
 # Output of turbo prune
 /out
+
+# Turborepo
```
freemvmt, line 55: turbo wants to do this automatically on some invocations so I just let it :)

package.json
```diff
@@ -15,10 +15,13 @@
...
     "logs": "docker compose logs --tail 30 -f",
     "prepare": "husky",
+    "prebuild": "./scripts/check-containers.sh dev",
+    "pretest": "./scripts/check-containers.sh e2e",
     "typecheck": "turbo run typecheck",
     "lint": "turbo run lint",
     "check": "turbo run typecheck lint",
-    "build": "turbo run build"
+    "build": "turbo run build",
+    "test": "turbo run test --concurrency=1"
```
freemvmt, line 24: enforce running 1 test suite/package at a time to avoid overwhelming computer (bearing in mind that we are necessarily already running the test docker stack, and running e2e + editor tests on top of that is **heavy**)

README.md
```diff
@@ -106,10 +106,18 @@ The root of the project has several scripts set up to help you manage your docke
 - `pnpm typecheck` - type-check every workspace (`tsc --noEmit` / `astro check`)
 - `pnpm lint` - ESLint + Prettier checks (Editor, API, E2E)
 - `pnpm check` - `typecheck` and `lint` together
-- `pnpm build` - build the Editor, API and LPS
+- `pnpm build` - build the Editor, API and LPS (run `pnpm run up` first so LPS can hit Hasura)
+- `pnpm test` - test all packages (run `pnpm tests` first to spin up test containers)
 
 Turbo caches by content hash, so re-running an unchanged task is near-instant (`>>> FULL TURBO`). This can be scoped to one package with `--filter` (e.g. `pnpm turbo run build --filter=api.planx.uk`).
 
+Note that to take advantage of this 'multiplayer' cache shared between local and CI runs, you have to login using the `TURBO_TOKEN` secret, and link your local environment to the remote cache (hosted on Vercel):
```
freemvmt, line 114: Notes for logging in to remote cache from local, as discussed [here](https://github.com/theopensystemslab/planx-new/pull/6986#discussion_r3795602296)

turbo.json
```diff
@@ -1,4 +1,8 @@
 {
+  "remoteCache": {
+    "teamId": "team_7liLUNk0ekSOy74MY60qmGD3",
```
freemvmt, line 3: not a secret, just the ID version of our team slug (turbo adds automatically when you login/link)

freemvmt (commented):
Commenting this up by way of explanation 🔧

apps/editor.planx.uk/src/@planx/components/ContactInput/Editor.test.tsx
DafyddLlyr: We've made this change a few times for the performance wins - I think there's no issue switching away from `.type()` if there's not a specific reason we're testing a typing interaction (e.g. search debounce or something).

apps/editor.planx.uk/vitest.config.ts
DafyddLlyr: Works for me!

.env.example
DafyddLlyr, line 1: Thanks for this - I'm sure I had this here at some point but it must have been overwritten / not commited 🤔

pnpm-workspace.yaml
```diff
@@ -9,10 +9,10 @@ packages:
   - scripts/*
 
 catalog:
-  '@opensystemslab/planx-core': github:theopensystemslab/planx-core#9ecd850
-  '@types/jsdom': ^28.0.3
-  '@types/node': ~24.10.15
-  '@vitest/eslint-plugin': ^1.6.20
+  "@opensystemslab/planx-core": github:theopensystemslab/planx-core#48cfed3ca5a1387f3be30cb1f0712284a629cb33
```
DafyddLlyr: I think this is now out of date with `main`. If you bump to the lastest `main` commit that would be helpful as we'll bring in a bunch of dependabot fixes - https://github.com/theopensystemslab/planx-core/commits/main/

DafyddLlyr (approved):
Thanks for this - all looking good! Only small question is around the `planx-core` hash used which I think has now drifted.

> _Do the [turbo runs in CI context](https://github.com/theopensystemslab/planx-new/blob/main/.github/workflows/pull-request.yml#L82) definitely contribute to the same shared remote cache as we're attempting to interact with here? They read more immediately to me as just making use of the GitHub Action cache rather than connecting to the Vercel API?_

My understanding is yes - it's a shared cache via the Vercel API. We can test this by - 
 - busting the cache (e.g. adding a newline, console.log etc)
 - building
 - re-building elsewhere, seeing "remote cache hit"

The tricky thing is env vars - local env vars can differ from CI, so a local artifact doesn't always match a remote artifact, triggering a rebuild not a hit.

.env.example
freemvmt: things do get lost haha! luckily we have convergence :saluting_face:

pnpm-workspace.yaml
```diff
@@ -9,10 +9,10 @@ packages:
   - scripts/*
 
 catalog:
-  '@opensystemslab/planx-core': github:theopensystemslab/planx-core#3ecedcd
-  '@types/jsdom': ^28.0.3
-  '@types/node': ~24.10.15
-  '@vitest/eslint-plugin': ^1.6.20
+  "@opensystemslab/planx-core": github:theopensystemslab/planx-core#fd0d1c4aa3879f1b73afef7ef7eb27579a0d2acb
```
freemvmt, line 12: bumping `planx-core` to [latest](https://github.com/theopensystemslab/planx-core/commits/main/)

freemvmt:
> The tricky thing is env vars - local env vars can differ from CI, so a local artifact doesn't always match a remote artifact, triggering a rebuild not a hit.

@DafyddLlyr yes good point about env, I did wonder if the subtle differences between local/CI runs would mean they would never share cache hits. But obviously the CI cache is the main win anyways.

PR #7188, opened 2026-08-18
feat: Sanitise user-derived values injected into GovNotify emails

DafyddLlyr:
## What does this PR do?
This PR adds a `sanitiseContentFor` property to the `NotifyConfig` type. We then populate this, per-template, with a list of user-derived properties (values entered by end-users when completing a PlanX application).

Please see Trello card for full write up of the problem - https://trello.com/c/j7sBXNxL/3735-sanitise-all-user-generated-content-displayed-in-govnotify-emails

## Testing
Enter a value which should be sanitised, trigger an email - content is sanitised and not rendered as a link (full workflow listed on Trello card above).

image

image


zz-hh-aa (approved):
Worked as expected, thanks for writeup context!

PR #7190, opened 2026-08-18
feat(editor): Improve pattern counts to include clones and nested flows

DafyddLlyr:
## What does this PR do?
Updates `getComponentCount()` to an improved `getPatternCounts()` helper which - 
 - Counts nested flows separately
 - Correctly handles clones
 - Still ignores `_root` and `Option` nodes

image


ianjon3s (approved):
Works great, appreciate the clearness of the narrative provided by tests

PR #7191, opened 2026-08-18
style: grouped submissions log detail modal

zz-hh-aa:
This PR handles much of the styling for the (still feature flagged!) grouped submissions log submission detail modal. It:
- Creates `SubmissionDetails` and `SubmissionEventsHistory` components for the two main info panels in the modal
    - Creates a `StatusIcon` for each event in `SubmissionEventsHistory`
    - In `SubmissionDetails`, fetches the team logo with new query / hook

Before this PR:
image

After:
image

There are still some style nits to work through but I thought I would separate the PRs here. Next is the bulk of button functionality:
- View submission (and route)
- Download submission
- Resubmit buttons for failed events
- Close modal

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/SubmissionDetailModal.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/SubmissionDetailModal.tsx
@@ -4,6 +4,7 @@ import Button from "@mui/material/Button";
 import Dialog from "@mui/material/Dialog";
 import DialogActions from "@mui/material/DialogActions";
 import DialogContent from "@mui/material/DialogContent";
+import Grid from "@mui/material/Grid";
 import ListItem from "@mui/material/ListItem";
 import Typography from "@mui/material/Typography";
 import { useNavigate } from "@tanstack/react-router";
@@ -11,6 +12,9 @@ import DelayedLoadingIndicator from "components/DelayedLoadingIndicator/DelayedL
 import { useStore } from "pages/FlowEditor/lib/store";
 
 import type { Submission } from "../types";
+import { DownloadSubmissionButton } from "./DownloadSubmissionButton";
+import { SubmissionDetails } from "./SubmissionDetails";
+import { ViewSubmissionButton } from "./ViewSubmissionButton";
 
 const GET_SUBMISSION_EVENTS = gql`
   query GetSubmissionEvents($sessionId: uuid!) {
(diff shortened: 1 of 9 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/SubmissionDetailModal.tsx
```diff
@@ -56,50 +60,50 @@ const SubmissionDetailModal: React.FC<SubmissionDetailModalProps> = ({
...
+                to: "/app/$team/submissions",
+                params: {
+                  team: teamSlug,
+                },
+              })
+            }
+          >
+            Back
+          </Button>
+        </DialogActions>
```
DafyddLlyr, line 105: This should be a sibling of `DialogContent`, not a child. This is why it's a nested element in the PR screenshot, not a full length element like other modals.

See https://mui.com/material-ui/react-dialog/

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/queries.ts
```diff
@@ -0,0 +1,12 @@
+import { gql } from "@apollo/client";
+
+export const GET_TEAM_LOGO = gql`
```
DafyddLlyr, line 3: Returning id values here will allow us to get this from the Apollo cache if available, we might just need to rejig the query to go team -> theme -> logo, and not theme -> logo.

Having a cachable key (id) will also save us re-fetching on each render (e.g. open details modal, 1st fetch, close and re-open, 2nd fetch...).

DafyddLlyr (changes requested):
Really solid..! Nice sized PR, code all very clear and concise.

A few small suggestions below, no showstoppers!

It would also be nice to have this new method available as a Story - it should just be the case of setting the flag in `window.localStorage` (maybe as part of the `loader` function?). We used to have a few stories which did this but I don't think there are any live currently. Not a deal breaker if not a simple one-liner..!

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/queries.ts
zz-hh-aa: Good thinking! Done in [78c2598](https://github.com/theopensystemslab/planx-new/pull/7191/commits/78c25988379653aceba7a5d1e873f7d54e6f2233)

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/SubmissionDetailModal.tsx
zz-hh-aa: Ah thanks 🔍 my next PR is all of the buttons for the modal, so I think I'd like to fix this one forward / handle it there, if that's alright?

DafyddLlyr: Yep no problem 👌

DafyddLlyr (approved):
Great..!

PR #7192, opened 2026-08-18
fix(editor): Setup `onError` fallback for logo component

DafyddLlyr:
## What's the problem?
If a logo URL errors (4xx, 5xx) then it just shows a broken image icon, and does not fall back to plain text.

This was raised here during a GitHub outage - https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1787063943428389

## What's the solution?
Use the existing `onError` prop to handle the fallback ([docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#image_loading_errors)).

## Testing
Either update the Pizza DB to have a broken logo URL for a given team, or block the URL via dev tools. The header will fall back to using the plain text team name -

image


ianjon3s (approved):
Working nicely, thanks for picking this up

PR #7193, opened 2026-08-18
fix(editor): Spacing and layout for Question and Checklist text inputs

DafyddLlyr:
## What's the problem?
Spacing is missing beneath the text / title inputs on Question and Checklist components.

## What's the cause?
This was introduced here when an `ErrorWrapper` was added for the new error condition - https://github.com/theopensystemslab/planx-new/pull/7127

A new wrapping element broke the `:last-child` styling, and the `marginBottom` was not applied as a result.

## What's the solution?
Reach for the existing `errorMessage` prop instead of using a wrapper - the `Input` element has a built in `ErrorWrapper` already.

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/apps/editor.planx.uk/src/ui/shared/InputRow.tsx
+++ b/apps/editor.planx.uk/src/ui/shared/InputRow.tsx
@@ -7,7 +7,7 @@ const Root = styled(Box, {
   shouldForwardProp: (prop) => prop !== "childRow",
 })<BoxProps & { childRow?: boolean }>(({ childRow }) => ({
   display: "flex",
-  alignItems: "center",
+  alignItems: "flex-end",
   width: "100%",
   "&:not(:last-child)": {
     marginBottom: 5,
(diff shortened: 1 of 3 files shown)


apps/editor.planx.uk/src/ui/shared/InputRow.tsx
```diff
@@ -7,7 +7,7 @@ const Root = styled(Box, {
   shouldForwardProp: (prop) => prop !== "childRow",
 })<BoxProps & { childRow?: boolean }>(({ childRow }) => ({
   display: "flex",
-  alignItems: "center",
+  alignItems: "flex-end",
```
DafyddLlyr, line 10: This change is required to keep the `ImgInput` aligned. I've checked all instances of this I can find and this change doesn't seems to have a negative effect, but I'd appreciate a quick cross-check if possible (cc @ianjon3s ?)

Here's what it looks like if this value remains as `center` - 

Image

ianjon3s: As this is a shared component (editor/public) I've run through the Accessibility 2026 flow locally, and I can't see any issues. There are however numerous places where `<InputRow>` is used unnecessarily, I've made a note to go through a remove these.

ianjon3s (approved):
Tested locally and fixes the issue. Public interface tested but I've not given the editor a complete run-through - let's keep an eye on this.

PR #7194, opened 2026-08-18
feat: Abstract mosaic pattern for each pattern

ianjon3s:
## What does this PR do?

Produces an abstract mosaic grid pattern for each pattern:
- A grid pattern is deterministic and unique to each pattern
- Uses a colour palette in variance of the lime green used for patterns
- Each component produces a 1x1 square, nested flows produce a 2x2 square
- Grid scales depending on number of elements present

https://github.com/user-attachments/assets/4a7dfca4-df78-4d9c-a29f-0583c098eecc

### DIFF
--- a/apps/editor.planx.uk/src/components/DelayedLoadingIndicator/DelayedLoadingIndicator.tsx
+++ b/apps/editor.planx.uk/src/components/DelayedLoadingIndicator/DelayedLoadingIndicator.tsx
@@ -24,7 +24,7 @@ const SpinnerRoot = styled(Box, {
   }),
 }));
 
-const EllipsesText = styled(Typography)(() => ({
+export const EllipsesText = styled(Typography)(() => ({
   "&::after": {
     display: "inline-block",
     animation: "ellipsis steps(1,end) 1s infinite",
--- a/apps/editor.planx.uk/src/components/DelayedLoadingIndicator/DelayedLoadingIndicator.tsx
+++ b/apps/editor.planx.uk/src/components/DelayedLoadingIndicator/DelayedLoadingIndicator.tsx
@@ -2,6 +2,7 @@ import Box from "@mui/material/Box";
 import CircularProgress from "@mui/material/CircularProgress";
 import { styled } from "@mui/material/styles";
 import Typography from "@mui/material/Typography";
+import { EllipsesText } from "components/EllipsesText/EllipsesText";
 import React, { useEffect, useState } from "react";
 
 export interface Props {
@@ -22,21 +23,6 @@ const SpinnerRoot = styled(Box, {
   ...(inline && {
     padding: 0,
   }),
-}));
-
-export const EllipsesText = styled(Typography)(() => ({
-  "&::after": {
-    display: "inline-block",
-    animation: "ellipsis steps(1,end) 1s infinite",
-    content: '""',
-  },
-  "@keyframes ellipsis": {
-    "0%": { content: '""' },
-    "25%": { content: '"."' },
-    "50%": { content: '".."' },
-    "75%": { content: '"..."' },
-    "100%": { content: '""' },
-  },
 }));
 
 const DelayedLoadingIndicator: React.FC<Props> = ({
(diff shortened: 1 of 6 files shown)


apps/editor.planx.uk/src/components/DelayedLoadingIndicator/DelayedLoadingIndicator.tsx
DafyddLlyr, line 1: As this is now shared and used elsewhere it's worth splitting into it's own file. A dev could make a change here without realising it will have impacts elsewhere.

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/PatternMosaic.tsx
```diff
@@ -0,0 +1,62 @@
...
+const CELL_SIZE = 36;
+const GAP = 0.12;
+
+interface Props {
+  seed: string;
+  counts: PatternCounts;
+  sx?: SxProps<Theme>;
+}
+
+// Renders a pattern's mosaic as a self-contained SVG
```
DafyddLlyr: nit: These should be JSDocs not plain comments. Making them JSDocs will mean they're readable via IDEs etc.

I'd keep `//` comments for in-line comments, not top-level function/variable explanations which may be needed outside the context of a single file.

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/utils.ts
```diff
@@ -30,3 +30,142 @@ export const getPatternCounts = (graph: Graph): PatternCounts => {
...
+// Deterministic string hash, used to seed the mosaic's PRNG per-pattern
+const hashString = (value: string): number => {
+  let hash = 0;
+  for (let i = 0; i < value.length; i++) {
+    hash = (hash * 31 + value.charCodeAt(i)) | 0;
+  }
+  return hash >>> 0;
+};
+
+// Sequence generator using mulberry32 forumla to produce a pattern based on each seed
```
DafyddLlyr: Same JSDoc comment as above for most the functions in this file 👍

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/utils.ts
```diff
@@ -30,3 +30,142 @@ export const getPatternCounts = (graph: Graph): PatternCounts => {
...
+  colorIndex: number;
+}
+
+export interface Mosaic {
+  cols: number;
+  rows: number;
+  squares: MosaicSquare[];
+}
+
+const PALETTE_SIZE = 10;
```
DafyddLlyr: What happens if this drifts from the `.length` of `MOSAIC_PALETTE`?

DafyddLlyr (approved):
A few very minor comments. Code looks good, can't test as the Pizza is currently down. Approving to unblock this, will test on staging once merged or on Pizza if I catch it up before the merge.

apps/editor.planx.uk/src/components/DelayedLoadingIndicator/DelayedLoadingIndicator.tsx
ianjon3s: Dedicated file created for the ellipses element 👍

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/PatternMosaic.tsx
ianjon3s: Noted, I've converted all top-level comments to JSDoc and will stick to this format

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/PatternsTab/utils.ts
ianjon3s: Good catch, I've now calculated using `.length` to keep in sync

DafyddLlyr (approved):

ianjon3s:
Thanks @DafyddLlyr , will merge once CI passes

PR #7196, opened 2026-08-18
fix: Allow settings tabs to scroll

ianjon3s:
## Issue

As reported in Slack, settings tabs overflow the container and scroll is not enabled:
https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1786094923420729

## Solution

- Enable scroll on tabs

https://7196.planx.pizza/app/barnet/apply-for-a-lawful-development-certificate/settings/visibility

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/SettingsLayout.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/SettingsLayout.tsx
@@ -1,9 +1,10 @@
 import Box from "@mui/material/Box";
 import Container from "@mui/material/Container";
-import { styled } from "@mui/material/styles";
+import { styled, useTheme } from "@mui/material/styles";
 import type { SvgIconProps } from "@mui/material/SvgIcon";
 import Tabs, { tabsClasses } from "@mui/material/Tabs";
 import Typography from "@mui/material/Typography";
+import useMediaQuery from "@mui/material/useMediaQuery";
 import { useNavigate } from "@tanstack/react-router";
 import React from "react";
 import { useLocation } from "react-use";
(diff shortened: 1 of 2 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/SettingsLayout.tsx
```diff
@@ -78,6 +78,8 @@ const SettingsLayout: React.FC<Props> = ({
           {settingsLinks && (
             <TabList>
               <Tabs
+                variant="scrollable"
+                scrollButtons={false}
```
DafyddLlyr: Agree with no scroll buttons, but we might need something here as a visual indicator - often a fade or glow us used to indicate "more". There's a change that a configuration of labels lines up pretty well at a given screen size and perfectly hides an option.

In the screenshot below there's no indication that a tab is cut off - 

Image

DafyddLlyr (changes requested):
https://github.com/user-attachments/assets/e49393a3-eb49-4628-a04a-d50066b58279

Hitting some weird interactions here on first load - the scroll doesn't seem to kick in correctly on first pass. There's also an odd use of the loading spinner here which isn't consistent with other pages (we can fix here, ignore for now, or fix this as a follow up).

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/SettingsLayout.tsx
ianjon3s: I've spent a bit more time with this, agree that having a blind overflow is less than ideal.

Steps I've taken:
- Reduce (what is currently quite generous) padding/margin slightly so that current tabs fit within viewport
- Remove icons at a smaller breakpoint so that tabs do not overflow until smaller screens
- Introduce scrolling only at smaller screens where we know an overflow will happen

This does put emphasis on us not introducing any more tabs, but I think if we do find we need more, having an overflow menu that "swallows up" tabs (similar to GitHub's example below) will be a strong solution to look at

image

DafyddLlyr (approved):
This is great...! Thanks for taking the extra time here - the outcome now is much strong imho 👌

PR #7202, opened 2026-08-19
chore: Refactor GovPay business logic into hook, away from UI

DafyddLlyr:
## What does this PR do?
Moves all GovPay business logic out the the UI component, and into a new `useGovPay()` hook. There are zero functional changes here, purely organisational - no unit tests or E2E tests updated at all.

## Context
As part of the Stripe migration I've been looking at how we can share the UI, but offer different functionality based on a team's payment provider. Originally I did think of having a high level conditional check that toggled between GovPay and Stripe but decided against this as - 
 - The fundamental control flow will remain the same for creating a payment. Keeping the current robust setup is a nice win I think?
 - There's a bunch of shared UI
 
 ## Next steps...
 Make a placeholder `useStripe()` hook, toggle based on feature flag/payment provider.

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/Pay/Public/providers/types.ts
+++ b/apps/editor.planx.uk/src/@planx/components/Pay/Public/providers/types.ts
@@ -1,0 +1,14 @@
+export type PaymentProviderName = "govpay" | "stripe";
+
+export interface PaymentActions {
+  startNewPayment: () => Promise<void>;
+  refetchPayment: () => Promise<void>;
+  resumeExistingPayment: () => Promise<void>;
+  handleSuccess: () => void;
+}
+
+export interface UsePaymentProviderResult {
+  actions: PaymentActions;
+  passportKey: string;
+  hasExistingPayment: boolean;
+}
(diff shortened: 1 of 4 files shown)


apps/editor.planx.uk/src/@planx/components/Pay/Public/providers/types.ts
```diff
@@ -0,0 +1,14 @@
+export type PaymentProviderName = "govpay" | "stripe";
```
jessicamcinchak, line 3: 🔮 !

apps/editor.planx.uk/src/@planx/components/Pay/Public/providers/types.ts
```diff
@@ -0,0 +1,14 @@
...
+export interface PaymentActions {
+  startNewPayment: () => Promise<void>;
+  refetchPayment: () => Promise<void>;
+  resumeExistingPayment: () => Promise<void>;
+  handleSuccess: () => void;
+}
+
+export interface UsePaymentProviderResult {
+  actions: PaymentActions;
+  passportKey: string;
```
jessicamcinchak: Should this actually be strict type and not generic `string;` ?

jessicamcinchak (approved):
Great !

PR #7199, opened 2026-08-19
fix(api): Instantiate single OS proxy at module level

DafyddLlyr:
## What's the problem?
When looking at logs this morning for API downtime there's a bunch of OS proxy instantiation logs (as usual). These seem to align with CPU spikes in AWS CloudWatch.

I'm pretty unsure if this is the issue but it seems worth a fix / worthwhile to rule out, but reading through the `http-proxy-middleware` docs shows we're using a bit of an anti-pattern here. Rather than instantiate a proxy per request (of which there will be very many when panning for map tiles for example), we can instantiate once at module level.

Testing locally shows no issues hitting OS APIs, and much quieter logs when panning the map for example.

The below logs are generated when panning or zooming on the map (plus a whole lot more) - a new proxy per request.

```sh
2026-08-19 13:01:41 [HPM] Proxy created: /  -> https://api.os.uk
2026-08-19 13:01:41 [HPM] Subscribed to http-proxy events: [ 'error', 'proxyRes', 'close' ]
2026-08-19 13:01:41 [HPM] GET /proxy/ordnance-survey/maps/vector/v1/vts/resources/styles?srs=3857 ~> https://api.os.uk
2026-08-19 13:01:41 [HPM] Proxy created: /  -> https://api.os.uk
2026-08-19 13:01:41 [HPM] Subscribed to http-proxy events: [ 'error', 'proxyRes', 'close' ]
2026-08-19 13:01:41 [HPM] GET /proxy/ordnance-survey/maps/vector/v1/vts/tile/19/174109/261220.pbf?srs=3857 ~> https://api.os.uk
2026-08-19 13:01:41 [HPM] Proxy created: /  -> https://api.os.uk
2026-08-19 13:01:41 [HPM] Subscribed to http-proxy events: [ 'error', 'proxyRes', 'close' ]
2026-08-19 13:01:41 [HPM] GET /proxy/ordnance-survey/maps/vector/v1/vts/tile/19/174108/261220.pbf?srs=3857 ~> https://api.os.uk
2026-08-19 13:01:41 [HPM] Proxy created: /  -> https://api.os.uk
2026-08-19 13:01:41 [HPM] Subscribed to http-proxy events: [ 'error', 'proxyRes', 'close' ]
2026-08-19 13:01:41 [HPM] GET /proxy/ordnance-survey/maps/vector/v1/vts/tile/19/174108/261221.pbf?srs=3857 ~> https://api.os.uk
[...]

PR #7208, opened 2026-08-20
chore(api): Reduce JWT validaiton noise in CloudWatch logs

DafyddLlyr:
## What does this PR do?
Reduced noise in AWS CloudWatch logs (all API logs in fact) by changing calls to `console.error()` with `console.debug()`.

Right now AWS CloudWatch logs are polluted with these logs, which appear genuine to me - over a period yesterday morning (see https://github.com/theopensystemslab/planx-new/pull/7199) there were large numbers of calls from a small number of JWTs (same `expiresAt` timestamp) which looks like a few logged out users.

## Next steps...
Once deployed to `main` and `production`, sense check logs over the next few days 👌


freemvmt (approved):
genuinely a great and very needed edit tysm 👏

PR #7210, opened 2026-08-20
fix(editor): Configure more robust fetch policy for `usePatterns()`

DafyddLlyr:
## What does this PR do?
Change the `fetchPolicy` to `"cache-and-network"` for the `usePatterns()` hook. This ensures that we always check the "copiable" status of the pattern list before allowing users to insert them. Without this change it's possible for the cache to contain stale items - allowing users to insert patterns no longer "copiable" (until the next refresh).

Please see https://opensystemslab.slack.com/archives/C5Q59R3HB/p1787215523036939?thread_ts=1787209920.554459&cid=C5Q59R3HB (OSL Slack) for further context.


zz-hh-aa (approved):
Works as expected!

PR #7214, opened 2026-08-20
fix: sync `flow_note_*` tables from prod db in correct order

freemvmt:
As per title - see [Slack](https://opensystemslab.slack.com/archives/C04EEBR1E0P/p1787199487798469).


jessicamcinchak (approved):
Thanks for fixing this !

PR #7217, opened 2026-08-20
Production deploy

DafyddLlyr:
We can maybe get the Newcastle SSL cert work here also.


freemvmt (approved):

freemvmt:
> We can maybe get the Newcastle SSL cert work here also.

Yes that would be great - cutover PR is at #7223 :)

DafyddLlyr:
@freemvmt I've gone ahead and deployed this as I can see the Newcastle thread is still ongoing, and I'm keen to get patterns across to production.

We can still have another rollout later today without any issues 👌

PR #7218, opened 2026-08-20
fix: truncate flow notes as part of `reset_flows` action

freemvmt:
#7214 did not fix nightly db sync because that script uses the `reset_flows` option and the `TRUNCATE` command to `flows` table does cascade through to because they are essentially disconnected.

That is, neither `flow_note_content` or `flow_note_positions` has an FK to flows.

Therefore we add an additional line here to explicitly truncate the former, which will then cascade to the latter.

See also [here](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1787226339748609) and [here](https://opensystemslab.slack.com/archives/C5Q59R3HB/p1787239070911579) on Slack.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ The root of the project has several scripts set up to help you manage your docke
 - `pnpm recreate` will build and (re)start your docker containers from scratch
 - `pnpm destroy` will remove volumes (i.e. database data) for your docker containers (dev or e2e) and can be a useful hard reset when necessary
 - `pnpm sync-data` will sync production records with modified data in your database
+- `pnpm test-sync` emulates the command performed by the nightly staging sync (see `sync-staging-db.yml`)
 - `pnpm clean-data` will sync production records and reset any modified data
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -93,6 +93,7 @@ The root of the project has several scripts set up to help you manage your docke
 - `pnpm recreate` will build and (re)start your docker containers from scratch
 - `pnpm destroy` will remove volumes (i.e. database data) for your docker containers (dev or e2e) and can be a useful hard reset when necessary
 - `pnpm sync-data` will sync production records with modified data in your database
+- `pnpm test-sync` emulates the command performed by the nightly staging sync (see `sync-staging-db.yml`)
```
jessicamcinchak, line 96: Hadn't realised this was skipped in README, good catch 👍

jessicamcinchak (approved):
Makes sense! `reset_flows` conceptually now inclusive of "notes" too ➕

DafyddLlyr (approved):

PR #7222, opened 2026-08-20
[chore] update db sync workflow to use node v24

freemvmt:
Should resolve the following warning seen in latest runs of Nightly staging database sync action. 

> Warning: Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: andstor/file-existence-action@v2, pulumi/setup-pulumi@v2. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/


jessicamcinchak (approved):
Thanks !

PR #7201, opened 2026-08-19
feat: submission detail modal buttons

zz-hh-aa:
# What's in this PR? 
New button
- Close modal button
    - Refactors / shares `CloseButton` now (there were a couple similar close buttons that existed elsewhere)

Versions of existing buttons; will have `Grouped` suffix removed as a last step / when feature flag comes off
- DownloadSubmissionButtonGrouped
- ViewSubmissionButtonGrouped
- OpenResponseButtonGrouped
- ResubmitButtonGrouped

# Testing
- In the Testing team, I've made a simple submission with some mock applications, please switch the `Grouped_submissions` feature flag on and then view and test modals, buttons, etc.: https://7201.planx.pizza/app/testing/submissions

Locally, I have tested submissions log buttons with feature flag off to ensure no regressions too. 

Screenshot 2026-08-19 130019

# Next steps
- More manual testing
- Some refactoring / tidying
- Tests!
- Stories
- Remove feature flag and rename / delete parallel features

PR #7227, opened 2026-08-21
feat: parallel routes for $team/$flow/submissions and $team/submissions

zz-hh-aa:
Follow up to #7201 

# What's in this PR?
This PR sets up parallel submission routes for `$team/submissions?detail=$sessionId` and `$team/$flow/submissions?detail=$sessionId`. Missing parallel route was highlighted by Daf in #7201 and suggestion for this route is below. The `?detail=$sessionId` route renders the detail modal, and `submissions/$sessionId` is the full page `SubmissionHTML`.

I realised that the issue being bounced out of `$flow` (eg from `$team/$flow/submissions`) back up to `$team` was afflicting the existing view submission route. 

```mermaid
graph TD;
    team["$team"] --> flow["$flow"];
    team --> submissions;
    flow --> submissions;
    submissions --> sessionId["$sessionId<br/>`SubmissionHTML`"]
    submissions --> detail["?detail=$sessionId<br/>`SubmissionDetailModal`"];

```

# Testing
- Turn feature flag on!
- Try `$team/$flow/submissions` route: https://7227.planx.pizza/app/testing/simple-submission-oz/submissions
    - Click on application row
    - Close modal
    - Re-open application row and try "View submission"
    - Confirm that all routes work as expected and are still within `$flow/`
- Try `$team/submissions` route: https://7227.planx.pizza/app/testing/submissions
    - Click on application row
    - Close modal
    - Re-open application row and try "View submission"
    - Confirm that all routes work as expected
- Can double check that existing, non-feature flagged routes work as expected


apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/SubmissionDetailModal.tsx
```diff
@@ -50,8 +50,28 @@ const SubmissionDetailModal: React.FC<SubmissionDetailModalProps> = ({
   sessionId,
 }) => {
   const navigate = useNavigate();
+  const params = useParams({ strict: false });
   const [teamSlug] = useStore((state) => [state.teamSlug]);
 
+  const handleClose = () => {
```
DafyddLlyr: I think we can just `navigate({ to: ".." })` here to go back a step? This should be type-safe.

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/SubmissionsGrouped.tsx
```diff
@@ -44,7 +44,7 @@ const SubmissionsGrouped: React.FC<SubmissionsProps> = ({ flowSlug }) => {
   const filteredSubmissions = submissions.filter(
     (submission) => !flowSlug || slugify(submission.flowName) === flowSlug,
   );
-
+  console.log({ flowSlug });
```
DafyddLlyr: 👀

DafyddLlyr (changes requested):
> _The actual submission view had to be rendered as a sibling of /submissions and not a child as intended (eg `$team/submissions/$sessionId/view`). While I could get the navigation to work as expected, I couldn't get the rendering up--I think this is because we're using nested routes and the Outlet component, which acts as a placeholder for the child route / will render all ancestors too._

Just had a play around with this and it is solvable by some pretty convoluted routing / matching - please see https://github.com/theopensystemslab/planx-new/commit/e407052116950e593c58e023fb6b12c4f339eaf7

To get this working we need to change the route tree so the submissions table lives in the layout (`route.tsx`) and persists across all child routes using `<Outlet />`.

I'm not entirely sure this complexity is worth it - have a play around and let me know wha you think. It might actually be simpler to have the modal be `/submissions?detail=$sessionId` and then the full page be `/submissions/$sessionId` or something.

This would also solve some state / loading issues on this current iteration - we get a weird "Loading..." panel on the right hand side of the page when opening the modal, and the table re-renders on close (flash of white).

My changes above fix this by setting `pendingMs: Infinity` on the layout, so that it never hits a loading state on return and using `pendingComponent: () => null` on the `$sessionId` layout. Both are sort of work-arounds that _I think_ the query params URL will fix natively.

---

So strong feelings on the ultimately approach we decide on here, but at a minimum we should fix the two loading states on modal open/close 👍

zz-hh-aa:
Thanks @DafyddLlyr! I checked out your commit and had a go at implementing your suggestion: 
> It might actually be simpler to have the modal be /submissions?detail=$sessionId and then the full page be /submissions/$sessionId or something.

Restructure in [5c33cca](https://github.com/theopensystemslab/planx-new/pull/7227/commits/5c33cca48e638ec6d0461ec03ae005770b535881); good shouts on the simpler routes. Routes seem to be working and avoid the awkward loading state--inclined to go with this option, if it works for you! PR description also adjusted accordingly.

DafyddLlyr (changes requested):
Agree the modal is much smoother on open close - lets stick with this..!

A few minor comments - 

1. The loader is still present, I think my comment above caught how to avoid this. Throttling network here to exaggerate the effect.

https://github.com/user-attachments/assets/9df63715-0244-40ec-8d6b-78262b0cfcac

2. Can't close modal via `Esc` key or `backdropClick` as with other ephemeral modals. This might be because it's it's own route. Not a showstopper but worth looking into.

3. Placement of closing X button - still seems a little off (see Ian's previous comment on this).

zz-hh-aa:
Thank @DafyddLlyr! 

> 1. The loader is still present, I think my comment above caught how to avoid this. Throttling network here to exaggerate the effect.

I tried out `pendingMs` and `pendingComponent: () => null` but didn't have any luck, changed what gets returned in the component instead (`Dialog` with `DelayedLoadingIndicator` in it) which has gotten rid of the side panel--further style thoughts welcome (sorry still kinda quick even with throttling)

https://github.com/user-attachments/assets/e7aec221-9b4f-4116-891c-4a5a07a75cd4

> 2. Can't close modal via Esc key or backdropClick as with other ephemeral modals. This might be because it's it's own route. Not a showstopper but worth looking into.

Passed `Dialog` `handleClose` also!

> 3. Placement of closing X button - still seems a little off (see Ian's previous comment on this).

Gonna do a separate PR with all the style changes (which I've noted down) if that's alright!

DafyddLlyr (approved):

PR #7226, opened 2026-08-21
feat: List templates on explore page

ianjon3s:
## What does this PR do?

- Sets explore as viewable for a user with edit permission only (making the below template interaction make sense)
- Introduces list of templates to the explore page, with information about which are already subscribed to
- Sets up a generic `SearchListItem` that will be eventually used to list all flow types as search results
- Sets up Storybook file for templates widget

image

### Testing

https://storybook.7226.planx.pizza/?path=/docs/editor-components-explore-templateswidget--docs

Toggle `explore` feature flag to test in app:
https://7226.planx.pizza/app/barnet/explore

PR #7228, opened 2026-08-21
feat: Return template details as search result

ianjon3s:
## What does this PR do?

- Builds on initial work to list templates on explore page - https://github.com/theopensystemslab/planx-new/pull/7226
- Adds a generic `SearchListItemDetail` that will later be used for 2-panel view of search results
- Adds `TemplateDetailsModal` to display templates as an isolated search result
- Warns about adding a duplicate template to a team with an additional dialog confirmation step

image

image

### Testing

Toggle `explore` feature flag to test in app:
https://7228.planx.pizza/app/barnet/explore

### DIFF
--- a/apps/editor.planx.uk/src/pages/Explore/components/SearchListItemDetail.tsx
+++ b/apps/editor.planx.uk/src/pages/Explore/components/SearchListItemDetail.tsx
@@ -1,0 +1,85 @@
+import Box from "@mui/material/Box";
+import Divider from "@mui/material/Divider";
+import Tooltip from "@mui/material/Tooltip";
+import Typography from "@mui/material/Typography";
+import React from "react";
+import { FONT_WEIGHT_SEMI_BOLD } from "theme";
+import CheckCircleIcon from "ui/icons/CheckCircle";
+
+import type { SearchResult } from "./SearchResult";
+
+interface SearchListItemDetailProps {
+  result: SearchResult;
+}
+
+export const SearchListItemDetail: React.FC<SearchListItemDetailProps> = ({
+  result: {
+    icon,
+    sourceTeam,
+    statusLabel,
+    title,
+    meta,
+    tag,
+    description,
+    relatedItems,
+  },
+}) => {
+  return (
+    <>
+      <Box sx={{ display: "flex", alignItems: "center", gap: 1, mb: 2 }}>
+        {icon}
+        {sourceTeam && (
+          <Typography
+            variant="body1"
+            sx={{ fontWeight: FONT_WEIGHT_SEMI_BOLD }}
+          >
+            {sourceTeam}
+          </Typography>
+        )}
+      </Box>
+      {statusLabel && (
+        <Box sx={{ display: "flex", alignItems: "center", gap: 0.33, mb: 0.5 }}>
+          <CheckCircleIcon color="success" sx={{ fontSize: 20 }} />
+          <Typography
+            variant="body2"
+            sx={{ fontWeight: FONT_WEIGHT_SEMI_BOLD }}
+          >
+            {statusLabel}
+          </Typography>
+        </Box>
+      )}
+      <Typography variant="h3" component="h2">
+        {title}
+      </Typography>
+      {meta && (
+        <Typography variant="body2" sx={{ color: "text.secondary", mt: 0.5 }}>
+          {meta}
+        </Typography>
+      )}
+      {tag && <Box sx={{ mt: 1.5, display: "flex" }}>{tag}</Box>}
+      {description && (
+        <Typography variant="body1" sx={{ mt: 2 }}>
+          {description}
+        </Typography>
+      )}
+      {relatedItems && relatedItems.items.length > 0 && (
+        <>
+          <Divider sx={{ my: 2 }} />
+          <Typography
+            variant="body1"
+            sx={{ fontWeight: FONT_WEIGHT_SEMI_BOLD, mb: 1.5 }}
+          >
+            {relatedItems.label}
+          </Typography>
+          <Box sx={{ display: "flex", gap: 1, flexWrap: "wrap" }}>
+            {relatedItems.items.map((item) => (
+              <Tooltip key={item.key} title={item.tooltip}>
+                <span>{item.icon}</span>
+              </Tooltip>
+            ))}
+          </Box>
+        </>
+      )}
+    </>
+  );
+};
--- a/apps/editor.planx.uk/src/pages/Explore/components/SearchListItemDetail.tsx
+++ b/apps/editor.planx.uk/src/pages/Explore/components/SearchListItemDetail.tsx
@@ -71,7 +71,7 @@ export const SearchListItemDetail: React.FC<SearchListItemDetailProps> = ({
           >
             {relatedItems.label}
           </Typography>
-          <Box sx={{ display: "flex", gap: 1, flexWrap: "wrap" }}>
+          <Box sx={{ display: "flex", gap: 0.5, flexWrap: "wrap" }}>
             {relatedItems.items.map((item) => (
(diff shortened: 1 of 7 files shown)


apps/editor.planx.uk/src/pages/Explore/components/TemplateDetailsModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+        primaryColour?: string | null;
+        logo?: string | null;
+      };
+    };
+  }[];
+}
+
+const GET_TEMPLATE_DETAILS = gql`
+  query GetTemplateDetails($id: uuid!, $systemTeams: [String!]!) {
+    flows(where: { id: { _eq: $id } }, limit: 1) {
```
DafyddLlyr: `flows_by_pk` is quicker and simpler...!

apps/editor.planx.uk/src/pages/Explore/components/TemplateDetailsModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+      name: string;
+      theme: {
+        primaryColour?: string | null;
+        logo?: string | null;
+      };
+    };
+  }[];
+}
+
+const GET_TEMPLATE_DETAILS = gql`
```
DafyddLlyr, line 45: Let's also grab appropriate ids for all child entities in this query (operations, published flows, etc).

apps/editor.planx.uk/src/pages/Explore/components/TemplateDetailsModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+        details.operations[0].createdAt,
+        details.operations[0].actor,
+      ).formatted
+    : undefined;
+
+  const usedByTeams = [...(details?.usedByTeams ?? [])].sort((a, b) => {
+    if (a.team.id === teamId) return -1;
+    if (b.team.id === teamId) return 1;
+    return 0;
+  });
```
DafyddLlyr, line 185: I think we can skip this by just using `order_by` in the query?

DafyddLlyr: Actually I guess we're trying to say "sort but stick my team first" here right? If so and we need this then this works well, don't think we can do this via GQL.

apps/editor.planx.uk/src/pages/Explore/components/TemplateDetailsModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+          if (confirmed) handleAddToTeam();
+        }}
+        title="Add template to your team?"
+        confirmText="Continue"
+        cancelText="Cancel"
+      >
+        <Typography>
+          You already subscribe to this template, subscribing again would mean
+          maintaining more than one instance of this template.
+        </Typography>
```
DafyddLlyr, line 268: Is this ever a desirable outcome, do we want to allow this? If not when we should just not enable the feature (I guess it is possible today though via the plain "Add flow").

apps/editor.planx.uk/src/pages/Explore/components/SearchListItemDetail.tsx
```diff
@@ -0,0 +1,85 @@
...
+          >
+            {relatedItems.label}
+          </Typography>
+          <Box sx={{ display: "flex", gap: 1, flexWrap: "wrap" }}>
+            {relatedItems.items.map((item) => (
+              <Tooltip key={item.key} title={item.tooltip}>
+                <span>{item.icon}</span>
+              </Tooltip>
+            ))}
+          </Box>
```
DafyddLlyr: The spacing here is pretty painful on the eyes for me..! 🤓 

Image

Getting a strong Hermann grid illusion effect which makes lots of grey dots flash up and it's quite uncomfortable to look at.

https://en.wikipedia.org/wiki/Grid_illusion

apps/editor.planx.uk/src/pages/Explore/components/SearchListItemDetail.tsx
```diff
@@ -0,0 +1,85 @@
...
+          <Typography
+            variant="body1"
+            sx={{ fontWeight: FONT_WEIGHT_SEMI_BOLD, mb: 1.5 }}
+          >
+            {relatedItems.label}
+          </Typography>
+          <Box sx={{ display: "flex", gap: 1, flexWrap: "wrap" }}>
+            {relatedItems.items.map((item) => (
+              <Tooltip key={item.key} title={item.tooltip}>
+                <span>{item.icon}</span>
```
DafyddLlyr, line 77: On my second monitor these logo look pretty rough at this scale, maybe they need to be a touch bigger?

Image

DafyddLlyr (changes requested):
A few small questions, main issue is the flashing grey dots..! (is it just me?) 🩶

apps/editor.planx.uk/src/pages/Explore/components/TemplateDetailsModal.tsx
ianjon3s: Thanks, updated in https://github.com/theopensystemslab/planx-new/pull/7228/changes/f034f13ec5d601b8092102b69d01ec09541b9789

apps/editor.planx.uk/src/pages/Explore/components/TemplateDetailsModal.tsx
ianjon3s: I've updated to include IDs in the query
https://github.com/theopensystemslab/planx-new/pull/7228/changes/e7e25af025d228c253a47889e2a62218d1674b05

apps/editor.planx.uk/src/pages/Explore/components/TemplateDetailsModal.tsx
ianjon3s: Indeed, the intention is to intersect the list to display the team in view first, matching the label of (for example) `Subscribed to by Barnet and 16 other teams:`

apps/editor.planx.uk/src/pages/Explore/components/TemplateDetailsModal.tsx
ianjon3s: Let's think more on this, as there are cases where teams have multiples of templates, for example Doncaster:

image

apps/editor.planx.uk/src/pages/Explore/components/TemplateDetailsModal.tsx
DafyddLlyr: Yep - makes sense - thanks for clarifying.

apps/editor.planx.uk/src/pages/Explore/components/SearchListItemDetail.tsx
ianjon3s: I've updated the badges to be round, which seems to lessen the effect (very much open to your opinion on that of course)

I did try a "stack" approach by layering up the icons, but with such dense colours it didn't look good at all.

image

DafyddLlyr: Works for me, thanks! We can revisit when we reach flows with 100+ teams 😉

apps/editor.planx.uk/src/pages/Explore/components/TemplateDetailsModal.tsx
DafyddLlyr: In which case - let's move ahead with this as-is. A warning is a lot better than our current setup 👌

DafyddLlyr (approved):
Changes all look good - thanks @ianjon3s , @joecarver !

PR #7234, opened 2026-08-21
feat: show ITP and session started in submission history

zz-hh-aa:
# What does this PR do
Preventing empty `<SubmissionEventHistory>` (which Daf pointed out in #7201) which were rendering for submissions with "Sending" and "Invited to pay" statuses. 

- Adds session started and payment request times to `submission_services_log`
- Renders them in the `SubmissionEventHistory` component, ensuring that view response + resubmit don't render for them

Before:
image

After: 
image

(Aware there's some awkward spacing here, struggled to fix it--might bring to dev call or Ian)

# Why
Some histories were empty because `payment_requests` was queried in the `submission_grouped` view but not in `submission_services_log`, which is where the list of all events was being stored. (Ian's design also included "Started session" in the history.)

First I tried creating a new view (since existing `submission_services_log` was complex and had performance issues before. This led me down a rabbit hole of convoluted types and aggregation logic, so I went back to adding CTEs to `submission_services_log`. This works, makes for simpler code but is still imperfect and we might have some performance issues on prod--thoughts welcome!

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/Submissions.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/Submissions.tsx
@@ -18,7 +18,10 @@ const Submissions: React.FC<SubmissionsProps> = ({ flowSlug }) => {
     gql`
       query GetSubmissions($team_id: Int!) {
         submissions: submission_services_log(
-          where: { team_id: { _eq: $team_id } }
+          where: {
+            team_id: { _eq: $team_id }
+            event_type: { _nin: ["Started session", "Invited to pay"] }
+          }
           order_by: { created_at: desc }
         ) {
           flowId: flow_id
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/OpenResponseButtonGrouped.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/OpenResponseButtonGrouped.tsx
@@ -7,10 +7,11 @@ import { DataTableModal } from "ui/shared/DataTable/components/DataTableModal";
 import type { Attempt } from "../types";
 import { FormattedResponse } from "./FormattedResponse";
 
-type Props = { attempt: Attempt; sessionId: string };
+type Props = { attempt: Attempt; sessionId: string; disabled?: boolean };
 
 export const OpenResponseButtonGrouped = (props: Props) => {
   const [modalIsOpen, setModalIsOpen] = useState(false);
+  if (props.disabled) return;
 
   const parseResponse = ({ eventType, status, response }: Attempt) => {
     let data;
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/SubmissionEventsHistory.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/SubmissionEventsHistory.tsx
@@ -103,6 +103,8 @@ const SubmissionEvent: React.FC<{
   isMostRecent: boolean;
 }> = ({ groupedEvent: { sessionId, events: attempts }, isMostRecent }) => {
   const { eventType, createdAt, status } = attempts[0];
+  const hideResponse =
+    eventType === "Invited to pay" || eventType === "Started session";
 
   return (
     <Box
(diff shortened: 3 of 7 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/OpenResponseButtonGrouped.tsx
```diff
@@ -7,10 +7,11 @@ import { DataTableModal } from "ui/shared/DataTable/components/DataTableModal";
 import type { Attempt } from "../types";
 import { FormattedResponse } from "./FormattedResponse";
 
-type Props = { attempt: Attempt; sessionId: string };
+type Props = { attempt: Attempt; sessionId: string; disabled?: boolean };
 
 export const OpenResponseButtonGrouped = (props: Props) => {
   const [modalIsOpen, setModalIsOpen] = useState(false);
+  if (props.disabled) return;
```
zz-hh-aa: I passed the optional `disabled` prop here to avoid nested ternaries in the parent component

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/SubmissionEventsHistory.tsx
```diff
@@ -147,6 +163,7 @@ const SubmissionEvent: React.FC<{
                 <OpenResponseButtonGrouped
                   attempt={attempts[0]}
                   sessionId={sessionId}
+                  disabled={hideResponse}
```
zz-hh-aa, line 162: I passed the optional disabled prop here to avoid nested ternaries here, since `OpenResponseButtonGrouped` is the only component in this conditional block that might need hiding based on the event type

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/Submissions.tsx
```diff
@@ -18,7 +18,10 @@ const Submissions: React.FC<SubmissionsProps> = ({ flowSlug }) => {
     gql`
       query GetSubmissions($team_id: Int!) {
         submissions: submission_services_log(
-          where: { team_id: { _eq: $team_id } }
+          where: {
+            team_id: { _eq: $team_id }
+            event_type: { _nin: ["Started session", "Invited to pay"] }
```
zz-hh-aa, line 23: Because the view itself has changed, I'm filtering out session started + invited to pay events here (so they don't show up in the non-feature-flagged submissions log)

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/SubmissionEventsHistory.tsx
```diff
@@ -103,6 +115,8 @@ const SubmissionEvent: React.FC<{
   isMostRecent: boolean;
 }> = ({ groupedEvent: { sessionId, events: attempts }, isMostRecent }) => {
   const { eventType, createdAt, status } = attempts[0];
+  const hideResponse =
+    eventType === "Invited to pay" || eventType === "Started session";
```
jessicamcinchak, line 119: 👍 In the future, I could imagine the "response" for "Invited to pay" could include a copy of the payment magic link which sometimes councils are looking for to help agents/applicants debug, but very happy for this to be a "later" thing / not now !

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/Submissions.tsx
jessicamcinchak: Makes sense! Thanks for carefully balancing new & old here

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/utils.ts
```diff
@@ -17,3 +19,21 @@ export const getConsolidatedStatus = (
...
+  isMostRecent: boolean,
+  status: string | undefined,
+  eventType: Submission["eventType"] | SessionEvent,
+): eventType is Submission["eventType"] => {
+  return (
+    isMostRecent &&
+    status !== "Success" &&
+    !NON_RESUBMITTABLE_EVENT_TYPES.includes(eventType)
+  );
+};
```
jessicamcinchak, line 39: 👍

jessicamcinchak (approved):
Changes all look good here to me! Agree big `submission_services_log` view might be one to revisit in future, but these should be "lightweight" events to add to it now in meantime.

PR #7237, opened 2026-08-24
docs: Add `grep` command to `pendingCouncilDnsRecords` fetch

DafyddLlyr:
Please see comment below.

### DIFF
--- a/doc/how-to/how-to-setup-custom-domains.md
+++ b/doc/how-to/how-to-setup-custom-domains.md
@@ -104,7 +104,7 @@ This adds the domain to the mining cert as an SAN; ACM generates the CNAME the c
 Read the validation record:
 
 ```sh
-pulumi stack output pendingCouncilDnsRecords --stack production --json
+pulumi stack output pendingCouncilDnsRecords --stack production --json | grep -i -C 2 "{COUNCIL}"
 ```
 
 Get the shared CDN domain for the routing record:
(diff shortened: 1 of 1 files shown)


doc/how-to/how-to-setup-custom-domains.md
```diff
@@ -104,7 +104,7 @@ This adds the domain to the mining cert as an SAN; ACM generates the CNAME the c
 Read the validation record:
 
 ```sh
-pulumi stack output pendingCouncilDnsRecords --stack production --json
+pulumi stack output pendingCouncilDnsRecords --stack production --json | grep -i -C 2 "{COUNCIL}"
```
DafyddLlyr, line 107: This fetches the two "context" lines (above and below) - https://explainshell.com/explain?cmd=grep+-i+-C+2

This means that instead of a long json which needs to be scrolled we just get a small window matching the council name, e.g.

```sh
  },
  {
    "domain": "planningservices.dummyCouncil.gov.uk",
    "validationCname": {
      "name": "dummy name",
      "target": "dummy target"
    }
```

freemvmt: Very nice, this is already what I was doing locally but when I originally wrote the HOWTO I didn't want to go in too hard with perfect commands. But in retrospect there's no reason not to, these are copy and paste jobs right.

freemvmt (approved):
✨

doc/how-to/how-to-setup-custom-domains.md
DafyddLlyr: Yeah I've also been doing it locally for a while, never saving the command to docs, and having to painfully remember the right syntax each time 😂

freemvmt: thanks for taking the time to commit :grin:

PR #7241, opened 2026-08-24
chore: Bump osl/map

DafyddLlyr:
- Bumps to @opensystemslab/map@1.0.0-alpha.15 to resolve a11y issue
- Moves map to shared pnpm catalog - a11y fix is also needed in LPS


freemvmt (approved):
excellent!

PR #7242, opened 2026-08-25
feat[alt notes]: new `Note` component type and sanitise function on flatten

jessicamcinchak:
- Adds a `Notes v2` feature flag to separate this work from original implementation
- Adds a new compent type `Note` (`{"type": 999}`) which relies on https://github.com/theopensystemslab/planx-core
  - This means standalone notes are _always_ in the graph and tracked by ShareDB operations
  - The "Show notes" graph toggle determines if they are visible or not; even when hidden, ShareDB still accounts for their position in all nearby operations
- Incorporates new component type into "Search"
- Adds a `sanitiseNotes` step to `dataMerged` function which runs when flattening flows (when checking for changes to publish, and on load of `/preview` & `/draft` routes)

**Not yet handled:** 
- Attached notes https://github.com/theopensystemslab/planx-new/pull/7265
- "Visibility" rules for templates and source templates https://github.com/theopensystemslab/planx-new/pull/7271
- DB-level permissions for `flows.data` (eg `getFlowData` API helper still relies on `$public` client; this is the _only_ questionable use of `$public` client, the rest are all querying published flows which will be "sanitised" on next publish) - see discussion here https://opensystemslab.slack.com/archives/C5Q59R3HB/p1787759810242429
- Disabling and migrating "old" ways of creating standalone notes

**To test:**
- Turn on the feature flag
- Confirm you see the new "Note" option in the component select modal
- Create a "Note" component, maybe a few!
- Create other component types and populate "Internal notes"
- Load a `/draft`, `/preview` or `/published` route and inspect flow data (eg `window.api.getState().flow`)
  - For a given `Note` component type, confirm it doesn't have `data: { note: "xx" }`
  - For any other component type with an internal notes, confirm that `data` no longer includes `notes` prop
  - When you compare to the same flow inspection in the editor graph page, you'll see full note _content_

### DIFF
--- a/apps/api.planx.uk/shared/dataMerged.ts
+++ b/apps/api.planx.uk/shared/dataMerged.ts
@@ -1,5 +1,6 @@
 import type { FlowGraph } from "@opensystemslab/planx-core/types";
 import { ComponentType } from "@opensystemslab/planx-core/types";
+import cloneDeep from "lodash/cloneDeep.js";
 
 import { getFlowData, getMostRecentPublishedFlowVersion } from "../helpers.js";
 import type { Node } from "../types.js";
@@ -51,7 +52,25 @@ export const dataMerged = async (
     }
   }
 
-  return mergedGraph as FlowGraph;
+  return redactNotesFromFlow(mergedGraph) as FlowGraph;
+};
+
+const redactNotesFromFlow = (flow: MergedGraph): MergedGraph => {
+  const redactedFlow = cloneDeep(flow);
+
+  for (const [nodeId, node] of Object.entries(redactedFlow)) {
+    // Leave "Note" components in place in the graph, but fully delete their data/content
+    if (node?.type === ComponentType.Note) {
+      delete redactedFlow[nodeId]?.["data"];
+      continue;
+    }
+    // For any component type that has "internal notes", remove only this prop
+    if (node?.data?.notes) {
+      delete redactedFlow[nodeId]?.["data"]?.["notes"];
+    }
+  }
+
+  return redactedFlow;
 };
 
 const fetchAndMergeStack = async (
--- a/apps/api.planx.uk/shared/dataMerged.ts
+++ b/apps/api.planx.uk/shared/dataMerged.ts
@@ -52,10 +52,10 @@ export const dataMerged = async (
     }
   }
 
-  return redactNotesFromFlow(mergedGraph) as FlowGraph;
+  return sanitiseNotes(mergedGraph) as FlowGraph;
 };
 
-const redactNotesFromFlow = (flow: MergedGraph): MergedGraph => {
+export const sanitiseNotes = (flow: MergedGraph): MergedGraph => {
   const redactedFlow = cloneDeep(flow);
 
   for (const [nodeId, node] of Object.entries(redactedFlow)) {
--- a/apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
@@ -36,12 +36,12 @@ function NoteComponent(props: Props) {
         <ModalSectionContent>
           <InputRow>
             <Input
-              name="note"
+              name="text"
               format="large"
               minRows={4}
               multiline={true}
               placeholder="Note"
-              value={formik.values.note}
+              value={formik.values.text}
               onChange={formik.handleChange}
               disabled={props.disabled}
             />
--- a/apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
@@ -7,7 +7,7 @@ import Input from "ui/shared/Input/Input";
 import InputRow from "ui/shared/InputRow";
 
 import type { EditorProps } from "../shared/types";
-import type { Note} from "./model";
+import type { Note } from "./model";
 import { parseContent, validationSchema } from "./model";
 
 type Props = EditorProps<ComponentType.Note, Note>;
--- a/apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
@@ -1,0 +1,60 @@
+import { ComponentType } from "@opensystemslab/planx-core/types";
+import { useFormikWithRef } from "@planx/components/shared/useFormikWithRef";
+import { ModalFooter } from "ui/editor/ModalFooter";
+import ModalSection from "ui/editor/ModalSection";
+import ModalSectionContent from "ui/editor/ModalSectionContent";
+import Input from "ui/shared/Input/Input";
+import InputRow from "ui/shared/InputRow";
+
+import type { EditorProps } from "../shared/types";
+import type { Note} from "./model";
+import { parseContent, validationSchema } from "./model";
+
+type Props = EditorProps<ComponentType.Note, Note>;
+
+export default NoteComponent;
+
+function NoteComponent(props: Props) {
+  const formik = useFormikWithRef(
+    {
+      initialValues: parseContent(props.node?.data),
+      onSubmit: (newValues) => {
+        props.handleSubmit?.({
+          type: ComponentType.Note,
+          data: newValues,
+        });
+      },
+      validationSchema,
+    },
+    props.formikRef,
+  );
+
+  // Notes will never be "templated", therefore no template fields here
+  return (
+    <form onSubmit={formik.handleSubmit} id="modal">
+      <ModalSection>
+        <ModalSectionContent>
+          <InputRow>
+            <Input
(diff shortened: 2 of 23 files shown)


apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
```diff
@@ -0,0 +1,60 @@
...
+              onChange={formik.handleChange}
+              disabled={props.disabled}
+            />
+          </InputRow>
+        </ModalSectionContent>
+      </ModalSection>
+      <ModalFooter
+        formik={formik}
+        disabled={props.disabled}
+        showTags={true}
```
jessicamcinchak, line 54: I'm assuming editors may still want to tags Notes as "To review" especially!

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Question.tsx
```diff
@@ -80,9 +80,11 @@ const Question: React.FC<Props> = React.memo((props) => {
           "card",
           "decision",
           "type-" + TYPES[props.type as TYPES],
+          `${props.type === TYPES.Note ? "question" : ""}`,
           {
             isDragging,
             isClone: isCloneValue,
+            isNote: props.type === TYPES.Note,
```
jessicamcinchak, line 87: These `classNames` are hacky for right now but get the desired sticky-yellow with folded corner!

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/ComponentsTab/componentData.ts
```diff
@@ -17,6 +18,21 @@ export interface Category {
 const title = (type: TYPES): string => COMPONENT_TITLES[type] ?? String(type);
 
 export const ALL_CATEGORIES: Category[] = [
+  ...(hasFeatureFlag("NOTES_V2")
+    ? [
+        {
+          label: "",
```
jessicamcinchak: I'm assuming `Note` component types should have no section label/category, and be top of list like first implementation proposed.

apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Search/SearchResultCard/getDisplayDetailsForResult.tsx
```diff
@@ -34,8 +34,11 @@ const keyFormatters: KeyMap = {
         ? "Option (description)"
         : "Description",
   },
+  "data.note": {
+    getDisplayKey: () => "Note",
+  },
   "data.notes": {
-    getDisplayKey: () => "Internal notes",
+    getDisplayKey: () => "Internal notes", // TODO rename here and in-modals?
```
jessicamcinchak, line 38: Now handled https://github.com/theopensystemslab/planx-new/pull/7265

apps/api.planx.uk/shared/dataMerged.ts
```diff
@@ -51,7 +52,25 @@ export const dataMerged = async (
     }
   }
 
-  return mergedGraph as FlowGraph;
+  return sanitiseNotes(mergedGraph) as FlowGraph;
```
jessicamcinchak, line 55: Important to note this part is _not_ actually feature-flagged! All publishes will start stripping internal `notes` props once this merges. (New standalone `Note` component type _is_ feature flagged so won't exist in real content yet). 

I don't actually think this is a huge problem (and there's no rush to merge here), but one to be clear about !

apps/api.planx.uk/shared/dataMerged.ts
```diff
@@ -51,7 +52,25 @@ export const dataMerged = async (
...
+      delete redactedFlow[nodeId]?.["data"];
+      continue;
+    }
+    // For any component type that has "internal notes", remove only this prop
+    if (node?.data?.notes) {
+      delete redactedFlow[nodeId]?.["data"]?.["notes"];
+    }
+  }
+
+  return redactedFlow;
```
jessicamcinchak, line 73: Also good to be explicit about: 
- Now that we strip away `notes` fields and `Note`-component `data` on flatten
- If you have _only_ updated notes since your last publish and check for changes again, we will display "No new changes to publish" !
- Your notes edits are still captured in "History" and restore-able

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/ComponentsTab/componentData.ts
DafyddLlyr: I agree it should be different or prominent, but at first glance it looks like we've made a mistake and missed this label. Maybe it needs a divider beneath it or something?

Very minor UI point, we can easily revisit this.

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/ComponentsTab/componentData.ts
```diff
@@ -17,6 +18,21 @@ export interface Category {
...
+          items: [
+            {
+              type: TYPES.Note,
+              slug: "note",
+              title: title(TYPES.Note),
+              description: "Add a sticky note visible to editors only",
```
DafyddLlyr: question(non-blocking): I'm not sure if we should be going for "editors" or "Editors" for these - I tend to use "Editors" in Slack etc but I'm not sure ever really refer to these user types anywhere outside the teams pages.

apps/editor.planx.uk/src/pages/FlowEditor/floweditor.scss
```diff
@@ -834,6 +834,7 @@ $fontMonospace: "Source Code Pro", monospace;
   }
   align-items: center;
 
+  // TODO retire `.question.` prefix
```
DafyddLlyr, line 837: 👍

apps/api.planx.uk/shared/dataMerged.ts
DafyddLlyr: Thanks for highlighting - no issues here, I think this is an improvement!

DafyddLlyr (commented):
Code all looks good, no major comments below at all - all pretty straightforwards..!

## Testing...

- [x] Turn on the feature flag
- [x] Confirm you see the new "Note" option in the component select modal
- [x] Create a "Note" component, maybe a few!
- [x] Create other component types and populate "Internal notes"
- [x] Load a /draft, /preview or /published route and inspect flow data (eg window.api.getState().flow)
  - [x] For a given Note component type, confirm it doesn't have data: { note: "xx" }
  - [x] For any other component type with an internal notes, confirm that data no longer includes notes prop
  - [x] When you compare to the same flow inspection in the editor graph page, you'll see full note content

Some very minor issues it's very possible these are already picked up elsewhere in the stack -

 - History shows "undefined" as the note, as opposed to maybe a snippet

Image

 - We should truncate notes after a while

Image

 -  Adding notes when "Toggle notes" is off is a bit of a strange experience. Same as adding a data field or image thought I guess if their toggles are off.

apps/editor.planx.uk/src/@planx/components/Note/model.ts
```diff
@@ -0,0 +1,21 @@
+import type { SchemaOf } from "yup";
+import { object, string } from "yup";
+
+import type { BaseNodeData } from "../shared";
+import { baseNodeDataValidationSchema, parseBaseNodeData } from "../shared";
+
+export interface Note extends BaseNodeData {
+  text: string;
```
jessicamcinchak, line 8: cc @DafyddLlyr - by renaming `note` to `text` here, we're now more consistent with other component types and naturally picked up correctly in "History" (good spot!) ➕ 

Image



And, perhaps most importantly, don't need to add a new custom prop name to this list 😅
Image

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/ComponentsTab/componentData.ts
jessicamcinchak: Added a comment to revisit styling here so we don't forget ! 

Also a wider testing question to come back to 🔖 : 
- Should we only show the `Note` option to select _when_ "Show notes" toggle is on? Also agree with your point it's a bit odd to add a `Note` when "show" toggle is off because no graph feedback!

apps/editor.planx.uk/src/pages/FlowEditor/components/forms/AddComponentModal/ComponentsTab/componentData.ts
jessicamcinchak: good catch, updated to capital-e Editors!

DafyddLlyr (approved):
Sorry - just realised I never gave this a ✅ after my comments!

PR #7245, opened 2026-08-25
fix(a11y): Enhanced Text Input - DAC_Labels_or_Instructions_01/02

joecarver:
From page 23 in the revised review.

This one was a little bit more involved, more to do with the UX flow when writing a new description, and how the inline text box presents a different UX compared to the other two options which take the user to a new page. So this PR changes the flow so that all three options take the user to the same page, and the text box is optionally pre-filled depending on which option they chose, and the warning about the input not being re-checked is shown consistently.

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/types.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/types.tsx
@@ -4,10 +4,11 @@ import type { EnhancedTextInput, TaskAction } from "../types";
 
 export type Props = PublicProps<EnhancedTextInput>;
 
+export type Step = "input" | "selection" | "modification";
+
 interface FormValuesBase {
   userInput: string;
   selectedOption: TaskAction | null;
-  customDescription: string;
 }
 
 interface FormValuesIdle extends FormValuesBase {
--- a/apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/ModifyUserInput.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/ModifyUserInput.tsx
@@ -16,6 +16,7 @@ import type { EnhancedTextInput } from "../types";
 import type { FormValues } from "./types";
 
 const HEADING_ID = "confirm-project-description-heading";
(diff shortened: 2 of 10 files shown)


apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/types.tsx
```diff
@@ -4,10 +4,11 @@ import type { EnhancedTextInput, TaskAction } from "../types";
 
 export type Props = PublicProps<EnhancedTextInput>;
 
+export type Step = "input" | "selection" | "modification";
```
DafyddLlyr, line 7: 👍

DafyddLlyr (approved):
There's an existing issue with "back" navigation here that may merit a second pair of eyes - we hit the same loading state again unnecessarily. None of the inputs have changed, so we should be able to avoid this. You also get this issue on staging when going "back" from the confirm page.

Changes here all look good - I'd cross-post to #planx-design with details of this and a Storybook as it's a pretty significant UX change (we spent a while with UR here trying to find the best version). Just to keep people in the loop!

joecarver:
> There's an existing issue with "back" navigation here that may merit a second pair of eyes - we hit the same loading state again unnecessarily. None of the inputs have changed, so we should be able to avoid this. You also get this issue on staging when going "back" from the confirm page.

@DafyddLlyr What I'm seeing is that 'back' always goes back to the previous flow element i.e. in the accessibility testing flow it's the notice. I assumed this was just the standard behaviour of the back button. As this is a single element with multiple steps, do we not need a separate back button to navigate between the separate steps? Or should the existing back button be overloaded to handle this?

joecarver:
> There's an existing issue with "back" navigation here that may merit a second pair of eyes - we hit the same loading state again unnecessarily. None of the inputs have changed, so we should be able to avoid this. You also get this issue on staging when going "back" from the confirm page.

@DafyddLlyr Have fixed this now by storing the input in the parent components state and passing it down, which prevents the `enhanceProjectDescription` as the value has not changed

DafyddLlyr (approved):
Thanks for resolving this one @joecarver ..!

I can now navigate back without re-triggering the "enhance" step unnecessarily 👍 

Flagging this one for @ianjon3s also as there's a change to the UX here that's worth having on all our radars..!

PR #7247, opened 2026-08-25
feat: group event history by type

zz-hh-aa:
Rewrites the `groupEvents` function to group at the highest level by event type and not time. The query returns the events sorted by time desc so no further sorting required. 

Some of the event histories are still quite long, so made the `SubmissionDetails` panel sticky while I was at it.


jessicamcinchak (approved):
Great!

PR #7249, opened 2026-08-25
chore(editor): Rename GovPay variables and components

DafyddLlyr:
Minor housekeeping, follow up to https://github.com/theopensystemslab/planx-new/pull/7211

This PR rename GovPay-specific variables and components to provider-agnostic names. 

A few notable exceptions - 
- `govPayMetadata` data property / Formik field name - maps to stored flow data in the DB
- `GovPayMetadata` type - defined in `planx-core`
- `GOV_UK_PAY_URL` - I think we want this to stay as a GovPay-specific endpoint, Stripe will get its own alongside (maybe `/pay/stripe` which we can eventually consolidate)

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/Pay/Editor/Editor.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/Pay/Editor/Editor.tsx
@@ -22,8 +22,8 @@ import InputRow from "ui/shared/InputRow";
 import { Switch } from "ui/shared/Switch";
 
 import type { EditorProps } from "../../shared/types";
-import { GovPayMetadataSection } from "./GovPayMetadataSection";
 import { InviteToPaySection } from "./InviteToPaySection";
+import { PaymentMetadataSection } from "./PaymentMetadataSection";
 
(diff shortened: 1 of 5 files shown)


zz-hh-aa (approved):
Thanks for explaining the exceptions! For my own Stripe context was wondering: on the DB side, I guess columns like `payment_requests.govpay_metadata` and `payment_requests.govpay_payment_id` will be supplemented by equivalent columns for Stripe? Before later being removed or appended with `_deprecated`?

jessicamcinchak (approved):

DafyddLlyr:
> Thanks for explaining the exceptions! For my own Stripe context was wondering: on the DB side, I guess columns like payment_requests.govpay_metadata and payment_requests.govpay_payment_id will be supplemented by equivalent columns for Stripe? Before later being removed or appended with _deprecated?

Yep exactly - not 100% sure of the DB solution we'll end up on, but I just wanted to highlight that this structure is intentionally not touched here.

We'll separately need to manage migration (?) / population of a new column when it comes to Stripe data.

PR #7250, opened 2026-08-25
fix(editor): Additional validation for Stripe metadata

DafyddLlyr:
## What does this PR do?
This PR adds an additional check for payment metadata to conform to Stripe requirements. All other requirements are looser than GovPay, we can come back and expand this later.

Docs - https://docs.stripe.com/metadata

## Testing
I've checked this is safe to roll out, and that no existing metadata contains `[` or `]` with the following check -

```sql
SELECT
  f.id AS flow_id,
  f.slug AS flow_slug,
  t.name AS team_name,
  node_id,
  m.value ->> 'key' AS metadata_key,
  m.value ->> 'value' AS metadata_value
FROM flows f
JOIN teams t ON t.id = f.team_id,
LATERAL jsonb_each(f.data) AS nodes(node_id, node_data),
LATERAL jsonb_array_elements(node_data -> 'data' -> 'govPayMetadata') AS m(value)
WHERE node_data -> 'data' ? 'govPayMetadata'
  AND m.value ->> 'type' = 'static'
  AND (m.value ->> 'value' LIKE '%[%' OR m.value ->> 'value' LIKE '%]%');
```

Would appreciate a second pair of eyes here to make sure this query doesn't miss anything. Dropping the final clause gives me the expected results - a list of all fields.

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/Pay/model.test.ts
+++ b/apps/editor.planx.uk/src/@planx/components/Pay/model.test.ts
@@ -96,6 +96,39 @@ describe("Payment Metadata Schema", () => {
     expect(result).toEqual(input);
   });
 
+  test("static values cannot contain square brackets", async () => {
+    const input = [
+      ...defaults,
+      { key: "myKey", value: "value[0]", type: "static" },
+    ];
+    const errors = await validate(input);
+    expect(errors).toHaveLength(1);
+    expect(errors[0]).toMatch(
+      /Static values cannot contain \[ or \] characters/,
+    );
+  });
+
+  test("static values reject lone opening bracket", async () => {
+    const input = [
+      ...defaults,
+      { key: "myKey", value: "value[with", type: "static" },
+    ];
+    const errors = await validate(input);
+    expect(errors).toHaveLength(1);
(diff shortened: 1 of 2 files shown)


jessicamcinchak (approved):
This looks good! 

Per your testing query, I think it's probably important to note that in this line below, `'%[%'` _might_ not pick up `'[%'` and vice versa for the end-square bracket!? I think _if_ we do have any existing brackets, the `'[%'` is actually perhaps most likely and worth double-checking for explicitly?

Postgres `%` is picky this way and I never quite remember the exact behavior / have been fooled by false-confident returns before! 

```
  AND (m.value ->> 'value' LIKE '%[%' OR m.value ->> 'value' LIKE '%]%');
```

DafyddLlyr:
@jessicamcinchak Thanks - good challenge to cross-check this one. As per [the docs](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-LIKE) _"a percent sign (%) matches any sequence of zero or more characters."_ which I read to mean it will pick up on strings ending or starting in `[` or `]`.

I've tested this locally, by making metadata which does end with `]`, and another which starts with `[` and the above does spot them - 

image

PR #7254, opened 2026-08-25
feat: searchable flows db changes

joecarver:
Implements the flow search functionality at the DB level, via a combination of:
- New Postgres view `searchable_views` so we can narrow the columns returned and more cleanly scope permissions 
- New Postgres extension pg_trgm which uses trigram matching to give fuzzy search w/ similarity scores
- trgm Index on `searchable_views.name`
  - do we want any other fields searchable + indexed? Haven't benchmarked the time implications of this yet but can do
- Postgres function `search_views` which executes the query against `searchable_flows`
  - this can then be executed in a graphql query
- Hasura metadata for new function and view
- Introspection and behaviour tests for new function and view


apps/hasura.planx.uk/metadata/databases/default/functions/public_search_flows.yaml
```diff
@@ -0,0 +1,9 @@
+function:
+  name: search_flows
+  schema: public
+configuration:
+  exposed_as: query
+permissions:
+  - role: teamEditor
+  - role: teamAdmin
```
zz-hh-aa: Nit: don't think this is needed here, since `teamAdmin` exists _in addition to_ `teamEditor` per-user, and we [never assign `teamAdmin` as a default role](https://github.com/theopensystemslab/planx-new/blob/a4aee24d4e7d704289ee2ce86d294b659ef522bd/apps/api.planx.uk/modules/auth/service/jwt.ts#L51) (it's always in the list of `x-hasura-allowed-roles` instead, and manually overrides the default role in select operations, eg when managing team members).

apps/hasura.planx.uk/metadata/databases/default/tables/public_searchable_flows.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - name: team
+    using:
+      manual_configuration:
+        column_mapping:
+          team_id: id
+        insertion_order: null
+        remote_table:
+          name: teams
+          schema: public
+select_permissions:
```
zz-hh-aa: Just checking my understanding here--cause `public.flows` has many select permissions, this view is a straightforward way of limiting the search ability to admins + editors at the db level? Anything else I missed?

apps/hasura.planx.uk/migrations/default/1787578731000_add_search_flows_function/up.sql
```diff
@@ -0,0 +1,43 @@
+-- Powers the explore page "Search PlanX" feature: fuzzy search over flows & templates, by name
+-- ranked by similarity using pg_trgm
+
+CREATE EXTENSION IF NOT EXISTS pg_trgm;
```
DafyddLlyr: A heads up on postgres extensions - in order to install them on AWS RDS we'll need to manually redeploy the data layer. Please see `shared_preload_libraries` here - 

https://github.com/theopensystemslab/planx-new/blob/joe/search-backend/infrastructure/data/index.ts#L31-L46

apps/hasura.planx.uk/metadata/databases/default/tables/public_searchable_flows.yaml
joecarver: yeah pretty much! more specifically, because we run the SELECT query from a function, and the function declares `RETURNS SETOF <table_name>`, if we use `flows` here then the function would inherit `flows` permission set. 

So by defining a view we can narrow the permission set (at the view creation step) and keep this tracked separately.

apps/hasura.planx.uk/migrations/default/1787578731000_add_search_flows_function/up.sql
```diff
@@ -0,0 +1,43 @@
...
+
+CREATE INDEX flows_name_trgm_idx ON public.flows
+USING GIN (name gin_trgm_ops);
+
+-- question - do we want any other columns to be searchable? should index those also if so
```
DafyddLlyr: Yes - we want summary, description, and limitations also searchable here. I guess with `name` taking a higher weighting?

apps/hasura.planx.uk/tests/search_flows.test.js
```diff
@@ -0,0 +1,143 @@
...
+    test.each(["public", "analyst", "teamViewer"])(
+      "%s role cannot call search_flows",
+      async (role) => {
+        const i = await introspectAs(role);
+        expect(i.queries).not.toContain("search_flows");
+      },
+    );
+  });
+
+  describe("behaviour", () => {
```
DafyddLlyr: It's actually very nice to see these tests here! 👍 

However, we try to keep these introspection tests scoped to permission checking only as opposed to behaviour (there's places like `lowcal_sessions` we don't do this properly though!). I think the reason it's a bit misleading / confusing is that these behavioural tests do not test the permissions at all - we're calling `gqlAdmin` here instead of `gqlWithRole`.

Functional tests like this may be better in e2e/tests/api.

DafyddLlyr (commented):
A few broad questions..!

1. `pg_trgm` vs native Postgres full-text search (`tsvector` / `tsquery`)
`pg_trgm` is new to me but it looks like it gives us fuzzy search (typos!) which is nice - what are the deciding factors to use this over the native text search? Does us wanting to include additional fields, potentially with weightings, change this decision at all?

No strong feelings on which way we go, just wanted to understand the motivations and trade-offs here.

Regardless, wrapping in a function is nice and the way to go here as it allows us to change this implementation in future without breaking the API surface.

2. The view and permissions model
I'm not too sure we need this, and I don't think it's really scoping permission (I might just be missing something here!). If `search_flows` just `RETURNS SETOF flows`, I think the standard Hasura permissions will apply which should be fine. This would let us drop the view and the column list (repeated x3). I can see adding or amending columns here in future being quite an tricky process - needing to thread it through a few places.

joecarver:
> The view and permissions model
I'm not too sure we need this, and I don't think it's really scoping permission (I might just be missing something here!). If search_flows just RETURNS SETOF flows, I think the standard Hasura permissions will apply which should be fine. This would let us drop the view and the column list (repeated x3). I can see adding or amending columns here in future being quite an tricky process - needing to thread it through a few places.

Hmm yeah, you're right, well the columns in the view are a subset of the columns in `flows`, so it technically _could_ have scoped permissions, but I suppose if flows already grants wider permissions to the same roles then it's not actually meaningful. And yeah it would be harder to keep in sync. Will remove it any query flows directly.

apps/hasura.planx.uk/tests/search_flows.test.js
joecarver: done here https://github.com/theopensystemslab/planx-new/pull/7254/commits/dd254ac11af6832d12c9b84acccd780e22de29d4 and prev tests removed https://github.com/theopensystemslab/planx-new/pull/7254/commits/98d4db4c7f5b99131ce87bb9b3b5e0cbf0ee2eda

DafyddLlyr: Feel free to also add `@regression` to these - it's fine to run them once a day, this isn't likely to have much cross-over with other features.

apps/hasura.planx.uk/migrations/default/1787578731000_add_search_flows_function/up.sql
```diff
@@ -0,0 +1,24 @@
+-- Powers the explore page "Search PlanX" feature: search over flows & templates
+-- by name, summary, description, and limitations - ranked by relevance using
+-- Postgres full text search, with name weighted highest, then summary, description/limitations
+-- HTML is stripped from description/limitations before indexing
```
DafyddLlyr, line 4: Well spotted..!

apps/hasura.planx.uk/migrations/default/1787578731000_add_search_flows_function/up.sql
```diff
@@ -0,0 +1,24 @@
...
+
+ALTER TABLE public.flows ADD COLUMN search_vector tsvector
+GENERATED ALWAYS AS (
+  setweight(to_tsvector('english', coalesce(name, '')), 'A')
+  || setweight(to_tsvector('english', coalesce(summary, '')), 'B')
+  || setweight(to_tsvector('english', regexp_replace(coalesce(description, ''), '<[^>]+>', ' ', 'g')), 'B')
```
DafyddLlyr, line 10: Obligatory link 😂 https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454

DafyddLlyr: This should be fine though, as these are simple rich text inputs with a limited subset of tags.

I can't directly test this as the Pizza Hasura is hitting an inconsistent metadata error - you may need to destroy and recreate as the migrations have likely churned here.

Image

DafyddLlyr: Approved however as we have test coverage and I'll come back to this once merged and give it a pass over..! ✅

DafyddLlyr (approved):
Appreciate the thought that's gone into this and the effort to try out both methods. Not been able to test directly due to Hasura - it's probably worth being cautious and destroying the Pizza, recreating, and testing, before proceeding to `main` juuust incase.

PR #7255, opened 2026-08-25
style(editor): SubmissionDetailModal

zz-hh-aa:
Picks up style suggestions from Ian in #7201 + one from Alki in S&T

1. `SubmissionDetailModal` close button in line with modal heading 
2. Moves `Badge` from `Explore` into shared components and reuses in `SubmissionDetails`
3. `SubmissionDetails` reuses `DescriptionList`
4. `SubmissionDetailModal` responsive layout (collapses to one column on smaller screens)
5. `ViewSubmissionButtonGrouped` and `DownloadSubmissionButtonGrouped` use `startIcon` prop
6. `ResubmitButtonGrouped`: replaced `<Link>` with `<Button variant="link">`
7. Adjusts spacing and makes border consistent in `SubmissionEventsHistory`
8. Removes cell highlight when clicking a row--not totally sure about this one!


apps/editor.planx.uk/src/ui/shared/DataTable/DataTable.tsx
```diff
@@ -147,6 +147,9 @@ export const DataTable = <T,>({
                 backgroundColor: "action.hover",
               },
             }),
+            "& .MuiDataGrid-cell:focus": {
+              outline: "none",
```
zz-hh-aa, line 152: This was originally flagged by Alki in S&T--clicking a row link highlights the cell that was clicked (not the row, which is the actual link). 

I had a look at where we use DataTable elsewhere (eg PlatformAdminPanel) and don't think this is an issue

Image

But do we think council editors will miss this highlight in their submissions log?

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/SubmissionEventsHistory.tsx
```diff
@@ -75,7 +75,7 @@ export const SubmissionEventsHistory: React.FC<{ events: Submission[] }> = ({
         background: "background",
         border: 1,
         margin: 1,
-        borderColor: "border.main",
+        borderColor: "lightgrey",
```
zz-hh-aa, line 78: To be consistent with DescriptionList bottom border

apps/editor.planx.uk/src/ui/shared/DataTable/DataTable.tsx
```diff
@@ -147,6 +147,9 @@ export const DataTable = <T,>({
                 backgroundColor: "action.hover",
               },
             }),
+            "& .MuiDataGrid-cell:focus": {
```
DafyddLlyr: nit: We should import `gridClasses` here

apps/editor.planx.uk/src/ui/shared/DataTable/DataTable.tsx
DafyddLlyr: I think this looks better without..! The alternative here would be to make the actual row clickable in the submissions log as opposed to each cell - this is maybe closer to our intention here.

DafyddLlyr (approved):
Great!

PR #7256, opened 2026-08-25
integrate eslint/prettier into `infrastructure/`

freemvmt:
As per title, just as a drive-by while I was experimenting with importing the Scanii stack via Pulumi.

Just followed READMEs in `packages/` and ran `lint:fix` etc across repo.

All changes should be purely cosmetic, I skimmed it all and looks OK to me.

Also bundles in the subjection of the `eslint-config` package to it's own (`base`) rules :cyclone:


DafyddLlyr (approved):
Great! Thanks for this - worth picking up for sure.

PR #7265, opened 2026-08-26
feat[alt notes]: handle internal notes

jessicamcinchak:
- "Show notes" toggle now displays internal notes on the graph, using same code pattern as displaying data fields, tags, help text, etc
- "Internal notes" component modal section renamed to plain "Note" with consistent sticky icon - intention is for this to feel "same as" standalone note component type for editors I think!
  - Now also always displayed at bottom so that in-modal info heirarchy matches graph node heirarchy when all attributes toggled on (eg note is below tags)

**Not yet handled:** 
- Display rules for nodes in templated flows (eg always hide same as standalone note components unless in templated folder/attached to templated node?)
- Internal note input field and graph node display on "options" https://github.com/theopensystemslab/planx-new/pull/7269


DafyddLlyr (approved):
Appreciate the small and well-scoped PR..!

Not tested manually as Pizza is having issues, but I'll cross-check elsewhere. Code all looks simple and solid.

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/AttachedNote.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { useStore } from "pages/FlowEditor/lib/store";
+import React from "react";
+
+export const AttachedNote: React.FC<{
+  note: string;
+}> = ({ note }) => {
+  const showNotes = useStore((state) => state.showNotes);
+
+  // TODO also return null if templated flow and *not* attached to a templated node?
+  if (!showNotes) return null;
```
DafyddLlyr, line 11: Just spotted this in the next PR, but it's most relevant here.

We check `showNotes` internally here, but also at every call site. It's cleaner and simpler I think to just always have `<AttachedNote/>` without conditions in each parent, and internally here choose to return `null`. This means less duplication and easier to change the condition (as noted here).

jessicamcinchak: Yep good spot - cleaned up duplication!

PR #7268, opened 2026-08-26
fix(editor): Populate required `orderedFlow` before loading source templates

DafyddLlyr:
## What's the problem?
Airbrake has been reporting a recurring `TypeError: Cannot read properties of undefined (reading 'edges')` error for about a month. Error tracked here - https://planx.airbrake.io/projects/329753/groups/4381666065751580350

## What's the cause?
I believe this regression was introduced in #7003 . This set the `<Customise/>` tab as default for source templates but did not update the route files to also set `orderedFlow` before the page loaded, or manage loading states for a missing `flow`.

## What's the solution?
Calculate `orderedFlow` once upstream, and compute it before loading either source templates or templated flows. This means by the time the `<Customise/>` panel loads it always has data ready.

I've also updated this component to directly read from `orderedFlow` instead of re-sorting the `flow` in order to show customised nodes.

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Customisations/index.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Customisations/index.tsx
@@ -2,37 +2,26 @@ import { gql, useSubscription } from "@apollo/client";
 import Box from "@mui/material/Box";
 import List from "@mui/material/List";
 import Typography from "@mui/material/Typography";
-import { sortIdsDepthFirst } from "@planx/graph";
 import DelayedLoadingIndicator from "components/DelayedLoadingIndicator/DelayedLoadingIndicator";
 import { useStore } from "pages/FlowEditor/lib/store";
-import React, { useEffect } from "react";
 
 import { CustomisationCard } from "./CustomisationCard";
 import type { FlowEdits } from "./types";
 
 const Customisations = () => {
-  const [flowId, flow, isTemplatedFrom, isTemplate, setOrderedFlow] = useStore(
+  const [flowId, isTemplatedFrom, isTemplate, orderedFlow] = useStore(
     (state) => [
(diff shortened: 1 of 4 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Sidebar/Customisations/index.tsx
```diff
@@ -2,37 +2,26 @@ import { gql, useSubscription } from "@apollo/client";
...
-  const customisableNodeIds = new Set(
-    Object.entries(flow)
-      .filter(([_nodeId, nodeData]) => Boolean(nodeData.data?.isTemplatedNode))
-      .map((entry) => entry[0]),
-  );
-  const sortedCustomisableNodeIds =
-    sortIdsDepthFirst(flow)(customisableNodeIds);
+  const sortedCustomisableNodeIds = (orderedFlow ?? [])
+    .filter((node) => Boolean(node.data?.isTemplatedNode))
+    .map((node) => node.id);
```
jessicamcinchak, line 24: Nice simplification ➕

apps/editor.planx.uk/src/routes/_authenticated/app/$team/$flow/-route.test.tsx
```diff
@@ -121,4 +122,66 @@ describe("connectToFlowRoute (ShareDB connection guard)", () => {
...
+        templatedFrom: "",
+        template: undefined,
+        isPattern: false,
+      });
+
+      await connectToFlowRoute("lambeth", "someFlow");
+
+      expect(mockSetOrderedFlow).not.toHaveBeenCalled();
+    });
+  });
```
jessicamcinchak, line 186: 👍

jessicamcinchak (approved):
Good spot & fix, thanks for extra test coverage here too!

PR #7269, opened 2026-08-26
feat[alt notes]: support internal notes on options

jessicamcinchak:
- Adds `notes` prop to `BaseOptionsEditor`
- Displays notes attached to option graph nodes

See here: https://storybook.7269.planx.pizza/?path=/story/planx-components-question--with-editor (four inputs per option is a lot but not quite as noisy as I was anticipating as long as we're happy to keep this note to a single-line?)


apps/editor.planx.uk/src/@planx/components/Option/model.tsx
```diff
@@ -43,6 +44,7 @@ export const optionValidationSchema = object({
     text: string().required().trim(),
     val: string(),
     exclusive: mixed().oneOf([true, undefined]),
+    notes: string(),
```
jessicamcinchak, line 47: By naming the same as internal notes prop, this is automatically picked up by `sanitiseNotes` on flatten. So worth it for convenience even if plural `notes` is a little semantically annoying to repeat (at least to me 😅)!

e2e/tests/ui-driven/src/helpers/addComponent.ts
```diff
@@ -453,6 +455,8 @@ async function createComponentOptionsWithDataValues(
     await page
       .getByRole("option", { name: option.dataValue, exact: true })
       .click();
+    // Close the 'data field' autocomplete
+    await page.keyboard.press("Escape");
```
jessicamcinchak, line 459: See https://opensystemslab.slack.com/archives/C01E3AC0C03/p1788221424129119 - thanks again Joe for extra eyes here!

joecarver (approved):
Nice!

freemvmt (approved):
Looks good to me! ✨

PR #7271, opened 2026-08-27
feat[alt notes]: handle visibility of Note components in templated flows

jessicamcinchak:
- Hides `Note` component types from the graph when: 
  - "Show notes" toggle is off
  - It's a templated flow and the `Note` is _not_ within a templated folder
- Decision to "hide" rather than strip out completely on templated flow creation is important distinction here as we do _not_ want to have consider (or ever "re-map") stripped out notes when reconciling templated flow updates, etc! 

**Not yet handled here:**
- Attached internal notes visibility on _other_ component types' graph nodes in templated flows. Maybe connect to `disabled` prop? Want to double-check expectations here with content team, and if "hiding" only applies to graph or in-modals too in this case?

**Testing:**
- Mock source template has multiple `Notes`
[Screencast from 2026-08-26 21-35-20.webm](https://github.com/user-attachments/assets/53ed050c-4422-497c-ad87-2c851423978d)
- Mock templated flow onlly displays `Notes` that are in the templated folder
[Screencast from 2026-08-26 21-34-11.webm](https://github.com/user-attachments/assets/2a687293-6045-4a61-b81f-86162b8f18d1)


apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Node.tsx
```diff
@@ -36,6 +44,22 @@ const Node: React.FC<any> = (props) => {
...
+    flow,
+    indexedParent,
+  );
+  const parentIsChildOfTemplatedInternalPortal =
+    nodeIsChildOfTemplatedInternalPortal(flow, indexedParent);
+  const renderNoteComponent =
+    showNotes &&
+    (!isTemplatedFrom ||
+      parentIsTemplatedInternalPortal ||
+      parentIsChildOfTemplatedInternalPortal);
```
DafyddLlyr: This gets calculated for each `Node` (a full scan of `orderedFlow`), but it's only ever needed for `Note` components. 

We should either move it into the `case TYPES.Note:` block, or make a specific new hanger (instead of `Question`) which handles this. I think the former is more consistent and simpler!

jessicamcinchak: Good catch, thanks & sorry I was a bit lazy with initial implementation of this! 

Now updated to be scoped to `TYPES.Note` block, agree specific hanger component is future goal but still relying on `Question.isNote` SCSS sticky styles so not going to untangle just yet if that's alright!

DafyddLlyr (approved):

PR #7273, opened 2026-08-27
fix: Ensure error boundary always has access to the component call stack

DafyddLlyr:
## What's the problem?
One of the largest open issues on Airbrake (59 occurrences) is just `Error: undefined`. This is essentially unactionable as there's no associated call stack we can dig into. We can't understand the root cause - it could be one issue, or many disparate issues. Please see occurrences here - https://planx.airbrake.io/projects/329753/groups/4345444345772610066

## What's the cause?
This is happening because `logger.notify(error)` is getting called in the body of `<ErrorFallback/>`, which doesn't have access to the call stack unless an existing `Error` object was passed in (but we can see it just receives and `undefined). The [react-error-boundary docs](https://github.com/bvaughn/react-error-boundary#logging-errors) show that we should report from the boundary component's `onError(error, info)` callback prop, where `info.componentStack` is available, not from the `<ErrorFallback/>` UI.

## What's the solution?
I've made a basic wrapper (`AppErrorBoundary` - better names welcome!) which enforces this pattern - we always call `logError()` to hit Airbrake with the full information.

All instances of `<ErrorBoundary/>` have now been swapped out for the new `<AppErrorBoundary/>`,  and some imports tidied up in these files.

## Next steps...
So the `undefined` errors will continue to throw at the same rate! But now each time there will be a proper call stack in place allowing us to actually diagnose and fix these issues.

### DIFF
--- a/apps/editor.planx.uk/src/components/Error/AppErrorBoundary.tsx
+++ b/apps/editor.planx.uk/src/components/Error/AppErrorBoundary.tsx
@@ -1,10 +1,11 @@
 import React, { type ComponentType, type PropsWithChildren } from "react";
-import {
-  ErrorBoundary,
-  type ErrorBoundaryProps,
-  type FallbackProps,
-  type OnErrorCallback,
+import type {
+  ErrorBoundaryProps,
+  FallbackProps,
+  OnErrorCallback,
 } from "react-error-boundary";
+// eslint-disable-next-line no-restricted-imports
+import { ErrorBoundary } from "react-error-boundary";
 
 import ErrorFallback from "./ErrorFallback";
 import { logError } from "./logError";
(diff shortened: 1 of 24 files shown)


apps/editor.planx.uk/src/components/Error/AppErrorBoundary.tsx
```diff
@@ -0,0 +1,42 @@
...
+  onError?: OnErrorCallback;
+  onReset?: ErrorBoundaryProps["onReset"];
+  resetKeys?: ErrorBoundaryProps["resetKeys"];
+}>;
+
+/**
+ * Wraps `react-error-boundary` so that every caught error is reported
+ * to Airbrake with its React component stack via logError()
+ */
+export const AppErrorBoundary: React.FC<AppErrorBoundaryProps> = ({
```
DafyddLlyr, line 25: The alternative here would be to pass the prop `onError={logError}` to every instance of `<ErrorBoundary/>`.

I've opted for this approach + ESLint rules as it's more robust and sustainable I think than always remembering to include an optional prop.

jessicamcinchak (approved):
Thanks for taking a look at this one!

PR #7276, opened 2026-08-27
docs(iac): Check both validation and routing records when onboarding or migrating domains

DafyddLlyr:
Coventry have only set up the validation record (which we checked) but not the routing record (not checked).

This PR updates to docs to include this exhaustive step - meaning we fail early and give the councils feedback before proceeding.


freemvmt (approved):
Nice! I definitely thought I included this originally, cause I've been doing it, but must've just been a thing in my head 🫠

DafyddLlyr:
In all fairness one should be a proxy for the other, but I think we've just been caught out here as Coventry got two emails but only actioned one.

PR #7278, opened 2026-08-27
fix: internal back button positioning

joecarver:
can be tested here https://7278.planx.pizza/lambeth/project-description-accessibility-testing/published?analytics=false


DafyddLlyr (approved):
Works as expected...! Thanks for the quick fix. Did you manage to locate the regression at all?

PR #7279, opened 2026-08-27
[api] require user-uploaded files to be scanned before serving

freemvmt:
Enforce that `user-data` objects (uploaded via `/file/` POST routes) have been scanned and tagged by Scanii before they can be served (by the `/file/` GET routes). See [ticket](https://trello.com/c/g4q2hI9k).

We're introducing these changes as a side-quest before enabling new file types (see this other [ticket](https://trello.com/c/kkt7dKkB)).

## TODO

- [x] Add tests
- [x] Prepare a script to manually backwrite Scanii tags onto all objects in `user-data`, after which we could remove the `ENFORCE_SCAN_FROM` var (no urgency here)


apps/api.planx.uk/modules/file/service/getFile.ts
```diff
@@ -1,40 +1,132 @@
...
+ * S3 signals a missing object as `NoSuchKey` on GetObject and `NotFound` on HeadObject,
+ * the latter without a usable message - fall back to the response status.
+ */
+const isNotFound = (error: unknown): boolean => {
+  if (!(error instanceof Error)) return false;
+  if (error.name === "NoSuchKey" || error.name === "NotFound") return true;
+
+  const { $metadata } = error as { $metadata?: { httpStatusCode?: number } };
+  return $metadata?.httpStatusCode === 404;
+};
```
DafyddLlyr, line 134: It's possible there's an S3 error type we can import or validate against here to avoid this casting.

apps/api.planx.uk/modules/file/service/uploadFile.ts
```diff
@@ -33,13 +34,19 @@ export const uploadPrivateFile = async (
...
+   */
+  { scanExempt = false }: { scanExempt?: boolean } = {},
 ) => {
   const s3 = s3Factory();
 
   const { params, key, fileType } = generateFileParams(file, filename, filekey);
 
   params.Metadata = {
     is_private: "true",
+    ...(scanExempt && { [SCAN_EXEMPT_METADATA_KEY]: "true" }),
```
DafyddLlyr: Just to confirm - these values must be strings right and booleans are not accepted?

apps/api.planx.uk/modules/file/service/scanStatus.ts
```diff
@@ -0,0 +1,97 @@
...
+  const { TagSet } = await s3.getObjectTagging({
+    Bucket: process.env.AWS_S3_BUCKET,
+    Key: key,
+  });
+  const tags = Object.fromEntries(
+    (TagSet ?? []).map(({ Key, Value }) => [Key, Value]),
+  );
+
+  // tagged, but not by Scanii (unexpected, but we guard against it)
+  if (!tags[SCANII_ID_TAG]) return PENDING;
```
DafyddLlyr, line 77: 👍

apps/api.planx.uk/modules/file/service/uploadFile.ts
```diff
@@ -33,13 +34,19 @@ export const uploadPrivateFile = async (
   file: Express.Multer.File,
   filename: string,
   filekey?: string,
+  /**
+   * Optionally mark the file as exempt from Scanii checks. Should only be used for files
+   * the API generates itself from trusted data - never for anything a user uploads.
+   */
+  { scanExempt = false }: { scanExempt?: boolean } = {},
```
DafyddLlyr: nit: I'd lean towards a scarier name here, like React's `dangerouslySetInnerHTML` or something.

apps/api.planx.uk/modules/send/utils/exportZip.ts
```diff
@@ -240,7 +241,17 @@ export class ExportZip {
     const decodedS3Key = decodeURIComponent(s3Key);
     const file = await getFileFromS3(decodedS3Key);
 
-    if (file) this.zip.addFile(name, file.body);
+    // Fail the whole zip rather than silently handing a council an incomplete submission.
+    // A pending scan should resolve, so failing here lets the send event's Hasura retry kick in.
```
DafyddLlyr, line 245: 👍

apps/api.planx.uk/modules/file/docs.yaml
```diff
@@ -157,9 +180,11 @@ paths:
...
-        "204":
-          description: Successful deletion
+        "200":
+          $ref: "#/components/responses/DownloadFile"
         "404":
           $ref: "#/components/responses/MissingFile"
         "500":
           $ref: "#/components/responses/ErrorMessage"
+        "503":
+          $ref: "#/components/responses/FileScanPending"
```
DafyddLlyr, line 190: Good call! https://http.dog/503

doc/how-to/how-to-identify-missing-files.md
DafyddLlyr, line 1: Thanks for the updates here!

doc/how-to/how-to-upgrade-nodejs.md
DafyddLlyr, line 1: Again - appreciate this docs update 👍

infrastructure/data/README.md
DafyddLlyr, line 1: Great!

.env.example
```diff
@@ -51,6 +51,9 @@ FILE_API_KEY_NEXUS=👻
 FILE_API_KEY_SOUTHWARK=👻
 FILE_API_KEY_TEWKESBURY=👻
 
+# ISO8601 date from which API refuses to serve user-data files not scanned + tagged by Scanii
+ENFORCE_SCAN_FROM=""
```
DafyddLlyr: We also need to update the `.env` files on S3 so Pizzas have access to this value (they upload to staging S3).

DafyddLlyr (approved):
Solid - really appreciate the level of docs and comments here!

apps/api.planx.uk/modules/file/service/getFile.ts
freemvmt: Indeed we can resort to looking for the base error class `S3ServiceException` if the first two don't match. Nice suggestion!

apps/api.planx.uk/modules/file/service/uploadFile.ts
freemvmt: Yes exactly. `params` is typed as `PutObjectCommandInput`, which when you drill down has `Metadata?: Record<string, string>` - the SDK sends these over the wire as `x-amz-meta-xxx` headers, so naturally the value should be a string (and metadata is stored as lowercase so we stick to that too)

more here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html

apps/api.planx.uk/modules/file/service/uploadFile.ts
freemvmt: yep this is a good shout! went with `dangerouslySkipMalwareScan` :skull_and_crossbones:

DafyddLlyr: Yep this is great and suitably scary 👻 ☠️ ⚠️

.env.example
freemvmt: Thanks for this provocation. I was feeling a bit fuzzy about all the different environments and what behaviour we're expecting on each, so I got into this a bit more.

Most immediately I realised that pizzas don't upload to the `user-data-xxx` bucket on staging, but another bucket which we've manually provisioned (i.e. it's not in Pulumi/IaC) called `pizza-user-uploads`, as per the `.env`.

We're not scanning stuff that lands in that bucket. Do we want to? I'm leaning towards _yes_, for various reasons, e.g. parity of pizzas with other envs, minimising attack surface, peace of mind etc. What do you think?

Should be as easy as adding an extra trigger to the staging lambda, I think.

I got claude to make a table of all the different contexts for my own sanity:

| Environment | `NODE_ENV` | S3 target | Scanii? | Guard |
| --- | --- | --- | --- | --- |
| Local dev | `development` | Minio (`pizza-user-uploads` dir) | no | **off** |
| API vitest | `test` | client mocked | n/a | off (stubbed per-test) |
| e2e / integration CI | `test` | Minio | no | **off** |
| Pizza (per-PR) | `pizza` | real S3, `pizza-user-uploads` | **no** | **off** |
| Staging | `staging` | real S3, `user-data-*` | yes | **on** |
| Production | `production` | real S3, `user-data-*` | yes | **on** |

DafyddLlyr: Very good spot - this is for sure an existing gap. Yes, I'd vote to manually add an extra trigger to the staging lambda here for sure.

It's entirely possible that somebody will want to test the a pipeline which goes from Pizza -> council, so we do need this bucket covered by Scanii.

The varying envs are challenging - we have both `NODE_ENV` and `APP_ENVIRONMENT`, and there's various idiosyncrasies here and there we should look at rationalising sometime soon!

freemvmt: Nice, glad you agree. I tried adding the bucket as trigger, but it can't tag the objects because that stack doesn't have appropriate permissions (I forgot this but you do have to give the bucket name on instantiation).

So we'll need to make a new Scanii stack specific to the pizza uploads bucket.

PR #7281, opened 2026-08-28
fix: throw notFound in submissions log route if user not authorised

zz-hh-aa:
Also made Subscription page follow this pattern while I was here


DafyddLlyr (approved):
Thanks!

PR #7282, opened 2026-08-28
fix: make submission availability a condition of resubmit button

zz-hh-aa:
Testing on staging with older data is surfacing various cases I wasn't able to test locally... Previously only checked for `isPlatformAdmin` and not `isSubmissionAvailable` also! Pulls logic up into parent component because the boolean value is used for conditional rendering in various places.


DafyddLlyr (approved):
Nice catch - this highlights a few things I think...

- We need test coverage here to that we don't introduce regressions like this. There's quite a number of permutations a submission can be in
- Storybook stories for each permutation should allow us to spot and manually test these issues and build on top of this with confidence going forwards

It's generally easier to do both of these upfront first - we can generate a series of mock submissions which match all permutations and use them for both RTL tests and stories.

zz-hh-aa:
Thanks @DafyddLlyr, yep I've got a local branch going with tests and storybook! Hopefully open for review later today.

PR #7291, opened 2026-08-28
fix: grouped submissions log nits

zz-hh-aa:
They're all small so made one PR for first round of bugs / nits flagged in [testing thread](https://opensystemslab.slack.com/archives/C5Q59R3HB/p1787921006379939?thread_ts=1787919522.714559&cid=C5Q59R3HB)

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/EventsLogGrouped.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/EventsLogGrouped.tsx
@@ -91,6 +91,9 @@ const EventsLogGrouped: React.FC<EventsLogGroupedProps> = ({
       field: "address",
       headerName: "Address",
       width: 250,
+      columnOptions: {
+        valueFormatter: (value: string | null) => value || "Sanitised",
+      },
     },
     {
       field: "consolidatedStatus",
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/EventsLogGrouped.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/EventsLogGrouped.tsx
@@ -79,7 +79,7 @@ const EventsLogGrouped: React.FC<EventsLogGroupedProps> = ({
     {
       field: "flowName",
       headerName: "Service",
-      width: 250,
+      width: 350,
       type: ColumnFilterType.SINGLE_SELECT,
       customComponent: (params) => <strong>{`${params.value}`}</strong>,
       columnOptions: {
@@ -90,7 +90,7 @@ const EventsLogGrouped: React.FC<EventsLogGroupedProps> = ({
     {
       field: "address",
       headerName: "Address",
-      width: 250,
+      width: 350,
       columnOptions: {
         valueFormatter: (value: string | null) => value || "Sanitised",
       },
@@ -98,7 +98,7 @@ const EventsLogGrouped: React.FC<EventsLogGroupedProps> = ({
     {
       field: "consolidatedStatus",
       headerName: "Status",
-      width: 125,
+      width: 200,
       type: ColumnFilterType.SINGLE_SELECT,
       customComponent: StatusChipGrouped,
       columnOptions: {
@@ -108,7 +108,7 @@ const EventsLogGrouped: React.FC<EventsLogGroupedProps> = ({
     {
       field: "eventCreatedAt",
       headerName: "Date",
-      width: 125,
+      width: 200,
       columnOptions: {
         valueFormatter: dateFormatter,
       },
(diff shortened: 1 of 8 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/EventsLogGrouped.tsx
```diff
@@ -91,6 +91,9 @@ const EventsLogGrouped: React.FC<EventsLogGroupedProps> = ({
       field: "address",
       headerName: "Address",
       width: 250,
+      columnOptions: {
+        valueFormatter: (value: string | null) => value || "Sanitised",
+      },
```
DafyddLlyr, line 96: I _think_ we may also need to account for address not provided. It's possible for a council to have a service which doesn't ask this (e.g. the service shown by Doncaster last S&T).

Could we have three options here - 
 - Address
 - N/A (or just empty) - used if the record is not sanitised, but no address is present
 - Sanitised - used if there was an address, but we can no longer show it

Not a showstopper and we can fix forwards here - it might need a view update to check this status.

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/SubmissionDetailModal.tsx
```diff
@@ -121,6 +122,9 @@ const SubmissionDetailModal: React.FC<SubmissionDetailModalProps> = ({
   const events = data?.submissions || [];
   const latestEvent = events[0];
   const submittedAt = getSubmittedAt(events);
+  const isSanitised = submittedAt
+    ? hasBeenSanitised(new Date(submittedAt))
+    : false;
```
DafyddLlyr, line 120:
```suggestion
  const isSanitised = submittedAt && hasBeenSanitised(new Date(submittedAt))
```

`hasBeenSanitised()` already returns a boolean.

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/utils.ts
```diff
@@ -39,3 +42,10 @@ export const canResubmit = (
     !NON_RESUBMITTABLE_EVENT_TYPES.includes(eventType)
   );
 };
+
+export const hasBeenSanitised = (submittedAt: Date) => {
+  const sixMonthsAgo = new Date();
+  sixMonthsAgo.setMonth(sixMonthsAgo.getMonth() - 6);
+
+  return submittedAt < sixMonthsAgo;
+};
```
DafyddLlyr, line 60: nit: Magic number..! I think we have access to `RETENTION_PERIOD_MONTHS` here or another value.

Either way, I think a stronger pattern is to should encode the retention period, and compare this with `submittedAt` (same exact calcuation and result, just better semantics and maintainability?).

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/SubmissionDetailModal.tsx
zz-hh-aa: `submittedAt` is potentially undefined here which is why I check first--but I could move it to within the `hasBeenSanitised` function if that seems more appropriate?

DafyddLlyr: The idea here is that the `&&` is doing the boolean check for us, without needing the wrapping ternary.

If `submittedAt` is falsy, it will short-circuit and not evaluate the second half. It also acts as a type guard so that `submittedAt` is always truthy within the `new Date()` caller.

Initially, it's a `string | undefined`
image

After the `&&` it's narrowed to just `string`
image

DafyddLlyr: To be honest this is all functionally the same thing...! I feel the second version is more concise and readable but this is always somewhat subjective 😅 

Very much non-blocking, just wanted to flag an alternative pattern here as we're repeating a similar method with `submissionExpirationDate` and `isSubmissionAvailable` below.

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/utils.ts
zz-hh-aa: Good call, done here [ead21f0](https://github.com/theopensystemslab/planx-new/pull/7291/commits/ead21f05062f0f2f4be033f2b9f5f900ff70aecb)

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/SubmissionDetailModal.tsx
zz-hh-aa: Gonna leave as is--even though the `&&` does the boolean check--to avoid the downstream type issue

apps/editor.planx.uk/src/pages/FlowEditor/components/Submissions/components/EventsLogGrouped.tsx
zz-hh-aa: Good shout on view changes, opened here https://github.com/theopensystemslab/planx-new/pull/7317

jessicamcinchak (approved):
Changes all look good here, including responses to first round of feedback - didn't spot anything else !

PR #7292, opened 2026-08-28
fix(ci): Update dependabot to run correctly for new monorepo setup

DafyddLlyr:
## What's the problem?
Dependabot has not been reliably picking up packages since we moved to the monorepo. We can see error logs as follows for each app - 

```sh
Dependabot could not run pnpm to update your dependencies due to a configuration error
Please review your pnpm setup. Dependabot encountered the following error when running it:

Updating workspaces from inside a workspace subdirectory is not supported. Both `pnpm-lock.yaml` and `pnpm-workspace.yaml` exist in a parent directory. Dependabot should only update from the root workspace.
```

Source: https://github.com/theopensystemslab/planx-new/network/updates/1544330629

Some PRs do succeed to open, but most of the outstanding dependabot issues show the following - 

image

## What's the cause?
We're now using a single pnpm workspace (one root `pnpm-lock.yaml` and one `pnpm-workspace.yaml`), and Dependabot cannot update `npm` from a subdirectory. This means every per-app job was erroring with as above.

## What's the solution?
Collapse the per-app setup in favour of a single root `npm` configuration, pointing at the root `pnpm-workspace.yaml`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,10 +19,14 @@ updates:
       sharedb:
         patterns: ["sharedb", "@types/sharedb"]
     ignore:
-      # Node version is managed manually, only allow patches
-      - dependency-name: "@types/node"
+      # Majors only across the workspace so we don't fall too far behind
+      # Security updates bypass this and still raise PRs for CVEs
+      - dependency-name: "*"
         update-types:
-          ["version-update:semver-major", "version-update:semver-minor"]
+          ["version-update:semver-patch", "version-update:semver-minor"]
+      # Node version is managed manually - also ignore major versions
+      - dependency-name: "@types/node"
+        update-types: ["version-update:semver-major"]
     # Matches pnpm's `minimumReleaseAge` in pnpm-workspace.yaml
     cooldown:
       default-days: 7
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,84 +2,28 @@
 # fixes for vulnerabilities/CVEs (i.e. security updates) are handled automatically as per repo settings
 version: 2
 
-# TODO: increase use of groups to collect version upgrades in logical fashion and reduce PRs
-multi-ecosystem-groups:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -100,102 +44,20 @@ updates:
...
     open-pull-requests-limit: 1
 
-  # E2E tests
-  - package-ecosystem: "npm"
-    directory: "/e2e"
-    schedule:
-      interval: "monthly"
-    labels: ["e2e", "tests", "pnpm"]
-    commit-message:
-      prefix: "[skip pizza] "
```
joecarver, line 188: by removing this prefix and not adding it to the root config, does it mean we'll always be rebuilding the pizza where we weren't before? I can see its checked in the `pull-request.yml` workflows

.github/dependabot.yml
```diff
@@ -100,102 +44,20 @@ updates:
...
-    ignore:
-      - dependency-name: "*"
-        update-types:
-          ["version-update:semver-patch", "version-update:semver-minor"]
```
joecarver, line 192: similar question here, before we were ignoring all dependencies for patch or minor updates, now it looks like we are just ignoring `@types/node`?

joecarver (commented):
Got a couple of questions around potentially increasing PR/CI volume as a result of these changes... I might be misunderstanding though!

freemvmt (approved):
Ah amazing thanks for fixing this! Was at the edge of my awareness also :)

.github/dependabot.yml
DafyddLlyr: Yes - previously we were skipping the pizza for certain types of deps.

We don't need a Vultr instance for e2e test updates or Hasura introspection tests (there's nothing novel on these envs - the changes would only run on CI).

We have to omit this option now as there's currently no way for us to segregate out a "e2e dep" vs a "api dep" now that we're using workspaces.

PR #7293, opened 2026-08-31
fix(api): Update Resend config to allow Pizza and Staging "submit" emails

DafyddLlyr:
## What's the problem?
As noted [in this thread](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1787931852552329) (OSL Slack) we can't send "submit" emails from staging / pizza, only production. This isn't ideal as this is something we need to test, and we need positive status on "send to email" events in order to test the submissions pages.

We intentionally limit "welcome" emails as we don't want to trigger these on each data-sync event, but having access to "submit" emails is very helpful.

## What's the cause?
Resend was using an old API key on Pizzas, limiting the addresses we could use to submit. On staging there's a mismatch between the domain and `RESEND_FROM_ADDRESS` which blocks emails.

## What's the solution?
- Add and verify `notifications.planx.dev` for the staging Resend API key (see screenshot below)
- Update `RESEND_FROM_ADDRESS` to match each env's verified domain 
  - Prod → `notifications.planx.uk`
  - Staging, dev, pizza → `notifications.planx.dev`
- Update API key on `.env` (via S3)
- Refactored the "welcome" template route to make the env guarding more explicit

## Testing
 - Adding a new user on this Pizza does not trigger a "welcome" email from Resend
   - Hasura logs show this as a skipped success (HTTP 200)
 - Triggering a "submit" event does send an email (configure your own address for a "send to email" event)
   - This is safe as we do not sync council emails from production - there's no increased risk of us sending councils emails from staging/dev/pizza


apps/api.planx.uk/modules/sendEmail/routes.ts
```diff
@@ -35,8 +36,9 @@ router.post(
 );
 
 router.post(
-  "/send-email/:template(welcome)",
```
DafyddLlyr, line 38: Changed to just `/welcome` as we never need access to this variable.

infrastructure/application/Pulumi.production.yaml
```diff
@@ -84,8 +84,7 @@ config:
     secure: AAABADAL9YkrUfJaDga7MaVRD/dZaElsnFnnC+7IvqqlgBFywXFwVKnSMbVZ1SMFoitG8C6bKHhl2BPwOHcuvkzMKblwYzvu
   application:ordnance-survey-api-key:
     secure: AAABANmVR/iyxoja6JGBOBjhlijbMiqdqpNRwqS2p+ZusuwiYdKBACIi5y/vi2HMg7+ydXWR0SPVqtQuK5O/tA==
-  application:resend-from-address:
-    secure: AAABACSDFBG/Z8fAVReyp/KD41l9YAsuqldyHSYDCBmZaY7EqAgsod8TjdXYLvOR4CQ8c07XTJ8UsirXOwQ=
+  application:resend-from-address: noreply@notifications.planx.uk
```
DafyddLlyr, line 87: Intentionally no longer a secret - these emails are not sensitive, and are shown to users who receive these messages.

infrastructure/application/Pulumi.staging.yaml
```diff
@@ -81,8 +81,7 @@ config:
     secure: AAABAMqkRWGvu4THnzddRYpNddFfIFFKCqEE3gm9DkYuVJRCjZBCWpKogsT+X4j3R2f9V8OyeZxHK53yIpl8rQ==
   application:resend-api-key:
     secure: AAABAFmzQejlInptwqtYYDEGS+U+aizq8a/+vGvFei2iMyHKwiwCgKmuGoN3o+7D/tbH0wgZIWtEW5p6nZ5p7DZTyyw=
-  application:resend-from-address:
-    secure: AAABAIyo5sjwMg+bJ1iRAU44fOPdQNsdpJqTdcazQibTLktJdxQHi+hDi/0rIyrbviDizNaMNtrWcf9EZE0=
+  application:resend-from-address: noreply@notifications.planx.dev
```
DafyddLlyr, line 84: Image

planx.dev now set up as a domain on Resend (via CloudFlare DNS records).

apps/api.planx.uk/modules/sendEmail/controller.ts
```diff
@@ -86,14 +86,6 @@ export const confirmationEmailController: ConfirmationEmail = async (
...
 export const resendEmailController: ResendEmail = async (_req, res, next) => {
   const { payload } = res.locals.parsedReq.body;
-  const { template } = res.locals.parsedReq.params;
-
-  // call sendEmail only in production or when testing
-  if (!["production", "test"].includes(process.env.APP_ENVIRONMENT!)) {
-    return res.status(200).send({
-      message: `Skipping ${template} email: APP_ENVIRONMENT is not production or test`,
-    });
-  }
```
DafyddLlyr, line 96: Moved to a middleware as it took me a minute to work out this only applies to the "welcome" template. It feels more explicit to have it up at the route level.

apps/api.planx.uk/modules/sendEmail/routes.ts
joecarver: does this change mean we can remove `case "welcome":` from `auth/middleware useSendEmailAuth` (line 83) as we are matching for the welcome route exacty?

joecarver (approved):
Looks good! Manual tests both behaved as expected for me and the new test coverage is also v helpful. Just a very tiny comment about dead code in a switch case but not super important.

apps/api.planx.uk/modules/sendEmail/routes.ts
DafyddLlyr: We can't do this with the current setup as we use an exhaustive switch (thanks to `handleInvalidTemplate()` accepting a `never`).

It would actually be better though to be clear and consistent here though I guess and stick with `useSendEmailAuth()`.

DafyddLlyr: Hit merge too soon here 🙃 One liner to fix this here - https://github.com/theopensystemslab/planx-new/pull/7303

PR #7295, opened 2026-08-31
fix(ci): Fix Docker collisions on Vultr

DafyddLlyr:
## What's the problem?
This aims to resolve a pretty common CI failure - 

```sh
Container storybook Error response from daemon: Conflict. The container name "/storybook" is already in use by container "c7d78faf781cdd582ff4a8af56a95a1e412b6139bfb74d567780187daca0ecda". You have to remove (or rename) that container to be able to reuse that name. 
Error response from daemon: Conflict. The container name "/storybook" is already in use by container "c7d78faf781cdd582ff4a8af56a95a1e412b6139bfb74d567780187daca0ecda". You have to remove (or rename) that container to be able to reuse that name.
2026/08/31 09:47:59 Process exited with status 1
```

## What's the cause?
We get this when there's two pushes within relatively short succession (second push starts before 1st CI run completes). 

Each PR runs on a single Vultr machine, with fixed container and network names. Two pushes within one ~9-min build window leads to two runs that both run `down → build → up` on the same host. This race condition is what leads to the error - two instances running at once, but trying to share container and network names.

```mermaid
sequenceDiagram
    participant Dev as PR pushes
    participant D1 as Deploy 1
    participant D2 as Deploy 2
    participant H as Pizza

    Dev->>D1: Push 1
    activate D1
    Note over D1: Build 1
    Dev->>D2: Push 2 (shortly after)
    activate D2
    Note over D2: Build 2
    D1->>H: ✅ Creates network + Storybook, Caddy etc
    deactivate D1
    D2->>H: ❌ "Container /storybook already in use"
    deactivate D2
 ```

## What's the solution?
Two parts here - 
 1. Add concurrency to the GHA - a second push should kill the first run, there's no need to proceed with tests, builds etc. This doesn't stop `create.sh` / `update.sh` scripts from running on the Vultr machine though, it just drops the SSH connection.
 
 2. Use a lock to make sure subsequent runs of the script can't collide and overlap. We just need to wait for the previous (short) `docker compose` command to finish before starting on the next one.

### DIFF
--- a/scripts/pullrequest/create.sh
+++ b/scripts/pullrequest/create.sh
@@ -1,31 +1,31 @@
 #!/usr/bin/env bash
 set -o errexit -o pipefail
 
-# run from project root
+# Run from project root
 cd "$(dirname $0)/../.."
 
 echo "root:$SSH_PASSWORD" | chpasswd
 
-# check if swap space is available - see link for more on updating swap:
+# Check if swap space is available - see link for more on updating swap:
 # https://www.digitalocean.com/community/tutorials/how-to-add-swap-space-on-ubuntu-22-04
 swapon --show
 
-# caddy runs up against kernel buffer limits of host OS, so we increase them
-# see https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes
+# Caddy runs up against kernel buffer limits of host OS, so we increase them
+# See https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes
 sysctl -w net.core.rmem_max=7500000
 sysctl -w net.core.wmem_max=7500000
 
-# ensure docker does not pull caddy image via ipv6 (no subnet configured, it will fail)
+# Ensure docker does not pull caddy image via ipv6 (no subnet configured, it will fail)
 sysctl -w net.ipv6.conf.all.disable_ipv6=1
 sysctl -w net.ipv6.conf.default.disable_ipv6=1
 
-# set env for this shell
+# Set env for this shell
 set -o allexport
 source .env.pizza
 DOCKER_BUILDKIT=1
 set +o allexport
 
-# fallback to building caddy image if pull fails
+# Fallback to building caddy image if pull fails
 PIZZA_FAILOVER=""
 if docker pull "$VULTR_CR_URN/caddy-vultr:latest"; then
   echo "Caddy image pulled successfully"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: scripts/pullrequest/create.sh [hunk 1]
```diff
@@ -34,9 +34,22 @@ else
...
-  -f docker-compose.pizza.yml $PIZZA_FAILOVER \
-  -f docker-compose.seed.yml \
-  up --build --wait
+function compose() {
+  docker compose \
+    -f docker-compose.yml \
+    -f docker-compose.pizza.yml $PIZZA_FAILOVER \
+    -f docker-compose.seed.yml \
+    "$@"
+}
```
DafyddLlyr, line 43: Pattern pinched from here - https://github.com/theopensystemslab/planx-new/blob/main/scripts/start-containers-for-tests.sh#L10-L17

scripts/pullrequest/create.sh
```diff
@@ -34,9 +34,22 @@ else
...
+
+# Build outside the lock. Images don't collide, so this keeps the lock short-lived
+compose build
+
+# Use a lock to ensure overlapping runs of this script (pushes to GH in short succession) can't collide
+exec 9>/tmp/pizza-deploy.lock
+flock 9
```
joecarver: do we need to timeout on `flock` in case one of the following commands hangs? or are we comfortable that the overall CI timeout will handle this?

scripts/pullrequest/create.sh
```diff
@@ -34,9 +34,22 @@ else
...
+function compose() {
+  docker compose \
+    -f docker-compose.yml \
+    -f docker-compose.pizza.yml $PIZZA_FAILOVER \
+    -f docker-compose.seed.yml \
+    "$@"
+}
+
+# Build outside the lock. Images don't collide, so this keeps the lock short-lived
+compose build
```
joecarver: question: trying to get my head round how this works and just want to double check if its safe to run the build outside of the lock... is the following a case that could happen?

- push commit 1, then push commit 2 shortly after
- commit 2 build completes before commit 1 build
- commit 1 has then set the tagged image as it finished later
- both commits then run `up` which uses the image built by commit 1
- pizza version is then running commit 1 but reports that its up to date

let me know if i've misunderstood anything here...

joecarver (commented):
Looks like an elegant fix to an annoying problem. Got a couple questions around concurrency safety though..

scripts/pullrequest/create.sh
DafyddLlyr: Yep good spot - your understanding here is right. I've moved the build inside the lock now to ensure we get the correct sequence of events.

DafyddLlyr: Failing on Vultr now 🤔 Moved back to draft whilst I figure it out.

PR #7296, opened 2026-08-31
feat: search UI & queries/hooks

joecarver:
### What's in this PR?
- Hooks & gql snippets for executing search queries
- New UI components for rendering search results
- Move template details logic to `useTemplateDetails` so it can be re-used in the new `TemplateDetailsPanel` as well as the existing `TemplateDetailsModal`

### To note
- Noticed a bug with the DB search implementation where e.g. 'app' matches results but 'appl' does not, due to the stem matching approach. Will fix this in a separate PR
- This is a fairly large PR, so I'd be happy to split it out if preferred. Could split it either as hooks vs UI components, or as initial results vs DetailsPanels  - also worth nothing that a good ~200+ lines of this is just a hook refactor into a separate file. Give me a shout for preferred review approach or if this is fine already!

### DIFF
--- a/apps/editor.planx.uk/src/pages/Explore/components/SearchListItemDetailActions.tsx
+++ b/apps/editor.planx.uk/src/pages/Explore/components/SearchListItemDetailActions.tsx
@@ -22,7 +22,7 @@ export const SearchListItemDetailActions: React.FC<
       <DialogActions
         sx={{
           bgcolor: "background.paper",
-          justifyContent: onClose ? "flex-end" : "flex-start",
+          justifyContent: "flex-end",
         }}
       >
         {onClose && (
--- a/apps/editor.planx.uk/src/pages/Explore/components/SearchListItemDetailActions.tsx
+++ b/apps/editor.planx.uk/src/pages/Explore/components/SearchListItemDetailActions.tsx
@@ -44,7 +44,8 @@ export const SearchListItemDetailActions: React.FC<
           <Button
             onClick={secondaryAction.onClick}
             disabled={secondaryAction.disabled}
-            sx={{ backgroundColor: "background.default" }}
+            variant="contained"
+            color="secondary"
           >
             {secondaryAction.label}
           </Button>
--- a/apps/editor.planx.uk/src/pages/Explore/components/SearchListItemDetailActions.tsx
+++ b/apps/editor.planx.uk/src/pages/Explore/components/SearchListItemDetailActions.tsx
@@ -7,18 +7,24 @@ import type { SearchResult } from "./SearchResult";
 
 interface SearchListItemDetailActionsProps {
   primaryAction?: SearchResult["primaryAction"];
+  secondaryAction?: SearchResult["secondaryAction"];
   onClose?: () => void;
 }
 
 export const SearchListItemDetailActions: React.FC<
   SearchListItemDetailActionsProps
-> = ({ primaryAction, onClose }) => {
-  if (!onClose && !primaryAction) return null;
+> = ({ primaryAction, secondaryAction, onClose }) => {
+  if (!onClose && !primaryAction && !secondaryAction) return null;
 
   return (
     <>
       <Divider />
-      <DialogActions>
+      <DialogActions
+        sx={{
+          bgcolor: "background.paper",
+          justifyContent: onClose ? "flex-end" : "flex-start",
+        }}
+      >
         {onClose && (
           <Button variant="contained" color="secondary" onClick={onClose}>
             Close
@@ -32,6 +38,15 @@ export const SearchListItemDetailActions: React.FC<
             disabled={primaryAction.disabled}
           >
             {primaryAction.label}
+          </Button>
+        )}
+        {secondaryAction && (
+          <Button
+            onClick={secondaryAction.onClick}
+            disabled={secondaryAction.disabled}
+            sx={{ backgroundColor: "background.default" }}
+          >
+            {secondaryAction.label}
           </Button>
         )}
       </DialogActions>
--- a/apps/editor.planx.uk/src/pages/Explore/components/SearchModal.tsx
+++ b/apps/editor.planx.uk/src/pages/Explore/components/SearchModal.tsx
@@ -153,6 +153,10 @@ export const SearchModal: React.FC<SearchModalProps> = ({ open, onClose }) => {
                 bgcolor:
                   index === tabIndex ? "text.primary" : "background.default",
                 color: index === tabIndex ? "background.default" : undefined,
+                "&:hover, &:focus, &:focus-visible, &:active": {
+                  bgcolor: index === tabIndex ? "text.primary" : "action.hover",
+                  color: index === tabIndex ? "background.default" : undefined,
+                },
               }}
             />
           ))}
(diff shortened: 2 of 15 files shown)


apps/editor.planx.uk/src/pages/Explore/components/SearchModal.tsx
```diff
@@ -0,0 +1,251 @@
...
+}
+
+const TABS: SearchTab[] = [
+  { label: "All flows", where: undefined },
+  {
+    label: "Templates",
+    where: {
+      is_template: { _eq: true },
+      can_create_from_copy: { _eq: true },
+    },
```
DafyddLlyr: It feels a bit awkward to be mixing up GQL / db logic here within the frontend component directly, it also leads to snake_case pollution here,

What about a map of variables to GQL filters (`FlowWhere`)?

```ts
export type FlowFilter = "all" | "templates" | "copyable";

const FILTER_WHERE: Record<FlowFilter, FlowsWhere | undefined> = {
  all: undefined,
  templates:  { is_template: { _eq: true },  can_create_from_copy: { _eq: true } },
  copyable:   { is_template: { _eq: false }, can_create_from_copy: { _eq: true } },
};
```

apps/editor.planx.uk/src/pages/Explore/components/useCopyFlowToTeam.ts
DafyddLlyr, line 1: This doesn't behave quite the same as the current copy functionality as there's no warning of uniqueness violation, just a generic failure. We should handle this consistently.

The answer might be to share the use of this hook across both instances where this is called. Happy for this to be a fix forward as the PR is pretty big as-is, and it's a relatively small tweak.

apps/editor.planx.uk/src/pages/Explore/components/useSearchFlows.ts
```diff
@@ -0,0 +1,91 @@
...
+      primaryColour: string | null;
+      logo: string | null;
+    };
+  };
+  publishedFlows: { hasSendComponent: boolean }[];
+  operations: {
+    createdAt: string;
+    actor?: { firstName: string; lastName: string };
+  }[];
+}
```
DafyddLlyr, line 45: Let's use camelCase for these interfaces, see comments about on `FlowsWhere` which I think will need to remain as snake_case.

DafyddLlyr (commented):
Very solid - well structured and easy to read!

A few minor comments directly on the code, and a few small usability / design comments here - 

 - Search box should be focussed when I load the modal so I can start typing right away
 - No handling of "no results found" state
 - I think we could make the search input more prominent on the Explore page, it feels quite invisible currently
 - Badges are not legible when in both focus and active states

https://github.com/user-attachments/assets/f472f77f-1dcd-4043-a261-3e2dec9fd182

 - Whitespace - quite a lot in the top right which is unused. Notion has the search bar going full width and then uses two columns below

Image

Image

- Maybe we should include `team.name` and / or `team.slug` in the search vector. I can imagine somebody trying to match on this (e.g. after watching a certain team demo a service at S&T).

ianjon3s:
Working brilliantly, very impressive stuff 😀

Couple of general style comments:
- The search modal needs a max width so it's usable on larger screens, I think `contentWide` would fit well
- Badges were changed to be circular due to feedback on the explore templates PR, the layout has also been changed slightly so that the action buttons sit below the badges when viewing template details, let's keep that consistent
image

apps/editor.planx.uk/src/pages/Explore/components/SearchListItemDetailActions.tsx
```diff
@@ -34,6 +40,15 @@ export const SearchListItemDetailActions: React.FC<
             {primaryAction.label}
           </Button>
         )}
+        {secondaryAction && (
+          <Button
+            onClick={secondaryAction.onClick}
+            disabled={secondaryAction.disabled}
+            sx={{ backgroundColor: "background.default" }}
```
ianjon3s: Secondary buttons should be marked up `variant="contained" color="secondary"`

joecarver:
> * Search box should be focussed when I load the modal so I can start typing right away

Adding this option raises a a11y warning - "The autoFocus prop should not be used, as it can reduce usability and accessibility for users. (eslint jsx-a11y/no-autofocus)" - https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/8f75961d965e47afb88854d324bd32fafde7acfe/docs/rules/no-autofocus.md 

Agree might be a nice UX for some users but maybe best to leave it? @ianjon3s any thoughts?

joecarver:
> * Badges were changed to be circular due to feedback on the explore templates PR, the layout has also been changed slightly so that the action buttons sit below the badges when viewing template details, let's keep that consistent

https://github.com/theopensystemslab/planx-new/pull/7296/commits/6fe2ef44b9b55e3c9b5a07b56994e242ec9b0dcb

> The search modal needs a max width so it's usable on larger screens, I think contentWide would fit well

https://github.com/theopensystemslab/planx-new/pull/7296/commits/29fa4ab435216f8230d44fbaa9fbd792b1c4623a

apps/editor.planx.uk/src/pages/Explore/components/SearchListItemDetailActions.tsx
joecarver: https://github.com/theopensystemslab/planx-new/pull/7296/commits/6b3b3a5addb71e7115bd838a31d35979ba03a337

apps/editor.planx.uk/src/pages/Explore/components/useSearchFlows.ts
joecarver: 😆 eventually I'll remember this - thanks! 

https://github.com/theopensystemslab/planx-new/pull/7296/commits/38be6ab1e44bd9f25b1a1339bdcd3e02a34f3ca1

apps/editor.planx.uk/src/pages/Explore/components/SearchModal.tsx
joecarver: https://github.com/theopensystemslab/planx-new/pull/7296/commits/4824a8926bbc0d7b991d43017f0e13b650006f36

joecarver:
> * Maybe we should include `team.name` and / or `team.slug` in the search vector. I can imagine somebody trying to match on this (e.g. after watching a certain team demo a service at S&T).

Can incorporate this with other upcoming search backend changes - to be discussed in dev call tomorrow

joecarver:
> * Search box should be focussed when I load the modal so I can start typing right away
> * No handling of "no results found" state
> * I think we could make the search input more prominent on the Explore page, it feels quite invisible currently
> * Badges are not legible when in both focus and active states
> * Whitespace - quite a lot in the top right which is unused. Notion has the search bar going full width and then uses two columns below

These have also all been fixed except the autoFocus thing - comment here https://github.com/theopensystemslab/planx-new/pull/7296#issuecomment-5494219993

apps/editor.planx.uk/src/pages/Explore/components/useCopyFlowToTeam.ts
joecarver: working on this in a separate branch as suggested

apps/editor.planx.uk/src/pages/Explore/components/useSearchFlows.ts
DafyddLlyr: It's an annoying one sorry...! I think we could eventually largely solve via extracting GQL types to a package and setting up codegen, but then a linting rule for snake_case would still flag if we're connecting to an external API (e.g. OS, Stripe).

DafyddLlyr (approved):
Thanks for all the follow ups!

PR #7297, opened 2026-08-31
fix(a11y): focus on main-content on load [DAC_SR_Feedback_07]

joecarver:
You can verify this by opening browser console and running:
`document.addEventListener('focusin', e => console.log('focus →', e.target))`,

then stepping through the flow, on each new page you should see: 
`focus → <main id=​"main-content" tabindex=​"-1" class=​"css-ddl2xc">​…​</main>​`

### DIFF
--- a/e2e/tests/ui-driven/src/a11y-flows/focus-navigation.spec.ts
+++ b/e2e/tests/ui-driven/src/a11y-flows/focus-navigation.spec.ts
@@ -42,30 +42,22 @@ test.describe("Focus navigation accessibility", () => {
     await tearDownTestContext();
   });
 
-  test("skip link works consistently across multiple form steps", async ({
+  test("focus moves to main content consistently across multiple form steps", async ({
     page,
   }) => {
+    const mainContent = page.locator("#main-content");
+
     await fillInEmail({ page, context });
     await clickContinue({ page, waitForResponse: true });
 
     await answerQuestion({ page, title: "Question 1", answer: "A" });
     await clickContinue({ page, waitForLogEvent: true });
 
-    await page.keyboard.press("Tab");
-
-    const skipLink = page.getByRole("link", { name: "Skip to main content" });
-    await expect(skipLink).toBeFocused();
-
-    await skipLink.click();
+    await expect(mainContent).toBeFocused();
 
     await answerQuestion({ page, title: "Question 2", answer: "One" });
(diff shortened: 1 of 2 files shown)


e2e/tests/ui-driven/src/a11y-flows/focus-navigation.spec.ts
```diff
@@ -42,30 +42,22 @@ test.describe("Focus navigation accessibility", () => {
...
 
     await answerQuestion({ page, title: "Question 1", answer: "A" });
     await clickContinue({ page, waitForLogEvent: true });
 
-    await page.keyboard.press("Tab");
-
-    const skipLink = page.getByRole("link", { name: "Skip to main content" });
-    await expect(skipLink).toBeFocused();
-
-    await skipLink.click();
```
DafyddLlyr, line 59: I think this is the correct behaviour we want to preserve - we should expect the first `Tab` to hit the skip link.

Examples - 
 - https://www.planning.data.gov.uk/
 - https://www.gov.uk/
 - https://www.theguardian.com/
 - https://www.diy.com/

DafyddLlyr (commented):
I'm not sure this is quite right - we seem to have lost the Skip Link behaviour to acheive this one.

The report is pretty sparse on this point - I'm not able to recreate the issue as shown. On initial load I can never get focus on the footer initially as described?

e2e/tests/ui-driven/src/a11y-flows/focus-navigation.spec.ts
joecarver: ooh yes makes sense, yeah I've tested this on a bunch of components and can't replicate it either. Shall we close this PR and leave it / mark it as cannot replicate?

DafyddLlyr: Yeah, maybe worth a cross-check with @ianjon3s , but as it's a usability issue and we can't replicate I think closing it as "cannot replicate" makes sense to me!

ianjon3s: Agree, let's leave as cannot replicate, as it seems like a niche occurrence rather than an replicable behaviour (which would have surely seen it flagged as an A or AA issue)

joecarver:
Closing as cannot replicate

PR #7298, opened 2026-08-31
fix(a11y): correct header and close button on Planning Constriant modal [DAC_SR_Feedback_08]

joecarver:
The report also mentioned "The escape button closed the modal; this was an issue as it meant if I had not finished my task, I was required to re-open the modal. Usually, I would only expect the modal to close either after I have made a selection or I close it manually." - I didn't fix this as it's my understanding that Escape should be expected to close modals (https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/) - and perhaps the reviewers issue was that they actually didn't know it was a modal therefore the escape behaviour was unexpected? Keen to hear anyone else's take on this though...

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/PlanningConstraints/Public/Modal.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/PlanningConstraints/Public/Modal.tsx
@@ -161,7 +161,7 @@ export const OverrideEntitiesModal = ({
           aria-label="Close"
           onClick={closeModal}
           data-testid="override-modal-close-button"
-          sx={{ color: "grey.600", marginTop: 2, marginRight: 2 }}
+          sx={{ color: "grey.600", marginTop: 1, marginRight: 2 }}
         >
           <Close />
         </IconButton>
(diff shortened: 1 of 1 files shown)


apps/editor.planx.uk/src/@planx/components/PlanningConstraints/Public/Modal.tsx
```diff
@@ -139,9 +141,31 @@ export const OverrideEntitiesModal = ({
...
+          id="dialog-heading"
+          sx={{ flex: 1 }}
+        >
+          I don't think this constraint applies to this property
+        </DialogTitle>
+        <IconButton
+          aria-label="Close"
+          onClick={closeModal}
+          data-testid="override-modal-close-button"
+          sx={{ color: "grey.600", marginTop: 2, marginRight: 2 }}
```
ianjon3s: Vertical alignment is a bit off

Image

ianjon3s (approved):
Agree on maintaining `esc` to close the modal, and adding a close button seems like a strong belt-and-braces approach. One small comment about button alignment

PR #7300, opened 2026-08-31
fix(a11y): upload error messaging and focus fixing [DAC_LV_Feedback_02]

joecarver:
- Properly focus on elements during upload process
- More specific error messages listing specific file types needed


ianjon3s (approved):
Works really well, focus passed to descriptive error.

I'm going to suggest and make one copy change before merging, changing:

`Upload and label: [missing info]`

to 

`Upload and label files containing: [missing info]`

PR #7301, opened 2026-08-31
fix(editor): Fixed height for patterns popover to solve async issues

DafyddLlyr:
## What's the problem?
The "Patterns" tab of the `<AddComponentModal/>` does not re-orientate itself on open, it's always top down. Opening the popover on a hanger close to the bottom of the screen makes in un-navigable.

Thread here with video (OSL Slack) - https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1787734171912219

## What's the cause?
The contents of the patterns tab are loaded asynchronously and then it never re-adjusts it's position.

## What's the solution?
I've just set a fixed `height` here instead of a dynamic height + max. This means it's always 480px, which is really only an issue until we have a few additional patterns (where we would have hit this point anyway). I think even if it did re-adjust on load, it would feel quite janky, and only be a short-lived solution.

Another option (previously discussed) would be pre-warming the cache with the required patterns but this feels like overkill..!

image


augustlindemer (approved):
Nice simple solution! Should work indefinitely, if we have many more patterns we'd just get a vertical scroll as we do in Components.

ianjon3s (approved):
Works perfectly

PR #7302, opened 2026-08-31
fix(editor): Resolve Tanstack Router warning for `CatchAllComponent`

DafyddLlyr:
Resolves the following warning shown when running the Editor - 

```sh
[tanstack-router] These exports from "/Users/dafyddllyr/Code/planx-new/apps/editor.planx.uk/src/routes/$.tsx" will not be code-split and will increase your bundle size:
- CatchAllComponent
For the best optimization, these items should either have their export statements removed, or be imported from another location that is not a route file.
```

image

The solution is not to export a component from a Route file elsewhere, but just to create the component as standalone and import it as a pure component, not from within a route file.


joecarver (approved):
Nice and clean and removes the warning, thanks for doing it!

PR #7306, opened 2026-09-01
feat(stripe): add DB columns and permissions for Stripe account IDs

joecarver:
- add column `team_integrations.production_stripe_account_id`
- add column `team_integrations.staging_stripe_account_id`
- Allows `api` hasura role to update these - actual user permissions will need to be checked at the API layer
- This field will be written as part of the OAuth callback handler, after successful Stripe account connect

### DIFF
--- a/apps/hasura.planx.uk/metadata/databases/default/tables/public_team_integrations.yaml
+++ b/apps/hasura.planx.uk/metadata/databases/default/tables/public_team_integrations.yaml
@@ -17,11 +17,13 @@ insert_permissions:
         - production_file_api_key
         - production_govpay_secret
         - production_power_automate_api_key
+        - production_stripe_account_id
         - staging_bops_secret
         - staging_bops_submission_url
         - staging_file_api_key
         - staging_govpay_secret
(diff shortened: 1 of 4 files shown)


DafyddLlyr (approved):
Great...!

I'm reminded each time we revisit this that the `team_integrations` setup and secrets management is quite awkward here. I'd be keen to re-open https://github.com/theopensystemslab/planx-new/pull/6492/ sometime to come up with a more solid method of managing our keys.

It might be the case that we can use these Stripe secrets as a test case, but this is possibly overley complicating time sensitive work.

joecarver:
> I'm reminded each time we revisit this that the `team_integrations` setup and secrets management is quite awkward here.

Correct me if I'm wrong but I don't necessarily think that applies here, as we are not storing API keys for councils in this table, just their Stripe Account ID. Perhaps we want to consider those as secrets as well though? From I think these are fine to be considered public as they are useless without API keys/auth etc

DafyddLlyr:
Ah great - much simpler if we can store in plaintext 👍

PR #7307, opened 2026-09-01
refactor: CloseButton as reusable component

ianjon3s:
## What does this PR do?

`CloseButton` has recently been created as a reusable `IconButton` with styling only, meaning that when using it with a properly marked up `IconButton` as a child, two `<button>` elements are nested together:

image

As there are numerous uses of the close button across both the editor and public interfaces, it's worth going the whole way to making this a reusable component, meaning that:
- Button is marked up with correct `title` and `aria-label`
- Size/padding can be standardised with an optional `small` size
- Button inherits colour of parent

### Components updated

**Editor:**
- Component modal (`FormModal`)
- `NoteEditorDialog` 
- `EnvironmentSelect`
- `TeamSelect`
- `NotificationsPanel`
- `FeatureFlagsPanel`

**Public:**
- `MoreInfo` (help text)
- Feedback panel

### Also fixes

- Content page (public) to have correct `back` button rather than close icon (which is currently incorrectly placed in the header
 
**Current:**

image

**Fixed:**

image

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/shared/Preview/MoreInfo.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/shared/Preview/MoreInfo.tsx
@@ -1,31 +1,26 @@
-import CloseIcon from "@mui/icons-material/Close";
 import Box from "@mui/material/Box";
 import Container from "@mui/material/Container";
 import Drawer from "@mui/material/Drawer";
-import IconButton from "@mui/material/IconButton";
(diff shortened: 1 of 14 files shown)


zz-hh-aa (approved):
Thanks for picking this up! Tested in all the listed components, looks and works as expected ✨

ianjon3s:
Thanks for testing @zz-hh-aa !

PR #7308, opened 2026-09-01
chore: Start GCP subdomain process

DafyddLlyr:
Certificates layer deployed and GCP emailed ✅


joecarver (approved):
all good (despite my initial confusion at the PR name wondering when we started using GCP over AWS 😅 )

PR #7317, opened 2026-09-01
feat: explain sanitised addresses in submissions log

zz-hh-aa:
Alters db views `submissions_grouped` and `submission_services_log` to show 'No longer retained' when applications >6mo old. (As Daf suggested [in comment](https://github.com/theopensystemslab/planx-new/pull/7291#discussion_r3881680525))

This gives us 3 options for displaying addresses:
1. Address
2. Empty (eg for services that don't have addresses, such as "Register to attend Public Inquiry")
3. "No longer retained" for older applications

Copy suggestions welcome--this felt like a middle ground between succinct and technical "Sanitised" and verbose "Submission is older than 6months".

### DIFF
--- a/apps/hasura.planx.uk/migrations/default/1788270607778_explain_sanitised_address_in_submissions_grouped/up.sql
+++ b/apps/hasura.planx.uk/migrations/default/1788270607778_explain_sanitised_address_in_submissions_grouped/up.sql
@@ -1,0 +1,174 @@
+CREATE
+OR REPLACE VIEW "public"."submissions_grouped" AS WITH payments AS (
+  SELECT
+    ps.session_id,
+    'Pay' :: text AS event_type,
+    initcap(ps.status) AS status,
+    ps.created_at
+  FROM
+    (
+      payment_status ps
+      LEFT JOIN payment_status_enum pse ON ((pse.value = ps.status))
+    )
+  WHERE
+    (
+      (ps.status <> 'created' :: text)
+      AND (
+        ps.created_at >= '2024-01-01 00:00:00+00' :: timestamp with time zone
+      )
+    )
+),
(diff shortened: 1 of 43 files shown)


apps/hasura.planx.uk/migrations/default/1788270607778_explain_sanitised_address_in_submissions_grouped/up.sql
```diff
@@ -0,0 +1,173 @@
...
+          ((ls.data -> 'passport' :: text) -> 'data' :: text) -> '_address' :: text
+        ) ->> 'single_line_address' :: text
+      ),
+      (
+        (
+          ((ls.data -> 'passport' :: text) -> 'data' :: text) -> '_address' :: text
+        ) ->> 'title' :: text
+      )
+    )
+  END AS address
```
jessicamcinchak, line 162: For services that do _not_ have any FindProperty, could we simply apply a _third_ COALESCE condition here of `"Not applicable"` so that db view is alway correct at source? 

Thanks for remembering this scenario too in PR description !!

jessicamcinchak (commented):
I think you have one too many sets of migrations here, can one be removed from branch and pizza rebuilt (from destroy) to confirm new Hasura comes up correct? 

Also one minor comment! Otherwise great, thanks for finding nice clear copy compromise here!

zz-hh-aa:
Thanks @jessicamcinchak, both pairs of migrations are actually intentional! One for `submissions_grouped` view and another for `submission_services_log`. Renamed the commits when I edited the `COALESCE` so hopefully that's clearer now. 

On re-read your comment is very clear, I just totally misunderstood, thanks for clarifying!

PR #7319, opened 2026-09-01
refactor: Move explore nav menu item to first position

ianjon3s:
## What does this PR do?

- Repositions `Explore` to be in first place in the navigation menu
- Renames to `Explore & search`
- Updates icon to be indicative of search


apps/editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
```diff
@@ -364,6 +350,13 @@ function EditorNavMenu() {
     return accessibleByCurrentUserRole;
   };
 
+  const exploreRoute: Route = {
+    title: "Explore & search",
+    Icon: FindInPageIcon,
+    route: `/app/${teamSlug}/explore`,
+    accessibleBy: ["platformAdmin", "teamAdmin", "teamEditor"] satisfies Role[],
```
jessicamcinchak: Just double-checking we're intentionally omitting analysts here?

jessicamcinchak (approved):
Looks good, much prefer this ordering !

apps/editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
ianjon3s: Good catch, I've added analysts to the list for explore, so that they can view it on any team

PR #7321, opened 2026-09-02
fix(ci): Add new `flow.search_vector` to data sync script

DafyddLlyr:
~~To be merged once https://github.com/theopensystemslab/planx-new/pull/7322/changes/738a9bc579e861ecfe9a2a3f5b5f4fb07b2c7775 reaches production.~~

Not going to re-run this manually once merged due to notes testing on staging, just let it complete overnight.

`pnpm test-sync` running locally on this branch ✅

### DIFF
--- a/scripts/seed-database/write/flows.sql
+++ b/scripts/seed-database/write/flows.sql
@@ -25,7 +25,8 @@ CREATE TEMPORARY TABLE sync_flows (
   deleted_at timestamptz,
   email_template text,
   is_service boolean,
-  is_pattern boolean
+  is_pattern boolean,
+  search_vector tsvector
   );
 \copy sync_flows FROM '/tmp/flows.csv' WITH (FORMAT csv, DELIMITER ';');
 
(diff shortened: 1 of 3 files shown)


scripts/seed-database/write/flows.sql
```diff
@@ -25,7 +25,8 @@ CREATE TEMPORARY TABLE sync_flows (
   deleted_at timestamptz,
   email_template text,
   is_service boolean,
-  is_pattern boolean
+  is_pattern boolean,
+  search_vector tsvector
```
DafyddLlyr, line 29: This only needs to be added here, we can't directly add data into it - it's self-populating.

Adding `search_vector` elsewhere in the script leads to the following error - 

```sh
psql:write/flows.sql:110: ERROR:  cannot insert a non-DEFAULT value into column "search_vector"
DETAIL:  Column "search_vector" is a generated column.
```

jessicamcinchak (approved):
Appreciate fixes (& timing coordination w/ notes migration) here!

PR #7323, opened 2026-09-02
feat: explain sanitised addresses in submissions log v2

zz-hh-aa:
Re-opens #7317 after stack rebase mishap. 

Alters db views submissions_grouped and submission_services_log to show 'No longer retained' when applications >6mo old, with additional `COALESCE` option for `Not applicable` if no address. (Per Daf [original suggestion](https://github.com/theopensystemslab/planx-new/pull/7291#discussion_r3881680525) and [Jess's feedback](https://github.com/theopensystemslab/planx-new/pull/7317#discussion_r3905472180))

NB: there are _two_ sets of migrations here for two different views, which might look like duplicate set of migrations (one for `submissions_grouped` and one for `submission_services_log`). 

This gives us 3 options for displaying addresses:
- Address
- "No longer retained" for older applications
- "Not applicable" when no address

Copy suggestions welcome--this felt like a middle ground between succinct and technical "Sanitised" and verbose "Submission is older than 6months".


freemvmt (approved):
Makes sense! Have to say I didn't parse all the SQL!

PR #7325, opened 2026-09-02
chore(api): Log cause of payment parsing error to Airbrake

DafyddLlyr:
Aiming to get more information on this category of error - https://planx.airbrake.io/projects/329753/groups/4355606157377770771?tab=notice-detail

Logging the `cause` directly to Airbrake should allow us to pick up the Zod error originating here - 

https://github.com/theopensystemslab/planx-core/blob/2a081237d30d242087ba95b74140626f9f08ee11/src/utils/feeBreakdown.ts#L184-L188


zz-hh-aa (approved):
Thanks!

PR #7326, opened 2026-09-02
feat(Stripe): Onboarding (api + frontend)

joecarver:
### What's in this PR
- api routes/handlers/tests for Stripe redirects and connect status
- frontend changes to call these endpoints
Was planning to put the frontend changes in a separate branch but they're quite minimal, and they make it easier to fully test the API changes.
- adds `stripe` sdk package to give us a bit of typing and wrappers for authorizeUrl and oAuth token exchange

### How does it work?
- e.g. for buckinghamshire user on local 
- Council user goes to e.g. `http://localhost:3000/app/buckinghamshire/settings/payments`
- Frontend queries `http://localhost:7002/stripe/connect/buckinghamshire/status` to check whether `team_integrations.staging_stripe_account_id` has been set for Bucks
- If so, display a 'connected' badge, the account ID, and a 'test' badge
- If not, display a 'Connect Stripe Account' button
- Clicking Connect Stripe Account makes a request to the API `http://localhost:7002/stripe/connect/buckinghamshire` - which generates the oauth url and returns a 302 redirect response to redirect the browser to the URL
- User then connects a new or existing Stripe account via Stripe's onboarding UI
- When complete, Stripe calls `http://localhost:7002/stripe/connect/callback` (this is configured [here](https://dashboard.stripe.com/acct_1UAntHEAzmRwH26v/test/settings/connect/onboarding-options/oauth)
- Our API checks the response is valid, uses the returned `code` to get a `token` from which we can extract the `account_id`
- Account ID is stored in our DB `team_integrations` table for staging or prod as relevant
- API returns a 302 redirect response back to the payments setting page, with an error attached
- Payment settings page now shows success state
- Any errors in the pre-oauth or post-auth endpoint trigger a 302 redirect back to the payments setting page, with an error message attached

### What's not here
- I didn't add any detailed information to the connect status yet - we may want detailed info for the 'started by didn't finish' state, although i'm not sure this is actually relevant given we are using standard oauth connect. We may also want more detailed information for a connected account, which we would query separately from Stripe using the account ID. But would probably be good to scope this out a bit more first


apps/api.planx.uk/modules/auth/requireTeamMembership.ts
```diff
@@ -1,8 +1,48 @@
...
+  );
+
+  if (!user.isPlatformAdmin && !isUserInTeam) {
+    throw new ServerError({
+      status: 403,
+      message: `Access denied. User ${userId} is not a member of team ${teamLabel} with permission to edit it`,
+    });
+  }
+};
+
```
joecarver, line 45: Refactored this shared functionality out of `requireTeamMembership` so it can be re-used by the Stripe check, which needs to fetch the teamId using the slug, rather than from the request.

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Payment/Onboarding/index.tsx
```diff
@@ -5,54 +5,54 @@ import CircularProgress from "@mui/material/CircularProgress";
...
-  businessName: string;
-  email: string;
-  chargesEnabled: boolean;
-  payoutsEnabled: boolean;
-}
+export const Onboarding: React.FC = () => {
+  const teamSlug = useStore((state) => state.teamSlug);
+  const toast = useToast();
+  const navigate = useNavigate();
+  const search = useSearch({ strict: false });
```
DafyddLlyr: I don't think we should use `{ strict: false }` here.

This is always at known static route and we've already configured the `paymentsSearchSchema` here to account for this.

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Payment/Onboarding/index.tsx
```diff
@@ -5,54 +5,54 @@ import CircularProgress from "@mui/material/CircularProgress";
...
+  const { stripeConnected, stripeError } = search;
 
-const PLACEHOLDER_ACCOUNT: ConnectedAccount = {
-  accountId: "acct_1234567890",
-  businessName: "Example Council",
-  email: "payments@example.council.gov.uk",
-  chargesEnabled: true,
-  payoutsEnabled: true,
-};
+  const stripeConnectStatusQuery = useStripeConnectStatus(teamSlug);
```
DafyddLlyr: nit: I'd be tempted to desctucture this for brevity below (no repeated `stripeConnectStatusQuery.`)

apps/editor.planx.uk/src/routes/_authenticated/app/$team/settings/payments.tsx
```diff
@@ -1,6 +1,15 @@
 import { createFileRoute, redirect } from "@tanstack/react-router";
+import { zodValidator } from "@tanstack/zod-adapter";
 import { hasFeatureFlag } from "lib/featureFlags";
 import PaymentSettings from "pages/FlowEditor/components/Settings/Team/Payment";
+import { z } from "zod";
+
+export const paymentsSearchSchema = z.object({
+  stripeConnected: z.boolean().optional(),
+  stripeError: z.string().optional(),
```
DafyddLlyr: We could use the `string & {}` trick here to map the known and used error states that we're handling (unless there's an easy `const` import we can rely on). This would both explicitly show us the handled errors here, as well as give us _some_ type safety downstream.

DafyddLlyr: https://www.youtube.com/watch?v=lraHlXpuhKs&t=590s

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Payment/Onboarding/index.tsx
```diff
@@ -5,54 +5,54 @@ import CircularProgress from "@mui/material/CircularProgress";
...
+    if (!stripeConnected && !stripeError) return;
+
+    if (stripeConnected) {
+      toast.success("Stripe account connected successfully");
+      stripeConnectStatusQuery.refetch();
+    } else if (stripeError) {
+      toast.error(
+        stripeError === "access_denied"
+          ? "Stripe connection was cancelled"
+          : "Failed to connect Stripe account, please try again",
```
DafyddLlyr: The suggestion above for using a `const` of values for `stripeError` would also allow us to create a map of status: message. This is a nice pattern as the number of statues handled tends to grow over time, and then mapping messages back in ends up a little messy.

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Payment/Onboarding/index.tsx
```diff
@@ -5,54 +5,54 @@ import CircularProgress from "@mui/material/CircularProgress";
...
- * @todo Replace with real Stripe Connect onboarding
- */
-const connectToStripe = async (): Promise<ConnectedAccount> => {
-  console.log("[Onboarding] Starting Stripe Connect onboarding...");
-  console.log("[Onboarding] Redirecting to Stripe Connect OAuth...");
-  await new Promise((resolve) => setTimeout(resolve, 2000));
-  console.log("[Onboarding] Stripe Connect account linked successfully");
-  return PLACEHOLDER_ACCOUNT;
-};
+  useEffect(() => {
```
DafyddLlyr, line 30: An alternative here could be to handle this on the route via the `loader`, then passing a `stripeResult` to the component for the toast to handle.

I don't think either is better or worse, just an observation on a possibly more Tanstack idiomatic solution 👍

apps/api.planx.uk/modules/stripe/service.ts
```diff
@@ -0,0 +1,149 @@
...
+      query GetStripeAccountId($teamId: Int!) {
+        team_integrations(where: { team_id: { _eq: $teamId } }) {
+          accountId: ${column}
+        }
+      }
+    `,
+    { teamId },
+  );
+
+  return team_integrations[0]?.accountId ?? null;
```
DafyddLlyr: Guess the nit? 😉 🐍

apps/api.planx.uk/modules/stripe/controller.ts
```diff
@@ -0,0 +1,95 @@
...
+  InitiateConnectController,
+} from "./types.js";
+
+export const initiateConnect: InitiateConnectController = async (
+  req,
+  res,
+  next,
+) => {
+  try {
+    const team = res.locals.team as Team;
```
DafyddLlyr: You should be able to set the generic `TLocals` in `ValidatedRequestHandler` to avoid this casting throughout.

apps/api.planx.uk/modules/stripe/middleware.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Look up the team and confirm the requesting user is a teamEditor/teamAdmin for it (or a platformAdmin)
+ */
+export const requireStripeConnectTeamAuth: RequestHandler = async (
+  _,
+  res,
+  next,
+) => {
+  const { teamSlug } = res.locals.parsedReq.params as { teamSlug: string };
```
DafyddLlyr: I think this casting is also not necessary - we can type this function not as a plain `RequestHandler` but a `ValidatedRequestHandler<typeof connectSchema, never, { team: Team }>`

apps/api.planx.uk/modules/stripe/middleware.ts
```diff
@@ -0,0 +1,75 @@
...
+
+// Read the connect state from the session and clear it so it can't be reused
+export const consumeConnectState = (
+  req: RequestWithSession,
+): StripeConnectSessionState | undefined => {
+  const state = req.session?.stripeConnect as
+    StripeConnectSessionState | undefined;
+  req.session!.stripeConnect = undefined;
+  return state;
+};
```
DafyddLlyr, line 61: Rather than cast here and just having a "trust" boundary, the safer option would be to parse the response (e.g. using Zod).

If we make a Zod schema the source of `StripeConnectSessionState`, we can then parse this back out of the sessions store.

```ts
const stripeConnectSessionStateSchema = z.object({
  teamId: z.number(), 
  teamSlug: z.string(), 
  nonce: z.string(),
});

type StripeConnectSessionState = z.infer<typeof stripeConnectSessionStateSchema>;

...

// Inside consumeConnectState()
return stripeConnectSessionStateSchema.safeParse(raw).data;
```

apps/api.planx.uk/modules/stripe/middleware.ts
```diff
@@ -0,0 +1,75 @@
...
+  teamSlug: string;
+  nonce: string;
+}
+
+/**
+ * Only the `session` property is needed for these middleware functions
+ */
+interface RequestWithSession {
+  session?: Record<string, unknown> | null;
+}
```
DafyddLlyr, line 18: This is set to `undefined` below, is the `null` case actually covered?

apps/api.planx.uk/modules/stripe/service.ts
```diff
@@ -0,0 +1,149 @@
...
+      }
+    `,
+    { teamId, accountId },
+  );
+};
+
+export const getStripeAccountId = async (
+  teamId: number,
+): Promise<string | null> => {
+  const column = stripeAccountIdColumn();
```
DafyddLlyr, line 133: Nice simple solution 👌

apps/api.planx.uk/modules/stripe/service.ts
```diff
@@ -0,0 +1,149 @@
...
+};
+
+// Staging uses Stripe test mode keys, production uses live mode keys
+export const getStripeMode = (): "test" | "live" =>
+  process.env.APP_ENVIRONMENT === "production" ? "live" : "test";
+
+/**
+ * `team_integrations` stores separate columns per environment
+ */
+const stripeAccountIdColumn = (): string =>
```
DafyddLlyr: nit: You can drop the `: string` here and have TS infer a tighter type.

Image

apps/api.planx.uk/modules/stripe/service.test.ts
```diff
@@ -0,0 +1,190 @@
...
+    process.env.API_URL_EXT = "https://api.example.com";
+    mockAuthorizeUrl.mockReturnValue(
+      "https://connect.stripe.com/oauth/authorize?mock=1",
+    );
+  });
+
+  afterEach(() => {
+    process.env = { ...originalEnv };
+    mockAuthorizeUrl.mockReset();
+  });
```
DafyddLlyr: Using `vi.stubEnv()` here is a bit more vitest-idiomatic (admittedly not applied consistently across the API codebase currently!). This should save a bit of boilerplate here.

DafyddLlyr (approved):
Great..! Solid piece of work as always, conforms well to the codebase's existing patterns which is great to see. Easy to read and understand.

Almost all comments below are minor and just general TS comments I think.

It would be great to increase test coverage here as this area of the codebase is going to be quite mission-critical in the next few months. This could look like more frontend and API tests, or possibly E2E (would be nice for our debugging, but I feel like having a whole mock "connect" account would be a lot of work for little reward). I'd be inclined to focus on more unit tests I think.

Image

apps/api.planx.uk/modules/stripe/service.ts
```diff
@@ -0,0 +1,149 @@
...
+    ? "production_stripe_account_id"
+    : "staging_stripe_account_id";
+
+export const saveStripeAccountId = async (
+  teamId: number,
+  accountId: string,
+): Promise<void> => {
+  const column = stripeAccountIdColumn();
+
+  await $api.client.request(
```
DafyddLlyr, line 115: We _might_ want to think about permissions on this table now that writing is a genuine allowed action, but I think sticking with `$api` and not modifying permissions is the right move for now. It could allow us to simplify the auth middleware chain here by relying on Hasura permissions, but I don't think we need to widen this just yet (and we can't scope permissions per-column, just per-row).

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Payment/Onboarding/index.tsx
joecarver: https://github.com/theopensystemslab/planx-new/pull/7326/commits/679eaaea530b76b5775baf4e6824b0d3df2df908

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Payment/Onboarding/index.tsx
joecarver: https://github.com/theopensystemslab/planx-new/pull/7326/commits/4c60809f1b31fed2f67192a6dbbfbe5cee8be51d

apps/editor.planx.uk/src/routes/_authenticated/app/$team/settings/payments.tsx
joecarver: https://github.com/theopensystemslab/planx-new/pull/7326/commits/e217741161c2645b0de05d6be4d6107b984561a7

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Payment/Onboarding/index.tsx
joecarver: good shout! always a good idea to pull this kind of stuff up to the route matching level https://github.com/theopensystemslab/planx-new/pull/7326/commits/a8e61658345d22c676ba2353607b5d7d76d574ac

apps/api.planx.uk/modules/stripe/service.ts
joecarver: https://github.com/theopensystemslab/planx-new/pull/7326/commits/e35dbf607e664d894ab9199edcaf484c5ed083b5

apps/api.planx.uk/modules/stripe/controller.ts
joecarver: https://github.com/theopensystemslab/planx-new/pull/7326/commits/99ada97c107ac6d870043aeadd63b025a95cff02

apps/api.planx.uk/modules/stripe/middleware.ts
joecarver: https://github.com/theopensystemslab/planx-new/pull/7326/commits/feb9e1765f14b9da76a07d4ea9663acdacfdadba

apps/api.planx.uk/modules/stripe/middleware.ts
joecarver: https://github.com/theopensystemslab/planx-new/pull/7326/commits/eed44740a61a8d31b49df74f6754b8d5eb82605a

apps/api.planx.uk/modules/stripe/service.ts
joecarver: https://github.com/theopensystemslab/planx-new/pull/7326/commits/0e5652513859f78715df2e0c55a2a0923ee28296

apps/api.planx.uk/modules/stripe/middleware.ts
joecarver: https://github.com/theopensystemslab/planx-new/pull/7326/commits/8905d8b39463b7392df2b89dd4a411be0a736565

apps/api.planx.uk/modules/stripe/service.test.ts
joecarver: https://github.com/theopensystemslab/planx-new/pull/7326/commits/0dce13fec7d226c9c39c9a37847e724cf8dfe814

apps/api.planx.uk/modules/stripe/service.ts
joecarver: Noted - makes sense, leaving for now

joecarver:
> It would be great to increase test coverage here as this area of the codebase is going to be quite mission-critical in the next few months. This could look like more frontend and API tests, or possibly E2E (would be nice for our debugging, but I feel like having a whole mock "connect" account would be a lot of work for little reward). I'd be inclined to focus on more unit tests I think.

@DafyddLlyr Thanks for the thorough review. I addressed all the code style comments, and added some more tests:

controller API tests: https://github.com/theopensystemslab/planx-new/pull/7326/commits/d639ae7f8face0b90c80978486cdf12786aafc2f

onboarding UI tests: https://github.com/theopensystemslab/planx-new/pull/7326/commits/2ec85239602cba4f464b7e1cb12bba53d6faf8db

apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Payment/Onboarding/index.tsx
DafyddLlyr: Actually that worked out better than expected 😅 Feels clearer and simpler now!

DafyddLlyr (approved):
This is great thanks @joecarver - appreciate the follow up to the comments and additional testing here!

PR #7329, opened 2026-09-02
[api] upgrade AI SDK + enforce `disallowPromptTraining`

freemvmt:
As per title. See [Slack thread](https://opensystemslab.slack.com/archives/C0ARCE260LF/p1786457050736379) for motivation.

In passing through we also expand caught errors (so that changes in account-level settings which invalidate hardcoded model choice will be clear to see), simplify the `getModel` util (which doesn't throw), and add tests for full coverage on `api.planx.uk/modules/ai/`.

Vercel migration guide: https://ai-sdk.dev/docs/migration-guides/migration-guide-7-0

### DIFF
--- a/apps/api.planx.uk/modules/ai/projectDescription/service.test.ts
+++ b/apps/api.planx.uk/modules/ai/projectDescription/service.test.ts
@@ -1,0 +1,203 @@
+import { APICallError } from "ai";
+// see: https://ai-sdk.dev/docs/ai-sdk-core/testing
+import { MockLanguageModelV4 } from "ai/test";
+
+import { DEFAULT_MODEL_ID } from "../constants.js";
+import { GATEWAY_STATUS } from "../types.js";
+import { enhanceProjectDescription } from "./service.js";
+
+const mockLogAiGatewayExchange = vi.fn();
+const mockGetModel = vi.fn();
+
+vi.mock("../logs.js", () => ({
+  logAiGatewayExchange: (args: unknown) => mockLogAiGatewayExchange(args),
+}));
+
+vi.mock("../utils.js", () => ({
+  getModel: () => mockGetModel(),
+}));
+
+const mockFlowId = "123e4567-e89b-12d3-a456-426614174000";
+const endpoint = "/ai/project-description/enhance";
+
+/** Captures the options `generateText` passes down to the model. */
+let modelCallOptions: Record<string, unknown> | undefined;
+
+// a model returning a schema-compliant object, plus the gateway metadata the audit log reads
+const stubModel = (
+  overrides: {
+    enhancedDescription?: string;
+    status?: string;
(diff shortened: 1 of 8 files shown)


apps/api.planx.uk/modules/ai/projectDescription/service.test.ts
```diff
@@ -0,0 +1,203 @@
+import { APICallError } from "ai";
+// see: https://ai-sdk.dev/docs/ai-sdk-core/testing
+import { MockLanguageModelV4 } from "ai/test";
```
DafyddLlyr, line 4: Neat!

apps/api.planx.uk/modules/ai/projectDescription/service.ts
```diff
@@ -79,6 +83,7 @@ export const enhanceProjectDescription = async (
       ? { ok: false, error: output.status }
       : { ok: true, value: output.enhancedDescription };
   } catch (error) {
+    // full list of AI SDK errors: https://ai-sdk.dev/docs/reference/ai-sdk-errors
```
DafyddLlyr, line 87: nit(minor): As this `else if` chain grows I think a `switch/case` might read a little better..!

apps/api.planx.uk/modules/ai/projectDescription/service.ts
```diff
@@ -101,3 +115,31 @@ export const enhanceProjectDescription = async (
...
+const getGatewayRejectionName = (error: unknown): string | undefined => {
+  const cause = error instanceof Error ? error.cause : undefined;
+  if (!APICallError.isInstance(cause)) return undefined;
+
+  const body = cause.data as
+    { error?: { param?: { name?: unknown } } } | undefined;
+  const name = body?.error?.param?.name;
+
+  return typeof name === "string" ? name : undefined;
+};
```
DafyddLlyr: Adding `@ai-sdk/gateway` as a direct dependency might also be a way forward here and avoids us having to maintain this function and `isUnroutableRequestError()` above.

DafyddLlyr (approved):
One meaningful comment below - it seems simpler and more maintainable to add `@ai-sdk/gateway` as a direct dependency here instead of just a transitive dep through `ai`.

apps/api.planx.uk/modules/ai/projectDescription/service.ts
freemvmt: yes nice, it's a bit unwieldy, will do that!

apps/api.planx.uk/modules/ai/projectDescription/service.ts
freemvmt: Yes agree it's a complicated workaround when we could just manage one more package. Will implement that :)

PR #7330, opened 2026-09-02
chore: revert `FlowNotes` across editor and e2e

jessicamcinchak:
Reverting editor-side implementation of "FlowNotes", particularly because a few pieces of this (eg context menu items) aren't fully feature-flagged and leaking into testing of latest notes and visible on live envs. 

Please note I have _not_ touched `.scss` file here @ianjon3s - I suspect that once this merges there will be quite a few sticky-related orphaned classnames in there to additionally tidy up. 

I'll pick up DB, Hasura & API-side reverts in a separate follow-up PR.

**To test:** 
- Please confirm editor routing and graph operations all working as expected / as normal !


zz-hh-aa (approved):
Tested notes-related operations (as listed in [testing thread](https://opensystemslab.slack.com/archives/C5Q59R3HB/p1788309537561849)) and general editor functions like creating flows, archiving flows, copying flows, creating from templates, creating from templates with both editable and non-editable notes, etc. All works as expected!

PR #7334, opened 2026-09-03
feat[alt notes]: set visibility rules for templates

jessicamcinchak:
**Current rules:** 
- In a source template, a Note component type should never be "templated" aka made customisable, so hide this section in editor modal
- In a source template, graph note visbility is controlled as usual by "Show notes" 
- In a templated flow, even when "Show notes" is toggled on, standalone and attached notes should always be hidden from the graph view unless:
  - It is an attached note to a templated node
  - It is a standalone or attached note within a templated folder

**Open questions:**
- In a templated flow, when I open the read-only modal for a non-templated node, I can still see its' "Note" input (disabled) - is this okay or would we expect to be hidden / fully invisible?
- [Jonny's user story / testing feedback] In a source template, if I update only the content of a note and want to push out updates, how can I get around "Found no changes to publish" ? 
  - Current workaround for this would be remove + re-add note to trigger _position_ change that will be captured by diff and therefore _is_ publishable
  - Longer term, we can adjust publish diff logic to have special rule for source templates (eg found changes _or_ has operations since last publish?). The blocker here is _just_ the "check for changes diff" (relies on flattened data which now has sanitised note content), the actual function to push out updates to templated flow children reads from live flow data (unflattened, contains all notes)
- What is role of attached notes alongside templated node "instructions" going forward / how does content team anticipate using either? Should instructions have graph visibility same way attached notes do?

I think / hope that all these template open questions are revisitable and don't hold up wider feature release/call for testing among councils ! Publishing questions only impact source template authors (us).

**Testing flows:**
- Source template example https://7334.planx.pizza/app/opensystemslab/notes-in-templates-testing
- Templated flow example https://7334.planx.pizza/app/opensystemslab/notes-in-templates-testing-templated
- :exclamation: Please note that I've run full content migration against this pizza too since staging now reset - see `temp_data_migrations_audit` table on this pizza's Hasura

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/AttachedNote.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/AttachedNote.tsx
@@ -1,15 +1,45 @@
 import Box from "@mui/material/Box";
 import { useStore } from "pages/FlowEditor/lib/store";
+import {
+  nodeIsChildOfTemplatedInternalPortal,
+  nodeIsTemplatedInternalPortal,
+} from "pages/FlowEditor/utils";
 import React from "react";
 
+import { getParentId } from "../lib/utils";
+
 export const AttachedNote: React.FC<{
+  nodeId: string;
   note: string;
   variant?: "option";
-}> = ({ note, variant }) => {
-  const showNotes = useStore((state) => state.showNotes);
+}> = ({ nodeId, note, variant }) => {
+  const [showNotes, isTemplatedFrom, flow, orderedFlow] = useStore((state) => [
+    state.showNotes,
+    state.isTemplatedFrom,
+    state.flow,
+    state.orderedFlow,
+  ]);
 
-  // TODO also return null if templated flow and *not* attached to a templated node?
-  if (!showNotes) return null;
+  // In templated flows, always hide `AttachedNote` in the graph
+  //   unless it is attached to a templated node or within a templated folder
+  const isAttachedToTemplatedNode = flow[nodeId]?.data?.isTemplatedNode;
+  const parent = getParentId(nodeId);
+  const indexedParent = orderedFlow?.find(({ id }) => id === parent);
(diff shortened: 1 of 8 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/AttachedNote.tsx
```diff
@@ -1,15 +1,45 @@
...
+  //   unless it is attached to a templated node or within a templated folder
+  const isAttachedToTemplatedNode = flow[nodeId]?.data?.isTemplatedNode;
+  const parent = getParentId(nodeId);
+  const indexedParent = orderedFlow?.find(({ id }) => id === parent);
+  const parentIsTemplatedInternalPortal = nodeIsTemplatedInternalPortal(
+    flow,
+    indexedParent,
+  );
+  const parentIsChildOfTemplatedInternalPortal =
+    nodeIsChildOfTemplatedInternalPortal(flow, indexedParent);
```
DafyddLlyr, line 33: This logic is reused in a number of places now - worth refactoring into a hook or store function?

DafyddLlyr (approved):
> _In a templated flow, when I open the read-only modal for a non-templated node, I can still see its' "Note" input (disabled) - is this okay or would we expect to be hidden / fully invisible?_

I think "disabled" is a stronger signal that it's not unintentionally missing - it's just something you can't touch (like you can with other nodes). I think "disabled" is more consistent, at the cost of a slight amount of visual noise of seeing something you can never interact with.

> _In a source template, if I update only the content of a note and want to push out updates, how can I get around "Found no changes to publish" ?_

Checking `operations` as you suggested makes sense here for source templates 👍

apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/AttachedNote.tsx
jessicamcinchak: Yes good shout ! Am going to merge this as-is for migration/prod deploy timing now, but on my list to come back and clean up!

PR #7349, opened 2026-09-03
fix(a11y): Announce progressive loading stages to screen readers

ianjon3s:
## Issue

Relates to:
p.38 (V2 report) - DAC_Status_Messages_01
https://trello.com/c/Js7Xo4FE/3753-aa-dacstatusmessages01

`ProgressiveLoading` inserts as new page content with its `aria-live="polite"` status region already populated. Screen readers will usually only announce changes to an existing live region, so this insertion does not get announced, and focus is left on the `Continue` button that triggered it - a user non-sighted has no indication anything happened.

> Screen reader users are not alerted to the fact that content has been added to the page,
and focus stays on the ‘Continue’ control so that it may appear to users that nothing has
happened. Screen reader users would be required to search the page for changes to identify
what has occurred.

## Solution

Move focus into the region on mount instead, so the stages get read out and focus doesn't stay on the `Continue` button that triggered the loading state.

image

### DIFF
--- a/apps/editor.planx.uk/src/ui/shared/ProgressiveLoading.test.tsx
+++ b/apps/editor.planx.uk/src/ui/shared/ProgressiveLoading.test.tsx
@@ -42,6 +42,12 @@ describe("ProgressiveLoading", () => {
     expect(screen.getByRole("status")).not.toHaveAttribute("aria-hidden");
   });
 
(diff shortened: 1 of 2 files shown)


ianjon3s:
Closing in favour of a fresh PR, repeated Vultr timeouts

PR #7352, opened 2026-09-03
chore: Skip pizza build for dependabot runs

DafyddLlyr:
Follow up to https://github.com/theopensystemslab/planx-new/pull/7292

PRs opened by dependabot need additional permissions to access build images for Vultr - newly opened ones are timing out on this step.

I actually don't think we need to build a Pizza each time though here - the CI process + reviewing package changelogs should be enough. I rarely use a Pizza as part of this maintenance process so it seems both simpler and quicker to skip this step.

Another added bonus here is that dependabot PRs are most likely to leave orphaned instances behind on Vultr (maybe another permissions problem?) and it would be nice to not have to routinely clean these up.

Once reviewed and merged, I'll rebase open PRs - not totally sure this will get picked up properly there (I don't know if the existing commit messages will get amended on a rebase), but worth a shot!

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,9 @@
 # NB. this file configures dependabot's approach to general version updates of dependencies
 # fixes for vulnerabilities/CVEs (i.e. security updates) are handled automatically as per repo settings
+#
+# Every entry carries a "[skip pizza]" commit-message prefix - we rarely need these environments for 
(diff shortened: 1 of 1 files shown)


zz-hh-aa (approved):
Thanks for picking up! 🧹

freemvmt (approved):
yes nice!

PR #7357, opened 2026-09-03
fix(a11y): Correctly associate header as label on ModifyUserInput

ianjon3s:
## Issue

Relates to:
p.21 - DAC_Label_in_Name_01
https://trello.com/c/0SRg9qUj/3747-enhanced-text-input-daclabelinname01

Previously addressed with https://github.com/theopensystemslab/planx-new/pull/7244 , but testing in browser still shows a broken reference.

- Header is not correctly associated as a label for text input when modifying project description (final step)

> The label which enables users to identify the purpose of the field ‘Confirm your project
description’ is not contained within the accessible name of the component ‘Project
description’. Voice activation users rely on the visible label to present the programmatically
associated accessible name by which they may reference the component to their associative
technology. However, if this label is not contained within the accessible name of the control,
assistive technologies will not recognise this component and users may struggle to access it.

## Solution

As suggested in the report, follow the [Making labels and legends headings](https://design-system.service.gov.uk/get-started/labels-legends-headings/) guidance from GOV.UK, placing a `<label>` inside the page `<h1>`, to correctly reuse the page title as label.

## Testing

https://7357.planx.pizza/lambeth/project-description-accessibility-testing/preview

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/ModifyUserInput.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/ModifyUserInput.tsx
@@ -1,9 +1,6 @@
 import Box from "@mui/material/Box";
 import Typography from "@mui/material/Typography";
-import {
-  DESCRIPTION_TEXT,
-  ERROR_MESSAGE,
-} from "@planx/components/shared/constants";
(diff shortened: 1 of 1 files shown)


jessicamcinchak (approved):
Thanks for spotting/re-adjusting this one!

PR #7363, opened 2026-09-03
[api] fix expected `ScaniiFindings` tag on scanned objects

freemvmt:
#7279 introduced a bug, which we fix here (we thought the `ScaniiFindings` tag attached to any scanned S3 object would be empty, but actually it defaults to the literal string `"None"`).

In the meantime, anything uploaded to staging in the last week or so (since staging `ENFORCE_SCAN_FROM` value) will 404 on GET.


DafyddLlyr (approved):
Good spot and quick fix..!

PR #7365, opened 2026-09-04
fix(lps): Configure prettier

DafyddLlyr:
Quick config issue I spotted - we aren't pointing to our custom config, just the standard prettier config.

I've ran `pnpm lint:fix` on this branch and there's no changes to pick up 👌


zz-hh-aa (approved):
Nice catch :)

PR #7369, opened 2026-09-04
refactor: split data sanitation cron job into four separate triggers

zz-hh-aa:
Our sanitise applications cron trigger has been flaky ([see card](https://trello.com/c/pg5PmbFV)). We tried [doubling the timeout](https://github.com/theopensystemslab/planx-new/pull/7181/changes) as a first easy measure, which didn't do the trick. `sanitiseApplicationData` was trying to run 10 different across different tables. 

Instead of creating 10 separate cron triggers here, this tries separating them into 4 (raw application data, audit records, Hasura events and feedback). The order is kept the same, so I don't think we should hit any cascade issues?


apps/api.planx.uk/modules/webhooks/service/sanitiseApplicationData/index.ts
```diff
@@ -1,27 +1,68 @@
...
+  if (operationFailed)
+    await postToSlack(results, `${sanitationType} Data Sanitation`);
 
   return { operationFailed, results };
 };
 
+export const sanitiseRawApplicationData = async () => {
+  const results = await createOperationsHelper(
+    sanitiseRawApplicationDataOperations(),
+    "Raw Application",
```
freemvmt, line 37: nit, maybe like this?

```suggestion
    "Raw Application Data",
```

apps/hasura.planx.uk/metadata/cron_triggers.yaml
DafyddLlyr, line 1: The `schedule` arguments here move this from a daily run to a monthly run for 3 of the 4 jobs.

`0 2 15 * *` means 2am on the 15th of each month  - https://crontab.guru/#0_2_15_*_*

Image

apps/hasura.planx.uk/metadata/cron_triggers.yaml
```diff
@@ -25,8 +25,47 @@
     - name: authorization
       value_from_env: HASURA_PLANX_API_KEY
   comment: 'Re-checks GOV.UK Pay for sessions which began a payment but were never submitted, and logs any paid-but-unsubmitted sessions to #planx-notifications-internal'
-- name: sanitise_application_data
-  webhook: '{{HASURA_PLANX_API_URL}}/webhooks/hasura/sanitise-application-data'
+- name: sanitise_audit_records_data
```
DafyddLlyr, line 28: Auth headers missing here, this will 401 on each run I think.

DafyddLlyr (changes requested):
Two bugs / issues spotted below.

One main question here though about the structure - there's quite a lot of duplication now and there's more things to track. If the problem we're trying to fix is a timeout (the whole process taking more than 60s / 120s) can we just try a fire and forget approach?

This would mean - 
 - Cron job triggers, API request made
 - API immediately returns a [HTTP 202 ](https://http.dog/202)
 - Task runs on API until completed
 - Any failures logged to Slack as per usual

The downside here is we lose our success statuses which are logged in Hasura within the response currently. The truth is though that we rarely (never?) rely on these - even if we just persisted success outcomes to logs (calling `console.log()`, checking AWS CloudWatch if we ever need to see the success status) this is all we'd lose here.

The upside would be that we get to keep the current approach, and don't hit any tricky cascade / sequential issues in future (we don't have a true guarantee that process 1 won't take longer than the gap in the cron jobs, and end up with overlaps). Similarly, we don't have a guarantee that any of the 4 individual tasks won't trip the same timeout issue. It's also just a bit more complex to mentally track 4 jobs.

apps/api.planx.uk/modules/webhooks/controller.ts
```diff
@@ -214,22 +219,73 @@ export const deleteSessionController: DeleteSessionController = async (
   }
 };
 
-export const sanitiseApplicationDataController: SanitiseApplicationData =
+export const sanitiseRawApplicationDataController: SanitiseApplicationData =
```
freemvmt, line 222: I wonder if since all 4 of these handlers are basically identical except for the function and the error message, you could make a factory function to generate each? would be very neat!

freemvmt (commented):
Had a couple comments but I saw Daf proposed an overhaul in a separate review which would make them redundant anyway, but will leave them up anyway :)

zz-hh-aa:
#7394 was simpler so we merged that instead

PR #7370, opened 2026-09-04
style: Note tidy up styles

ianjon3s:
## What does this PR do?

Tidies up and adds subtle note styles to component authoring:
- Remove obsolete styles from `floweditor.scss`
  - `.attached-note` no longer used
  - decorative `.note-connector` replaced with actual hanger
  - `.note-card` replaced with styled component
  - complete removal of paper fold visual treatment
- Note input on question options has subtle note background:
image

- Note input on component has subtle note background:
image

- Standalone note input has subtle note background:
image

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
@@ -1,4 +1,3 @@
-import { lighten, useTheme } from "@mui/material/styles";
 import { ComponentType } from "@opensystemslab/planx-core/types";
 import { useFormikWithRef } from "@planx/components/shared/useFormikWithRef";
 import { ModalFooter } from "ui/editor/ModalFooter";
@@ -29,19 +28,16 @@ function NoteComponent(props: Props) {
     },
     props.formikRef,
   );
-  const theme = useTheme();
 
   // Notes should never be "templated", therefore disable `showTemplateConfiguration` in footer here
   return (
     <form onSubmit={formik.handleSubmit} id="modal">
-      <ModalSection
-        sectionBackgroundColor={lighten(theme.palette.background.note, 0.6)}
-      >
+      <ModalSection>
         <ModalSectionContent>
           <InputRow>
             <Input
               name="text"
-              format="large"
+              format="note"
               minRows={4}
               multiline={true}
               placeholder="Note"
--- a/apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
@@ -1,3 +1,4 @@
+import { lighten, useTheme } from "@mui/material/styles";
 import { ComponentType } from "@opensystemslab/planx-core/types";
 import { useFormikWithRef } from "@planx/components/shared/useFormikWithRef";
 import { ModalFooter } from "ui/editor/ModalFooter";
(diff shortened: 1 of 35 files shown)


apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
```diff
@@ -28,11 +29,14 @@ function NoteComponent(props: Props) {
...
   );
+  const theme = useTheme();
 
   // Notes should never be "templated", therefore disable `showTemplateConfiguration` in footer here
   return (
     <form onSubmit={formik.handleSubmit} id="modal">
-      <ModalSection>
+      <ModalSection
+        sectionBackgroundColor={lighten(theme.palette.background.note, 0.6)}
+      >
```
jessicamcinchak: I think I much prefer the `backgroundColor` set on `Input` (like in OptionsEditor) rather than on the whole `ModalSection` - could we do that consistently? Is there a good argument not to that I'm missing like contrast?

apps/editor.planx.uk/src/@planx/components/shared/BaseOptionsEditor/index.tsx
```diff
@@ -110,6 +110,7 @@ export const BaseOptionsEditor: React.FC<Props> = (props) => {
             <Input
               value={props.value.data.notes || ""}
               placeholder="Note"
+              format="note"
```
jessicamcinchak, line 113: 🙌

apps/editor.planx.uk/src/pages/FlowEditor/floweditor.scss
```diff
@@ -12,6 +12,7 @@ $folderBackground: #b5d1f7;
 $folderBorder: #2c4b88;
 $primaryAction: #0010a4;
 $noteBackground: #fffdb0;
+$lockedNoteBackground: #f5f4d2;
```
jessicamcinchak, line 15: 👍

apps/editor.planx.uk/src/pages/FlowEditor/floweditor.scss
```diff
@@ -857,31 +724,19 @@ $fontMonospace: "Source Code Pro", monospace;
...
-      &::before {
-        @include paperFoldFront;
-        .flow-locked & {
-          border-bottom-color: $lockedNodeBackground;
-        }
-      }
-      &::after {
-        @include paperFoldBack;
-      }
-    }
```
jessicamcinchak, line 884: Appreciate you sifting through this one 🧹

jessicamcinchak (commented):
Thank you! One question/suggestion below to keep modal color consistently on input background rather than whole section background, but otherwise all looks great!

apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
ianjon3s: Great shout, much cleaner and extendable, have updated (more context of bigger diffs in my main comment)

apps/editor.planx.uk/src/pages/FlowEditor/floweditor.scss
ianjon3s: Always good to trim stuff off the `.scss` file 😀

ianjon3s:
Thanks for this suggestion @jessicamcinchak, agree it's a simpler solution that mirrors our current use of `format="data"`

The standalone note input had `format="large"` present, which I initially looked at ways of preserving. However digging deeper, it seems that `format="large"` is currently doing an identical job as `format="bold"` in just making input text semi-bold (which isn't actually needed on the note input either way), with the other styles being repetitions of default styles (height, width, background-color).

I've removed the obsolete `format="large"` and converted all current uses to `format="bold"` (which better explains what it's actually doing), and added a couple of missing cases of `format="bold"` when QAing through components.

jessicamcinchak (approved):
Looks great, thanks for consistent input `note` styles here and also cleaning up the `large`/`bold` issue too !

PR #7371, opened 2026-09-04
refactor: submissions log queries and hooks

zz-hh-aa:
Small outstanding TODO to make queries / hooks follow same pattern


DafyddLlyr (approved):
Perfect!

PR #7374, opened 2026-09-04
chore[alt notes]: add editor modal tests for `Note` component

jessicamcinchak:
`Note` component types uniquely should _not_ display templated node configuration inputs to editors when in a source template - this adds test coverage for this scenario as well as basic general flow rendering/validation.

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
+++ b/apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
@@ -31,7 +31,11 @@ function NoteComponent(props: Props) {
 
   // Notes should never be "templated", therefore disable `showTemplateConfiguration` in footer here
   return (
-    <form onSubmit={formik.handleSubmit} id="modal">
+    <form
+      onSubmit={formik.handleSubmit}
+      id="modal"
+      data-testid="noteEditorForm"
+    >
       <ModalSection>
         <ModalSectionContent>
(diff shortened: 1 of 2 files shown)


apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
```diff
@@ -44,6 +48,7 @@ function NoteComponent(props: Props) {
               value={formik.values.text}
               onChange={formik.handleChange}
               disabled={props.disabled}
+              errorMessage={formik.errors.text as string}
```
jessicamcinchak: Had to cast `as string` here else was misreading `text` as a union type - despite being a `string` in model.ts. Any ideas? 
Image

DafyddLlyr:
```suggestion
              errorMessage={getIn(formik.errors, "text")}
```

I think something like this will do it...! We need to `import { getIn } from "formik"` which will handle the variety of types which `formik.errors` can end up as.

DafyddLlyr (approved):

apps/editor.planx.uk/src/@planx/components/Note/Editor.tsx
jessicamcinchak: :bulb: thank you!

PR #7375, opened 2026-09-04
fix(editor): Tidy up redundant React imports across frontend

DafyddLlyr:
I've been picking these up ad-hoc as I've spotted them, but it's an easy find+replace to just drop them all.

No longer needed as of React 19, missed as part of the migration.

image


ianjon3s (approved):
Thanks for working through these!

PR #7380, opened 2026-09-07
feat: Autofocus on search input when opening modal (a11y compliant method)

ianjon3s:
## What does this PR do?

Discussed earlier as part of broader search work: https://github.com/theopensystemslab/planx-new/pull/7296#issuecomment-5494219993

Adding a native `autoFocus` property throws an error, as it moves focus unexpectedly on load. Setting focus programatically using a ref ensures that focus only passes when the input is loaded, meaning focus moves only once the dialog is fully open, as a predictable result of the user opening it.


DafyddLlyr (approved):
Perfect!

PR #7381, opened 2026-09-07
fix: Pattern and template hover states

ianjon3s:
## What does this PR do?

Tidies up a few bits around patterns:
- Hover state on flow table responsive to flow type (rather than generic grey)
- Removes diagonal pattern on flow card header
- Consolidates `FlowTemplateIndicator` and `PatternIndicator` into generic `FlowIndicator`


apps/editor.planx.uk/src/pages/Flows/components/FlowCard/index.tsx
```diff
@@ -74,16 +73,21 @@ const FlowCard: React.FC<Props> = ({ flow, view }) => {
       >
         {isAnyTemplate && (
           <CardBanner>
-            <FlowTemplateIndicator
+            <FlowIndicator
               isSourceTemplate={isSourceTemplate}
               isTemplatedFlow={isTemplatedFlow}
+              isPattern={isPattern}
```
DafyddLlyr, line 79: nit: (This has come up before as a good idea for a wider refactor) We should look at introducing a single type here across the app instead of a series of booleans.

One to take a closer look at some other time I think, just rearing it's head again here 😅

DafyddLlyr (approved):
Great - thank you..!

apps/editor.planx.uk/src/pages/Flows/components/FlowCard/index.tsx
ianjon3s: I'll take a look at this!

PR #7368, opened 2026-09-04
chore: Update dependencies

DafyddLlyr:
Resolves 41/45 open dependabot issues.

Ran `pnpm update -r` and `pnpm dedupe` to update packages based on our stated version ranges (`~`, `^` etc).

Will take a look at the remaining 4 issues next, as well as the two sets of pinned versions (MUI + TSR).

Regression tests passing here - https://github.com/theopensystemslab/planx-new/actions/runs/33866434079 ✅

PR #7382, opened 2026-09-07
chore: Bump Tanstack Router

DafyddLlyr:
Follow up to https://github.com/theopensystemslab/planx-new/pull/7368#discussion_r3933475898


apps/editor.planx.uk/src/components/LoadingOverlay.tsx
DafyddLlyr, line 1: `LoadingOverlay` is a blocking MUI `Backdrop` component (blocking pointer events). The following TSR update changes the timing of the pending state, which makes this flash on certain navigation events (leading to failing E2E tests) - https://github.com/TanStack/router/pull/7805

The `RouteLoadingIndicator` actually already handles this (and did prior to introducing TSR), so this is just a reversion to what we had before now this issue has been solved.

apps/editor.planx.uk/src/routes/_authenticated/route.tsx
DafyddLlyr, line 1: Another change due to https://github.com/TanStack/router/pull/7805

On local and Pizza environments, logging in sends you back to the app with the JWT in the URL as a `?jwt=...` param. We read that param, save it to a cookie, and then tidy the URL to remove the token.

Before, all of this happened inside the auth store and we removed the param using `window.history.replaceState`.

The problem: TanStack Router now watches `history` for changes. So when we called `replaceState` mid-login, the router treated it as a brand new navigation and re-ran the route's auth guard (while we were still fetching the user). This means the user looks not logged in, and we get kicked back to the login page. You can actually navigate to `/app/:team` manually without issue as the cookie is set (you have been authenticated).

The solution: instead of rewriting the URL ourselves manually, we make this a router action. The auth store no longer touches `history` at all, there's no re-run of the auth guard, and TSR handles it all.

We can test this by logging into the Pizza - this works as expected.

jessicamcinchak (approved):
Pizza all working as expected for me here, thanks for notes on this one!

Very exciting to now support direct navigtaion on login! Eg if I try to go here when not logged in https://7382.planx.pizza/app/medway/flows, I get redirected to login screen, but then put at _correct_ route on success ✅ 🙌

PR #7383, opened 2026-09-07
chore: Bump MUI to v9.4.0

DafyddLlyr:
Follow up to https://github.com/theopensystemslab/planx-new/pull/7368#discussion_r3933465929

Relies on https://github.com/theopensystemslab/planx-core/pull/1064

As noted below, I'd rather merge this once the a11y re-test has completed, so I'm leaving this as draft until then. Reviews welcome in the meantime!


apps/editor.planx.uk/src/@planx/components/EnhancedTextInput/Public/ProjectDescription.test.tsx
```diff
@@ -657,7 +657,15 @@ describe("basic layout and behaviour", () => {
     ).toBeVisible();
 
     // Exit sidebar
-    await user.keyboard("{Esc}");
```
DafyddLlyr, line 660: `Esc` still works perfectly in-browser, but jsdom can't capture this.

apps/editor.planx.uk/src/@planx/components/Feedback/Public/tests/Public.accessibility.test.tsx
```diff
@@ -34,16 +34,16 @@ describe("when the Feedback component is rendered", async () => {
 
     const ratingButtons = screen.getAllByRole("button");
 
-    // user tabs through all rating buttons and selects the last one
+    // The rating buttons are an exclusive ToggleButtonGroup, which uses a roving tabindex
```
DafyddLlyr, line 37: Please see top item here - https://github.com/mui/material-ui/releases/tag/v9.3.0

This has never been flagged by an a11y audit, but I'd maybe say we don't merge this until we've passed the re-test?

apps/editor.planx.uk/src/theme.ts
DafyddLlyr, line 1: Drawer now works as expected without these overrides. Due to changes in https://github.com/mui/material-ui/blob/master/CHANGELOG.md#muimaterial901 if we leave this in the `Drawer` flies off to the left when closed 😅

apps/editor.planx.uk/src/@planx/components/Feedback/Public/tests/Public.accessibility.test.tsx
freemvmt: have you had a go at this? does it still feel reasonably intuitive? the fact they've flagged it in that update with the use of the wheelchair emoji* would suggest it's supposed to be an improvement?

*ngl kinda surprised this is still normal practice

freemvmt (approved):

apps/editor.planx.uk/src/@planx/components/Feedback/Public/tests/Public.accessibility.test.tsx
DafyddLlyr: Yeah this seems correct to me, as these are effectively Radio buttons. The `Feedback` component now acts like this (even though it's a `button` and not `input` under the hood still) - 

- https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/radio
- https://mui.com/material-ui/react-toggle-button/

PR #7386, opened 2026-09-08
chore(postgres): Update to Debian 13 (Trixie)

DafyddLlyr:
### What's the problem?
This is a follow up to the issues initially addressed here - https://github.com/theopensystemslab/planx-new/pull/7379

CI is failing this morning with the following error - 

```sh
E: Release file for http://deb.debian.org/debian-security/dists/bullseye-security/InRelease is expired (invalid since 12h 2min 30s). Updates for this repository will not be applied.
```

## What's the cause?
Our `postgres` Docker image relies on Debian 11 Bullseye which hit end of LTS on 31/08/26. As as a result the error above is thrown - there are no more security releases available for this OS. 

Debian releases here - https://www.debian.org/releases/

## What's the solution?
Upgrade! This PR takes us to Debian 13 which has an end of LTS date of 30/06/30.

I've shifted from the PostGIS image ([which does not have a newer Debian release for Postgres 16](https://hub.docker.com/r/postgis/postgis/tags)) across to plain [`postgres:16-trixie`](https://hub.docker.com/layers/library/postgres/16-trixie/images/sha256-f5745ce5089cd73eb8ed78267752540bb632b4083ce6d3f0dd80806b70b65edb) in order to get us up and running. The alternative (to stick with `postgis` Docker images) would be to upgrade PostgreSQL from 16 to 18 across both Docker and AWS RDS which I'm not too keen to get into just now. There's a knock on impact of switching away from PostGIS, see code comment on the migration below.

### DIFF
--- a/apps/hasura.planx.uk/migrations/default/1739787604658_create_postgis_extensions/up.sql
+++ b/apps/hasura.planx.uk/migrations/default/1739787604658_create_postgis_extensions/up.sql
@@ -1,0 +1,3 @@
+CREATE EXTENSION IF NOT EXISTS fuzzystrmatch;
+CREATE EXTENSION IF NOT EXISTS postgis_topology;
+CREATE EXTENSION IF NOT EXISTS postgis_tiger_geocoder;
--- a/apps/hasura.planx.uk/migrations/default/1739787604658_create_postgis_extensions/down.sql
+++ b/apps/hasura.planx.uk/migrations/default/1739787604658_create_postgis_extensions/down.sql
@@ -1,0 +1,3 @@
+DROP EXTENSION IF EXISTS postgis_tiger_geocoder;
+DROP EXTENSION IF EXISTS postgis_topology;
+DROP EXTENSION IF EXISTS fuzzystrmatch;
--- a/apps/postgres/Dockerfile
+++ b/apps/postgres/Dockerfile
@@ -1,5 +1,5 @@
-FROM postgis/postgis:16-3.5
+FROM postgres:16-trixie
 
 RUN apt-get update \
-  && apt-get install -y --no-install-recommends postgresql-16-cron \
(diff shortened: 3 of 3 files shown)


apps/hasura.planx.uk/migrations/default/1739787604658_create_postgis_extensions/up.sql
DafyddLlyr, line 1: **What is this migration doing?**

This is a **historic** migration, placed back in time just before https://github.com/theopensystemslab/planx-new/blob/main/apps/hasura.planx.uk/migrations/default/1739787604659_update%20extensions/up.sql#L1-L4 🕚 

Without this, Hasura will fail with the following error - 

```sh
level=error msg="extension \"fuzzystrmatch\" does not exist"
  statement: SELECT PostGIS_Extensions_Upgrade(); ... ALTER EXTENSION fuzzystrmatch UPDATE;
level=fatal msg="applying migrations failed on database(s): default"
```

This is because...
 - The previous `postgis` Docker image we used already invoked `CREATE EXTENSION IF NOT EXISTS fuzzystrmatch;`
 - The current `postgres` Docker image does not

I _think_ this migration will just be skipped when we hit `main` - if it runs it will be a no-op as the extensions already exist. There is a slight chance we hit an inconsistent metadata / other migration error on merge to `main` as a result of this time travelling migration. If this is the case I'll revert and run this SQL as a script / entrypoint on Docker setup, but using a migration like this felt a bit cleaner and more explicit. Opinions most welcome, I admit it is a little bit weird to be doing it like this.

freemvmt: It seems reasonable, I think I follow! I guess we'll find out when it hits main 😁

freemvmt (approved):
Clever 🤓

zz-hh-aa (approved):

PR #7355, opened 2026-09-03
[api][editor] finalise file types

freemvmt:
Re-enable all file types, with attendant fixes/countermeasures as discussed in [this comment](https://github.com/theopensystemslab/planx-new/pull/7039#pullrequestreview-4884317780).

Original PR was #7039, with a crucial intervention meanwhile in the form of #7279.

Relates to [this ticket](https://trello.com/c/kkt7dKkB) - see full technical plan there, for which it handles the 2 crucial remaining points on the technical plan which were blocking it from staging.

PR #7389, opened 2026-09-08
[editor] align dropzones & enforce file validation by extension

freemvmt:
Bundles some forward-fixes for the enabling of many more file types in #7355. Relates to [this ticket](https://trello.com/c/kkt7dKkB). See also relevant Slack thread [here](https://opensystemslab.slack.com/archives/C5Q59R3HB/p1788521820727899).

The big thing this PR brings is parity across the frontend/Editor and the API as to what file types are accepted for upload.

There are 2 instances of [react-dropzone](https://react-dropzone.js.org/) integrated into components in the Editor. Due to a quirk of how that package handles the `accept` option (it reads MIME type first and is satisfied by that even if extension doesn't match), the `Dropzone` component was allowing files through which would then be rejected by the API, making for pretty bad UX. Meanwhile the `PublicFileUploadButton` component had the same issue as well as no file size constraint.

So here we align the behaviour for both uses of `react-dropzone`, including adding custom validation to both to enforce file-type-validation-by-extension, all based on the canonical allowlist in the workspace `file-upload` package which the analogous logic in the API also references - so we have a single source of truth across the stack.

### DIFF
--- a/apps/editor.planx.uk/src/@planx/components/shared/extensionValidator.ts
+++ b/apps/editor.planx.uk/src/@planx/components/shared/extensionValidator.ts
@@ -1,0 +1,35 @@
+import { getFileExtension, isAllowedExtension } from "@planx/file-upload";
+import type { FileError } from "react-dropzone";
+import { ErrorCode } from "react-dropzone";
+
+/**
+ * Extension checks for react-dropzone's `validator` option.
+ * See: https://react-dropzone.js.org/examples/validator#custom-validation
+ *
+ * react-dropzone's `accept` logic accepts files over-zealousy. That is, it accepts a file if its browser-reported
+ * MIME type OR its extension matches, so any MIME type we include also admits every other extension the browser maps
+ * onto that type. For example, the catchall 'application/octet-stream' admits .exe or .com files, which we don't allow.
+ * We therefore validate strictly by extension on the frontend, so they never hit the API (which will reject them anyway).
+ *
+ * We deliberately reuse react-dropzone's own `file-invalid-type` code so these rejections flow through
+ * the same handling as its built-in ones (see handleRejectedUpload).
+ */
+const invalidType = (file: File): FileError => ({
+  // we reuse react-dropzone's own `file-invalid-type` code so rejections flow through handleRejectedUpload
+  code: ErrorCode.FileInvalidType,
+  message: `File extension not allowed: ${file.name}`,
+});
+
+export const validateExtension = (file: File): FileError | null =>
+  isAllowedExtension(file.name) ? null : invalidType(file);
+
+/**
+ * As above, but for dropzones accepting a narrower list than the full allowlist (e.g. images only).
+ * Pass the extensions - with leading dots - which this dropzone accepts.
+ */
+export const createExtensionValidator =
+  (allowedExtensions: string[]) =>
+  (file: File): FileError | null =>
+    allowedExtensions.includes(getFileExtension(file.name))
+      ? null
+      : invalidType(file);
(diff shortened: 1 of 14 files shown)


apps/editor.planx.uk/src/@planx/components/shared/extensionValidator.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Extension checks for react-dropzone's `validator` option.
+ * See: https://react-dropzone.js.org/examples/validator#custom-validation
+ *
+ * react-dropzone's `accept` logic accepts files over-zealousy. That is, it accepts a file if its browser-reported
+ * MIME type OR its extension matches, so any MIME type we include also admits every other extension the browser maps
+ * onto that type. For example, the catchall 'application/octet-stream' admits .exe or .com files, which we don't allow.
+ * We therefore validate strictly by extension on the frontend, so they never hit the API (which will reject them anyway).
+ *
```
joecarver: helpful comment!

packages/file-upload/allowedFileTypes.ts
```diff
@@ -47,7 +52,7 @@ export const ALLOWED_EXTENSIONS_BY_MIME_TYPE: Record<string, string[]> = {
   "video/x-ms-wmv": [".wmv"],
   // GML (Geographic Markup Language)
   "application/gml+xml": [".gml"],
-  // binary files without custom MIME, for which we fall back to catchall octet-stream
+  // binary files with no registered and reliable MIME type, so we fall back to catchall octet-stream
```
DafyddLlyr, line 55: Helpful!

packages/file-upload/allowedFileTypes.ts
```diff
@@ -73,8 +78,25 @@ export const PREVIEWABLE_MIME_TYPES: ReadonlySet<string> = new Set([
...
  */
 export const ALLOWED_EXTENSIONS: string[] = Array.from(
   new Set(Object.values(ALLOWED_EXTENSIONS_BY_MIME_TYPE).flat()),
 );
+
+/**
+ * Returns lowercased extension, with leading dot, e.g. '.pdf'.
+ * Returns  "" for a filename with no extension, or for dotfiles ().
+ *
+ * Modelled on node's `path.extname`, which we avoid using so this package can be imported by the frontend.
```
DafyddLlyr, line 91: Was just going to ask this question...! 😅

apps/editor.planx.uk/src/ui/editor/ImgInput/ImgInput.tsx
DafyddLlyr, line 1: Nice catch!

apps/editor.planx.uk/src/ui/shared/PublicFileUploadButton.tsx
```diff
@@ -3,14 +3,28 @@ import Image from "@mui/icons-material/Image";
...
 import { useCallback, useEffect, useState } from "react";
-import type { FileWithPath } from "react-dropzone";
+import type { FileRejection, FileWithPath } from "react-dropzone";
 import { useDropzone } from "react-dropzone";
 
 export type AcceptedFileTypes = Record<string, ImageFileExtensions[]>;
 
-export type ImageFileExtensions = ".jpg" | ".jpeg" | ".png" | ".svg" | ".ico";
+// all raster and vector graphic extensions from canonical allowlist in file-upload package
+export type ImageFileExtensions =
```
DafyddLlyr, line 17: There's probably a neat TS way of making this a valid subset of the types set out in the package, with their associated category (i.e. `image/*`) using `infer`.

This is just for DX / type-safety though really, all the runtime validation below is still needed.

I'll have a quick timeboxed look at this later 👍

DafyddLlyr (approved):
Great...!

apps/editor.planx.uk/src/ui/shared/PublicFileUploadButton.tsx
freemvmt: Yes I did think about doing something like that, but then we're actually not catching **all** `image/*` MIME types from the map - i.e. the `.dwg` and `.dxf` extensions - so then doing some clever TS stuff seemed like more pain that it was worth. but very happy if you want to go for it!

PR #7390, opened 2026-09-08
[api][editor] wait on public file scans before returning

freemvmt:
Builds on #7389. Again, relates to [this ticket](https://trello.com/c/kkt7dKkB), and relevant Slack thread [here](https://opensystemslab.slack.com/archives/C5Q59R3HB/p1788521820727899).

This time we handle another issue - that of logos not appearing after upload in [team settings](https://editor.planx.dev/app/opensystemslab/settings/design), because it can't be fetched until it's been scanned. Like so:

image

This was a general problem we'd have seen for any public file upload, so here we add logic such that the Editor tries successively to get a 200 from the API, rather than bailing out on the first 503.

[MDN on CORP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Resource-Policy) may be a useful reference here, I had to reap up :nerd_face:

### DIFF
--- a/apps/api.planx.uk/modules/file/controller.ts
+++ b/apps/api.planx.uk/modules/file/controller.ts
@@ -9,6 +9,7 @@ import {
   getFileFromS3,
   type GetFileResult,
   headFileInS3,
+  PUBLIC_FILE_CROSS_ORIGIN_RESOURCE_POLICY,
 } from "./service/getFile.js";
 import { uploadPrivateFile, uploadPublicFile } from "./service/uploadFile.js";
 import { buildFilePath, safeDecode } from "./service/utils.js";
@@ -96,6 +97,25 @@ type DownloadNext = Parameters<DownloadController>[2];
 const SCAN_RETRY_AFTER_SECONDS = 30;
 
 /**
+ * Public files are worth caching. Team logos in particular are served on many pages,
+ * and each uncached hit costs us a full S3 read into memory.
+ *
+ * Without this, browsers fall back to heuristic freshness based on `Last-Modified`.
+ * See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching
+ */
+const PUBLIC_FILE_CACHE_CONTROL_SUCCESS = {
+  "cache-control": "public, max-age=3600",
+} as const;
+
+/**
+ * Note the public GET route has no useNoCache (unlike private), so we set this on error responses to
+ * keep failures out of caches, e.g. a cached 503 would leave a file looking broken when it's servable.
+ */
+const PUBLIC_FILE_CACHE_CONTROL_SUCCESS_FAILURE = {
+  "cache-control": "no-store",
+} as const;
+
+/**
  * Map a failed read of the user-data bucket onto an HTTP response.
  *
  * Request bodies should carry a stable code: councils and their integrations can branch on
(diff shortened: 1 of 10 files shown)


apps/api.planx.uk/modules/file/service/getFile.ts
```diff
@@ -16,7 +27,7 @@ import { s3Factory } from "./utils.js";
 const buildHeaders = (file: GetObjectCommandOutput) => {
   const headers: Record<string, string> = {
     // allows the editor to load public-bucket images cross-origin
-    "cross-origin-resource-policy": "cross-site",
```
freemvmt, line 19: This was a mistake from earlier PR - this option doesn't exist!

apps/api.planx.uk/modules/file/controller.ts
```diff
@@ -95,6 +96,25 @@ type DownloadNext = Parameters<DownloadController>[2];
...
 const SCAN_RETRY_AFTER_SECONDS = 30;
 
+/**
+ * Public files are worth caching. Team logos in particular are served on many pages,
+ * and each uncached hit costs us a full S3 read into memory.
+ *
+ * Without this, browsers fall back to heuristic freshness based on `Last-Modified`.
+ * See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching
+ */
+const PUBLIC_FILE_CACHE_CONTROL_SUCCESS = {
```
freemvmt, line 106: These are really just drive by improvements to caching

apps/editor.planx.uk/src/lib/api/fileUpload/requests.ts
```diff
@@ -33,3 +34,50 @@ export const uploadPrivateFile: UploadFunction = async (file, onProgress) =>
...
   handleUpload(file, "/file/public/upload", onProgress);
+
+/**
+ * Delay before each retry when waiting for an uploaded file to become servable.
+ *
+ * Most objects are scanned in less than 3s, but it can take significantly longer,
+so we ramp up the delay after each attempt, ending on the API's blanket 30s `Retry-After`.
+ */
+export const DEFAULT_POLL_DELAYS_MS = [
+  1_000, 2_000, 4_000, 8_000, 15_000, 30_000, 30_000,
```
freemvmt, line 45: Most public files are going to be small PNGs, SVGs etc, so we'd expect Scanii to do its thing within 1s. Certainly most files are handled within 3s, based on the last 2 weeks of data on prod.

See panels for the `Submit` and `Callback` lambdas respectively:

Image

Image

apps/editor.planx.uk/src/lib/api/fileUpload/requests.ts
```diff
@@ -33,3 +34,50 @@ export const uploadPrivateFile: UploadFunction = async (file, onProgress) =>
...
+/**
+ * Wait until an uploaded file can actually be fetched.
+ *
+ * An upload's 200 only means bytes reached S3. The API will not serve a file until Scanii has tagged it,
+ * answering `503 FILE_SCAN_PENDING` with a `Retry-After` until then, so an URL handed back will be broken
+ * for as long as the scan takes. Callers should wait on this before treating an upload as complete.
+ *
+ * Bonus: `delaysMs` is injectable so tests need no fake timers.
+ */
+export const waitForPublicFile = async (
```
DafyddLlyr, line 59: I had originally hoped we could solve this with the `` elements `onLoad` and/or `onError` but on reading up it seems that was likely a little naive 👍

DafyddLlyr (approved):
Great! Tested and working as expected here 👌

apps/editor.planx.uk/src/lib/api/fileUpload/requests.ts
freemvmt: Yep totally, I thought the same at first and did investigate it, because it's what we do for a limited list (`PREVIEWABLE_MIME_TYPES`) of file types in the case of private file uploads (i.e. just show the local file in-memory on the client with `blob:*`) - but in that case we don't have the option to retrieve them from the API and show in browser.

Also here one of the file types commonly expected for public uploads is SVGs for logos, which being an 'active' file type, we have expressly not included in the above-mentioned list, so seemed better to build out the proper wait functionality.

PR #7393, opened 2026-09-09
feat: search query optimisations

joecarver:
Write a new migration to change the vector generation and the query execution:

- Team name is now included in the search vector
  - this required using manual triggers rather than a `GENERATED ALWAYS` column
- Only relevant columns are returned in the search query as it was taking a while to serialize the results (mainly due to `flows.data`)
  - this required returning explicit NULL values for all non-needed columns, so we can continue to return results like `RETURNS SETOF public.flow` from the function
- Match prefixes as well as lexemes (e.g. so `appl` matches `apply` and `application` etc)

### DIFF
--- a/apps/hasura.planx.uk/migrations/default/1788400000000_search_flows_include_team_name_and_limit_columns/up.sql
+++ b/apps/hasura.planx.uk/migrations/default/1788400000000_search_flows_include_team_name_and_limit_columns/up.sql
@@ -56,10 +56,17 @@ SET search_vector = public.compute_flow_search_vector(f.name, t.name, f.summary,
 FROM public.teams t
 WHERE f.team_id = t.id;
 
+
+--
 -- return null values for columns we don't need in the search results
 -- this improves search performance
 CREATE OR REPLACE FUNCTION public.search_flows(search text)
 RETURNS SETOF public.flows AS $$
+  WITH q AS (
+    -- prefix-match each word in the search term (e.g. "appl" matches "apply", "application")
+    SELECT to_tsquery('english', string_agg(lexeme || ':*', ' & ')) AS tsq
+    FROM unnest(tsvector_to_array(to_tsvector('simple', search))) AS lexeme
+  )
   SELECT
     id,
     team_id,
@@ -88,8 +95,8 @@ RETURNS SETOF public.flows AS $$
     NULL::boolean AS is_service,
     NULL::boolean AS is_pattern,
     NULL::tsvector AS search_vector
-  FROM public.flows
+  FROM public.flows, q
   WHERE deleted_at IS NULL
-    AND search_vector @@ plainto_tsquery('english', search)
-  ORDER BY ts_rank(search_vector, plainto_tsquery('english', search)) DESC;
+    AND search_vector @@ q.tsq
+  ORDER BY ts_rank(search_vector, q.tsq) DESC;
 $$ LANGUAGE sql STABLE;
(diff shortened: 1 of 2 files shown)


apps/hasura.planx.uk/migrations/default/1788400000000_search_flows_include_team_name_and_limit_columns/up.sql
```diff
@@ -0,0 +1,102 @@
...
+    NULL::boolean AS is_listed_on_lps,
+    NULL::text AS category,
+    NULL::uuid AS submission_email_id,
+    NULL::timestamptz AS deleted_at,
+    NULL::text AS email_template,
+    NULL::boolean AS is_service,
+    NULL::boolean AS is_pattern,
+    NULL::tsvector AS search_vector
+  FROM public.flows, q
+  WHERE deleted_at IS NULL
```
jessicamcinchak, line 99: Per testing feedback here, I think we also need to explicitly filter out `archived` flows, not just `deleted`? https://opensystemslab.slack.com/archives/C5Q59R3HB/p1788350690391159

joecarver: ah yes - was working on this in a separate branch but you're right, should be applied at query level.

joecarver: https://github.com/theopensystemslab/planx-new/pull/7393/commits/c7bfa1ea6193348e13d1a4e94ff1b4032c8bca4f

jessicamcinchak: I think `deleted_at` and `archived_at` need to be an `OR` not `AND` else we won't pick up _only_ archived flows here? (A flow can be only archived, not deleted; but can never be deleted without first being archived I think? So maybe _only_ `WHERE archived_at IS NOT NULL` is sufficient alone too?)

joecarver: Hmm, just writing out the cases here as I'm not sure...

- Flow is archived but not deleted: `deleted_at IS NULL` --> true, `archived_at IS NULL` --> false.
  - With AND --> `true AND false` = `false` --> flow correctly excluded 
  - With OR --> `true OR false` = `true` --> flow wrongly included

So I think we do want `AND` not `OR` as it currently is.

Or, as you suggested, only check `archived_at`, and ignore `deleted_at` - but this seems a little less robust? I don't think we have any DB constraints on the relationship between these two columns so maybe best to keep both in the query.

jessicamcinchak: If I test directly on the pizza, search is still returning archived (un-deleted) flows for me though :thinking: 

Screenshot from 2026-09-10 08-42-40
Screenshot from 2026-09-10 08-42-17

joecarver: ahh its probably because the Vultr Upsert doesn't completely reprovision the database, so any changes to already-applied migrations are not picked up. Let me destroy and recreate

jessicamcinchak: Also re: db constraint, we can simply check this! Even though not formally enforced at schema-level (agree this would be more robust option in future), this does return 0 as expected on prod :white_check_mark: 
```
select count(*) from flows
where archived_at is null
and deleted_at is not null;
```

joecarver: yeah can confirm the `archived_at` clause is applied now, having fully re-created the pizza, so should be working! I'd err towards leaving in the `deleted_at` clause also, just in case, but happy to remove if you prefer.

jessicamcinchak: Thanks for rebuilding pizza, and apologies for slow review loop here overall! 

Now getting expected results:
- :heavy_check_mark: Flows that are archived-only omitted from search
- :heavy_check_mark: Flows that are archived + deleted omitted from search

jessicamcinchak (approved):

PR #7394, opened 2026-09-09
refactor: use fire-and-forget approach for data sanitation cron trigger

zz-hh-aa:
This PR is an alternative version of #7369 following Daf's comment. Instead of splitting the one cron trigger into 4 separate ones (with the same 10 operations run), this trigger returns a 202 right away, the API runs the operation, and if any errors they are logged to Slack. 

I tested the endpoint locally (both a success and a failure--the latter by revoking API privileges on updating `lowcal_sessions` and saw a Slack notification as expected. I could also create some dummy data to test the actual cron part.

### DIFF
--- a/apps/api.planx.uk/modules/webhooks/controller.ts
+++ b/apps/api.planx.uk/modules/webhooks/controller.ts
@@ -3,6 +3,7 @@ import isNull from "lodash/isNull.js";
 import { ServerError } from ".././../errors/index.js";
 import { sendSlackMessage } from "../slack/utils.js";
 import { analyzeSessions } from "./service/analyzeSessions/index.js";
+import type { AnalyzeSessions } from "./service/analyzeSessions/types.js";
 import { getDailyFailedSubmissions } from "./service/dailyFailedSubmissions/index.js";
 import type { FailedSubmissionController } from "./service/dailyFailedSubmissions/types.js";
 import { softDeleteSession } from "./service/deleteSession/index.js";
@@ -217,9 +218,11 @@ export const deleteSessionController: DeleteSessionController = async (
 export const sanitiseApplicationDataController: SanitiseApplicationData =
   async (_req, res, next) => {
     try {
-      const { operationFailed, results } = await sanitiseApplicationData();
-      if (operationFailed) res.status(500);
-      return res.json(results);
+      res.status(202).json({ message: "Sanitation job started" });
+
+      sanitiseApplicationData().catch((error) => {
+        console.error("Unhandled error in sanitiseApplicationData", error);
+      });
     } catch (error) {
       return next(
         new ServerError({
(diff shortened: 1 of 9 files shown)


apps/hasura.planx.uk/metadata/databases/default/functions/public_search_flows.yaml
```diff
@@ -2,6 +2,7 @@ function:
   name: search_flows
   schema: public
 configuration:
+  custom_root_fields: {}
```
zz-hh-aa, line 5: This and below (in `public_team_integrations`) are auto-generated Hasura metadata changes

apps/hasura.planx.uk/metadata/cron_triggers.yaml
```diff
@@ -30,11 +30,6 @@
   schedule: 0 2 * * *
   include_in_metadata: true
   payload: {}
-  retry_conf:
```
zz-hh-aa, line 33: Deleted since reset back to defaults (no longer need 120 second timeout)

apps/api.planx.uk/modules/webhooks/controller.ts
```diff
@@ -217,9 +218,11 @@ export const deleteSessionController: DeleteSessionController = async (
 export const sanitiseApplicationDataController: SanitiseApplicationData =
   async (_req, res, next) => {
     try {
-      const { operationFailed, results } = await sanitiseApplicationData();
-      if (operationFailed) res.status(500);
-      return res.json(results);
+      res.status(202).json({ message: "Sanitation job started" });
```
freemvmt, line 221:
```suggestion
      res.status(202).json({ message: "Sanitisation job started" });
```

? this is a super nit... I'm basically unsure of the difference between 'sanitation' and 'sanitisation' haha

entirely in your hands! just a curiosity 🤓

freemvmt (approved):
This is great Ollie! Seems such a simple change.

I think as long as failures still post to Slack, as you tested, this is perf ✨

apps/api.planx.uk/modules/webhooks/controller.ts
zz-hh-aa: TIL sanitation is more about cleaning (eg surfaces) and sanitisation is to do with data... I was going to update but we use 'sanitation' fairly frequently in the codebase--so I am going to leave it!

DafyddLlyr: Yep I think this is possibly my bad second language English coming back to bite us sorry 😂

DafyddLlyr (commented):
Agreed - comparing the two PRs this seems like the solution which is both simpler and more maintainable.

apps/api.planx.uk/modules/webhooks/controller.ts
zz-hh-aa: Easily swapped!

PR #7395, opened 2026-09-09
Search UI tweaks

joecarver:
Implements suggested changes from the [Slack thread](https://opensystemslab.slack.com/archives/C5Q59R3HB/p1788350690391159):

- Hide 'Testing' team's flows from results
- 'Templates' and 'Flows I can copy' chips show all flows in the set until a term is entered
- Details panel shows on result hover as well as click
- Use more chip variants for styling


ianjon3s (changes requested):
Working really well, couple of comments:

Selection on hover feels snappy, but it creates narrow channels that must be navigated to keep a selection (even after clicking). Selecting the last card in the list illustrates this well:

image

I suggest we keep selection to click only based on this.

Also a bit of gap needed between chips:
image

joecarver:
thanks @ianjon3s - have changed these now

ianjon3s (approved):
Great work 🚀

PR #7042, opened 2026-07-21
feat: notes feature - full stack

joecarver:
**What's in this PR?**
- all other notes PRs in the stack have been merged into here now

PR #7398, opened 2026-09-09
chore[alt notes]: revert `FlowNotes` across API, Hasura, and seed scripts

jessicamcinchak:
Follows on from #7330 as final cleanup of alt-notes proposal

Essentially manually reverts #7082 and #7151

**To test:**
- Confirm pizza data sync has succeeded, omitting `flow_notes_*` tables
- Confirm you can still copy flows as expected (with unique IDs, published by default on creation)


apps/api.planx.uk/modules/flows/copyFlow/service.ts
```diff
@@ -23,38 +17,17 @@ const copyFlow = async ({
...
+    await createFlow({
+      teamId,
+      slug,
+      name,
+      isTemplate: false,
+      flowData: uniqueFlowData,
+      copiedFrom: flowId,
+      isService: flow.isService,
+      isPattern: flow.isPattern,
+    });
```
jessicamcinchak, line 30: `/copy` endpoint no longer has make multiple requests to `flow_notes_content` and `flow_note_positions` so postgres function (which offered more transaction control) has been dropped. 

This file, and its' test suite, matches latest commit history pre-FlowNotes again.

apps/hasura.planx.uk/migrations/default/1788992417384_drop_copy_function_flow_note_tables/up.sql
```diff
@@ -0,0 +1,5 @@
+DROP FUNCTION IF EXISTS public.copy_flow(uuid, integer, text, text, jsonb, boolean, boolean, text, integer);
+DROP FUNCTION IF EXISTS public.remap_note_placement(jsonb, text);
+DROP FUNCTION IF EXISTS public.rename_node_id(text, text);
+DROP TABLE IF EXISTS public.flow_note_content CASCADE;
+DROP TABLE IF EXISTS public.flow_note_positions CASCADE;
```
jessicamcinchak, line 5: Only `copy_flow` was created as a "tracked" function, but we want to drop all functions that were part of original "up" migration so that they don't silently linger in our schema !

apps/hasura.planx.uk/metadata/databases/default/tables/tables.yaml
```diff
@@ -29,8 +29,6 @@
 - "!include public_feedback_type_enum.yaml"
 - "!include public_flow_comments.yaml"
 - "!include public_flow_history.yaml"
-- "!include public_flow_note_content.yaml"
-- "!include public_flow_note_positions.yaml"
```
jessicamcinchak, line 33: Hasura automatically deleted these two rows as expected, but did _not_ automatically drop the corresponding individual yaml files here. I think they were never actually properly "tracked" (which the "Hasura change summary" seems to confirm)? 

I've manually deleted each instead. Same goes for `copy_flow` in `/functions`

apps/api.planx.uk/helpers.ts
```diff
@@ -425,10 +425,6 @@ const getChildren = (
  * (replace last n characters), leaving `_root` and nullish ids untouched.
  */
 function renameNodeId(nodeId: string, replaceValue: string): string;
-function renameNodeId(
-  nodeId: string | null | undefined,
-  replaceValue: string,
-): string | null | undefined;
```
jessicamcinchak, line 431: This is a really strange block where `renameNodeId` (introduced in #7042) is defined 3x times over. I _think_ it's attempting to workaround known tedious type issues with `flowData[node]` and `flowData[node]["edges"]` possibly being `undefined` ? 

I've been able to successfully remove 1 instance, but there's still 2. Can anyone else fill in the blanks for me here of why?

DafyddLlyr: This is an example of an overloaded function (it's basically an alternative way of declaring three valid types for this function).

I think this is essentially right as is - we're saying if we take in a `string` argument, we return one (1st overload). If we take in a `string | null | undefined`, we return a `string | null | undefined`.

The alternative way of declaring this would be to use generic (this is what we tend to use in the codebase instead of overloads).

This is the same as something like this - 

```ts
function renameNodeId<T extends string | null | undefined>(
  nodeId: T,
  replaceValue: string,
): T {
```

DafyddLlyr: Actually, can we just drop the second definition here?

The following works for me and doesn't through any type check errors - 

```ts
function renameNodeId(nodeId: string, replaceValue: string): string {
  if (!nodeId || nodeId === "_root") return nodeId;
  return nodeId.slice(0, -replaceValue.length) + replaceValue;
}
```

DafyddLlyr (approved):
- Confirm pizza data sync has succeeded, omitting `flow_notes_*` tables ✅ 
 - Confirm you can still copy flows as expected (with unique IDs, published by default on creation) ✅

apps/api.planx.uk/helpers.ts
jessicamcinchak: Thanks for explanation and suggestion here! 

Indeed works (see [ea7e3d6](https://github.com/theopensystemslab/planx-new/pull/7398/commits/ea7e3d6be523d636dfad01a4fdbc1d3f5e9e67ea)), I had originally tried dropping the first definition thinking I needed to keep the more complex type which did throw errors, but this seems to work a charm!

PR #7396, opened 2026-09-09
chore[alt notes]: refactor `showNoteInTemplatedFlow` helper and add tests

jessicamcinchak:
Previously introduced a bit of repeated logic to determine visibility among standalone `Note` and `AttachedNote` in templated flows, now refactored to shared function with test coverage :white_check_mark:

PR #7399, opened 2026-09-09
fix [alt notes]: account for undefined ID on initial component creation and correctly show in graph

jessicamcinchak:
Realised I introduced a small bug to staging upon merging #7396 

To test: 
- On staging: toggle "show notes" on in a regular, non-templated flow. Add a standalone "Note" component and it doesn't display on graph :x: 
- On this pizza: repeat and expect "Note" to display as corrected !

### DIFF
--- a/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Node.tsx
+++ b/apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Node.tsx
@@ -107,7 +107,6 @@ const Node: React.FC<any> = (props) => {
       //   from the source template unless within a templated folder
       const showNoteComponent =
         showNotes &&
-        node.id &&
         (!isTemplatedFrom ||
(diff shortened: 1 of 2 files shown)


apps/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Node.tsx
```diff
@@ -107,7 +107,6 @@ const Node: React.FC<any> = (props) => {
       //   from the source template unless within a templated folder
       const showNoteComponent =
         showNotes &&
-        node.id &&
```
jessicamcinchak, line 110: Realising `node.id` isn't guaranteed here upon first creation of a new `Node`! Therefore we can't require here but instead we can handle `string | undefined` argument in helper function.

zz-hh-aa (approved):
Works as expected, thanks!

DafyddLlyr (approved):

PR #7401, opened 2026-09-10
feat: Add calendar explore widget

ianjon3s:
## What does this PR do?

- Embeds PlanX google calendar as an explore page widget
- Reuses `ToggleButtonGroup` with `StyledToggleButton` to switch between calendar views
- Reuses `localStorage` method used for `flowCardView` to persist user selection of view

image
image


apps/editor.planx.uk/src/pages/Explore/components/CalendarWidget.tsx
```diff
@@ -0,0 +1,108 @@
...
+const VIEW_PARAMS: Record<CalendarView, string> = {
+  grid: "&mode=MONTH&wkst=2",
+  list: "&mode=AGENDA",
+};
+
+const embedUrl = (view: CalendarView) =>
+  `https://calendar.google.com/calendar/embed?src=${CALENDAR_SRC}` +
+  "&ctz=Europe%2FLondon" +
+  VIEW_PARAMS[view] +
+  "&showTitle=0&showPrint=0&showTabs=0&showCalendars=0&showTz=0";
```
DafyddLlyr: nit: Managing this as `URLSearchParams()` is probably a little cleaner than using a raw string.

apps/editor.planx.uk/src/ui/editor/DashboardWidget/index.tsx
```diff
@@ -80,7 +82,7 @@ export const DashboardWidget: React.FC<DashboardWidgetProps> = ({
           />
         )}
       </Box>
-      {link && <WidgetLink {...link} />}
+      {headerAction ?? (link && <WidgetLink {...link} />)}
```
DafyddLlyr: nit: This is a little complex - we've got two completing props for this single slot. I would see `link` as just a type of headerAction component right.

The challenge here is we need `headerAction` for styling above. If there's no way to unify or simplify this we might want to look at setting stricter (and more correct) props that describe this setup. Right now a dev could pass in both a link and a header action, but headerAction would win out due to this line.

```ts
type HeaderSlot =
  | { link?: never; headerAction?: never } // neither
  | { link: WidgetLinkOptions; headerAction?: never } // link only
  | { headerAction: React.ReactNode; link?: never }; // action only
```

DafyddLlyr (approved):
Great! A few small nits below, no showstoppers.

apps/editor.planx.uk/src/pages/Explore/components/CalendarWidget.tsx
ianjon3s: Agreed, a much cleaner way to show options, updated in https://github.com/theopensystemslab/planx-new/pull/7401/changes/dadedf96d8d45a6fa8d2e8df893f17437c498a45

apps/editor.planx.uk/src/ui/editor/DashboardWidget/index.tsx
ianjon3s: Thanks for this, makes a lot of sense to refactor. I think I've configured this in a way that forces a mutually exclusive selection (with styling maintained): https://github.com/theopensystemslab/planx-new/pull/7401/changes/5d03575ab2708ae7d1777c9a377c8263fccb871f

apps/editor.planx.uk/src/pages/Explore/components/CalendarWidget.tsx
DafyddLlyr: Yep this is much nicer - thanks!

apps/editor.planx.uk/src/ui/editor/DashboardWidget/index.tsx
DafyddLlyr: Fab - also much clearer and simpler imho! 👍 Slight cost of more complex types up front often can pay off like this.

DafyddLlyr (approved):

PR #7402, opened 2026-09-10
feat: platform_dashboard_stats - db table and hasura metadata

joecarver:
Wrote this a month or so ago and somehow forgot to create the PRs, so, here it is...

- Creates a postgres view to expose our overall stats, as required for the 'PlanX In Numbers' panel on the explore page
- Adds Hasura metadata and introspection tests


ianjon3s (approved):
Approving from functional pov, works great

apps/hasura.planx.uk/migrations/default/1785761000000_create_platform_dashboard_stats_view/up.sql
```diff
@@ -0,0 +1,65 @@
...
+WITH lpa_teams AS (
+  SELECT id, created_at
+  FROM teams
+  WHERE name NOT IN ('Open Digital Planning', 'Open Systems Lab', 'PlanX', 'Templates', 'Testing', 'WikiHouse')
+),
+filtered_flows AS (
+  SELECT f.id, f.status
+  FROM flows f
+  JOIN lpa_teams t ON t.id = f.team_id
+  WHERE f.deleted_at IS NULL
```
jessicamcinchak, line 11: nit: I'd expect we want to filter 'archived' here same as discussed in search, rather than deleted only?

apps/hasura.planx.uk/migrations/default/1785761000000_create_platform_dashboard_stats_view/up.sql
```diff
@@ -0,0 +1,65 @@
+CREATE OR REPLACE VIEW "public"."platform_dashboard_stats" AS
+WITH lpa_teams AS (
+  SELECT id, created_at
+  FROM teams
+  WHERE name NOT IN ('Open Digital Planning', 'Open Systems Lab', 'PlanX', 'Templates', 'Testing', 'WikiHouse')
```
zz-hh-aa, line 5: Not sure if we discussed already but do we want to filter out Trial teams? 

We also have a couple other non LPA teams: 
- Council Onboarding
- Environment Agency (Trial)
- Historic England (Trial)
- Lambeth Sandbox (Trial)
- Madetech (Trial)
- Planning Advisory Service (Trial)
- Scotland (Trial)
- TPX (Trial)

Omitted LPA Trial teams from the above list ^

apps/hasura.planx.uk/migrations/default/1785761000000_create_platform_dashboard_stats_view/up.sql
```diff
@@ -2,7 +2,12 @@ CREATE OR REPLACE VIEW "public"."platform_dashboard_stats" AS
 WITH lpa_teams AS (
   SELECT id, created_at
   FROM teams
-  WHERE name NOT IN ('Open Digital Planning', 'Open Systems Lab', 'PlanX', 'Templates', 'Testing', 'WikiHouse')
+  WHERE name NOT IN (
+    'Open Digital Planning', 'Open Systems Lab', 'PlanX', 'Templates', 'Testing', 'WikiHouse',
+    'Council Onboarding', 'Environment Agency (Trial)', 'Historic England (Trial)',
```
zz-hh-aa, line 7: Sorry @joecarver only just saw these--I maybe wasn't so clear in my comment before!

The team names don't actually include the string `Trial`, I just marked which ones were--trial teams are stored in in `team_settings.is_trial` and include a few more LPAs.

PR #7403, opened 2026-09-10
feat: hook up NumbersWidget with real stats

joecarver:
Adds graphql query and hook, passing the stats to the Widget.


ianjon3s (approved):
Approving from functional pov, works great

PR #7384, opened 2026-09-07
fix(temp): Drop timeout from Vultr update step

DafyddLlyr:
Please see thread - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1788520367596309

This is a temporary drop the the `command_timeout` on the Vultr update step whilst I continue to debug this. Dropping the line means CI is not held up whilst I get to the bottom of this one.

[The previous fix](https://github.com/theopensystemslab/planx-new/pull/7372) appears to have worked (no rebuild I can spot in logs), but we're still hitting the 10 min mark on ocassion.

```sh
  net.ipv6.conf.all.disable_ipv6 = 1
  net.ipv6.conf.default.disable_ipv6 = 1
  latest: Pulling from planx/caddy-vultr
  Digest: sha256:139b7072c51b3b160fd4841e37806d8ea18b742a6e5f45bc05fc9910ff0662c3
  Status: Image is up to date for lhr.vultrcr.com/planx/caddy-vultr:latest
  lhr.vultrcr.com/planx/caddy-vultr:latest
  Caddy image pulled successfully
  ```

PR #7407, opened 2026-09-10
[ci] bump vultr update timeout to 15m

freemvmt:
We commented out the `command_timeout` for the Vultr update step in #7384, but it hasn't worked because that just drops us down to the [appleboy/ssh-action default timeout](https://github.com/appleboy/ssh-action#%EF%B8%8F-ssh-command-settings), which is also 10 minutes!

So bumping it here to 15 mins explicitly to hopefully abate CI problems - intended as a temporary measure until we resolve whatever is slowing things down.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -364,7 +364,7 @@ jobs:
           host: ${{ env.FULL_DOMAIN }}
           username: root
           key: ${{ secrets.SSH_PRIVATE_KEY }}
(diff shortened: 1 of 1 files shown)


DafyddLlyr (approved):
Thank you - I totally did not check the default value 😅

DafyddLlyr:
@freemvmt I'm going to go ahead and merge this as we're hitting this issues across a few PRs currently.

PR #7408, opened 2026-09-10
chore: Stripe database migrations

jessicamcinchak:
**Covers** 
- https://github.com/orgs/theopensystemslab/projects/11/views/1?pane=issue&itemId=223916455
- https://github.com/orgs/theopensystemslab/projects/11/views/1?pane=issue&itemId=223918875
- https://github.com/orgs/theopensystemslab/projects/11/views/1?pane=issue&itemId=223919437

**Key decisions**
- We do _not_ want to change any existing Gov Pay columns/functionality and introduce Stripe in an additive-only way as much as possible
- This specifically means:
  - We'll add new `stripe_`-prefixed columns in `payment_status` & `payment_requests` and we'll create a new table `stripe_payment_status_enum` rather than trying to reconcile/combine Gov Pay & Stripe statuses

**Changes here**
- Add new columns and tables
- Configure "check constraints" to ensure each record can only reflect a single provider (Gov Pay or Stripe)
- Configure permissions to match existing Gov Pay permissions
- Add introspection test coverage

:white_check_mark:  Regression tests passed against this branch https://github.com/theopensystemslab/planx-new/actions/runs/34839935176

**Not covered here / follow-up PRs** 
- Align planx-core's `PaymentRequest` type? https://github.com/theopensystemslab/planx-core/blob/main/src/types/payment-request.ts (it's currently omitting `_metadata` columns so I don't expect to fail in meantime? also see regression test success for confidence)
- Update nightly sync scripts once new columns are on prod
- Populate "real" values for Stripe payment status enums (via migration or just direct on prod?)
- Handle `lowcal_sessions.data.govUkPayment` and `lowcal_sessions.data.passport.data.["application.fee.reference.govPay"]`

### DIFF
--- a/apps/hasura.planx.uk/metadata/databases/default/tables/public_payment_status.yaml
+++ b/apps/hasura.planx.uk/metadata/databases/default/tables/public_payment_status.yaml
@@ -11,6 +11,12 @@ object_relationships:
         remote_table:
           name: lowcal_sessions
           schema: public
+  - name: payment_status_enum
+    using:
+      foreign_key_constraint_on: status
+  - name: stripe_payment_status_enum
+    using:
+      foreign_key_constraint_on: stripe_status
 insert_permissions:
   - role: api
     permission:
--- a/apps/hasura.planx.uk/migrations/default/1789071868856_squashed/up.sql
+++ b/apps/hasura.planx.uk/migrations/default/1789071868856_squashed/up.sql
@@ -1,0 +1,66 @@
+-- PAYMENT REQUESTS: new columns and constraints
+alter table "public"."payment_requests" 
+add column "stripe_payment_id" text null unique;
+
+alter table "public"."payment_requests" 
+add column "stripe_metadata" jsonb not null default '[]'::jsonb;
+
+-- payment_id and payment_metadata are nullable until ITP request is paid, so check needs to allow both null or max one not-null
+alter table "public"."payment_requests" 
+add constraint "single_provider_id_or_neither" 
+check (num_nonnulls(govpay_payment_id, stripe_payment_id) <= 1);
+
+alter table "public"."payment_requests" 
+add constraint "single_provider_metadata_or_neither" 
+check (num_nonnulls(govpay_metadata, stripe_metadata) <= 1);
+
+-- PAYMENT STATUS: new columns, constraints, and new enum table fkey
+alter table "public"."payment_status" 
+add column "stripe_metadata" jsonb null;
+
+alter table "public"."payment_status" 
+add column "stripe_payment_id" text null;
+
+alter table "public"."payment_status" 
+add column "stripe_status" text null;
+
+-- Existing columns need to become nullable, with exactly one not-null constraint instead
+alter table "public"."payment_status" 
+alter column "payment_id" drop not null;
+comment on column "public"."payment_status"."payment_id" is E'gov_pay_payment_id';
+
+alter table "public"."payment_status" 
+alter column "status" drop not null;
+comment on column "public"."payment_status"."status" is E'gov_pay_status';
+
+alter table "public"."payment_status" 
+add constraint "single_provider_id" 
+check (num_nonnulls(payment_id, stripe_payment_id) = 1);
+
+alter table "public"."payment_status" 
+add constraint "single_provider_status" 
+check (num_nonnulls(status, stripe_status) = 1);
+
+alter table "public"."payment_status" 
+add constraint "single_provider_metadata" 
+check (num_nonnulls(gov_pay_metadata, stripe_metadata) = 1);
+
+-- Create new enum table for Stripe, one existing record is required to toggle "Set table as enum" in console
+comment on table "public"."payment_status_enum" is E'GOV.UK Pay statuses';
+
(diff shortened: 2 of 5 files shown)


apps/hasura.planx.uk/metadata/databases/default/tables/public_payment_status.yaml
```diff
@@ -11,31 +11,43 @@ object_relationships:
         remote_table:
           name: lowcal_sessions
           schema: public
+  - name: payment_status_enum
+    using:
+      foreign_key_constraint_on: status
+  - name: stripe_payment_status_enum
+    using:
+      foreign_key_constraint_on: stripe_status
```
jessicamcinchak, line 19: These were Hasura-suggested "relationships" that make sense / surprised we didn't have them already !

apps/hasura.planx.uk/migrations/default/1789071868856_squashed/up.sql
```diff
@@ -0,0 +1,63 @@
...
+add column "stripe_payment_id" text null unique;
+
+alter table "public"."payment_requests" 
+add column "stripe_metadata" jsonb not null default '[]'::jsonb;
+
+-- payment_id is nullable until ITP request is paid, so check needs to allow one not-null OR both null
+--   same goes for payment_metadata, but jsonb uses `[]` default and is not nullable so don't enforce num_nonnulls check constraint here
+alter table "public"."payment_requests" 
+add constraint "single_provider_id_or_neither" 
+check (num_nonnulls(govpay_payment_id, stripe_payment_id) = 1 OR num_nulls(govpay_payment_id, stripe_payment_id) = 2);
```
jessicamcinchak, line 12: To expand on this comment about `_metadata` : 
- It's a bit odd/redundant that `payment_requests` captures `_metadata` at all as the corresponding `payment_status` record should always have it (and they should never differ) ! 
- `payment_status` allows `_metadata` to actually be `null`, whereas `payment_request` uses a default jsonb value and never-null. Rather than introduce inconsistent `check` constraints, I think we can just aim to retire `payment_request.*_metadata` columns in future? 
- Enforcing single `*_payment_id` will be enough alone to infer "provider" !

apps/hasura.planx.uk/migrations/default/1789071868856_squashed/up.sql
```diff
@@ -0,0 +1,63 @@
...
+CREATE TABLE "public"."stripe_payment_status_enum" (
+  "value" text NOT NULL, 
+  "comment" text NOT NULL, 
+  PRIMARY KEY ("value") , 
+  UNIQUE ("value")
+);
+COMMENT ON TABLE "public"."stripe_payment_status_enum" IS E'Stripe statuses';
+
+INSERT INTO "public"."stripe_payment_status_enum"("value", "comment") 
+VALUES (E'created', E'pending payment created');
```
jessicamcinchak, line 57: As noted above, this isn't a "real" value but at least one record is required to exist in table before can be toggled as "Set table as enum" in Hasura console !

apps/hasura.planx.uk/migrations/default/1789071868856_squashed/up.sql
```diff
@@ -0,0 +1,63 @@
...
+add constraint "single_provider_id" 
+check (num_nonnulls(payment_id, stripe_payment_id) = 1);
+
+alter table "public"."payment_status" 
+add constraint "single_provider_status" 
+check (num_nonnulls(status, stripe_status) = 1);
+
+alter table "public"."payment_status" 
+add constraint "single_provider_metadata" 
+check (num_nonnulls(gov_pay_metadata, stripe_metadata) = 1);
```
jessicamcinchak, line 43: Our decision here was to introduce equivalant `stripe_`-prefixed columns for every `gov_pay_`-prefixed preexisting column. Then a check constraint which says exactly one can be populated, other must be null.

This means we do _not_ have an explicit `payment_provider_enum` with values eg 'Stripe' | 'GovPay'` defined anywhere! We like this approach because it means we don't need to consider migrating/backfilling historic records in anyway.

In the future, there might be a nice use-case for a "computed column" to return plain "provider" to simplify analytics views, submissions log conditional displays, etc instead of always "coalescing" !

jessicamcinchak:
E2E and introspection tests both failing CI with docker error: 
```
minio Error pull access denied for minio/minio, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

Both pass successfully when ran locally for me :white_check_mark: Will retry tomorrow to see if remote issue/fluke?

apps/hasura.planx.uk/migrations/default/1789071868856_squashed/up.sql
DafyddLlyr: Yeah this is a bit odd - good catch.

I suspect it's possibly a permissions hangover - do we grant access to `payment_requests` but not `payment_status` for somebody invited to pay, and was `metadata` always on `payment_requests`?.

The solution might be to allow access to `payment_requests` (I think we touched on this the other day) to avoid this duplication as you suggest.

DafyddLlyr (approved):

PR #7410, opened 2026-09-11
feat(api): Stripe checkout endpoint

DafyddLlyr:
## What does this PR do?
This PR creates the `POST /stripe/checkout-session/:localAuthority` endpoint, which creates a hosted Stripe Checkout Session and returns its URL. Docs: https://docs.stripe.com/api/checkout/sessions/create

Many intentional `TODO`s left throughout the code, trying to keep this as small and simple as possible - there's no persistence, fee splitting, management of metadata or connected accounts etc here. This is just a way of making a simple payment via our API.

## Testing
You can test this via [the API docs](https://api.7410.planx.pizza/docs/#/stripe/post_stripe_checkout_session__localAuthority_) and should see the following behaviour - 

- I can make a request with the default params
- I get a Stripe Checkout URL returned
  - Amount is picked up via request (same as GovPay - we may want to think about this!)
  - Return URL works (if I hit back, or complete successfully)
- I can make a payment ([test card details here](https://docs.stripe.com/testing?locale=en-GB))
- I get returned to my supplied `returnURL` on success
- I can see the payment in PlanX sandbox account on success 

https://github.com/user-attachments/assets/bc46a266-691d-4809-a52b-d2d760e94cae


apps/api.planx.uk/modules/stripe/client.ts
```diff
@@ -0,0 +1,8 @@
+import assert from "assert";
+import Stripe from "stripe";
+
+assert(process.env.STRIPE_SECRET_KEY);
+
+export const stripe = new Stripe(process.env.STRIPE_SECRET_KEY, {
+  apiVersion: "2026-08-26.dahlia",
```
DafyddLlyr, line 7: Docs: https://docs.stripe.com/api/versioning?api-version=2026-08-26.preview&rds=1

apps/api.planx.uk/modules/stripe/client.ts
```diff
@@ -0,0 +1,8 @@
+import assert from "assert";
+import Stripe from "stripe";
+
+assert(process.env.STRIPE_SECRET_KEY);
+
+export const stripe = new Stripe(process.env.STRIPE_SECRET_KEY, {
```
DafyddLlyr, line 6: I'm using a module-singleton here to match [Stripe docs](https://github.com/stripe/stripe-node#usage). This means there's a single pooled connection.

apps/api.planx.uk/modules/stripe/client.ts
```diff
@@ -0,0 +1,8 @@
...
+  apiVersion: "2026-08-26.dahlia",
+});
```
joecarver, line 8: good shout to refactor here

apps/api.planx.uk/modules/stripe/checkout/controller.ts
```diff
@@ -0,0 +1,49 @@
...
+          },
+          quantity: 1,
+        },
+      ],
+      success_url: `${returnURL}?stripeSessionId={CHECKOUT_SESSION_ID}`,
+      cancel_url: `${returnURL}?cancelled=true`,
+      // TODO: Metadata population
+      metadata: { sessionId, flowId },
+      payment_intent_data: { metadata: { sessionId, flowId } },
+    });
```
joecarver, line 39: just to check as I can't remember where we landed on it, are we intending to store the checkout sessions ourselves as well? I think we decided that we would right, maybe this is just coming later?

apps/api.planx.uk/modules/stripe/checkout/controller.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Create a Stripe Checkout Session and return the URL
+ */
+export const createCheckoutSession: CreateCheckoutSessionController = async (
+  _req,
+  res,
+  next,
+) => {
+  const { localAuthority } = res.locals.parsedReq.params;
+  const { sessionId, flowId, amount, returnURL } = res.locals.parsedReq.body;
```
joecarver, line 14: it seems like you've thought of this but just to flag that `amount` won't be passed from the client, it'll be calculated server side based on what the transaction is

apps/api.planx.uk/modules/stripe/checkout/controller.ts
```diff
@@ -0,0 +1,49 @@
...
+          price_data: {
+            currency: "gbp",
+            product_data: { name: "Planning application fee" },
+            unit_amount: amount,
+          },
+          quantity: 1,
+        },
+      ],
+      success_url: `${returnURL}?stripeSessionId={CHECKOUT_SESSION_ID}`,
+      cancel_url: `${returnURL}?cancelled=true`,
```
joecarver: probably unlikely to happen but is there any world where we would want our return URL to already contain query parameters? if so this would break by having `?` twice in the query param string

joecarver (approved):
Solid stuff! Couple tiny questions / nits but I think you have it covered already

apps/api.planx.uk/modules/stripe/checkout/controller.ts
DafyddLlyr: Yep that's right, we want to derive from the user's passport ultimately as opposed to pass in from the frontend. It's here as a convenience for now (no need to setup and make valid sessions to test).

Most (all?) of the data needed for this endpoint can actually be derived from just a session ID.

apps/api.planx.uk/modules/stripe/checkout/controller.ts
DafyddLlyr: Yep thats's right, coming later alongside other db persistence steps 👍

apps/api.planx.uk/modules/stripe/checkout/controller.ts
DafyddLlyr: Great spot! Added here [43233f2](https://github.com/theopensystemslab/planx-new/pull/7410/commits/43233f27a3231d031a2d6631b434e52b78326ffb)

PR #7411, opened 2026-09-11
fix(a11y): LPS links open in new tab with screenreader text

ianjon3s:
## Issue

Relates to:
p.72 (v2.0 report) - DAC_CLL_Feedback_03
https://trello.com/c/gWjk5L0l/3752-usability-cognitive-feedback-03

External links on LPS do not consistently open in a new tab/window, and signal this by a visual arrow icon only.

## Solution

Ensure all external links open in a new tab, adding `(opens in a new tab)` as screenreader only text.

### Testing

### DIFF
--- a/apps/localplanning.services/astro.config.mjs
+++ b/apps/localplanning.services/astro.config.mjs
@@ -5,6 +5,7 @@ import sitemap from "@astrojs/sitemap";
 import tailwindcss from "@tailwindcss/vite";
 import { defineConfig, envField, fontProviders } from "astro/config";
 import icon from "astro-icon";
+import rehypeExternalLinks from "rehype-external-links";
 
 // BUILD_MODE must be an env var for Turbo caching
 // The dev / pizza / staging / prod builds must have their own cache-key
@@ -15,6 +16,20 @@ const isCloudfrontBuild = ["staging", "production"].includes(mode);
 export default defineConfig({
   site: process.env.SITE_URL,
   integrations: [react(), icon(), sitemap()],
+  markdown: {
+    rehypePlugins: [
+      [
+        rehypeExternalLinks,
+        {
+          target: "_blank",
+          rel: ["noopener", "noreferrer", "nofollow"],
+          properties: { className: ["external"] },
+          content: { type: "text", value: " (opens in a new tab)" },
+          contentProperties: { className: ["sr-only"] },
+        },
+      ],
+    ],
+  },
   env: {
     schema: {
       PUBLIC_PLANX_EDITOR_URL: envField.string({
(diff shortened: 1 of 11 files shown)


DafyddLlyr (commented):
Aware this is just a draft just now but two quick comments - 
 - We might want an `ExternalLink.astro` to handle this once consistently
 - For `.md` files it might be the case we can get `remark` to just apply the conventions we want consistently when converting to from markdown to html, as opposed to having to remember each time to follow a certain convention

ianjon3s:
Thanks for jumping in here @DafyddLlyr ! Agree on mitigating the repetition.

I've created an `ExternalLink.astro` component to reduce repetition in Astro templates. I've also sourced a plugin called [rehype-external-links](https://github.com/rehypejs/rehype-external-links) which works with `remark` to add default config for external links.

apps/localplanning.services/astro.config.mjs
DafyddLlyr, line 1: Perfect...! I remember trying to whack-a-mole PlanX's external links and keen to do anything to avoid this 😅 

This looks like a great solution.

apps/localplanning.services/src/content/generic/about.md
```diff
@@ -12,10 +12,10 @@ Built by your local planning team in collaboration with planning professionals,
 
 ## Open Digital Planning
 
-_'Find Local Planning Services'_ has been developed as part of <a href="https://opendigitalplanning.org/" class="external" target="_blank" rel="noopener noreferrer nofollow">Open Digital Planning</a> - a growing community of innovators, including over 150 Local Planning Authorities in England, planning experts, designers, and technologists. It is driven by a shared ambition to make planning more accessible and consistent for everyone, and create an open data ecosystem.
+_'Find Local Planning Services'_ has been developed as part of [Open Digital Planning](https://opendigitalplanning.org/) - a growing community of innovators, including over 150 Local Planning Authorities in England, planning experts, designers, and technologists. It is driven by a shared ambition to make planning more accessible and consistent for everyone, and create an open data ecosystem.
```
DafyddLlyr, line 15: Much nicer to user raw markdown here also 👍

apps/localplanning.services/src/components/applications/ApplicationCard.tsx
DafyddLlyr, line 1: Ah yep I didn't think about React components that can't re-embed `.astro` components, but the solutions here work well  - we can make an `ExternalLink.tsx` in future if we find ourselves repeating this a lot.

DafyddLlyr (approved):
This is fab - thanks for taking another look here - much better outcome!

apps/localplanning.services/src/components/applications/ApplicationCard.tsx
ianjon3s: Indeed this is the one anomaly, will keep this in mind in the case of further React work on this site 👍

PR #7415, opened 2026-09-11
fix: Team data failing to load

ianjon3s:
## Issue

Reported by @augustlindemer in Slack: 
https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1789128854532539

When accessing a team from the root / team select page, data is not load correctly, causing flow list and other entries to be empty.

## Solution

Moved `setTeam()` from the `$team` route's loader (which TanStack router's 30s preload cache was silently skipping after a hover) into a `useEffect` in the route component, which only mounts on actual navigation - fixing the root list hover+click case.


DafyddLlyr (approved):
Great!

PR #7416, opened 2026-09-11
feat: create `team_invoice_details` table in db

zz-hh-aa:
Backend for #7417 

- Created a new `team_invoice_details` table as it felt like a lot of columns to add to an existing one. 
    - Also migrates existing teams to the table
- Creating new teams will also create a new entry in `team_invoice_details` with `null` values (to be updated when teams update their settings)
- Permissions model is the same as other settings
- Introspection tests too


apps/editor.planx.uk/src/pages/Teams/hooks/useCreateTeam.tsx
```diff
@@ -7,6 +7,7 @@ const CREATE_TEAM_MUTATION = gql`
     $domain: String
     $settings: team_settings_insert_input!
     $theme: team_themes_insert_input!
+    $invoice_details: team_invoice_details_insert_input!
```
zz-hh-aa, line 10: Following the same pattern that `team_settings` and other tables use--create row with `null` values when team is created, and then we only really need `update` permissions.

jessicamcinchak (approved):
Great ! 

Once this table on prod, I think we'll also want to remember to add to nightly staging sync scripts for easier future Stripe testing across environments :bookmark:

apps/editor.planx.uk/src/pages/Teams/hooks/useCreateTeam.tsx
DafyddLlyr, line 1: (non-blocking): I think you'll also need to update instances of `insert_teams_one` in `planx-core`.

These are called to create teams within the E2E tests, and we may need these invoice details as part of a full Stripe E2E test at some point.

apps/hasura.planx.uk/tests/team_invoice_details.js
DafyddLlyr, line 1: Does this get run? Vitest is matching on `*.test.*` filenames I think.

apps/hasura.planx.uk/tests/team_invoice_details.js
```diff
@@ -0,0 +1,110 @@
...
+    });
+
+    test("can query team_invoice_details", () => {
+      expect(i.queries).toContain("team_invoice_details");
+    });
+
+    test("cannot create, update, or delete team_invoice_details", () => {
+      expect(i).toHaveNoMutationsFor("team_invoice_details");
+    });
+  });
```
DafyddLlyr, line 109: This looks like it does not match the metadata file - there's no permissions granted for the `api` role.

DafyddLlyr (changes requested):
Small question about the tests!

apps/hasura.planx.uk/tests/team_invoice_details.test.js
zz-hh-aa: Thanks for your keen eye 👁️ ! Fixed here and updated API role tests too (I guess we'll probably need access later on but figured we could follow pattern of making permissions as granular as possible and opening them up as-and-when needed?) [4df27d2](https://github.com/theopensystemslab/planx-new/pull/7416/commits/4df27d2fa658a4134c702338fe57a6efed347643)

I also caught a `teamAdmin` mistake while updating these tests--I don't think I fully removed the inherited roles when we changed permissions models. PR incoming...

DafyddLlyr (approved):
Great...!

PR #7418, opened 2026-09-13
feat(api): Stripe webhook receiver

DafyddLlyr:
## What does this PR do?
This PR adds the `POST /stripe/webhook` endpoint to our API. After an applicants pays, Stripe calls us back by hitting this webhook. This endpoint verifies the request, `console.log()`s the event, and then acknowledges it (HTTPS 200). As with the previous PR - this is just a small slice, no persistence of this state that the frontend can verify to allow submission to proceed.

```mermaid
sequenceDiagram
    participant A as Applicant
    participant API as PlanX API
    participant S as Stripe

    Note over A,S: Start payment (checkout-session PR)
    A->>API: POST /stripe/checkout-session
    API->>S: Create Checkout Session
    S-->>API: { url }
    API-->>A: Redirect to hosted Checkout
    A->>S: Pay with test card

    Note over A,S: Webhook (this PR)
    S->>API: POST /stripe/webhook (payment_intent.succeeded)
    API->>API: Verify stripe-signature against STRIPE_WEBHOOK_SECRET
    API-->>S: 200 OK
```

### Testing
We can't currently test this on the Pizza because we need to configure each webhook in the Stripe account. As with Google Auth, we can't map wildcard URLs (e.g. `https://api.*.planx.pizza`). There's two main routes here I think...
 - No payments on Pizzas (pretty bad outcome!)
 - Setup a container running the Stripe CLI to listen for events (required for E2E, nicer setup for local)
   - This is slightly complicated by the fact that the API needs `STRIPE_WEBHOOK_SECRET` which is outputted by the CLI. I think this means we need to run this first, inject into a temp `.env`, and pass this to the API container. This is why I've swerved this here - I'll pick this up in it's own PR.

In the meantime, we can test this locally by running the Stripe CLI which will intercept requests (full details added to `README.md`).

```sh
stripe listen --forward-to localhost:7002/stripe/webhook
```

Copy the printed `whsec_…` into `STRIPE_WEBHOOK_SECRET` in `.env`, restart the API container, then trigger an event by making a payment via the `/stripe/checkout` endpoint locally. It's also possible to trigger events directly via the CLI.

Video demo talking this through posted on Slack, file size exceeds GitHub limit - https://opensystemslab.slack.com/archives/C01E3AC0C03/p1789387536898789


README.md
DafyddLlyr, line 1: This isn't great as a solution, maybe a small Docker container which handles this would be nice - which we can run with a `payments` profile (like `analytics` is used to spin up Metabase locally).

infrastructure/application/Pulumi.production.yaml
```diff
@@ -122,3 +122,5 @@ config:
   # This is a dummy placeholder value until we can switch on oAuth on prod and get the real value
   application:stripe-connect-client-id:
     secure: AAABABBT7TG68HoBqVdJHF6dbPr8FZveetwTUZpPyJ92pLYokKu8I58qrA==
+  application:stripe-webhook-secret:
```
DafyddLlyr, line 125: Dummy value, webhook endpoint not configured on production.

infrastructure/application/Pulumi.staging.yaml
```diff
@@ -108,3 +108,5 @@ config:
     secure: AAABAO7YAHmy748uzNOvl59Q18FnKBSJmjiGj2qiDuUAxGnHBxWHkXMtDxv0eEOp8jQtFRoAG+MgKWBwuS5dB+Hepw==
   application:stripe-secret-key:
     secure: AAABAGWwIOWw5cQTSVpgrmFFUCw6U/OqomMVuo8mZEiMNpyFzNY+ZHUtdY8KKektvSuEXuPvbNDSrCh4LVJDJmBA4iyTnphgrjSA1s2g9qbxeL70C7/tjlTvuJLz7ucjdK3GEC/kZJnF4sELdE6Uchn/iLpMy4F3b0jp6lGeSM8slyna7hFR1MGllA==
+  application:stripe-webhook-secret:
```
DafyddLlyr, line 111: Real value, endpoint configured on Sandbox. At dev call tomorrow it's maybe worth talking through our sandbox setup for Stripe just to make sure I have the right understanding.

apps/api.planx.uk/modules/stripe/checkout/checkout.test.ts
DafyddLlyr, line 1: Missed in previous reorganisation.

apps/api.planx.uk/modules/stripe/webhook/controller.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Handle a verified inbound Stripe webhook event
+ *
+ * Signature verification and parsing happens upstream in `verifyStripeWebhook()`
+ */
+export const handleStripeWebhook: StripeWebhookController = (_req, res) => {
+  const { stripeEvent: event } = res.locals;
+
+  switch (event.type) {
+    case "payment_intent.succeeded":
+    case "payment_intent.payment_failed":
```
joecarver, line 13: nit: do we want consts or an enum for these? / does the Stripe SDK not export anything?

apps/api.planx.uk/modules/stripe/webhook/middleware.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Verify the `stripe-signature` header and construct a Stripe.Event
+ *
+ * Docs: https://docs.stripe.com/webhooks/signature
+ */
+export const verifyStripeWebhook: RequestHandler = (req, res, next) => {
+  const webhookSecret = process.env.STRIPE_WEBHOOK_SECRET;
+  if (!webhookSecret) {
+    console.error("STRIPE_WEBHOOK_SECRET is not configured");
+    return res.status(500).send("Webhook secret not configured");
+  }
```
joecarver, line 15: for the `STRIPE_SECRET_KEY`, we check its set in the client file on build, so the api can't even start if its not set. Do we want to do the same here for the `STRIPE_WEBHOOK_SECRET` or is the per-request check enough in this case do you think?

joecarver: Ah just saw the note in the README so I guess this is intentional to allow for non-payment dev environments, makes sense then, happy to leave this!

apps/api.planx.uk/server.ts
```diff
@@ -51,7 +51,12 @@ app.set("trust proxy", 1);
 
 app.use(defaultCors);
 
-app.use(bodyParser.json({ limit: "100mb" }));
+const jsonParser = bodyParser.json({ limit: "100mb" });
+
+app.use((req, res, next) =>
+  // The Stripe webhook needs the raw, unparsed, request body for signature verification, so we skip bodyParser here
+  req.path === "/stripe/webhook" ? next() : jsonParser(req, res, next),
```
joecarver: nit: is it worth having a `const` for the path we match on here? and re-using it in the route definition. Just to make sure we don't accidentally drift, which I think would break by consuming the body too early.

apps/api.planx.uk/modules/stripe/webhook/controller.ts
```diff
@@ -0,0 +1,23 @@
...
+      // TODO: Persist payment status, don't just log
+      console.log(`Stripe event ${event.id} ${event.type}`);
+      break;
+    default:
+      console.log(`Ignoring unhandled Stripe event ${event.id} ${event.type}`);
```
joecarver, line 18: question: do we want to also handle the event `payment_intent.created` ? It could potentially be useful to update something on the payments table here so we have better traceability? I'm not sure what failure cases can occur specifically but maybe its possible for the `payment_intent.created` event to fire but then neither of the `payment_intent.succeeded` | `payment_intent.succeeded` to fire. In which case, if we don't track `payment_intent.created` then we have an incomplete record of what happened on our side.

joecarver (approved):
Great stuff, nice and clean and clear to see how it will be built on. Left a couple of tiny qs/nits.

On the pizza dev stuff, I think it's fine to leave for now but do we want to raise a ticket to make sure we address it properly?

There was a similar issue with the onboarding stuff, which needs the redirect URLs to be manually added in Stripe settings. For pizza dev I was just manually adding URLs for the pizza as I needed it. (here https://dashboard.stripe.com/acct_1UAntHEAzmRwH26v/test/settings/connect/onboarding-options/oauth)

Obviously this is not ideal but it does save adding any more architecture to solve the problem. And obviously its only gonna be certain situations in which we need to test payments on Pizzas...

joecarver:
> * Setup a container running the Stripe CLI to listen for events (required for E2E, nicer setup for local)

also with this approach, the solution sounds nice but would we also have trouble with Authentication? The Stripe CLI made me auth via the browser which was a pretty manual step. Not sure if that part can be automated too?

apps/api.planx.uk/modules/stripe/webhook/controller.ts
DafyddLlyr: This is a great question.

https://github.com/theopensystemslab/planx-new/pull/7450 is still in draft but you can see I'm handing writes as part of the checkout endpoint here - meaning that "created" means "checkout session created".

The reasoning here is two-fold - 
 - Matched GovPay convention already in place (maybe not a good thing per-se!)
 - `payment_intent.created` fires immediately as a user presses "Pay" on the checkout page, not when they navigate to the page. I'm not sure quite how useful this would be - maybe we want to capture both events?
 
I hit this when working on the above PR and when I open for review and write it up I was going to cover a few of these questions around statuses and what we want to capture when. 

I think I'm leaning towards the pure-webhook approach you're advocating for here (write on `payment_intent.created`) as it keeps a single Stripe ID (`pi_XXX`) as opposed to a mix of checkout + payment intent IDs (`cs_XXX` + `pi_XXX`).

apps/api.planx.uk/modules/stripe/webhook/middleware.ts
DafyddLlyr: Yep this is it - we don't have this value on Pizzas / E2E (we could stub something though).

DafyddLlyr:
Thanks for the review and thoughtful comments @joecarver 👍 

On the Pizza / Stripe issue I've made a ticket here to look at this specifically - https://github.com/orgs/theopensystemslab/projects/11?pane=issue&itemId=248426672

> _also with this approach, the solution sounds nice but would we also have trouble with Authentication? The Stripe CLI made me auth via the browser which was a pretty manual step. Not sure if that part can be automated too?_

Using the Stripe CLI docker image I think we can authenticate via API key - https://docs.stripe.com/cli/docker

This one is not _that_ critical - I think we can live for a little while with payments being a local/staging setup. I'm more concerned I think about E2E tests though where we face a similar issue. This may prove to be a blocker here if we want to test the full workflow.

apps/api.planx.uk/modules/stripe/webhook/controller.ts
joecarver: Yeah interesting, obviously we still need the checkout endpoint in either case so it can return the redirect. And we still need the webhook handler for the other payment updates.

So the question becomes - is there any distinction between checkout created and payment intent created - i.e. is it possible where one happens without the other, and do we need to care about that? If so we should keep both. If not, maybe best to do it all via webhook just for the sake of keeping all persistence logic in the same place... 

Kind of tricky to test and answer that question as it seems a super specific failure case which might not even exist. I guess there's no harm in tracking both events though. 

I'm happy with either approach for now, wouldn't be too tricky to extend it later on, we'd just be missing some historic data potentially.

apps/api.planx.uk/modules/stripe/webhook/controller.ts
DafyddLlyr: These are already typed as `event.type` comes from `Stripe.Event`.

image

image

apps/api.planx.uk/server.ts
DafyddLlyr: Agreed! Implemented in 581320d1f5ce909c725e4ea549e4469c35240ef6

PR #7421, opened 2026-09-14
fix: Replace minio image

DafyddLlyr:
## What's the problem?
CI is failing with the following error as of last night - 

```sh
 minio Error pull access denied for minio/minio, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
 ```

## What's the cause?
Minio [no longer offers Docker images](https://github.com/minio/minio/issues/21647#issuecomment-3418675115) and has changed its business model - so we need to [find an alternative](https://trello.com/c/PjuGigXk/3557-swap-out-deprecated-minio-image?filter=minio).

## What's the solution?
This is the quickest / laziest fix to unblock CI which is just to point to a mirror at quay.io

As per the Trello ticket linked above, we should find an alternative here..!


joecarver (approved):

freemvmt:
> As per the Trello ticket linked above, we should find an alternative here..!

Thanks for this quick fix Daf, starting work on the ticket today.

PR #7426, opened 2026-09-14
fix: correctly filter trial teams from planx in numbers stats

joecarver:
Adds a new migration to replace the `platform_stats_view` with largely the same view, just with proper filtering for teams which are on trial, using the db column `is_trial` rather than the string matching before.


apps/hasura.planx.uk/migrations/default/1789400000000_fix_platform_dashboard_stats_trial_teams/up.sql
```diff
@@ -0,0 +1,67 @@
+CREATE OR REPLACE VIEW "public"."platform_dashboard_stats" AS
+WITH lpa_teams AS (
+  SELECT t.id, t.created_at
+  FROM teams t
+  JOIN team_settings ts ON ts.team_id = t.id
+  WHERE t.name NOT IN ('Open Digital Planning', 'Open Systems Lab', 'PlanX', 'Templates', 'Testing', 'WikiHouse')
```
zz-hh-aa: Sorry one more--think you want `Council Onboarding` in here too, otherwise looks good / thanks for fix!!

joecarver: done thank you! https://github.com/theopensystemslab/planx-new/pull/7426/commits/56e753f0b9ee7b0354f7aba4f9f39ab9fd6690df

zz-hh-aa (approved):
Thanks! ✨

PR #7429, opened 2026-09-14
feat: team_trial_history table

joecarver:
Similar to how we track flow status changes, this table tracks changes to the `team.is_trial` column, allowing us to surface events such as 'Team X joined PlanX' in the Activity panel on Explore page.


jessicamcinchak:
Larger question: if the goal is to just surface _when_ a team joins PlanX - can we simply leverage existing `teams.created_at` timestamp for this (with `WHERE is_trial = false` condition)  rather than directly tracking `is_trial` changes (since as discussed in thread we don't anticipate ever really surfacing this or frequently/if-ever changing _back on_ once off)?

joecarver:
> Larger question: if the goal is to just surface _when_ a team joins PlanX - can we simply leverage existing `teams.created_at` timestamp for this (with `WHERE is_trial = false` condition) rather than directly tracking `is_trial` changes (since as discussed in thread we don't anticipate ever really surfacing this or frequently/if-ever changing _back on_ once off)?

@jessicamcinchak 

The idea here is to capture when a team moves from trial to non-trial i.e. goes live. If we just looked at the `team.created_at` timestamp then we'd miss any teams that were initially created in trial mode (which is most of them right?)

We want to detect when a team joins PlanX in two ways:

- newly created and `is_trial = false`
- `is_trial` is flipped from `true` to `false`

This table will give us the timestamped events for the second case.

This approach means we'll also be tracking teams that go back to trial mode if that ever happens - but this won't be surfaced as activity.

Does this make sense or am I overcomplicating things?

jessicamcinchak:
Just re-read through Slack thread again around this one and I _think_ we might be overcomplicating but I also see how you got to this with latest discussion. 

I think I'm just feeling confused about: 
- "LPAs on PlanX" - in my mind this _should_ include trial teams (and only filter non-LPA teams) because they can already be doing onboarding steps, subscribed to offline templates (included in that count on search result cards etc) ? Then +/- 30 day change would only have to reference `teams.created_at` ? 
- If we don't want trial teams in this number, maybe it should be rephrased as "**Live** LPAs on PlanX" (eg team has at least one online flow rather than simply isn't trial status anymore; by validation rule flows can only be online once non-trial anyways)? I think this is both: already an existing analytics dashboard or CRM metric (?); and possible to calculate with existing tables rather than introducing new one? 

Just hesitant to introduce a new definition here ! Counting LPAs already tedious because we have so many non-council teams to blacklist!

joecarver:
> * "LPAs on PlanX" - in my mind this _should_ include trial teams (and only filter non-LPA teams) because they can already be doing onboarding steps, subscribed to offline templates (included in that count on search result cards etc) ? Then +/- 30 day change would only have to reference `teams.created_at` ?

Yep, completely agreed on how we count LPAs, and  the 'PlanX in numbers' stats panel already uses this approach, exposed by the `platform_dashboard_stats` view (this PR https://github.com/theopensystemslab/planx-new/pull/7402 and corrected in this PR https://github.com/theopensystemslab/planx-new/pull/7426)

The table added here is here to drive the 'Activity on PlanX' panel, which shows events for teams 'joining' - where 'joining' should only be non-trial teams. So we need a way to differentiate between 'team created' and 'team goes live'. Sorry, is a bit confusing having both at once especially as the Slack thread was discussing both! Can chat about it in dev call later to make sure we're all on the same page.

---
this is coming in the next PR but in case this helps clarify the intended use, another migration which creates a view for displaying these activity events will look like:
```sql
CREATE OR REPLACE VIEW "public"."platform_activity_feed" AS
WITH team_first_active AS (
    -- either when a team created with is_trial=false or has is_trial set to false later
    SELECT team_id, MIN(event_start) AS event_start
    FROM team_trial_history
    WHERE is_trial = false
    GROUP BY team_id
),
team_joined AS (
    -- "team joined PlanX" events, timestamped with when they went non-trial
    SELECT
        'team_joined:' || t.id::text AS id,
        'team_joined' AS type,
        tfa.event_start AS event_time,
        t.name AS team_name,
        NULL::text AS flow_name
    FROM teams t
    JOIN team_settings ts ON ts.team_id = t.id
    JOIN team_first_active tfa ON tfa.team_id = t.id
    WHERE ts.is_trial IS NOT TRUE
      AND t.name <> ALL (
        ARRAY ['Open Digital Planning'::text, 'Open Systems Lab'::text, 'PlanX'::text, 'Templates'::text, 'Testing'::text, 'WikiHouse'::text]
      )
),
service_online AS (
    -- "service set online" events, timestamped with when it first went online
    SELECT
        'service_online:' || f.id::text AS id,
        'service_online' AS type,
        flow_first_online_at(f.*) AS event_time,
        t.name AS team_name,
        f.name AS flow_name
    FROM flows f
    JOIN teams t ON t.id = f.team_id
    WHERE f.is_service IS TRUE
      AND f.archived_at IS NULL
      AND flow_first_online_at(f.*) IS NOT NULL
      AND t.name <> ALL (
        ARRAY ['Open Digital Planning'::text, 'Open Systems Lab'::text, 'PlanX'::text, 'Templates'::text, 'Testing'::text, 'WikiHouse'::text]
      )
)
SELECT * FROM team_joined
UNION ALL
SELECT * FROM service_online
ORDER BY event_time DESC;
```

joecarver:
Closing after discussion in dev call, in which we agreed that 'team joined planX' events can also include trial teams

PR #7430, opened 2026-09-14
feat: Editor nav menu refresh

ianjon3s:
## What does this PR do

Refresh of editor nav menu to address the following:
- Removal of dark "PlanX" header
- Introduction of PlanX dot branding
- Branded dot carries over to favicon
- ~~Dedicated style for `Explore & search` menu item~~
- Team menu items grouped with `TeamSelect`
- When permission is `TeamView`, introduce `VIEW ONLY` chip

**Team editor vs team viewer:**
image


apps/editor.planx.uk/src/components/EditorNavMenu/components/NavMenuHeader.tsx
```diff
@@ -2,21 +2,42 @@ import Box from "@mui/material/Box";
...
+  alignItems: "center",
+  justifyContent: "center",
+  gap: theme.spacing(0.75),
+  padding: theme.spacing(0.5, 0.9),
+  lineHeight: 0.5,
+  fontWeight: FONT_WEIGHT_SEMI_BOLD,
 }));
 
+const LogoIcon = styled(Box)<{ teamcolour?: string }>(
+  ({ theme, teamcolour }) => ({
```
DafyddLlyr: Do we need a `shouldForwardProp` here? (not ran this locally to check if an alert is thrown)

apps/editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
```diff
@@ -227,6 +231,12 @@ function EditorNavMenu() {
           },
         ],
       },
+    ],
+    [teamSlug, lpsBaseUrl, referenceCode, teamAnalyticsLink],
+  );
+
+  const teamDocumentationSections: MenuSection[] = useMemo(
```
DafyddLlyr: nit: As we're using the React Compiler we no longer need `useMemo()` unless there's a specific behaviour you're targeting here or a rendering issue you're trying to fix. Same applies to `teamGroupedSections` above.

zz-hh-aa (approved):
Works great, tested with + without `platformAdmin` privileges, with + without `EXPLORE` feature flag, as `teamEditor` and `teamViewer`.

apps/editor.planx.uk/src/utils/favicon.ts
DafyddLlyr, line 1: This is also new to me (I've not used it yet), but React 19 now hoists any document metadata, where ever it is in the component tree to the document head. I think this means we just need to drop a `<link/>` somewhere. 

This approach may not work here with a dynamic favicon but it's worth a look at because it's more React-idiomatic and potentially simpler.

Docs: https://react.dev/reference/react-dom/components/link#special-rendering-behavior

DafyddLlyr (approved):
Looks great, and all working as expected on Pizza.

A few nits and suggestions below, no showstoppers!

apps/editor.planx.uk/src/components/EditorNavMenu/EditorNavMenu.tsx
ianjon3s: Thanks for this, I've removed the wrapper in https://github.com/theopensystemslab/planx-new/pull/7430/changes/51602a888ecb5af0766172a9951a2cfd71317d5c

apps/editor.planx.uk/src/components/EditorNavMenu/components/NavMenuHeader.tsx
ianjon3s: Good catch, added in https://github.com/theopensystemslab/planx-new/pull/7430/changes/34912e091874178bd1f775a77167b1f07089d6ed

apps/editor.planx.uk/src/utils/favicon.ts
ianjon3s: I've looked further into this, as you suspected, any reference of a favicon in `<link>` will be loaded in on first mount and would prevent any dynamic manipulation, as we are doing here

DafyddLlyr: Thanks for cross-checking this!

DafyddLlyr (approved):

PR #7432, opened 2026-09-14
feat: add autocomplete to SetValue field values

augustlindemer:
Does intentionally **not** include planning data values. This component is used for manipulating data rather than checking against it like questions or checklists do. This is routinely used for things like application types, property types, and project types, for which we want to suggest values from the schema. Planning data however should not routinely be manipulated in this way.


augustlindemer:
[Set value component to suggest values like other components](https://trello.com/c/Bs9TO20I/3764-set-value-component-to-suggest-values-like-other-components)

jessicamcinchak (approved):
Works great, nice extension! 

Planning constraints _will_ still autocomplete if naturally part of the `flowSchema` which I think is okay and still within your described intention, right? Eg add a `SetValue` here https://7432.planx.pizza/app/planx-academy/201-geospatial-integrations

Otherwise fully agree not a case we should support/encourage!

augustlindemer:
Yeah I think that's expected, current `flowSchema` values should consistently show irrespective of what they are.

PR #7436, opened 2026-09-15
fix: drop `teamAdmin` inherited role

zz-hh-aa:
While fixing test issues in #7416 I realised the `inherited_roles.yaml` file still existed, which should have been deleted in #6840. Must have missed it, my bad!

This PR:
- Deletes the `inherited_roles` YAML 
    - This is because we decided not to use inherited roles here because custom checks basically nullified the benefits of inherited roles) and instead switched to a model where all `teamAdmin`s have `teamEditor` role by default, and we manually assign `x-hasura-role` based on the operation
- Fixes a missing `select` permission for `teamAdmin`
- Updates all tests accordingly

I have tested:
- viewing team members page as `teamEditor` and `teamAdmin`
- creating and updating team members as `teamAdmin`
- general flow operations as `teamAdmin`, verifying via JWT that requests are still made with default `teamEditor` role


DafyddLlyr (approved):
Good spot - thanks for coming back and tidying this up.

I've also triggered regression tests on this branch (out of an abundance of caution!) - please feel free to merge to `main` once these pass ✅

PR #7438, opened 2026-09-15
chore: add team_invoice_details to sync

zz-hh-aa:
Will wait till #7416 merges to prod. 

Adds `team_invoice_details` to sync scripts for Stripe testing ease, as per [Jess's comment](https://github.com/theopensystemslab/planx-new/pull/7416#pullrequestreview-5202911676)


scripts/seed-database/container.sh
```diff
@@ -33,6 +33,7 @@ tables=(
   # feedback
   # flow_comments
   # lowcal_sessions
+  # team_invoice_details
```
zz-hh-aa, line 36: Left it commented out for now since not all of us will need it

DafyddLlyr: We may need to revisit this if we end up with dependencies on these fields (e.g. e2e tests, pizza manual testing) but no issues with this for now!

DafyddLlyr (approved):

PR #5931, opened 2025-12-18
chore: re-introduce Slack notification on flow status change

jessicamcinchak:
I think this was accidently dropped somewhere in the midst of refactoring Settings pages and data fetching methods (last notification was Oct 15)!

See https://opensystemslab.slack.com/archives/C01E3AC0C03/p1765995536685979 

Original implementation was #3535

PR #7445, opened 2026-09-15
feat: add copy permission Slack notification

augustlindemer:
This PR introduces a Slack notification for changes to a flow's copy permission settings, following the approach in #5931.


joecarver (approved):
nice :) 

the CI failure here is unrelated - exists upstream on main as noted here https://opensystemslab.slack.com/archives/C01E3AC0C03/p1789545862137589

PR #7446, opened 2026-09-16
feat: Activity on PlanX panel for Explore page

joecarver:
Adds the ActivityPanel to explore page:

- new db migration + metadata for `platform_activity_feed` view
  - this currently uses the new (unmerged) `team_trial_history` table to give us timestamped data of when teams go 'live' i.e. exit trial mode. There is some ongoing discussion around the approach in this PR https://github.com/theopensystemslab/planx-new/pull/7429
- frontend query bits - hook and gql snippets
- new `ActivityWidget` component with Storybook entry
- Also refactored `Dot` out of `NotificationCard` so it can be re-used here 

A couple of questions:

- Currently the 'Team X joined PlanX' has no subtitle - do we want one?
- Currently the rows are variable height - do we want to fix the height to make it consistent? The height varies in two ways atm:
  - whether the entry has a subtitle
  - the width of the panel pushing the title/subtitle onto multiple lines


apps/hasura.planx.uk/migrations/default/1789100000000_create_platform_activity_feed_view/up.sql
```diff
@@ -0,0 +1,51 @@
...
+        t.slug AS team_slug,
+        f.name AS flow_name,
+        f.slug AS flow_slug
+    FROM flows f
+    JOIN teams t ON t.id = f.team_id
+    WHERE f.is_service IS TRUE
+      AND f.archived_at IS NULL
+      AND flow_first_online_at(f.*) IS NOT NULL
+      AND t.name <> ALL (
+        ARRAY ['Open Digital Planning'::text, 'Open Systems Lab'::text, 'PlanX'::text, 'Templates'::text, 'Testing'::text, 'WikiHouse'::text]
```
zz-hh-aa: Sorry, annoying I know--this is supposed to filter out all non-LPAs, right? There's a few more...

Council Onboarding, Environment Agency, Historic England, Lambeth Sandbox, PlanX Academy, Planning Advisory Service (PAS), Strategic and Local Plan, TPX

apps/hasura.planx.uk/metadata/databases/default/tables/public_platform_activity_feed.yaml
```diff
@@ -0,0 +1,64 @@
...
+        - id
+        - type
+        - event_time
+        - team_name
+        - team_slug
+        - flow_name
+        - flow_slug
+      filter: {}
+    comment: ""
+  - role: teamAdmin
```
zz-hh-aa: Nit related to [teamAdmin permissions](https://github.com/theopensystemslab/planx-new/pull/7436): the role doesn't really need permissions on anything except for `users` and `team_members` tables 'cause it's [never returned as a default role](https://github.com/theopensystemslab/planx-new/blob/e22987c9d151f3c122199b479d68b7faec9e5788/apps/api.planx.uk/modules/auth/service/jwt.ts#L51)

(Maybe we should improve docs for this somewhere?)

apps/hasura.planx.uk/migrations/default/1789100000000_create_platform_activity_feed_view/up.sql
zz-hh-aa: Wondering also if we need a better way of surfacing when a team isn't an LPA--seems like there are a couple different allow lists to maintain now? Maybe a db column... 

Otherwise it seems a bit convoluted for whoever's making the team to have to remember that they need to update these lists, or we'll mistakenly show non-LPA activity then have to revert?

zz-hh-aa (approved):
Looks / works great! Thanks for `Dot` refactor also. I've left a couple minor comments & will defer design questions (maybe to @ianjon3s ?)

apps/hasura.planx.uk/migrations/default/1789100000000_create_platform_activity_feed_view/up.sql
joecarver: thank you! and yeah agreed... seems to keep coming up. A boolean column on `teams` seems like a good solution - be good to hear from others to see if there's any reason why we cant do this.

ianjon3s: We do handle some filtering in the app to define system teams (currently used for routing to the dashboard):

https://github.com/theopensystemslab/planx-new/blob/main/apps/editor.planx.uk/src/lib/systemTeams.ts

joecarver: I've updated the list in the query for this PR now but a more robust DB-side solution sounds wise, can pick that up separately here.

apps/hasura.planx.uk/metadata/databases/default/tables/public_platform_activity_feed.yaml
joecarver: done!

apps/hasura.planx.uk/migrations/default/1789100000000_create_platform_activity_feed_view/up.sql
joecarver: PR for the is_lpa boolean here https://github.com/theopensystemslab/planx-new/pull/7511 

have rebased this branch onto the is_lpa branch, so hopefully can get them merged in that order!

PR #7452, opened 2026-09-16
Release Stripe Onboarding

joecarver:
- Removes the feature flag check from the payment settings route, exposing the Stripe onboarding section
- Adds the feature flag round the `Provider` render, hiding the GovPay->Stripe switch
  - Q: do we want to still display this Provider section but just disable the switching? Might make it more clear for councils that although they have onboarded Stripe, they are not using it yet...


DafyddLlyr (approved):
I think we shouldn't unflag the `<Provider/>` block as it's currently quite misleading - the buttons do nothing but they can look like a "success" state. I think there's too much of a risk somebody hits "Migrate to Stripe" instead of "Connect to Stripe". 

Once the `<Provider/>` component is properly wired up we should disable if not already connected which will help this.

Image

joecarver:
> I think there's too much of a risk somebody hits "Migrate to Stripe" instead of "Connect to Stripe".

I guess the idea is 'migrate to stripe' would be not even visible for now - the provider section would just show 'current payment provider: GovPay' and a message saying that switching to Stripe is not yet ready... just worried that someone would think that 'connecting' stripe means its already switched over.

zz-hh-aa:
Sorry just merged the invoice details settings forms, so you'll have to rebase (but [those can be un-feature-flagged](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1789572242180389?thread_ts=1789571981.462029&cid=C01E3AC0C03))

jessicamcinchak:
Agree that displaying "Provider" section _without_ toggle (or fully disabled toggle so success messages not possible) for context seems like good instinct here! We don't want folks to think "connecting" their account = switching which seems very plausible to me if _only_ first section is shown :heavy_plus_sign:

DafyddLlyr:
Yep - missed this when considering what to show - it's not a simple binary of show/hide the whole provider block! I think as long as the button is hidden we're good.

joecarver:
Screenshot 2026-09-16 at 16 54 11

cool, just pushed to give this UI ^

Also just noticed the copy in the connect section actually says 'Once connected, Stripe will handle payment processing for all application fees collected through your services.' so will remove that section as its kind of meaningless (and re-iterated more clearly in the following section)

DafyddLlyr (approved):
Perfect - thank you!

PR #7453, opened 2026-09-16
fix(a11y): Identify warning containers as labelled region

ianjon3s:
## Issue

Relates to: p.15 of retest document - DAC_Information_and_Relationships_01

- `WarningContainer` does not identify as a region with identifiable label

## Solution

- Output as `<section>` with assigned `aria-labelledby` pointing to contained text
- Additionally, prepend screenreader only text `Warning:` to replace visual icon

### Testing

Select "invite to pay" to see warning container flagged in report:
https://7453.planx.pizza/testing/warning-container-test/preview


jessicamcinchak (approved):
All makes sense, thanks for dilligently catching all instances of this one - even those out of scope of audit !

DafyddLlyr (approved):

ianjon3s:
Thanks @jessicamcinchak , there may be a better way to programatically label these, but to be safe I've applied to all instances of `<WarningContainer>`

PR #7454, opened 2026-09-16
fix(a11y): Button labels for file name unknown

ianjon3s:
## Issue

Relates to: p.17 of retest doc - DAC_Headings_and_Labels_03

Files with name unknown are output as `undefined`, meaning buttons relating to that file do not make sense.

I can replicated this by navigating/continuing on from Upload & label and then back to the component, see here:
https://editor.planx.dev/testing/upload-file-name/preview

## Solution

Belt and braces approach:
- Ensure file name is included in cached slot
- Output suggest (by report) text if file name cannot be loaded

### Testing

https://7454.planx.pizza/testing/upload-file-name/preview


apps/editor.planx.uk/src/@planx/components/FileUploadAndLabel/model.ts
```diff
@@ -71,7 +71,7 @@ export interface FormattedUserFile {
    * This allows us to restore the UI on refresh / resume / back
    */
   cachedSlot: Omit<FileUploadAndLabelSlot, "file"> & {
-    file: Pick<FileWithPath, "path" | "type" | "size">;
+    file: Pick<FileWithPath, "name" | "path" | "type" | "size">;
```
jessicamcinchak, line 74: 👍

jessicamcinchak (approved):
Good pragmatic solution here!

DafyddLlyr (commented):
Great!

PR #7455, opened 2026-09-16
[DX] hook up workspace package builds

freemvmt:
Integrating an in-workspace package (e.g. `./packages/file-upload`) that needs building/compiling has caused some pain for devs, e.g. when stopping and starting the Docker stack after pulling latest `main`. See [Slack thread](https://opensystemslab.slack.com/archives/C01E3AC0C03/p1787495642146919).

To smooth out these issues, here we do the following:
- add a `build-packages` script and run it on `pnpm install`, `pnpm start`, `pnpm dev`
- also hook it in to git cycle via husky's `post-merge` (should handle pulls from `main`)
- bind-mount `file-upload/` into API dev container's node packages so that re-builds are picked up
- further bind-mount all packages `dist/`'s and include in `tsx watch` so that API container also restarts on re-builds
- add a `watch:packages` script at root so that anyone developing `file-upload` (or another workspace dependency) has a method to reflect those edits throughout the stack on save

We also extend the `file-upload` README to explain all this, and leave various comments throughout.


packages/file-upload/README.md
```diff
@@ -4,8 +4,25 @@ Exports constants related to file uploads, for consumption by the API and Editor
...
+
+That copy does not track changes to this package — `docker-compose.local.yml` therefore bind-mounts this package over that copy, so the dev container always reads the host's `dist/`.
+
+## Editing this package
+
+Consumers read `dist/`, never the TS source, so an edit here is invisible everywhere until the package is rebuilt. Nothing rebuilds it _on save_ — the hooks above all fire at specific junctures (e.g. `pnpm i`, `git pull`, `pnpm dev`).
+
+If you do want edits to reflect live in `dist/` on save, run `pnpm watch:packages` from root in a second shell. Once `dist/` is rebuilt, the two consumers acknowledge it by different methods:
+
+- The **Editor** picks it up picks it up thanks to Vite, which resolves this package to its real path outside `node_modules` and serves it as source, so will be subject to HMR (hot module replacement) without dev-server restart.
```
joecarver:
```suggestion
- The **Editor** picks it up thanks to Vite, which resolves this package to its real path outside `node_modules` and serves it as source, so will be subject to HMR (hot module replacement) without dev-server restart.
```

package.json
```diff
@@ -12,17 +12,20 @@
     "clean-data": "docker compose -f ./docker-compose.yml -f ./docker-compose.seed.yml run seed-database reset_all",
     "dev": "./scripts/start-containers-for-dev.sh",
     "tests": "./scripts/start-containers-for-tests.sh",
+    "prestart": "./scripts/build-packages.sh",
```
joecarver: the package commands `recreate` and `analytics` also start the stack of docker containers. do we want matching `prerecreate` and `preanalytics` hooks as well?

joecarver (approved):
Great stuff, really thorough fixes. Tested locally with a few weird setups and seems to be reliably starting every time for me. Thanks!

package.json
freemvmt: great catch, will add these in!

freemvmt:
> Great stuff, really thorough fixes. Tested locally with a few weird setups and seems to be reliably starting every time for me. Thanks!

fab @joecarver thanks for putting this stuff through its paces!

PR #7464, opened 2026-09-17
feat(api): Get checkout status endpoint

DafyddLlyr:
## What does this PR do?
This adds a new endpoint so that on return to PlanX, the `Pay` component can confirm a Stripe payments status - allowing the user to proceed to `Send`.

On return from Steipe the redirect itself can't be trusted to confirm payment, so as per [Stripe's recommended pattern](https://docs.stripe.com/checkout/fulfillment) the return page asks our API to retrieve the Checkout Session and read its status, rather than trusting the URL alone. 

The webhook remains should remain as the source of truth for fulfilment, this is the synchronous check for the returning applicant and will work for all card payments (all we're allowing for now whilst still supporting GovPay).


jessicamcinchak (approved):
Great! Very straightforward

PR #7466, opened 2026-09-17
feat: Setup `team_setting.payment_provider`, configure `<Provider/>` to read field

DafyddLlyr:
## What does this PR do?
 - Adds `team_setting.payment_provider` column which can hold "stripe", "govpay" or `null`
   - Column populated by checking `team_integrations` values - if a team has GovPay keys, they're using GovPay
 - Relies on https://github.com/theopensystemslab/planx-core/pull/1072

## Context
Please see https://opensystemslab.slack.com/archives/C01E3AC0C03/p1789643457725249?thread_ts=1789571981.462029&cid=C01E3AC0C03 (OSL Slack)

Now that the `<Provider/>` column is shown we can't just use GovPay as a fallback (default) for all teams - we need to account for the three possible states.

This column is intentionally not synced from production as it's likely teams will want to test on staging and persist over a few days, instead of having to constantly toggle this.


apps/editor.planx.uk/src/pages/FlowEditor/components/Settings/Team/Payment/Provider/index.tsx
```diff
@@ -143,15 +148,23 @@ const Provider: React.FC = () => {
...
+                    label={PROVIDER_LABELS[provider]}
+                    color="info"
+                    size="small"
+                  />
+                </Box>
+                {renderProviderAction()}
+              </>
+            ) : (
+              <Typography variant="body2" sx={{ color: "text.secondary" }}>
+                No payment provider connected to your account
```
jessicamcinchak: nit: Because you may be able to _onboard_  / _connect_ your Stripe account before it's actually "active", can this simply say: "No active payment provider" ?

apps/hasura.planx.uk/migrations/default/1789644819612_alter_table_public_team_settings_add_payment_provider/up.sql
```diff
@@ -0,0 +1,13 @@
...
+  add constraint "team_settings_payment_provider_check"
+  check ("payment_provider" in ('govpay', 'stripe'));
+
+comment on column "public"."team_settings"."payment_provider" is E'Payment provider used to route applicant payments: govpay, stripe, or null.';
+
+update "public"."team_settings" ts
+set "payment_provider" = 'govpay'
+from "public"."team_integrations" ti
+where ts."team_id" = ti."team_id"
+and ti."production_govpay_secret" is not null;
```
jessicamcinchak: We want to support _unique_ providers on staging and prod, but I don't think this migration will correctly populate default govpay when it runs on staging if relying on this `production_` column check?

Maybe simple solution is to run `UPDATE` step only manually on each env rather than via Hasura?

DafyddLlyr: Nice spot - thank you.

This has now been dropped from the `up.sql` and I can run the following manually once merged (additional `OR` check for `staging_` column).

```sql
update "public"."team_settings" ts
set "payment_provider" = 'govpay'
from "public"."team_integrations" ti
where ts."team_id" = ti."team_id"
  and (
    ti."staging_govpay_secret" is not null
    or ti."production_govpay_secret" is not null
  );
  ```

jessicamcinchak (approved):

apps/hasura.planx.uk/migrations/default/1789644819612_alter_table_public_team_settings_add_payment_provider/up.sql
jessicamcinchak: Perfect !

PR #7471, opened 2026-09-17
test(editor): Drop hardcoded dates, use relative dates

DafyddLlyr:
## What's the problem?
Test in `SubmissionDetailModal.test.tsx` are failing across a number of PRs without having been directly modified.

## What's the casue?
We've gone past the 28 day expiry period of the hardcoded `created_at` dates in the mocks used here. This causes the tests to fail as they rely on these dates to dynamically set the status.

## What's the solution?
Use dates relative to `new Date()` (i.e. the timestamp the tests run from), and preserve the sequence of events by appending a minute to each one from this base date.


zz-hh-aa (approved):

jessicamcinchak (approved):
Smart ! Thank you!

PR #7474, opened 2026-09-17
feat: add local authority district exception to customisable node options lock

augustlindemer:
Workspaces covering multiple councils need to customise the authority check in our source templates. This PR adds an exception to questions sorting on `property.localAuthorityDistrict` that allows non-platform admins to add, delete, and reorder options.

If we ever learn of more edge cases that require graph structure changing edits in templates, we can turn this into a list of values instead.


jessicamcinchak (approved):
Simple + pragmatic solution!

PR #7477, opened 2026-09-17
fix: sync `team_settings.payment_provider`

jessicamcinchak:
Data sync currently failing !


DafyddLlyr (approved):
Approved to unblock, but I think we want to not overwrite this value - teams will legitimately want to have one provider on staging, but another on production, whilst we're migrating and testing.

I missed that we do need to include it though if we're going a `SELECT *` here. `published_flows` and `team_integrations` list a subset of columns for the sync - maybe this is the pattern we need to follow here.

jessicamcinchak:
Closing in favor of #7479

PR #7478, opened 2026-09-17
chore: add FME, Stripe and service charge details to admin panel

jessicamcinchak:
Per https://trello.com/c/R13JkYC3/3758-show-fme-stripe-connect-connected-pay-provider-in-admin-panel


apps/editor.planx.uk/src/types.ts
```diff
@@ -117,6 +117,10 @@ export interface AdminPanelData {
   actionColour?: string;
   liveFlows: LiveFlow[] | null;
   isTrial: boolean;
+  fmeEnabled: boolean;
+  stripeConnected?: string;
+  paymentProvider?: string;
+  serviceCharges?: number; // pg SUM() returns null instead of 0
```
DafyddLlyr, line 123: This _might_ trip a type issue somewhere down the line as `null` does not conform to type `undefined`. 

Could we `COALESCE(SUM(column_name), 0)` so we always return a number?

jessicamcinchak: Ah sorry this is not a very clear comment on my part - so how I understand this: 
- PG returns number > 0 or null if no charges for a given team (we never actually care about showing sum in actual table, it's just easy agg function for lateral join here I think)
- We query `teams_summary` view via GraphQL which will always return `null` as `undefined`, so this type matches GraphQL response ! 

Not an issue adding `coalesce` here either, but I don't think the `null` should ever actually crop up outside of the raw db view values?

DafyddLlyr: Perfect - thanks for the explanation 👍

DafyddLlyr (approved):
Nice and simple - great! Approved from a code POV, appreciate there's an open `#planx-design` thread on this one.

PR #7479, opened 2026-09-18
fix(ci): Fix `team_settings` data sync

DafyddLlyr:
We're intentionally not overwriting the value of `team_settings.payment_provider` here - we want a team to continue with GovPay on prod, but be actively testing staging on Stripe over a number of days/weeks.


jessicamcinchak (approved):
Great ! Happy for this to replace #7477

PR #7481, opened 2026-09-18
feat(api): Ensure `/validate-session` endpoint accounts for Stripe payment status

DafyddLlyr:
## What does this PR do?
This PR ensures that we skip Save & Return reconciliation for Stripe applicants who have initiated payment, which copies the existing GOV.UK Pay behaviour. Without this applicants will go through a reconciliation process on return from Stripe.

Note - the "Resume" page is still showing in the video below (locally) as we don't get `document.referrer` over HTTP - same as current GovPay setup

https://github.com/user-attachments/assets/025da0a8-9ec4-4b7f-99cb-1e1c88529eda


apps/api.planx.uk/modules/saveAndReturn/service/validateSession.ts
```diff
@@ -23,6 +25,47 @@ export type ReconciledSession = {
...
+
+const STRIPE_INITIATED_STATUSES: StripePaymentStatus[] = [
+  "created",
+  "processing",
+  "succeeded",
+];
+
+/**
+ * Stripe writes to payment_status via webhook (on payment_intent.created)
+ */
```
jessicamcinchak, line 56: Helpful data model reminders here 👍

jessicamcinchak (approved):
Great ! Thanks for improving lots of test coverage here too

